PR #1952, opened 2021-02-25
Bonsai Tries: Respect account deletion across transacitons

shemnon:
## PR description

In a block if another account is deleted vs self destruct the sstore
operation could get an incorrect original value, and if the new write is
the same as the value before delete a SLOAD cost will mistakenly be
applied.  Found in mainnet block 11905274 for account
0x7f150bd6f54c40a34d7c3d5e9f56.

## Changelog


atoulme:
Looks good to me, would there be a way to have a test to back it up?

matkt (approved):
Sounds good to me.

PR #1954, opened 2021-02-25
Update Docker files due to hadolint

shemnon:
Hadolint introduced a new linting rule prohibiting using the same env
var in it's definition. Move to a two step definition.

## Changelog


atoulme (approved):
LGTM

PR #1960, opened 2021-02-28
Reduce hash calculations in Bonsai Tries

shemnon:
Setting contract code on a bonsai account results in a re-hashing. To
reduce this only set the code if the code was changed.  In some cases
this was responsible for 40% of load.

## Changelog


matkt (approved):
LGTM

PR #1933, opened 2021-02-21
Replacing the static references of SECP256K1 by an interface

daniel-iobuilders:
Signed-off-by: Daniel Lehrner <daniel@io.builders>

## PR description
At the moment Besu only supports the Ethereum standard SECP256K1 key algorithm. Due to NIST compliance, we would like to add support for SECP256R1 keys.
In this first PR all the static reference calls to SECP256K1 are replaced by an interface. This is a first step to allow different elliptic curves to be added by future PRs. 

## Fixed Issue(s)
-

## Changelog

- [ ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #1967, opened 2021-03-02
Remove hard coded elliptic curve lengths

daniel-iobuilders:
Signed-off-by: Daniel Lehrner <daniel@io.builders>

As discussed in #1933, the byte size for different variables of the elliptic curve cryptography are hard coded and don't allow the future implementation of algorithms other than SECP256*. This PR removes all these hard coded byte sizes and replaces them by dynamic ones. This allows the future implementation of algorithms like SECP384R1 (NIST P-384).
This is a simple refactoring, no existing functionality was modified or removed, neither was a new one added.


crypto/src/main/java/org/hyperledger/besu/crypto/SECPSignature.java
```diff
@@ -88,28 +97,41 @@ private static void checkInBounds(
...
+    final int signatureLength = calculateSignatureLength(privateKeyLength);
     checkArgument(
-        bytes.size() == BYTES_REQUIRED, "encoded SECP256K1 signature must be 65 bytes long");
-
-    final BigInteger r = bytes.slice(0, 32).toUnsignedBigInteger();
-    final BigInteger s = bytes.slice(32, 32).toUnsignedBigInteger();
-    final byte recId = bytes.get(64);
-    return SECPSignature.create(r, s, recId, curveOrder);
+        bytes.size() == signatureLength,
+        new StringBuilder()
```
lucassaldanha, line 105: We are missing the spaces between the appends, aren't we?

PR #1973, opened 2021-03-03
Allow to disable ErrorProne

hmijail:
## PR description
Simplifies javac calls to allow the build to be run by other analyzers.

Example:
```
./gradlew installDist -Davt.disableErrorProne=true
```

Signed-off-by: Horacio Mijail Anton Quiles <hmijail@gmail.com>

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -188,6 +188,10 @@ allprojects {
     ]
 
     options.errorprone {
+      if ('true'.equalsIgnoreCase(System.getProperty('avt.disableErrorProne'))) {
+        enabled = false
+      }
+
       excludedPaths = '.*/(generated/*.*|.*ReferenceTest_.*)'
 
       // Our equals need to be symmetric, this checker doesn't respect that.


RatanRSur (approved):

shemnon (changes requested):
I'm concerned that adding backdoors like this will result in the checks being stripped.   These are essential for our security posture.

Can a similar counter-check be put in place to prevent distributions being created if the flag is set?

hmijail:
Should be doable I guess? I’ll check. 

In any case, can you give an example of what threat scenario you have in mind? Someone who builds their own distribution with custom flags surely knows what they’re doing and/or can do much worse than disabling ErrorProne anyway.

shemnon:
I think the burden of proof actually lies on why we actually need to put this back door in the main repo in the first place. Can this analysis be done from a forked repo kept in sync?  

Also, can you point me to where this analysis is being done?

hmijail:
I'm confused. I sent you days ago the report about using Infer on our codebases, asking specifically for your feedback. If you couldn't read it then, I hope it will be published at some point outside of ConsenSys, at least in part.

I didn't ask for proof of anything, but for the threat scenario you have in mind, because I don't get it. 
Personally, I fail to see how disabling EP through a flag turns into a "back door". It's much riskier to allow to build without tests, but that is what happens when one uses `installDist`, which is what the official docs recommend.
And if EP is essential for our security posture, I would imagine that reducing friction to run more powerful analyzers would be a welcome first step to improve security.
So please let me know what kind of proof should I provide, and of what exactly.

Finally: of course, everything can be done from a forked repo. But again, my goal was precisely to remove that friction.

shemnon:
You were able to generate those reports without any change to the main repository.  Can the process you did continue to be the same process for the reports?

hmijail:
My report is finished, this is not about that. And of course I can do everything without touching the repo. But the purpose of the PR is to improve the build system so it can be used with other analyzers in the future, without having to re-discover and re-fix the problem.
If this is not considered an improvement, then the PR makes no sense and can be discarded.

I am not sure how to address the "conditionals in the main build path". Do you mean that you would prefer a `installDistWithoutErrorProne` task without extra flags?

shemnon:
An entirely separate target would be better.  I don't want to see the enable property in the global configuration toggled by an external property.

I would also be interested in documentation explaining the mechanics of how this is a problem for the static analyzer you are using in your report and why this solves it (and argument why it's the correct solution). That would be of more use to future static analysis efforts than an arbitrary property with no documentation. It would also be of use to the errorprone team in case they were interested in modifying errorprone so it can co-exist with more analyzers.

hmijail:
The root of the problem is that ErrorProne is run by adding to the javac call [a string of tokens separated by spaces](https://errorprone.info/docs/installation#command-line).
Infer, like other analyzers, captures the compiler call to do its job. 
This capture breaks the string of tokens into the separate tokens, and the resulting reconstructed javac call is wrong. 

The bug is in the interaction between Gradle and Infer, and I already reported it to Infer long ago. But no idea of how prioritary it is for them to fix it, if it is fixable at all. (I will refresh their bug report with a new example anyway)

Regarding the "arbitrary property with no documentation", there are currently other 5 properties to which build.gradle reacts, and are only documented with a few comment lines. Not sure if they are "arbitrary". Should I add a comment to `avt.disableErrorProne`?

shemnon:
Evidence of past misbehavior is not justification for bad practice.  If we hold that standard we will never improve the codebase.

Don't document avt.disableErorrProne.  Find another solution, that was my original request.

Issue #1120, opened 2020-06-22
Use 'Transitions' in the genesis file to change Mining Beneficiary and/or Block reward in IBFT2


rain-on:
Assumes network-level block reward and mining beneficiary has already been complete for IBFT2.
===

Extend the transitions section of the Genesis file to allow the mining beneficiary and/or the block reward to be changed at any block.

This will mean the "Transitions" will need to be applied to an existing ProtocolSchedule by:
* Inserting a new milestone at the specified block (but otherwise identical to the preceding milestone), and * modify the existing behaviour of all latter milestones with the change specified in the transition.

vmichalik:
Closed by #1977

PR #1962, opened 2021-03-01
Allow for switching ProtocolSchedule rules outside of Milestones

rain-on:
This is the first step in supporting switchable consensus mechanisms.

Specifically this allows additional protocol specs to be inserted to the protocol schedule at milestones other than that explicitly specified in the genesis config.

## Changelog

- [ ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #1977, opened 2021-03-03
Implement transitions for IBFT2 block rewards

VladLupashevskyi:
## PR description

Here I implemented the transitions for block rewards on IBFT2 (#1120).

Could you please have a look at it and tell me whether the approach I've chosen is correct? There are no tests yet, I would appreciate if somebody could give me directions in the code for tests implementation.

I assume it should be enough if I create spec with following piece and check the balances of sealer on different blocks:

```
"transitions": {
        "ibft2": [
        {
          "block": 10,
          "blockreward": "1"
        },
        {
          "block": 15,
          "blockreward": "5"
        },
        {
          "block": 20,
          "blockreward": "0"
        }
        ]
      }
```


rain-on:
It looks really nice, but ;) (there's always a but).
If you wait for #1962 to go in, you may find it way easier to make this change

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -134,13 +135,13 @@ public boolean isSuccessful() {
   protected AbstractBlockProcessor(
       final MainnetTransactionProcessor transactionProcessor,
       final TransactionReceiptFactory transactionReceiptFactory,
-      final Wei blockReward,
+      final Function<Long, Wei> blockRewardCalculator,
```
rain-on, line 138: hoping we don't need a calculator - but rather, we can ensure the blockProcessor at each Milestone knows what reward to expect.

rain-on:
It was easier to throw up an example of what i was thinking as done in #1979 - it has a few problems due to how Transitions work between IBFT2 and QBFT (but we'll resolve that in another PR).

But basically, there _should_ be no need to modify BlockProcessors etc - instead, we create 'interim' Milestones at the blocks defined in the Transitions - the behaviour of these new milestones is the same as the previous milestone, but alters the Blockreward.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
VladLupashevskyi: Yeah... just passed scheduler to block processor

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/RewardTraceGeneratorTest.java
```diff
@@ -90,7 +90,7 @@ public void assertThatTraceGeneratorReturnValidRewardsForMainnetBlockProcessor()
         new MainnetBlockProcessor(
             transactionProcessor,
             transactionReceiptFactory,
-            blockReward,
+            protocolSchedule,
```
rain-on, line 93: The ProtocolSchedule is effectively a list of ProtocolSpecs - each of which defines the behaviour for each block in a milestone.
As such - Besu creates a _new_ MainnetBlockProcessor for each milestone, each potentially with a different blockReward.
Thus, the goal is for a ProtocolSchedule to have a number of BlockProcessors, rather than a single BlockProcessor having access to the schedule.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/goquorum/GoQuorumBlockProcessor.java
```diff
@@ -61,15 +61,15 @@
   public GoQuorumBlockProcessor(
       final MainnetTransactionProcessor transactionProcessor,
       final TransactionReceiptFactory transactionReceiptFactory,
-      final Wei blockReward,
+      final ProtocolSchedule protocolSchedule,
```
rain-on, line 64: see comment above regarding relationships.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -123,7 +123,7 @@ public boolean isSuccessful() {
 
   protected final AbstractBlockProcessor.TransactionReceiptFactory transactionReceiptFactory;
 
-  final Wei blockReward;
+  final ProtocolSchedule protocolSchedule;
```
rain-on, line 126: see above comment regarding relationships

rain-on (commented):
Hey, I raised a comment inside the code, but will duplicate here - the interface into the AbstractBlockProcessor _should_ take the block-reward value (as per existing code), not the protocolSchedule.
This is because the ProtocolSchedule (effectively) has a list of BlockProcessors, one for each milestone - where each milestone defines the appropriate block reward.

VladLupashevskyi:
@rain-on Thanks for your comment - that definitely makes sense. I missed it somehow that BlockProcessor gets created for each ProtocolSpec. Fixed it. Now tests are left. Do you think approach described earlier would be sufficient?

As for CircleCI - I was not following my fork, but it was available in the list, because I was signed up there via GitHub. So maybe there was an issue, however it's really strange. I just deauthorized it - let's see...

VladLupashevskyi:
After I've done this - no jobs were running on CircleCI at all.
I tried to authorize again, follow and then unfollow the project, but it does not really help :(

VladLupashevskyi:
@rain-on I've rebased the branch on current master branch and cherry-picked your commit with test to here. I've also run tests locally.

CHANGELOG.md
```diff
@@ -189,6 +189,8 @@ This release contains the activation blocks for Berlin across all supported test
 * Distributions and maven artifacts have been moved off of bintray [\#1886](https://github.com/hyperledger/besu/pull/1886)
 * admin_peers json RPC response now includes the remote nodes enode URL
 * add support for keccak mining and a ecip1049_dev network [\#1882](https://github.com/hyperledger/besu/pull/1882)
+* Introduced transitions for IBFT2 block rewards [\#1977](https://github.com/hyperledger/besu/pull/1977)
```
jframe, line 192: This needs to move to 21.7.0-RC1 section now in the changelog

jframe (commented):
Just getting back to this now.
Overall looks really good, just need to move the changelog entry into the right release version section.

CHANGELOG.md
jframe: Actually since this is a trivial change, I'm happy to just move this to correct place in the changelog before I merge it.

jframe (approved):

PR #1978, opened 2021-03-04
use cloudsmith for orion

macfarla:

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -131,7 +131,6 @@ allprojects {
     maven { url "https://splunk.jfrog.io/splunk/ext-releases-local" }
     maven { url "https://dl.bintray.com/open-telemetry/maven" }
     maven { url "https://dl.cloudsmith.io/public/consensys/quorum-mainnet-launcher/maven/"}
-    maven { url "https://dl.cloudsmith.io/public/consensys/orion/maven/"}
   }
 
   dependencies { errorprone "com.google.errorprone:error_prone_core" }
--- a/build.gradle
+++ b/build.gradle
@@ -131,6 +131,7 @@ allprojects {
     maven { url "https://splunk.jfrog.io/splunk/ext-releases-local" }
     maven { url "https://dl.bintray.com/open-telemetry/maven" }
     maven { url "https://dl.cloudsmith.io/public/consensys/quorum-mainnet-launcher/maven/"}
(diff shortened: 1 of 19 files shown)


build.gradle
```diff
@@ -131,6 +131,7 @@ allprojects {
     maven { url "https://splunk.jfrog.io/splunk/ext-releases-local" }
     maven { url "https://dl.bintray.com/open-telemetry/maven" }
     maven { url "https://dl.cloudsmith.io/public/consensys/quorum-mainnet-launcher/maven/"}
+    maven { url "https://dl.cloudsmith.io/public/consensys/orion/maven/"}
```
shemnon, line 134: Why do we have a separate maven repository for every ConsenSys product?  Shouldn't there be one unified ConsenSys maven repository?

pinges (approved):
I guess we can talk about the seperate repos, but get this change merged in.

shemnon (changes requested):
This only works because the consensys bintray is still in the repository list - 

https://github.com/hyperledger/besu/pull/1978/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R130

Orion is not being pulled from the cloudsmith repos because the layout there is not maven and POMs are not being published.

build.gradle
macfarla: good question. the cloudsmith docs point to this format of url ie owner/repo/maven so I can't see how to have a unified maven repo https://help.cloudsmith.io/docs/gradle-repository#download--installing but def something we can look into

macfarla:
> This only works because the consensys bintray is still in the repository list -
> 
> https://github.com/hyperledger/besu/pull/1978/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R130
> 
> Orion is not being pulled from the cloudsmith repos because the layout there is not maven and POMs are not being published.

yep. working on it!

build.gradle
macfarla: you are right - there is a central repo. closing this PR

shemnon:
Another wrinkle... besu depends on 20.10.1 which is not currently in the cloudsmith repository.  Can the older version be hand uploaded?  Possibly all older versions?

PR #1980, opened 2021-03-04
Allow forbidding unprotected transactions

### LINKED ISSUE #1975: Disable sending of non eip155 replay protected tx by default
Geth recently implemented this change here: https://github.com/ethereum/go-ethereum/pull/22339. It would be nice if Besu would also disable sending of non EIP155 transactions by default (see reasoning in linked PR).

atoulme:

## PR description
This PR changes the way transactions are accepted over JSON-RPC and allows forbidding transactions that do not have EIP-155 provisions, ie the chainId is not provided.
This PR sets the stage to forbid unprotected transactions, but for now Besu will continue allowing unprotected transactions until the next major release.
The flag `--require-tx-replay-protection` is added, and is set to false by default.

## Fixed Issue(s)
Fixes #1975

## Changelog


atoulme:
This is now ready for review.

NicolasMassart:
@atoulme could you add a detailed description of the change for us to have all the material do document this change please?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -817,6 +817,13 @@ void setBannedNodeIds(final List<String> values) {
       arity = "1")
   private final Wei txFeeCap = DEFAULT_RPC_TX_FEE_CAP;
 
+  @Option(
+      names = {"--rpc-allow-unprotected-txs"},
+      description =
+          "Allow for unprotected (non EIP155 signed) transactions to be submitted via RPC (default: ${DEFAULT-VALUE})",
+      arity = "1")
+  private final Boolean unprotectedTransactionsAllowed = false;
```
shemnon, line 825: This breaks backwards compatibility, so it will have to wait for the next quarterly release of the default is to deny.
I would recommend adding the flag with the default to allow and then at the next quarterly release rc cycle flip the flag to deny.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -817,6 +817,13 @@ void setBannedNodeIds(final List<String> values) {
       arity = "1")
   private final Wei txFeeCap = DEFAULT_RPC_TX_FEE_CAP;
 
+  @Option(
+      names = {"--rpc-allow-unprotected-txs"},
```
shemnon, line 821: "unprotected" is a bit of a broad term.  "Replay unprotected" is also a bit of a mouthfull.

How about `--rpc-require-eip155-txs`?  or `--rpc-require-chainid-in-txs`?

Similarly I think a more descriptive flag across the code than just "unprotected."  eip-155 is more precise but a bit harder to grock when new to the code, so `requireChainIdTXs" may be more readable.

acceptance-tests/tests/src/test/resources/clique/clique.json
```diff
@@ -1,12 +1,7 @@
 {
   "config": {
     "chainId": 4,
-    "homesteadBlock": 1,
```
shemnon, line 4: These seem orthogonal to the PR scope, which is already quite large due to the AT changes.  Please move to a different PR.  Same with ibft and qbft.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -142,6 +142,9 @@ public boolean addTransactionHash(final Hash transactionHash) {
 
   public ValidationResult<TransactionInvalidReason> addLocalTransaction(
       final Transaction transaction) {
+    if (transaction.getChainId().isEmpty() && !configuration.isUnprotectedTransactionsAllowed()) {
```
shemnon, line 145: Such a large test impact for such a little 3 line change ;)

shemnon: TIL you can't do in-line emojis in github comments.

shemnon (commented):
Code looks reasonable.  Just need to iterate on the specific wording and making sure not to ship breaking changes outside of quarterly releases.

acceptance-tests/tests/src/test/resources/clique/clique.json
atoulme: I had to do this to make acceptance tests pass or the first block cannot be minted.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
atoulme: I believe I copied the flag from geth.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
atoulme: it was a bit daunting yeah.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
atoulme: Makes sense.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
shemnon: We should do better.

atoulme: lol ok no problem.

acceptance-tests/tests/src/test/resources/clique/clique.json
shemnon: I don't follow.  Acceptance tests were working before, right?  What changed to require this change?  The protocol specs should simply upgrade each block as it passes the milestone, why does it suddenly need to start at a future fork for reference tests?

atoulme: It looks like I tried to send a eip155 transaction to a frontier block, and it didn't really like that.

shemnon: My thought then is that the ATs for eip-155 transactions should wait until a certain block is mined or a separate set of AT genesis files should be used that start right after spurious monkey.

atoulme: Not to drag this thread along too much, but there is a fine point here that this change introduces, which pretty much says you'd better start your chain after a certain ethereum hard fork.

I'm not quite sure what the point of having private chains with frontier blocks is.

It would also make sense if Besu could deal with EIP-155 transactions and wait out until the hard fork has happened to introduce them. I guess the same thing will happen with EIP-1559 type transactions next, if/when it gets adopted.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
atoulme: Going with `--rpc-require-chainid-in-txs`

atoulme:
I changed the logic to keep unprotected transactions enabled for now.

acceptance-tests/tests/src/test/resources/clique/clique.json
shemnon: We still have to be able to full import mainnet and classic, even if unprotected transactions are prohibited in a future fork.  So I am still uncomfortable with the broad changes occurring to the ATs implied by these changes.  I am of the opinion a different solution should be found, such as adding `clique_berlin.json` et.al, or ensuring that the ATs are past the spurious dragon blocks before submitting EIP-155 transactions.  I am also uneasy moving wholesale to EIP-155 transactions in the ATs.  While they should be present the should not make pre-EIP-155 TX formats extinct in the AT stack.

atoulme: So you’re saying:
- tie eip-155 tx flag to the protocol schedule
- revert all genesis file changes
- Make all tests pass again

I’ll do it.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
atoulme: done

acceptance-tests/tests/src/test/resources/clique/clique.json
atoulme: Done

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -831,6 +831,14 @@ void setBannedNodeIds(final List<String> values) {
       arity = "1")
   private final Wei txFeeCap = DEFAULT_RPC_TX_FEE_CAP;
 
+  @Option(
+      names = {"--rpc-require-chainid-in-txs"},
+      description =
+          "Allow for unprotected (non EIP155 signed) transactions to be submitted via RPC (default: ${DEFAULT-VALUE})",
+      arity = "1")
+  // TODO: set default to false for next major release
+  private final Boolean unprotectedTransactionsAllowed = true;
```
mbaxter, line 840: This looks inverted to me: if `--rpc-require-chainid-in-txs=true`, then shouldn't `unprotectedTransactionsAllowed == false`?  Suggest changing this variable to `requireTxReplayProtection` or similar.  I think the flag name could be clearer as well, maybe: `--require-tx-replay-protection`.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -109,6 +116,10 @@ public MainnetTransactionValidator(
       return signatureResult;
     }
 
+    if (transaction.getChainId().isEmpty() && !acceptUnprotectedTransactions) {
```
mbaxter, line 119: We can only enforce this if there's actually a `chainId` specified.  I think this should be:
```suggestion
    if (!acceptUnprotectedTransactions && chainId.isPresent() && transaction.getChainId().isEmpty()) {
```

We should also make sure we have tests for all possible permutations here (with / without chainId, with / without unprotected txs allowed, txs with / without a chainId).

atoulme: I missed that! Thanks for pointing it out. I will add tests for it.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
atoulme: That sounds easier to parse indeed. I'll change to your suggestion.

shemnon (commented):
Still some dangling uses of "unprotected transaction" in the code.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidatorTest.java
```diff
@@ -389,6 +445,88 @@ public void goQuorumCompatibilityModeSuccessZeroGasPrice() {
...
+            .createTransaction(senderKeys);
+    when(gasCalculator.transactionIntrinsicGasCostAndAccessedState(transaction))
+        .thenReturn(new GasAndAccessedState(Gas.of(0)));
+    final ValidationResult<TransactionInvalidReason> result =
+        validator.validate(transaction, Optional.empty());
+    assertThat(result.isValid()).isTrue();
+  }
+
+  @Test
+  public void shouldAllowUnprotectedTransactionsWhenChainIdSetAndReplayProtectionNotSet() {
```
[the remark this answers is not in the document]
atoulme: done

shemnon:
LGTM, but since @mbaxter had some comments I will wait at least 24h to give her a chance to validate the changes address her issues.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -109,6 +116,11 @@ public MainnetTransactionValidator(
       return signatureResult;
     }
 
+    if (chainId.isPresent() && transaction.getChainId().isEmpty() && requireTxReplayProtection) {
```
mbaxter, line 119: Thought about this some more - this is actually the wrong place to validate.  This validator is executed on transactions received over gossip, and during block processing.  If we validate here, we're going to be unable to process blocks that contain unprotected transactions, and we're going to censor gossiped transactions.  We should move this validation to the JSON-RPC layer.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -832,6 +832,14 @@ void setBannedNodeIds(final List<String> values) {
       arity = "1")
   private final Wei txFeeCap = DEFAULT_RPC_TX_FEE_CAP;
 
+  @Option(
+      names = {"--require-tx-replay-protection"},
+      description =
+          "Requires transactions to be EIP-155 to be submitted via RPC (default: ${DEFAULT-VALUE})",
+      arity = "1")
```
mbaxter, line 839: A few more nits here:
* Suggest moving this to `RPCOptions`
* Suggest renaming to `--strict-tx-replay-protection-enabled` to be more in line with our other boolean option names.  
* Would be nice to allow no params with a fall-back to true
* Suggest some tweaks to the description as well, see suggestion below

```suggestion
  @Option(
      names = {"--strict-tx-replay-protection-enabled"},
      description =
          "Require transactions submitted via JSON-RPC to use replay protection in accordance with EIP-155 (default: ${DEFAULT-VALUE})",
      fallbackValue = "true",
      arity = "0..1")
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
atoulme: Ah, it was initially in the `TransactionPool` class under the `addLocalTransaction` method.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
atoulme: Sounds good!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
mbaxter: Would be good to get test coverage showing we can process blocks with a mix of un/protected txs with this option enabled.

mbaxter:
> LGTM, but since @mbaxter had some comments I will wait at least 24h to give her a chance to validate the changes address her issues.

Thanks - I usually check back in on PR's at most once a day :D

vmichalik:
@mbaxter @atoulme @shemnon any blockers on this PR we can help with? looks like it was 95% per the above thread!

mbaxter:
Hey @vmichalik - no blockers that I'm aware of.  There is a good bit more work to do here, however (see [this comment](https://github.com/hyperledger/besu/pull/1980#discussion_r601547780)) - we basically need to move the tx replay checks out of the core `ProtocolSpec` and probably into the `TransactionPool`.  

I'm going to go ahead and close this PR as it needs to be reworked.  @atoulme - feel free to reopen if you want to keep pushing on this.  At this point, though, it's probably easier to just open a new PR.

atoulme:
Please reopen and let’s get it done. Wanna take a stab at it?

mbaxter:
Yeah - I can try to take a look at this 👍

mbaxter:
fyi - started some work here: https://github.com/hyperledger/besu/pull/2497.  Just need to make sure that the new error is reported out to users correctly and will try to add some more tests as well.

jflo:
Should this be closed in favor of https://github.com/hyperledger/besu/pull/2497 ?

atoulme:
Sure.

PR #1984, opened 2021-03-06
Cache MessageDigests

shemnon:
Fetching the hash from the java security providers is often longer than
the hashing itself, so we should cache results from the providers in a
reusable fashion.

## Changelog


shemnon:
This will eliminate ~10% of the time wasted in mainnet block processing.  More time was spent fetching the hash algo then actually executing it.

crypto/src/main/java/org/hyperledger/besu/crypto/Hash.java
```diff
@@ -25,24 +27,41 @@
...
   private static final String BLAKE2BF_ALG = "BLAKE2BF";
 
+  private static final Supplier<MessageDigest> KECCAK256_SUPPLIER =
+      Suppliers.memoize(() -> messageDigest(KECCAK256_ALG));
+  private static final Supplier<MessageDigest> SHA256_SUPPLIER =
+      Suppliers.memoize(() -> messageDigest(SHA256_ALG));
+  private static final Supplier<MessageDigest> RIPEMD160_SUPPLIER =
+      Suppliers.memoize(() -> messageDigest(RIPEMD160_ALG));
+  private static final Supplier<MessageDigest> BLAKE2BF_SUPPLIER =
+      Suppliers.memoize(() -> messageDigest(BLAKE2BF_ALG));
```
RatanRSur, line 41: How do you feel about just initializing these statically instead of using Suppliers?

shemnon: The security providers may not be properly set up when the class initializes.  

For example, the blake2f provider is loaded via besu code on startup, and since the algorithms are loaded by string reference the code that adds blake2f to the providers is not guaranteed to have run since there is no class file dependency.  And this is the wrong place to do that provider setup.

RatanRSur (approved):

PR #1988, opened 2021-03-08
Notes about Berlin upgrade

timbeiko:
On March 5, the Berlin upgrade was changed to remove EIP-2315 ([issue](https://github.com/ethereum/pm/issues/263#issuecomment-791473406), [call](https://www.youtube.com/watch?v=xWfR-WxjmYg)). This PR updates the changelog to make it clear that 21.1.1 is our Berlin-compatible release, and that 21.1.0 users need to upgrade if they are using public networks.


matkt (approved):
LGTM

Issue #1687, opened 2020-12-10
Issue with Besu keeping files open when sender does not actively close them


pegartillo95:
**Besu client 20.10.1 accepts indiscriminately all incoming HTTP RPC connections, making JVM eventually collapse, if the connections are not closed by the invokers. Details below:**

##### Setup used for the test.
Client info:
- OS of the machine running the client: Ubuntu 18.04
- Client used to send the transactions: Artillery version 1.6.1
- Our client sending the transactions consist on a Artillery client on which we have a prebuilt json transaction. For each new sent transaction Artillery only needs to update the nonce and sign the new transaction.

Server info:
- OS of all the machines running the besu nodes: RedHat 7
- Version of all Besu Nodes: 20.10.1

Network info:
- Consensus Protocol: IBFT2
- Block time: 5 seconds.
- Topology of the network: 4 validators, 2 boot nodes and 1 regular node (this is the one to which we are sending the transactions)

### Test step by step
I'll explain now our testing procedure at the same time trying to show you some of the results we got.
The purpose we had when we found this issue was to stress test our private Besu network.
First we started sending transactions at a speed or about 250 txs/second, and we realized that suddenly after around 15 seconds the node was suddenly dying. So we started decreasing the number of txs/second little by little and we saw that it didn't matter how many txs/seconds we were sending the node will always crash after having processed around 3600 txs.
The number was always close to that amount it doesn’t matter if we sent 1 txs/second or 200 txs/second.

After that we realized that this was probably happening because of the number of open files that besu was keeping alive, which was continuously increasing during the test.
Here you can see some of our results for the number of open files we were getting in the case of sending only 1 tx/second, the metrics are taken from the machine running the regular node.
```
After roughly 15 minutes:
--------------------------------------
$ sudo lsof | grep java | wc -l
5206

$ sudo lsof | grep besu | wc -l
79494
--------------------------------------


After roughly 25 minutes:
--------------------------------------
$ sudo lsof | grep java | wc -l
8323

$ sudo lsof | grep besu | wc -l
151516
---------------------------------------


After roughly 35 minutes:
--------------------------------------
$ sudo lsof | grep java | wc -l
11946

$ sudo lsof | grep besu | wc -l
234714
--------------------------------------


At the point of having received around 3600, just before crashing
--------------------------------------
$ sudo lsof | grep java | wc -l
13803

$ sudo lsof | grep besu | wc -l
278623
--------------------------------------
```
[...]

PR #1996, opened 2021-03-10
limit the number of active RPC HTTP connections

macfarla:

Added option to set an upper limit on RPC HTTP connections. Once this limit is reached, any further connections will be immediately closed. 

This will prevent Besu from crashing if the client keeps opening new connections, and does not close connections.

See #1687 

## Changelog


RatanRSur:
So is the main value here to prevent the user from accidentally creating too many connections to the node? Because from the issue it seems like they were intentionally keeping them alive, which we should allow them to do since it's their node.

macfarla:
> So is the main value here to prevent the user from accidentally creating too many connections to the node? Because from the issue it seems like they were intentionally keeping them alive, which we should allow them to do since it's their node.

This won't close any existing connections, just stop accepting new connections. So if they set keepalive that should still work. Basically this should stop Besu from crashing.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -264,6 +275,34 @@ private void validateConfig(final JsonRpcConfiguration config) {
...
+    return connection -> {
+      if (activeConnectionsCount.get() >= maxActiveConnections) {
+        LOG.warn(
+            "Max Active Connections limit {} reached. Rejecting all new connections.",
+            maxActiveConnections);
+        // disallow new connections to prevent DoS
+        LOG.warn(
+            "Rejecting new client connection: remoteAddress: {} connection count: {}",
+            connection.remoteAddress(),
+            activeConnectionsCount.incrementAndGet());
```
lucassaldanha, line 289: We should probably log only one of these msgs. I believe we don't the two.

I think we can keep the msg when we reject the connection and change the msg slightly:
```
Rejected new HTTP connection from 1921.68.0.1. Max 80 active HTTP connections limit reached.
```
(or something similar)

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -264,6 +275,34 @@ private void validateConfig(final JsonRpcConfiguration config) {
...
+        connection.close();
+      } else {
+        LOG.warn(
```
lucassaldanha, line 292: This shouldn't be a warning. Change it to a debug message.

And we can change the msg slightly:
```
Opened HTTP connection from 192.160.1.1. Total of active HTTP connections: 12/80.
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -264,6 +275,34 @@ private void validateConfig(final JsonRpcConfiguration config) {
...
+            "New client connection: remoteAddress: {} connection count: {}",
+            connection.remoteAddress(),
+            activeConnectionsCount.incrementAndGet());
+      }
+      connection.closeHandler(
+          c ->
+              LOG.warn(
```
lucassaldanha, line 299: Not a warning. Change it to debug :)

And the msg:
```
Closed HTTP connection from 192.168.1.1. Total of active HTTP connection: 11/80
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketService.java
```diff
@@ -91,10 +91,8 @@ private WebSocketService(
                 new HttpServerOptions()
                     .setHost(configuration.getHost())
                     .setPort(configuration.getPort())
-                    .setHandle100ContinueAutomatically(true)
-                    .setCompressionSupported(true)
-                    .addWebSocketSubProtocol("undefined"))
-            .webSocketHandler(websocketHandler())
+                    .setWebsocketSubProtocols("undefined"))
```
lucassaldanha, line 94: What are these options that were removed?

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/LimitConnectionsJsonRpcHttpServiceTest.java
```diff
@@ -0,0 +1,46 @@
...
+import okhttp3.Response;
+import org.junit.Test;
+
+public class LimitConnectionsJsonRpcHttpServiceTest extends JsonRpcHttpServiceTestBase {
+
+  @Test
+  public void limitActiveConnections() throws Exception {
+    OkHttpClient newClient = new OkHttpClient();
+    int i;
+    for (i = 0; i < JsonRpcConfiguration.DEFAULT_MAX_ACTIVE_CONNECTIONS; i++) {
```
lucassaldanha, line 31: Is it possible to run this test with a lower limit? Probably 1? Then the second connection should fail.

It seems to be a waste of resources (and test time) opening 80 real connections to test the limit if we can reduce the config before starting the server.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketService.java
macfarla: this was not meant to be changed...

macfarla: reverted

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/LimitConnectionsJsonRpcHttpServiceTest.java
macfarla: refactored to add this

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/OpenTelemetryAcceptanceTest.java
```diff
@@ -191,11 +191,11 @@ public void traceReporting() {
   @Test
   public void traceReportingWithTraceId() {
 
+    OkHttpClient okClient = new OkHttpClient();
```
macfarla, line 194: moved this outside the wait so the connection gets reused

lucassaldanha (approved):

PR #1998, opened 2021-03-10
Update Ethereum Node Record When We Pass Network Upgrade Blocks

RatanRSur:
## Fixed Issue(s)

#1998 

## Changelog


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -178,52 +176,54 @@ protected PeerDiscoveryAgent(
...
+    final Integer discoveryPort = localNode.get().getEnodeURL().getDiscoveryPortOrZero();
+    final Integer listeningPort = localNode.get().getEnodeURL().getListeningPortOrZero();
+    return existingNodeRecord
+        .filter(
+            nodeRecord ->
+                id.equals(nodeRecord.get(EnrField.PKEY_SECP256K1))
+                    && addressBytes.equals(nodeRecord.get(EnrField.IP_V4))
+                    && discoveryPort.equals(nodeRecord.get(EnrField.UDP))
+                    && listeningPort.equals(nodeRecord.get(EnrField.TCP))
+                    && forkIdSupplier.get().equals(nodeRecord.get("eth")))
```
garyschulte, line 199: non-blocking feedback would be nice to have an enum rather than "eth" magic string.  I see that it is an existing practice though

besu/src/test/java/org/hyperledger/besu/RunnerBuilderTest.java
```diff
@@ -133,8 +144,53 @@ public void enodeUrlShouldHaveAdvertisedHostWhenDiscoveryDisabled() {
...
+            BlockDataGenerator.BlockOptions.create()
+                .setBlockNumber(1)
+                .setParentHash(genesisBlock.getHash()));
+    blockchain.appendBlock(block, gen.receipts(block));
+    assertThat(
+            storageProvider
+                .getStorageBySegmentIdentifier(BLOCKCHAIN)
+                .get("local-enr-seqno".getBytes(StandardCharsets.UTF_8)))
+        .isNotNull();
+  }
```
garyschulte, line 195: is this actually checking that the sequence has incremented or just that it isn't null?  could we assert that it has gone from 0 -> 1 ?

garyschulte (commented):
LGTM, non-blocking feedback

besu/src/test/java/org/hyperledger/besu/RunnerBuilderTest.java
RatanRSur: I'm really glad you mentioned this because in looking at it I actually found out I had an unrelated test bug (nulls vs optionals). I also expanded it a bit to update multiple times.

garyschulte (approved):
👍

matkt (approved):
Looking at the code this seems good to me. I wonder if it is possible to take advantage of the next goerli fork to verify that this feature works as we want?

RatanRSur:
Good idea, I should add a log line anyways.

PR #100, opened 2019-10-10
[PIE-2026] Separate in-sync from sync-status listeners

mbaxter:
## PR description
In order to prepare for updates to our `SyncStatus` events, this PR reworks sync-related listener logic so that we have a clean separation between consumers who care whether the node is "in-sync" according to some threshold vs consumers who care about the actual `SyncStatus` updates.

Changes:
* Add the ability to add "in-sync" listeners with arbitrary sync tolerances
* Deduplicate "in-sync" analysis by having `SyncStatusNodePermissioningProvider` subscribe to in-sync events directly
* Rework `SyncStateTest` to use `EthProtocolManagerTestUtil` (less mocking of peer behavior) so that various listeners are set up correctly
* Try to name subscriber methods more consistently
* Simplify `Synchronizer` subscription management by just delegating to `SyncState`

PR #99, opened 2019-10-10
[PIE-2023] Fixed casing of dynamic MetricCategorys

mark-terry:
Signed-off-by: Mark Terry <mark.terry@consensys.net>

## PR description
Updated the MetricCategoryRegistry to simplify dynamic `MetricCategory` instantiation and make them consistent with the existing static entries.

## Fixed Issue(s)
#PIE-2023

PR #1999, opened 2021-03-10
Update BonsaiLayeredWorldState implementation

matkt:
Signed-off-by: Karim TAAM <t2am.karim@gmail.com>

## PR description

Update BonsaiLayeredWorldState implementation to find valid code, storage and account corresponding to a block

### Test performed

- Fast sync bonsai trie goerli

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -47,59 +47,67 @@
...
+    return layeredWorldStatesByHash.containsKey(blockHash)
         || persistedState.blockHash().equals(blockHash)
         || worldStateStorage.isWorldStateAvailable(rootHash, blockHash);
   }
 
   @Override
   public Optional<MutableWorldState> getWorldState(final Hash rootHash, final Hash blockHash) {
-    if (layeredWorldStates.containsKey(blockHash)) {
-      return Optional.of(layeredWorldStates.get(blockHash));
-    } else {
```
shemnon, line 95: If we don't load TrieLog layered world states from disk how do we then do traces beyond near chain head?  We won't be able to do eth_call 1MM blocks back but we should be able to do trace_* from 1MM blocks back if we have the TrieLogs.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/processor/BlockReplay.java
```diff
@@ -146,7 +146,7 @@ public BlockReplay(
       return Optional.empty();
     }
     final MutableWorldState mutableWorldState =
-        worldStateArchive.getMutable(previous.getStateRoot(), previous.getHash()).orElse(null);
+        worldStateArchive.getWorldState(previous.getStateRoot(), previous.getHash()).orElse(null);
```
shemnon, line 149: Looks like there is a lingering type/name issue.  If the return type is a MutableWorldState then the method name should be getMutable or getMutableWorldState.  This should either go to just a `WorldState` or the method name should be `getMutable`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -129,15 +124,21 @@ public UInt256 getStorageValue(final Address address, final UInt256 key) {
     while (currentLayer != null) {
       final Optional<UInt256> maybeValue =
           currentLayer.trieLog.getStorageBySlotHash(address, slotHash);
-      if (maybeValue.isPresent()) {
+      final Optional<UInt256> maybeOriginalValue =
+          currentLayer.trieLog.getOriginalStorageBySlotHash(address, slotHash);
```
shemnon, line 128: I don't think this is a good place for getOriginalStorageBySlotHash.  Original in the sense of EIP-2200 gas costing.  I think we need to go to renaming BonsaiValue `getOriginal` as `getPrior` to remove that confusion.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -37,30 +37,31 @@
 /** A World State backed first by trie log layer and then by another world state. */
 public class BonsaiLayeredWorldState implements MutableWorldState, BonsaiWorldView, WorldState {
 
-  private final BonsaiWorldStateArchive worldStateArchive;
-  private final BonsaiWorldView parent;
+  private Optional<BonsaiWorldView> parent;
```
shemnon, line 40: Is this being re-worked so as only used for blocks that were before the persisted state?  (eg persisted #100, this is a trie log for #99)  In that case `parent` may not be the best name.  perhaps `nextTrieLog`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
matkt: Indeed I just put back the loading of the layer from the disk but in this case I put the nextLayer to empty and I load it on demand (call getNextWorldView) so as not to load all layers for nothing

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/processor/BlockReplay.java
matkt: Done renamed to getMutable

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
matkt: indeed I wanted to rename it in another PR but why not do it now

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
matkt: So we can have a layer or a persistate Worldstate so I renamed it to `nextWorldView` because those are both BonsaiWorldView.

Layer9 (parent = = Layer10), Layer10 (parent= = Layer11), Layer11 (parent==PersistedWorldState)

matkt:
@shemnon Do you have any other remarks before its merge into master ?

shemnon (approved):

PR #2002, opened 2021-03-10
Use scalars for consensus round encoding

rain-on:
Signed-off-by: Trent Mohay <trent.mohay@consensys.net>

## Changelog


consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validation/RoundChangePayloadValidatorTest.java
```diff
@@ -139,18 +139,6 @@ public void roundChangeWithZeroTargetRoundFails() {
     assertThat(messageValidator.validate(signedPayload)).isFalse();
   }
 
-  @Test
-  public void roundChangeWithNegativeTargetRoundFails() {
```
jframe, line 143: Why remove this test?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/payload/CommitPayload.java
```diff
@@ -26,8 +26,8 @@
 import java.util.Objects;
 import java.util.StringJoiner;
 
-public class CommitPayload implements Payload {
-  private static final int TYPE = QbftV1.COMMIT;
+public class CommitPayload extends QbftPayload {
+  public static final int TYPE = QbftV1.COMMIT;
```
jframe, line 30: why does this need to be public? couldn't see where it is used

jframe (approved):
LGTM

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/payload/CommitPayload.java
rain-on: yup - should be private, was part of a secondary piece of work I was tinkering with

PR #2003, opened 2021-03-11
Update Qbft payload signing to interop

rain-on:
## Changelog


usmansaleem (approved):
LGTM

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/payload/QbftPayload.java
```diff
@@ -29,4 +31,14 @@ protected void writeConsensusRound(final RLPOutput out) {
   protected static ConsensusRoundIdentifier readConsensusRound(final RLPInput in) {
     return new ConsensusRoundIdentifier(in.readLongScalar(), in.readIntScalar());
   }
+
+  @Override
+  public Hash hashForSignature() {
```
jframe, line 36: Can you add a test for this hashing

rain-on: without just reproducing the literal Rlp-ing, not sure how.

PR #2008, opened 2021-03-12
Add SECP256R1 support

daniel-iobuilders:
Signed-off-by: Daniel Lehrner daniel@io.builders

At the moment Besu only supports the Ethereum standard secp256k1 key algorithm. Due to NIST compliance, we would like to add support for secp256r1 keys.

No existing functionality was modified nor removed.


RatanRSur:
Hi @daniel-iobuilders ! We've added this to our backlog and will take a look at our earliest convenience.

acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/transaction/account/TransferTransaction.java
```diff
@@ -64,33 +75,67 @@ public TransferTransaction(
...
-            recipient.getAddress(),
-            Convert.toWei(transferAmount, transferUnit).toBigIntegerExact());
+    final RawTransaction transaction = createRawTransaction();
 
-    return toHexString(
+    return Numeric.toHexString(
         TransactionEncoder.signMessage(transaction, sender.web3jCredentialsOrThrow()));
   }
 
+  private String signedTransactionDataWithSignatureAlgorithm(
```
lucassaldanha, line 99: It looks like this method could be used to sign any transaction and not only a transfer transaction, is that right?
If that is the case, it would be better to move it to a utility class of some sort. Thoughts?

acceptance-tests/tests/src/test/resources/crypto/secp256r1.json
```diff
@@ -0,0 +1,24 @@
...
+    "chainId": 1981,
+    "ecCurve": "secp256r1",
+    "constantinoplefixblock": 0,
+    "ethash": {
+      "fixeddifficulty": 1000
+    }
+  },
+  "nonce": "0x0",
+  "timestamp": "0x58ee40ba",
+  "extraData": "0xf83ea00000000000000000000000000000000000000000000000000000000000000000d594c2ab482b506de561668e07f04547232a72897daf808400000000c0",
```
lucassaldanha, line 12: You can remove the extraData from the genesis file because it is not needed for ethhash.

crypto/src/main/java/org/hyperledger/besu/crypto/SECP256R1.java
```diff
@@ -0,0 +1,39 @@
...
+  private static final Logger LOG = LogManager.getLogger();
+  public static final String CURVE_NAME = "secp256r1";
+
+  public SECP256R1() {
+    super(CURVE_NAME, SecP256R1Curve.q);
+  }
+
+  @Override
+  public void enableNative() {
+    LOG.info("Native secp256r1 requested but not available");
```
lucassaldanha, line 32: This should probably be an ERROR log. And perhaps throwing a runtime exception up.
My concern is the user thinking that the node is using a native implementation when it is not.

crypto/src/test/java/org/hyperledger/besu/crypto/SECP256R1Test.java
```diff
@@ -0,0 +1,127 @@
...
+  protected static String suiteStartTime = null;
+  protected static String suiteName = null;
+
+  @BeforeClass
+  public static void setTestSuiteStartTime() {
+    suiteStartTime =
+        LocalDateTime.now(ZoneId.systemDefault())
+            .format(DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss"));
+    suiteName(SECP256K1Test.class);
+    ;
```
lucassaldanha, line 47: Don't need this extra semicolon.

lucassaldanha: (please remove it from `SECP256K1Test` as well) :)

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -208,7 +212,12 @@ public NodeRecord updateNodeRecord() {
                       nodeRecordFactory.createFromValues(
                           sequenceNumber,
                           new EnrField(EnrField.ID, IdentitySchema.V4),
-                          new EnrField(EnrField.PKEY_SECP256K1, Functions.compressPublicKey(id)),
+                          new EnrField(
```
lucassaldanha, line 215: I'm not 100% sure if we should change the key from `secp256k1`. I understand that it will be weird to have the key saying `k1` and the actual value is an `r1` public key. However, the ENR spec defined the key name to be `secp256k1` (unfortunately).

Maybe for now we leave the key as it was. I'll follow up and see if is is possible to update the spec.

For reference: https://eips.ethereum.org/EIPS/eip-778

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/ENRResponsePacketDataTest.java
```diff
@@ -29,6 +32,9 @@
 import org.junit.Test;
 
 public class ENRResponsePacketDataTest {
```
lucassaldanha, line 34: Same as before, let's keep the `secp256k1` as key for now.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
lucassaldanha: @RatanRSur it would be great to hear your thoughts on this.

acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/transaction/account/TransferTransaction.java
daniel-iobuilders: You are right, I've moves it to a utility class

acceptance-tests/tests/src/test/resources/crypto/secp256r1.json
daniel-iobuilders: :+1:

crypto/src/main/java/org/hyperledger/besu/crypto/SECP256R1.java
daniel-iobuilders: As discussed I change it to a warning, because using the native implementation is set by default and should not could cause errors when not manually deactivated

crypto/src/test/java/org/hyperledger/besu/crypto/SECP256R1Test.java
daniel-iobuilders: :+1:

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
daniel-iobuilders: After testing it: The `compressPublicKey` function needs to use the same elliptic curve as the node id, if not an error is caused. We decided to keep the key name as `EnrField.PKEY_SECP256K1` to be in compliance with the standard even though the value could be derived with another elliptic curve.

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/ENRResponsePacketDataTest.java
daniel-iobuilders: :+1:

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
RatanRSur: What was done before with getting the name of the curve is fine for the mainnet side as it uses the same value as the key.

On the NIST side, it changes this EIP non-functionally. However, strict adherence to the EIP is not really required as changing the cryptography already breaks mainnet compatibility. I would probably allow the name to change since it's less confusing for future people in NIST-land and adds some incompatibility to the ENRs. I don't think it's likely node operators will accidentally pollute themselves with NIST ENRs gotten by extra-protocol means but it's possible.

crypto/src/test/java/org/hyperledger/besu/crypto/SECP256R1Test.java
```diff
@@ -0,0 +1,126 @@
...
+
+  protected static String suiteStartTime = null;
+  protected static String suiteName = null;
+
+  @BeforeClass
+  public static void setTestSuiteStartTime() {
+    suiteStartTime =
+        LocalDateTime.now(ZoneId.systemDefault())
+            .format(DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss"));
+    suiteName(SECP256K1Test.class);
```
lucassaldanha, line 46: I believe this is supposed to be SECP256R1 instead of K1.

daniel-iobuilders: You are right, I've changed it

lucassaldanha (approved):

PR #2014, opened 2021-03-13
fix exponent overflow

RatanRSur:


matkt (approved):
LGTM, just looked at the code

PR #2033, opened 2021-03-16
Increase node diversity when downloading blocks

matkt:

## PR description

As a node operator, I don't want to see "Failed to download block 0xhex from peer 0xhex..." message from the same node over and over. I want to see besu fail over to another node and try again rather than failing from the same node

### Tes performed
 - Run goerli node

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/GetBlockFromPeersTask.java
```diff
@@ -0,0 +1,99 @@
...
+    return new GetBlockFromPeersTask(
+        peers, protocolSchedule, ethContext, hash, blockNumber, metricsSystem);
+  }
+
+  @Override
+  protected void executeTask() {
+    LOG.debug("Downloading block {} from peers {}.", hash, peers.stream().map(EthPeer::toString));
+    final Optional<EthPeer> peer =
+        peers.stream().filter(ethPeer -> !triedAndInvalidPeers.contains(ethPeer)).findFirst();
+    if (peer.isPresent() && !peer.get().isDisconnected()) {
```
RatanRSur, line 77: If even one peer is disconnected then we stop trying for others as well? I'm working on making this method simpler anyways, I'll show you my changes once I write them up.

matkt: Just updated the code thanks to your PR. thanks

RatanRSur (approved):

PR #2039, opened 2021-03-17
Update qbft extra data encoding to match GoQuorum qbft encoding

jframe:

## PR description
Update the QBFT extra data encoding to match what is used in the GoQuorum implementation.

In particular
* RLP encode round number as a scalar
* No vote is encoded as an empty list
* Encoding without commit seal, the seals are represented by an empty list
* Encoding without roundchange, the roundchange is represented by scalar of 0
* Roundchange message represents no block as an empty list

## Changelog


usmansaleem (approved):
LGTM

rain-on (approved):

PR #2043, opened 2021-03-18
Fix transaction encoding/decoding issues.

abdelhamidbakhta:
Signed-off-by: Abdelhamid Bakhta <abdelhamid.bakhta@consensys.net>


plugin-api/src/main/java/org/hyperledger/besu/plugin/data/TransactionType.java
```diff
@@ -18,8 +18,8 @@
 
 public enum TransactionType {
   FRONTIER(0xf8 /* doesn't end up being used as we don't serialize legacy txs with their type */),
-  ACCESS_LIST(0x1),
-  EIP1559(0xf /* placeholder value until we know what the real type byte will be */);
```
RatanRSur, line 22: If we know this is going to be the transaction type, we should remove this comment

abdelhamidbakhta: makes sense

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -634,32 +637,44 @@ private static Bytes frontierPreimage(
...
+                rlpOutput.writeList(
+                    accessList.get(),
+                    (accessListEntry, accessListEntryRLPOutput) -> {
+                      accessListEntryRLPOutput.startList();
+                      rlpOutput.writeBytes(accessListEntry.getAddress());
+                      rlpOutput.writeList(
+                          accessListEntry.getStorageKeys(),
+                          (storageKeyBytes, storageKeyBytesRLPOutput) ->
+                              storageKeyBytesRLPOutput.writeBytes(storageKeyBytes));
+                      accessListEntryRLPOutput.endList();
```
RatanRSur, line 670: We should probably deduplicate this between here and `TransactionEncoder::encodeAccessListInner`

abdelhamidbakhta: Yes

abdelhamidbakhta: Done https://github.com/hyperledger/besu/pull/2043/commits/093e27344a6aa281edb14edc75169c73bf08c910

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionEncoder.java
```diff
@@ -148,38 +148,58 @@ public static void encodeAccessListInner(
...
+    out.writeBigIntegerScalar(transaction.getV());
+    writeSignature(transaction, out);
+  }
+
   private static void writeSignatureAndRecoveryId(
       final Transaction transaction, final RLPOutput out) {
-    out.writeBigIntegerScalar(transaction.getV());
+    out.writeIntScalar(transaction.getSignature().getRecId());
     writeSignature(transaction, out);
   }
```
RatanRSur, line 204: Since the access list reference tests now have the bytes of the transaction, I think I can wire those together to get better coverage for this. I'll do that now and think that should go in before this.

abdelhamidbakhta: ok i am not ready to merge anyway. I will wait until i have all multi client integration tests passing

abdelhamidbakhta: Converted the PR to Draft to avoid issues.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthSendRawTransaction.java
```diff
@@ -64,12 +68,14 @@ public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
 
     final Transaction transaction;
     try {
-      transaction = DomainObjectDecodeUtils.decodeRawTransaction(rawTransaction);
-    } catch (final InvalidJsonRpcRequestException e) {
+      transaction = Transaction.readFrom(RLP.input(Bytes.fromHexString(rawTransaction)));
```
garyschulte, line 71: Did we determine that the rpc format is expecting RLP encoded txData rather than transaction byte prefixed opaque data?   https://github.com/hyperledger/besu/pull/2100#issuecomment-812051091

koalasensei: It works with both since`Transaction.readFrom` calls `TransactionDecorder.decodeForWire` which has the logic to decode both mode

abdelhamidbakhta: It works with both since `Transaction.readFrom` calls `TransactionDecorder.decodeForWire` which has the logic to decode both mode

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
RatanRSur: Can we not use the method you introduced here if we use `accessList.ifPresent`?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionEIP1559Test.java
```diff
@@ -14,65 +14,70 @@
...
+            .value(Wei.ZERO)
+            .gasLimit(30000)
+            .gasPremium(Wei.of(2))
+            .payload(Bytes.EMPTY.trimLeadingZeros())
+            .feeCap(Wei.of(new BigInteger("5000000000", 10)))
+            .gasPrice(null)
+            .to(Address.fromHexString("0x000000000000000000000000000000000000aaaa"))
+            .accessList(accessListEntries)
+            .guessType()
+            .signAndBuild(
```
RatanRSur, line 63: Not important but we could probably use the `BlockDataGenerator` 1559 transaction generator here

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
```diff
@@ -399,7 +399,6 @@ private Transaction eip1559Transaction(final Bytes payload, final Address to) {
     return Transaction.builder()
```
RatanRSur, line 399: We definitely need to fix this to randomly include/not include access lists. Probably makes it more worth it to use the generator in the `TransactionEIP1559Test` because then we would immediately notice that this can't include access lists.

RatanRSur (dismissed):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
abdelhamidbakhta: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -152,7 +152,7 @@ public Transaction(
     if (Objects.equals(transactionType, TransactionType.ACCESS_LIST)) {
       checkState(
           maybeAccessList.isPresent(), "Must specify access list for access list transaction");
-    } else {
+    } else if (!Objects.equals(transactionType, TransactionType.EIP1559)) {
       checkState(
           maybeAccessList.isEmpty(),
           "Must not specify access list for non-access list transaction");
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done

matkt (approved):
seems good for me

Issue #1931, opened 2021-02-20
Update ENR record when passing milestone blocks


shemnon:
The ENR record as currently constituted has a field holding the fork id of the node.  This fork ID should change when we roll over to the new fork.

We should broadcast the forkID of of the best _known_ block and when that crosses the barrier accept that fork id.  We will also need to write the forkid to the persistent storage when needed.

So probably a listener on the blockchain object, that has a reference to the protocol schedule and whatever class holds the ENR.  When we detect a change in the listener from observed values we fire off an event to the object calculating the ENR to rebuild it.

shemnon:
Re-opening as the ENR did not update after the goerli fork.  Perhaps it wasn't pushed down to the p2p layer?


```
dannos-mbp:~ dannoferrin$ devp2p enrdump  `devp2p -verbosity 1 discv4 requestenr enode://6ff032786b390dda567836c446b78f30e5544fc96d4f32d338e7b4537e7418bdff9b7c5843664e618f974205c5039e64ef250d3b4d1683f89968d5a245a693b0@127.0.0.1:30303`
Node ID: 65fc3954d0600096009ecb0e16c6846f0fda51c48bac02cfe4aec173540edb99
URLv4:   enode://6ff032786b390dda567836c446b78f30e5544fc96d4f32d338e7b4537e7418bdff9b7c5843664e618f974205c5039e64ef250d3b4d1683f89968d5a245a693b0@76.25.183.38:30303
Record has sequence number 3 and 6 key/value pairs.
  "eth"       cac984c25efa5c83441064
  "id"        "v4"
  "ip"        76.25.183.38
  "secp256k1" a1026ff032786b390dda567836c446b78f30e5544fc96d4f32d338e7b4537e7418bd
  "tcp"       30303
  "udp"       30303
```

shemnon:
The key is the 'eth' field.  The future fork block should be null and the "eth" entry should have become `"eth"       c7c684757a1c4780`

RatanRSur:
Ok, will take a look.

RatanRSur:
Where do I get this `devp2p` executable you're using?

shemnon:
geth.  `brew install ethereum`

PR #2047, opened 2021-03-19
Update `NodeRecord` on `DiscoveryPeer`

RatanRSur:
## PR description
Review suggestion: hide whitespace

We wrote the saved the new node record to disk but we didn't set it on the `DiscoveryPeer` class so we were still responding with the old one.

## Fixed Issue(s)
#1931 

## Changelog

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
@@ -166,7 +166,7 @@ public abstract class PeerDiscoveryAgent {
                 this.localNode = Optional.of(ourNode);
                 isActive = true;
                 LOG.info("P2P peer discovery agent started and listening on {}", localAddress);
-                ourNode.setNodeRecord(updateNodeRecord());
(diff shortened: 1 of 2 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -191,43 +191,46 @@ public NodeRecord updateNodeRecord() {
...
+                  LOG.info("Writing node record to disk. {}", nodeRecord);
+                  final KeyValueStorageTransaction keyValueStorageTransaction =
+                      keyValueStorage.startTransaction();
+                  keyValueStorageTransaction.put(
+                      Bytes.wrap(SEQ_NO_STORE_KEY.getBytes(UTF_8)).toArray(),
+                      nodeRecord.serialize().toArray());
+                  keyValueStorageTransaction.commit();
+                  return nodeRecord;
+                });
+    localNode.get().setNodeRecord(newNodeRecord);
```
matkt, line 232: This is an optional so it might be better to use an ifpresent

matkt (changes requested):
Looks good just one comment

matkt (approved):
LGTM

PR #2051, opened 2021-03-22
Bonsai trie: fix account has an empty state change

matkt:
Co-authored-by:  Danno Ferrin danno.ferrin@gmail.com

## PR description

Fix a consensus issue on goerli 4,462,228 : Account 0x0 had an empty state change.

### Test peformed : 
- FULL SYNC GOERLI (in progress)

## Changelog


RatanRSur:
So I don't doubt that this fixes the issue. And I don't understand Bonsai Tries too well (yet 😅) but one question I'd like to know the answer to is:

Is there a way to make it so that this account is never iterated over in the first place? It looks like it comes from storageToUpdate in BonsaiWorldStateUpdater. Is it possible that there's a better place to fix this upstream where this account never ends up in that collection in the first place if it's empty?

matkt:
> So I don't doubt that this fixes the issue. And I don't understand Bonsai Tries too well (yet 😅) but one question I'd like to know the answer to is:
> 
> Is there a way to make it so that this account is never iterated over in the first place? It looks like it comes from storageToUpdate in BonsaiWorldStateUpdater. Is it possible that there's a better place to fix this upstream where this account never ends up in that collection in the first place if it's empty?

I did some modification to avoid the loop.

RatanRSur:
I see that it says `TEST-FIX`, let me me know when it's ready for rereview

matkt:
> I see that it says `TEST-FIX`, let me me know when it's ready for rereview

I think you forgot to refresh the page before the review. this comment is no longer present in the last commit 🤔

RatanRSur:
You're right :)

RatanRSur (approved):
LGTM! It's a bigger diff but ultimately I think it's nice to have some invariant like "there won't be an entry with empty pending values" so it's easier to reason about the code

matkt:
> LGTM! It's a bigger diff but ultimately I think it's nice to have some invariant like "there won't be an entry with empty pending values" so it's easier to reason about the code

the goerli fullsync worked

PR #2054, opened 2021-03-23
Updated CLI option validation to handle dependent options with multiple main options.

### LINKED ISSUE #2038: Update log message when GoQuorum mode enabled
When GoQuorum mode is enabled, the privacy key and URL CLI options are needed.  A log message currently indicates they will be ignored when privacy is not enabled. 

- [ ] Sensible log message 

Check log message is there for GoQuorum mode if GoQuorum APIs enabled.

mark-terry:
Signed-off-by: Mark Terry <mark.terry@consensys.net>

## PR description
Updates the dependent CLI option validation to handle options that can be dependent on multiple main options.

## Fixed Issue(s)
Fixes #2038.

## Changelog

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
@@ -76,8 +76,8 @@ public class CommandLineUtils {
    * @param isMainOptionCondition the conditions to test dependent options against. If all
    *     conditions are true, dependent options will be checked.
    * @param dependentOptionsNames a list of option names that can't be used if condition is met.
-   *     Example: if --miner-coinbase is in the list and condition is that --miner-enabled should
-   *     not be false, we log a warning.
+   *     Example: if --min-gas-price is in the list and condition is that either --miner-enabled or
+   *     --goquorum-compatibility-enabled should not be false, we log a warning.
    */
   public static void checkMultiOptionDependencies(
       final Logger logger,
--- a/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
@@ -26,6 +26,8 @@ import picocli.CommandLine;
 public class CommandLineUtils {
   public static final String DEPENDENCY_WARNING_MSG =
       "{} has been ignored because {} was not defined on the command line.";
+  public static final String MULTI_DEPENDENCY_WARNING_MSG =
+      "{} ignored because none of {} was defined.";
 
   /**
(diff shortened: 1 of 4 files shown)


mark-terry:
New output looks like:
image

besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
```diff
@@ -51,20 +51,59 @@ public static void checkOptionDependencies(
...
+   *     if not won't
+   * @param dependentOptionsNames a list of option names that can't be used if condition is met.
+   *     Example: if --miner-coinbase is in the list and condition is that --miner-enabled should
+   *     not be false, we log a warning.
+   */
+  public static void checkMultiOptionDependencies(
+      final Logger logger,
+      final CommandLine commandLine,
+      final List<String> mainOptions,
+      final List<Boolean> isMainOptionCondition,
```
macfarla, line 84: I think it would be more intuitive to pass a single boolean expression rather than a list

besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
```diff
@@ -51,20 +51,59 @@ public static void checkOptionDependencies(
...
     }
   }
+
+  /**
+   * Check if options are passed that require an option to be true to have any effect and log a
+   * warning with the list of affected options. Multiple main options may be passed to check
+   * dependencies against.
+   *
+   * <p>Note that in future version of PicoCLI some options dependency mechanism may be implemented
+   * that could replace this. See https://github.com/remkop/picocli/issues/295
```
macfarla, line 68: is anything happening with 295?

besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
mark-terry: As in, pass a single parameter of `--privacy-enabled && --goquorum-compatibility-enabled` ?

besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
mark-terry: It's been implemented, but it'd require rework and I don't think it addresses this issue:
https://picocli.info/#_argument_groups

mark-terry:
Message updated to: 
image

besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
macfarla: yes - they don't get used separately, and passing them in separately sort of makes it seem like they will be

macfarla (approved):

PR #2057, opened 2021-03-23
Rename repolinter action

ryjones:
Signed-off-by: Ry Jones <ry@linux.com>

## PR description

Rename action to be more obvious.

### DIFF
--- a/.github/workflows/repolinter.yml
+++ b/.github/workflows/repolinter.yml
@@ -14,16 +14,11 @@ on:
       - main
 
 jobs:
-  build:
+  lint-repo:
     runs-on: ubuntu-latest
     container: hyperledger-tools.jfrog.io/repolinter:0.10.0
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
       - name: Lint Repo
-        run: bundle exec /app/bin/repolinter.js --rulesetUrl https://raw.githubusercontent.com/hyperledger-labs/hyperledger-community-management-tools/master/repo_structure/repolint.json --format markdown > /repolinter-report.md
-      - name: Save repolinter-report file
-        uses: actions/upload-artifact@v2
-        with:
-          name: repolinter-report
-          path: /repolinter-report.md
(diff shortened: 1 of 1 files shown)


.github/workflows/repolinter.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v2
       - name: Lint Repo
+        continue-on-error: true
```
NicolasMassart, line 24: Why not just keep the linter output in the console output of the action instead of an artefact?
It would enable checking the lint result code and get the output visible without the need to download and open the artefact because you can't view it online. Plus the output in the console is nicely coloured.

NicolasMassart: Plus I would change the job name from "build" to something like "linting-repo" as it's may be confusing for users who think it's building Besu.

.github/workflows/repolinter.yml
```diff
@@ -14,16 +14,11 @@ on:
...
       - name: Checkout Code
         uses: actions/checkout@v2
       - name: Lint Repo
-        run: bundle exec /app/bin/repolinter.js --rulesetUrl https://raw.githubusercontent.com/hyperledger-labs/hyperledger-community-management-tools/master/repo_structure/repolint.json --format markdown > /repolinter-report.md
-      - name: Save repolinter-report file
-        uses: actions/upload-artifact@v2
-        with:
-          name: repolinter-report
-          path: /repolinter-report.md
+        run: bundle exec /app/bin/repolinter.js --rulesetUrl https://raw.githubusercontent.com/hyperledger-labs/hyperledger-community-management-tools/master/repo_structure/repolint.json --format markdown
```
NicolasMassart, line 24: if you want it to display even better on the terminal, you can remove the `--format markdown` flag.

ryjones: this was how it came in from @shemnon . I was considering making it tee to a file, but while there are a lot of changes to possibly make, it isn't super interesting to me

NicolasMassart: did you intend to close your PR?

PR #2076, opened 2021-03-25
Use istanbul 100 subprotocol in QBFT

jframe:

## PR description
Use istanbul/100 subprotocol in qbft to match the sub-protocol used in Quorum for qbft.

## Changelog


usmansaleem (approved):
LGTM

PR #2088, opened 2021-03-29
Adds DNS discovery URL CLI command

### LINKED ISSUE #927: Support DNS discovery
# Description
As an operator of a Besu node, I want to be able to use DNS discovery to complement UDP-based discovery, using the mechanism defined with EIP-1459. 

### Acceptance Criteria
* An operator can define a host to start DNS discovery from (ex: example.com).
* In the absence of an explicit host, if the operator runs a named network (mainnet, ropsten), it uses the proper host name.
* DNS discovery may be explicitly disabled.

### LINKED ISSUE #1631: Add DNS Discovery as a CLI Flag
### Description
As a node operator, I want to be able to set the DNS Discovery URL via CLI so that I can use DNS Discovery URLs for networks other than the pre-configured named mainnet networks. 

### Acceptance Criteria
* DNS discovery URL is set via command line flag, such as `--discovery-dns` and the node uses it.  For a stand alone test a manually configured testnet should be used instead of the `--network` flag and DNS discovery URL is echoed in the info log

RichardH92:
## PR description
Adds a CLI flag `--dns-discovery-url` for specifying the DNS discovery URL
## Fixed Issue(s)
fixes #1631 

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Additions and Improvements
 
+- Adds `--dns-discovery-url` CLI command
+
 ### Bug Fixes
 
 - Fixed issue in discv5 where nonce was incorrectly reused.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Additions and Improvements
 
-- Adds `--discovery-dns-url` CLI command
+- Adds `--discovery-dns-url` CLI command [\#2088](https://github.com/hyperledger/besu/pull/2088)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Additions and Improvements
 
-- Adds `--dns-discovery-url` CLI command
+- Adds `--discovery-dns-url` CLI command
 
 ### Bug Fixes
 
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -1060,19 +1060,16 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
 
   @CommandLine.Option(
       names = {"--api-gas-price-blocks"},
-      paramLabel = MANDATORY_PATH_FORMAT_HELP,
       description = "Number of blocks to consider for eth_gasPrice (default: ${DEFAULT-VALUE})")
   private final Long apiGasPriceBlocks = 100L;
 
   @CommandLine.Option(
       names = {"--api-gas-price-percentile"},
-      paramLabel = MANDATORY_PATH_FORMAT_HELP,
       description = "Percentile value to measure for eth_gasPrice (default: ${DEFAULT-VALUE})")
   private final Double apiGasPricePercentile = 50.0;
 
   @CommandLine.Option(
       names = {"--api-gas-price-max"},
-      paramLabel = MANDATORY_PATH_FORMAT_HELP,
       description = "Maximum gas price for eth_gasPrice (default: ${DEFAULT-VALUE})")
   private final Long apiGasPriceMax = 500_000_000_000L;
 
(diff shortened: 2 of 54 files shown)


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2405,6 +2412,10 @@ private EthNetworkConfig updateNetworkConfig(final NetworkName network) {
         builder.setBootNodes(new ArrayList<>());
       }
       builder.setDnsDiscoveryUrl(null);
+    } else if (dnsDiscoveryUrl != null) {
```
shemnon, line 2415: I am of the opinion that the CLI flag should override any and all other configurations (default and genesis file).  So I don't think this should be an else if, and it should be after any other attempts to set the value (which would also be after the --network flag is considered).  This _may_ be the correct place but I don't have my IDE handy at the moment.

This should also be validated with a test.  One approach would be to do a config with `--network=goerli` and `--discovery-dns-url=some-bogus-value` and see if the builder is configured with the bogus value.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1089,6 +1089,13 @@ void setBannedNodeIds(final List<String> values) {
           "Specifies the static node file containing the static nodes for this node to connect to")
   private final Path staticNodesFile = null;
 
+  @SuppressWarnings({"FieldCanBeFinal", "FieldMayBeFinal"}) // PicoCLI requires non-final Strings.
+  @CommandLine.Option(
+          names = { "--dns-discovery-url" },
```
shemnon, line 1094: Hmm... this feels like discovery should come first.  perhaps `--discovery-dns-url`.  @NicolasMassart any opinions on this or ideas on who it should be run by?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
NicolasMassart: Here it only applies if genesis is null, otherwise if genesis is not null it matches the first condition and skips this else if.
That means that you can override discovery dns url only for known networks as you have to define either a genesis or a network.
For what I understood, known networks all already have their config available (see `DiscoveryConfiguration.java`) and even if you may want to override it, it's not the main expected use case.
The expectation, to me, is that you can use DNS discovery even with a custom network, that means in the case when you use a custom genesis file.
So the option should be taken in account in any cases, not only for predefined networks.

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -988,6 +988,20 @@ public void genesisAndNetworkMustNotBeUsedTogether() throws Exception {
         .startsWith("--network option and --genesis-file option can't be used at the same time.");
   }
 
+  @Test
+  public void testDnsDiscoveryUrlEthConfig() throws Exception {
+    final ArgumentCaptor<EthNetworkConfig> networkArg =
+            ArgumentCaptor.forClass(EthNetworkConfig.class);
+
+    parseCommand("--dns-discovery-url", "test123");
```
NicolasMassart, line 996: Testing with fake/wrong values is fine, but you should also respect the value syntax: `enrtree://<key>@<fqdn>` as defined in [eip-1459](https://eips.ethereum.org/EIPS/eip-1459).
It will not test that the DNS query works, it's not in the scope of the test here (and mocked), but it helps making sure it at least works with the method you are testing. Plus it helps making the tests understandable for readers. Don't forget that tests are used to test... obviously... but also to understand how things works. Tests are the first doc after code itself. We look at test very frequently to write the doc and create examples.

RichardH92:
Hey @shemnon @NicolasMassart thanks for the feedback! Sorry it's taken me a sec to get to the comments, but I think I have addressed them. Mind re-reviewing when you get a chance?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Additions and Improvements
 
+- Adds `--discovery-dns-url` CLI command
```
shemnon, line 7: Please add a PR link

shemnon (approved):

shemnon:
Also, DCO tag is missing on the final commit.  Please run the commands documented here - https://github.com/hyperledger/besu/pull/2088/checks?check_run_id=2260621607

PR #2099, opened 2021-03-31
Fix flaky test

macfarla:

Simplify this test setup by making only one node a miner. This also improves performance of the test.
Before: was failing 20% of the time when running locally (100 runs)
After: passing 100% of the time (100 runs)

## Changelog


pinges (approved):
LGTM

PR #2100, opened 2021-03-31
Address EIP-2930 transactions not in EIP-2718 typed transactions

garyschulte:
## Fixed Issue(s)

json-rpc fix for access list transactions which are RLP encoded rather than EIP-2718 typed transactions

this was discovered when manually running the ethereum reference tests for EIP-2930.  It isn't clear whether these transactions should be considered valid or not given that EIP-2930 specifies a typed transaction envelope, but this 
PR correctly handles them.

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Bug Fixes
 
+- Fixed an issue with access list transactions which were RLP encoded (non EIP-2718)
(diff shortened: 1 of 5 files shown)


garyschulte:
we determined that the tests are sending invalid json-rpc format and that rather than attempting to process these requests, we should instead fix the tests.  The reason these tests are not failing for geth is because the tests are not using json-rpc to make the state transitions.

PR #2108, opened 2021-04-06
Added a limit of back layers that can be loaded

matkt:

## PR description

Added a limit of back layers that can be loaded
- By default : 512
- Add new experimental flag (--Xbonsai-maximum-back-layers-to-load)
- Fix Eth_call to not return null when worldsate is not available

## Changelog


abdelhamidbakhta (approved):
Looks good to me.

PR #2112, opened 2021-04-07
Qbft to start round timer after block period expiry

### LINKED ISSUE #2095: QBFT round changes for initial round to sent after block period
QBFT currently sends a round change as soon as the request timeout has expired. This can result in a round change being sent on round 0 before there is a block that has been created. Without a block the network might stall if the network goes into round change.

From the issue raised in the interop discussion channel
> For the new height, the round change would not happen until, the block period has passed and the node has the proposal ready to propose.

jframe:

## PR description
Qbft to start round change timer after block period and proposal received

## Fixed Issue(s)

fixes #2095

## Changelog


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftRound.java
```diff
@@ -87,8 +88,13 @@ public QbftRound(
     this.messageFactory = messageFactory;
     this.transmitter = transmitter;
     this.bftExtraDataCodec = bftExtraDataCodec;
+    this.roundTimer = roundTimer;
 
-    roundTimer.startTimer(getRoundIdentifier());
+    if (getRoundIdentifier().getRoundNumber() == 0) {
```
usmansaleem, line 93: should there be an illegalstateexception (or invalid protocol exception) if rounder number is negative? or is it already sanitized above in the chain?

jframe: The round number always starts from 0 and is only ever incremented by 1. Don't think we need check that the number isn't negative here. The QbftBlockHeightManager should ensure that we create round change with the correct number.

usmansaleem: 👍🏼

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -99,50 +95,65 @@ public QbftBlockHeightManager(
...
     }
 
-    startNewRound(currentRound.getRoundIdentifier().getRoundNumber() + 1);
+    startNewRound(qbftRound.getRoundIdentifier().getRoundNumber() + 1);
 
     try {
       final RoundChange localRoundChange =
           messageFactory.createRoundChange(
-              currentRound.getRoundIdentifier(), latestPreparedCertificate);
+              currentRound.get().getRoundIdentifier(), latestPreparedCertificate);
```
usmansaleem, line 156: should it qbftRound instead of currentRound??

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -151,7 +162,8 @@ public void roundExpired(final RoundExpiry expire) {
       LOG.warn("Failed to create signed RoundChange message.", e);
     }
 
-    transmitter.multicastRoundChange(currentRound.getRoundIdentifier(), latestPreparedCertificate);
+    transmitter.multicastRoundChange(
+        currentRound.get().getRoundIdentifier(), latestPreparedCertificate);
```
usmansaleem, line 166: same as previous message, should it be change to qbftRound?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -239,14 +260,15 @@ private void startNewRound(final int roundNumber) {
...
+              roundFactory.createNewRoundWithState(
+                  parentHeader, futureRoundStateBuffer.get(roundNumber)));
       futureRoundStateBuffer.keySet().removeIf(k -> k <= roundNumber);
     } else {
-      currentRound = roundFactory.createNewRound(parentHeader, roundNumber);
+      currentRound = Optional.of(roundFactory.createNewRound(parentHeader, roundNumber));
     }
     // discard roundChange messages from the current and previous rounds
-    roundChangeManager.discardRoundsPriorTo(currentRound.getRoundIdentifier());
+    roundChangeManager.discardRoundsPriorTo(currentRound.get().getRoundIdentifier());
```
usmansaleem, line 271: is it correct assumption that currentRound won't be empty here?

usmansaleem: nvm ... it will always be available.

usmansaleem (approved):
LGTM - minor comments.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -99,50 +95,66 @@ public QbftBlockHeightManager(
...
   }
 
   @Override
   public void handleBlockTimerExpiry(final ConsensusRoundIdentifier roundIdentifier) {
-    if (roundIdentifier.equals(currentRound.getRoundIdentifier())) {
-      currentRound.createAndSendProposalMessage(clock.millis() / 1000L);
+    startNewRound(0);
+    final QbftRound qbftRound = currentRound.get();
+
+    if (!finalState.isLocalNodeProposerForRound(qbftRound.getRoundIdentifier())) {
```
saltiniroberto, line 110: We must check that the node is not the proposer for round 0, not for the current round.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -99,50 +95,66 @@ public QbftBlockHeightManager(
...
           roundIdentifier,
-          currentRound.getRoundIdentifier());
+          qbftRound.getRoundIdentifier());
     }
   }
 
   @Override
   public void roundExpired(final RoundExpiry expire) {
-    if (!expire.getView().equals(currentRound.getRoundIdentifier())) {
+    if (currentRound.isEmpty()) {
```
saltiniroberto, line 126: Can this ever be true?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -239,14 +260,15 @@ private void startNewRound(final int roundNumber) {
     LOG.debug("Starting new round {}", roundNumber);
     if (futureRoundStateBuffer.containsKey(roundNumber)) {
       currentRound =
-          roundFactory.createNewRoundWithState(
-              parentHeader, futureRoundStateBuffer.get(roundNumber));
+          Optional.of(
```
saltiniroberto, line 263: Why are we using Optional here?

saltiniroberto (commented):
This must be fixed as, otherwise, we may have a safety issue.
https://github.com/hyperledger/besu/pull/2112#discussion_r616333254

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
jframe: The current round is 0 at this point. As initialise the round above with startNewRound(0). I've added a comment and verify in the tests to show that this is the case. Also have this returning immediately is the round has already started before we receive the block expiry.

jframe:
> This must be fixed as, otherwise, we may have a safety issue.
> [#2112 (comment)](https://github.com/hyperledger/besu/pull/2112#discussion_r616333254)

Investigating this further we are fortunately always checking whether we are proposer against round 0 as we initialise the current round to zero in the same method. I've updated the tests to verify this is the case as well.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -229,24 +256,30 @@ public void handleRoundChangePayload(final RoundChange message) {
       final RoundChangeArtifacts roundChangeMetadata = RoundChangeArtifacts.create(result.get());
 
       if (finalState.isLocalNodeProposerForRound(targetRound)) {
-        currentRound.startRoundWith(
-            roundChangeMetadata, TimeUnit.MILLISECONDS.toSeconds(clock.millis()));
+        if (currentRound.isEmpty()) {
```
saltiniroberto, line 259: I don't think that this `if` block should be here.

saltiniroberto: It is ok to move from "no round" to any round as a consequence of a quorum of RoundChange messages

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
```diff
@@ -229,24 +256,30 @@ public void handleRoundChangePayload(final RoundChange message) {
...
+          startNewRound(0);
+        }
+        currentRound
+            .get()
+            .startRoundWith(roundChangeMetadata, TimeUnit.MILLISECONDS.toSeconds(clock.millis()));
       }
     }
   }
 
   private void startNewRound(final int roundNumber) {
```
saltiniroberto, line 269: @jframe What do you think about addiing a check that roundNumber > currentRound?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
jframe: This doesn't prevent moving from no round to a round. The if block is just so that we have a QbftRound object to call startRoundWith if this happens before the block expiry has occurred.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftBlockHeightManager.java
jframe: Don't think is necessary after all. The only way we can move to a arbitrary round is from a proposal or a round change. And both of these check the age of the message, so anything that for a prior round is logged and discarded by the handleRoundChangePayload on line 241 or by handleProposalPayload on line 186.

saltiniroberto (approved):

PR #2119, opened 2021-04-07
bump jacoco to support java 15 jdk asm

garyschulte:
## PR description

bump jacoco plugin version to 0.8.6 in order to unblock builds on jdk 15:
https://github.com/jacoco/jacoco/releases

ideally we could wait for 0.8.7 to be released where 15 and 16 are officially supported and 17 is experimental, but the release schedule for jacoco is unpredictable.

without this bump, when trying to execute tests we get unsupported class type errors.

Also, gradle does not support java 16 until version 7.x which is not yet released.  So **jdk 15 is the latest jdk which besu can build** on.

## Changelog


matkt (commented):
Have we had the opportunity to test to be sure it's work correctly ?

garyschulte:
> Have we had the opportunity to test to be sure it's work correctly ?

jacoco is just a test coverage tool, so it only really matters at build time.  I was able to build with 11, 14 and 15 after making the change, but that is the extent of the testing I did.

garyschulte:
I can confirm that besu builds, runs and passes general state tests with jdk 15 :)

matkt (approved):
LGTM

PR #2132, opened 2021-04-09
handle json-rpc raw transactions encoded in either

garyschulte:
## PR description
resubmitting this PR.  The recent 1559 merge changed the behavior of EthSendRawTransaction with the intent to work for both rlp encoded raw transactions *and* byte prefixed typed transactions.  However the result was that properly encoded 2718 transactions fail.

Dimitry updated retesteth to use the 2718 style encoding, so we need to fix this.  

## Changelog


matkt (commented):
Obviously we change the plugin API. Is this modification does not bring regression with plugins ?

matkt (approved):
LGTM

Issue #2125, opened 2021-04-08
`eth_estimateGas` with `nonce` is rejected on non-strict mode


ricardolyn:
### Description
As a Developer, I want `eth_estimateGas` to ignore extra parameters so that only build the Tx once for multiple RPC methods. 

### Acceptance Criteria
* The non-strict behaviour of `eth_estimateGas` should ignore if the user sends the `nonce` as part of the parameters (same way Geth/GoQuorum ignores). non-strick means not passing the `strict` argument as specified on https://besu.hyperledger.org/en/stable/Reference/API-Objects/#transaction-call-object.

### Steps to Reproduce (Bug)
1. send `eth_estimateGas` with a `nonce` as part of the params.

**Expected behavior:**

Besu responds with the estimated gas.

**Actual behavior:**

Besu rejects with an error 
```
Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "nonce" (class org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.JsonCallParameter), not marked as ignorable (11 known properties: "gasPrice", "value", "gasLimit", "gas", "from", "to", "feeCap", "gasPremium", "strict", "data", "payload"]).
```

**Frequency:** 100%

### Versions (Add all that apply)
* Software version: v21.1.3
* Docker Version: 
```
Client: Docker Engine - Community
 Cloud integration: 1.0.9
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:13:00 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:15:47 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Additional Information

[...]

PR #2133, opened 2021-04-12
Estimate gas: ignore nonce

macfarla:
Ignore nonce when supplied to eth_estimateGas or eth_call. This is to better support web3j calls as per #2125 

## Changelog


pinges (approved):
LGTM

PR #2135, opened 2021-04-12
Fix invalid hash during test

matkt:
## PR description

Fix an invalid hash during EVM testing. The sender's account was not deleted when it was empty

## Changelog


abdelhamidbakhta (approved):
Looks good to me.

PR #2137, opened 2021-04-12
Fix Invalid trace issue with bonsai trie 

matkt:

## PR description

Fix the bonsai layers bug that impacted the tracing. This came during a reorg the link between layers was not valid
This resulted in invalid traces based on bad blocks history

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -59,15 +59,30 @@ public BonsaiWorldStateArchive(
...
-    worldStateStorage = new BonsaiWorldStateKeyValueStorage(provider);
-    persistedState = new BonsaiPersistedWorldState(this, worldStateStorage);
-    layeredWorldStatesByHash = new HashMap<>();
+    this.worldStateStorage = new BonsaiWorldStateKeyValueStorage(provider);
+    this.persistedState = new BonsaiPersistedWorldState(this, worldStateStorage);
+    this.layeredWorldStatesByHash = new HashMap<>();
     this.maxLayersToLoad = maxLayersToLoad;
+    blockchain.observeBlockAdded(
+        event -> {
+          if (event.isNewCanonicalHead()) {
```
RatanRSur, line 68: (optional) Maybe an `if(!event.isCanonicalHead()) return;` here? Makes it easier to scan.

RatanRSur (approved):

PR #2138, opened 2021-04-13
Fix invalid error codes on some JSON RPC apis

matkt:

## PR description

This bugs were found during ethspams testing

- Fix bad error code when we give a block ahead (http code 200 instead of 400  in this case). In the eip-1898 it is recommended to use -32XXX error code and not -32602 when the block is not found

- Modify the code so the blockchainQueries classes does not use the persistent worlstate with bonsai trie. 

## Changelog


atoulme (approved):
LGTM please add a changelog entry

PR #2141, opened 2021-04-14
qbft to not include round number in proposal block

### LINKED ISSUE #2134: Qbft proposal to not include round number
Change the Qbft proposal block to not include round number and instead only populate the round number in the extra data on the commit phase before the commit seal is calculated. This is to match the approach taken in GoQuorum for the round number in the block.

jframe:

## PR description
Change QBFT so that the proposal block no longer includes the round number and uses the default value of 0 in the extra data. Round number is populated before the commit seal is calculated and included in the final block.

## Fixed Issue(s)

fixes #2134 

## Changelog


usmansaleem (approved):
LGTM

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidatorTest.java
```diff
@@ -167,30 +167,6 @@ public void validationFailsWhenMessageMismatchesExpectedHeight() {
     verifyNoMoreInteractions(blockValidator);
   }
 
-  @Test
```
saltiniroberto, line 170: What about having a test that shows that we now validate Proposal messages independently of the round number in the block?

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validation/RoundChangeMessageValidatorTest.java
```diff
@@ -227,35 +227,6 @@ public void prepareFromNonValidatorFails() {
     assertThat(messageValidator.validate(message)).isFalse();
   }
 
-  @Test
```
saltiniroberto, line 230: What about testing that the round number in the block does not affect the validation?

saltiniroberto (approved):
LGTM.

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidatorTest.java
jframe: done

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validation/RoundChangeMessageValidatorTest.java
jframe: done

PR #2143, opened 2021-04-15
Make public key decode errors debug level

### LINKED ISSUE #2142: Invalid packets in discovery agent throw an error, but should only be logged as a debug message
Recently a lot of error messages appear in the log when Besu syncs one of the test nets. The error messages look like this:

```
2021-04-12 18:34:42.924+00:00 | vert.x-eventloop-thread-1 | ERROR | VertxPeerDiscoveryAgent | Encountered error while handling packet
java.lang.IllegalArgumentException: Could not recover public key
	at org.hyperledger.besu.crypto.SECP256K1.recoverFromSignatureNative(SECP256K1.java:420)
	at org.hyperledger.besu.crypto.SECP256K1$PublicKey.recoverFromSignature(SECP256K1.java:539)
	at org.hyperledger.besu.ethereum.p2p.discovery.internal.Packet.<init>(Packet.java:83)
	at org.hyperledger.besu.ethereum.p2p.discovery.internal.Packet.decode(Packet.java:119)
	at org.hyperledger.besu.ethereum.p2p.discovery.VertxPeerDiscoveryAgent.lambda$handlePacket$6(VertxPeerDiscoveryAgent.java:221)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:313)
	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:832)
```

[A PR of mine ](https://github.com/hyperledger/besu/pull/1933) was changing SECP256K1, [so the first assumption was that it was causing this error messages](https://github.com/hyperledger/besu/pull/1933#issuecomment-818600272). [After investigating it and testing it with different releases of Besu](https://github.com/hyperledger/besu/pull/1933#issuecomment-819350912), I found out that the PR in question did not cause it. The same error messages appear with older releases of Besu, like 21.1.0 or 20.10.4, which were released before the PR was merged.

This can be verified and reproduced by using one of the mentioned releases or the current one with the following config file:
```
miner-enabled=false
graphql-http-enabled=false
sync-mode="FULL"
pruning-enabled=true
rpc-ws-enabled=false
data-path="/home/USER/chains/ropsten"
rpc-http-enabled=false
network="ROPSTEN"
```

After about 5 - 15 minutes of syncing the error messages will appear in the log.

The problem is that the discovery agent receives packets with invalid signatures and throws an exception afterwards. This packets are not causing any problems, they are simply invalid and should not create an error. I suggest that instead of throwing an exception it should just create a debug message, so it is not cluttering the logs.

This is the part of the code that I would suggest to change:
```
// org.hyperledger.besu.ethereum.p2p.discovery.internal.Packet.<init>(Packet.java:83)

    this.publicKey =
        SIGNATURE_ALGORITHM
            .get()
            .recoverPublicKeyFromSignature(keccak256(signedPayload), this.signature)
            .orElseThrow(
                () ->
                    new PeerDiscoveryPacketDecodingException(
                        "Invalid packet signature, " + "cannot recover public key"));
```

I suggest to catch the `IllegalArgumentException` thrown by `org.hyperledger.besu.crypto.SignatureAlgorithm.recoverPublicKeyFromSignature*`  and create a debug log message with `Invalid packet signature cannot recover public key` instead.

Please tell me your thoughts about this change. If it is considered to be alright I could create the PR for it.

Original reporters of the issue: @matkt @3nprob

3nprob:
## PR description
Catches packets where deciding public key failed, reducing the log level from ERROR to DEBUG.

## Fixed Issue(s)
fixes #2142

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Bug Fixes
 
+- Reduced logging level of public key decoding failure of malformed packets. [\#2143](https://github.com/hyperledger/besu/pull/2143)
+
 ### Early Access Features
 
 #### Previously identified known issues
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/Packet.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/Packet.java
@@ -119,11 +119,13 @@ public class Packet {
     final PacketData packetData;
     try {
       packetData = deserializer.deserialize(RLP.input(message, PACKET_DATA_INDEX));
+      return new Packet(packetType, packetData, Bytes.wrapBuffer(message));
     } catch (final RLPException e) {
       throw new PeerDiscoveryPacketDecodingException("Malformed packet of type: " + packetType, e);
+    } catch (final IllegalArgumentException e) {
+      throw new PeerDiscoveryPacketDecodingException(
+          "Failed decoding packet of type: " + packetType, e);
     }
-
-    return new Packet(packetType, packetData, Bytes.wrapBuffer(message));
   }
 
   public Buffer encode() {
--- a/ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PacketTest.java
+++ b/ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PacketTest.java
@@ -17,6 +17,7 @@ package org.hyperledger.besu.ethereum.p2p.discovery.internal;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import org.hyperledger.besu.ethereum.p2p.discovery.Endpoint;
+import org.hyperledger.besu.ethereum.p2p.discovery.PeerDiscoveryPacketDecodingException;
 
 import java.util.Optional;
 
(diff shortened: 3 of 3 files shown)


daniel-iobuilders:
I think it would be good to add a test as well to make sure that a `PeerDiscoveryPacketDecodingException` is thrown when the signature of a packet is invalid. You can use the following as message with an invalid signature for `Packet.decode`:

```
    BigInteger invalidMessageBI =
        new BigInteger(        "43f91d11b3338b4dbdf16db4f9fa25d7b4e2db81e6fd63f8f6884dfaea851e106f8f692c77169b387bde7c38832cf2d37a9b97b1553d07587ebe251ee21ee36e0ed54fd9218e3feea3bd13ca6982b25c204d5186e7ec5373ea664c91d42467b30102f3cb842f3ee37b82040e82765fa04139782abaccbc8fd290a7fde1ff138943fa9659f7bd67f97c97b09893d1ee8a84607806e108", 16);
    Buffer message = Buffer.buffer(invalidMessageBI.toByteArray());
```

shemnon:
I like the idea of a test, but asserting on specific logging outputs has proven to produce flakey tests.  I would have to be a lower level unit test specific to `Packet.decode`

daniel-iobuilders:
@shemnon Yes, I was thinking the same. Basically in `org.hyperledger.besu.ethereum.p2p.discovery.internal.PacketTest` you could add this test:

```
  private static final String INVALID_SIGNATURE_PACKET = "43f91d11b3338b4dbdf16db4f9fa25d7b4e2db81e6fd63f8f6884dfaea851e106f8f692c77169b387bde7c38832cf2d37a9b97b1553d07587ebe251ee21ee36e0ed54fd9218e3feea3bd13ca6982b25c204d5186e7ec5373ea664c91d42467b30102f3cb842f3ee37b82040e82765fa04139782abaccbc8fd290a7fde1ff138943fa9659f7bd67f97c97b09893d1ee8a84607806e108";
  
    @Test(expected = PeerDiscoveryPacketDecodingException.class)
  public void invalidSignatureShouldThrowPeerDiscoveryPacketDecodingException() {
    decode(INVALID_SIGNATURE_PACKET);
  }
```

matkt:
Any news regarding this PR  ? is it ready to merge ?

shemnon:
Needs a test.  The OP either needs to merge @daniel-iobuilders' suggestion or a maintainer needs to add it by hand.

3nprob:
@shemnon I did add a test but:

When doing `gradlew build` I get an unrelated failing test that I *think* is unrelated to this change:

```
org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.DebugTraceBlockTest > shouldReturnCorrectResponse FAILED
    java.lang.AssertionError at DebugTraceBlockTest.java:135

1863 tests completed, 1 failed

> Task :ethereum:api:test FAILED

```

Building and testing takes ~1h on my workstation and viewing CI results on this repo is not public so I haven't been able to free up the necessary time to get to the bottom of what's going on here (and it was close to a decade since I last did serious Java)

3nprob:
Hmm, rebuilding from the same commit a couple of times and this error doesn't appear consistently - maybe an unrelated race condition?

shemnon (approved):
Sounds like a flakey test.  The gold standard however is did the CircleCI integration pass the flakey test, and it did.  So it's approved.  

Is it ready to merge?

3nprob:
Thanks @shemnon! (actually on my first attempt CircleCIR failed but test results are private so couldnt see in circlecir why ;))

PR #2154, opened 2021-04-19
Implement faucet for Aleut devnet

abdelhamidbakhta:
## PR description

Implements `debug_faucet` JSON RPC endpoint to enable users to request `ETH` on Aleut devnet. The endpoint takes one parameter which corresponds to the address of the recipient.

Adds 2 experimental flags:
-  `--Xfaucet-value`: Faucet value denominate in ETH (default: 5 ETH)
- `--Xfaucet-private-key`: Faucet account private key (default: first account in `dev.json` genesis)

### Sample request
```
{
   "method":"debug_faucet",
   "id":1,
   "jsonrpc":"2.0",
   "params":[
      "0xC3298C6341f82468309302611e24D3003Bc79B46"
   ]
}
```

### Success response
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0xb7d3183aff348d2d5994a978417129a4987d9f425295cc8e85b6563d2895a7bb"
}
```

### Unauthorized response
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -40100,
        "message": "Unauthorized"
    }
}
```
## Changelog

Signed-off-by: Abdelhamid Bakhta <abdelhamid.bakhta@consensys.net>

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugFaucet.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugFaucet.java
@@ -35,6 +35,10 @@ import org.hyperledger.besu.ethereum.rlp.RLPException;
 import org.hyperledger.besu.ethereum.transaction.TransactionInvalidReason;
 
 import java.math.BigInteger;
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import com.google.common.base.Suppliers;
@@ -44,6 +48,7 @@ import org.apache.tuweni.bytes.Bytes32;
 public class DebugFaucet implements JsonRpcMethod {
   private static final com.google.common.base.Supplier<SignatureAlgorithm> SIGNATURE_ALGORITHM =
       Suppliers.memoize(SignatureAlgorithmFactory::getInstance);
+  private static final long FAUCET_ACCESS_INTERVAL_SECONDS = TimeUnit.HOURS.toSeconds(6);
 
   private final Supplier<BlockchainQueries> blockchainQueries;
   private final Supplier<TransactionPool> transactionPool;
@@ -51,6 +56,7 @@ public class DebugFaucet implements JsonRpcMethod {
   private final KeyPair faucetKeyPair;
   private final Address faucetAddress;
   final Wei faucetAmount;
+  private final Map<String, Instant> access = new HashMap<>();
 
   public DebugFaucet(
       final BlockchainQueries blockchainQueries,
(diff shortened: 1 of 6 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugFaucet.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import com.google.common.base.Suppliers;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+
+public class DebugFaucet implements JsonRpcMethod {
+  private static final com.google.common.base.Supplier<SignatureAlgorithm> SIGNATURE_ALGORITHM =
```
RatanRSur, line 49: Why not use a plain java Supplier here?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugFaucet.java
```diff
@@ -0,0 +1,139 @@
...
+      transaction =
+          Transaction.builder()
+              .chainId(chaindId)
+              .nonce(nonce)
+              .value(faucetAmount)
+              .gasLimit(30000)
+              .payload(Bytes.EMPTY.trimLeadingZeros())
+              .gasPrice(Wei.of(1000))
+              .to(Address.fromHexString(recipientAddress))
+              .guessType()
```
RatanRSur, line 105: The intention behind `guessType` was that it would be used when we're parsing transactions from json and they don't explicitly say the transaction type. I think we should be explicit whenever we're building the transaction.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionEIP1559Test.java
```diff
@@ -35,44 +34,44 @@
...
-    final List<AccessListEntry> accessListEntries =
-        List.of(
-            new AccessListEntry(
-                Address.fromHexString("0x000000000000000000000000000000000000aaaa"),
-                List.of(Bytes32.ZERO)));
+    /*final List<AccessListEntry> accessListEntries =
+    List.of(
+        new AccessListEntry(
+            Address.fromHexString("0x000000000000000000000000000000000000aaaa"),
+            List.of(Bytes32.ZERO)));*/
```
RatanRSur, line 51: commented out code in this file

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/DebugJsonRpcMethods.java
```diff
@@ -77,25 +80,40 @@ protected RpcApi getApiGroup() {
...
+            chainId -> {
+              final DebugFaucet debugFaucet =
+                  new DebugFaucet(
+                      blockchainQueries,
+                      transactionPool,
+                      chainId,
+                      ExperimentalFeatures.faucetPrivateKey,
+                      Wei.fromEth(ExperimentalFeatures.faucetValue));
+              debugJsonRpcMethods.put(debugFaucet.getName(), debugFaucet);
+            });
```
RatanRSur, line 116: How come this is conditional on chainId? Could we instead make the private key field in the experimental options class `Optional<String>` and make this operation conditional on that?

PR #2163, opened 2021-04-20
Fix: Native libraries of secp256k1 and alt bn 128 can be disabled

### LINKED ISSUE #2150: Native libraries of secp256k1 and alt bn 128 cannot be disabled
Both `SECP256K1.class` and `AltBN128AddPrecompiledContract.class` have a Java implementation and another one using native libraries. The native library implementations are enabled by default in both cases. `BesuCommand` has the options `--Xsecp256k1-native-enabled` and `--Xaltbn128-native-enabled` which can be set to `false` to disable the native libraries and use the Java implementations. 

But right now the disabling of the libraries does not work:

`SECP256K1.class` and `AltBN128AddPrecompiledContract.class` have a boolean flag to enable or disable the native libraries, which is initialized with `true`:
```
private boolean useNative = true;
```

In `BesuCommand` the handling of the options looks like this:
```
  private void configureNativeLibs() {
    if (unstableNativeLibraryOptions.getNativeAltbn128()) {
      AbstractAltBnPrecompiledContract.enableNative();
    }
    if (unstableNativeLibraryOptions.getNativeSecp256k1()) {
      SignatureAlgorithmFactory.getInstance().enableNative();
    }
  }
```

`unstableNativeLibraryOptions.getNativeAltbn128()` and `unstableNativeLibraryOptions.getNativeSecp256k1()` return `true` by default and `false` when one of the corresponding options is set to `false`.

If one of the options is set to `false` in the command line, it simply skips the `enableNative` function call, but because the `useNative` variable is initialized with `true`, it never changes to `false`. There is also no other function available to set it to `false`. Because `useNative` can never be changed to `false` the native library implementations are always used.

I would suggest to initialize `useNative` with `false`. The native libraries would still be enabled by default, because in `BesuCommand` the `enableNative` functions will be called by default and set `useNative` to `true`. But if `--Xsecp256k1-native-enabled` or `--Xaltbn128-native-enabled` are set to `false` the `enableNative` functions will not be called and `useNative` stays `false` which is the intended behavior.

Please tell me your thoughts about the suggested change.

daniel-iobuilders:
## PR description
This PR fixes the ability to disable the native library implementations for secp256k1 and altBn128 via the command line. I added a log message in case of using the Java implementation, to make it clear to the user which implementation is used in both cases.

## Fixed Issue(s)
fixes #2150 

## Changelog


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1180,6 +1180,7 @@ public void run() {
 
     try {
       configureLogging(true);
+      instantiateSignatureAlgorithmFactory();
```
daniel-iobuilders, line 1183: I moved the instantiation of the signature algorithm here, to execute it before the native libraries are configured. Otherwise the wrong signature algorithm could be configured in a future release when more than one is supported in Besu.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 ### Bug Fixes
 
 - Fixed `NullPointerException` when crossing network upgrade blocks when peer discovery is disabled. [\#2140](https://github.com/hyperledger/besu/pull/2140)
+- Fixed disabling of native libraries for secp256k1 and altBn128. [\#2163](https://github.com/hyperledger/besu/pull/2163)
```
lucassaldanha, line 27: We need to move this into 21.1.6

crypto/src/main/java/org/hyperledger/besu/crypto/SECP256K1.java
```diff
@@ -44,7 +44,7 @@
 
   private static final Logger LOG = LogManager.getLogger();
 
-  private boolean useNative = true;
+  private boolean useNative = false;
```
lucassaldanha, line 47: Add a comment on why we set this to false even though using native is the default behaviour.

CHANGELOG.md
daniel-iobuilders: Done. Thanks!

crypto/src/main/java/org/hyperledger/besu/crypto/SECP256K1.java
daniel-iobuilders: In the end I changed it back to `true`. If not the tests that don't use BesuCommand, would be using the Java implementation instead of the native one. So it would have changed the test behavior. I've added a `disableNative` function instead and call it when the native libraries should be disabled.

lucassaldanha (approved):

PR #2164, opened 2021-04-20
Bonsai trie : fix consensus issue on ropsten

matkt:

## PR description

Fix an invalid originalValue for a specific storage slot :

in the same block we are doing
1 ) create a contract
2) change the value of a slot
3) change a second time the value of the same slot. the original is empty while it should be the previous value of the step 2

Why because when we create a contract we clear the storage so the method return 0x 

https://github.com/hyperledger/besu/blob/e13dd77c8d3a4ad6b003d2c71526a39818195ea8/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java#L314

## Changelog


abdelhamidbakhta (approved):
Looks good to me.

PR #2172, opened 2021-04-21
check that doesn't break during release

RatanRSur:
`dev` isn't present in the version string during a release. `besu` is, however.


matkt (approved):
LGTM

PR #2184, opened 2021-04-23
fix GoQuorum privacy

pinges:
The /receive RPC method that was used by Besu to retrieve private transactions in qoQuorum privacy mode from Tessera has been removed. This PR changes Besu to use the same RPC that is used by GoQuorum.


macfarla (approved):
LGTM

PR #2185, opened 2021-04-23
Ignoring changelog modification on CI

matkt:

## PR description

During releases we can waste time waiting for the pipeline while we only change the CHANGELOG. This PR will not run the tests if we only modify the CHANGELOG

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,19 @@ notify:
     - url: $HUBOT_URL
 
 commands:
+
+  check-tests-needed:
+    description: "Checks if tests are needed"
+    steps:
+      - checkout
+      - run:
+          name: Checking for changes
+          command: |
+            changes=`git --no-pager diff --name-only master...HEAD`
+            if [ "$changes" = ".circleci/config.yml" ]; then
+              echo "Only found a CHANGELOG change. Stopping build"
+              circleci-agent step halt
+            fi
   prepare:
     description: "Prepare"
     steps:
@@ -87,6 +100,7 @@ jobs:
   assemble:
     executor: besu_executor_xl
     steps:
+      - check-tests-needed
       - prepare
       - run:
           name: DCO check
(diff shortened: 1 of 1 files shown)


RatanRSur (dismissed):

ajsutton:
I can't say I'm a fan of this approach. It adds another piece of complexity to CI and means the build we actually release never gets tested at all.  At minimum we should continue running the build on master branches and apply this skip only to the PR builds.

We've solved the issues with waiting for builds in Teku by moving the changelog out of git - only the unreleased changes are there so we can update it as we build stuff, then at any point post release delete entries that have now been released.  The official changelog is then the GitHub releases page, though it could just as easily be a wiki page.

I'd also recommend using tag based releasing which dramatically reduces release times as well but that's a separate issue.

RatanRSur:
Those are good points and we keep pushing off the tag-based release discussion. I think we should park this until we have more clarity on what we want to do on that front.

matkt:
Preventing the pipeline from executing on a CHANGELOG modification is not going to prevent testing the release (I voluntarily let the pipeline execute when there is modification of the gradle.properties).

The steps I'm trying to make faster are as follows :
- Several PR were merged without modifying the CHANGELOG (often detected just before the release)
- Add the download link in the CHANGELOG after the PR
- And prepare the CHANGELOG for the new snapshot version of Besu

Cleaning the CHANGELOG and keeping only unrelease things is a very good idea but unfortunately this is not going to be enough to correct the problems listed above.

**I think indeed the solution is using tag based releasing**

Maybe we need to look at that

ajsutton:
> Cleaning the CHANGELOG and keeping only unrelease things is a very good idea but unfortunately this is not going to be enough to correct the problems listed above

If the change log only contained unreleased items you wouldn't bother updating it for the missed changes - just add them to where you publish the release notes for the release. They'd be deleted again from the change log in git straight after the release anyway. 

The other alternative is to update the change log in the same pr as you update the gradle properties to trigger the release. That used to be standard practice and can still be done with gar-e - just add additional commits to the pr it creates before approving it.

vmichalik:
@matkt are you still wanting to merge this?

PR #2197, opened 2021-04-28
fix selection of receive RPC

pinges:
There are two RPC in Tessera for GET "/transaction/{hash}". This makes sure we are using the right one.


macfarla (approved):
LGTM

PR #2203, opened 2021-04-29
reduce blocks required for permissioning related tests

macfarla:
There is no reason to wait for 50 blocks. Reduce the number of blocks so tests run quicker

## Changelog


mark-terry (approved):
Looks reasonable.

PR #2204, opened 2021-04-29
Fix entropy

matkt:

## PR description

 Use non-blocking randomness for acceptance tests

## Changelog


acceptance-tests/tests/src/test/resources/acceptanceTesting.security
```diff
@@ -2,5 +2,6 @@
 ### FOR ACCEPTANCE TESTING ONLY ###
 # Disable OS provided entropy
 securerandom.source=
+java.security.egd=
```
macfarla, line 5: I think since we already have securerandom.source set this won't have any effect

matkt: yes indeed it does not correct anything . I will close this PR

PR #2213, opened 2021-05-03
Ask block when not present

matkt:

## PR description

Two nodes stopped importing this weekend. This PR is a fix candidate. The nodes were late and some blocks were not stored in the pending blocks cache (because there is a size limit with announced blocks). At the time of importing the missing block, Besu did not find the block and stopped; Here is my theory. So I implemented a mechanism that directly asks the missing blocks to peers in these cases

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -143,6 +145,20 @@ private void onBlockAdded(final BlockAddedEvent blockAddedEvent) {
...
+          .map(ProcessableBlockHeader::getNumber)
+          .ifPresent(
+              minAnnouncedBlockNumber -> {
+                long distance =
+                    minAnnouncedBlockNumber
+                        - protocolContext.getBlockchain().getChainHeadBlockNumber();
+                if (distance < config.getBlockPropagationRange().upperEndpoint()
+                    && minAnnouncedBlockNumber > newBlock.getHeader().getNumber()) {
+                  retrieveMissingAnnouncedBlock(newBlock.getHeader().getNumber() + 1);
+                }
```
RatanRSur, line 160: We already check that a block is within the import range in `handleNewBlockFromNetwork`. Do you see an issue in always attempting to get the `chainHead +1`th block if there are any saved blocks? For example:

```java
      long nextExpectedBlockNumber = newBlock.getHeader().getNumber() + 1;
      if (pendingBlocksManager
          .lowestAnnouncedBlock()
          .map(ProcessableBlockHeader::getNumber)
          .filter(minAnnouncedBlockNumber -> minAnnouncedBlockNumber > nextExpectedBlockNumber)
          .isPresent()) {
        retrieveMissingAnnouncedBlock(nextExpectedBlockNumber);
      }
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -245,11 +261,29 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
     }
   }
 
+  @SuppressWarnings("unused")
```
RatanRSur, line 264: Why do we need this?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
matkt: this is to avoid triggering this mechanism when we are starting a fullsync from scratch for example and/or when we are too far from the head. Just to use this when we are in a "block by block" kind of sync.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
matkt: Yep true, I will fix that

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
RatanRSur: Ah I see. Sounds good.

RatanRSur (approved):

PR #2218, opened 2021-05-04
PermissioningService Besu Plugin

antonydenyer:
# PermissioningService Besu Plugin

An additional plugin extension point has been added to besu to enable developers to intercept p2p connections and messages.

```java
public interface PermissioningService extends BesuPluginService {
  void registerNodePermissioningProvider(NodeConnectionPermissioningProvider provider);
  void registerNodeMessagePermissioningProvider(NodeMessagePermissioningProvider provider);
}
```

## NodeConnectionPermissioningProvider
This will allow plugin users to hook into node connections and when a node attempts to send messages using the following interface.
```java
@FunctionalInterface
public interface NodeConnectionPermissioningProvider {
  /**
   * Can be used to intercept the initial connection to a peer. Note that once a connection is
   * established it's bi-directional.
   *
   * @param sourceEnode the originators enode
   * @param destinationEnode the enode you are about to send to
   * @return if you can connect
   */
  boolean isConnectionPermitted(final EnodeURL sourceEnode, final EnodeURL destinationEnode);
}
```

## NodeMessagePermissioningProvider

If a provider has been registered through the plugin API all messages sent to a peer will use the hook to check if the message can be sent to a peer.

```java
@FunctionalInterface
public interface NodeMessagePermissioningProvider {
  /**
   * Can be used to intercept messages before they are sent from besu. 
   *
   * Note! this method is called on every message send.
   *
   * @param destinationEnode the enode you are about to send to
   * @param code devp2p code for the message
   * @return if we can send the message to the peer
   */
  boolean isMessagePermitted(final EnodeURL destinationEnode, final int code);
}
```


.circleci/config.yml
```diff
@@ -163,7 +163,7 @@ jobs:
     steps:
       - prepare
       - attach_workspace:
-          at: ~/project
+          at: ~/projectby
```
ricardolyn, line 166: typo or necessary change?

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/peers/EnodeURL.java
```diff
@@ -32,7 +32,7 @@
 import com.google.common.primitives.Ints;
 import org.apache.tuweni.bytes.Bytes;
 
-public class EnodeURL {
+public class EnodeURL implements org.hyperledger.besu.plugin.data.EnodeURL {
```
ricardolyn, line 35: should the `EnodeURL` class be named more about what implementation is doing? for p2p, etc

ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/node/provider/SyncStatusNodePermissioningProvider.java
```diff
@@ -91,7 +91,9 @@ private void handleInSyncEvent(final boolean isInSync) {
    *     otherwise
    */
   @Override
-  public boolean isPermitted(final EnodeURL sourceEnode, final EnodeURL destinationEnode) {
+  public boolean isConnectionPermitted(
+      final org.hyperledger.besu.plugin.data.EnodeURL sourceEnode,
```
ricardolyn, line 95: I think we really need to rename to avoid the package specification when using `EnodeURL` interface

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/PermissioningService.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package org.hyperledger.besu.plugin.services;
+
+import org.hyperledger.besu.plugin.services.permissioning.NodeConnectionPermissioningProvider;
+import org.hyperledger.besu.plugin.services.permissioning.NodeMessagePermissioningProvider;
+
+public interface PermissioningService extends BesuPluginService {
+  void registerNodePermissioningProvider(NodeConnectionPermissioningProvider provider);
+
+  void registerNodeMessagePermissioningProvider(NodeMessagePermissioningProvider provider);
```
ricardolyn, line 23: what about `unregister`? it should be there as plugins might want to deregister too.

Moreover, I see some naming convention of other services where is used the convention `addListener`/`removeListener`. shouldn't we use also here add/remove?

antonydenyer: The listeners are different, you're observing something that has happened. You are not changing the behaviour of the node. You may want to listen to blocks, then after block number n listen to events for 100 blocks or something. Either way, the act of observing doesn't change the behaviour of the node. 

Whereas with things that change behaviour I don't it make sense, would you run a node with permissioning enabled and then turn permissioning off whilst it's running. Maybe you would, but you would need to know information about the block height when making that decision.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/peers/EnodeURL.java
antonydenyer: This is a difficult one, the `EnodeURL` introduces a circular dependency between the plugin and the implementation. That's why I introduced the interface. From an OO naming point of view, nothing really makes sense. Hungarian`IEnodeURL` isn't used anywhere else in besu `Impl` is but would mean there would be about 600 changes!

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins/PermissioningPluginTest.java
```diff
@@ -0,0 +1,103 @@
...
+  @Test
+  public void transactionsAreNotSendToBlockPendingTransactionsNode() {
+    final Account account = accounts.createAccount("account-one");
+    final Amount balance = Amount.ether(20);
+
+    TransferTransaction tx = accountTransactions.createTransfer(account, balance);
+
+    Hash txHash = aliceNode.execute(tx);
+
+    minerNode.execute(minerTransactions.minerStop());
```
macfarla, line 92: does this test rely on the mining stopping before the tx we just sent is mined? Won't that be flaky?

besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -697,6 +704,7 @@ public Runner build() {
       final Blockchain blockchain) {
     final Collection<EnodeURL> fixedNodes = getFixedNodes(bootnodesAsEnodeURLs, staticNodes);
 
+    // TODO: extract the different types of permissioning into their own plugin
```
macfarla, line 707: can this TODO be a GI

besu/src/test/java/org/hyperledger/besu/plugins/TestPermissioningPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+          if (sourceEnode.toString().contains(bobNode)
+              || destinationEnode.toString().contains(bobNode)) {
+
+            boolean isBobTalkingToAlice =
+                sourceEnode.toString().contains(aliceNode)
+                    || destinationEnode.toString().contains(aliceNode);
+            if (isBobTalkingToAlice) {
+              LOG.info("BLOCK CONNECTION from {}, to {}", sourceEnode, destinationEnode);
+            } else {
+              LOG.info("ALLOC CONNECTION from {}, to {}", sourceEnode, destinationEnode);
```
macfarla, line 54: s/ALLOC/ALLOW

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthPeerTest.java
```diff
@@ -402,18 +421,30 @@ private void messageStream(
   }
 
   private EthPeer createPeer() {
-    return createPeer(Collections.emptyList());
+    return createPeer(Collections.emptyList(), Collections.emptyList());
+  }
+
+  private EthPeer createPeerWithPermissioningProvider(
```
macfarla, line 427: nit: s/createPeerWithPermissioningProvider/createPeerWithPermissioningProviders

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthPeerTest.java
```diff
@@ -402,18 +421,30 @@ private void messageStream(
...
   }
 
   private EthPeer createPeer(final PeerValidator... peerValidators) {
-    return createPeer(Arrays.asList(peerValidators));
+    return createPeer(Arrays.asList(peerValidators), Collections.emptyList());
   }
 
-  private EthPeer createPeer(final List<PeerValidator> peerValidators) {
-    final Set<Capability> caps = new HashSet<>(singletonList(EthProtocol.ETH63));
-    final PeerConnection peerConnection = new MockPeerConnection(caps);
```
macfarla, line 414: caps not needed?

ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/node/provider/SyncStatusNodePermissioningProvider.java
macfarla: +1

jitpack.yml
```diff
@@ -0,0 +1,2 @@
+jdk:
```
macfarla, line 1: is this needed?

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/permissioning/NodeConnectionPermissioningProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+@FunctionalInterface
+public interface NodeConnectionPermissioningProvider {
+  /**
+   * Can be used to intercept the initial connection to a peer. Note that once a connection is
+   * established it's bidirectional.
+   *
+   * @param sourceEnode the originators enode
+   * @param destinationEnode the enode you are about to send to
+   * @return if you can connect
```
macfarla, line 27: if the connection is permitted

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/permissioning/NodeConnectionPermissioningProvider.java
```diff
@@ -0,0 +1,30 @@
...
+import org.hyperledger.besu.plugin.data.EnodeURL;
+
+@FunctionalInterface
+public interface NodeConnectionPermissioningProvider {
+  /**
+   * Can be used to intercept the initial connection to a peer. Note that once a connection is
+   * established it's bidirectional.
+   *
+   * @param sourceEnode the originators enode
+   * @param destinationEnode the enode you are about to send to
```
macfarla, line 26: the enode you are attempting to connect to

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/permissioning/NodeConnectionPermissioningProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.hyperledger.besu.plugin.data.EnodeURL;
+
+@FunctionalInterface
+public interface NodeConnectionPermissioningProvider {
+  /**
+   * Can be used to intercept the initial connection to a peer. Note that once a connection is
+   * established it's bidirectional.
+   *
+   * @param sourceEnode the originators enode
```
macfarla, line 25: s/originators/originator's

macfarla (commented):
some comments

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins/PermissioningPluginTest.java
antonydenyer: Good spot, thanks

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthPeerTest.java
antonydenyer: Surprisingly not! It doesn't do anything in this test. Changed it to use `mock(PeerConnection.class)` nothing is being validated against, it's just there to make you don't get null pointers.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/peers/EnodeURL.java
ricardolyn: I think is worth changing (IntelliJ can handle the 600 changes, but if you prefer, can be done in a separate PR).

should the interface `EnodeURL` live inside plugin data? maybe needs to be something more common (not only for plugins).
moreover, in the class `EnodeURL`, is there anything specific about this implementation, like used for `p2p` or `peer`? instead of using `Impl`.

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/PermissioningService.java
ricardolyn: non-blocking:

you allow also multiple providers. shouldn't be only 1? as if I follow your thoughts, then one provider can affect other one, so there can be still issues. 

plus, as you have multiple providers, might always good to unregister them even for memory management (except if we were using WeakReferences which will be cleared up automatically).

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/peers/EnodeURL.java
antonydenyer: > I think is worth changing (IntelliJ can handle the 600 changes, but if you prefer, can be done in a separate PR).
100% a separate or first to get rid of the noise on this commit.
> should the interface EnodeURL live inside plugin data? maybe needs to be something more common (not only for plugins).
Perhaps not, we would need to create a new module to contain such things. 
>  moreover, in the class EnodeURL, is there anything specific about this implementation, like used for p2p or peer? instead of using Impl.
Nope :disappointed:

ricardolyn (approved):
👍

macfarla:
OK I've looked at how we do this for other plugin API types eg Address. the origin "impl" just needs to implement the fully qualified plugin interface, and everywhere else we use the plugin interface. So SyncStatusNodePermissioningProvider can change to importing the new EnodeURL interface, instead of the existing impl. And then we only need the fully qualified type in one spot.

ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/node/provider/SyncStatusNodePermissioningProvider.java
```diff
@@ -18,18 +18,18 @@
 
 import org.hyperledger.besu.ethereum.core.Synchronizer;
 import org.hyperledger.besu.ethereum.p2p.peers.EnodeURL;
```
macfarla, line 20: so here import org.hyperledger.besu.plugin.data.EnodeURL instead

macfarla (commented):
see comments on changing the imports to the interface and then the need for the fully qualified type is only in one spot - and then I will be happy to approve :)

antonydenyer:
> > I will let @macfarla propose a new name for this if she agrees. I don't know enough about the code to propose that. my point is only the package import on each time doesn't seem good.
> 
> I can't think of anything better than EnodeURL... Also happy for that to be a separate PR



> see comments on changing the imports to the interface and then the need for the fully qualified type is only in one spot - and then I will be happy to approve :)

I don't think we can get away with not having a different name, it's just too messy! 

https://github.com/hyperledger/besu/pull/2283

macfarla (approved):

PR #2227, opened 2021-05-05
Fix tracer crash

matkt:
## PR description

Fix a crash with tracer on Staticcall and callcode

## Test performed 

- Ropsten 
>curl --location --request POST 'http://127.0.0.1:8545' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"debug_standardTraceBlockToFile","params":["0x0d9db1248c91a8a2f9b37442d50e7c38b2344853f35b1862e1cf6275ce792467", {"disableMemory":true}],"id":1}'

- Test with some FuzzyVM tests

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/StandardJsonTracer.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/StandardJsonTracer.java
@@ -128,14 +128,13 @@ public class StandardJsonTracer implements OperationTracer {
 
   private Gas computeCallGas(
       final AbstractCallOperation callOperation, final MessageFrame messageFrame) {
-    final GasCalculator gasCalculator = callOperation.gasCalculator();
     // as part of EIP 150 the returned costGas is gas - base * 63 / 64.
     // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md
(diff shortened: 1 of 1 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/StandardJsonTracer.java
```diff
@@ -128,14 +128,13 @@ public void traceExecution(
 
   private Gas computeCallGas(
       final AbstractCallOperation callOperation, final MessageFrame messageFrame) {
-    final GasCalculator gasCalculator = callOperation.gasCalculator();
     // as part of EIP 150 the returned costGas is gas - base * 63 / 64.
```
garyschulte, line 131: this is non-blocking, but should we change this comment also?  it isn't base cost * 63/64  but calculated operation cost * 63/64, right?

garyschulte, line 131: non-blocking feedback, we should change this comment to be "cost" rather than "base".  AFAICT call cost includes memory expansion, value transfer, and new account creation costs

garyschulte (approved):
concise and LGTM.

PR #2237, opened 2021-05-06
Add new fork for london, baikal and aleut

matkt:

## Changelog

### DIFF
--- a/config/src/main/java/org/hyperledger/besu/config/StubGenesisConfigOptions.java
+++ b/config/src/main/java/org/hyperledger/besu/config/StubGenesisConfigOptions.java
@@ -38,8 +38,10 @@ public class StubGenesisConfigOptions implements GenesisConfigOptions {
   private OptionalLong istanbulBlockNumber = OptionalLong.empty();
   private OptionalLong muirGlacierBlockNumber = OptionalLong.empty();
   private OptionalLong berlinBlockNumber = OptionalLong.empty();
+  private OptionalLong londonBlockNumber = OptionalLong.empty();
+
   // TODO EIP-1559 change for the actual fork name when known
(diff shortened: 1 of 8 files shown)


config/src/main/java/org/hyperledger/besu/config/StubGenesisConfigOptions.java
```diff
@@ -266,9 +282,8 @@ public OptionalLong getEcip1017EraRounds() {
     getIstanbulBlockNumber().ifPresent(l -> builder.put("istanbulBlock", l));
     getMuirGlacierBlockNumber().ifPresent(l -> builder.put("muirGlacierBlock", l));
     getBerlinBlockNumber().ifPresent(l -> builder.put("berlinBlock", l));
-    // TODO EIP-1559 change for the actual fork name when known
-    getEIP1559BlockNumber().ifPresent(l -> builder.put("aleutblock", l));
-
+    getLondonBlockNumber().ifPresent(l -> builder.put("londonBlock", l));
```
garyschulte, line 285: do we not also need baikalBlockNumber?  or are you thinking we will just going to use londonBlock only and keep aleut for backward compatibility?

matkt: No we did this for the Berlin testnets (yoloBlock) . In the genesis we will have baikalBlock and internally we consider it as a londonBlocK. We will remove this mechanism after testing.

For aleut we need it because, for now, is different from London. Aleut is a testnet with just EIP1559 and BASEFEE while baikal is a testnet that wants to be like london (with all EIPS).

For now for retro compatibility I prefer to keep it like that. To avoid consensus issue in this testnet

garyschulte (approved):

PR #2240, opened 2021-05-06
haveged + urandom

macfarla:

Install haveged on AT executor to increase entropy available for non-java processes.

Also setting secure source to /dev/urandom/ explicitly (but not sure that is relevant)

## Changelog


pinges (approved):
We should give this a go!

Issue #2250, opened 2021-05-08
Besu discovery packets are rejected by peers


atoulme:
### Description
Besu discovery packets are rejected by Core-Geth with the following message:

```
TRACE[05-01|00:33:00.410] >> PING/v4                               id=0cf4136259a482a1 addr=3.16.171.213:30303 err=nil
TRACE[05-01|00:33:00.509] << PONG/v4                               id=0cf4136259a482a1 addr=3.16.171.213:30303 err=nil
TRACE[05-01|00:33:01.010] >> FINDNODE/v4                           id=0cf4136259a482a1 addr=3.16.171.213:30303 err=nil
DEBUG[05-01|00:33:01.099] Bad discv4 packet                        addr=3.16.171.213:30303 err="bad hash"
TRACE[05-01|00:33:01.511] FINDNODE failed                          id=0cf4136259a482a1 failcount=384 dropped=false err="RPC timeout"
```

### Steps to Reproduce
1. Run a geth node next to Besu and connect to it for discovery
2. Observe the logs

**Expected behavior:**
Besu should communicate properly with Core-Geth.

**Actual behavior:**
All discovery packets are rejected.

**Frequency:**
Always

atoulme:
After some investigation, what appears to happen is that Besu sends packet which are longer than 1280 bytes. This is because most networks out there will drop bytes after that size.
The packet is truncated, and the hash doesn't check the packet contents anymore, so it's impossible to receive peers.

PR #2252, opened 2021-05-08
Fix discovery packet size

atoulme:

## PR description
Make sure Besu sends NEIGHBORS packets under MTU size.

## Fixed Issue(s)
#2250 

## Changelog

### DIFF
--- a/ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgentTest.java
+++ b/ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgentTest.java
@@ -224,19 +224,19 @@ public class PeerDiscoveryAgentTest {
     final IncomingPacket neighborsPacket = incomingPackets.get(0);
     assertThat(neighborsPacket.fromAgent).isEqualTo(agent);
 
-    // Assert that we only received 16 items.
+    // Assert that we only received 13 items.
     assertThat(neighborsPacket.packet.getPacketData(NeighborsPacketData.class).isPresent())
         .isTrue();
     final NeighborsPacketData neighbors =
         neighborsPacket.packet.getPacketData(NeighborsPacketData.class).get();
     assertThat(neighbors).isNotNull();
-    assertThat(neighbors.getNodes()).hasSize(16);
+    assertThat(neighbors.getNodes()).hasSize(13);
 
-    // Assert that after removing those 16 items we're left with either 4 or 5.
+    // Assert that after removing those 13 items we're left with either 4 or 5.
     // If we are left with 5, the test peer was returned as an item, assert that this is the case.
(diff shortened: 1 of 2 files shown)


ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgentTest.java
```diff
@@ -224,19 +224,19 @@ public void neighborsPacketLimited() {
...
     assertThat(neighborsPacket.fromAgent).isEqualTo(agent);
 
-    // Assert that we only received 16 items.
+    // Assert that we only received 13 items.
     assertThat(neighborsPacket.packet.getPacketData(NeighborsPacketData.class).isPresent())
         .isTrue();
     final NeighborsPacketData neighbors =
         neighborsPacket.packet.getPacketData(NeighborsPacketData.class).get();
     assertThat(neighbors).isNotNull();
-    assertThat(neighbors.getNodes()).hasSize(16);
```
shemnon, line 233: If the purpose of the change is to ensure packet size on the wire is < 1144 bytes then the unit test should test the encoded packet size.

atoulme: Added something, though this test doesn't encode packets, bypassing the whole UDP transport.

shemnon (approved):

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgentTest.java
```diff
@@ -224,19 +224,20 @@ public void neighborsPacketLimited() {
...
+    assertThat(neighbors.getNodes()).hasSize(13);
+    assertThat(neighborsPacket.packet.encode().length()).isLessThanOrEqualTo(1280); // under max MTU
 
-    // Assert that after removing those 16 items we're left with either 4 or 5.
+    // Assert that after removing those 13 items we're left with either 4 or 5.
```
shemnon, line 236: nit.  7 or 8
```suggestion
    // Assert that after removing those 13 items we're left with either 7 or 8.
```

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgentTest.java
```diff
@@ -224,19 +224,20 @@ public void neighborsPacketLimited() {
...
     // If we are left with 5, the test peer was returned as an item, assert that this is the case.
```
shemnon, line 237: nit, 8
```suggestion
    // If we are left with 8, the test peer was returned as an item, assert that this is the case.
```

PR #2261, opened 2021-05-10
Bonsai trie : clone updater during persist

matkt:
## PR description

Clone the updater during the persistence of a new block in order to prevent the updater from being modified between the generation of the stateroot and the creation of the layer. The clone will be a temporary clone in the persistent method scope. Bug found during a mainnet synchronization.

Fix a bug that occurred during a rollback of the persistent state. Indeed a null or ZERO slot have the same meaning but was considered different during the rollback resulting in an error.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -639,6 +649,13 @@ private void rollStorageChange(
     }
   }
 
+  private boolean isSlotEquals(final UInt256 expectedValue, final UInt256 existingSlotValue) {
+    final UInt256 sanitizedExpectedValue = (expectedValue == null) ? UInt256.ZERO : expectedValue;
+    final UInt256 sanitizedExistingSlotValue =
+        (existingSlotValue == null) ? UInt256.ZERO : existingSlotValue;
```
garyschulte, line 655: Optional.ofNullable is more readable than ternary's IMO :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -47,20 +47,30 @@
 public class BonsaiWorldStateUpdater extends AbstractWorldUpdater<BonsaiWorldView, BonsaiAccount>
     implements BonsaiWorldView {
 
-  private final Map<Address, BonsaiValue<BonsaiAccount>> accountsToUpdate = new HashMap<>();
-  private final Map<Address, BonsaiValue<Bytes>> codeToUpdate = new HashMap<>();
-  private final Set<Address> storageToClear = new HashSet<>();
+  private Map<Address, BonsaiValue<BonsaiAccount>> accountsToUpdate = new HashMap<>();
+  private Map<Address, BonsaiValue<Bytes>> codeToUpdate = new HashMap<>();
+  private Set<Address> storageToClear = new HashSet<>();
```
garyschulte, line 52: non-blocking feedback:  we could have a private constructor that takes these params, and keep them final.  We'd have to add a matching protected constructor in AbstractWorldStateUpdater too though.

garyschulte (approved):
LGTM.  Non-blocking feedback just so you know I did 👀    :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
matkt: oops I wanted to make the change but you were faster than me :sweat_smile:

PR #2265, opened 2021-05-11
Qbft invalid timestamp

jframe:
## PR description
Fix error where we get an invalid header timestamp occasionally. This seems to be due slight deviation from the block period. It's not likely we can ever get the timing exact as we need determine the delay before scheduling the block expiry event so there can be a very discrepancy.

## Changelog


usmansaleem (approved):
LGTM

PR #2269, opened 2021-05-11
Split test acceptanceTests

matkt:

## PR description

Even if we set up tests in parallel we just create several docker that are on the same machine and that therefore queries the same source of entropy. I tried to split the test into different steps.

I  tried 5 times and it's fine everytime

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,7 +208,7 @@ jobs:
           no_output_timeout: 40m
           command: |
             CLASSNAMES=$(circleci tests glob "acceptance-tests/tests/src/test/java/**/*.java" \
-              | grep "/privacy/"
+              | grep "/privacy/" \
               | sed 's@.*/src/test/java/@@' \
               | sed 's@/@.@g' \
               | sed 's/.\{5\}$//' \
@@ -231,7 +231,7 @@ jobs:
           no_output_timeout: 40m
           command: |
             CLASSNAMES=$(circleci tests glob "acceptance-tests/tests/src/test/java/**/*.java" \
-              | grep "/permissioning/"
+              | grep "/permissioning/" \
               | sed 's@.*/src/test/java/@@' \
               | sed 's@/@.@g' \
               | sed 's/.\{5\}$//' \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -185,6 +185,53 @@ jobs:
...
+              | sed 's@/@.@g' \
+              | sed 's/.\{5\}$//' \
+              | circleci tests split --split-by=timings --timings-type=classname)
+            # Format the arguments to "./gradlew test"
+            GRADLE_ARGS=$(echo $CLASSNAMES | awk '{for (i=1; i<=NF; i++) print "--tests",$i}')
+            ./gradlew --no-daemon acceptanceTest $GRADLE_ARGS
+      - capture_test_results
+      - capture_test_logs
+
+  acceptanceTestsPrivacy:
```
macfarla, line 199: make this a required step

.circleci/config.yml
```diff
@@ -185,6 +185,53 @@ jobs:
...
+              | sed 's@/@.@g' \
+              | sed 's/.\{5\}$//' \
+              | circleci tests split --split-by=timings --timings-type=classname)
+            # Format the arguments to "./gradlew test"
+            GRADLE_ARGS=$(echo $CLASSNAMES | awk '{for (i=1; i<=NF; i++) print "--tests",$i}')
+            ./gradlew --no-daemon acceptanceTest $GRADLE_ARGS
+      - capture_test_results
+      - capture_test_logs
+
+  acceptanceTestsPermissioning:
```
macfarla, line 222: make this a required step

.circleci/config.yml
```diff
@@ -185,6 +185,53 @@ jobs:
...
+    parallelism: 8
```
macfarla, line 223: looking at the output, this step now only takes 3 min so could prob reduce the concurrency but we can tune that later (not sure how this affects the credit usage)

ajsutton (commented):
This doesn't actually change where the containers come from - whether you use parallel or separate steps CircleCI will just allocate an executor from its pool. 

All this does is make the config more complex and makes the balancing of tests across containers less effective.

PR #2281, opened 2021-05-12
try to use the hostName property from log4j2

atoulme:

## PR description
This adds the ability to default to the `hostName` property, filled by log4j2, before we default to sending localhost.

## Changelog


usmansaleem (approved):
LGTM

PR #2282, opened 2021-05-13
Modify Gas estimation logic for GoQuorum mode

macfarla:
If in GoQuorum privacy mode, use a fully filled data field to get max value for gas estimation.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -157,12 +163,25 @@ public TransactionSimulator(
...
     final Wei gasPrice = callParams.getGasPrice() != null ? callParams.getGasPrice() : Wei.ZERO;
     final Wei value = callParams.getValue() != null ? callParams.getValue() : Wei.ZERO;
-    final Bytes payload = callParams.getPayload() != null ? callParams.getPayload() : Bytes.EMPTY;
+
+    // If GoQuorum privacy enabled, and value = zero, do simulation with max non-zero bytes.
+    // It is possible to have a data field that has a lower intrinsic value than the PTM hash
+    // so this checks the tx as if we were to place a PTM hash (with all non-zero values).
+    // This means a potential over-estimate of gas, rather than the exact cost to run right now.
+    final Bytes payload =
+        GoQuorumOptions.goQuorumCompatibilityMode && value.isZero()
```
pinges, line 172: could we only do that for private Txs?  What are they doing in Quorum?

pinges (commented):
Please check my comment ..

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
macfarla: I think the idea is that you give a gas estimate that will work whether it ends up being a private or a public tx. So in GoQuorum it's the same logic for all tx https://github.com/ConsenSys/quorum/pull/644/files

pinges (approved):
LGTM

PR #2284, opened 2021-05-13
refactor: expose EnodeUrl to plugins

antonydenyer:
Couldn't figure out a nice way of not having EnodeURLImpl.

If we kept it as EnodeURL we could add static imports in for things like `fromURI` but the problem is they are mostly used as method references. So you can either have method references with the class name or lambdas with a static import! 

I think it's the least worse option at the moment.


macfarla (approved):
LGTM as the least worst option

PR #2293, opened 2021-05-18
parallelism on unit tests and ref tests

macfarla:

Add parallelism to unit tests and ref tests since they now take the longest

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,7 @@ jobs:
             build\distributions\besu\bin\besu.bat --version
 
   unitTests:
+    parallelism: 2
     executor: besu_executor_xl
     steps:
       - prepare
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -127,6 +127,7 @@ jobs:
             build\distributions\besu\bin\besu.bat --version
 
   unitTests:
+    parallelism: 2
```
ajsutton, line 130: This will just run all the tests on two boxes because there's nothing that's splitting the tests across them.  Se the `CLASSNAMES=$(circleci tests glob` stuff in the `acceptanceTests` task.

ajsutton (changes requested):

.circleci/config.yml
macfarla: yes of course. ok prob isn't worth spending time on now then.

PR #2296, opened 2021-05-19
Eip 1559 tx sorting refactor

garyschulte:
## PR description

this is a refactor of PendingTransactions, moving the actual queue out into an interface, and adds a basic implementation that prioritizes by transaction revenue.

The basic implementation keeps just a single transaction pool, without other pre-optimized indexed collections, and simplified locking.  We should quantify the performance of this approach before using this impl.  PR marked as DO NOT MERGE for this reason.

## Changelog


garyschulte:
stale

PR #2302, opened 2021-05-20
errata discovered in 21.1.6

garyschulte:

## PR description
A couple quick fixes for bugs I found In scale testing 21.1.6 pending transactions:
* baseFee = null was not handled correctly in PendingTransactions.updateBaseFee for a fork block
* the transaction gas budget calculator was 2x'ing the already 2x'd gasLimit at and past fork block

we should probably write some tests that cover these cases, but it is late and the release is already out there...

## Changelog


matkt (approved):
LGTM, Maybe we need to update the CHANGELOG for that

PR #2305, opened 2021-05-20
add log blooms for quorum private transactions

pinges:
When private transactions are executed in goquorum compatible mode the receipts of these private transactions are stored in place of the public marker transaction. To make these receipts discoverable by calls that use the bloom filters the blooms filters of the private transactions have to be combined with the blooms of the public transactions in the right places. 

### DIFF
--- a/container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTestBase.java
+++ b/container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTestBase.java
@@ -42,7 +42,7 @@ import org.web3j.quorum.Quorum;
 
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class ContainerTestBase {
-  private final String besuImage = System.getProperty("containertest.imagename");
+  private final String besuImage = "hyperledger/besu:21.1.7-SNAPSHOT";
   private final String goQuorumVersion = "21.1.0";
   private final String tesseraVersion = "21.1.1";
 
@@ -157,6 +157,7 @@ public class ContainerTestBase {
         .withClasspathResourceMapping(hostGenesisPath, besuContainerGenesisPath, BindMode.READ_ONLY)
         .withClasspathResourceMapping(hostKeyPath, containerKeyPath, BindMode.READ_ONLY)
         .withCommand(
+            "--auto-log-bloom-caching-enabled=false",
             "--genesis-file",
             besuContainerGenesisPath,
             "--network-id",
--- a/container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTestBase.java
+++ b/container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTestBase.java
@@ -42,7 +42,9 @@ import org.web3j.quorum.Quorum;
 
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class ContainerTestBase {
-  private final String besuImage = "hyperledger/besu:21.1.7-SNAPSHOT";
+  // A docker image can be built using gradlew distDocker
+  // To use that local image change besuImage to "hyperledger/besu:<VERSION>-SNAPSHOT"
+  private final String besuImage = System.getProperty("containertest.imagename");
   private final String goQuorumVersion = "21.1.0";
   private final String tesseraVersion = "21.1.1";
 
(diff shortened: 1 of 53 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -121,6 +166,31 @@ public Hash getHash() {
     return hash.get();
   }
 
+  public LogsBloomFilter getLogsBloom(final boolean forFilter) {
```
macfarla, line 169: would it be clearer to name this param forFilterPrivateLogs if it only applies when there's private logs?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeaderBuilder.java
```diff
@@ -87,7 +87,7 @@ public static BlockHeaderBuilder fromHeader(final BlockHeader header) {
   }
 
   public static BlockHeaderBuilder fromBuilder(final BlockHeaderBuilder fromBuilder) {
-    BlockHeaderBuilder toBuilder =
+    final BlockHeaderBuilder toBuilder =
```
macfarla, line 90: why is this called toBuilder when the method is called fromBuilder? (I realise you did not change this...)

macfarla (commented):
I think it makes sense. Do we need additional tests?

macfarla (approved):
overall looks good. some minor comments

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
pinges: changed it to "addPrivateLoom".

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeaderBuilder.java
pinges: I think it makes sense. This is the builder that we are copying to. fromBuilder is the builder that we are copying from.

container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTestBase.java
```diff
@@ -42,7 +42,7 @@
 
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class ContainerTestBase {
-  private final String besuImage = System.getProperty("containertest.imagename");
+  private final String besuImage = "hyperledger/besu:21.1.7-SNAPSHOT";
```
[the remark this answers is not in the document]
pinges: Good catch! No that was my change to make it work locally!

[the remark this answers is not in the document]
pinges: I've added a comment on how to use a locally built image.

container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTests.java
```diff
@@ -184,12 +234,31 @@ public void contractShouldBeDeployedOnlyToBesuNode() throws IOException, Transac
             goQuorumPollingTransactionReceiptProcessor,
             logValue);
 
+    while (!checked.get()) {
+      Thread.sleep(1000);
```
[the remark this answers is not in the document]
pinges: Very good catch! Changed it so it would not wait forever

container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTests.java
```diff
@@ -184,12 +234,36 @@ public void contractShouldBeDeployedOnlyToBesuNode() throws IOException, Transac
             goQuorumPollingTransactionReceiptProcessor,
             logValue);
 
+    int secondsWaited = 0;
+    while (!checked.get()) {
+      Thread.sleep(1000);
+      secondsWaited++;
+      if (secondsWaited > 30) {
+        fail("Waited more than 30 seconds for log.");
```
macfarla, line 242: if we fail here, does the dispose get skipped - should it be in a finally{}

macfarla (approved):

container-tests/tests/src/test/java/org/hyperledger/besu/tests/container/ContainerTests.java
pinges: Does not really matter. The nodes are going to be stopped!

PR #2311, opened 2021-05-20
GoQuorum: Gas estimate to cover PMT

macfarla:
If GoQuorum mode and value = 0, modify tx simulator result to return estimate which will cover a PMT transaction

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/GasCalculator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/GasCalculator.java
@@ -440,7 +440,5 @@ public interface GasCalculator {
   };
 
   // what would be the gas for a PMT with hash of all non-zeros
-  default Gas getMaximumPmtCost() {
-    return Gas.ZERO;
-  }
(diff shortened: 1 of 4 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/GasCalculator.java
```diff
@@ -438,4 +438,9 @@ default Gas modExpGasCost(final Bytes input) {
   default long getMaxRefundQuotient() {
     return 2;
   };
+
+  // what would be the gas for a PMT with hash of all non-zeros
+  default Gas getMaximumPmtCost() {
```
pinges, line 443: Do we really need the default implementation?

pinges (approved):
Have one comment with a question ...

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/GasCalculator.java
macfarla: removed

PR #2342, opened 2021-05-31
Fix invalid upfrontgascost modification for EIP1559

matkt:

## PR description

Obviously the last 1559 modification made causes a consensus issue on Baikal while the 4 rules are valid
- The block http://18.224.51.102:3000/block/14498
- The modification https://github.com/hyperledger/besu/pull/2338/files#diff-8f3af0f8da1b352c31d41501c14cf92a66947bc3dc2a2f0d617c1691c380c70fR539
This change decrement the balance of the wallet here https://github.com/hyperledger/besu/blob/3267501c8cd2a1cd116b8b295e5baedab12bd3b5/[…]erledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java while we do not know yet if it is maxFeePerGas that will be used. 
And we already calculate the valid value needed for upfrontgascost here https://github.com/hyperledger/besu/blob/3267501c8cd2a1cd116b8b295e5baedab12bd3b5/[…]erledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java

This PR also fix somme issues regarding the transaction pool for eip1559 and a tracing issue

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -536,8 +536,10 @@ public Wei getUpfrontGasCost() {
    * @return the up-front cost for the gas the transaction can use.
    */
   public Wei getUpfrontGasCost(final Wei gasPrice) {
-    return Wei.of(getGasLimit())
-        .multiply(Wei.of(getMaxFeePerGas().orElse(gasPrice).getAsBigInteger()));
+    if (gasPrice == null || gasPrice.isZero()) {
+      return Wei.ZERO;
+    }
```
RatanRSur, line 541: I'm trying to find where this could be null from? I'm hoping this isn't test code concerns leaking into production code.

matkt: I mostly revert this part because it should not change with my fix. but I think that if we have to remove this check it will be preferable to remove it in another PR non EIP1559 in order to identify it more easily if a regression that we would not have seen appears

garyschulte: it could logically be null though right?  gasPrice should be absent from a 1559 transaction

RatanRSur (approved):

abdelhamidbakhta (approved):
Looks good to me.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
matkt: Yes I think it can maybe possible, I just had a usecase when the gaslimit is null

PR #2361, opened 2021-06-02
ci/cd: Require Quorum Acceptance Tests to pass

ricardolyn:
Signed-off-by: Ricardo Silva <ricardo.silva@consensys.net>

### Summary

As the work is done for the privacy interop, we remove now the ignore of the results so if it fails, it breaks the build.

## Changes

* Removed `|| true` to make the job fail if the tests fail.
* Make the publish of docker images depend on the Quorum AT.
* Add the `nosupport` tag as privacy polishing to be done later.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -205,7 +205,7 @@ jobs:
       - run:
           name: Quorum Acceptance Tests (failures ignored)
           no_output_timeout: 40m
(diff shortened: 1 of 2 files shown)


mark-terry (approved):
LGTM.

macfarla (approved):

PR #2363, opened 2021-06-02
[Cleanup] Remove Gas Budget Calculator

RatanRSur:
## PR description
Since there are no separate gas budgets for eip-1559 and non-eip-1559 txs. We can return to the simpler method of just comparing against the gas limit.

This is a non-functional change.

## Changelog


garyschulte (approved):
Love all the 🟥

PR #2366, opened 2021-06-02
Fixes for 1559 retesteth behaviors

garyschulte:

quick pr to unblock retesteth testing, remove aleut and calaveras from retesteth chain genesis.  

Aleut config was an artifact from when we did not have London configs.


matkt (approved):
LGTM

PR #2374, opened 2021-06-04
1559 blockchain reftests

### LINKED ISSUE #2350: Ensure besu is passing reference tests with retesteth
todo: add details

garyschulte:
## PR description
* change transaction pool behavior to not reject transactions based on chain head's baseFee
* add short form of Wei for maxFeePerGas and maxPriorityFeePerGas
* ~~address edge case of allowable maximum delta gasLimit per block~~
* update reference tests module to latest

This pr gets besu passing all of the London blockchain tests using the retesteth engine.  There will be a follow up pr to address issues the few remaining GeneralStateTests failures for Berlin and London

## Fixed Issue(s)

fixes #2350 for London Blockchain tests

## Changelog


ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/GasLimitElasticityValidationRuleTest.java
```diff
@@ -55,19 +55,19 @@
     return Arrays.asList(
         new Object[][] {
           {20000000, 10000000, 10, true},
-          {20019530, 10000000, 10, true},
```
matkt, line 58: This test was link to this PR https://github.com/ethereum/EIPs/pull/3566
the same result as geth . Did this test changed ?

matkt (changes requested):
I'm not sure about the modification regarding  the GasLimitRangeAndDelta but maybe I missed a modification

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/GasLimitElasticityValidationRuleTest.java
garyschulte: These changes were necessary due to the change from <= to < for the gas limit delta rules.  if these are the same cases geth is using, I will double check the spec to ensure this is the correct behavior at the bounds of the max allowable gasLimit delta.  

Either these cases are right, or the reference test cases are right.

matkt: If there is this kind of modification  it might be necessary to wrap it on a check is1559Block to not impact the blocks before london IMO but maybe I miss something. And to be sure that Geth has the same result? Or to check  if there is a new version of the test  ? 🤔

matkt (approved):
LGTM when the pipeline will be green

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/GasLimitElasticityValidationRuleTest.java
garyschulte: thanks for catching this.  It was indeed a problem in the reference tests rather than our implementation.

Issue #2370, opened 2021-06-03
[Discussion] Unrestricted Privacy Encryption options/recomendations


antonydenyer:
### Description
As an enterprise user, I want to be able to encrypt `unrestricted` private transactions. Unrestricted privacy is where the private/side-channel transaction is distributed to all participants in the network but only some can decrypt it. Besu will implement this using a pre-compile in the same way as `restricted` privacy. The key difference is that the payload for the transaction will be stored on-chain in the data property (aka the private payload) https://github.com/hyperledger/besu/pull/2324

There are numerous tactics and options on how best to encrypt private data. I don't think besu can recommend best practice with this regard, the best option will depend on the use case and trade-offs the user is willing to accept.

As such, I propose that we provide a plugin hook to allow users to do this. This still comes with it's own set of trade-offs:

### 1 - bytes in bytes out
Besu could pass the rlp encoded `PrivateTransaction` to the plugin and receive back the encrypted private payload.

```java
  Bytes decrypt(final Bytes input);
  Bytes encrypt(final Bytes input);
```
This provides the most flexible option for the user. However, if they wanted to do anything useful they would need to deserialize the input to find out what the `privacyGroupId` is and encrypt the data on a per-group basis.

### 2 - bytes and meta in bytes out
In addition, besu could pass some metadata with the raw private transaction to help with encryption. This is essentially what happens with the existing `restricted` privacy implementation. 

```java
  Bytes decrypt(final Bytes input);
  Bytes encrypt(final Bytes input, Bytes privacyGroupId);
```
It gives the user a bit more help with encryption. But no extra help with decryption. 

### 3 - PrivateTransaction in bytes out
This would be the easiest from the user point of view but the hardest from an implementation point of view. We would be opening up the internals of besu to the plugin.

```java
  PrivateTransaction decrypt(final Bytes input);
  Bytes encrypt(final PrivateTransaction input);
```

#### Serialization
When upgrading a network any changes to `PrivateTransaction` would need to be non-breaking. This is currently the case when using an enclave, except that you can 'migrate' data in an enclave but you can't 'migrate' data on-chain. When adding properties to `PrivateTransaction` sensible defaults would be required.

alberto-hernandez:
Hi Anthony,
I will add to the options a more generall approach for the encryption/decription process. In all the options it's supposed that it only will be used one key for the whole process and this could not be flexible for alll scenarios such us rotate the key every N blocks, etc.

[...]

PR #2377, opened 2021-06-04
Privacy payload plugin

antonydenyer:
The Privacy Plugin Service will allow people to implement their own strategies for private transactions. At a very high-level besu will delegate the encoding and decoding of the private marker transaction payload to the plugin. This gives the power to the implementor to decide on how to distribute the private transaction. They could use a private enclave network and notarise the location (as happens currently with Tessera) or they could use some other strategy like onchain encryption.


antonydenyer:
Implementation for https://github.com/hyperledger/besu/issues/2370

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/PrivateTransaction.java
```diff
@@ -0,0 +1,149 @@
...
+   * The public key of the recipients of this private transaction. This will
+   *
+   * <p>Only present if no privacy group {@link #getPrivacyGroupId} is present.
+   *
+   * @return public keys of the recipients
+   */
+  Optional<List<Bytes>> getPrivateFor();
+
+  /**
+   * The identifier for the group of identifier participants.
```
macfarla, line 133: The identifier for the group of participants.

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/PrivateTransaction.java
```diff
@@ -0,0 +1,149 @@
...
+   * @return the quantity of Wei per gas unit paid.
+   */
+  Quantity getGasPrice();
+
+  /**
+   * A scalar value equal to the maximum amount of gas that should be used in executing this
+   * transaction. This is paid up-front, before any computation is done and may not be increased
+   * later.
+   *
+   * @return the maximum amount of gas that should be used in executing this * transaction.
```
macfarla, line 51: the maximum amount of gas that should be used in executing this transaction

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/PrivateTransaction.java
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Value corresponding to the 'V' component of the signature of the transaction.
+   *
+   * @return the 'V' component of the signature
+   */
+  BigInteger getV();
+
+  /**
+   * Value corresponding to the 'V' component of the signature of the transaction.
```
macfarla, line 79: s/V/R

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/PrivateTransaction.java
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * @return the 'V' component of the signature
```
macfarla, line 81: s/V/R

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/PrivateTransaction.java
```diff
@@ -0,0 +1,149 @@
...
+   */
+  BigInteger getR();
+
+  /**
+   * Value corresponding to the 'V' component of the signature of the transaction.
```
macfarla, line 86: s/V/S

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/PrivateTransaction.java
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * @return the 'V' component of the signature
```
macfarla, line 88: s/V/S

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/PrivateTransaction.java
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * The public key of the sender of this private transaction.
+   *
+   * @return public key of the sender
+   */
+  Bytes getPrivateFrom();
+
+  /**
+   * The public key of the recipients of this private transaction. This will
```
macfarla, line 124: The public keys of the recipients of this private transaction.

macfarla:
Overall looks good. Some comments and spotless is complaining.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/UnrestrictedPrivacyPrecompiledContract.java
```diff
@@ -22,21 +22,24 @@
...
 import org.hyperledger.besu.ethereum.vm.MessageFrame;
 
+import java.util.Optional;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
 
 public class UnrestrictedPrivacyPrecompiledContract extends PrivacyPrecompiledContract {
```
pinges, line 35: Maybe we should name this something like PluginUsingPrivacyPrecompiledContract.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/UnrestrictedPrivacyPrecompiledContract.java
```diff
@@ -46,14 +49,22 @@ public Bytes compute(final Bytes input, final MessageFrame messageFrame) {
       return Bytes.EMPTY;
     }
 
-    final Hash pmtHash = messageFrame.getTransactionHash();
+    final Optional<org.hyperledger.besu.plugin.data.PrivateTransaction> pluginPrivateTransaction =
+        privacyParameters
+            .getPrivacyService()
+            .getUnrestrictedPayloadEncryptionProvider()
+            .decryptMarkerPayload(
```
pinges, line 56: this method could be renamed to something like "getPrivateTransaction" ...

pinges (commented):
Not quite through all the files, but will post the current comments now!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/UnrestrictedPrivacyPrecompiledContract.java
```diff
@@ -46,14 +49,22 @@ public Bytes compute(final Bytes input, final MessageFrame messageFrame) {
...
+                messageFrame.getBlockHeader().getNumber(), messageFrame.getTransaction());
```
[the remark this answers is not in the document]
antonydenyer: Agreed and that's what I had previously, however, it's become apparent that to facilitate decryption you might want to find out some information about the originating transaction. Specifically the `sender` of the pmt , now the question is would that be enough? I don't know, so figured just put the full transaction in there and be done with it. The alternative is we have something like `decryptMarkerPayload(blockNumber, sender, payload)` and hope that it's enough for the plugin implementor.

besu/src/main/java/org/hyperledger/besu/services/PrivacyServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import org.hyperledger.besu.plugin.services.privacy.PrivacyPayloadEncryptionProvider;
+
+public class PrivacyServiceImpl implements PrivacyService {
+  private PrivacyPayloadEncryptionProvider privacyPayloadEncryptionProvider =
+      new UnencryptedPayloadEncryptionService();
+  private PrivacyGroupAuthProvider privacyGroupAuthProvider =
+      (privacyGroupId, privacyUserId, blockNumber) -> true;
+
+  @Override
+  public void setUnrestrictedPayloadEncryptionProvider(
```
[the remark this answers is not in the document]
antonydenyer: PMT Payload Provider?

pinges (commented):
I'm not sure whether we want to support legacy private txs ...
But I have a few other comments as well!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PluginPrivacyController.java
```diff
@@ -119,9 +124,16 @@ public Transaction createPrivateMarkerTransaction(
...
+
+    if (pluginPrivateTransaction.isEmpty()) {
+      return Optional.empty();
+    }
 
-    final PrivateTransaction privateTransaction = PrivateTransaction.readFrom(input);
+    final PrivateTransaction privateTransaction = readFrom(pluginPrivateTransaction.get());
 
     final String internalPrivacyGroupId =
         privateTransaction.determinePrivacyGroupId().toBase64String();
```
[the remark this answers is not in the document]
pinges: s.a.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PluginPrivacyController.java
```diff
@@ -175,6 +187,9 @@ public long determineEeaNonce(
       final String privacyUserId) {
```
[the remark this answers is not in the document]
pinges: s.a.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PluginPrivacyController.java
```diff
@@ -119,9 +124,16 @@ public Transaction createPrivateMarkerTransaction(
     final BlockHeader blockHeader =
         blockchain.getBlockHeader(transactionLocation.getBlockHash()).orElseThrow();
 
-    final BytesValueRLPInput input = new BytesValueRLPInput(transaction.get().getPayload(), false);
+    final Optional<org.hyperledger.besu.plugin.data.PrivateTransaction> pluginPrivateTransaction =
+        privacyPluginService
+            .getPayloadProvider()
+            .getPrivateTransactionFromPayload(blockHeader.getNumber(), transaction.get());
```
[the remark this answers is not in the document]
antonydenyer: We definitely need blockNumber in the plugin for ~decrypting~ `getPrivateTransactionFromPayload`. You're right when we ~encrypt~ `generateMarkerPayload` the plugin can not know for certain. 

Let's say we have a simple scheme to use a different key every 1000 blocks. When you call `generateMarkerPayload` the plugin could ask the node for the block height (this would, of course, be increasing on a different thread so you can not use it as a guarantee) if the height is after 1000 use key A and if it's after 2000 use key B. It could also reject the transaction if the block height is too close to the crossover. 

When you come to ~decrypt~ `getPrivateTransactionFromPayload` you need the block number for the transaction. It's not in `Transaction` you could get it using `eth_getTransactionByHash`. 

I think it's useful, but happy to remove it as there are ways to get the info you need.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PluginPrivacyController.java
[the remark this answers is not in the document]
antonydenyer: It's easy to have in, but understand if it's a long term goal to deprecate `eea_sendRawTransaction` with `privateFrom` and `privateFor` then we should not continue to implement it.

besu/src/main/java/org/hyperledger/besu/services/PrivacyPluginPluginServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    return privacyPluginPayloadProvider;
+  }
+
+  @Override
+  public void setPrivacyGroupAuthProvider(final PrivacyGroupAuthProvider privacyGroupAuthProvider) {
+    this.privacyGroupAuthProvider = privacyGroupAuthProvider;
+  }
+
+  @Override
+  public PrivacyGroupAuthProvider getPrivacyGroupAuthProvider() {
```
pinges, line 52: Could you please check whether we need a similar check that we have for the getPayloadProvider. In BesuCommand there is a check to make sure that we have a PayloadProvider, but no check for the AuthProvider.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PluginPrivacyController.java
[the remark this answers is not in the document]
pinges: I think we should remove it. In the generateMarkerPayload(privateTransaction, privacyUserId) you are not passing it in and in general I do not like it if something is only "about" right. I guess if you do want to rotate keys you could always do that based on time, or what sometimes is done, based on the amount of data that has been encrypted with a key. 
In case you rotate keys you would generally provide some kind of key id in the payload to enable the receiving side to find the right key.

pinges (approved):
I've left two comments. Please have a look at them.

besu/src/main/java/org/hyperledger/besu/services/PrivacyPluginPluginServiceImpl.java
antonydenyer: Maybe, I defaulted it to `true` on the assumption most people won't be using it. I figured from an implementor's point of view it was the best option, works out of the box but can override. 

https://github.com/hyperledger/besu/blob/9526379221a1c0d84ce9c7e550cc221c60530e3e/besu/src/main/java/org/hyperledger/besu/services/PrivacyPluginPluginServiceImpl.java#L30

pinges (approved):
LGTM

PR #2379, opened 2021-06-07
QBFT to use fixed IBFT mixhash value for headers

jframe:
## PR description
Use the fixed IBFT mixedHash value in QBFT as this is what is being done in Quorum.

## Changelog


usmansaleem (approved):
LGTM

PR #2380, opened 2021-06-07
eth_getCode: add support for GoQuorum private tx

macfarla:
Attempt to retrieve from GoQuorum private state first, otherwise get from public state as before.

I have tested this by:
Node1 deploys a contract, private for Node4. Nodes 1 and 4 can get the code; Nodes 2 and 3 do not. getCode for public tx works as before (all nodes get the right value). Attempting to getCode for non-existent address gives 0x.

## Changelog


mark-terry (approved):
LGTM.

PR #2386, opened 2021-06-08
Fix nullpointer error with trace_block

matkt:

## PR description

This PR fixes a nullpointer that there was on the transaction https://rinkeby.etherscan.io/tx/0x9eec6e2e4a5356be4e930f448d4a772869b1b5a4f2c0d9a1e98fc6faed5d0f4d

>curl --data '{"method":"trace_block","params":["84CD39"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST http://besu:8545 | jq

## Test performed

Tested and reproduced locally

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java
```diff
@@ -465,20 +465,29 @@ private static String firstNonNull(final String defaultValue, final String... va
...
 
   private static FlatTrace.Context handleHalt(
+      final List<FlatTrace.Builder> flatTraces,
       final Deque<FlatTrace.Context> tracesContexts,
       final FlatTrace.Context currentContext,
       final TraceFrame traceFrame) {
-    final FlatTrace.Builder traceFrameBuilder = currentContext.getBuilder();
+    final FlatTrace.Builder traceFrameBuilder;
+    if (currentContext == null) {
+      traceFrameBuilder = flatTraces.get(flatTraces.size() - 1);
```
garyschulte, line 474: looks like a scary assumption, but afaict in the current usage we should always have at least 1 element in the list

garyschulte (approved):
lgtm, but might want to wait for feedback from Rai

abdelhamidbakhta (approved):
Looks good to me.

PR #2391, opened 2021-06-08
Add ValidateFirstOutboundMessage just before DeFramer

perusworld:
This takes care of the current requirements as well as provide the possibility of adding other channel handlers before that like for example a TLS handler


atoulme (approved):
Changes are ok with me. They do not alter the functionality. Curious what you’ll do next ;)

lucassaldanha (approved):

Issue #2389, opened 2021-06-08
1559-Tx Null-Pointer-Exception


MariusVanDerWijden:
### Description
When sending eip-1559 transactions without gas price over RPC, the RPC fails with an null pointer exception
```
2021-06-08 19:20:55.914+02:00 | vert.x-worker-thread-18 | ERROR | JsonRpcHttpService | Error processing JSON-RPC requestBody
java.lang.NullPointerException: Cannot invoke "org.hyperledger.besu.ethereum.core.Wei.toLong()" because the return value of "org.hyperledger.besu.ethereum.core.Transaction.getGasPrice()" is null
	at org.hyperledger.besu.ethereum.api.query.BlockchainQueries.lambda$gasPrice$29(BlockchainQueries.java:821)
	at java.base/java.util.stream.ReferencePipeline$5$1.accept(ReferencePipeline.java:229)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:658)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:274)
	at java.base/java.util.stream.LongPipeline$1$1.accept(LongPipeline.java:177)
	at java.base/java.util.stream.Streams$RangeLongSpliterator.forEachRemaining(Streams.java:228)
	at java.base/java.util.Spliterator$OfLong.forEachRemaining(Spliterator.java:763)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:550)
	at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
	at java.base/java.util.stream.LongPipeline.toArray(LongPipeline.java:521)
	at org.hyperledger.besu.ethereum.api.query.BlockchainQueries.gasPrice(BlockchainQueries.java:823)
	at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.EthGasPrice.response(EthGasPrice.java:54)
	at org.hyperledger.besu.ethereum.api.jsonrpc.JsonRpcHttpService.process(JsonRpcHttpService.java:725)
[...]

PR #2396, opened 2021-06-09
Transaction.gasPrice Optional

garyschulte:

## PR description

Systemic treatment of gasPrice as Optional has revealed several areas where we are still assuming gasPrice is not null and where we do not have specific treatment of 1559 fee fields

## Fixed Issue(s)

#2389 
## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetMinerDataByBlockHash.java
```diff
@@ -92,6 +92,7 @@ public static MinerDataResult createMinerDataResult(
                           t.getTransaction().getHash());
                   return t.getTransaction()
                       .getGasPrice()
+                      .get()
```
garyschulte, line 95: mark with TODO for 1559 txn's with no gasPrice

garyschulte: changed to effectiveGas, no TODO

jflo (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionEncoder.java
```diff
@@ -92,7 +91,7 @@ static void encodeAccessList(final Transaction transaction, final RLPOutput rlpO
     encodeAccessListInner(
         transaction.getChainId(),
         transaction.getNonce(),
-        transaction.getGasPrice(),
+        transaction.getGasPrice().get(),
```
matkt, line 94: Not important but why not use orElseThrow as for maxFeePerGas ?

matkt (approved):
LGTM

matkt:
@garyschulte maybe you have to ask those who work on the privacy features to be sure that there is no impact

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionEncoder.java
garyschulte: Yup, will change to orElseThrow() on next merge of master

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -820,7 +820,10 @@ private TransactionWithMetadata transactionByHeaderAndIndex(
                         .orElseThrow(
                             () -> new IllegalStateException("Could not retrieve block #" + l)))
             .flatMap(Collection::stream)
-            .mapToLong(t -> t.getGasPrice().toLong())
+            // TODO: currently we omit 1559 transactions from this calculation, what should
+            //       gasPrice() return for this endpoint post-London? (substitute effective gas?)
```
lucassaldanha, line 824: Just a sanity check. Was this TODO supposed to be here? I mean, did we forget to do something or is this a reminder for a future change?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -112,7 +112,8 @@ public MainnetTransactionValidator(
       return signatureResult;
     }
 
-    if (goQuorumCompatibilityMode && !transaction.getGasPrice().isZero()) {
+    if (goQuorumCompatibilityMode
+        && !transaction.getGasPrice().map(Wei::isZero).orElse(Boolean.TRUE)) {
```
lucassaldanha, line 116: This clause is getting harder and harder to understand. Maybe we could encapsulate the "tx cost zero" logic into a method in the Transaction class?

Something like: `transaction.hasGasPrice()` or `transaction.isGasPriceZero()` (I'm not great with names, feel free to come up with a better name for this method).

I recon this would make this if statement easier to read:

```
if (goQuorumCompatibilityMode && transaction.hasGasPrice()) {
    ....
}
```

ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/TransactionSmartContractPermissioningController.java
```diff
@@ -194,7 +196,8 @@ private static Bytes encodeTransaction(final Transaction transaction) {
         encodeAddress(transaction.getSender()),
         encodeAddress(transaction.getTo()),
         transaction.getValue().toBytes(),
-        transaction.getGasPrice().toBytes(),
+        // TODO: no treatment of 1559 fields here, is Bytes32.ZERO an appropriate fallback?
```
lucassaldanha, line 199: Same question as to the other TODO. Did we forget this here or will we deal with this later?

lucassaldanha (approved):
Overall it is looking good and I don't think it will cause any side effects on private txs.

I've left a few comments/suggestions. Nothing major.

lucassaldanha:
One thing that I'm thinking that we might want is to simplify the access to the tx gas price when we don't want to handle the optional value (e.g. when optional.empty means zero and not-empty means a Wei value).

e.g.
`return transaction.getGasPrice().orElse(Wei.ZERO)` and `transaction.getGasPrice().get().getValue().longValue()`
could become:
`return transaction.getGasPriceValue()` and `transaction.getGasPriceValue().longValue()`

What I'm thinking is that having a method like `transaction.getGasPriceValue()` or something, that "hides" the optional value in favour of a Wei or numeric might simplify the access to this property in some places.

ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/TransactionSmartContractPermissioningController.java
garyschulte: added to #2441, thanks again for catching

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
garyschulte: this is a reminder I forgot to remove - it is added to #2441.  Thanks for catching

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
garyschulte: encapsulated in core Transaction.  I went ahead and added 1559 fields just in case.

PR #2399, opened 2021-06-09
update to latest reference tests develop branch

garyschulte:
## PR description
* update ref tests to latest develop
* add London to the list of forks tested
* handle missing baseFee for legacy tests
* ignore `transactionSequence` for blockchain test cases (per-transaction block inclusion/exclusion reason)
* remove aleut spec

## Changelog

### DIFF
--- a/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/GeneralStateReferenceTestTools.java
+++ b/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/GeneralStateReferenceTestTools.java
@@ -60,7 +60,8 @@ public class GeneralStateReferenceTestTools {
     final String eips =
         System.getProperty(
             "test.ethereum.state.eips",
-            "Frontier,Homestead,EIP150,EIP158,Byzantium,Constantinople,ConstantinopleFix,Istanbul,Berlin,London");
+            "Frontier,Homestead,EIP150,EIP158,Byzantium,Constantinople,ConstantinopleFix,Istanbul,Berlin," +
+                "London,Shanghai,Cancun");
     EIPS_TO_RUN = Arrays.asList(eips.split(","));
   }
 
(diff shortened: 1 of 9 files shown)


ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/GeneralStateReferenceTestTools.java
```diff
@@ -60,7 +60,7 @@ private static MainnetTransactionProcessor transactionProcessor(final String nam
     final String eips =
         System.getProperty(
             "test.ethereum.state.eips",
-            "Frontier,Homestead,EIP150,EIP158,Byzantium,Constantinople,ConstantinopleFix,Istanbul,Berlin");
+            "Frontier,Homestead,EIP150,EIP158,Byzantium,Constantinople,ConstantinopleFix,Istanbul,Berlin,London");
```
RatanRSur, line 63: Is there something we can do to make this automatic? At the very least, can we add Shanghai, Cancun, etc. here now so that we don't forget them?

garyschulte: will pre-add them.  trying to introspect into the test jsons to get the full set of forks is probably just too messy.

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/GeneralStateTestCaseSpec.java
```diff
@@ -120,6 +136,7 @@ public int hashCode() {
   }
 
   /** Represents the "post" part of a general state test json _for a specific hard-fork_. */
+  @JsonIgnoreProperties(ignoreUnknown = true)
```
RatanRSur, line 139: Why is this needed?

RatanRSur (approved):
Just 1 more question, probably not a blocker.

RatanRSur:
Oh, and can we also remove baikal from the list of state test forks?

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/GeneralStateTestCaseSpec.java
garyschulte: The latest test artifacts have transaction exception / failure expectations that we are not currently using.

garyschulte:
> Oh, and can we also remove baikal from the list of state test forks?

The only reference I can find to baikal is in protocol schedules, and Karim has repurposed it for Calaveras.

PR #2406, opened 2021-06-10
2004 reduce logging bigints

### LINKED ISSUE #2004: Reduce logging of big ints
Because of this we're dumping big logs:

Ethereum/hive#458

jflo:
## PR description
Avoids logging (potentially large) corrupt HELLO messages when connecting with peers.

## Fixed Issue(s)

fixes #2004 
## Changelog
Failures to parse HELLO now include byte length differences on nodeId in the exception message, as well as guidance to adjust the debug level to trace if user is really interested in seeing the entire peerInfo.


jflo:
I'm considering updating the DeFramer unit test to mock out the logger and check that it is called as expected in cases of valid and invalid HELLOs

garyschulte (approved):
🚢

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/rlpx/connections/netty/DeFramer.java
```diff
@@ -109,12 +109,12 @@ protected void decode(final ChannelHandlerContext ctx, final ByteBuf in, final L
         try {
           peerInfo = HelloMessage.readFrom(message).getPeerInfo();
         } catch (final RLPException e) {
-          LOG.debug("Received invalid HELLO message", e);
+          LOG.warn("Received invalid HELLO message, set log level to TRACE for message body", e);
```
matkt, line 112: Are we sure that passing this log in warning does not risk spamming the logs of a node running on the mainnet or on ropsten etc ? Do we often receive invalid messages of this type ?

jflo: Great point. I looked at log aggregation for all of our canaries and only see 1 occurrence of this exception. This should not result in logspam at warn instead of debug.

matkt: ok 👍

PR #2424, opened 2021-06-11
update eth_sendRawTransaction JSON RPC method

matkt:

## PR description
We are not returning an error message when we are calling sendRawTransaction with “max priority fee per gas greater than max fee per gas” and we are adding the transaction to the pool. This transaction must throw an error. This PR update the JSON RPC API to fix that 

## Changelog


abdelhamidbakhta (approved):
Looks good to me.

PR #2339, opened 2021-05-29
use baseFee instead of baseFeePerGas for retesteth genesis config

garyschulte:
Signed-off-by: garyschulte <garyschulte@gmail.com>

## PR description

minor tweak to the baseFee config parameter for retesteth command since  that rpc command spec has not (yet?) changed to baseFeePerGas

## Changelog

- [ ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #2427, opened 2021-06-11
change test_setChainParams to baseFeePerGas 

garyschulte:

## PR description
change test_setChainParams baseFee back to baseFeePerGas now that retesteth is using that param name

## Changelog


matkt:
I'm not sure but If it is a rollback. Why we do not find the same modification present in this PR ? https://github.com/hyperledger/besu/pull/2339/files

garyschulte:
> I'm not sure but If it is a rollback. Why we do not find the same modification present in this PR ? https://github.com/hyperledger/besu/pull/2339/files

The `test_` namespace spec is owned by Ori and Dimitry, and until recently (Friday) the genesis config was using `baseFee` instead of baseFeePerGas.

matkt (approved):
LGTM

PR #2432, opened 2021-06-14
eth_feeHistory

### LINKED ISSUE #2430: Add feeHistory JSON RPC API
Description : https://github.com/ethereum/eth1.0-specs/pull/212/files

RatanRSur:
The spec for this implementation is listed in the issue below.

## Fixed Issue(s)

fixes #2430 

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -0,0 +1,242 @@
...
+            .map(blockHeader -> blockHeader.getBaseFee().orElse(0L))
+            .orElseGet(
+                () ->
+                    protocolSchedule
+                        .getByBlockNumber(nextBlockNumber)
+                        .getEip1559()
+                        .map(
+                            eip1559 -> {
+                              final BlockHeader lastBlockHeader =
+                                  blockHeaders.get(blockHeaders.size() - 1);
```
matkt, line 111: I think we can have a out of bound exception if blockCount==0

Maybe we need to add a check to avoid blockcount<1

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -0,0 +1,242 @@
...
+    final List<Double> gasUsedRatios =
+        blockHeaders.stream()
+            .map(blockHeader -> blockHeader.getGasUsed() / (double) blockHeader.getGasLimit())
+            .collect(toUnmodifiableList());
+
+    final Optional<List<List<Long>>> maybeRewards =
+        maybeRewardPercentiles.map(
+            rewardPercentiles ->
+                LongStream.range(oldestBlock, oldestBlock + blockCount)
+                    .mapToObj(blockNumber -> blockchain.getBlockByNumber(blockNumber).get())
```
matkt, line 130: we can have an issue if blockCount > head block number

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -0,0 +1,242 @@
...
+              >= rewardPercentiles.get(rewardIndex)) {
+        rewards.add(transaction.getEffectivePriorityFeePerGas(baseFee));
+        rewardIndex++;
+      }
+    }
+    return rewards;
+  }
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  public static class FeeHistory {
```
matkt, line 185: NIT: For consistency with the current code and specification, I would say that it's better to rename the class to FeeHistoryResult and to use the Jackson and Immutable annotations

Spec : https://github.com/ethereum/eth1.0-specs/pull/212/files#diff-c56024dc5c5982b3df8a0e683e2e340ff092b384030ce455e4a4b45239a43b81R235
Location in the code : https://github.com/hyperledger/besu/tree/e230a445bb457bbac27bf4c384da449ccbd3f043/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results
Sample of result class : https://github.com/hyperledger/besu/tree/e230a445bb457bbac27bf4c384da449ccbd3f043/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results

matkt (changes requested):

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -263,15 +263,13 @@ private ProcessableBlockHeader createPendingBlockHeader(final long timestamp) {
...
-            eip1559.computeBaseFee(
-                newBlockNumber,
-                parentHeader.getBaseFee().orElseThrow(),
-                parentHeader.getGasUsed(),
-                eip1559.targetGasUsed(parentHeader));
       }
+      baseFee =
+          eip1559.computeBaseFee(
+              newBlockNumber,
+              parentHeader.getBaseFee().orElseThrow(),
```
matkt, line 270: Maybe I'm missing something but if we're in the fork block. The parent has no baseFee and this will throw an exception ? 🤔

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
```diff
@@ -0,0 +1,138 @@
...
+                .getResult())
+        .isEqualTo(
+            new EthFeeHistory.FeeHistory(
+                10,
+                List.of(47177L, 53074L),
+                List.of(0.9999999992132459),
+                Optional.of(List.of(List.of(1524742083L)))));
+  }
+
+  @Test
```
garyschulte, line 107: suggest adding negative test cases for troublesome blockCounts like 0, -1 , etc

garyschulte (changes requested):
suggestions re: arithmetic and input validation

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -0,0 +1,192 @@
...
+        blockHeaders.stream()
+            .map(blockHeader -> blockHeader.getGasUsed() / (double) blockHeader.getGasLimit())
+            .collect(toUnmodifiableList());
+
+    final Optional<List<List<Long>>> maybeRewards =
+        maybeRewardPercentiles.map(
+            rewardPercentiles ->
+                LongStream.range(oldestBlock, oldestBlock + blockCount)
+                    .mapToObj(blockNumber -> blockchain.getBlockByNumber(blockNumber))
+                    .flatMap(Optional::stream)
```
garyschulte, line 131: 👍

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -0,0 +1,190 @@
...
+    final List<Transaction> transactionsAscendingEffectiveGasFee =
+        transactions.stream()
+            .sorted(
+                Comparator.comparing(
+                    transaction -> transaction.getEffectivePriorityFeePerGas(baseFee)))
+            .collect(toUnmodifiableList());
+
+    // We need to weight the percentile of rewards by the gas used in the transaction.
+    // That's why we're keeping track of the cumulative gas used and checking to see which
+    // percentile markers we've passed
```
garyschulte, line 169: 🙏

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -327,6 +327,18 @@ public Wei getGasPrice() {
...
+  public long getEffectivePriorityFeePerGas(final Optional<Long> maybeBaseFee) {
+    return maybeBaseFee
+        .filter(__ -> getType().supports1559FeeMarket())
+        .map(
+            baseFee ->
+                Math.max(
+                    0,
+                    Math.min(
+                        getMaxPriorityFeePerGas().get().getAsBigInteger().longValue(),
+                        getMaxFeePerGas().get().getAsBigInteger().longValue() - baseFee)))
```
garyschulte, line 339: this is going to throw for frontier transactions post-london

garyschulte (changes requested):
LGTM, just need to address effectivePriorityFee for legacy transactions post-london

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
RatanRSur: Does the `filter` above not sift those out?

garyschulte: ah yes it does, quite compact 👍

garyschulte (approved):
🚢

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
garyschulte: Forgot about the fork block thing  ☝️

garyschulte (changes requested):

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceTest.java
```diff
@@ -1699,8 +1699,20 @@ private void assertTransactionResultMatchesTransaction(
...
+    Optional.ofNullable(result.getString("maxFeePerGas"))
+        .ifPresent(
+            maxPriorityFeePerGas ->
+                assertThat(Wei.fromHexString(maxPriorityFeePerGas))
+                    .isEqualTo(transaction.getMaxFeePerGas().get()));
+    Optional.ofNullable(result.getString("maxPriorityFeePerGas"))
+        .ifPresent(
+            gasPriceString ->
+                assertThat(Wei.fromHexString(gasPriceString))
+                    .isEqualTo(transaction.getMaxPriorityFeePerGas().get()));
```
garyschulte, line 1715: these lambda parameter names are a bit jumbled.  Test should pass, but gasPrice -> maxFeePerGas, maxFeePerGas -> maxPriorityFeePerGas, maxPriorityFeePerGas -> gasPriceString

🤯

garyschulte (changes requested):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -0,0 +1,190 @@
...
+    final BlockParameter highestBlock = request.getRequiredParameter(1, BlockParameter.class);
+    final Optional<List<Double>> maybeRewardPercentiles =
+        request.getOptionalParameter(2, Double[].class).map(Arrays::asList);
+
+    final long chainHeadBlockNumber = blockchain.getChainHeadBlockNumber();
+    final long resolvedHighestBlockNumber =
+        highestBlock
+            .getNumber()
+            .orElse(
+                highestBlock.isEarliest()
```
matkt, line 75: why this check ? if it's earlist the highestBlock.getNumber() will not return automaticaly the genesis ? https://github.com/hyperledger/besu/blob/e230a445bb457bbac27bf4c384da449ccbd3f043/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/BlockParameter.java#L42

matkt:
I'm just concerned about the performance of this API when blockcount == 500 on calaveras. I wanted to try and after 2 minutes and several thread blocked we have the result. I also  wanted to test with a blockcount == 10 starting from the block 6400 and I had 23 seconds. These blocks are filled by transactions 

What is the blockcount requested in the common case ?

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
```diff
@@ -619,7 +620,7 @@ public KeyPair generateKeyPair() {
     private boolean hasTransactions = true;
     private TransactionType[] transactionTypes = TransactionType.values();
     private Optional<Address> coinbase = Optional.empty();
-    private Optional<Long> baseFee = Optional.empty();
+    private Optional<Optional<Long>> maybeBaseFee = Optional.empty();
```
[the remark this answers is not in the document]
RatanRSur: Yeah because we need a way to encode if the basefee is unspecified. This is important for knowing whether to use a default base fee or the optional.empty the caller gives explicitly to the builder.

RatanRSur:
I think block counts of 500 would be very unusual but 10 sounds reasonable.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceTest.java
RatanRSur: wow, how did I manage to do that haha

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
RatanRSur: Ah yes, so I just need to check for the latest and pending cases.

garyschulte:
> I think block counts of 500 would be very unusual but 10 sounds reasonable.

How about we silently map anything >10 to 10.  We might fail some test cases in hive, but this seems like an acceptable tradeoff for now.

garyschulte (approved):
🚢

Issue #2409, opened 2021-06-10
Blockchain Tests


sajz:
Executes multiple blocks with different transactions

[BlockchainTests](https://github.com/ethereum/tests/tree/develop/BlockchainTests)

Success Criteria:
* passing all blockchain tests with besu reference test pipeline
* passing all blockchain tests with retesteth test runner

garyschulte:
all blockchain tests pass besu reference test pipeline
with #2434, all blockchain tests are passing again for retesteth engine

PR #2434, opened 2021-06-15
ensure test_importRawBlock appends to chain head

garyschulte:

## PR description
Ensures `test_importRawBlock` appends the block to the chain even if the block already exists in storage.  This behavior is limited to the test endpoint, rather than pushing it down into the block importer.

## Fixed Issue(s)

#2409 

## Changelog


abdelhamidbakhta (approved):
Looks good to me.

PR #2445, opened 2021-06-16
Update eth call post 1559

matkt:
## PR description

Update eth_call to work correctly after the london fork 

Update integration tests to be able to test JSON RPC methods after the london fork

## Changelog


garyschulte:
The eth_call changes LGTM 👍 , the diff includes some tracing changes and artifacts though - is that expected?  Is this PR based on a branch that hasn't merged yet?

matkt:
> The eth_call changes LGTM 👍 , the diff includes some tracing changes and artifacts though - is that expected? Is this PR based on a branch that hasn't merged yet?

I updated the code to fix the pipeline and I merged the master. The tracing is no longer present in the PR

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -151,52 +152,78 @@ public TransactionSimulator(
...
+
+    if (header.getBaseFee().isEmpty()) {
+      transactionBuilder.gasPrice(gasPrice);
+    } else if (protocolSchedule.getChainId().isPresent()) {
+      transactionBuilder
+          .maxFeePerGas(callParams.getMaxFeePerGas().orElse(gasPrice))
+          .maxPriorityFeePerGas(callParams.getMaxPriorityFeePerGas().orElse(gasPrice));
+    } else {
+      return Optional.empty();
+    }
```
garyschulte, line 206: is there a protocol schedule where we would have a base fee but no chain id?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -400,15 +400,16 @@ public void shouldReturnFailureResultWhenProcessingFailsByHash() {
...
     mockWorldStateForAccount(Hash.ZERO, callParameter.getFrom(), 1L);
 
     final Transaction expectedTransaction =
         Transaction.builder()
             .type(TransactionType.EIP1559)
             .chainId(BigInteger.ONE)
             .nonce(1L)
-            .gasPrice(callParameter.getGasPrice())
+            .chainId(BigInteger.ONE)
+            .type(TransactionType.EIP1559)
```
garyschulte, line 412: duplicates of 408-409

garyschulte (changes requested):
just a little copypasta on one test, otherwise lgtm

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
matkt: I removed this part because I found that the chainId is already set here 
https://github.com/hyperledger/besu/pull/2445/files#diff-49639cb13fe67367ebb91b4b732316588dd971719829ba51071141026b7888a9L188

It was useless to do that on this part

To answer the question, no it is not possible regarding the spec

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
matkt: ah 🤦  thanks I just removed that

garyschulte (approved):
🚢

PR #2449, opened 2021-06-18
Bft messages have duplicate prepares in roundChange message

jframe:
## PR description
In IBFT2 and QBFT the round change message could be created with duplicate prepares being included in the round change message. These messages are rejected as being invalid by Besu.

This fixes the equality and hashCode on the message so that we don't add duplicate messages.

## Changelog


usmansaleem (approved):
LGTM

usmansaleem (approved):
LGTM

PR #2481, opened 2021-06-25
Implement eth_getBlockByNumber while downloading state.

### LINKED ISSUE #2209: Determine correct behavior of eth_getBlockByNumber, latest
Currently, we return the latest chain head even if we don't have state for it. For example, in fast sync. At least one other client seems to return the latest head that we have the state for, which means while fast syncing they return 0. We should work out what the behavior should be.

This has implications for how hive testing is currently done.

jflo:
## PR description
Introduces a dependency from eth_getBlockByNumber onto the Synchronizer, in order to determine download state, and current most recent block that we have state for. JsonRPC test classes also were refactored to provide a Synchronizer.

## Fixed Issue(s)

fixes #2209 

## Changelog
When asking for the latest block while downloading state, will return most recent block we have state for. While still downloading block headers, will return genesis block.

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetBlockByNumber.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetBlockByNumber.java
@@ -23,6 +23,7 @@ import org.hyperledger.besu.ethereum.api.query.BlockchainQueries;
 import org.hyperledger.besu.ethereum.chain.Blockchain;
 import org.hyperledger.besu.ethereum.core.BlockHeader;
 import org.hyperledger.besu.ethereum.core.Hash;
+import org.hyperledger.besu.ethereum.core.Synchronizer;
 
 import java.util.function.Supplier;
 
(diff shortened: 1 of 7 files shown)


vmichalik:
Hey @jflo, it would be useful for posterity to have a descriptive PR title. How about "Implement correct behavior of eth_getBlockByNumber"?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetBlockByNumber.java
```diff
@@ -20,27 +20,39 @@
...
 
 import com.google.common.base.Suppliers;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class EthGetBlockByNumber extends AbstractBlockParameterMethod {
 
   private final BlockResultFactory blockResult;
   private final boolean includeCoinbase;
+  private static final Logger log = LogManager.getLogger();
```
matkt, line 38: For more consistency with the other classes I would say that it is necessary to do like this
> private static final Logger LOGGER = LogManager.getLogger();

ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/DummySynchronizer.java
```diff
@@ -0,0 +1,63 @@
...
+
+package org.hyperledger.besu.ethereum.retesteth;
+
+import org.hyperledger.besu.ethereum.core.Synchronizer;
+import org.hyperledger.besu.plugin.data.SyncStatus;
+import org.hyperledger.besu.plugin.services.BesuEvents;
+
+import java.util.Optional;
+
+public class DummySynchronizer implements Synchronizer {
```
matkt, line 24: NIT : I would have added a description to the class to understand the goal of this class faster

matkt (approved):
LGTM

PR #2490, opened 2021-06-30
GraphQL schema + adapter updates to handle GoQuorum transaction fields

mark-terry:
## PR description
Updates the GraphQL schema + adapter to handle GoQuorum transaction fields

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLDataFetchers.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLDataFetchers.java
@@ -289,32 +289,32 @@ public class GraphQLDataFetchers {
 
   private TransactionAdapter updatePrivatePayload(final Transaction transaction) {
     final GoQuorumEnclave enclave = goQuorumPrivacyParameters.get().enclave();
-    final Bytes enclavePayload;
+    Bytes enclavePayload;
 
     try {
       // Retrieve the payload from the enclave
       enclavePayload =
           Bytes.wrap(enclave.receive(transaction.getPayload().toBase64String()).getPayload());
-
-      // Return a new transaction containing the retrieved payload
-      return new TransactionAdapter(
-          new TransactionWithMetadata(
-              new Transaction(
-                  transaction.getNonce(),
-                  transaction.getGasPrice(),
-                  transaction.getMaxPriorityFeePerGas(),
-                  transaction.getMaxFeePerGas(),
-                  transaction.getGasLimit(),
-                  transaction.getTo(),
-                  transaction.getValue(),
-                  transaction.getSignature(),
-                  enclavePayload,
-                  transaction.getSender(),
-                  transaction.getChainId(),
-                  Optional.ofNullable(transaction.getV()))));
     } catch (final Exception ex) {
       LOG.debug("An error occurred while retrieving the GoQuorum transaction payload: ", ex);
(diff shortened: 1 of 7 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLDataFetchers.java
```diff
@@ -259,7 +275,46 @@ public GraphQLDataFetchers(final Set<Capability> supportedCapabilities) {
...
+    final GoQuorumEnclave enclave = goQuorumPrivacyParameters.get().enclave();
+    final Bytes enclavePayload;
+
+    try {
+      // Retrieve the payload from the enclave
+      enclavePayload =
+          Bytes.wrap(enclave.receive(currTransaction.getPayload().toBase64String()).getPayload());
+
+      // Create a new transaction containing the retrieved payload
+      final Transaction newTransaction =
```
macfarla, line 300: I don't like the names currTransaction and newTransaction. Can we make currTransaction -> transaction and make the new Transaction call inline so it doesn't need a name

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLDataFetchers.java
```diff
@@ -259,7 +275,46 @@ public GraphQLDataFetchers(final Set<Capability> supportedCapabilities) {
...
+              currTransaction.getGasLimit(),
+              currTransaction.getTo(),
+              currTransaction.getValue(),
+              currTransaction.getSignature(),
+              enclavePayload,
+              currTransaction.getSender(),
+              currTransaction.getChainId(),
+              Optional.ofNullable(currTransaction.getV()));
+      return new TransactionAdapter(new TransactionWithMetadata(newTransaction));
+    } catch (final Exception ex) {
```
macfarla, line 315: can we be more specific about what this exception is?

macfarla (commented):
Does this make the GQ AT pass?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLDataFetchers.java
mark-terry: Happy to do that if you find it clearer.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLDataFetchers.java
mark-terry: The most common exception I'd expect here is an EnclaveClientException, but I'm not sure it's the only one.

macfarla (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/internal/pojoadapter/TransactionAdapter.java
```diff
@@ -185,4 +185,16 @@ public TransactionAdapter(final TransactionWithMetadata transactionWithMetadata)
     }
     return results;
   }
+
+  public Optional<Boolean> getIsPrivate() {
```
pinges, line 189: does that really have to be an Optional? I think it should just return the boolean ...

PR #2493, opened 2021-07-01
refcator response constructor

RatanRSur:
This is a non-functional change that sets me up to better support eth66.


matkt (approved):
LGTM

PR #2495, opened 2021-07-02
test simple poc blacklist peers

matkt:

## Changelog


vmichalik:
Hey @matkt just as a point of order we moved on from whitelist/blacklist to allowlist/denylist as conventions

PR #2501, opened 2021-07-06
SECP256R1 BC now using non-deterministic variant

### LINKED ISSUE #2500: Change SECP256R1 K calculation to non-deterministic variant
### Description

ECDSA exists in two variants:
1) The original proposed standard, which uses a random element in its calculation. This means that signing the same data with the same private key several times, creates different results each time.
2) [RFC 6979](https://datatracker.ietf.org/doc/html/rfc6979#section-3.2) proposes a deterministic variant, which uses HMAC to replace the random element.

The SECP256K1 implementation in Besu uses the deterministic variant. When SECP256R1 was implemented, the same variant was used. However, the deterministic variant is not NIST-approved. Besu should use the non-deterministic variant. This matches the native implementation (OpenSSL) that we are using as well.

### Acceptance Criteria
* When using SECP256K1, the deterministic K calculation should be used
* When using SECP256R1, the non-deterministic K calculation should be used

lucassaldanha:
## PR description

- Updated our `SignatureAlgorithm` interface with a method to return the K calculator implementation (BC)
- Each concrete implementation returns the expected calculator:
  - SECP256K1: deterministic calculator
  - SECP256R1: non-deterministic calculator

## Fixed Issue(s)
Fixes #2500

## Changelog
N/A

### DIFF
--- a/crypto/src/main/java/org/hyperledger/besu/crypto/SignatureAlgorithm.java
+++ b/crypto/src/main/java/org/hyperledger/besu/crypto/SignatureAlgorithm.java
@@ -20,6 +20,7 @@ import java.util.function.UnaryOperator;
 
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
+import org.bouncycastle.crypto.signers.DSAKCalculator;
 import org.bouncycastle.math.ec.ECPoint;
 
 public interface SignatureAlgorithm {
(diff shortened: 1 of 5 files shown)


crypto/src/main/java/org/hyperledger/besu/crypto/SignatureAlgorithm.java
```diff
@@ -54,6 +55,8 @@ SECPSignature normaliseSignature(
 
   String getCurveName();
 
+  DSAKCalculator getKCalculator();
```
daniel-iobuilders, line 58: I think this function does not have to be part of the interface, as it only used internally.

lucassaldanha: Good point. I've removed it from the interface! 👍

pinges (approved):
LGTM

PR #2502, opened 2021-07-06
Fix cleared storage issue

matkt:

## PR description

A consensus issue was detected on the mainnet with bonsai. This came from a call to selfdestruct on a specific account. 

block : https://etherscan.io/txsInternal?block=11114732

account : 0x0000000000007f150bd6f54c40a34d7c3d5e9f56

The slot had been cleared but did not have the flag isCleared to true in this specific case

## Changelog


garyschulte (approved):
seems like a good integration test case!

PR #2505, opened 2021-07-06
Prioritize with nonce distance

jflo:
## PR description

Prevent multiple transactions from the same address from evicting others which are waiting in a full transaction pool. When we have to evict a transaction from a full pool, we now also consider how far it is from the lowest nonce from the same address already in the pool. We are also increasing the size of the pending transaction pool, so it should support the new 10MB maximum message size previously introduced by pr [#2485](https://github.com/hyperledger/besu/pull/2485) .

## Changelog
Default value for `--tx-pool-max-size`  is now 32K.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransactions.java
```diff
@@ -105,8 +106,22 @@
                           .getMaxFeePerGas()
                           .map(maxFeePerGas -> maxFeePerGas.getValue().longValue())
                           .orElse(transactionInfo.getGasPrice().toLong()))
+              .thenComparing(transactionInfo -> distanceFromNextNonce(transactionInfo))
               .thenComparing(TransactionInfo::getSequence)
               .reversed());
+
+  private Long distanceFromNextNonce(final TransactionInfo incomingTx) {
+    TransactionsForSenderInfo inPool = transactionsBySender.get(incomingTx.getSender());
```
matkt, line 114: I think we have final here

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransactions.java
```diff
@@ -105,8 +106,22 @@
...
+    if (inPool == null) { // nothing in pool, you're next
```
matkt, line 115: are we sure we can have only a null but not empty inPool ?

matkt (changes requested):
Some comments

matkt (approved):
LGTM

besu/src/test/java/org/hyperledger/besu/services/BesuEventsImplTest.java
```diff
@@ -436,7 +432,7 @@ public void transactionDroppedEventDoesNotFireAfterUnsubscribe() {
     transactionPool.addLocalTransaction(TX2);
 
     assertThat(result.get()).isNotNull();
-    serviceImpl.removeTransactionAddedListener(id);
+    serviceImpl.removeTransactionDroppedListener(id);
```
garyschulte, line 435: nice catch.

garyschulte (approved):
🚢

PR #2517, opened 2021-07-08
switch to hex integers for eth_feeHistory

RatanRSur:
https://github.com/ethereum/eth1.0-specs/blob/master/json-rpc/spec.json#L200


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/FeeHistory.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Nullable
+    @JsonProperty("reward")
+    List<List<String>> getReward();
+
+    static FeeHistoryResult from(final FeeHistory feeHistory) {
+      return ImmutableFeeHistoryResult.of(
+          "0x" + Long.toHexString(feeHistory.getOldestBlock()),
+          feeHistory.getBaseFeePerGas().stream()
+              .map(baseFeePerGas -> "0x" + Long.toHexString(baseFeePerGas))
```
matkt, line 59: Why not using Quantity.create to have the hex string value in this case ?

matkt (changes requested):
Small comment

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/FeeHistory.java
RatanRSur: Oh right! Forgot that existed. Will do.

matkt (approved):
LGTM

PR #2536, opened 2021-07-13
p2p tls support (command line options and acceptance tests)

perusworld:
This is the final part of the changes to support p2p over tls. this has command line options to support p2p over tls enablement and acceptance tests to cover them as well.
The keystores are generated using scripts from https://github.com/perusworld/besu-isolated-networks/blob/p2p-over-ssl/scripts/gen-at-keys.sh


acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/bft/BftMiningAcceptanceTest.java
```diff
@@ -123,4 +123,52 @@ public void shouldStillMineWhenANonProposerNodeFailsAndHasSufficientValidators()
...
+      minerNode2.execute(accountTransactions.createIncrementalTransfers(sender, receiver, 1));
+      cluster.verify(receiver.balanceEquals(1));
+
+      minerNode3.execute(accountTransactions.createIncrementalTransfers(sender, receiver, 2));
+      cluster.verify(receiver.balanceEquals(3));
+
+      minerNode4.execute(accountTransactions.createIncrementalTransfers(sender, receiver, 3));
+      cluster.verify(receiver.balanceEquals(6));
+
+      if (minerNode1.getTLSConfiguration().isEmpty()) {
```
lucassaldanha, line 158: I'm a little bit confused by this if statement here. What is the reason for this?

besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -190,6 +192,18 @@ public RunnerBuilder p2pEnabled(final boolean p2pEnabled) {
...
+    return this;
+  }
+
+  public RunnerBuilder p2pTLSConfiguration(final Optional<TLSConfiguration> p2pTLSConfiguration) {
+    if (null != p2pTLSConfiguration) {
+      this.p2pTLSConfiguration = p2pTLSConfiguration;
+    }
+    return this;
+  }
+
```
lucassaldanha, line 206: Are we using both these methods?

besu/src/main/java/org/hyperledger/besu/cli/options/stable/P2PTLSConfigOptions.java
```diff
@@ -0,0 +1,131 @@
...
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+import org.apache.logging.log4j.Logger;
+import picocli.CommandLine;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.ParameterException;
+
+public class P2PTLSConfigOptions {
```
lucassaldanha, line 33: We probably want to add the -X to the properties as an experimental property for this first release. That means they might change between minor versions (helpful if we decide to add/remove/rename properties later on).

perusworld: done

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/bft/BftMiningAcceptanceTest.java
perusworld: if you are referring to the if statement, that is needed as we are using the parameterized tests with multiple config settings the CRL config is present only for the TLS related parameterized configs and hence those nodes are rejected during connection attempts.

besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
perusworld: yeah had to add them here to avoid conditional code from the callers :-)

lucassaldanha (approved):

PR #2545, opened 2021-07-15
Tx replacement rules

### LINKED ISSUE #2529: Change behavior of tx replacement price bump requirements
Right now we require a percentage increase on the effective priority fee per gas. This causes undesirable properties because most txs stuck in the pool and eligible for replacement will have an effective priority fee per gas of 0. That means that the price bump will be 0. We should require a bump of both the feeCap and maxPriorityFeePerGas.

garyschulte:
## PR description
refactor of transaction replacement rules:
* split into gasPrice and 1559FeeMarket rules
* consider maxFeePerGas as well as maxPriorityFeePerGas in the transaction bump logic for 1559 fee market

1559 fee market logic is effectively:
* replace when BOTH:
  * new effective price  > prior effective price plus percent bump 
  * new effective priority fee is >= prior effective priority fee
* OR, replace when BOTH:
  * new effective price == prior effective price 
  * new effective priority fee > prior effective priority fee plus percent bump
* ELSE, don't replace

## Fixed Issue(s)

fixes #2529

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolReplacementHandler.java
```diff
@@ -28,7 +28,10 @@
   private final List<TransactionPoolReplacementRule> rules;
 
   public TransactionPoolReplacementHandler(final Percentage priceBump) {
-    this(asList(new TransactionReplacementByPriceRule(priceBump)));
+    this(
+        asList(
```
matkt, line 32: I wonder if the rules list can be defined based on the fork as we do for the rest here
https://github.com/hyperledger/besu/blob/1772f8b1f89686e674e3ccaf333ae7f6355d1d36/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java#L458

matkt (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolReplacementHandler.java
garyschulte: I pushed a commit that plumbs protocolSchedule down to TransactionPoolReplacementHandler.  LMK what you think, I am on the fence about it.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolReplacementHandler.java
```diff
@@ -14,35 +14,50 @@
...
-  private final List<TransactionPoolReplacementRule> rules;
+  private final Map<TransactionPoolReplacementRule, Function<BlockHeader, Boolean>> rules;
+
+  public TransactionPoolReplacementHandler(
+      final ProtocolSchedule protocolSchedule, final Percentage priceBump) {
 
-  public TransactionPoolReplacementHandler(final Percentage priceBump) {
-    this(asList(new TransactionReplacementByPriceRule(priceBump)));
+    this.rules =
+        ImmutableMap.of(
```
matkt, line 37: I was thinking of another way (https://github.com/hyperledger/besu/blob/1772f8b1f89686e674e3ccaf333ae7f6355d1d36/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java#L111) to do it because like that we have behavior logic based on the milestone in two different places in the code

I think we can create a transactionPoolRplacementHandler based on the milestone : 

```java
static ProtocolSpecBuilder londonDefinition(
      final Optional<BigInteger> chainId,
      final OptionalInt configContractSizeLimit,
      final OptionalInt configStackSizeLimit,
      final boolean enableRevertReason,
      final GenesisConfigOptions genesisConfigOptions,
      final boolean quorumCompatibilityMode) {
    ...
    return berlinDefinition(
            chainId,
            configContractSizeLimit,
            configStackSizeLimit,
            enableRevertReason,
            quorumCompatibilityMode)
        .gasCalculator(LondonGasCalculator::new)
        ....
        .transactionPoolReplacementHandler(asList(new TransactionReplacementByGasPriceRule(priceBump),  new TransactionReplacementByFeeMarketRule(priceBump)))
        ....      
        .name(LONDON_FORK_NAME);
  }
```


```java
static ProtocolSpecBuilder frontierDefinition
    ...
    return frontierDefinition(
        ....
        .transactionPoolReplacementHandler(asList(new TransactionReplacementByGasPriceRule(priceBump)))
        ....      
        .name(FRONTIER_FORK_NAME);
  }
```

matkt (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolReplacementHandler.java
garyschulte: I was reluctant to move the TransactionPoolReplacementRule interface into `core`, but creating a handler interface in there is more palatable.  :+1:

matkt (approved):
LGTM but we need another PR to clean some parts of the 1559 code

PR #2570, opened 2021-07-22
Bft validator provider

jframe:
## PR description
QBFT uses the same validator management logic that IBFT2 and IBFT use and allows for a lot of code reuse. This will make it possible for bft based consensus mechanism to use it's own specific mechanism to manage validators i.e. determining what the validators are and the mechanism to add and remove them.

This is needed so that QBFT can move to using transaction based validator management using smart contracts.

## Changelog


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/validator/blockbased/BlockValidatorProvider.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public Collection<Address> getValidatorsAfterBlock(final BlockHeader header) {
+    return voteTallyCache.getVoteTallyAfterBlock(header).getValidators();
+  }
+
+  @Override
+  public Optional<VoteProvider> getVoteProvider() {
+    return Optional.of(voteProvider);
```
usmansaleem, line 82: `Optional.ofNullable` ??

usmansaleem (approved):
LGTM - one comment around using Optional.ofNullable.

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/validator/blockbased/BlockValidatorProvider.java
jframe: VoteProvider is mandatory for BlockValidatorProvider and we create vote provider on line 67 so it will always exist

PR #2573, opened 2021-07-22
Added abstract BftBesuControllerBuilder

lucassaldanha:
## PR description

Introduced an abstract `BftBesuControllerBuilder` with supplier methods that enable concrete implementations to provide the desired concrete `ExtraDataCodec` and `BftBlockInterface`.

## Changelog


besu/src/main/java/org/hyperledger/besu/controller/QbftBesuControllerBuilder.java
```diff
@@ -76,19 +75,24 @@
...
   private static final Logger LOG = LogManager.getLogger();
   private BftEventQueue bftEventQueue;
   private BftConfigOptions bftConfig;
   private ValidatorPeers peers;
-  private final BftExtraDataCodec bftExtraDataCodec = new QbftExtraDataCodec();
-  private final BftBlockInterface blockInterface = new BftBlockInterface(bftExtraDataCodec);
+
+  @Override
+  protected Supplier<BftExtraDataCodec> bftExtraDataCodec() {
+    return Suppliers.memoize(QbftExtraDataCodec::new);
```
lucassaldanha, line 94: This is where next we will be returning the PKI-specific codec used only when running on "PKI mode".

jframe (approved):
LGTM

PR #2574, opened 2021-07-23
Qbft to use validator contract

jframe:
## PR description
Qbft to retrieve validators from the validator contract instead of from the block header.

* Adds a transaction validator provider to read the current validators from the contract
* Enables transaction based validators if validatorconractaddress has a value in the Qbft config

## Changelog


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ValidatorContractController.java
```diff
@@ -0,0 +1,92 @@
...
+    this.contractAddress = contractAddress;
+    this.transactionSimulator = transactionSimulator;
+
+    try {
+      this.getValidatorsFunction =
+          new Function(
+              GET_VALIDATORS,
+              List.of(),
+              List.of(new TypeReference<DynamicArray<org.web3j.abi.datatypes.Address>>() {}));
+    } catch (Exception e) {
```
usmansaleem, line 53: nit: final

usmansaleem (approved):
LGTM

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ValidatorContractController.java
jframe: done

usmansaleem (approved):
LGTM

PR #2575, opened 2021-07-23
2496 ping pong encoding error

### LINKED ISSUE #2496: Erroneous PING and PONG encoding of enr sequence number
The besu client does not properly encode the enr sequence number in PONG responses. 

See https://github.com/hyperledger/besu/blob/e230a445bb457bbac27bf4c384da449ccbd3f043/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PongPacketData.java#L72 

Essentially, the enr sequence is written as `writeBytes( seqAsAUint64.toBytes());`

Example Pong
```
&Pong{
            To:         Endpoint{net.ParseIP("127.0.0.1").To4(), 3322, 5544},
            ReplyTok:   common.HexToHash("0x0000000000000000000000000000000000000000000000000000000000001234").Bytes(),
            Expiration: 1136239445,
            ENRSeq:     0,
        },
```
( it's UDP 3322, TCP 5544)

Besu encodes the example above as `0xf83bcb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a355880000000000000000`. 
```
rlpdump -hex f83bcb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a355880000000000000000
[
  [
    7f000001,
    0cfa,
    15a8,
  ],
  0000000000000000000000000000000000000000000000000000000000001234,
  43b9a355,
  0000000000000000,
]
```

If the sequence number is `1`, it's encoded as `0xf83bcb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a355880000000000000001`

```
 rlpdump -hex f83bcb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a355880000000000000001
[
  [
    7f000001,
    0cfa,
    15a8,
  ],
  0000000000000000000000000000000000000000000000000000000000001234,
  43b9a355,
  0000000000000001,
]
```
So apparently the ENR sequence number is encoded as `8` bytes, not as an RLP integer. 

This was not an _obvious_ problem until geth implemented a change in how packets are decoded, in https://github.com/ethereum/go-ethereum/commit/3e6f46caec51d82aef363632517eb5842eef6db6 . Before this change: 

- Any 'junk' at the end would be ignored. Thus, the `0000000000000001` element was silently discarded, the enr sequence number was always `0`, when decoding a PONG from besu. 
- Geth would output `f3cb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a35580` for the zero-case

```
rlpdump -hex f3cb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a35580
[
  [
    7f000001,
    0cfa,
    15a8,
  ],
  0000000000000000000000000000000000000000000000000000000000001234,
  43b9a355,
  "",
]
```
- Geth would output the following for the `1`-case

```
rlpdump -hex f3cb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a35501
[
  [
    7f000001,
    0cfa,
    15a8,
  ],
  0000000000000000000000000000000000000000000000000000000000001234,
  43b9a355,
  01,
]
```

Now, however, geth willl reject the 'junk' at the end. 
On output, geth willl output the following for the `0`-case
```
 rlpdump -hex f2cb847f000001820cfa8215a8a000000000000000000000000000000000000000000000000000000000000012348443b9a355
[
  [
    7f000001,
    0cfa,
    15a8,
  ],
  0000000000000000000000000000000000000000000000000000000000001234,
  43b9a355,
]
```

This issue was found in hive

Screenshot_2021-07-02 https hivetests ethdevops io

Unfortunately, the test-framework also contained another false positive, which caused the 'legit' error from not always being shown (the false positive being ` PONG 'to' endpoint mismatch: got {IP:172.17.0.7 UDP:44025 TCP:44025}, want {IP:172.17.0.7 UDP:44025 TCP:0}`). 

I think the fix would be to just use `writeLongScalar` to write the ENR sequence number. 

cc @fjl @garyschulte 

Related geth-pr with some encoding examples: https://github.com/ethereum/go-ethereum/pull/23155 
Related hive logs: https://hivetests.ethdevops.io/?page=v-pills-results-tab&suite=1625138437-05d8c3c144ebe42dbca73551d1d2f73a.json

jflo:
There are clients that send PING messages without a from field. The devp2p spec doesn't indicate that these are optional, but we will try to handle them as if they were.

The encoding for ENRseq also was implemented incorrectly, and has been changed from a byte array to a 64bit scalar.

Both of these changes have been implemented to be backwards compatible with existing clients.

## Fixed Issue(s)

fixes #2496 

## Changelog


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PingPacketData.java
```diff
@@ -56,48 +59,62 @@ private PingPacketData(
...
   }
 
   public static PingPacketData readFrom(final RLPInput in) {
     in.enterList();
     // The first element signifies the "version", but this value is ignored as of EIP-8
     in.readBigIntegerScalar();
-    final Optional<Endpoint> from = Endpoint.maybeDecodeStandalone(in);
-    final Endpoint to = Endpoint.decodeStandalone(in);
+    Optional<Endpoint> from = Optional.empty();
+    Optional<Endpoint> to = Optional.empty();
```
garyschulte, line 84: initializing `to` isn't strictly necessary here.  nitpick

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PingPacketData.java
```diff
@@ -56,48 +59,62 @@ private PingPacketData(
...
         enrSeq = UInt64.valueOf(in.readLongScalar());
+        LOG.debug("read PING enr as long scalar");
       } catch (MalformedRLPInputException malformed) {
+        LOG.debug("failed to read PING enr as scalar, trying to read bytes instead");
         enrSeq = UInt64.fromBytes(in.readBytes());
       }
     }
     in.leaveListLenient();
-    return new PingPacketData(from, to, expiration, enrSeq);
+    return new PingPacketData(from, to.get(), expiration, enrSeq);
```
garyschulte, line 108: silly question.  do we care if there is a `to` field on a decoded ping packet?  I don't see the `to` field getting used for anything outside of tests, string serialization, and and `writeTo`.    `writeTo` implies we are sending it rather than receiving it...  So for a decoded packet, is `to` useless?  Do we repeat decoded ping messages?

If we are receiving this packet, maybe we should relax the to requirement and use `.orElse()` or perhaps change the `to` member to Optional<Endpoint> maybeTo and allow malformed ping packet decoding where there are no endpoints at all. 🤔

garyschulte (commented):
open question about the value of a ping packet with a missing `to` field.  Otherwise LGTM

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PingPacketData.java
jflo: Thats a very good question. You would think the `to` field is far more optional than the `from` field. If we are ignoring what is in the PING in favor of address values at lower layers, then all address fields should be treated as optional. I'll raise this with ACD.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PingPacketData.java
```diff
@@ -56,48 +59,62 @@ private PingPacketData(
...
   public static PingPacketData readFrom(final RLPInput in) {
     in.enterList();
     // The first element signifies the "version", but this value is ignored as of EIP-8
     in.readBigIntegerScalar();
-    final Optional<Endpoint> from = Endpoint.maybeDecodeStandalone(in);
-    final Endpoint to = Endpoint.decodeStandalone(in);
+    Optional<Endpoint> from = Optional.empty();
+    Optional<Endpoint> to = Optional.empty();
+    if (in.nextIsList()) {
+      to = Endpoint.maybeDecodeStandalone(in);
```
matkt, line 86: are you managing the case where the list is present but we don't have an IP, just the port (from field)

jflo: nope, i did not consider that. I'll have to add that.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/Endpoint.java
```diff
@@ -36,13 +35,11 @@
...
 public class Endpoint {
-  private final String host;
+  private final Optional<String> host;
   private final int udpPort;
   private final Optional<Integer> tcpPort;
 
-  public Endpoint(final String host, final int udpPort, final Optional<Integer> tcpPort) {
-    checkArgument(
-        host != null && InetAddresses.isInetAddress(host), "host requires a valid IP address");
+  public Endpoint(final Optional<String> host, final int udpPort, final Optional<Integer> tcpPort) {
```
garyschulte, line 42: non-blocking nitpick, but this signature implies host is optional.  IMO it would be better to expose host as essential, but handle it being nullable internally.  e.g.:
```this.host =  Optional.ofNullable(host)```

$0.02

garyschulte (approved):
non-blocking feedback about exposing host as Optional, otherwise 🚢

PR #2576, opened 2021-07-25
Add support for mining ommers, up to 8 blocks behind the head job

### LINKED ISSUE #2572: block creator: allow submitting 8-block deep uncle solutions
### Description
Besu is capable of mining the latest block only. Like in Geth, it should also accept 8-block deep uncles besides the latest pending block.

```java
public boolean submitSolution(final PoWSolution solution) {
    final Optional<PoWSolverJob> jobSnapshot = currentJob;
    if (jobSnapshot.isEmpty()) {
      LOG.debug("No current job, rejecting miner work");
      return false;
    }

    final PoWSolverJob job = jobSnapshot.get();
    final PoWSolverInputs inputs = job.getInputs();
    if (!inputs.getPrePowHash().equals(solution.getPowHash())) {
      // ! REJECTS POTENTIAL UNCLE HERE
      LOG.debug("Miner's solution does not match current job");
      return false;
    }
    final Optional<PoWSolution> calculatedSolution = testNonce(inputs, solution.getNonce());

    if (calculatedSolution.isPresent()) {
      LOG.debug("Accepting a solution from a miner");
      currentJob.get().solvedWith(calculatedSolution.get());
      return true;
    }
    LOG.debug("Rejecting a solution from a miner");
    return false;
  }
```

atoulme:

## PR description
This allows miners to send solutions for blocks up to 8 blocks behind head, to be included as ommers.

## Fixed Issue(s)
Fixes #2572

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
@@ -26,10 +26,14 @@ import java.util.concurrent.TimeUnit;
 
 import com.google.common.base.Stopwatch;
 import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.concurrent.ExpiringMap;
 import org.apache.tuweni.units.bigints.UInt256;
 
 public class PoWSolver {
 
+  private static final long POW_JOB_TTL = 1000 * 60 * 5; // 5 minutes
+  private static final int MAX_OMMER_DEPTH = 8;
   private static final Logger LOG = getLogger();
 
   public static class PoWSolverJob {
@@ -80,6 +84,7 @@ public class PoWSolver {
   private final Subscribers<PoWObserver> ethHashObservers;
   private final EpochCalculator epochCalculator;
   private volatile Optional<PoWSolverJob> currentJob = Optional.empty();
+  private final ExpiringMap<Bytes, PoWSolverJob> currentJobs = new ExpiringMap<>();
 
   public PoWSolver(
       final Iterable<Long> nonceGenerator,
@@ -98,12 +103,13 @@ public class PoWSolver {
   public PoWSolution solveFor(final PoWSolverJob job)
       throws InterruptedException, ExecutionException {
     currentJob = Optional.of(job);
+    currentJobs.put(job.getInputs().getPrePowHash(), job, System.currentTimeMillis() + POW_JOB_TTL);
     if (stratumMiningEnabled) {
       ethHashObservers.forEach(observer -> observer.newJob(job.inputs));
     } else {
       findValidNonce();
     }
-    return currentJob.get().getSolution();
+    return job.getSolution();
   }
 
   private void findValidNonce() {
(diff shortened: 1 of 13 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
```diff
@@ -26,10 +26,14 @@
 
 import com.google.common.base.Stopwatch;
 import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.concurrent.ExpiringMap;
 import org.apache.tuweni.units.bigints.UInt256;
 
 public class PoWSolver {
 
+  private static final long POW_JOB_TTL = 1000 * 60 * 5; // 5 minutes
```
matkt, line 35: is this value related to something ?

atoulme: Not really. 13s per block on average, so 13 times 8 is 108s, times 2 makes 3 minutes and a half, and buffered to 5 minutes to ensure we catch stragglers. We can use a different approach. I have another ticket to add meters to count accepted solutions vs rejected solutions.

matkt: ok it was to make sure I didn't miss anything. did you had the opportunity to test it ? I can test it on ropsten if you want

atoulme: I did not. I kinda wanted folks to review as well. I added some unit tests to make sure it made sense.

shemnon: My instinct says this should be configurable.  Probably via an `--X` flag.

atoulme: Is 5 minutes an ok default?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
```diff
@@ -149,22 +155,50 @@ public void cancel() {
...
+      return false;
+    }
+    final PoWSolverInputs inputs = jobToTestWith.getInputs();
+
     final Optional<PoWSolution> calculatedSolution = testNonce(inputs, solution.getNonce());
 
     if (calculatedSolution.isPresent()) {
       LOG.debug("Accepting a solution from a miner");
-      currentJob.get().solvedWith(calculatedSolution.get());
+      currentJobs.remove(solution.getPowHash());
```
shemnon, line 200: Does the miner want multiple ommers minable?  Removing on the first solution precludes that.

atoulme: Right now, we only accept one solution per block (the `solveFor` method blocks until the future completes). This repeats the same mechanism but with ommer blocks. We could potentially choose to allow more than one solution per block, turn those into ommers and replace the proposed block with the block with the best nonce. That's why I wanted a review, I'm just not sure how far to push this.

atoulme: To be clear, that'd be best addressed as a separate PR, because changing `solveFor` feels like a change worth doing on its own, whether we accept mining ommers or not.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
shemnon: maybe 2x high?  But since we are just keeping block headers (right?) from 5 min ago unless things get _very_ hairy I don't see it causing resource problems.  If we were keeping TXes it could be problematic.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
```diff
@@ -149,22 +155,50 @@ public void cancel() {
...
     if (jobSnapshot.isEmpty()) {
       LOG.debug("No current job, rejecting miner work");
       return false;
     }
 
-    final PoWSolverJob job = jobSnapshot.get();
-    final PoWSolverInputs inputs = job.getInputs();
-    if (!inputs.getPrePowHash().equals(solution.getPowHash())) {
-      LOG.debug("Miner's solution does not match current job");
+    PoWSolverJob headJob = jobSnapshot.get();
```
garyschulte, line 164: might as well get rid of jobSnapshot reference.

shemnon:
Plumb the `--X` flag through to configure retention and I'll approve it unless someone from ConsenSys objects.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
atoulme: Yep, only block headers. It should be fine.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
```diff
@@ -26,11 +26,15 @@
 
 import com.google.common.base.Stopwatch;
 import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.concurrent.ExpiringMap;
 import org.apache.tuweni.units.bigints.UInt256;
 
 public class PoWSolver {
 
+  private static final int MAX_OMMER_DEPTH = 8;
```
shemnon, line 35: For another PR, another `--X` candidate.

shemnon (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PoWSolver.java
atoulme: https://github.com/hyperledger/besu/pull/2584 is open

PR #2577, opened 2021-07-26
EIP1559 Cleanup : remove experimental flags 

matkt:

## PR description

We need to clean the ExperimentalEIPs class by removing this flags

This was configurable when developing the EIP now the values have been defined in the spec and cannot change via a flags without creating a consensus bug issue on the mainnet

## Changelog

### DIFF
--- a/config/src/main/java/org/hyperledger/besu/config/experimental/ExperimentalEIPs.java
+++ b/config/src/main/java/org/hyperledger/besu/config/experimental/ExperimentalEIPs.java
@@ -14,31 +14,8 @@
  */
 package org.hyperledger.besu.config.experimental;
 
-import picocli.CommandLine.Option;
-
 /**
  * Flags defined in this class must be used with caution, and strictly reserved to experimental
  * EIPs.
  */
-public class ExperimentalEIPs {
-  // To make it easier for tests to reset the value to default
-  public static final long EIP1559_BASEFEE_DEFAULT_VALUE = 1000000000L;
-
-  @Option(
-      hidden = true,
-      names = {"--Xeip1559-basefee-max-change-denominator"},
-      arity = "1")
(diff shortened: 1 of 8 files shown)


config/src/main/java/org/hyperledger/besu/config/experimental/ExperimentalEIPs.java
```diff
@@ -24,21 +24,9 @@
   // To make it easier for tests to reset the value to default
   public static final long EIP1559_BASEFEE_DEFAULT_VALUE = 1000000000L;
```
RatanRSur, line 25: What do you think about deleting this class entirely and moving these constants? There are no `ExperimentalEIPs` anymore.

garyschulte (commented):
LGTM, will :+1: after you push the delete of ExperimentalEIPs

config/src/main/java/org/hyperledger/besu/config/experimental/ExperimentalEIPs.java
```diff
@@ -24,21 +24,9 @@
...
-      hidden = true,
-      names = {"--Xeip1559-basefee-max-change-denominator"},
-      arity = "1")
-  public static Long basefeeMaxChangeDenominator = 8L;
-
   @Option(
       hidden = true,
       names = {"--Xeip1559-initial-base-fee"},
       arity = "1")
   public static Long initialBasefee = EIP1559_BASEFEE_DEFAULT_VALUE;
```
garyschulte, line 31: Do we still have a use case for this cli option?  it seems awfully niche and is otherwise configurable via genesis file

garyschulte (approved):
lets merge as is and address the feedback on follow-up PRs

jflo (approved):

PR #2580, opened 2021-07-27
Created PkiQbftExtraData

lucassaldanha:
## PR description

- Created `PkiQbftExtraData` and `PkiQbftExtraDataCodec`

## Changelog


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
```diff
@@ -0,0 +1,124 @@
...
+    } else {
+      vote = Optional.of(decodeVote(rlpInput));
+    }
+
+    final int round = rlpInput.readIntScalar();
+    final List<SECPSignature> seals =
+        rlpInput.readList(
+            rlp -> SignatureAlgorithmFactory.getInstance().decodeSignature(rlp.readBytes()));
+
+    final Optional<Bytes> cms;
```
jframe, line 60: Unless you care about the cms strictly being at the end of the list you simplify the encoding to just create another rlp list with the qbft extra data and the cms i.e. [qbftExtraData, cms] where qbftExtraData = [vanity, vote, validators, round, seals]

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
```diff
@@ -0,0 +1,124 @@
...
+    if (rlpInput.nextIsList() && rlpInput.nextSize() == 0) {
```
jframe, line 61: I thought the cms was mandatory?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
jframe, line 60: Is there a richer data type for the cms? thought was a pki signature of some sort

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
```diff
@@ -0,0 +1,124 @@
...
+      } else {
+        encoder.writeEmptyList();
+      }
+    } else {
+      encoder.writeIntScalar(0);
+      encoder.writeEmptyList();
+    }
+
+    if (encodingType == EncodingType.ALL) {
+      if (extraData.getCms().isPresent()) {
```
jframe, line 108: Looks like the cms is excluded when calculating the onchain block hash and committed seal? Seems like you probably always want the cms to be included as part of the hash?

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodecTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+  private final String EMPTY_CMS_RAW_HEX_ENCODING_STRING =
+      "0xf8f1a00102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20ea940000000000000000"
+          + "000000000000000000000001940000000000000000000000000000000000000002d7940000000000000000"
+          + "00000000000000000000000181ff83fedcbaf886b841000000000000000000000000000000000000000000"
+          + "0000000000000000000001000000000000000000000000000000000000000000000000000000000000000a"
+          + "00b841000000000000000000000000000000000000000000000000000000000000000a0000000000000000"
+          + "00000000000000000000000000000000000000000000000100c0";
+
+  // Arbitrary bytes representing an non-empty CMS
```
jframe, line 53: nit: a instead of an

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
lucassaldanha: When the extra data is created, there is no cms. We add the cms later on (as a "stamp").

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
lucassaldanha: We can't have the CMS as part of the on-chain hash calculation because it has the block hash inside of it as a signed payload.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
lucassaldanha: There is but it is tied to the Bouncy Castle API. I have decided to keep it "raw" in the extra data so we don't add a hard dependency on any library at this point.

lucassaldanha: I have implemented your suggestion. However, I've settled on `[qbftExtraData, [cms]]`. I've added the CMS field into its own list. In the case we want to add other PKI-related elements in the future, it is easier to keep them aggregated.

jframe (approved):

PR #2581, opened 2021-07-27
Add ability to expose getWork/submitWork JSON-RPC methods over the stratum port

### LINKED ISSUE #2065: Add support for JSON-RPC getWork/submitWork from the stratum port
The current stratum port implementation works well with stratum1 and ethproxy methods. However, when this port is opened to the world, we see users point at this port and perform HTTP requests.
1. We should simply ignore HTTP traffic that doesn't fit the stratum pattern. It is not interesting.
2. Because miners out there get misconfigured and folks may only have ability to use with getWork/submitWork, we should port over from the HTTP JSON-RPC service the 2 methods eth_getWork and eth_submitWork. These methods can then be safely exposed without exposing the JSON-RPC service.

atoulme:

## PR description
Adds the JSON-RPC methods getWork/submitWork to the stratum port, so miners can communicate with Besu over the stratum port. This allows to keep the JSON-RPC port closed to miner traffic.

## Fixed Issue(s)
Fixes #2065

## Changelog

### DIFF
--- a/ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumConnection.java
+++ b/ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumConnection.java
@@ -19,6 +19,8 @@ import static org.apache.logging.log4j.LogManager.getLogger;
 import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
 import java.util.function.Consumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.google.common.base.Splitter;
 import io.vertx.core.buffer.Buffer;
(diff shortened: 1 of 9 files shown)


ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumConnection.java
```diff
@@ -60,23 +64,49 @@ void handleBuffer(final Buffer buffer) {
...
+        }
       }
     }
   }
 
-  void close(final Void aVoid) {
+  void close() {
     if (protocol != null) {
       protocol.onClose(this);
     }
```
shemnon, line 112: Possibly out of scope, but in `handleMessage()` (next method) if the protocol is null the protocol is set but then the message is not handled.  `canHandle` is called but handling is never performed.  should `canHandle` be `maybeHandle` if it also conditionally handles?

shemnon (approved):

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumConnection.java
atoulme: That's a better name, yes. It changed this time around since this new protocol needs to handle the first message, when others would simply ignore it.

Issue #2579, opened 2021-07-27
misleading warning on startup: --min-gas-price ignored


macfarla:
The following warning is being reported: "--min-gas-price ignored because none of --miner-enabled or isQuorum (in genesis file) was defined." 

Using the quickstart. 

### Description
As an [Actor], I want [feature] so that [why]. 

### Acceptance Criteria
* [Criteria 1]

### Steps to Reproduce (Bug)
1. [Step 1]
2. [Step 2]
3. [Step ...]

**Expected behavior:** [What you expect to happen]

**Actual behavior:** [What actually happens]

**Frequency:** [What percentage of the time does it occur?]

### Versions (Add all that apply)
* Software version: [`besu --version`]
* Java version: [`java -version`]
* OS Name & Version: [`cat /etc/*release`]
* Kernel Version: [`uname -a`]
* Virtual Machine software & version: [`vmware -v`]
* Docker Version: [`docker version`]
* Cloud VM, type, size: [Amazon Web Services I3-large]

drinkcoffee:
This is very confusing when some other issue is causing the quick start to not work (in my case it was because of networking issues). Zero gas cost transactions are still accepted. As such, what is being shown in the error message is very misleading

drinkcoffee:
Thank you for fixing this @macfarla !

PR #2582, opened 2021-07-27
simplify --min-gas-price dependency check

macfarla:

See #2579 
For goQuorum mode, min-gas-price must be zero and this is checked separately so I think this warning is adequate. 

## Changelog


mark-terry (approved):
LGTM.

PR #2583, opened 2021-07-27
Add stratum metrics

### LINKED ISSUE #2325: Expose metrics related to mining
### Description
As an operator, I want to know about the population of miners connected to my client to understand how mining is going.

### Acceptance Criteria
* Expose metrics using the metrics system that capture the following:
** current difficulty
** number of connected miners
** number of connections over time
** number of disconnections over time

atoulme:

## PR description
Adds a few metrics to record the stratum port activity.

## Fixed Issue(s)
Fixes #2325

## Changelog


ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
```diff
@@ -126,6 +158,12 @@ public void newJob(final PoWSolverInputs poWSolverInputs) {
     for (StratumProtocol protocol : protocols) {
       protocol.setCurrentWorkTask(poWSolverInputs);
     }
+    UInt256 difficulty = poWSolverInputs.getTarget().divide(divisor);
+    if (!difficulty.fitsLong()) {
```
shemnon, line 162: Prometheus uses floating points in the transport.  Why trim to long when we could just convert to Double?

shemnon: We will be in floating point quite a bit before Long.MAX_VALUE, it only seems to get 9 digits of precision.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
```diff
@@ -39,20 +44,27 @@
 public class StratumServer implements PoWObserver {
 
   private static final Logger logger = getLogger();
+  private static final UInt256 divisor =
```
shemnon, line 47: `ALL_CALS` and more descriptive?  `BIGINT_TWO_E_256`?
(also, it's bigger than a UInt256 so the type is wrong).

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
atoulme: We don't just support Prometheus anymore, we have OpenTelemetry now ❤️ 
I can switch to Double to make sure the value always fits, good point.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
atoulme: Not sure I get this. The value is 1 * 10^32, I think I can maybe represent this in a less obtuse way.

shemnon: I had my brain in hex.

atoulme: I was incorrect, it's 2^256.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
```diff
@@ -39,20 +45,26 @@
...
 
   private static final Logger logger = getLogger();
+  private static final UInt256 DIFFICULTY_1_TARGET = UInt256.valueOf(2).pow(256);
 
   private final Vertx vertx;
   private final int port;
   private final String networkInterface;
   private final AtomicBoolean started = new AtomicBoolean(false);
+  private final AtomicLong numberOfMiners = new AtomicLong(0);
+  private final AtomicDouble difficultyGauge = new AtomicDouble(0.0);
```
shemnon, line 55: It's not actually a gauge, we get the gauge from `metricsSystem.createGauge`.  Perhaps `currentDifficulty`?

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
```diff
@@ -89,11 +114,18 @@ public StratumServer(
   }
 
   private void handle(final NetSocket socket) {
+    connectionsCount.inc();
+    numberOfMiners.incrementAndGet();
```
shemnon, line 118: It's interesting the atomic APIs don't have a simple `imcrement` or `add` method, they all require returning a value. but `set` works just fine. </rant>

shemnon (approved):
Approved once the field is renamed.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
atoulme: Indeed, I was looking for it. It's ok, not a big deal.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
atoulme: alright

atoulme: done

PR #2584, opened 2021-07-27
Add unstable CLI option for max ommers depth

atoulme:

## PR description
Add an unstable CLI parameter to specify the max ommers depth, with a default value of 8.

## Fixed Issue(s)
Relates to issue #2572 and PR #2576

## Changelog


shemnon:
The `MinigParameters` constructor is too large and should probably move to a builder pattern.  In `BesuCommand` the construction should move to a method for readability, and in the tests all default options should not be set.  This will also improve readability so you know what `false` goes to which option.

shemnon (approved):

PR #2586, opened 2021-07-28
Feature/1559 cleanup remove experimentaleip

garyschulte:
## PR description
Remove ExperimentalEIPs
remove CLI option for initial baseFee
plumb FeeMarket into ProtocolSpec

This PR stops short of moving the EIP1559 object/methods/data into the london FeeMarket to keep the scope bounded.  A subsequent cleanup PR will tackle that.

## Changelog


ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -160,7 +160,7 @@ void parse(
     final CommandLine commandLine = new CommandLine(this);
     commandLine.addMixin("Dagger Options", daggerOptions);
     // Usage of static command line flags is strictly reserved for experimental EIPs
```
macfarla, line 162: now that it's no longer experimental - need to make it non-static

config/src/main/java/org/hyperledger/besu/config/FeeMarketConfigOptions.java
```diff
@@ -12,21 +12,16 @@
...
+import picocli.CommandLine;
 
-/**
- * Flags defined in this class must be used with caution, and strictly reserved to experimental
- * EIPs.
- */
-public class ExperimentalEIPs {
-  // To make it easier for tests to reset the value to default
-  public static final long EIP1559_BASEFEE_DEFAULT_VALUE = 1000000000L;
+public class FeeMarketConfigOptions {
```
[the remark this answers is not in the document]
garyschulte: This is the only one currently.  I am not sure how much value it is going to provide however, since it is just a configuration for the transition block when the block is not at genesis.  Once mainnet and the major test nets have transitioned, I doubt it will be of much actual use.  That was why I was reluctant to plumb it into the main configuration options and usage.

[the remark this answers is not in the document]
garyschulte: The other experimental eip1559 options like slack coefficient and max change denominator have already been dropped.  Seeing as how we do not have an existing use case for modifying these values *and* geth is specifying these values should be ignored for mainnet:
 https://github.com/ethereum/go-ethereum/blob/master/core/genesis.go#L60-L65
 
 I am just going to remove the initial basefee option and get rid of this class entirely.  If a compelling use case comes along for a 1559-style fee market that uses different constants, we can implement it at that time.  Until then, imo the fewer arcane options the better.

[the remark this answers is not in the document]
macfarla: 👍

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionEIP1559Test.java
```diff
@@ -39,22 +40,24 @@ public void buildEip1559Transaction() {
...
+                    Bytes32.fromHexString("0x34Ee1552077D197D511b3C615459b8E597d9eE04"))));
     final Transaction tx =
         Transaction.builder()
-            .chainId(new BigInteger("1559", 10))
+            .chainId(new BigInteger("1981", 10))
             .nonce(0)
             .value(Wei.ZERO)
             .gasLimit(30000)
-            .maxPriorityFeePerGas(Wei.of(2))
+            //            .maxPriorityFeePerGas(Wei.of(2))
```
jflo, line 52: Any reason to keep these commented out lines in here?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -259,13 +258,14 @@ public PendingTransactions getPendingTransactions() {
...
-        && !eip1559
-            .map(eip1559 -> eip1559.isEIP1559(chainHeadBlockHeader.getNumber()))
-            .orElse(false)) {
-      return ValidationResult.invalid(
-          TransactionInvalidReason.INVALID_TRANSACTION_FORMAT,
-          "EIP-1559 transaction are not allowed yet");
+    if (transaction.getType().equals(TransactionType.EIP1559)) {
+      final long blknum = chainHeadBlockHeader.getNumber();
+      ProtocolSpec spec = protocolSchedule.getByBlockNumber(blknum);
+      if (!spec.getEip1559().map(eip1559 -> eip1559.isEIP1559(blknum)).orElse(false)) {
```
jflo, line 264: This is an improvement on readability. Could we take it a step further and push all the navigation of the Optional into the Protocol spec itself?

jflo: might come in handy below as well.

jflo (approved):
A couple non-blocking notes. This is a big improvement on readability.

jflo (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
garyschulte: I am going to defer this until the subsequent cleanup PR, because I am intending to remove the Optional<EIP1559> entirely.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionEIP1559Test.java
garyschulte: fixed!

PR #2592, opened 2021-07-28
BFT validator RPCs to use validator provider

jframe:
## PR description
Update BFT rpcs for metrics, validators querying to use the validator provider so it will work with transaction based voting in qbft.

## Changelog


usmansaleem (approved):
LGTM

usmansaleem (approved):
Looks good to me.

PR #2593, opened 2021-07-29
Update Spotless and google-java-format

shemnon:
## PR description

Update spotless to 5.14.2 and java-google-format to 1.10.0.
Both of these are needed to support Java 16 and Gradle 7.x.

## Changelog


atoulme (approved):
LGTM

PR #2594, opened 2021-07-29
Bugfix/2546 retesteth clock

### LINKED ISSUE #2546: Reject on high timestamp difference for blocks - why and why isn't it consistent
### Description

When I run a test case with a large difference in timestamps between two consecutive blocks, I get this error in the log:
```
2021-07-17 17:06:42.628+00:00 | vert.x-worker-thread-19 | INFO  | TimestampBoundedByFutureParameter | Invalid block header: timestamp 20000010000 is greater than the timestamp margin 1626541617
```

And the block is rejected with this RPC reply:
```
{
  "jsonrpc" : "2.0",
  "id" : 13,
  "error" : {
    "code" : -32000,
    "message" : "Could not import Block"
  }
}
```

This seems weird, for two reasons:

1. I can't find in the Yellow Paper any criteria for the block timestamp except for inequality (49) on page 7 (see https://ethereum.github.io/yellowpaper/paper.pdf). So if we ever had a lull of about thirty minutes during which no blocks were created, it would break consensus.

2. Even if there is such a limit, shouldn't it be consistent? I ran the same test twice with a minute between them, and the limit seems to grow by one microsecond every second (see the first and last lines):

```
2021-07-17 17:15:27.477+00:00 | vert.x-worker-thread-5 | INFO  | TimestampBoundedByFutureParameter | Invalid block header: timestamp 20000010000 is greater than the timestamp margin 1626542142
2021-07-17 17:16:29.222+00:00 | vert.x-worker-thread-6 | INFO  | ProtocolScheduleBuilder | Protocol schedule created with milestones: [London: 0]
2021-07-17 17:16:29.247+00:00 | vert.x-worker-thread-13 | INFO  | CalculatedDifficultyValidationRule | Invalid block header: difficulty 131072 does not equal expected difficulty 131136
2021-07-17 17:16:29.254+00:00 | vert.x-worker-thread-16 | INFO  | CalculatedDifficultyValidationRule | Invalid block header: difficulty 131072 does not equal expected difficulty 3686656
2021-07-17 17:16:29.257+00:00 | vert.x-worker-thread-17 | INFO  | CalculatedDifficultyValidationRule | Invalid block header: difficulty 131072 does not equal expected difficulty 131136
2021-07-17 17:16:29.260+00:00 | vert.x-worker-thread-18 | INFO  | TimestampBoundedByFutureParameter | Invalid block header: timestamp 20000010000 is greater than the timestamp margin 1626542204
```

garyschulte:
## Fixed Issue(s)

fixes #2546 
expose a protected method to set/update the clock in TimestampBoundedByFutureParameter in order to support the mutable RetestethClock

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/TimestampBoundedByFutureParameter.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/TimestampBoundedByFutureParameter.java
@@ -18,7 +18,6 @@ import org.hyperledger.besu.ethereum.core.BlockHeader;
 import org.hyperledger.besu.ethereum.mainnet.DetachedBlockHeaderValidationRule;
 
 import java.time.Clock;
-import java.util.concurrent.TimeUnit;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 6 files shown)


garyschulte:
Do not merge, there are outstanding issues with this change when running reference tests with retesteth engine

garyschulte:
Making this change breaks many existing reference tests.  Closing for now.

PR #2596, opened 2021-07-29
Gradle 7.1.1 and Dependency Version Upgrades

shemnon:
## PR description
 
Upgrade gradle to 7.1.1.
Upgrade most dependencies to current versions

## Changelog


atoulme (approved):
LGTM

shemnon:
Not all versions were updated

* Versions with non-release markings (rc, alpha, etc) we were not already using were skipped.
* Others didn't merge simply or cleanly and will be left for another PR
  * Dagger 
  * Kubernetes client
  * OpenTelemetry
  * JMH
* Some dependencies advertised major version upgrades and were not taking to the most recent version
  * vertx
  * web3j
* Rocks DB should have performance testing before upgrading

atoulme (approved):
Still looking good.

nat/src/main/java/org/hyperledger/besu/nat/upnp/BesuUpnpServiceConfiguration.java
```diff
@@ -229,6 +229,11 @@ public Executor getRegistryListenerExecutor() {
     return executorService;
   }
 
+  @Override
+  public Executor getRemoteListenerExecutor() {
+    return executorService;
+  }
+
```
macfarla, line 236: is this change related to the gradle upgrade?

shemnon: It's related to the UPNP version upgrade.  They needed a new method implemented as they added a method with a non-default implementation to one of their interfaces (in violation of semver).

PR #2602, opened 2021-08-02
Qbft validator contract config

jframe:
## PR description
Qbft specific config with validator contract address option

## Changelog


usmansaleem (approved):
LGTM

PR #2604, opened 2021-08-03
PKI QbftContext and QbftBesuControllerBuilder update

lucassaldanha:
## PR description

- Created `PKIQbftContext` that is used with Besu on PKI-QBFT mode.
- `QbftBesuControllerBuilder` uses the presence of the PKI keystore to decide if it should instantiate the PKI-specific classes 'PkiQbftExtraDataCodec' and 'PkiQbftContext'
- The wiring of the PKI keystore will be done in a follow-up PR. At the moment it will always be empty.

## Changelog


besu/src/main/java/org/hyperledger/besu/controller/QbftBesuControllerBuilder.java
```diff
@@ -262,11 +275,20 @@ protected BftContext createConsensusContext(
...
         convertBftForks(configOptions.getTransitions().getQbftForks());
 
-    return new BftContext(
-        BlockValidatorProvider.forkingValidatorProvider(
-            blockchain, epochManager, bftBlockInterface().get(), bftValidatorForkMap),
-        epochManager,
-        bftBlockInterface().get());
+    if (pkiKeyStore.isPresent()) {
+      return new PkiQbftContext(
+          BlockValidatorProvider.forkingValidatorProvider(
```
jframe, line 280: Would be nice to move the creation of the validatorProvider out of the if/else into it's own variable

jframe (approved):
LGTM

PR #2609, opened 2021-08-03
Fix bug with value transfer

### LINKED ISSUE #2606: ether value is not supported for private transactions
### Description
Updating to **21.7.1** currently all mainnet transactions fail with ``` ether value is not supported for private transactions ```
Transaction signature (r,s,v) with v == 37 or v == 38 

According to https://eips.ethereum.org/EIPS/eip-155; (replay protection)

chain_id == 1 ; v should be either 37 or 38. Minimum. 

But in here https://github.com/hyperledger/besu/blob/992c6563870ea87874e80b93d3dc6ce93dce880f/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java#L629

V = 37 & v = 38; should still be valid with mainnet txs unless if including the private txs fields such as privateFor, privateFrom, privacyGroupId, etc... 

What am i missing?

Note: With previous version **21.1.7** works fine

macfarla:

Clarify isGoQuorumPrivateTransaction method. v value is set directly for GoQuorum private transactions, and they don't have a chainId. All other tx derive their v value from the chainId.

fixes #2606 

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
@@ -626,11 +626,15 @@ public class Transaction
   /**
    * Returns whether or not the transaction is a GoQuorum private transaction. <br>
    * <br>
-   * A GoQuorum private transaction has its <i>v</i> value equal to 37 or 38.
+   * A GoQuorum private transaction has its <i>v</i> value equal to 37 or 38, and does not contain a
+   * chainId.
    *
    * @return true if GoQuorum private transaction, false otherwise
    */
   public boolean isGoQuorumPrivateTransaction() {
+    if (chainId.isPresent()) {
+      return false;
+    }
     return v.map(
             value ->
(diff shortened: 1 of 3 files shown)


garyschulte (approved):
LGTM!

lucassaldanha (changes requested):
We also need to fix the TransactionPool.

https://github.com/hyperledger/besu/blob/main/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java#L294-L297

It would be better to consolidate the logic into a single place.

macfarla:
> We also need to fix the TransactionPool.
> 
> https://github.com/hyperledger/besu/blob/main/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java#L294-L297
> 
> It would be better to consolidate the logic into a single place.

Updated. And added a TransactionPool test. Thanks @lucassaldanha

lucassaldanha (approved):

PR #2621, opened 2021-08-10
Added PKI QBFT Block Creation Behaviour

lucassaldanha:
## PR description
- Refactored QbftRound, introducing the `CreateBlockForProposalBehaviour` Functional Interface.
  -  The default implementation uses the `BlockCreator`
  - The PKI implementation has the logic to add the CMS message into the proposed block' extra data
- Added PKI Logic into `ProposalPayloadValidator` 

## Changelog


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
```diff
@@ -89,4 +104,58 @@ private boolean validateBlock(final Block block) {
 
     return true;
   }
+
+  /*
+    Validate CMS in block header (only for PKI-mode)
+  */
+  private boolean validateCms(final Block block) {
```
jframe, line 111: Think this would be nicer if it was separated out as it's own validator probably created in the ProposalValidator just like the ProposalPayloadValidator is

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
```diff
@@ -89,4 +104,58 @@ private boolean validateBlock(final Block block) {
...
+    final PkiQbftContext pkiQbftContext;
+    try {
+      pkiQbftContext = protocolContext.getConsensusState(PkiQbftContext.class);
+    } catch (final ClassCastException e) {
+      // TODO-lucas Do we have a better way of doing this?
+      // Not running on PKI-enabled mode
+      return true;
+    }
```
jframe, line 119: Perhaps pass the PkiBlockCreationConfiguration as an optional constructor argument to the MessageValidatorFactory you could then use that to test if PKI is enabled without the need to cast or use the context.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
```diff
@@ -89,4 +104,58 @@ private boolean validateBlock(final Block block) {
...
+
+    final BftExtraData extraData =
+        pkiQbftContext.getBlockInterface().getExtraData(block.getHeader());
+    final PkiQbftExtraData pkiExtraData = (PkiQbftExtraData) extraData;
+    final KeyStoreWrapper trustStore =
+        pkiQbftContext.getPkiBlockCreationConfiguration().getTrustStore();
+
+    if (pkiExtraData.getCms().isPresent()) {
+      // TODO-lucas Do we need to do this or could I just use 'block.getHeader().getHash()'
```
jframe, line 128: You should be able to use block.getHeader().getHash()

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
```diff
@@ -89,4 +104,58 @@ private boolean validateBlock(final Block block) {
...
+      final Hash proposedBlockHash =
+          BftBlockHeaderFunctions.forOnChainBlock(new PkiQbftExtraDataCodec())
+              .hash(block.getHeader());
+
+      LOG.info(">>> Validating CMS for block {}", proposedBlockHash);
```
jframe, line 133: nit: Is the >>> left over from debugging? Haven't noticed other logging using that style is all

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
```diff
@@ -89,4 +104,58 @@ private boolean validateBlock(final Block block) {
...
+      CertStore crlCertStore = null;
+      if (trustStore.getCRLs() != null) {
+        try {
+          crlCertStore =
+              CertStore.getInstance(
+                  "Collection", new CollectionCertStoreParameters(trustStore.getCRLs()));
+        } catch (final Exception e) {
+          throw new RuntimeException("Error reading PKI Block Creation TrustStore CRL file", e);
+        }
+      }
```
jframe, line 144: Seems like this something that should belong in the pki package alongside the CmsValidator?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
```diff
@@ -89,4 +104,58 @@ private boolean validateBlock(final Block block) {
...
+      final CmsValidator cmsValidator = new CmsValidator(trustStore, crlCertStore);
+
+      if (!cmsValidator.validate(pkiExtraData.getCms().get(), proposedBlockHash)) {
+        LOG.info(">>> CMS Validation INVALID for block {}", proposedBlockHash);
+        return false;
+      } else {
+        LOG.info(">>> CMS Validation VALID for block {}", proposedBlockHash);
+        return true;
+      }
+    } else {
```
jframe, line 155: return false in this case? Would have though that all proposals must have a CMS if PKI mode is being used

PR #2624, opened 2021-08-11
Feature/1559 cleanup tx price calc

### LINKED ISSUE #2559: Clean TransactionPriceCalculator
This class need a cleanup 

```java
public interface TransactionPriceCalculator {
  Wei price(Transaction transaction, Optional<Long> baseFee);

  static TransactionPriceCalculator frontier() {
    return (transaction, baseFee) -> transaction.getGasPrice().get();
  }

  static TransactionPriceCalculator eip1559() { `//change method name to london`
    return (transaction, maybeBaseFee) -> {
      final Wei baseFee = Wei.of(maybeBaseFee.orElseThrow());
      if (!transaction.getType().supports1559FeeMarket()) {
        return transaction.getGasPrice().get();
      }
      final Wei maxPriorityFeePerGas = transaction.getMaxPriorityFeePerGas().orElseThrow();
      final Wei maxFeePerGas = transaction.getMaxFeePerGas().orElseThrow();
      Wei price = maxPriorityFeePerGas.add(baseFee);
      if (price.compareTo(maxFeePerGas) > 0) {
        price = maxFeePerGas;
      }
      return price;
    };
  }
}
```

In some parts of the code we use either frontier or eip1559 depending on the type of transaction. This is not good practice. Do not use **frontier** after the fork. the **eip1559 price calculator** handles non-EIP1559 transactions after london without any issues

garyschulte:
## PR description
* roll TransactionPriceCalculator into FeeMarket
* create a LegacyFeeMarket to use prior to London
* remove explicit usages of TransactionPriceCalculator static constructors
* remove TransactionPriceCalculator from ProtocolSpec (added to FeeMarket)
* make FeeMarket non-optional in ProtocolSchedule
* create BaseFeeMarket interface for London

## Fixed Issue(s)

fixes #2559

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LegacyFeeMarket.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LegacyFeeMarket.java
@@ -30,11 +30,6 @@ public class LegacyFeeMarket implements FeeMarket {
   }
 
   @Override
-  public boolean implementsBaseFee() {
-    return false;
-  }
-
-  @Override
   public TransactionPriceCalculator getTransactionPriceCalculator() {
     return txPriceCalculator;
   }
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LegacyFeeMarket.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LegacyFeeMarket.java
@@ -16,40 +16,22 @@ package org.hyperledger.besu.ethereum.mainnet.feemarket;
 
 import org.hyperledger.besu.ethereum.core.Transaction;
 import org.hyperledger.besu.ethereum.core.Wei;
-import org.hyperledger.besu.ethereum.core.fees.BaseFee;
 import org.hyperledger.besu.ethereum.core.fees.TransactionPriceCalculator;
 
 import java.util.Optional;
 import java.util.function.Supplier;
 
-public class LondonFeeMarket implements BaseFeeMarket {
-  static final Long DEFAULT_BASEFEE_INITIAL_VALUE = 1000000000L;
-  static final Long DEFAULT_BASEFEE_MAX_CHANGE_DENOMINATOR = 8L;
-  static final Long DEFAULT_SLACK_COEFFICIENT = 2L;
+public class LegacyFeeMarket implements FeeMarket {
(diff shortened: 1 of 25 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LegacyFeeMarket.java
```diff
@@ -0,0 +1,47 @@
...
+  private final TransactionPriceCalculator txPriceCalculator;
+
+  public LegacyFeeMarket() {
+    this.txPriceCalculator = TransactionPriceCalculator.frontier();
+  }
+
+  @Override
+  public boolean implementsBaseFee() {
+    return false;
+  }
```
RatanRSur, line 35: Not saying we should do these but other options to consider:
* make this method `default` in `FeeMarket` and checks if the instance is `instanceof BaseFeeMarket`
* the calling context checks directly if the `FeeMarket` is `instanceof BaseFeeMarket` (this prevents these boolean methods from multiplying`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -80,10 +76,10 @@
...
   private final PeerTransactionTracker peerTransactionTracker;
   private final PeerPendingTransactionTracker peerPendingTransactionTracker;
-  private final TransactionPriceCalculator frontierPriceCalculator =
-      TransactionPriceCalculator.frontier();
-  private final TransactionPriceCalculator eip1559PriceCalculator =
-      TransactionPriceCalculator.eip1559();
+  //  private final TransactionPriceCalculator frontierPriceCalculator =
+  //      TransactionPriceCalculator.frontier();
+  //  private final TransactionPriceCalculator eip1559PriceCalculator =
+  //      TransactionPriceCalculator.eip1559();
```
RatanRSur, line 82: commented out code

RatanRSur (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LegacyFeeMarket.java
garyschulte: yeah good idea.  I am not a fan of instanceof, but default and an override default is pretty clean :+1:

PR #2626, opened 2021-08-11
Subcall spinnig

jflo:
## PR description

Adds caching for jumpdest analysis. Since contract code is immutable, there is no reason to re-compute the valid jump destinations every time a contract is run. This PR only solves a part of the problem during nested contract calls. There remains the issue of first time, repeated calls as would occur during a thundering herd event such as an NFT drop.

In order to even test this, it required opening up some class access modifiers due to the lack of an kind of inversion of control mechanism. This PR needs to be left in DRAFT status while that is discussed. It is likely that broadening access is not worth the increased testability.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetEvmRegistries.java
```diff
@@ -145,7 +145,7 @@ static EVM istanbul(final GasCalculator gasCalculator, final BigInteger chainId)
     return new EVM(registry, gasCalculator);
   }
 
-  static EVM london(final GasCalculator gasCalculator, final BigInteger chainId) {
+  public static EVM london(final GasCalculator gasCalculator, final BigInteger chainId) {
```
jflo, line 148: Same story here as with the gas calculator.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
```diff
@@ -36,6 +40,43 @@
...
+
+  private CodeLoader loader = new CodeLoader();
+
+  public CodeLoader getLoader() {
+    return loader;
+  }
+
+  public void setLoader(final CodeLoader cacheLoader) {
+    this.loader = cacheLoader;
+    this.codeCache =
```
atoulme, line 68: I'm a bit lost here. Why is there a separate codeCache field here?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
```diff
@@ -36,6 +40,43 @@
       new OperationResult(
           Optional.empty(), Optional.of(ExceptionalHaltReason.INSUFFICIENT_STACK_ITEMS));
 
+  public static class CodeLoader extends CacheLoader<Account, Code> {
```
atoulme, line 43: Please move the class to its own file and add unit tests for it?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
```diff
@@ -36,6 +40,43 @@
...
+    @Override
+    public Code load(final Account key) throws Exception {
+      if (!key.hasCode()) {
+        throw new IllegalArgumentException("account " + key + " has no executable code");
```
atoulme, line 47: this needs a test. This feels like a good way to harm Besu, one can craft an invalid transaction that makes it throw exceptions.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
```diff
@@ -36,6 +40,43 @@
...
+        CacheBuilder.newBuilder()
+            .maximumWeight(1024L * 1024L)
+            .weigher(this.scale)
+            .build(this.loader);
+  }
+
+  private final CodeScale scale = new CodeScale();
+
+  private LoadingCache<Account, Code> codeCache =
+      CacheBuilder.newBuilder().maximumWeight(1024L * 1024L).weigher(this.scale).build(this.loader);
```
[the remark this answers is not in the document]
atoulme: I would add a bit of config there. Also, why is it cached per Call operation? Wouldn't it be better to cache across calls?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
```diff
@@ -36,6 +40,43 @@
...
+  }
+
+  private static class CodeScale implements Weigher<Account, Code> {
+    @Override
+    public int weigh(final Account key, final Code value) {
+      return value.getSize();
+    }
+  }
+
+  private CodeLoader loader = new CodeLoader();
```
atoulme, line 60: final?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
```diff
@@ -36,6 +40,43 @@
...
+    @Override
+    public Code load(final Account key) throws Exception {
+      if (!key.hasCode()) {
+        throw new IllegalArgumentException("account " + key + " has no executable code");
+      }
+      return new Code(key.getCode());
+    }
+  }
+
+  private static class CodeScale implements Weigher<Account, Code> {
```
atoulme, line 53: Pretty sure this can be a lambda instead?

jflo:
Closing after discussion, since this would be extremely difficult to evict correctly on contract self destruct. We are likely going to have to pursue a much broader implementation at the world state or EVM level.

atoulme:
Where was the discussion?

PR #2632, opened 2021-08-13
refactor of  GasLimitCalculator into protocolSchedule

### LINKED ISSUE #2561: Clean GasLimitRangeAndDeltaValidationRule
Maybe rework **GasLimitRangeAndDeltaValidationRule** we have the same code on **AbstractBlockCreator**. We need this code to multiply the gaslimit by the slackCoefficient on fork block. we need to use a specific gasLimitCalculator fot that (ex **LongonGasLimitCalculator**)

**Step :** 
 1)  Move the TargetingGasLimitCalculator to the same package as AbstractGasLimitSpecification 
 2) Rename TargetingGasLimitCalculator to FrontierTargetingGasLimitCalculator  
 3) Create a LondonTargetingGasLimitCalculator
 4) Define the valid TargetingGasLimitCalculator depends on the milestone on the protocolSpec and not on the BesuControllerBuilder (this class already contains the protocolSpec)

**Example of code to change :** 
```java
 if (eip1559.isPresent() && eip1559.get().isForkBlock(header.getNumber())) {
    parentGasLimit = parent.getGasLimit() * eip1559.get().getFeeMarket().getSlackCoefficient();
  }
```

garyschulte:
## PR description

this is an extensive re-plumbing to make GasLimitCalculator part of ProtocolSchedule rather than being global

highlights:
* plumbing of `Optional<AtomicLong>` targetGasLimit into MiningParameters, MiningCoordinators, MiningExecutors, and Block Creators
* moving GasLimitCalculator interface and related implementations into ethereum:core
* creation of LondonTargetingGasLimitCalculator
* removal of 1559 specific forkblock gas limit logic in GasLimitRangeAndDeltaValidationRule and AbstractBlockCreator

## Fixed Issue(s)

fixes #2561 

## Changelog


shemnon:
Can the cleanup be moved to a different changeset?  Specifically Dropping AluetBlockNumber and EIP1559BlockNumber are fairly orthagonal and non-controversial.

garyschulte:
> Can the cleanup be moved to a different changeset? Specifically Dropping AluetBlockNumber and EIP1559BlockNumber are fairly orthagonal and non-controversial.

I can move those into a different PR pretty easily.  Feel free to ignore those for now and I will split those out on a subsequent commit

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/CliqueProtocolSchedule.java
```diff
@@ -56,9 +62,24 @@ public static ProtocolSchedule create(
 
     final EpochManager epochManager = new EpochManager(cliqueConfig.getEpochLength());
 
+    // TODO: roll EIP1559 object into fee market and remove these configs
+    final BaseFeeMarket londonFeeMarket =
+        FeeMarket.london(
+            targetGasLimit.stream()
```
shemnon, line 68: This doesn't feel right, or I'm reading it wrong.  This is taking what was a function without an input (how to calculate the gas limit) and replacing it with something that has a fixed input derived from a CLI flag.  

How do we update this value at runtime via RPC? 

Making operators restart their nodes to change the target is a loss of functionality.

garyschulte: the `miner_changeTargetGasLimit` endpoint should continue to function as normal, with the exception being that it needs to update the targetGasLimit for all of the GasLimitCalculators, since there isn't the notion of 'change the targetGasLimit as of this block number'.  

Previously there was only one global gas limit calculator to update, and it was plumbed via BesuCommand.  Now however there can be multiple calculators, each of which applies to their respective protocolSpecs.

shemnon: The OptionalLong is an immutable value however, it is not AtomicLong which can be updated.  Once set there is no changing the GasLimitCalculator, the RPCs would need to replace it within the schedule.

What should change per schedule is the function to interpret the configured flag, not a one-time evaluation of the flag with per-schedule meaning.

garyschulte: It is probably difficult to follow the change, but the actual targetGasLimit is non an OptionalLong, it is `private long`:

https://github.com/hyperledger/besu/pull/2632/files#diff-bf6252888b0185ced31a58f4ca5755847f8f913680d5fd23155bf364bc054d34R28

OptionalLong is just wrapping targetGasLimit because it is an optional parameter.  If there is not a targetGasLimit present, a constant gas limit calculator is used rather than a targeting gas limit calculator.

garyschulte: I can change the mutable long to AtomicLong.  the targetGasLimit update within the protocol schedule is happening here:
https://github.com/hyperledger/besu/pull/2632/files#diff-a498e6c44084bb5267fd823cb8591538e6b7f99dbb21bf0fc55438ae5bb4c03bR38-R48

shemnon: Merging the CLI setting of the gas limit and the schedule interpretation of that limit are not two things I would expect to be in the same class.  

I think we would get better utility by changing org.hyperledger.besu.ethereum.blockcreation.GasLimitCalculator#nextGasLimit to take two args, the current block limit and the CLI configure limit (if configured), so `long` and `OptionalLong`.  That way instead of diving into the schedules (which IMHO should be 100% immutable once set) we set the value in the AbstractBlockCreator, or the AbstractBlockCreator has an object that provides that value (a shared `AtomicLong`, a `Provider<Long>`, a custom class, etc.  

It fels like the value should be in `MiningParametes` since changing the Coinbase has to follow similar plumbing?  Then the CLI configuraiton gets passed into the various MinerExecutors the same way Coinbase gets passed around.  Another option would be the various `MiningCoordinators`

garyschulte: great call on MiningParameters.  That makes perfect sense to fold targetGasLimit into there, and keep gas limit calculators final/immutable

shemnon (dismissed):

shemnon:
LGTM assuming unit tests are just minor cleanup and nothing dramatic.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/LondonTargetingGasLimitCalculator.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class LondonTargetingGasLimitCalculator extends FrontierTargetingGasLimitCalculator {
+  private static final Logger LOG = LogManager.getLogger();
+  private final long londonForkBlock;
+
+  public LondonTargetingGasLimitCalculator(final long londonForkBlock) {
+    this(1024L, 5000L, Long.MAX_VALUE, londonForkBlock);
```
matkt, line 27: why not using DEFAULT_MIN_GAS_LIMIT and DEFAULT_MAX_CONSTANT_ADMUSTMENT_INCREMENT here ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/LondonTargetingGasLimitCalculator.java
```diff
@@ -0,0 +1,61 @@
...
+    if (targetGasLimit > currentGasLimit) {
+      nextGasLimit = Math.min(targetGasLimit, safeAddAtMost(currentGasLimit));
+    } else if (targetGasLimit < currentGasLimit) {
+      nextGasLimit = Math.max(targetGasLimit, safeSubAtMost(currentGasLimit));
+    } else {
+      nextGasLimit = currentGasLimit;
+    }
+
+    if (londonForkBlock == newBlockNumber) {
+      nextGasLimit = nextGasLimit * LondonFeeMarket.DEFAULT_SLACK_COEFFICIENT;
```
matkt, line 52: I wonder if it would be better if we pass the correct FeeMarket when defining the protocol schedule (passing feemarket in the constructor https://github.com/hyperledger/besu/pull/2632/files#diff-22b78733e37a697fa8d1d8a02d2a87fe5ccea9cf67c34ce5e6311f024c14abd6R480). Because if the DEFAULT_SLACK_COEFFICIENT may change in a future fork that would force us to duplicate that LondonTargetingGasLimitCalculator class just to change that call `LondonFeeMarket.DEFAULT_SLACK_COEFFICIENT` to ` NewForkFeeMarket.DEFAULT_SLACK_COEFFICIENT`. If we pass the fee market  in the constructor the problem is fixed

matkt (changes requested):
some comments and questions

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/LondonTargetingGasLimitCalculator.java
garyschulte: Ah, I added the constants after I added the London gas calculator.  :+1:

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/LondonTargetingGasLimitCalculator.java
garyschulte: Sure I think that is reasonable

matkt (approved):
LGTM

PR #2633, opened 2021-08-13
Fixed nonce bug in EthGetTransactionCount

rojotek:
## PR description

Fixed bug where EthGetTransactionCount would return a lower value for pending than latest when there are old transactions in the transaction pool.

Wrote a failing test to show the expected behaviour and ensured that the max of latest and pending is used.

## Changelog


atoulme (approved):
lgtm

vmichalik:
The elusive @rojotek contribution! Welcome back ;-)

macfarla (approved):
LGTM

PR #2634, opened 2021-08-13
Jumpdest analysis

### LINKED ISSUE #2607: jump opcode analysis
### Description
Every time an executable account (contract) is run, a brand new Code object is instantiated from the bytecode in the account. Upon encountering the first JUMP instruction, all valid jump locations are computed and stored in a bitmap on the Code object. Unfortunately, that Code object is not retained for future use, so this analysis must be repeated on the first JUMP instruction of every contract, every time it is run.

### Acceptance Criteria

1. Reuse of jump destinations on all subsequent contract executions. 
2. Runtime speed improvements proportional to the number of jump dests in the contract. Pathological EVMtool test case [attached](https://gist.github.com/jflo/6f46fd44fd706c3f26deaacadf5b1a45), should speed up considerably.
3. Cache entries must be evicted on Account deletion.
4. Cache entries must be evicted on contract self-destruct.
5. Entries should expire out of cache in LRU order when total cache weight is hit.

jflo:
## PR description

Adds ability to cache Code objects and use them for future calls which will avoid the re-computing of JUMP destinations. 
No additonal tests have been added to demonstrate performance improvement. The correct place for those remains an open question.

## Fixed Issue(s)

fixes #2607 

## Changelog


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 21.7.4
+### Additons and Improvements
```
atoulme, line 4: typo

ethereum/core/build.gradle
```diff
@@ -38,6 +38,7 @@ sourceSets {
 }
 
 dependencies {
+  implementation 'info.picocli:picocli:4.6.1'
```
atoulme, line 41: Can we avoid that somehow?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/CodeScale.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.hyperledger.besu.ethereum.core.contract;
+
+import org.hyperledger.besu.ethereum.core.Account;
+import org.hyperledger.besu.ethereum.vm.Code;
+
+import com.google.common.cache.Weigher;
+
+public class CodeScale implements Weigher<Account, Code> {
```
atoulme, line 23: Does it need to be public?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheOptions.java
```diff
@@ -0,0 +1,36 @@
...
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.hyperledger.besu.ethereum.core.contract;
+
+import picocli.CommandLine;
+
+public class ContractCacheOptions {
```
atoulme, line 20: Would move under unstable

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheOptions.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public static final String CONTRACT_CACHE_WEIGHT = "--contract-code-cache-weight-kb";
```
atoulme, line 22: Should be a `--X` unstable CLI option for now

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheOptions.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public static Long getContractCacheWeightKilobytes() {
+    return contractCacheWeightKilobytes;
+  }
+
+  @CommandLine.Option(
+      names = {CONTRACT_CACHE_WEIGHT},
+      description =
+          "size in kilobytes to allow the cache to grow to before evicting the least recently used",
```
atoulme, line 31: That description doesn't tell me what the cache does.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
```diff
@@ -199,7 +199,7 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
               .sender(sender(frame))
               .value(value(frame))
               .apparentValue(apparentValue(frame))
-              .code(new Code(contract != null ? contract.getCode() : Bytes.EMPTY))
+              .code(frame.getWorldState().getContract(contract).orElse(new Code()))
```
atoulme, line 202: So what happens if I try to get code on a EOA?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessorTest.java
```diff
@@ -92,6 +137,95 @@ public void shouldCallTransactionValidatorWithExpectedTransactionValidationParam
...
+
+    Account contractAccount = this.worldState.get(contractAddr);
+    Mockito.verify(this.worldState, times(1)).getContract(contractAccount);
+    Mockito.verify(loader, times(1)).load(contractAccount);
+
+    this.worldState.deleteAccount(contractAccount.getAddress());
+    this.worldState.commit();
+    assertThat(this.worldState.get(contractAccount.getAddress())).isNull();
+    assertThat(this.cache.size()).isEqualTo(0);
+  }
```
atoulme, line 227: Is there a test case to cover CREATE2?

atoulme:
What is the performance improvement observed? Is there a possibility to also persist the cache across restarts? Is there a way to add meters to see cache hits and misses?

ethereum/core/build.gradle
jflo: Sadly, no. This was necessary since putting the configuration classes in besu/main would cause a circular dependency between that and ethereum/core. I think this is important for moving towards a package-by-feature approach, which we will need more and more of in the future as we divide up the codebase..

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/CodeScale.java
jflo: No it does not, package-by-feature ftw!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
jflo: you get a new Code() with nothing in it. I suppose we could also put Bytes.EMPTY in the new Code() for consistency with previous, if that's what you're getting at.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessorTest.java
jflo: [Yes there is](https://github.com/hyperledger/besu/blob/367b6a74f1da5a88c9a6b9c84565f0f6f6ffd406/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/Create2OperationTest.java#L47), but since jumpdest analysis is done on execution, not creation, it was not impacted.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheOptions.java
jflo: See previous note about circular dependencies and package-by-feature, unless you are proposing an unstable packge being added here?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
atoulme: From your code changes it looks like you’d get an IllegalArgumentException.

ethereum/core/build.gradle
atoulme: Please expand a lot more - we’re dividing up the code base?

shemnon: When did we go to a package by feature approach?  Not that I'm against it, it's just that it seems like the sort of thing that should be discussed among more than just ConsenSys maintainer.

That being said, the current dogma would be to put this feature in a stand alone config object (or add it to an existing config object) and have the BesuCommand handle marshaling the features in and out.  Wiring in the PicoCLI flags binds the library code to a CLI style and makes it less suitable to library inclusion. It's mixing concerns and strongly tying us to one CLI library.

shemnon: At a high level, my thinking is that CLI configuration should be fully severable from the operation and packaging of the code.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/CodeCache.java
```diff
@@ -0,0 +1,161 @@
...
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheStats;
+import com.google.common.cache.LoadingCache;
+import com.google.common.collect.ImmutableMap;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class CodeCache implements LoadingCache<Account, Code> {
+
+  private final LoadingCache<Account, Code> cache;
```
shemnon, line 35: I would key by `Hash` instead of account.  Many accounts share code (such as most ERC-721s and ERC-20s) and this would reduce cache misses.  

Except for a convenience method (thunking account to account.getCodeHash) I would not use `Account` in this class, it's not a very reliable key.

shemnon: Switching to key by hash also opens up persisting the cache off-memory.

ethereum/core/build.gradle
jflo: Yes, that is my thinking too @shemnon , and seems to be in the spirit of the PicoCLI design. It naturally allows you to avoid these circular dependencies. I tried to implement this via a standalone config object and ended up writing more code than was valuable, to move around a single config option. Putting that config class in between the option and the CodeCache was what caused the circular dependency.

@atoulme I was referring to the last contributors call where we discussed separating out the EVM into it's own library. Between that and other discussions around providing MEV functionalities, it seems like we are heading in the direction of trying to modularize Besu. 

If there exists any documentation on why Besu is divided into the gradle projects and java packages that it is, I am more than happy to study it.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
jflo: sorry, I don't see it. which part would throw that?

garyschulte: perhaps a static Code.EMPTY constant would make the behavior more clear instead of a zero arg constructor.

ethereum/core/build.gradle
atoulme: @jflo there is a leap between saying we want to carve out the EVM so we can make it a library, and saying we're modularizing Besu, or applying what you call a package by feature approach.
Please do document your approach and submit it to contributors so we can all review.
* What are the modules?
* How is configuration and dependency injection working?
* What is the path to executables? - for example, you add picocli as a dependency right in here, but frankly, that should be decoupled away into a separate config module so you can have an ethereum lib.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/AbstractCallOperation.java
atoulme: I was referring to `getContract` hitting the cache, and I erroneously thought it would the IllegalArgumentException thrown in the CacheLoader class.

ethereum/core/build.gradle
jflo: A great deal of Besu is already packaged by feature. For instance, looking at the besu/ethereum/eth module, in the org.hyperledger.besu.ethereum.eth.sync package, I see many different classes with different functions, all serving the synchronization feature. I'm not sure why this seems like a leap? Is it the lack of an intermediary Configuration class between the Option and the class using it?

jflo: > Wiring in the PicoCLI flags binds the library code to a CLI style and makes it less suitable to library inclusion. It's mixing concerns and strongly tying us to one CLI library.

@shemnon can you elaborate on this? The flags are defined in the Options class, just like everywhere else, so a change in the CLI library still only requires updates to a single class. I think the part I'm really struggling with is the duplication of all the fields across the Options and Configuration class. I'm unsure why that is better and what it is buying me for this case.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/CodeCache.java
jflo: Added a wrapper class to intercept hashCode() from guava. Was surprised to see there was no way to add a custom comparator to the cache via its builder.

ethereum/core/build.gradle
shemnon: The options class should be in the besu module... `besu/src/main/java/org/hyperledger/besu/cli/options/unstable/`, which should have a `toDomainObject` method exporting the CLI value to a domain object that live in the core module.  Lie what `DnsOptions` does (a fairly simple example).

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheOptions.java
```diff
@@ -0,0 +1,42 @@
...
+  @CommandLine.Option(
+      names = {CONTRACT_CACHE_WEIGHT},
+      description =
+          "size in kilobytes to allow the cached"
+              + "contract bytecode to grow to before evicting the least recently used contract",
+      fallbackValue = "250000",
+      defaultValue = "250000",
+      hidden = true,
+      arity = "1")
+  public static Long contractCacheWeightKilobytes = 250000l;
```
shemnon, line 41: Don't use static

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/CodeCache.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.Optional;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.LoadingCache;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class CodeCache {
+
+  private final LoadingCache<WrappedAccount, Code> cache;
```
shemnon, line 29: In general this WrappedAccount scheme won't work.  I would instead suggest `CodeAndHash` where the Hash and Code object are stored as fields, the `equals` method first compares the hash and the code, and the `hashCode` method only returns the `hashCode` of the `Hash`.  Use that as the key instead of any other Account object.  There can be a static factory or constructor too that takes the `Account` object.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1588,7 +1591,7 @@ private void configure() throws Exception {
                     : SyncMode.FULL);
 
     ethNetworkConfig = updateNetworkConfig(getNetwork());
-
+    unstableContractCacheOptions.initConfig();
```
jflo, line 1594: I struggled a lot with were to put this. This is the moment where the value in the Option becomes available to the Configuration.

ethereum/core/build.gradle
jflo: Ok, i did this, but only up to a point. All of the other Configurations (such as `DnsOptions`) have a much shorter distance down into the codebase to travel than this one does. Most are used in the controller or service layers, while the contract code cache is used from multiple implementations of WorldState or WorldView. Doing this the exact same way would add the Configuration class to the signatures of tons of code. (I went down this path and have a branch somewhere if you would like to see the horrors). I thinks this singleton approach is a less bad compromise that 

1. keeps the Option as the single place that knows about PicoCli
2. avoids making dozens of other classes aware of a Configuration they don't need, and only pass down to lower layers.

ethereum/core/build.gradle
```diff
@@ -51,6 +52,7 @@ dependencies {
 
   implementation 'com.fasterxml.jackson.core:jackson-databind'
   implementation 'com.google.guava:guava'
+  implementation 'com.github.ben-manes.caffeine:caffeine:3.0.4'
```
jflo, line 54: Some of the [limitations in how weighted caches](https://groups.google.com/g/guava-discuss/c/P4f13-SH1Rc) are handled made this very hard to test. Upgrade was easy, and interfaces are very similar. Suggest using it elsewhere also.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+    this.contractCacheWeightKilobytes = contractCacheWeightKilobytes;
+  }
+
+  public static void init(final long contractCacheWeightKilobytes) {
+    if (INSTANCE == null) {
+      INSTANCE = new ContractCacheConfiguration(contractCacheWeightKilobytes);
+    }
+  }
+
+  public static void destroy() {
```
jflo, line 38: This was only needed for testing, shouldn't be useful at runtime once the cache is created.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.hyperledger.besu.ethereum.core.contract;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class ContractCacheConfiguration {
```
jflo, line 21: Could alternately implement this from an Enum

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessorTest.java
```diff
@@ -43,20 +64,46 @@
...
-
+  private MainnetMessageCallProcessor messageCallProcessor;
+  private CodeLoader loader;
+  private CodeCache cache;
   @Mock private Blockchain blockchain;
   @Mock private WorldUpdater worldState;
   @Mock private ProcessableBlockHeader blockHeader;
   @Mock private Transaction transaction;
   @Mock private BlockHashLookup blockHashLookup;
+  private final String manyJumps =
```
jflo, line 80: Need to load this from a file.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+  public static final ContractCacheConfiguration DEFAULT_CONFIG =
+      new ContractCacheConfiguration(250_000L);
+  private final long contractCacheWeightKilobytes;
+  private static ContractCacheConfiguration INSTANCE;
```
shemnon, line 25: Please, no static configuration.  Use the dependency injection pattern.

jflo: Can you elaborate? It sounds to me like you're saying not to use this Singleton at all, but I didn't see any comments on [my explanation as to why it was used](https://github.com/hyperledger/besu/pull/2634#discussion_r710321931) above.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -242,7 +243,7 @@ public Account get(final Address address) {
   }
 
   @Override
-  public Optional<Code> getContract(final Account account) {
+  public <T extends AccountState> Optional<Code> getContract(final T account) {
```
shemnon, line 246: Why the generic?  Account already implements AccountState.  You can do the getCodeHash at line 44 in CodeCahce without changing any types.

jflo: brainbroke, up too late last night. fixed.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/ContractCacheConfiguration.java
shemnon: Singletons are bad for unit testing.

jflo:
> What is the performance improvement observed? Is there a possibility to also persist the cache across restarts? Is there a way to add meters to see cache hits and misses?

The test provided was a pathological case run against London, Berlin, and Istanbul protocol spec. Runtime was reduced for the specific case tested.  Caution, napkin math:

London   14%
Berlin      21%
Istanbul    5%

So nothing to write home about, but it's an improvement. Sure we could measure cache hits and misses. I think in hindsight with what I learned doing this, I would just add the bitmap of the jumpdestinations to AccountState the first time it is run, and store it.

atoulme:
21% is pretty cool. I guess you’re talking about execution time.

jflo:
Yeah, thats on a pretty pathological case so I don't expect it to be the same improvement in regular mainnet usage. However, it was interesting to me that on a daily basis, there are [less than 10k](https://dune.xyz/queries/131376) unique contracts being used on mainnet. Given tx volume, that should be a lot of redundant jumpdest analysis that is eliminated.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/contract/CodeHash.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.hyperledger.besu.ethereum.core.contract;
+
+import org.hyperledger.besu.datatypes.Hash;
+
+import org.apache.tuweni.bytes.Bytes;
+
+/** key for use by the CodeCache */
+public class CodeHash {
```
matkt, line 23: Maybe you can use the Immutable  library fot that  https://immutables.github.io/

matkt (changes requested):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -42,6 +45,8 @@
   protected final long height;
   protected final TrieLogLayer trieLog;
   private final Hash worldStateRootHash;
+  private final CodeCache codeCache =
+      new CodeCache(ContractCacheConfiguration.getInstance().getContractCacheWeight());
```
garyschulte, line 49: I think this will be a problem.  A new layered worldstate gets created on every call to `BonsaiWorldStateArchive.persist`. I think we are going to end up with a lot of heap pressure from new codecache instances.  A singleton codecache should be safe and mitigate this.  @matkt, thoughts ?

garyschulte (changes requested):
As discussed on chat, a singleton codecache might simplify config plumbing.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
matkt: the number of layers in memory is limited to 512. So at most we will have 512 in memory. I don't know if the CodeCache is a large object. But maybe for this object we should create it only when we really use it. In many cases I think that the cache will be created in a layer but we do not need it or I miss something

jflo: this is out of date, it has been replaced with a singleton for the cache.

PR #2635, opened 2021-08-14
Refactoring CmsValidator (internal CRL resolution)

lucassaldanha:
## PR description

Loading CRL list inside CmsValidator (instead of externally)

## Changelog

### DIFF
--- a/pki/src/main/java/org/hyperledger/besu/pki/cms/CmsValidator.java
+++ b/pki/src/main/java/org/hyperledger/besu/pki/cms/CmsValidator.java
@@ -145,7 +145,7 @@ public class CmsValidator {
           new PKIXBuilderParameters(truststore.getKeyStore(), targetConstraints);
 
       // Adding CertStore with CRLs (if present, otherwise disabling revocation check)
-      loadCRLs(truststore)
+      createCRLCertStore(truststore)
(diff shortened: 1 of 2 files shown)


pki/src/main/java/org/hyperledger/besu/pki/cms/CmsValidator.java
```diff
@@ -178,4 +178,18 @@ private boolean isCertificateTrusted(
       throw new RuntimeException("Error validating certificate chain", e);
     }
   }
+
+  private Optional<CertStore> loadCRLs(final KeyStoreWrapper truststore) {
```
jframe, line 182: This is really loading CRL as such, maybe createCertStore?

jframe, line 182: Noticed there is CRLUtil that does something similar to this function, should that be used instead or this replaces that one?

lucassaldanha: CRLUtil is used when loading the CRL pem file. Not for the truststore. They are similar but not the same.

jframe (approved):
LGTM

PR #2645, opened 2021-08-17
Refactor QbftExtraDataCodec from new instance to injection

lucassaldanha:
## PR description
- Updated `RoundChange` and `ProposalPayload` `decode()` methods to accept a `BftExtraDataCodec` implementation (instead of creating a new instance of `QbftExtraDataCodec`.
- A bit of plumbing work all the way to `QbftBesuControllerBuilder`, where we choose the concrete codec instance to use.
- For the tests, we are using the `QbftExtraDataCodec` (to keep the previous behaviour)

## Changelog

### DIFF
--- a/consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/messagewrappers/ProposalTest.java
+++ b/consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/messagewrappers/ProposalTest.java
@@ -17,6 +17,7 @@ package org.hyperledger.besu.consensus.qbft.messagewrappers;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import org.hyperledger.besu.consensus.common.bft.BftExtraData;
+import org.hyperledger.besu.consensus.common.bft.BftExtraDataCodec;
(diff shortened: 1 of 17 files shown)


consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/messagewrappers/ProposalTest.java
```diff
@@ -41,16 +42,15 @@
 import org.junit.Test;
 
 public class ProposalTest {
+  private static final BftExtraDataCodec bftExtraDataCodec = new QbftExtraDataCodec();
```
jframe, line 45: Do we need a round trip test with CMS encoding? Might not be of much benefit as really just testing extraDataCodec?

lucassaldanha: If we want a round trip test for the Proposal, it probably makes more sense to do it in the follow-up PR will all the PKI extraData CMS changes (the new format, etc.).

jframe (approved):
LGTM

PR #2646, opened 2021-08-18
Add a labelled gauge to metrics

rolfyone:
The labelled gauge functionality will allow monitoring of gauge type data without creating many gauges, which can be particularly handy in tracking related data.

This would be particularly useful in a specific use case in Teku which uses the besu infrastructure to provide prometheus metrics.
https://github.com/ConsenSys/teku/issues/3892

There are a large number of gauges that are requested, but logically they're actually a single gauge with labels capturing context.

## Changelog


metrics/core/src/main/java/org/hyperledger/besu/metrics/opentelemetry/OpenTelemetryGauge.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.function.DoubleSupplier;
+import java.util.function.Supplier;
+
+import com.google.common.base.Preconditions;
+import io.opentelemetry.api.metrics.AsynchronousInstrument;
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.api.metrics.common.LabelsBuilder;
+
+public class OpenTelemetryGauge implements LabelledGauge {
```
ajsutton, line 31: nit: Should this be `OpenTelemetryLabelledGauge`?

metrics/core/src/main/java/org/hyperledger/besu/metrics/opentelemetry/OpenTelemetryGauge.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public void labels(final DoubleSupplier valueSupplier, final String... labelValues) {
+    Preconditions.checkArgument(
+        labelValues.length == labelNames.size(),
+        "label values and label names need the same number of elements");
+    final Labels labels = getLabels(labelValues);
+    if (!observationsMap.containsKey(labels)) {
+      observationsMap.put(labels, valueSupplier);
```
ajsutton, line 57: This has a race condition if some other thread is adding at the same time.
```suggestion
    if (observationsMap.putIfAbsent(labels, valueSupplier != null)) {
      throw new IllegalStateException("Already registered a gauge with labels " + Arrays.toString(labelValues));
```

metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusGauge.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.function.DoubleSupplier;
+
+import io.prometheus.client.Collector;
+
+public class PrometheusGauge implements LabelledGauge {
+  private final String metricName;
+  private final String help;
+  private final List<String> labelNames;
+  private final Consumer<Collector> collectorConsumer;
+  private final List<String> labelValuesCreated = new ArrayList<>();
```
ajsutton, line 31: This won't be thread safe, probably want to use a `ConcurrentHashSet`.  As with the open telemetry version need to try adding and throw if it already exists rather than doing the contains and add steps separately.

metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusMetricsSystem.java
```diff
@@ -141,6 +143,27 @@ public void createGauge(
...
+    final String metricName = convertToPrometheusName(category, name);
+    if (isCategoryEnabled(category)) {
+      final Consumer<Collector> addCollectorLambda =
+          (collector ->
+              collectors
+                  .computeIfAbsent(
+                      category, key -> Collections.newSetFromMap(new ConcurrentHashMap<>()))
+                  .add(collector));
+      final PrometheusGauge gauge =
+          new PrometheusGauge(metricName, help, List.of(labelNames), addCollectorLambda);
```
ajsutton, line 161: Shouldn't this just be calling `addCollectorUnchecked`?
```suggestion
      final PrometheusGauge gauge =
          new PrometheusGauge(metricName, help, List.of(labelNames), collector -> addCollectorUnchecked(category, collector));
```

atoulme:
Sounds ok to me as an optimization. I believe I'm due to update the metrics lib of OpenTelemetry, as 1.5.0 was apparently breaking a lot of APIs. It's also possible that we could now use OpenTelemetry API for all our needs going forward, since it is able to interface with Prometheus to export metrics.

metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusGauge.java
```diff
@@ -0,0 +1,62 @@
...
+      throw new IllegalArgumentException(
+          String.format("A gauge has already been created for label values %s", labelValuesString));
+    }
+  }
+
+  @Override
+  public List<MetricFamilySamples> collect() {
+    final List<MetricFamilySamples.Sample> samples = new ArrayList<>();
+    observationsMap.forEach(
+        (k, v) ->
```
ajsutton, line 57: nit: Would be good to use more meaningful variable names here, eg `(labels, valueSupplier)`.

metrics/core/src/main/java/org/hyperledger/besu/metrics/opentelemetry/OpenTelemetryGauge.java
```diff
@@ -0,0 +1,72 @@
...
+  private Labels getLabels(final String... labelValues) {
+    final LabelsBuilder labelsBuilder = Labels.builder();
+    for (int i = 0; i < labelNames.size(); i++) {
+      labelsBuilder.put(labelNames.get(i), labelValues[i]);
+    }
+    return labelsBuilder.build();
+  }
+
+  private void updater(final AsynchronousInstrument.DoubleResult doubleResult) {
+    observationsMap.forEach((k, v) -> doubleResult.observe(v.getAsDouble(), k));
```
ajsutton, line 70: nit: Would be good to use more meaningful variable names here, eg `(labelValues, valueSupplier)`.

ajsutton (approved):
LGTM.

PR #2657, opened 2021-08-18
Remove EIP-1702 account versioning

shemnon:
EIP-1702 versioning was a candidate for the Istanbul hard fork but was
removed prior to the first testnet. Other versioning techniques have
greater core dev mind share and one preparatory step landed in London,
EIP-3541, making EIP-1702 very unlikely to make it to mainnet.

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Support for mining ommers [#2576](https://github.com/hyperledger/besu/pull/2576)
 - Updated onchain permissioning to validate permissions on transaction submission [\#2595](https://github.com/hyperledger/besu/pull/2595)
 - Removed deprecated CLI option `--privacy-precompiled-address` [#2605](https://github.com/hyperledger/besu/pull/2605)
+- Removed code supporting EIP-1702. [#2657](https://github.com/hyperledger/besu/pull/2657)
 
 ### Bug Fixes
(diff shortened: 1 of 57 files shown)


lucassaldanha:
Changes look fine. Are we sure this change won't be backwards-incompatible in any way?

shemnon:
All "version 0" accounts don't persist a version number, only non-zero versions do, so as long as no versions are set to 1 it is backwards compatible. 

There are two and a half places where the account version is set to anything that is not a copy: The `MainnetContractCreationProcessor` and the `GenesisState` generator (the `PrivateStateGenesisAllocator` being the half as it's just the private form of genesis state).

To set a contract creation version to a non-zero value the contract creator would need to be set in java code.  That would be in the protocol spec code.  And none of that is configurable externally, you only get to pick from existing forks.  And none fo the forks set a version.  (pre-release Istanbul did, but that was changed fairly quick).

The version _could_ be set in a genesis definition, but that has never been documented. And would grant no value as nothing is set internally to use account versioning.

There is a lingering performance impact of this in the EVM.  Every opcode lookup as to go through _two_ array indirections.  One for version, one for opcode, before getting the operation.  Same with precompiles, one for the version and one for the precompile address. So eliminating this vestigial code will have positive performance implications.

atoulme (approved):
Changes look ok.

macfarla (approved):
LGTM

PR #2658, opened 2021-08-18
Feature/1559 cleanup remove 1559 object

### LINKED ISSUE #2562: Remove EIP1559 object
We need to remove this class. : 

- the method **isEIP1559** is useless we just need to check if we are in london thanks to the protocol MainnetProtocolSpecs
- the methods **isForkBlock/getForkBlock** need to be moved to the ProtocolShedule method (like that we will have something generic to retrieve the block number of a fork). To do that we will create a **getMilestoneBlock** method that return the block number of a fork name  : 

```java        
public interface ProtocolSchedule {

          ProtocolSpec getByBlockNumber(long number);

          public Stream<Long> streamMilestoneBlocks();

          **+++++ public Long getMilestoneBlock(final String forkName);  +++++**

          Optional<BigInteger> getChainId();

          void setTransactionFilter(TransactionFilter transactionFilter);

          void setPublicWorldStateArchiveForPrivacyBlockProcessor(
            WorldStateArchive publicWorldStateArchive);
        }

```

- remove **targetGasUsed** method . this method is always used with **computeBaseFee**. So maybe we need to merge this two methods

- **computeBaseFee** method move to BaseFee object for more consistency as a builder. This method will return a BaseFee object directly and not a long

```java
public class BaseFee {
  private final long value;
  private final long delta;
  private final long minNextValue;
  private final long maxNextValue;

  public BaseFee(final FeeMarket feeMarket, final long value) {
  this.value = value;
  this.delta = Math.floorDiv(value, feeMarket.getBasefeeMaxChangeDenominator());
  this.minNextValue = value - delta;
  this.maxNextValue = value + delta;
  }

  ....

  ++++ public static BaseFee computeBaseFee(
      ++++ final FeeMarket feeMarket
      final long blockNumber,
      final long parentBaseFee,
      final long parentBlockGasUsed,
      final long targetGasUsed) {
       ...
      return new new BaseFee(feeMarket, baseFee);
  }
}
```

garyschulte:
## PR description
* absorb EIP1559 objects into BaseFeeMarket
* remove explicit EIP1559 plumbing and rely on ProtocolSchedule where possible

## Fixed Issue(s)

fixes #2562

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -101,19 +103,19 @@ public JsonRpcResponse response(final JsonRpcRequestContext request) {
             .map(blockHeader -> blockHeader.getBaseFee().orElse(0L))
             .orElseGet(
                 () ->
-                    protocolSchedule
-                        .getByBlockNumber(nextBlockNumber)
-                        .getEip1559()
+                    Optional.of(protocolSchedule.getByBlockNumber(nextBlockNumber).getFeeMarket())
```
shemnon, line 106: Not a fan of this lambda style.  `Optional.of.filter.map.else` is really just `var x; if (x.y) ... else ...` and I think the latter reads better.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -265,17 +266,19 @@ private ProcessableBlockHeader createPendingBlockHeader(final long timestamp) {
...
-      // TODO roll eip1559 into feeMarket
-      final EIP1559 eip1559 = protocolSpec.getEip1559().orElseThrow();
-      baseFee =
-          eip1559.computeBaseFee(
-              newBlockNumber,
-              parentHeader.getBaseFee().orElse(0L),
-              parentHeader.getGasUsed(),
-              eip1559.targetGasUsed(parentHeader));
-    }
+    final Long baseFee =
```
shemnon, line 269: again... this reads better as 

```
var feeMarket = protocolSpec.getFeeMarket();
Long baseFee = feeMarket.implementsBaseFee()
      ? ((BaseFeeMarket)feeMarket).computeBaseFee(
          newBlockNumber,
          parentHeader.getBaseFee().orElse(0L),
          parentHeader.getGasUsed(),
          feeMarket.targetGasUsed(parentHeader))
      : null;
```

shemnon (approved):
LGTM.  Lamba complaint is purely optional.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
garyschulte: I can appreciate easier reading, but in these two cases I am doing it this way to avoid using explicit casts.  I have linter ptsd from prior project ;)

shemnon: Perhaps we could move `computeBaseFee(
      final long blockNumber,
      final long parentBaseFee,
      final long parentBlockGasUsed,
      final long targetGasUsed)` into the `FeeMarket` interface with a default implementation of `return 0L;`?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpec.java
```diff
@@ -72,8 +71,6 @@
 
   private final FeeMarket feeMarket;
```
matkt, line 72: just a little bit confused with FeeMarket and BaseFeeMarket. there is the use of FeeMarket here but elsewhere it is used BaseFeeMarket instead. Why not always used the interface ?

garyschulte: As part of the 1559 cleanup, `FeeMarket` is now a non-optional part of ProtocolSpec.  It includes TransactionPriceCalculator and minTransactionPriceInNextBlock, and is necessary for all of the forks.  There is a pre-1559 FeeMarket that is used for all ProtocolSpecs pre-London.  

`BaseFeeMarket` is a 1559-specific extension of FeeMarket and is used in all of the London-specific classes that need BaseFee constants and methods.

PR #2664, opened 2021-08-20
Remove duplicate AddOperation on MainnetEvmRegistries

matthewrkula:
## PR description
While familiarizing myself with the EVM code, I noticed that there is a duplicate `AddOperation` being added to the `MainnetEvmRegistries`. The `MainnetEvmRegistries#put()` function just assigns the operation to a map using the operation's opcode so removing this duplicate line has no effect on the EVM itself as it was just overwriting the previous line.

## Changelog


atoulme (approved):
LGTM thanks for the contrib

PR #2673, opened 2021-08-25
Update memory to use MutableBytes

matkt:

## PR description

This modification improves performance because we no longer have to copy the memory to a new array every time we want to read it.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LondonFeeMarket.java
```diff
@@ -116,4 +116,9 @@ public long computeBaseFee(
   public boolean isForkBlock(final long blockNumber) {
     return londonForkBlockNumber == blockNumber;
   }
+
```
matkt, line 119: this code will be removed when the fix will be on master (for testing purpose)

garyschulte (approved):
🚢

PR #2677, opened 2021-08-26
Adding QBFT PKI-backed Block Creation

### LINKED ISSUE #2653: Refactor CMS validation logic on ProposalPayloadValidator
### Description
At the moment, we are validating the CMS message in the `ProposalPayloadValidator`. However, it might be a good idea to have this logic decoupled from the "main" `ProposalPayloadValidator`. Either using inheritance (e.g. `PkiProposalPayloadValidator`) or composition (e.g. creating the concept of proposal payload validation rules).

For QbftRound, we will have a strategy for creating the proposal.

lucassaldanha:
## PR description
- Changed `cms` field in `PkiExtraData` from `Optional<Bytes>` to `Bytes`
- Updated PkiExtraData encoding/decoding. Now we are appending the CMS as the last element in the ExtraData list.
- Created `QbftContext` containing the optional `PkiBlockCreationConfiguration`. Used to determine if the node is running in PKI mode and to have access to the KeyStores.
- Created a new hashing strategy that excludes the CMS data from the hash calculation. This is used when calculating the signed hash inside the CMS message.
- Updated `QbftRound` with a new `createBlockForProposalBehaviour`. The default implementation just creates the block. The "PKI" implementation created the block and adds the CMS into the extra data.
- Updated `ProposalPayloadValidator` with logic to validate the CMS message in the proposal (when running in PKI-mode).
- Included `pki-qbft` parameter for PKI ATs. This will run tests using PKI QBFT.

## Pending Changes
- [x] Stop leaking PKI information to `BftExtraDataCodec` (we are leaking it through the `EXCLUDE_CMS` enum value)
- [x] Remove duplication on `PkiQbftExtraDataCodec#encode(..)` method (we are duplicating code from `QbftExtraDataCodec#encode`)

## Fixed Issue(s)
fixes #2653 

## Changelog


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftCreateBlockForProposalBehaviour.java
```diff
@@ -0,0 +1,95 @@
...
+    assert (bftExtraDataCodec instanceof PkiQbftExtraDataCodec);
+    this.bftExtraDataCodec = (PkiQbftExtraDataCodec) bftExtraDataCodec;
+  }
+
+  @Override
+  public Block create(final long headerTimeStampSeconds) {
+    final Block block = blockCreator.createBlock(headerTimeStampSeconds);
+    final QbftContext qbftContext = protocolContext.getConsensusState(QbftContext.class);
+    final Optional<PkiBlockCreationConfiguration> pkiBlockCreationConfig =
+        qbftContext.getPkiBlockCreationConfiguration();
```
jframe, line 60: We could just do this the once when initialising the PkiQbftCreateBlockForProposalBehaviour as the config doesn't change

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftCreateBlockForProposalBehaviour.java
```diff
@@ -0,0 +1,95 @@
...
+
+    if (pkiBlockCreationConfig.isEmpty()) {
```
jframe, line 62: Could just pass in the PkiBlockCreationConfiguration from the QbftRoundFactory which would make this check unnecessary and remove the need to pass in the protocol context

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftCreateBlockForProposalBehaviour.java
```diff
@@ -0,0 +1,95 @@
...
+      throw new IllegalStateException(
+          "Empty PkiBlockCreationConfiguration when create proposal in PKI-mode");
+    }
+
+    return replaceCmsInBlock(block, pkiBlockCreationConfig.get());
+  }
+
+  private Block replaceCmsInBlock(
```
jframe, line 70: Be good to have a unit test around this

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftRoundFactory.java
```diff
@@ -80,6 +83,16 @@ public QbftRound createNewRoundWithState(
     final QbftMessageTransmitter messageTransmitter =
         new QbftMessageTransmitter(messageFactory, finalState.getValidatorMulticaster());
 
+    final QbftContext qbftContext = protocolContext.getConsensusState(QbftContext.class);
+    final CreateBlockForProposalBehaviour createBlockForProposalBehaviour;
+    if (qbftContext.getPkiBlockCreationConfiguration().isPresent()) {
```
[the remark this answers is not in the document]
jframe: So it does, oh well guess this have to stay this way then

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftCreateBlockForProposalBehaviour.java
lucassaldanha: True!

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftCreateBlockForProposalBehaviour.java
lucassaldanha: Done!

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftCreateBlockForProposalBehaviour.java
lucassaldanha: True! Completely forgot about it. Will add!

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
```diff
@@ -37,54 +36,77 @@ public BftExtraData decodeRaw(final Bytes input) {
...
-    final BftExtraData bftExtraData = super.decodeRaw(extraDataListAsBytes);
-
-    final Optional<Bytes> cms;
-    if (rlpInput.nextIsList() && rlpInput.nextSize() == 0) {
-      cms = Optional.empty();
+    rlpInput.skipNext();
+    rlpInput.skipNext();
+    rlpInput.skipNext();
+    rlpInput.skipNext();
+    rlpInput.skipNext();
```
siladu, line 48: Forgive my naive understanding with very limited context...I'm wondering how robust this is to potential future changes...is it possible/likely that this list will change and make this brittle?

lucassaldanha: You are 100%. This implementation is very dependent on the RLP format and the `QbftExtraDataCodec` implementation. However, this is the case for most RLP structures, as they are very rigid (e.g. order dependent, etc.).

The best we can do is to prevent this breaking in the future is to add tests around the decoding/encoding of the expected input to ensure we catch any issues (e.g. if someone changes the "standard" RLP structure or change some details in the `QbftExtraDataCodec` implementation, we fail a unit test to flag that we need to update this class).

Another alternative that we have explored (and might bring back to solve the duplication in the encode() method) is to use an "envelope" list. This won't solve the coupling between this class and the QbftExtraData codec but it will definitely make the code easier to read (it will adhere better to the RLP library API).

lucassaldanha: To add a bit more context. We don't use a key/value structure on our RLP encoding, it is all positional. That also makes things a bit more complicated when parsing and extracting specific values from the RLP byte array :)

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BftExtraDataCodec.java
```diff
@@ -25,7 +25,8 @@
   protected enum EncodingType {
     ALL,
     EXCLUDE_COMMIT_SEALS,
-    EXCLUDE_COMMIT_SEALS_AND_ROUND_NUMBER
+    EXCLUDE_COMMIT_SEALS_AND_ROUND_NUMBER,
+    EXCLUDE_CMS // TODO-lucas How can we achieve this w/o changing the BftExtraDataCodec base class
```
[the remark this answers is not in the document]
lucassaldanha: We ended up using another approach that didn't rely on the EncodingType.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
```diff
@@ -20,71 +23,67 @@
...
-    encoder.startList(); // start envelope list
+    final BytesValueRLPOutput rlpOutput = new BytesValueRLPOutput();
+    rlpOutput.startList();
+    // Read through extraData RLP list elements and write them to the new RLP output
+    new BytesValueRLPInput(encoded, false)
+        .readList(RLPInput::readAsRlp).stream()
+            .map(RLPInput::raw)
+            .forEach(rlpOutput::writeRLPBytes);
+    // Only write CMS if includeCms is true
+    rlpOutput.writeBytes(includeCms ? ((PkiQbftExtraData) bftExtraData).getCms() : Bytes.EMPTY);
```
jframe, line 80: This looks like when encoding without a CMS will add a list item with empty bytes but the decoding allows for the empty CMS to be represented by empty bytes. For consistency be nice if the empty CMS case was represented not being in the list I think.

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodecTest.java
```diff
@@ -112,30 +169,70 @@ public void fullyPopulatedDataProducesCorrectlyFormedExtraDataObject() {
...
+            pkiBftExtraData.getVanityData(),
+            pkiBftExtraData.getSeals(),
+            pkiBftExtraData.getVote(),
+            pkiBftExtraData.getRound(),
+            pkiBftExtraData.getValidators());
+
+    final Bytes encoded = bftExtraDataCodec.encode(bftExtraData);
+
+    assertThat(encoded).isEqualTo(expectedRawDecoding);
   }
```
jframe, line 234: Be great to have some tests showing that pki qbft extra data without cms can be handled by the qbft extra data codec to ensure we have binary compatibility.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
```diff
@@ -20,71 +23,66 @@
...
-    final Bytes extraDataListAsBytes = rlpInput.currentListAsBytes();
-    final BftExtraData bftExtraData = super.decodeRaw(extraDataListAsBytes);
+    final RLPInput rlpInput = new BytesValueRLPInput(input, false);
 
-    final Optional<Bytes> cms;
-    if (rlpInput.nextIsList() && rlpInput.nextSize() == 0) {
-      cms = Optional.empty();
+    final Bytes cms;
+    final List<RLPInput> elements = rlpInput.readList(RLPInput::readAsRlp);
+    if (elements.size() > QBFT_EXTRA_DATA_LIST_SIZE) {
```
usmansaleem, line 50: is there a requirement or scenario to check the upper limit of elements? Is `5` size without cms?

usmansaleem: @jframe ^

usmansaleem (approved):
LGTM - one minor query around PkiExtraDataCodec list size.

usmansaleem (approved):
LGTM

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/pki/PkiQbftExtraDataCodec.java
jframe: Unfortunately the upper limit comes from Qbft extra data codec. Don't see any real way around knowing what the expected number of items for qbft extra data is. Fortunately we have tests around this so if the qbft extra data rlp list size changes our pki qbft extra data tests will fail.

PR #2690, opened 2021-08-29
ignore all unknown fields for eth_call and eth_estimateGas

macfarla:

Changed to ignore all unknown fields in JsonCallParameter. This matches geth's behaviour. Different third-party libraries send through extra fields from time to time, for example
ethers.js 5.4.2 sends eth_estimateGas with a `type` parameter which Besu was rejecting.

```
Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "type" (class org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.JsonCallParameter), not marked as ignorable (11 known properties: "gasPrice", "value", "gasLimit", "gas", "from", "to", "maxFeePerGas", "strict", "maxPriorityFeePerGas", "payload", "data"])
at [Source: (StringReader); line: 1, column: 42492] (through reference chain: org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.JsonCallParameter["type"])
```

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Added Labelled gauges for metrics [#2646](https://github.com/hyperledger/besu/pull/2646)
 - support for `eth/66` networking protocol [#2365](https://github.com/hyperledger/besu/pull/2365)
 - update RPC methods for post london 1559 transaction [#2535](https://github.com/hyperledger/besu/pull/2535)
+- Ignore `type` when supplied to eth_estimateGas or eth_call. [\#2690](https://github.com/hyperledger/besu/pull/2690)
 
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


emclab:
I have exactly same problem with ethersjs 5.4.2. ethers adds a "type" field in tx which is not recognized by server. What I did was upgrading to 5.4.6 and the problem disappeared so far.

macfarla:
> I have exactly same problem with ethersjs 5.4.2. ethers adds a "type" field in tx which is not recognized by server. What I did was upgrading to 5.4.6 and the problem disappeared so far.

Fixed by updating to ethers.js 5.4.6

shemnon (approved):

Issue #1317, opened 2020-08-17
Enable plugins to expose custom JSON RPC / GraphQL / WebSocket methods


aldoborrero:
### Description

As a plugin writer, I would like to have the possibility of exposing custom JSON RPC / GraphQL / WebSocket methods (I'll focus only on JSON RPC for the rest of the issue but it should apply to all of those).

Currently, the plugin API is limited by how we can interact with the outside world. Having the possibility of custom JSON RPC commands is immensely useful when you want to extend the capabilities of Besu beyond the default set of exposed methods.

In terms of the API, we can add the following to `BesuPlugin` interface:

```Java
...

/** Called when the plugin wants to expose custom JSON RPC methods.
* If not overridden, this method does nothing for convenience.
**/
default Map<String, JsonRpcMethod> createJsonRpcMethods() {
    return Collections.emptyMap();
}
```

This method would be called once at the `Registering` phase. The format for each method name could be the following:

```
<plugin-name-id>_<api-method>
```

Like this:

```
plugin-besu-exflo_retrieveTokenBalances
```

Each plugin is responsible of generating a custom response that can be serialized by Jackson.

We can reuse part of the work that has been done already in the class [`PluginsJsonRpcMethods`](https://github.com/hyperledger/besu/blob/master/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/PluginsJsonRpcMethods.java). That class receives the list of Plugins, so apart of registering the reload configuration, we can iterate for each plugin and see if they have defined a map of `JsonRpcMethods`.

Currently, neither of `JsonRpcMethod`, `JsonRpcMethods`, `JsonRpcRequestContext`... are exposed to `plugin-api` module. So a worthwhile discussion would be what is the best approach to expose those relevant entities to plugins.

Also, probably, I might be forgetting anything that could be crucial to take into account before any attempt of implementation.

### Acceptance Criteria

- It should not override existing JSON RPC commands (throw an error if the id of the method is one of the native ones).
- It should expose custom JSON RPC methods.
- It should expose custom GraphQL mutations/queries.
- It should allow custom WebSocket subscriptions/methods.

Maybe we can start with just only supporting JSON RPC and later on adding support to GraphQL and WebSockets.

### Additional Information

[Quorum has implemented a similar feature already](https://github.com/jpmorganchase/quorum/pull/956). 

If an approach is agreed upon, I would be happy to implement it.

antonydenyer:
[...]

PR #2704, opened 2021-08-31
Expose json rpc config to plugins through JsonRpcConfigurationService

antonydenyer:
## PR description
Expose json rpc config to plugins through JsonRpcConfigurationService. 
This is instead of exposing generic untyped arguments as explored in https://github.com/hyperledger/besu/pull/2661


plugin-api/src/main/java/org/hyperledger/besu/plugin/services/JsonRpcConfigurationService.java
```diff
@@ -0,0 +1,47 @@
...
+   * @return Host that JSON-RPC HTTP is listening on
+   */
+  String getJsonRpcHost();
+
+  /**
+   * The list of APIs enabled on JSON-RPC HTTP service
+   *
+   * @return list of APIs enabled on JSON-RPC HTTP service
+   */
+  Collection<String> getJsonRpcApis();
```
atoulme, line 46: Do you also need cors host and allowlist?

macfarla (approved):

vmichalik:
btw if you hadn't seen it already, #1317 requests this functionality

antonydenyer:
I ended up working around the issue in the end by getting what I wanted from picocli. It's a hack and relies on us continuing to use picocli.

```java
@Option(
      names = {"--plugin-permissioning-rpc-http-port"},
      defaultValue = "--rpc-http-port",
      hidden = true,
      preprocessor = BesuPreprocessor.class)
  private Integer port;


public static class BesuPreprocessor implements CommandLine.IParameterPreprocessor {
    @Override
    public boolean preprocess(
        final Stack<String> args,
        final CommandLine.Model.CommandSpec commandSpec,
        final CommandLine.Model.ArgSpec argSpec,
        final Map<String, Object> info) {
      argSpec.setValue(commandSpec.findOption(argSpec.defaultValue()).getValue());
      return true;
    }
  }
```

I don't think we should document this as 'the way to do it' but leaving here for reference until an agreement can be reached on how to expose config to plugins.

PR #2748, opened 2021-09-10
Cleanup metrics connections after idle timeout

rolfyone:
When metrics are in use in teku (which uses besu metrics), and the prometheus service restarts, connections can be left open and never closed.

These connections should close after an idle timeout to avoid causing issues with running out of ports.

Reported in teku issue https://github.com/ConsenSys/teku/issues/4327 .

Signed-off-by: Paul Harris <paul.harris@consensys.net>

## Changelog


ajsutton (approved):
LGTM.

PR #2752, opened 2021-09-10
Metrics idle timeout is now configurable

Neurone:
## PR description
#2748 introduced a timeout for metrics TCP connections. With this PR user can set a value in seconds for that timeout.

## Changelog


ajsutton (approved):
LGTM.  Thanks.

PR #2761, opened 2021-09-14
Gradle 7.2 upgrade / Java 17 Build Support

shemnon:
* Upgrade to Gradle 7.2, which supports Java 17
* Regenerate gradlew script and wrapper (as recommended)
* Suppress the removal warning for `AccessController` for now
* Update the JMH version to get rid of Gradle deprecation warning
* Add spotless `targetExcludes` so it won't re-format submodules

## Changelog


antonydenyer (approved):

build.gradle
```diff
@@ -514,6 +517,8 @@ task evmToolStartScripts(type: CreateStartScripts) {
   defaultJvmOpts = [
     "-Dsecp256k1.randomize=false"
   ]
+  unixStartScriptGenerator.template = resources.text.fromFile("${projectDir}/besu/src/main/scripts/unixStartScript.txt")
+  windowsStartScriptGenerator.template = resources.text.fromFile("${projectDir}/besu/src/main/scripts/windowsStartScript.txt")
   doLast {
```
antonydenyer, line 522: ${rootDir} instead maybe?

shemnon: The tasks are only in the root project.  The only place we use rootdir in the gradle scripts right now is in sub-projects.

PR #2768, opened 2021-09-15
Proposal for breaking change to the way plugins load cli options

antonydenyer:
# Description 

https://wiki.hyperledger.org/display/BESU/DRAFT+-+Pico+CLI+Plugin+Integration

The main change is to move preparePlugins into the begging of the run method. This means that all the cli parsing will happen before any plugins are called. The only way to register cli options will be to have them on the `BesuPlugin` e.g

https://github.com/hyperledger/besu/pull/2768/files#diff-2080ac0bf38c1e492921f4191b738b729195e0b1ad82657abc9103f19ac87cd8L1195

```
@AutoService(BesuPlugin.class)
public class MyPlugin implements BesuPlugin {

  @Option(
    names = {"--Xplugin-my-option"},
    defaultValue = "foo")
  String myOption;

  @Override
  public Optional<String> getName() {
    return Optional.of("my");
  }
}
```

The other breaking change is that you should override getName and provide a meaningful plugin name. This is used for the namespace when registering the cli arguments.

https://github.com/hyperledger/besu/pull/2768/files#diff-2080ac0bf38c1e492921f4191b738b729195e0b1ad82657abc9103f19ac87cd8R1204

### DIFF
--- a/acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/BesuNode.java
+++ b/acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/BesuNode.java
@@ -120,6 +120,7 @@ public class BesuNode implements NodeConfiguration, RunnableNode, AutoCloseable 
   private boolean isDnsEnabled = false;
   private Optional<Integer> exitCode = Optional.empty();
   private Optional<PkiKeyStoreConfiguration> pkiKeyStoreConfiguration = Optional.empty();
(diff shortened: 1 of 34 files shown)


Conanza00 (approved):

jflo:
Is this still active or can it be closed? We would like to clean up old outstanding PRs.

macfarla:
@antonydenyer this PR has been hanging around for a while - if it's still relevant, what do you think about moving it to a branch, maybe with a writeup about what's it's demonstrating - and we can revisit in the future when it's relevant?

PR #2789, opened 2021-09-21
Rollback TransactionGasCalculator

shemnon:
Rollback the transactionGasCalculator changes. This was done in
preparation for the EVM modularization.  Instead, destructure the data
needed to calculate the gas instead of passing in the whole transaction.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -274,7 +275,7 @@ public TransactionProcessingResult processTransaction(
         return TransactionProcessingResult.invalid(validationResult);
       }
 
-      final MutableAccount senderMutableAccount = sender.getMutable();
+      final var senderMutableAccount = sender.getMutable();
```
antonydenyer, line 278: I'm pretty happy using `var` everywhere but not sure if everyone else is?

antonydenyer (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
atoulme: Worth discussing next contributor call

PR #2792, opened 2021-09-21
Alternative proposal for breaking change to the way plugins load cli options

antonydenyer:
https://wiki.hyperledger.org/display/BESU/DRAFT+-+Pico+CLI+Plugin+Integration

The main change is to move preparePlugins into the begging of the run method. This means that all the cli parsing will happen before the method register is called. 

The only way to register cli options is to use the extra hook provided by `BesuPlugins::registerPicoCLIOptions`

### DIFF
--- a/plugin-api/src/main/java/org/hyperledger/besu/plugin/BesuPlugin.java
+++ b/plugin-api/src/main/java/org/hyperledger/besu/plugin/BesuPlugin.java
@@ -14,6 +14,8 @@
  */
 package org.hyperledger.besu.plugin;
 
+import org.hyperledger.besu.plugin.services.PicoCLIOptions;
+
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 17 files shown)


plugin-api/src/main/java/org/hyperledger/besu/plugin/BesuPlugin.java
```diff
@@ -36,6 +38,16 @@
     return Optional.of(this.getClass().getName());
   }
 
+  /**
+   * Called when the plugin is loaded so that you can register cli options to be parsed
+   *
+   * @param cliOptions the hook to register PicoCLIOptions.
+   */
+  default void registerPicoCLIOptions(final PicoCLIOptions cliOptions) {
```
shemnon, line 46: My concern is that, even with a default interface method, it forces all plugins to be PicoCLI aware and deeply embedded in the lifecycle.  Not every plugin cares or should care about PicoCLI, which is why I exposed it as a service.

antonydenyer: Yeah, I'm not a fan of this either. However, it is more explicit than the other proposal https://github.com/hyperledger/besu/pull/2768.

jflo:
Is this still active or can it be closed? We would like to clean up old outstanding PRs.

antonydenyer:
Related to https://wiki.hyperledger.org/display/BESU/DRAFT+-+Pico+CLI+Plugin+Integration and https://github.com/hyperledger/besu/pull/2768

Maybe useful if someone looks at plugins again.

PR #2805, opened 2021-09-27
reduce use of global quorum config

### LINKED ISSUE #2775: Replace GoQuorum static flag with dynamic flag
The static flag in (`GoQuorumOptions`)[https://github.com/hyperledger/besu/blob/78fe588c9dc032166250f0bfd649ca1b2cc9dc45/config/src/main/java/org/hyperledger/besu/config/GoQuorumOptions.java#L25] should be replaced.  

This is causing flakey tests when running multi-threaded tests, sometimes it is enabled and sometimes it is not. Because the switch is accessed in parts fo the code far from privacy it can cause non-deterministic behavior in parallel tests.

taccatisid:
## PR description

The use of the org.hyperledger.besu.config.GoQuorumOptions static variable is ugly. In particular does changing its value during tests cause flaky tests. To ameliorate the issue this PR

- removes unused code that did rely on GoQuorumOptions
- removes access to GoQuorumOptions deep in the call stack by looking it up in the transaction validator instead
- change the TransactionDecoder to explicitly take the  goQuorumCompatbility flag as an argument where access to the  transaction validator is not easy
- rewriting all tests that did change the GoQuorumOptions by mocking  the option change instead.

Signed-off-by: Taccat Isid <taccatisid@protonmail.com>

## Fixed Issue(s)

fixes #2775.

## Changelog


ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolTest.java
```diff
@@ -79,34 +78,36 @@
...
-      mock(TransactionPool.TransactionBatchAddedListener.class);
-  private final MetricsSystem metricsSystem = new NoOpMetricsSystem();
+  @Mock private MainnetTransactionValidator transactionValidator;
+  @Mock private PendingTransactionListener listener;
+  @Mock private TransactionPool.TransactionBatchAddedListener batchAddedListener;
+  @Mock private TransactionPool.TransactionBatchAddedListener pendingBatchAddedListener;
 
   @SuppressWarnings("unchecked")
-  private final ProtocolSchedule protocolSchedule = mock(ProtocolSchedule.class);
+  @Mock
```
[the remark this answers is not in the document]
taccatisid: Semi--it is the result of running spotlessApply which I assumed is the format we want. I agree that it silly.

macfarla (approved):
Getting rid of changing the value in tests should fix the flakiness. It's not perfect ie still having different ways of getting the flag value but it's better than it was.

taccatisid:
@shemnon Given that you raised the issue, do you want to have a look at this PR?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLDataFetchers.java
```diff
@@ -282,7 +283,9 @@ public GraphQLDataFetchers(final Set<Capability> supportedCapabilities) {
   private TransactionAdapter getTransactionAdapter(
       final TransactionWithMetadata transactionWithMetadata) {
     final Transaction transaction = transactionWithMetadata.getTransaction();
-    return goQuorumPrivacyParameters.isPresent() && transaction.isGoQuorumPrivateTransaction()
+    final boolean isGoQuorumPrivateTransaction =
+        transaction.isGoQuorumPrivateTransaction(GoQuorumOptions.goQuorumCompatibilityMode);
```
shemnon, line 287: Can the check of `GoQuorumOptions.goQuorumCompatibilityMode` be skipped and we presume that if the `goQuorumPrivacyParameters` are set that we need to goQuorum decode it?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetCode.java
```diff
@@ -64,7 +64,8 @@ protected BlockParameterOrBlockHash blockParameterOrBlockHash(
   @Override
   protected String resultByBlockHash(final JsonRpcRequestContext request, final Hash blockHash) {
     final Address address = request.getRequiredParameter(0, Address.class);
-    if (GoQuorumOptions.goQuorumCompatibilityMode && privacyParameters.isPresent()) {
+    final boolean isGoQuorumCompatibilityMode = GoQuorumOptions.goQuorumCompatibilityMode;
+    if (isGoQuorumCompatibilityMode && privacyParameters.isPresent()) {
```
shemnon, line 68: Same here, can we presume if `privacyParameters` is set we need goQuorum handling?  Or is there a goQuorum sub option in `privacyParametes`?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionGoQuorumTest.java
```diff
@@ -39,52 +37,47 @@
...
-  public static void beforeClass() {
-    GoQuorumOptions.goQuorumCompatibilityMode = true;
-  }
-
-  @AfterClass
-  public static void afterClass() {
-    GoQuorumOptions.goQuorumCompatibilityMode =
-        GoQuorumOptions.GOQUORUM_COMPATIBILITY_MODE_DEFAULT_VALUE;
-  }
+  private static final boolean goQuorumCompatibilityMode = true;
```
shemnon, line 40: Does this have to be static?  Can it be a regular field?

shemnon (commented):
This reduces the use of, but does not eliminate the static field configuration. While it reduces the window for test flakiness. And its presence still makes it possible that new flakiness could be introduced unless we are very vigilant in the side effects.  I think there is still room to eliminate it by changing a few classes from static methods to instance classes provided via Dependency Injection.

macfarla:
The second commit that fully eradicates the flag touches a lot of files. I think I prefer the "mostly gone" version. Can we memoize it so once it's set it can't be changed - that would prevent accidental misuse?

taccatisid:
> This reduces the use of, but does not eliminate the static field configuration. While it reduces the window for test flakiness. And its presence still makes it possible that new flakiness could be introduced unless we are very vigilant in the side effects. I think there is still room to eliminate it by changing a few classes from static methods to instance classes provided via Dependency Injection.

@shemnon could you have another superficial look at the updated PR? The code surface required to be touched turned out to be quite large. And in light of looming tickets like the separation of privacy into a plugin, we (revenant) feel that threading all those extra objects through the call stack is not an improvement of the codebase.

If we would instead just take the first of the two commits, and add a getter and a write once setter for the static `goQuorumCompatibilityMode` we would achieve the same safety effects with much less code churn.

shemnon:
I'll take a deeper look tonight, but making storage dependent on the genesis config doesn't look like a good solution.

taccatisid:
> The value of this bug is to go from `n` to zero. To go from `n` to some smaller `n` still leaves the core issue that it is not safe to test in parallel.

The issue that this is 100% safe for concurrent testing is easy to achieve without reducing that static flag use to zero. The first commit (2591f117a6d82c831687dd25ecc906c64ff3939b) already uses mocking to test the `goQorum = true` case without touching the global variable. To make it impossible to have this cause a problem in test we would just have to replace the `public static boolean` with a `private static boolean` and a getter and write once setter.

> I'll take a deeper look tonight, but making storage dependent on the genesis config doesn't look like a good solution.

It is already dependent, the PR just made that explicit. Storage needs to parse transactions which depends on the quorum flag which is stored in the genesis.

> I think we need to refactor the main culprit first: the static methods on TransactionDecoder. This needs to become a factory instance, and two fields would be if goQuorum mode is enabled and what the signature algorithm is (secp256k1 vs secp256r1).  The work will then be plumbing the factory instance into the places we need to get transactions decoded from RLP. I believe that will not impact the storage classes.

I considered this. It would be nicer code than what we have. But it would be a larger refactor: it would touch more or less all the places that the current PR touches but in a less mechanical way. 

And without further changes it would not decouple the storage classes from transaction decoding (and thus the quorum flag). This mostly results from us not having decoupled the ability to read, write, and (in-memory) represent transactions. Thus as soon as you do one of these, you need the quorum flag--even if you never use it because you do not parse transactions.

shemnon:
Consider two simultaneous tests, one decoding a mainnet encoded transaction and one decoding a goquorum private transaction.  The r values overlap and only the static field is used to declare one mainnet and one goquorum.  It is not 100% safe to use a static field for such a basic operation.

taccatisid:
When protecting the quorum flag from being changed during runtime like I proposed in [the most recent commit](https://github.com/hyperledger/besu/commit/5752224bd999d0494d9c45153ec8c07bc3901176#diff-5a66b8e2e487d1f8d8a615d67cec2cfb60f9047d6b027a427531dbe406cb8b1bR26), no tests could run concurrently with different flag values. Tests could not even change the quorum flag at all.

Instead, tests have to use mocks, static injection, or slightly adapt the tested class to locally overwrite the quorum config value. In the current code base, **three** test classes are playing with the static quorum flag. Changing them is much more economical than changing 200+ files as in the previous version of this PR.

shemnon:
I don't think any solution is going to "look good" unless something like Dagger is adopted, so the DI framework handles the plumbing of strange config settings to the far corners of the app.  Adding another hack (like single threading the static value) isn't any better than the current situation.

taccatisid:
After chats with revenant and @lucassaldanha we came to the conclusion that we dislike the protected static global variant of the PR a little less than the code churning one.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolTest.java
```diff
@@ -545,7 +543,10 @@ public void shouldNotNotifyPeerForPendingTransactionsIfItDoesntSupportEth65() {
     when(peerPendingTransactionTracker.isPeerSupported(peer, EthProtocol.ETH65)).thenReturn(false);
     when(peerPendingTransactionTracker.isPeerSupported(validPeer, EthProtocol.ETH65))
         .thenReturn(true);
-    when(transactionValidator.validate(any(), any(Optional.class), any())).thenReturn(valid());
+    // TODO: This mock should be called but is not. Make it happen and remove the lenient.
```
macfarla, line 546: tx validation does not occur here

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolTest.java
```diff
@@ -752,10 +761,12 @@ public void shouldIgnoreEIP1559TransactionWhenNotAllowed() {
             ImmutableTransactionPoolConfiguration.builder().txFeeCap(Wei.ONE).build());
     when(transactionValidator.validate(any(Transaction.class), any(Optional.class), any()))
         .thenReturn(valid());
-    when(transactionValidator.validateForSender(
```
macfarla, line 755: shouldIgnoreEIP1559TransactionWhenNotAllowed validateForSender() doesn't get called because the tx is rejected before that

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolTest.java
```diff
@@ -796,10 +807,12 @@ public void shouldIgnoreEIP1559TransactionBeforeTheFork() {
     when(protocolSpec.getFeeMarket()).thenReturn(FeeMarket.legacy());
     when(transactionValidator.validate(any(Transaction.class), any(Optional.class), any()))
         .thenReturn(valid());
-    when(transactionValidator.validateForSender(
-            any(Transaction.class),
-            nullable(Account.class),
-            any(TransactionValidationParams.class)))
+    lenient()
```
macfarla, line 810: shouldIgnoreEIP1559TransactionBeforeTheFork  validateForSender() doesn't get called because the tx is rejected before that

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolTest.java
```diff
@@ -359,7 +355,9 @@ public void shouldNotAddRemoteTransactionsThatAreInvalidAccordingToStateDependen
     givenTransactionIsValid(transaction2);
     when(transactionValidator.validate(eq(transaction1), any(Optional.class), any()))
         .thenReturn(valid());
-    when(transactionValidator.validateForSender(transaction1, null, true))
```
macfarla, line 362: this is the only one I'm still not sure of

macfarla (commented):
mock investigation

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolTest.java
macfarla: Ok it's because it's the _other_ validateForSender() that gets called here.
This works:
```
  @Test
  public void shouldNotAddRemoteTransactionsThatAreInvalidAccordingToStateDependentChecks() {
    givenTransactionIsValid(transaction2);
    when(transactionValidator.validate(eq(transaction1), any(Optional.class), any()))
        .thenReturn(valid());
    when(transactionValidator.validateForSender(eq(transaction1), eq(null), any(TransactionValidationParams.class)))
        .thenReturn(ValidationResult.invalid(NONCE_TOO_LOW));
    transactionPool.addRemoteTransactions(asList(transaction1, transaction2));

    assertTransactionNotPending(transaction1);
    assertTransactionPending(transaction2);
    verify(batchAddedListener).onTransactionsAdded(singleton(transaction2));
    verify(transactionValidator).validate(eq(transaction1), any(Optional.class), any());
    verify(transactionValidator).validateForSender(eq(transaction1), eq(null), any(TransactionValidationParams.class));
    verify(transactionValidator).validate(eq(transaction2), any(Optional.class), any());
    verify(transactionValidator).validateForSender(eq(transaction2), any(), any());
    verifyNoMoreInteractions(transactionValidator);
  }
```

macfarla (approved):
LGTM

macfarla:
This also fixes https://github.com/hyperledger/besu/issues/2813

PR #2809, opened 2021-09-27
Jumpdest cache bits

### LINKED ISSUE #2607: jump opcode analysis

jflo:
## PR description

Adds caching of valid jump destinations in a contract. This should reduce re-work previously required each time any function in the contract is called.

## Fixed Issue(s)

fixes #2607 

## Changelog


evm/build.gradle
```diff
@@ -40,6 +40,7 @@ dependencies {
   implementation 'org.apache.logging.log4j:log4j-api'
   implementation 'org.hyperledger.besu:bls12-381'
   implementation 'net.java.dev.jna:jna'
+  implementation 'com.github.ben-manes.caffeine:caffeine:3.0.4'
```
matkt, line 43: I think we need to move the version to the version.gradle file

matkt (approved):
LGTM

CHANGELOG.md
```diff
@@ -4,14 +4,15 @@
 ### Additions and Improvements
 * The EVM has been factored out into a standalone module, suitable for inclusion as a library. [#2790](https://github.com/hyperledger/besu/pull/2790)
 * Low level performance improvements changes to cut worst-case EVM performance in half. [#2796](https://github.com/hyperledger/besu/pull/2796)
-
+* reduces need for JUMPDEST analysis via contract caching [#2607](https://github.com/hyperledger/besu/issues/2607)
```
shemnon, line 7: Nit: newline after changelist row before header.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -42,7 +42,6 @@
   protected final long height;
   protected final TrieLogLayer trieLog;
   private final Hash worldStateRootHash;
-
```
shemnon, line 45: Nit: this touch will make "git spelunking" hard.  Please undo before submit.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/JumpOperationTest.java
```diff
@@ -136,17 +143,48 @@ public void shouldHaltWithInvalidJumDestinationWhenLocationIsOutsideOfCodeRange(
...
+            .pushStackItem(UInt256.fromHexString("0x03"))
+            .code(toRun)
+            .build();
+    frame.setPC(CURRENT_PC);
+
+    OperationResult result = operation.execute(frame, evm);
+    assertThat(result.getHaltReason()).isEmpty();
+    Mockito.verify(toRun, times(1)).calculateJumpDests();
+
+    // do it again to prove we don't recalc
```
shemnon, line 177: Optional: A cache-miss (new code, maybe "0x02") and a `Mockito.verify(toRun, times(2)).calculateJumpDests();` would make it clear on reading that the verify doesn't reset between checks.

evm/src/main/java/org/hyperledger/besu/evm/Code.java
```diff
@@ -28,21 +32,38 @@
...
    * @param bytes The byte representation of the code.
+   * @param codeHash the Hash of the bytes in the code.
    */
-  public Code(final Bytes bytes) {
+  public Code(final Bytes bytes, final Hash codeHash) {
     this.bytes = bytes;
+    this.codeHash = codeHash;
+  }
+
+  public Code(final Bytes bytecode, final Hash codeHash, final long[] validJumpDestinations) {
```
shemnon, line 58: Static side effects are not appropriate for constructors.  If you need side effects do it in a factory method that calls the constructor and then does the side effects.

evm/src/main/java/org/hyperledger/besu/evm/worldstate/AbstractWorldUpdater.java
```diff
@@ -97,6 +98,10 @@ public EvmAccount getAccount(final Address address) {
 
   @Override
   public void deleteAccount(final Address address) {
+    Account removed = get(address);
+    if (removed != null && removed.hasCode()) {
+      JumpDestCache.getInstance().invalidate(removed.getCodeHash());
```
shemnon, line 103: This purge is unneeded.  If accounts A,B,C,... share the same code Create A, Create B, Delete B, Create C, Delete C, etc.  Even though B and later are deleted A remains and would benefit from the cached jumpdests.  I think we should rely on the the cache's expiration policy instead of pre-deleting (unless we have reference counting handy).

shemnon (commented):
I cannot approve this PR.  I have asked gently numerous times that the cache not be stored as a static field.  No other data structure in Besu does this. Dependency Injection is the design standard.

A better place would be to keep the cache *and the jump dest analysis logic* in the EVM class.  There is one EVM per hard form in the mainnet protocol spec, so they will not change that often during run time.

Putting this there would insulate us from future changes that are in progress.  One specifically is the EVM container format that will include a pre-computed jump table and will not have the code starting at the zero index of the container.  It also keeps us from hanging on to a static singleton.

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ContractCacheOptions.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class ContractCacheOptions implements CLIOptions<JumpdestCacheConfiguration> {
+
+  public static final String CONTRACT_CACHE_WEIGHT = "--Xcontract-code-cache-weight-kb";
+
+  @SuppressWarnings({"FieldCanBeFinal", "FieldMayBeFinal"})
+  @CommandLine.Option(
+      names = {CONTRACT_CACHE_WEIGHT},
+      description =
+          "size in kilobytes to allow the cached"
```
macfarla, line 34: nit: add space after cached

evm/src/main/java/org/hyperledger/besu/evm/worldstate/WorldView.java
```diff
@@ -20,7 +20,6 @@
 /** Generic interface for a view over the accounts of the world state. */
 public interface WorldView {
   WorldView EMPTY = address -> null;
-
```
macfarla, line 23: another spurious change

macfarla (commented):
couple of small comments

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ContractCacheOptions.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2: nit: new files should have * Copyright contributors to Hyperledger Besu  https://wiki.hyperledger.org/display/TSC/Copyright+and+License+Policy

matkt:
> I cannot approve this PR. I have asked gently numerous times that the cache not be stored as a static field. No other data structure in Besu does this. Dependency Injection is the design standard.
> 
> A better place would be to keep the cache _and the jump dest analysis logic_ in the EVM class. There is one EVM per hard fork in the mainnet protocol spec, so they will not change that often during run time.
> 
> Putting this there would insulate us from future changes that are in progress. One specifically is the EVM container format that will include a pre-computed jump table and will not have the code starting at the zero index of the container. It also keeps us from hanging on to a static singleton.

IMO we must try to find a consensus on how to do this before merging this PR

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/JumpOperationTest.java
jflo: That would not pass though, since we are spying on the Code instance. A new code being analyzed would not cause toRun to be re-analyzed. If we want to test explicitly for cache hits/misses, I can add that, but this is really watching to see that for a given Code which contains the Jump operation, the analysis is only run once.

jflo:
Closed in favor of https://github.com/hyperledger/besu/pull/2821

PR #2810, opened 2021-09-27
Make ExceptionalHaltReason an interface

shemnon:
Migrate ExceptionalHaltReason to an interface from an enum. The enum
made it difficult for downstream implementations to add custom reasons
as to why the EVM has halted. Switching to an interface is needed
because enums are final and not extensible.

## Changelog


macfarla (approved):
LGTM

Issue #2744, opened 2021-09-09
QBFT to log warning when using validator contract if genesis extra data has validator


jframe:
When using the validator contract in QBFT the validators no longer come from the extra data but we still need valid extra data for the genesis block. We should a log a warning if there are validators in the genesis block extra data and we are using a validator contract so the user is aware the extra data validators aren't being used.

In QbftBesuControllerBuilder:validateContext could add another condition like this:
```    
if (genesisConfig.getConfigOptions().getQbftConfigOptions().getValidatorContractAddress().isPresent()
        && !bftBlockInterface().get().validatorsInBlock(genesisBlockHeader).isEmpty()) {
      LOG.warn("Using validator but genesis block contains signers.");
    }
```

Existing warning "Genesis block contains no signers - chain will not progress." should only show if using validator header mode.

PR #2814, opened 2021-09-28
QBFT to log warning when using validator contract if genesis extra data has validator

siladu:
Also prevent existing warning from logging when using validator contract with no signers in genesis extra data.

Fixes https://github.com/hyperledger/besu/issues/2744


antonydenyer (approved):

usmansaleem (approved):
LGTM

besu/src/main/java/org/hyperledger/besu/controller/QbftBesuControllerBuilder.java
```diff
@@ -265,11 +265,44 @@ protected ProtocolSchedule createProtocolSchedule() {
...
+      final BlockHeader genesisBlockHeader) {
+    return genesisConfig
+            .getConfigOptions()
+            .getQbftConfigOptions()
+            .getValidatorContractAddress()
+            .isEmpty()
+        && noSignersExistIn(genesisBlockHeader);
+  }
+
+  private boolean signersExistIn(final BlockHeader genesisBlockHeader) {
```
jframe, line 298: nit: not convinced this method is adding a lot of value. For the one place this is used I think noSignersExistIn(genesisBlockHeader) is easier to read.

besu/src/main/java/org/hyperledger/besu/controller/QbftBesuControllerBuilder.java
```diff
@@ -265,11 +265,44 @@ protected ProtocolSchedule createProtocolSchedule() {
...
     }
   }
 
+  private boolean usingValidatorContractModeButSignersExistIn(
+      final BlockHeader genesisBlockHeader) {
+    return genesisConfig
+            .getConfigOptions()
+            .getQbftConfigOptions()
+            .getValidatorContractAddress()
+            .isPresent()
```
jframe, line 284: Would be nice to remove the duplication between this and usingValidatorBlockHeaderModeButNoSignersIn by creating isValidatorContractMode method

besu/src/main/java/org/hyperledger/besu/controller/QbftBesuControllerBuilder.java
siladu: In order to keep the positively worded method, I have changed `signersExistIn` and removed noSignersExistIn (using negation instead)

jflo (approved):

jframe (approved):
LGTM

PR #2816, opened 2021-09-28
fix: private contracts not able to call public contracts that call other public contracts

### LINKED ISSUE #2803: Private Transaction Failed when invoke a Public Contract with 2 Indirections
### Description
As an besu developer
I want to deploy a private contract which read from public state with two inderections
so my logic executed correctly according to private state read/write rules.

### Acceptance Criteria
* The lastes version of Besu behaves in the similar way to Besu v21.1.5 versions

### Steps to Reproduce (Bug)
1. Clone the Repository https://github.com/albert0-hernandez/private-to-public-besu-tester.git
2. Follow the Readme steps to install, run & test
3. See that the test labelled as 2.c. failed

**Expected behavior:** 
With Besu versions up to 21.1.5 the test will succeeded with no problem.

1. Clone the Repository https://github.com/albert0-hernandez/private-to-public-besu-tester.git
2. Modify the base version of Besu from 21.7.4 to 21.1.5 on hyperledger-besu-tessera-network/config/templates/docker-compose-microservices.yaml
3. Follow the Readme steps to  run & test
3. See that the test labelled by 2.c. succeeded

**Actual behavior:** 
When you deploy a configuration of Smartcontracts that chain an execution like:
    Private SC (calls) -> Public SC (calls) -> Public SC 
The execution failed.

**Frequency:** 
100%

* Software version: >= 21.1.6 
* Java version: Official Dockerized version
* OS Name & Version: Docker

### Additional Information
This behaviour is reproduced migrating a validated fully tested SC solution on 21.1.3 to 21.7.4.
We created a sample project to better (and simplified) behaviour that we observe after migrating.

antonydenyer:
## PR description
Private contracts not able to call public contracts that call other public contracts. 

The root cause is that during `transferValue` the `MessageCallProcessor` attempts to get a mutable account. At this point the evm throws an exception as it is rightly not mutable!

This can be avoided in situations when there is no value to transfer.

## Fixed Issue(s)
fix #2803 Private Transaction Failed when invoking a Public Contract that calls another Public Contract

## Changelog

### DIFF
--- a/acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/privacy/PrivateContractPublicStateAcceptanceTest.java
+++ b/acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/privacy/PrivateContractPublicStateAcceptanceTest.java
@@ -22,6 +22,8 @@ import org.hyperledger.besu.tests.acceptance.dsl.privacy.ParameterizedEnclaveTes
 import org.hyperledger.besu.tests.acceptance.dsl.privacy.PrivacyNode;
 import org.hyperledger.besu.tests.web3j.generated.CrossContractReader;
 import org.hyperledger.besu.tests.web3j.generated.EventEmitter;
+import org.hyperledger.besu.tests.web3j.generated.RemoteSimpleStorage;
+import org.hyperledger.besu.tests.web3j.generated.SimpleStorage;
 import org.hyperledger.enclave.testutil.EnclaveType;
 
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


evm/src/main/java/org/hyperledger/besu/evm/processor/MessageCallProcessor.java
```diff
@@ -89,18 +90,25 @@ protected void codeSuccess(final MessageFrame frame, final OperationTracer opera
    * of the world state of this executor.
    */
   private void transferValue(final MessageFrame frame) {
-    final MutableAccount senderAccount =
-        frame.getWorldUpdater().getSenderAccount(frame).getMutable();
+    final EvmAccount senderAccount = frame.getWorldUpdater().getSenderAccount(frame);
```
macfarla, line 93: I think it might help to have some more comments? explaining what's going on - so we only need a mutable account if we're _actually_ transferring value ie value > 0

and if we're accessing public state from private state then value == 0 AND nothing is or needs to be mutable

macfarla (approved):
I think it's good. I like the added acceptance test. A few comments.

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/privacy/PrivateContractPublicStateAcceptanceTest.java
```diff
@@ -169,4 +171,43 @@ public void privateContractMustNotBeAbleToCallSelfDestructOnPublicContract() thr
         .returns(
             "0x", e -> ((PrivateTransactionReceipt) e.getTransactionReceipt().get()).getOutput());
   }
+
+  @Test
```
macfarla, line 175: did you verify that this test failed without the fix?

antonydenyer: Yes, https://github.com/hyperledger/besu/pull/2816/commits/342ee16f956a725af11262f50a66243d21ed1b22 causes the original test to fail which is correct as we are actually trying to send value to a public contract from a private contract. 

Added the additional test as a precaution.

evm/src/main/java/org/hyperledger/besu/evm/processor/MessageCallProcessor.java
antonydenyer: > I think it might help to have some more comments?

I'm always a bit reticent to put comments in. But in this case, you're right it's not clear what the intent/reason was.

Issue #2742, opened 2021-09-08
PeerDiscoveryTimestampsTest.lastContactedTimestampUpdatedOnOutboundMessage() is flakey


jflo:
### Acceptance Criteria
* Test passes consistently and repeatably.

### Steps to Reproduce (Bug)
1. Example failure can be found [here](https://app.circleci.com/pipelines/github/hyperledger/besu/10510/workflows/4de566cc-7861-467d-9488-758a7fdcc749/jobs/56455).

**Expected behavior:** 
Test always pass.

**Actual behavior:** 
Test sometimes fails when it runs "too fast" and the compared to timestamp is on the same millisecond.

**Frequency:** 
Multiple times a week in the course of CI.

jflo:
It might help to avoid using the wall clock, and instead try the `org.hyperledger.besu.testutil.TestClock` as a test fixture.

catcs:
I can work on this issue.

PR #2836, opened 2021-10-01
2742: Fix flaky timestamp unit test

catcs:
Signed-off-by: Caitlin Harding <chardingcs@gmail.com>

## PR description
The unit test `PeerDiscoveryTimestampsTest.lastContactedTimestampUpdatedOnOutboundMessage()` fails intermittently due to a timestamp precision issue. In the unit tests, we are comparing the before and after state of a timestamp to ensure it was updated. When the test runs too quickly, the before and after states appear equal due to a loss of precision.

I tried the following solutions:
1. Fixing the precision issue by using the `org.hyperledger.besu.testutil.TestClock` test fixture.
- This did not solve the problem because the underlying class used in the SUT (`DiscoveryPeer`) stores the timestamps as longs, so we are constrained by the precision of longs.
2. Using Mockito to spy on the peer and assert that the `setLastSeen` and `setLastContacted` setters are called.
- This did not solve the problem because the SUT creates the peer object under the hood, and there is no easy way to inject my spy version in there.

So I landed on a simple, albeit not fancy or elegant, solution of adding a 1 ms `sleep` in the test between the two send message calls to ensure the timestamps will be at least 1 ms different.

## Fixed Issue(s)

Fixes https://github.com/hyperledger/besu/issues/2742

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog). - Since this is "Internal technical improvements with no user impact", I think no changelog update is required.

### DIFF
--- a/ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryTimestampsTest.java
+++ b/ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryTimestampsTest.java
@@ -92,6 +92,12 @@ public class PeerDiscoveryTimestampsTest {
     firstDiscovered.set(fd);
 
     // Send another packet and ensure that timestamps are updated accordingly.
+    // Sleep beforehand to make sure timestamps will be different.
+    try {
+      Thread.sleep(1);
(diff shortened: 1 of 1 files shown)


jflo (approved):
1ms should be enough for this test to pass reliably.

PR #2837, opened 2021-10-01
Added subcommand for validate-config

ms2597:
## Changelog


antonydenyer:
Thanks for PR, you need to sign off your commits for contributions to be accepted. Can you amend your commit accordingly, please?

https://wiki.linuxfoundation.org/dco

macfarla:
Thanks for your contribution @ms2597 - the `assemble` step is failing on some formatting checks. Run './gradlew :spotlessApply' to fix locally and then commit the changes :)

macfarla:
@ms2597 were you still working on this? I've tested it out locally and it still needs some changes
* needs to return after printing an error otherwise it still prints "file is valid" even when it isn't
* the TomlParser it's using is the wrong one (will only check for permissioning config). 
I've made a new PR with some changes to make it work https://github.com/hyperledger/besu/pull/2994

PR #2866, opened 2021-10-08
Leave block p2p messages enabled

ajsutton:
## PR description
To support following the chain prior to TTD being reached, the NEW_BLOCK_HASHES and NEW_BLOCK messages must remain enabled.  

Potentially validation could/should be added to reject an new blocks if their parent already meets the total difficulty requirement for transition but that likely should live in the block import code rather than the networking code. That hasn't been done in this PR.

## Changelog


jflo (approved):
Could also just empty out mergeSupportExcludeMessages, but then the rest of the code to filter becomes dead and YAGNI, so..... i get it.

Issue #2868, opened 2021-10-11
Select correct ValidatorProvider when transitioning from validator contract to block header mode


siladu:
### Description
During the transition from contract mode to block header mode, when creating the extra data for a new block, we need to look ahead to the next block's validatorProvider to obtain the nonEmpty voteProvider associated with the BlockValidatorProvider.

### Acceptance Criteria
**Given** we are transitioning from contract to block header mode
**When** a block’s extra data is created
**Then** it should select the correct validatorProvider and not error upon selecting the TransactionValidatorProvider with an empty voteProvider.

(The following is a more robust AC which gives better asserts in an automated test than the above AC's "ensure it doesn't error")
**Given** we are transitioning from contract to block header mode
**When** a block’s extra data is created
**Then** it should use the same validators as the previous block (which will be the list of validators currently in the contract)

### Steps to Reproduce (Bug)
1. Start a node using a transition config like below:
```
"transitions": {
      "qbft": [
        {
          "block": 2,
          "validatorselectionmode": "contract",
          "validatorcontractaddress": "0x0000000000000000000000000000000000008888"
        },
        {
          "block": 4,
          "validatorselectionmode": "blockheader"
        }
      ]
    }
```
2. Wait for a block to be mined for the transition back to block header mode
3. See the block creation fail due to this error:
```
2021-10-06 21:10:01.007+02:00 | pool-8-thread-1 | ERROR | EventMultiplexer | State machine threw exception while processing event \{BlockTimerExpiry{Round Identifier=ConsensusRoundIdentifier{Sequence=4, Round=0}}\}
java.lang.IllegalStateException: Bft requires a vote provider
	at com.google.common.base.Preconditions.checkState(Preconditions.java:510)
	at org.hyperledger.besu.consensus.common.bft.blockcreation.BftBlockCreatorFactory.createExtraData(BftBlockCreatorFactory.java:108)
	at org.hyperledger.besu.consensus.qbft.blockcreation.QbftBlockCreatorFactory.createExtraData(QbftBlockCreatorFactory.java:93)
	at org.hyperledger.besu.consensus.common.bft.blockcreation.BftBlockCreatorFactory.lambda$create$1(BftBlockCreatorFactory.java:82)
	at org.hyperledger.besu.ethereum.blockcreation.AbstractBlockCreator.createBlock(AbstractBlockCreator.java:182)
[...]

PR #2869, opened 2021-10-11
Fix for Select correct ValidatorProvider when transitioning from validator contract to block header mode

siladu:
## PR description

During the transition from contract mode to block header mode, when creating the extra data for a new block, we need to look ahead to the next block's voteProvider to obtain the nonEmpty one associated with the BlockValidatorProvider.

This only applies to ForkingValidatorProvider which is currently only used for QBFT. However, we don't want QBFT to know it's using a ForkingValidatorProvider. Therefore we need to implement getVoteProviderAfterBlock on the ValidatorProvider interface even though it's implementation is equivalent to getVoteProvider.

Update qbft integration tests to expose this bug.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/2868

## Changelog


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/validator/ValidatorProvider.java
```diff
@@ -28,5 +28,9 @@
 
   Collection<Address> getValidatorsForBlock(BlockHeader header);
 
-  Optional<VoteProvider> getVoteProvider();
+  Optional<VoteProvider> getVoteProviderAtHead();
+
+  default Optional<VoteProvider> getVoteProviderAfterBlock(final BlockHeader header) {
```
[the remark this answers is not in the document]
siladu: The alternative was to make QBFT code use the ForkingValidatorProvider directly instead of the interface which also didn't seem great. On top of that, we'd have to refactor/duplicate common code from BftBlockCreatorFactory.createExtraData into QbftBlockCreatorFactory to account for this special case.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProvider.java
```diff
@@ -65,8 +65,14 @@ public ForkingValidatorProvider(
...
-  public Optional<VoteProvider> getVoteProvider() {
-    return resolveValidatorProvider(blockchain.getChainHeadHeader().getNumber()).getVoteProvider();
+  public Optional<VoteProvider> getVoteProviderAtHead() {
+    return resolveValidatorProvider(blockchain.getChainHeadHeader().getNumber())
+        .getVoteProviderAtHead();
+  }
+
+  @Override
+  public Optional<VoteProvider> getVoteProviderAfterBlock(final BlockHeader header) {
+    return resolveValidatorProvider(header.getNumber() + 1).getVoteProviderAtHead();
```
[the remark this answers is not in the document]
siladu: The resolver would handle this by yielding an empty forkSchedule for that block. However, you've highlighted that this would make for a good unit test so I'll add one :)

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/test/ValidatorContractTest.java
```diff
@@ -121,6 +126,122 @@ public void migratesFromBlockToValidatorContract() {
...
+    final BlockHeader block1 = context.getBlockchain().getBlockHeader(1).get();
+
+    // contract block extra data cannot contain validators or vote
+    assertThat(validatorProvider.getValidatorsForBlock(genesisBlock)).isEqualTo(block0Addresses);
+    assertThat(extraDataCodec.decode(genesisBlock).getValidators()).isEmpty();
+    assertThat(extraDataCodec.decode(genesisBlock).getVote()).isEmpty();
+
+    // uses previous block's validators
+    assertThat(validatorProvider.getValidatorsForBlock(block1)).isEqualTo(block0Addresses);
+    assertThat(extraDataCodec.decode(block1).getValidators()).containsExactly(NODE_ADDRESS);
```
jframe, line 162: Wonder if there is anything we can do to prove that the contract wasn't called to obtain the validators

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProvider.java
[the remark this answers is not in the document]
siladu: Slight clarification...the whole point is that it's looking ahead by one as this is used as part of creating the next block.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProvider.java
```diff
@@ -65,8 +65,14 @@ public ForkingValidatorProvider(
...
   @Override
-  public Optional<VoteProvider> getVoteProvider() {
-    return resolveValidatorProvider(blockchain.getChainHeadHeader().getNumber()).getVoteProvider();
+  public Optional<VoteProvider> getVoteProviderAtHead() {
+    return resolveValidatorProvider(blockchain.getChainHeadHeader().getNumber())
+        .getVoteProviderAtHead();
+  }
+
+  @Override
+  public Optional<VoteProvider> getVoteProviderAfterBlock(final BlockHeader header) {
```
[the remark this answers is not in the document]
siladu: Yep, I noticed this too and clarified in other comment thread 👍

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProviderTest.java
```diff
@@ -170,10 +173,50 @@ public void voteProviderIsDelegatesToHeadFork() {
         new ForkingValidatorProvider(
             blockChain, forksSchedule, blockValidatorProvider, contractValidatorProvider);
 
-    final VoteProvider voteProvider = Mockito.mock(VoteProvider.class);
-    when(contractValidatorProvider.getVoteProvider()).thenReturn(Optional.of(voteProvider));
+    validatorProvider.getVoteProviderAtHead();
```
jframe, line 176: Doesn't look like your test is using the vote provider that is returned

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProviderTest.java
```diff
@@ -170,10 +173,50 @@ public void voteProviderIsDelegatesToHeadFork() {
...
+  @Test
+  public void voteProviderIsDelegatesToHeadFork_whenHeadIsBlockFork() {
+    final BftForksSchedule<QbftConfigOptions> forksSchedule =
+        new BftForksSchedule<>(createBlockFork(0), emptyList());
+
+    final ForkingValidatorProvider validatorProvider =
+        new ForkingValidatorProvider(
+            blockChain, forksSchedule, blockValidatorProvider, contractValidatorProvider);
+
+    validatorProvider.getVoteProviderAtHead();
```
jframe, line 191: Same as above, doesn't look this is needed

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProviderTest.java
```diff
@@ -170,10 +173,50 @@ public void voteProviderIsDelegatesToHeadFork() {
...
+    final ForkingValidatorProvider validatorProvider =
+        new ForkingValidatorProvider(
+            blockChain, forksSchedule, blockValidatorProvider, contractValidatorProvider);
+
+    final VoteProvider voteProviderForBlockValidator = Mockito.mock(VoteProvider.class);
+    when(blockValidatorProvider.getVoteProviderAtHead())
+        .thenReturn(Optional.of(voteProviderForBlockValidator));
+    when(contractValidatorProvider.getVoteProviderAtHead()).thenReturn(Optional.empty());
+
+    SoftAssertions.assertSoftly(
```
jframe, line 212: Why do you need to use SoftAssertions?

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProviderTest.java
siladu: See other comment

jframe (approved):

PR #2871, opened 2021-10-11
add jacoco store artifacts

matkt:

## Changelog


jflo (approved):
What else does this need to move out of draft?

Issue #2795, opened 2021-09-22
Qbft RPCs should be disabled after switching to use validator contract


jframe:
Qbft voting RPCs should be disabled if we switch to using the validator contract with a transition

**Acceptance Criteria**
* After switching from a block header voting mode to validator contract mode the QBFT rpcs should return unimplemented error.
* Later if switching back to block header mode the RPCs should work again
* These should also work on initial startup with RPCs only working for block header voting
* Similarly needs to work for toggle on QBFT config validatorContractAddress field
* The qbft_getSignerMetrics should continue to work in both cases as it affected by voting changes

antonydenyer:
I think `qbft_getSignerMetrics` should remain available after a switch so that you can check who's doing what.

jframe:
Good point that should stay. We only need to disable voting QBFT rpcs.

PR #2881, opened 2021-10-12
Qbft RPCs should read validators from the contract when in contract mode

siladu:
~Created a partial copy of the ForkingValidatorProvider in order to use a read-only version of the BlockValidatorProvider (which contains its own voteTallyCache).~

Create readOnlyValidatorProvider in QbftBesuControllerBuilder and pass into QbftJsonRpcMethods constructor (better than using protocolContext as is only used once when creating the RPCs

This requires storing transactionValidatorProvider in a field and relying on its earlier creation during createConsensusContext()

Fixes https://github.com/hyperledger/besu/issues/2795


besu/src/test/java/org/hyperledger/besu/controller/QbftBesuControllerBuilderTest.java
```diff
@@ -0,0 +1,169 @@
...
+    when(worldStatePreimageStorage.updater())
+        .thenReturn(mock(WorldStatePreimageStorage.Updater.class));
+    when(storageProvider.createWorldStatePreimageStorage()).thenReturn(worldStatePreimageStorage);
+    when(synchronizerConfiguration.getDownloaderParallelism()).thenReturn(1);
+    when(synchronizerConfiguration.getTransactionsParallelism()).thenReturn(1);
+    when(synchronizerConfiguration.getComputationParallelism()).thenReturn(1);
+    when(observableMetricsSystem.createLabelledCounter(
+            any(), anyString(), anyString(), anyString()))
+        .thenReturn(labels -> null);
+    when(synchronizerConfiguration.getBlockPropagationRange()).thenReturn(Range.closed(1L, 2L));
```
siladu, line 119: If I could review my own code...I think there's a question a question of test value vs potential brittleness here. 
If future changes to the BesuControllerBuilder warrant more mocking changes then is it the test worth the hassle?
Those changes would also break BesuControllerBuilderTest, so someone would have to fix that too.

I'm not sure how often BesuControllerBuilder will change in a way that causes this test to fail.

besu/src/test/java/org/hyperledger/besu/controller/QbftBesuControllerBuilderTest.java
```diff
@@ -0,0 +1,169 @@
...
+        besuController
+            .getProtocolContext()
+            .getConsensusState(BftContext.class)
+            .getValidatorProvider();
+    assertThat(validatorProvider).isInstanceOf(ForkingValidatorProvider.class);
+  }
+
+  @Test
+  public void missingTransactionValidatorProviderThrowsError() {
+    ProtocolContext protocolContext = mock(ProtocolContext.class);
```
jframe, line 161: nit: final

besu/src/test/java/org/hyperledger/besu/controller/QbftBesuControllerBuilderTest.java
```diff
@@ -0,0 +1,169 @@
...
+            .transactionPoolConfiguration(poolConfiguration)
+            .nodeKey(nodeKey)
+            .storageProvider(storageProvider)
+            .gasLimitCalculator(gasLimitCalculator)
+            .evmConfiguration(EvmConfiguration.DEFAULT);
+  }
+
+  @Test
+  public void forkingValidatorProviderIsAvailableOnBftContext() {
+    BesuController besuController = qbftBesuControllerBuilder.build();
```
jframe, line 149: nit: final

besu/src/test/java/org/hyperledger/besu/controller/QbftBesuControllerBuilderTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    ValidatorProvider validatorProvider =
+        besuController
+            .getProtocolContext()
+            .getConsensusState(BftContext.class)
+            .getValidatorProvider();
+    assertThat(validatorProvider).isInstanceOf(ForkingValidatorProvider.class);
+  }
+
```
jframe, line 158: Is it possible to add a test that shows that the QbftJsonRpcMethods are actually created with a read only validator provider?

besu/src/test/java/org/hyperledger/besu/controller/QbftBesuControllerBuilderTest.java
jframe: Think overall there is more benefit to having the unit tests. But  like you said they could end up being brittle and also if either of these tests failed then Besu build would fail anyway as Qbft would never start up in the QBFT ATs. But still like the early warning and specific failures from the unit tests.

PR #2895, opened 2021-10-13
Add tracing of the EVM with OpenTelemetry

atoulme:

## PR description
Add OpenTelemetry tracing for the EVM as an experiment.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -164,7 +164,7 @@ protected AbstractBlockProcessor(
               blockHeader,
               transaction,
               miningBeneficiary,
-              OperationTracer.NO_TRACING,
+              OpenTelemetryTracing.INSTANCE,
```
shemnon, line 167: I have performance concerns about hard coding anything but a no-op tracer in this loop.  Can it be made CLI configurable and default to `NO_TRACING`?

atoulme: Yes it can. Let's do it.

atoulme:
FWIW I'm just playing with this - opinions about the value of this approach are very welcome.

PR #2926, opened 2021-10-19
[#2925] Update JDK 11 to latest version in Besu Docker images 

### LINKED ISSUE #2925: Update JDK 11 to latest version in Besu Docker images
### Description

[AdoptOpenJDK project has been superseded by Eclipse Temurin](https://blog.adoptopenjdk.net/2021/03/transition-to-eclipse-an-update/), and it is not more providing updates to JDK docker container images.

Use [Eclipse Temurin images for Besu](https://hub.docker.com/_/eclipse-temurin/) when building Besu images, in order to upgrade to the last JDK version.

fab-10:
AdoptOpenJDK project is now deprecated.

## PR description

Replace AdoptOpenJDK base image for Java 11, with the updated base image provided by Eclipse Temurin.

## Fixed Issue(s)

fixes #2925

## Changelog

### DIFF
--- a/docker/graalvm/Dockerfile
+++ b/docker/graalvm/Dockerfile
@@ -1,5 +1,5 @@
 
-FROM ghcr.io/graalvm/graalvm-ce:ol7-java11-21.0.0
+FROM ghcr.io/graalvm/graalvm-ce:ol7-java11
 ARG VERSION="dev"
 
 RUN adduser --home /opt/besu besu && \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Additions and Improvements
 - Upgrade CircleCI OpenJDK docker image to version 11.0.12. [#2928](https://github.com/hyperledger/besu/pull/2928)
+- Update JDK 11 to latest version in Besu Docker images. [#2925](https://github.com/hyperledger/besu/pull/2925)
 
 ### Bug Fixes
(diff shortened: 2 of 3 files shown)


shemnon (changes requested):
* There are two OpenJDK Images, 11 and latest.  Both should be updated.
* The variant name and file system location still says openjdk, that too should be updated.
* We should make sure that the [build](https://github.com/hyperledger/besu/blob/fdd935262b7fb7e11abdf895dde5a555f950e6b6/build.gradle#L603) tags the correct image as the [default image](https://github.com/hyperledger/besu/blob/fdd935262b7fb7e11abdf895dde5a555f950e6b6/build.gradle#L651).
* While we are updating docker images, we should update graalvm dokcer image to the most current java 11 image.

docker/graalvm/Dockerfile
```diff
@@ -1,5 +1,5 @@
 
-FROM ghcr.io/graalvm/graalvm-ce:ol7-java11-21.0.0
+FROM ghcr.io/graalvm/graalvm-ce:ol7-java11-21.3
```
shemnon, line 2:
```suggestion
FROM ghcr.io/graalvm/graalvm-ce:ol7-java11
```
(Optional) - This will keep us on the most recent Java 11 GraalVM

shemnon (approved):

shemnon (approved):

Issue #2901, opened 2021-10-14
Qbft transitioning to new validator selection mode uses the incorrect validator provider


jframe:
**Description**

When QBFT creates a new block it determines the validator provider that should be used for the new block incorrectly. It is using the validator provider for the current block instead of the next block. This could mean that incorrect set of validators could be used to create the next block.

**Acceptance Criteria**
- Use the validator provider appropriate for the next block returning the validators for the next block in the ForkingValidatorProvider

PR #2935, opened 2021-10-21
Transitions validator provider bug

siladu:
Fixes https://github.com/hyperledger/besu/issues/2901


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProvider.java
```diff
@@ -75,15 +80,15 @@ public ForkingValidatorProvider(
...
+      final BftForkSpec<QbftConfigOptions> forkSpec,
+      final Function<ValidatorProvider, Collection<Address>> getValidators) {
 
     // when moving to a block validator the first block needs to be initialised or created with
     // the previous block state otherwise we would have no validators
-    if (!forkSpec.getConfigOptions().isValidatorContractMode()) {
-      if (block > 0 && block == forkSpec.getBlock()) {
-        final long prevBlockNumber = block - 1L;
+    if (forkSpec.getConfigOptions().isValidatorBlockHeaderMode()) {
+      if (forkSpec.getBlock() > 0) {
```
siladu, line 90: Removing `block == forkSpec.getBlock()` changes the behaviour and breaks a unit test, need to figure out if changing the meaning of the unit test is the right fix.

consensus/qbft/src/test/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProviderTest.java
```diff
@@ -141,7 +141,8 @@ public void migratesFromContractToBlockValidatorProvider() {
     assertThat(validatorProvider.getValidatorsForBlock(genesisHeader))
         .isEqualTo(CONTRACT_ADDRESSES_1);
     assertThat(validatorProvider.getValidatorsForBlock(header1)).isEqualTo(CONTRACT_ADDRESSES_1);
-    assertThat(validatorProvider.getValidatorsForBlock(header2)).isEqualTo(BLOCK_ADDRESSES);
+    // should use previous validatorProvider's validators
+    assertThat(validatorProvider.getValidatorsForBlock(header2)).isEqualTo(CONTRACT_ADDRESSES_1);
```
siladu, line 145: TODO not sure if this is semantically correct - the removal of `block == forkSpec.getBlock()` as pointed out in the other comment is responsible for the difference

config/src/main/java/org/hyperledger/besu/config/QbftConfigOptions.java
```diff
@@ -20,6 +20,8 @@
 
   Optional<String> getValidatorContractAddress();
 
+  boolean hasValidatorOverrides();
```
jframe, line 23: This doesn't really belong as part of QbftConfigOptions as it's not a possible config option for QBFT. Might need to create another type to hold this information

jframe, line 23: This doesn't really belong as part of QbftConfigOptions as it's not a possible config option for QBFT. Might need to create another type to hold this information.

siladu: Have moved logic into BlockValidatorProvider

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
```diff
@@ -532,4 +533,21 @@ private static QbftConfigOptions createGenesisConfig(final boolean useValidatorC
     }
     return qbftConfigOptions;
   }
+
+  private static BftValidatorOverrides convertBftForks(final List<QbftFork> bftForks) {
+    final Map<Long, List<Address>> result = new HashMap<>();
+
+    for (final BftFork fork : bftForks) {
```
jframe, line 540: nit: there is probably a cleaner way to do this using streams. but also may not be worth the effort.

jframe (approved):
LGTM

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
siladu: I actually ripped this straight from https://github.com/hyperledger/besu/blob/a2f517ce32cf8763c858c3e4f051e45c2c56c200/besu/src/main/java/org/hyperledger/besu/controller/QbftBesuControllerBuilder.java#L346-L361

(which now I've looked is actually itself ripped from https://github.com/hyperledger/besu/blob/a2f517ce32cf8763c858c3e4f051e45c2c56c200/besu/src/main/java/org/hyperledger/besu/controller/IbftBesuControllerBuilder.java#L274-L289)

Maybe the original author didn't want too many nested lambdas? Agree it could be neater overall.

I didn't think it was worth removing duplication for the sake of including in the test, but now it's in three places...maybe...thoughts @jframe?

jframe: not worth the effort, think it's fine to stay as is

PR #2944, opened 2021-10-25
[1632] Add discovery options to genesis file

### LINKED ISSUE #1632: Add option to configure discovery in genesis json
### Description
As an administrator, I want to be able add bootnodes and dns discovery to the config json files so that all critical chain information can be setup in one file.

Currently if you have a custom network with custom bootnodes or DNS discovery you need to add that information to the CLI or the config.toml that proxies the CLI.  This feature would move that into the shared json file so that only node specific information (such as RPC ports) must be configured via CLI.  The cli flags should continue to work as well. 

### Acceptance Criteria
The new json section should look something like this, or highly similar

```
{
  "config": {
    // ...
    "discovery" : {
      "bootnodes": [
        "enode://...@0.1.2.3:4567",
        "enode://...@0.1.2.9:8765"
      ],
      "dns": "enrtree://ABASETHIRTYTWOSTRING@besu.your.corp.net"
    },
    // ...
  },
  // ...
}
```

`bootnodes` and `dns` are both optional, which means we may have an empty `discovery` object, which should not produce any errors even if it looks weird.  But if they put anything other than an object in the `discovery` field, such as a string, errors are fine.

diega:

## PR description
With this PR you can add a `discovery` object to the `config` field of the Genesis JSON file like this:
```json
{
  "config": {
    "discovery" : {
      "bootnodes": [
        "enode://...@0.1.2.3:4567",
        "enode://...@0.1.2.9:8765"
      ],
      "dns": "enrtree://ABASETHIRTYTWOSTRING@besu.your.corp.net"
    }
}
```
CLI args take precedence over the values in the Genesis file config

## Fixed Issue(s)
fixes #1632

## Changelog


diega:
@atoulme I added an acceptance test for overriding the _bootnodes_. Do you think an acceptance test for the DNS override is necessary as well? if that's the case, then I would need to mock a DNS Resolver with proper _enrtree_ responses

atoulme:
Bootnodes is enough. This enhancement looks great to me!

atoulme (approved):

PR #2954, opened 2021-10-26
Build arm64 besu docker images 

garyschulte:
## PR description

* Add arm64 docker image building 
* add arch suffix to docker image publishing
* add manifest task to  create a unified docker manifest (arm64 and amd64) image publishing

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,6 @@ executors:
     working_directory: ~/project
     environment:
       architecture: "amd64"
-      platform: "linux/amd64"
       GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.parallel=true -Dorg.gradle.workers.max=2
 
   besu_arm64_executor_med: # 2cpu, 8G ram
@@ -29,7 +28,6 @@ executors:
     working_directory: ~/project
     environment:
       architecture: "arm64"
-      platform: "linux/arm64"
       GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.parallel=true -Dorg.gradle.workers.max=2
 
   besu_executor_xl: # 8cpu, 16G ram
@@ -302,8 +300,7 @@ jobs:
       - prepare
       - attach_workspace:
           at: ~/project
-      - setup_remote_docker:
-          version: 20.10.14
+      - setup_remote_docker
       - run:
           name: hadoLint_openjdk_11
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -387,7 +387,6 @@ jobs:
       - prepare
       - attach_workspace:
           at: ~/project
-      - setup_remote_docker
       - run:
           name: Publish Docker
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,7 @@ executors:
     resource_class: medium
     working_directory: ~/project
     environment:
+      architecture: "amd64"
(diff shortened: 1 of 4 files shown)


ajsutton:
For the record, with Teku I wound up needing to use a machine executor and buildx to create the multi-arch docker image correctly.  I don't think I ever found a way to get an ARM docker executor in CircleCI (I could get an arm container but the remote docker attached to it was still x86).  https://github.com/ConsenSys/teku/pull/4632 is the PR for Teku.

macfarla:
@garyschulte - is this one superseded by https://github.com/hyperledger/besu/pull/3481

.circleci/config.yml
```diff
@@ -6,31 +6,41 @@ orbs:
...
+      - image: cimg/openjdk:11.0.13
     resource_class: small
     working_directory: ~/project
     environment:
       GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.parallel=true -Dorg.gradle.workers.max=2
 
   besu_executor_med: # 2cpu, 4G ram
     docker:
-      - image: cimg/openjdk:11.0
+      - image: cimg/openjdk:11.0.13
```
shemnon, line 17: > <openjdk-version> - The version of OpenJDK to use. This can be a full SemVer point release (such as 11.0.2) or just the minor release (such as 11.0). If you use the minor release tag, it will automatically point to future patch updates as they are released by AdoptOpenJDK & Oracle. \[https://circleci.com/developer/images/image/cimg/openjdk]

Do we really want to lock ourselves into a specific patch release?

.circleci/config.yml
```diff
@@ -18,6 +18,16 @@ executors:
     resource_class: medium
     working_directory: ~/project
     environment:
+      architecture: "amd64"
+      GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.parallel=true -Dorg.gradle.workers.max=2
+
+  besu_arm64_executor_med: # 2cpu, 8G ram
+    machine:
+      image: ubuntu-2004:202101-01
```
fab-10, line 26: could we remove the tag `202101-01` to get always the updated version?

.circleci/config.yml
```diff
@@ -318,6 +328,35 @@ jobs:
...
+            docker run --rm -i hadolint/hadolint < docker/graalvm/Dockerfile
+      - run:
+          name: build image
+          command: |
+            ./gradlew --no-daemon distDocker
+      - run:
+          name: test image
+          command: |
+            mkdir -p docker/reports
+            curl -L https://github.com/aelsabbahy/goss/releases/download/v0.3.9/goss-linux-arm -o ./docker/tests/goss-linux-amd64
```
fab-10, line 357: is the output file `./docker/tests/goss-linux-amd64` named correctly for ARM? or should it be `./docker/tests/goss-linux-arm64`?

garyschulte: This is the executable name that is used in `docker/test.sh`.  I think we could use the 'architecture' envar to name the executable.  IDK how much value there is in that over just changing the executable name to goss-linux 🤔 

It is probably better to err on the side of specificity, or at least not using the wrong arch name for the executable ;)

.circleci/config.yml
garyschulte: The circle-ci docs imply we need to be specific about the image tag:
https://circleci.com/docs/2.0/arm-resources/#overview

fab-10 (approved):
LGTM, a CHANGELOG entry for this will be nice

shemnon (approved):

macfarla (approved):
agree a changelog entry would be good, otherwise LGTM

PR #2957, opened 2021-10-27
Move plugin validation to happen after startPlugins

antonydenyer:
This is required because plugin implementation may need to wait until cli options have been bound to decide if they wish to initialise. 

Signed-off-by: Antony Denyer <git@antonydenyer.co.uk>


macfarla (approved):
LGTM

Issue #2800, opened 2021-09-23
SLF4J warning on startup


yorickdowne:
### Description

During startup, Besu prints a warning from SLF4J.

```
execution_1         | SLF4J: Class path contains multiple SLF4J bindings.
execution_1         | SLF4J: Found binding in [jar:file:/opt/besu/lib/log4j-slf4j-impl-2.14.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
execution_1         | SLF4J: Found binding in [jar:file:/opt/besu/lib/logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
execution_1         | SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
execution_1         | SLF4J: Actual binding is of type [org.apache.logging.slf4j.Log4jLoggerFactory]
```

### Versions (Add all that apply)
* Software version: besu/v21.7.4/linux-x86_64/openjdk-java-17
* Java version: openjdk-java-17
* OS Name & Version: Ubuntu 20.04
Besu built from source

PR #2962, opened 2021-10-27
2936 debug docker build

### LINKED ISSUE #2936: Debug variant of docker builds.
### Description
As a client Operator, I want a variant docker image that includes troubleshooting and debugging tools so I can diagnose issues, observe the system, and submit more detailed bug reports.

### Acceptance Criteria
For at least point releases (maybe RCs too?) we would like a docker image, based on the existing openjdk one, but adds:

- full jdk, including jmd, jtop, jstat etc
- wget, curl, jq
- exposed debug port
- prescriptively enable DEBUG rpc api, debug on 5005, native memory tracking
- a working/updated apt repo

That way I can drop a new image into my existing container management system for use when debugging.

jflo:
## PR description
Adds a debug profile, and updates the latest profile to openjdk-17

## Fixed Issue(s)
fixes #2936 

## Changelog


jflo:
fixes #2800

build.gradle
```diff
@@ -614,7 +614,8 @@ artifactoryPublish {
 def dockerVariants = [
   "openjdk-11",
   "graalvm",
-  "openjdk-latest"
+  "openjdk-latest",
+  "openjdk-11-debug"
```
fab-10, line 618: just an esthetic suggestion, what about moving it just below "openjdk-11" to have a nicer grouping?

ethereum/p2p/build.gradle
```diff
@@ -43,8 +43,12 @@ dependencies {
   implementation 'org.apache.logging.log4j:log4j-api'
   implementation 'org.apache.tuweni:tuweni-bytes'
   implementation 'org.apache.tuweni:tuweni-crypto'
-  implementation 'org.apache.tuweni:tuweni-devp2p'
```
fab-10, line 46: This change seems to be independent, should it go in its own PR?

jflo: It's a good question. I considered removing it, but since it solves for an existing open issue, i left it in and instead linked this PR to that issue. I think this falls under "clean as you go" but I admit that can be a subjective decision.

fab-10 (approved):
LGTM.

Nice to have, is a quick doc, with some examples, on how to interact with the debug container

macfarla (approved):

PR #2969, opened 2021-10-28
Private transaction nonce management improvement: PMT pool

macfarla:
Added a "PMT Pool" to keep track of pending private transactions, so these can be factored into private nonce management.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/ChainHeadPrivateNonceProvider.java
```diff
@@ -21,35 +21,73 @@
...
+        privateWorldStateArchive
+            .get(stateRoot, chainHeadHash)
+            .map(
+                privateWorldState -> {
+                  final Account account = privateWorldState.get(sender);
+                  return account == null ? 0L : account.getNonce();
+                })
+            .orElse(Account.DEFAULT_NONCE);
+
+    if (pendingPrivateNonceFromPmtPool == 0 && stateBasedPrivateNonce == 0) {
```
taccatisid, line 82: Isn't this case redundant and covered by the ones below?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
```diff
@@ -0,0 +1,113 @@
...
+    event
+        .getAddedTransactions()
+        .forEach(
+            tx -> {
+              LOG.debug("removing " + pmtPool.containsKey(tx.getHash()));
+              pmtPool.remove(tx.getHash());
+            });
+  }
+
+  public long getMaxMatchingNonce(final String sender, final String privacyGroupId) {
```
taccatisid, line 34: I think this function would be easier to use when returing an `Optional<Long>`. Right now the cases
- never sent a transaction, hence nonce is 0
- sent a transaction with nonce 0
- sent a transaction with a non-zero nonce but it was confirmed and now we do not know the current nonce

can not be distinguished. Around `nonce==0` this probably leads to wrong behaviour in the caller.

The optional could be used to denote that we do or don't know something about the current nonce.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
```diff
@@ -0,0 +1,113 @@
...
+
+    return pmtPool.values().stream()
+        .filter(
+            pmt ->
+                pmt.getSender().equals(sender)
+                    && pmt.getPrivacyGroupIdBase64().equals(privacyGroupId))
```
taccatisid, line 40: Just for me to know: is the private nonce unique per sender address, or per (sender address, privacyGroup)? From the code I assume the later, just want to double check.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
```diff
@@ -0,0 +1,113 @@
...
+        .forEach(
+            tx -> {
+              LOG.debug("removing " + pmtPool.containsKey(tx.getHash()));
+              pmtPool.remove(tx.getHash());
+            });
+  }
+
+  public long getMaxMatchingNonce(final String sender, final String privacyGroupId) {
+
+    return pmtPool.values().stream()
```
taccatisid, line 36: This performs an *O(n)* scan over all transactions. Are we sure the transaction pool is always small?

It this is an issue, we could always have a second data structure that maps from *(sender, privacy group)* to the *set(transaction hashes)* of those users.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
```diff
@@ -0,0 +1,113 @@
...
+import org.hyperledger.besu.ethereum.chain.BlockAddedEvent;
+import org.hyperledger.besu.ethereum.chain.BlockAddedObserver;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class PmtTransactionPool implements BlockAddedObserver {
```
taccatisid, line 13: Wouldn't we need some kind of persistence for this PmtTransactionPool?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  public void onBlockAdded(final BlockAddedEvent event) {
+    // remove from the map any transactions that went into this block
+    // TODO take into account re-orgs
+    event
+        .getAddedTransactions()
+        .forEach(
+            tx -> {
+              LOG.debug("removing " + pmtPool.containsKey(tx.getHash()));
```
taccatisid, line 29: For the normal transaction pool, when multiple transactions are submitted with the same *(sender, nonce)* then one will supersede the other depending on their relative timing and gas price. I assume this handled the same way for PMT transactions which--due to zero gas price--would always prefer the first submitted transaction.

Thus when a user submits multiple transactions with the same nonce, all but the committed one would stay in this PmtTransactionPool forever.

Shouldn't we instead delete all transactions from the same sender with nonce `<=` the one that was just committed to chain?

This may complicate reinsertion of transaction during re-orgs.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
macfarla: per (sender and privacyGroup)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
macfarla: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/ChainHeadPrivateNonceProvider.java
macfarla: gone

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
```diff
@@ -0,0 +1,128 @@
...
+              pmtPool.remove(tx.getHash());
+            });
+  }
+
+  public Optional<Long> getMaxMatchingNonce(final String sender, final String privacyGroupId) {
+
+    return pmtPool.values().stream()
+        .filter(
+            pmt ->
```
pinges, line 54: these are the trackers, not pmts. This is a bit misleading ...

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/privacy/methods/eea/PluginEeaSendRawTransaction.java
```diff
@@ -45,7 +46,13 @@ public PluginEeaSendRawTransaction(
...
       final PrivacyController privacyController,
       final GasCalculator gasCalculator) {
-    super(transactionPool, privacyIdProvider, privateMarkerTransactionFactory, publicNonceProvider);
+    super(
+        transactionPool,
+        new PmtTransactionPool(),
+        privacyIdProvider,
+        privateMarkerTransactionFactory,
+        publicNonceProvider);
+    // TODO fix this ^^ does plugin privacy need the PmtPool or not?
```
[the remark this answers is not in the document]
macfarla: done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/EeaJsonRpcMethods.java
```diff
@@ -35,16 +36,24 @@
...
   private final TransactionPool transactionPool;
+  private final PmtTransactionPool pmtTransactionPool;
   private final PrivacyParameters privacyParameters;
   private final NonceProvider nonceProvider;
 
   public EeaJsonRpcMethods(
       final BlockchainQueries blockchainQueries,
       final ProtocolSchedule protocolSchedule,
       final TransactionPool transactionPool,
+      final PmtTransactionPool pmtTransactionPool,
```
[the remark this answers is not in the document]
macfarla: done

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -73,6 +75,7 @@
       final Synchronizer synchronizer,
       final SyncState syncState,
       final TransactionPool transactionPool,
+      final PmtTransactionPool pmtTransactionPool,
```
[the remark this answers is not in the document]
macfarla: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/RestrictedDefaultPrivacyController.java
```diff
@@ -123,7 +123,7 @@ public String createPrivateMarkerTransactionPayload(
       final String privacyUserId,
       final Optional<PrivacyGroup> maybePrivacyGroup) {
     try {
-      LOG.trace("Storing private transaction in enclave");
+      LOG.info("Storing private transaction in enclave");
```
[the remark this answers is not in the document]
macfarla: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PmtTransactionPool.java
macfarla: done

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -324,6 +325,10 @@ public BesuController build() {
             miningParameters.getMinTransactionGasPrice(),
             transactionPoolConfiguration);
 
+    final PmtTransactionPool pmtTransactionPool = new PmtTransactionPool();
+    // TODO pass blockchain to the pmtPool and move this line there:
```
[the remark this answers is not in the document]
macfarla: done

Issue #2883, opened 2021-10-12
IPv6 peers cause an exception


yorickdowne:
### Description

IPv6 peers cause an exception, UnsupportedAddressType.

```
execution_1  | 2021-10-12 11:53:39.760+02:00 | vert.x-eventloop-thread-2 | WARN  | PeerDiscoveryAgent | Sending to peer DiscoveryPeer{status=bonding, enode=enode://4cb0b81ff60a8a96cf52c4d82739fc5a10287e60e4af79be04ef3ec882b6732604968189e141ccf2680d29870cde11940c406bd308c623eea13e010305c5e905@[240e:3b2:4e17:d990:292e:5907:1bdb:75c0]:30312, firstDiscovered=1634032393381, lastContacted=0, lastSeen=0} failed, packet: 0x938d964c3353e1fbbefa97676b7b0d5d1d895187292936ac3a42c89a8cbd4ea8e15ce106b88cdf7d85bf7858ae66b76f664b9eca02f71485bd7d6ca5ecf3ac757b72fc1cc535bb54ff3e72022af552c5b4ecaa470d06ddcae7e3254a975ad8b30001eb05cb847f00000182765f82765fd790240e03b24e17d990292e59071bdb75c08276688276688461655b5f0e
execution_1  | java.nio.channels.UnsupportedAddressTypeException
execution_1  |  at java.base/sun.nio.ch.Net.checkAddress(Unknown Source)
execution_1  |  at java.base/sun.nio.ch.DatagramChannelImpl.send(Unknown Source)
execution_1  |  at io.netty.channel.socket.nio.NioDatagramChannel.doWriteMessage(NioDatagramChannel.java:296)
execution_1  |  at io.netty.channel.nio.AbstractNioMessageChannel.doWrite(AbstractNioMessageChannel.java:143)
execution_1  |  at io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:949)
execution_1  |  at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.flush0(AbstractNioChannel.java:354)
execution_1  |  at io.netty.channel.AbstractChannel$AbstractUnsafe.flush(AbstractChannel.java:913)
execution_1  |  at io.netty.channel.DefaultChannelPipeline$HeadContext.flush(DefaultChannelPipeline.java:1372)
[...]

PR #2970, opened 2021-10-28
[2883] log nicer warnings when runtime isn't configured for ipv6

frankisawesome:
## PR description

See attached issue for more detailed description from the user's perspective. This exception is caused by the runtime not supporting ipv6 outbound traffic (docker by default doesn't), and the full stacktrace is not helpful and spammy. This PR makes the log TRACE level so users will need to specifically toggle to see the full trace.

Step 2 is to update our troubleshooting doc in besu-docs to include ipv6 related query.

Step 3 is to potentially update our quick start guide to include instructions on how to turn on ipv6 in docker

Step 4 (maybe) is to rewrite how PeerDiscoveryAgent is used, and maybe verify ipv6 connectivity and filter our ipv6 peers.

Step 2,3,4 and this PR are not dependent on each other.

## Fixed Issue(s)

https://github.com/hyperledger/besu/issues/2883

## Changelog
Catching `UnsupportedAddressTypeException` which are caused by runtime not having ipv6 support and display a nice error message rather than half a page long of unhelpful stacktrace


frankisawesome:
@pinges as discussed, thank you!

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -314,6 +315,11 @@ protected void handleOutgoingPacket(final DiscoveryPeer peer, final Packet packe
                 if (err instanceof SocketException && err.getMessage().contains("unreachable")) {
                   LOG.debug(
                       "Peer {} is unreachable, packet: {}", peer, wrapBuffer(packet.encode()), err);
+                } else if (err instanceof UnsupportedAddressTypeException) {
+                  LOG.warn(
+                      "Sending to peer failed - this is likely due to disabled ipv6 at runtime {}, packet: {}",
```
pinges, line 320: I have still not fully understood when this is happening. Is that because our local machine does not support IPv6, or is it the peer's node that does not support it?
We might want to make that very clear here.

frankisawesome: This only happens in docker when certain ipv6 supports are turned off out of the box - this has nothing to do with the peer node.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -314,6 +315,15 @@ protected void handleOutgoingPacket(final DiscoveryPeer peer, final Packet packe
                 if (err instanceof SocketException && err.getMessage().contains("unreachable")) {
                   LOG.debug(
                       "Peer {} is unreachable, packet: {}", peer, wrapBuffer(packet.encode()), err);
+                } else if (err instanceof UnsupportedAddressTypeException) {
+                  LOG.warn(
+                      "Unsupported address type exception when connecting to peer {}, this is likely due to ipv6 not being enabled at runtime. Enable debug trace to see full stacktrace",
```
jflo, line 320: Should clarify last sentence to read "Set logging level to TRACE to see full stacktrace", because as written it is ambiguous which level will produce the better diagnostics.

jflo (approved):

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
frankisawesome: Yeah I think I was conscious about the length of the explanation, but that would make it lot easier to understand, will add it in!

pinges (approved):
LGTM

PR #2979, opened 2021-10-31
Improve ValidatorContractController result validation

### LINKED ISSUE #2900: Improve ValidatorContractController result validation
### Description

When sending a msg to a smart contract, it is possible to get a Successful response with an empty result (`0x`).

The `ValidatorContractController` logic to read the validator list from a smart contract should never receive an empty result when calling the validators contract. If the response is Successful but has an empty result, it is an error (it is likely that the smart contract has been configured incorrectly, or hasn't been deployed). 

### Acceptance Criteria
* If the response from the validator smart contract is empty ('0x'), the application should log an error to let the operator know that something is wrong with the setup.

### Steps to Reproduce (Bug)
1. Start a QBFT node using smart-contract-based validator list, without deploying the contract

lucassaldanha:
## PR description

Add an extra check to ensure that a successful contract call with an empty result configures an error, as we never expect the validator contract to return an empty list of validators.

This is important because of the way that contract call works. It is possible to call an address that does not have a  contract deployed, and still get a successful call with an empty result.

## Fixed Issue(s)
fixes #2900

## Changelog

### DIFF
--- a/consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ValidatorContractController.java
+++ b/consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ValidatorContractController.java
@@ -99,8 +99,16 @@ public class ValidatorContractController {
   private List<Type> decodeResult(
       final TransactionSimulatorResult result, final Function function) {
     if (result.isSuccessful()) {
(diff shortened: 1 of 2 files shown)


jframe (approved):
LGTM

PR #2993, opened 2021-11-04
Log unused JSON params

### LINKED ISSUE #2705: Log unused JSON-RPC Request Fields
### Description
As part of [PR 2690](https://github.com/hyperledger/besu/pull/2690) we allowed all fields in a JSON-RPC call that are not recognized to be ignored.  This was done to support other libraries that would occasionally send "extra" fields (like "type" in eth_estimateGas).  While this increases compatibility, there is a risk we may be unaware of properties that we may want in the future.

To help this, as a node operator I want unknown properties to be appeneded to the console (at info level), including what the field and value is and what type it was trying to be added to.  The particular call would be nice but may be hard to marshal.

One possible approach is to add `@AnySetter` methods in key places and logging the results.  (for context a stackoverflow question: https://stackoverflow.com/a/31164723/8020)

### Acceptance Criteria
Calling any JSON-RPC with an unexpected field will result in the console log notifying me at ~INFO~ DEBUG level.

macfarla:
Fixes #2705. 

I have tested this locally. eg

```
{
    "jsonrpc": "2.0",
    "method": "eth_estimateGas",
    "params": [
        {
            "from": "0xFE3B557E8Fb62b89F4916B721be55cEb828dBd73",
            "to": "0xdd37f65db31c107f773e82a4f85c693058fef7a9",
            "nonce": 99,
            "value": "0x1"
        },
        "latest"
    ],
    "id": 1
}
```
gives me the expected result, with also this logged:
```
2021-11-04 11:35:17.486+10:00 | vert.x-worker-thread-14 | DEBUG | JsonCallParameter | unknown property - nonce with value - 99 and type - class java.lang.Integer caught during serialization
```

## Changelog


mark-terry (approved):
Change makes sense. Do you think the log message would be clearer if the variable values were enclosed in quotes?

ie: unknown property - "nonce" with value - "99" and type - "Integer" caught during serialization"

PR #2994, opened 2021-11-04
Added validate-config subcommand

### LINKED ISSUE #2759: Ability to validate besu config file / command line
### Description
As a developer, I want to be able to validate my besu config file (or command line) so that I can make sure it is valid without starting besu. 

I think it makes sense to do this as a besu subcommand.

### Acceptance Criteria
* `besu validate-config --config-file=besu.conf`
* `besu validate-config --unknown-option`
Above commands would print out any pico CLI exceptions, but not start besu

I would like to see any errors, or a "success" message, without needing to start besu

Currently I need to start besu with the given config file / command line and if it's valid (besu starts) I kill besu.

macfarla:
Fixes #2759 

I have tested this locally:

```
➜  besu git:(2837) ✗ besu validate-config --config-file ../besu-local-nodes/config/besu/besu1.conf
This toml config file is valid.%
➜  besu git:(2837) ✗ besu validate-config --config-file ../besu-local-nodes/config/besu/besu1.conf
picocli.CommandLine$ParameterException: Unknown option in TOML configuration file: xdata-path%
```

## Changelog


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1255,6 +1260,8 @@ private void addSubCommands(
         RLPSubCommand.COMMAND_NAME, new RLPSubCommand(resultHandler.out(), in));
     commandLine.addSubcommand(
         OperatorSubCommand.COMMAND_NAME, new OperatorSubCommand(resultHandler.out()));
+    commandLine.addSubcommand(
+        ValidateConfigSubCommand.COMMAND_NAME, new ValidateConfigSubCommand(resultHandler.out()));
```
antonydenyer, line 1264: Can you not inject it? 

`new ValidateConfigSubCommand(commandLine, resultHandler.out()));`

besu/src/main/java/org/hyperledger/besu/cli/subcommands/ValidateConfigSubCommand.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.PrintStream;
+import java.nio.file.Path;
+
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.ParentCommand;
+
+@Command(
+    name = COMMAND_NAME,
+    description = "This command provides Besu config validation.",
```
taccatisid, line 33: Just to clarify: this is only syntax validation. We could still use mutually incompatible flags/config values, pass in this verification, and then fail during app startup?

I am not saying we should change it, but I do wonder if we could describe the option more accurately. _This command provides Besu config **syntax** validation._ would not be helpful, though.

macfarla: this.out.print("TOML config file is valid on basic inspection. Further dependencies between related options is checked when Besu starts.");

macfarla: is that better?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
macfarla: yes. yes I can

besu/src/main/java/org/hyperledger/besu/cli/subcommands/ValidateConfigSubCommand.java
taccatisid: Will not win awards for brevity; but yes, it is better.

taccatisid (approved):

pinges (approved):
LGTM

PR #3014, opened 2021-11-08
More BFT tracing log messages

lucassaldanha:
## PR description
- Add a trace log message when `BftProtocolManager` receives a subprotocol message. This is the entry-point for BFT related logic in response to the message.

## Fixed Issue(s)
N/A

## Changelog


jframe (approved):
LGTM

PR #3016, opened 2021-11-08
separate step for sonar

macfarla:

Separate SonarQube into its own step since it uses a restricted context

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,6 +140,13 @@ jobs:
           command: |
             ./gradlew --no-daemon build
       - capture_test_results
+
+  sonarQube:
+    executor: besu_executor_med
+    steps:
+      - prepare
+      - attach_workspace:
+          at: ~/project
       - run:
           name: SonarQube
           no_output_timeout: 30m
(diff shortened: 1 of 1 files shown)


joshuafernandes (approved):
One more change I'd suggest @macfarla - in the workflow piece at the bottom of the page I'd set this to only run on the master branch similar to the dockerPublish job (ie it'll have all the rights to run things in the context)

macfarla:
> One more change I'd suggest @macfarla - in the workflow piece at the bottom of the page I'd set this to only run on the master branch similar to the dockerPublish job (ie it'll have all the rights to run things in the context)

the advantage with that would be that it would always have the permissions to run - the disadvantage would be that the author of the PR may not look at the output

pinges (approved):
LGTM

jflo:
I had issues in the past with trying it as a separate step, when remounting the project, the previous build wasn't there because it had run in a different executor. You might want to test that this generates test coverage correctly.

macfarla:
closing this. tbh it hasn't been a problem since diego became a maintainer.

Issue #3019, opened 2021-11-08
Update VoteTallyCache when transitioning to blockheader validatorselectionmode


siladu:
### Description
Transitioning to BLOCKHEADER mode uses the previous (contract-mode) block's validators but doesn't update the VoteTallyCache. This means subsequent blocks will get an old or possibly empty validator list.

### Acceptance Criteria
* `contract -> blockheader` should not stop mining
* `blockheader(vals1) -> contract (vals2) -> blockheader` should use vals2 from the transition block onwards

### Steps to Reproduce (Bug)
1. Set up a network using CONTRACT mode in the genesis block
2. Transition to BLOCKHEADER mode
3. Transition will work and produce the next block but then it will stop mining

**Expected behavior:**
Mining continues beyond the transition block

**Actual behavior:**
Mining stops the block after the transition

siladu:
This is be superseded by https://github.com/hyperledger/besu/issues/3038

PR #3028, opened 2021-11-09
Update VoteTallyCache when transitioning to blockheader validatorselectionmode

siladu:
TODO:
 - discuss that the cache is overriding validators for parent block (instead of block being produced) - maybe we should only set the cache once the block is successfully created? In reality this shouldn't matter since the parent block should be a contract-mode block and therefore have no cache entry.
 - discuss cache not just getting set once, but every execution of ForkingValidatorProvider.getValidators while producing the transition block.
 - try to recreate bug where network was stopped before node2 imported block, then got InvalidBlock

https://github.com/hyperledger/besu/issues/3019

### DIFF
--- a/consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProvider.java
+++ b/consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProvider.java
@@ -95,8 +95,13 @@ public class ForkingValidatorProvider implements ValidatorProvider {
         final long prevBlockNumber = blockNumber - 1L;
         final Optional<BlockHeader> prevBlockHeader = blockchain.getBlockHeader(prevBlockNumber);
         if (prevBlockHeader.isPresent()) {
-          return resolveValidatorProvider(prevBlockNumber)
-              .getValidatorsForBlock(prevBlockHeader.get());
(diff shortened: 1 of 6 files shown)


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validator/ForkingValidatorProvider.java
```diff
@@ -95,8 +95,13 @@ public ForkingValidatorProvider(
...
         final Optional<BlockHeader> prevBlockHeader = blockchain.getBlockHeader(prevBlockNumber);
         if (prevBlockHeader.isPresent()) {
-          return resolveValidatorProvider(prevBlockNumber)
-              .getValidatorsForBlock(prevBlockHeader.get());
+          final Collection<Address> validatorsForPreviousBlock =
+              resolveValidatorProvider(prevBlockNumber)
+                  .getValidatorsForBlock(prevBlockHeader.get());
+          // Update VoteTallyCache
+          blockValidatorProvider.setValidatorsForBlock(
+              prevBlockHeader.get(), validatorsForPreviousBlock);
```
siladu, line 103: Really, we want to be setting the validatorsForPreviousBlock against the yet to be created block, so it might make sense to update the cache once the block is created/imported

siladu:
Superseded by https://github.com/hyperledger/besu/pull/3043

PR #3032, opened 2021-11-10
improve perf trace

matkt:

## PR description

Optimize the performance of the tracing API by copying the input data only when it is necessary. Tested with some blocks on goerli

## Changelog


gezero (approved):
Looks good to me

PR #3033, opened 2021-11-10
Fix dns discovery issue 

matkt:

## PR description

We are creating but not starting the dns discovery daemon. After this modification the node find peers instantly.

## Changelog

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
@@ -224,8 +224,8 @@ public class DefaultP2PNetwork implements P2PNetwork {
                 }
                 dnsPeers.set(peers);
               });
-      dnsDaemon.start();
     }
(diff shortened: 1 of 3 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -224,6 +224,7 @@ public void start() {
                 }
                 dnsPeers.set(peers);
               });
+      dnsDaemon.start();
```
garyschulte, line 227: :+1:  amazing that this has been missing since it was originally implemented

garyschulte: we should add a test that ensures dnsDaemon is started if dnsDiscoveryURL is present

garyschulte (changes requested):
great find.  I think given the changes in Diego's PR #2944, we should add a test.

garyschulte (approved):
I am approving to unblock the PR, but needs approval from another contributor

macfarla (approved):
LGTM

PR #3039, opened 2021-11-11
Shift Optimization

shemnon:
## PR description

Reduce shift calculations to shifts that may have an actual result.

## Changelog

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/operation/SarOperation.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/operation/SarOperation.java
@@ -44,7 +44,7 @@ public class SarOperation extends AbstractFixedCostOperation {
     } else {
       final int shiftAmountInt = shiftAmount.toInt();
 
-      if (shiftAmountInt >= 256) {
+      if (shiftAmountInt >= 256 || shiftAmountInt < 0) {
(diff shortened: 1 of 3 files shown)


evm/src/main/java/org/hyperledger/besu/evm/operation/SarOperation.java
```diff
@@ -44,7 +44,7 @@ public SarOperation(final GasCalculator gasCalculator) {
     } else {
       final int shiftAmountInt = shiftAmount.toInt();
 
-      if (shiftAmountInt >= 256) {
+      if (shiftAmountInt >= 256 || shiftAmountInt < 0) {
```
macfarla, line 47: what happens if it's == 0

macfarla: also would be great to have a test since this caused an issue

shemnon: The bytes handler handles that efficiently.

macfarla (approved):

evm/src/main/java/org/hyperledger/besu/evm/operation/SarOperation.java
macfarla: 👍

Issue #2766, opened 2021-09-15
Clean up deprecation warnings


antonydenyer:
### Description
Some methods used by Besu are being deprecated and need to be migrated.

There are some easy wins that have simple method name changes e.g. 

https://github.com/hyperledger/besu/blob/main/nat/src/main/java/org/hyperledger/besu/nat/upnp/OkHttpStreamClient.java#L53
https://github.com/hyperledger/besu/blob/main/besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java#L1922

### Steps to Reproduce 
1. Temporarily add `'-Xlint:deprecation'` to `options.compilerArgs` in the project you want to clean up.
2. rebuild and look for warnings!

natashashams:
available to work on still?

natashashams:
@jflo can I take this up (from vGHC)?

macfarla:
👋  Hi @natashashams are you still interested in working on this?

sandrawangyx:
Hi @macfarla is anyone working on this issue atm? I am happy to help out with other issues.

macfarla:
@sandrawangyx would love your help on this one, thanks! Have assigned you.

sandrawangyx:
Thanks @macfarla, I've created a PR for it. It seems like it's going to be a big PR. (https://github.com/hyperledger/besu/pull/3511)
Before I proceed any further, can I get a brief review from the maintainers? just want to make sure that I am on the right track since this is my first contribution to besu.

macfarla:
Added some comments @sandrawangyx - looks pretty good so far!

antonydenyer:
There are a lot of warnings for different things! 
I'd recommend separate prs for each deprecation/concern e.g internal renames, okhttp etc rather than one large pr.

sandrawangyx:
another PR is (https://github.com/hyperledger/besu/pull/3536)

PR #3042, opened 2021-11-11
deprecation: replace calls to isEqualToComparingFieldByField 

macfarla:
with usingRecursiveComparison().isEqualTo

Simple find and replace. 

## Changelog


antonydenyer (approved):
LGTM

https://github.com/hyperledger/besu/issues/2766

shemnon (approved):
LGTM, pending some reversal based on unit and integration tests.

macfarla:
@atoulme Q for you -
the test failures here - a lot of them boil down to this problem - is this something to address in tuweni? I can't really see a nice way around it at the besu end. 
```
field/property 'payload.unsignedPayload.digest.delegate' differ:
- actual value  : 0xb58e0d975ca2e178b705d7c5a5ee243f51b7d8449097746bb93a47f17392ff85 (ArrayWrappingBytes32@f09a29b)
- expected value: 0xb58e0d975ca2e178b705d7c5a5ee243f51b7d8449097746bb93a47f17392ff85 (ByteBufferWrappingBytes32@1afb8449)
org.apache.tuweni.bytes.ArrayWrappingBytes32 can't be compared to org.apache.tuweni.bytes.ByteBufferWrappingBytes32 as ByteBufferWrappingBytes32 does not declare all ArrayWrappingBytes32 fields, it lacks these: [bytes]
```
or this one which I think is the inverse of ^
`org.apache.tuweni.bytes.ByteBufferWrappingBytes32 can't be compared to org.apache.tuweni.bytes.ArrayWrappingBytes32 as ArrayWrappingBytes32 does not declare all ByteBufferWrappingBytes32 fields, it lacks these: [byteBuffer]`

atoulme:
Not sure I can help much here - equality works:
`    assertEquals(new ArrayWrappingBytes32(new byte[32]), new ByteBufferWrappingBytes32(ByteBuffer.allocate(32)));`

These are different implementations of the Bytes object, one with a native byte array, and one with a ByteBuffer backing it. I am not sure what the effort here is, but if you try to compare the fields of the object, it won't work here.

Issue #3038, opened 2021-11-11
Require Validator Overrides For Blockheader Mode Transitions


siladu:
Currently, when we transition from `contract` to `blockheader` mode, we are retrieving the previous block's validators from the contract since there are no validators in the previous block's extra data. Using the previous block's validators isn't necessarily what the user wants to do when they transition, it's just a best guess. They also have the option of specifying a list of `validators` which will override whatever else was going to be used for that block.

When no validator overrides are specified, the above approach currently [suffers from a bug where a stale or empty list of validators are incorrectly used following the blockheader transition](https://github.com/hyperledger/besu/issues/3019). The `VoteTallyCache` is only used with blockheader mode so this causes complexity in repopulating it: we would have to make the cache aware of the previous block's contract validators. This would also impact other BFT code.

This proposed change is to force the configurer to specify `validators` at the same time as transitioning into `blockheader` mode. This means the cache doesn't need populating since the [`ForkingVoteTallyCache` already supports this](https://github.com/hyperledger/besu/blob/00b81b6c4b392d8fbc00576ab66ebfd3da586db7/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/validator/blockbased/ForkingVoteTallyCache.java#L40-L49).

Therefore this transition config would no longer be valid:
```
"transitions": {
  "qbft": [
    {
      "block": 10,
      "validatorselectionmode": "blockheader"
    }
  ]
}
```

If `"validatorselectionmode": "blockheader"` is present at start up we should error, ensuring that a non-empty validator list is also specified, i.e. 

```
"transitions": {
  "qbft": [
    {
      "block": 1,
      "validatorselectionmode": "blockheader",
      "validators": [
        "<0xValidatorAddress1>"
      ]
    }
  ]
}
```

PR #3043, opened 2021-11-11
Require blockheader mode validators

siladu:
A runtime is thrown at startup if this requirement is not met.

A second PR will remove some now redundant complexity in ForkingValidatorProvider

https://github.com/hyperledger/besu/issues/3038


lucassaldanha (approved):
LGTM

jframe (approved):
LGTM

PR #3048, opened 2021-11-12
Create ConsensusScheduleBesuControllerBuilder with no migration capabilities

### LINKED ISSUE #2997: Create a Switchable BesuControllerBuilder without migration capabilities.
This is to create the skeleton switchable BesuControllerBuilder that takes other BesuControllerBuilders and simply to delegate to the first BesuControllerBuilder for now. In the future this will allow switching between BesuControllerBuilders so that we can move between different consensus mechanisms.

**Acceptance Criteria**
* Genesis configuration added to allow for migration consensus mechanisms
* The switchable BesuControllerBuilder is used when the migration config is in use
* Existing consensus mechanism continue to function without any migration happening. i.e. this is just wrapping the BesuControllerBuilder and using the config.

lucassaldanha:
## PR description

Paving the way for the work on IBFT2 to QBFT migration.

- Created `startBlock` option for `QbftConfig`
- Created skeleton `ForkingBesuControllerBuilder`

fixes #2997


besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -191,6 +192,21 @@ BesuControllerBuilder fromGenesisConfig(
           genesisConfig.getConfigOptions(genesisConfigOverrides);
       final BesuControllerBuilder builder;
 
+      // TODO-lucas Too much in-line code?
+      if (configOptions.isQbft() && (configOptions.isIbft2() || configOptions.isIbftLegacy())) {
```
usmansaleem, line 196: may be add a comment mentioning something on the lines of `detect migration from ibft2/ibft to qbft...`

usmansaleem (approved):
LGTM

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
siladu: I don't think it looks too unreadable as it is. However, if we think it's too much I'd vote for an extracted method rather than a comment, e.g. `detectMigrationFromIbftToQbft()`

Although is that comment/name accurate? Are we able to detect that it's _from_ ibft and _to_ qbft?

besu/src/main/java/org/hyperledger/besu/controller/ForkingBesuControllerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+
+/*
+ This is a placeholder class for the QBFT migration logic. For now, all it does is to delegate any
+ BesuControllerBuilder to the first controller in the list.
+*/
+public class ForkingBesuControllerBuilder extends BesuControllerBuilder {
+
+  private static final Logger LOG = LogManager.getLogger();
+
+  private final List<BesuControllerBuilder> besuControllerBuilders = new ArrayList<>();
```
siladu, line 75: Could this be a Set?

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
siladu: Just realised you're talking about the whole if block, not just the condition :)
Still, I don't think think it's too bad. 
One to bear in mind for refactoring nearer the completion of the PR maybe, in case it changes.

siladu (commented):
Made a few in line comments, nothing major.
It's progressing as I expected, no surprises 👍

besu/src/main/java/org/hyperledger/besu/controller/ForkingBesuControllerBuilder.java
lucassaldanha: We could probably simplify it even further for now and keep a single instance of BesuControllerBuilder internally.

besu/src/main/java/org/hyperledger/besu/controller/ForkingBesuControllerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+            ethContext,
+            ethMessages,
+            scheduler,
+            peerValidators);
+  }
+
+  @Override
+  public BesuControllerBuilder storageProvider(final StorageProvider storageProvider) {
+    besuControllerBuilders.get(0).storageProvider(storageProvider);
+    return super.storageProvider(storageProvider);
```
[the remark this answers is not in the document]
lucassaldanha: They are not needed. IntelliJ added them and I forgot to change :)

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -191,6 +192,21 @@ BesuControllerBuilder fromGenesisConfig(
...
+        final BesuControllerBuilder originalControllerBuilder;
+        if (configOptions.isIbft2()) {
+          originalControllerBuilder = new IbftBesuControllerBuilder();
+        } else if (configOptions.isIbftLegacy()) {
+          originalControllerBuilder = new IbftLegacyBesuControllerBuilder();
+        } else {
+          throw new IllegalStateException("Invalid config");
+        }
```
jframe: This is duplicating a lot of the below code. Perhaps which builder to use could be just moved into it's own function.

besu/src/main/java/org/hyperledger/besu/controller/ForkingBesuControllerBuilder.java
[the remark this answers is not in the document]
lucassaldanha: > should it be the delegated builder's job to call super rather than this class

This would only initialise some fields in the delegate's builder class, not on the Forking.

I started removing the super's but then realised that we would endup with a lot of duplicated assignments.

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -191,6 +192,21 @@ BesuControllerBuilder fromGenesisConfig(
...
+        return new ForkingBesuControllerBuilder(
+                Arrays.asList(originalControllerBuilder, new QbftBesuControllerBuilder()))
```
jframe: Does this need to be hardcoded to QbftBesuControllberBuilder?

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
jframe: A seperate function would be nice for this

besu/src/main/java/org/hyperledger/besu/controller/ForkingBesuControllerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+
+import com.google.common.base.Preconditions;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/*
+ This is a placeholder class for the QBFT migration logic. For now, all it does is to delegate any
+ BesuControllerBuilder to the first controller in the list.
+*/
+public class ForkingBesuControllerBuilder extends BesuControllerBuilder {
```
jframe, line 71: Rename to maybe BftMigratingBesuControllerBuilder? Forking really describe what this is doing.

besu/src/main/java/org/hyperledger/besu/controller/ForkingBesuControllerBuilder.java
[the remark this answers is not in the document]
jframe: I had to same problem on my spike branch. This should just return this to satisfy the builder pattern. No need to call super.

config/src/main/java/org/hyperledger/besu/config/JsonQbftConfigOptions.java
```diff
@@ -34,6 +36,11 @@ public JsonQbftConfigOptions(final ObjectNode bftConfigRoot) {
...
   }
 
+  @Override
+  public OptionalLong getStartBlock() {
+    return JsonUtil.getLong(bftConfigRoot, START_BLOCK);
+  }
+
   @Override
   public Map<String, Object> asMap() {
     final Map<String, Object> map = super.asMap();
```
jframe, line 46: Need to add startBlock to the asMap method

config/src/main/java/org/hyperledger/besu/config/QbftConfigOptions.java
```diff
@@ -23,4 +24,8 @@
   default boolean isValidatorContractMode() {
     return getValidatorContractAddress().isPresent();
   }
+
+  default OptionalLong getStartBlock() {
```
jframe, line 28: Don't we think should have a default implementation that does nothing. It only makes sense for the MutableQbftConfigOptions to have an empty value.

config/src/main/java/org/hyperledger/besu/config/JsonQbftConfigOptions.java
```diff
@@ -34,6 +36,11 @@ public JsonQbftConfigOptions(final ObjectNode bftConfigRoot) {
     return JsonUtil.getString(bftConfigRoot, VALIDATOR_CONTRACT_ADDRESS).map(String::toLowerCase);
   }
 
+  @Override
+  public OptionalLong getStartBlock() {
```
jframe, line 40: What does empty startBlock represent? Seems a bit odd and I am wondering if instead we might be better defaulting to zero. That way startBlock has a consistent meaning.

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -206,5 +212,44 @@ BesuControllerBuilder fromGenesisConfig(
...
+      final Map<Long, BesuControllerBuilder> besuControllerBuilderSchedule = new HashMap<>();
+
+      final BesuControllerBuilder originalControllerBuilder;
+      if (configOptions.isIbft2()) {
+        originalControllerBuilder = new IbftBesuControllerBuilder();
+      } else if (configOptions.isIbftLegacy()) {
+        originalControllerBuilder = new IbftLegacyBesuControllerBuilder();
+      } else {
+        throw new IllegalStateException(
+            "Invalid genesis migration config. Migration is supported from IBFT (legacy) ot IBFT2 to QBFT)");
```
jframe, line 227: s/ot/to/

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -206,5 +212,44 @@ BesuControllerBuilder fromGenesisConfig(
...
+          genesisConfig.getConfigOptions().getQbftConfigOptions();
+      final Long qbftBlock = readQbftStartBlockConfig(qbftConfigOptions);
+      besuControllerBuilderSchedule.put(qbftBlock, new QbftBesuControllerBuilder());
+
+      return new ConsensusScheduleBesuControllerBuilder(besuControllerBuilderSchedule)
+          .genesisConfigFile(genesisConfig);
+    }
+
+    private Long readQbftStartBlockConfig(final QbftConfigOptions qbftConfigOptions) {
+      long startBlock =
```
jframe, line 241: nit: final

jframe (approved):

siladu (approved):
Minor comment about possible missing test, but maybe it's covered implicitly elsewhere

PR #3063, opened 2021-11-15
Adding filechannel name to the exception

gezero:

## PR description
When syncing on a disk that was not a SDD I got into problems with the task files. This PR expands exception to include recent file numbers for write/read to give at least some clue on which files are broken.

## Changelog


gezero:
Hmm, I need to figure out how to setup my IDE so that this does not happen again...

PR #3065, opened 2021-11-15
baseFee as Wei instead of long

### LINKED ISSUE #2785: Update basefee implementation to use Uint256 rather than long
Besu's implementation of EIP1559 is using a long instead of Uint256 for baseFee.  This is not likely to be a problem in the near future since baseFee would have to be in excess of 9 eth to cause an overflow, but we should match the types in the spec.

Acceptance criteria:
* update implementation and tests to use Uint256 baseFee

fab-10:
## PR description

this change breaks the compatibility, in the `BlockHeader.getBaseFee()` method but since that specific method is tagged as `@Unstable` I think it is ok to delivery this starting from the next RC
What do you think?

## Fixed Issue(s)

fixes #2785
## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -217,7 +218,7 @@ public void writeTo(final RLPOutput out) {
     out.writeBytes(mixHashOrRandom);
     out.writeLong(nonce);
     if (baseFee != null) {
-      out.writeLongScalar(baseFee);
+      out.writeUInt256Scalar(baseFee);
```
garyschulte, line 221: there appears to be a difference in writeUInt256Scalar and writeLongScalar.   writeLongScalar is always going to encode 8 bytes, whereas UInt256Scalar appears to trim leading zeros...  

This could be an encoding difference that might lead to a problem.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/EthHash.java
```diff
@@ -188,7 +188,7 @@ public static Bytes32 hashHeader(final SealableBlockHeader header) {
     out.writeLongScalar(header.getTimestamp());
     out.writeBytes(header.getExtraData());
     if (header.getBaseFee().isPresent()) {
-      out.writeLongScalar(header.getBaseFee().get());
+      out.writeUInt256Scalar(header.getBaseFee().get());
```
garyschulte, line 191: same concern here about difference in the write scalar implementation.  I am not sure if this could present a consensus problem or not

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/ProofOfWorkValidationRule.java
```diff
@@ -125,7 +125,7 @@ Hash hashHeader(final BlockHeader header) {
     out.writeLongScalar(header.getTimestamp());
     out.writeBytes(header.getExtraData());
     if (imlementsBaseFeeMarket() && header.getBaseFee().isPresent()) {
-      out.writeLongScalar(header.getBaseFee().get());
+      out.writeUInt256Scalar(header.getBaseFee().get());
```
garyschulte, line 128: same q

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionReplacementByFeeMarketRule.java
```diff
@@ -50,40 +48,38 @@ public boolean shouldReplace(
     }
 
     Wei newEffPrice = priceOf(newTransactionInfo.getTransaction(), baseFee);
-    Long newEffPriority =
-        newTransactionInfo.getTransaction().getEffectivePriorityFeePerGas(baseFee);
+    Wei newEffPriority = newTransactionInfo.getTransaction().getEffectivePriorityFeePerGas(baseFee);
 
     // bail early if price is or priority is not strictly positive
-    if (!(newEffPrice.toLong() > 0L)) {
+    if (newEffPrice.compareTo(Wei.ZERO) <= 0) {
```
garyschulte, line 54: Wei is unsigned, so I think this check is superfluous

garyschulte (changes requested):
Looks great and cleans up the code a lot.  My only concern is about the difference in `writeScalar` between long and UInt256.  I don't know for sure whether it will present an encoding or consensus issue, but figured I would highlight it.  Marking with request changes just in case.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionReplacementByFeeMarketRule.java
garyschulte: I guess == 0 is still valid though

ethereum/api/src/integration-test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcResponseUtils.java
```diff
@@ -94,7 +92,8 @@ public JsonRpcResponse response(
     final long gasUsed = unsignedLong(values.get(GAS_USED));
     final long timestamp = unsignedLong(values.get(TIMESTAMP));
     final long nonce = unsignedLong(values.get(NONCE));
-    final Long baseFee = values.containsKey(BASEFEE) ? unsignedLong(values.get(BASEFEE)) : null;
+    final Wei baseFee =
+        values.containsKey(BASEFEE) ? Wei.of(unsignedLong(values.get(BASEFEE))) : null;
```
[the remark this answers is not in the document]
fab-10: good catch! fixed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
fab-10: I double checked and the output is the same, indeed also the javadoc says that both are encoded without leading zero:
[writeLongScalar](https://github.com/fab-10/besu/blob/b8816e6210c33b869146108b8a35814f55344141/ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPOutput.java#L104)
[writeUInt256Scalar](https://github.com/fab-10/besu/blob/b8816e6210c33b869146108b8a35814f55344141/ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPOutput.java#L76)

I have add a[ unit test](https://github.com/fab-10/besu/blob/b8816e6210c33b869146108b8a35814f55344141/ethereum/rlp/src/test/java/org/hyperledger/besu/ethereum/rlp/BytesValueRLPOutputTest.java#L181) to verify that the encoding is the same.

Am I still missing something?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionReplacementByFeeMarketRule.java
fab-10: simplified to only check if the value is 0

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
garyschulte: You are right - both trim leading zeros.  I would swear the writeLongScalar implementation I looked at said it would write exactly 8 bytes.  Looking on main, I see that writeLongScalar is explicitly `writeBytes(Bytes.minimalBytes(v))` 🤷

garyschulte (approved):

PR #3069, opened 2021-11-16
Qbft migration protocol schedule

### LINKED ISSUE #2998: Unified protocol schedule for migrating consensus mechanism
Create a combined protocol schedule with the values from the underlying BesuControllerBuilders. This will allow for a single protocol schedule to be returned by the switable BesuControllerBuilder that can validate all blocks.

**Acceptance Criteria**
* A protocol schedule that contains all the forks all the supplied protocol schedules
* Combined protocol schedule has the correct config and header functions for specified

jframe:
## PR description
Create a combined protocol schedule for the QBFT migration that combines all the milestones from the delegate besu controllers protocol schedules.

fixes #2998  

## Changelog


besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
```diff
@@ -0,0 +1,334 @@
...
+  public ConsensusScheduleBesuControllerBuilder(
+      final Map<Long, BesuControllerBuilder> besuControllerBuilderSchedule) {
+    Preconditions.checkNotNull(besuControllerBuilderSchedule);
+    Preconditions.checkArgument(!besuControllerBuilderSchedule.isEmpty());
+    this.besuControllerBuilders.putAll(besuControllerBuilderSchedule);
+  }
+
+  @Override
+  protected void prepForBuild() {
+    besuControllerBuilders.get(0L).prepForBuild();
```
usmansaleem, line 84: instead of 0L, it might be more readable to use constant BLOCK_ZERO. It took me sometime to figure out that the key in the map is the block number.

usmansaleem (approved):
LGTM

besu/src/test/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilderTest.java
```diff
@@ -0,0 +1,149 @@
...
+  @Mock private BesuControllerBuilder delegateBesuControllerBuilder2;
+
+  @Test
+  public void createsCombinedProtocolScheduleWithMilestonesFromSingleProtocolSchedule() {
+    final StubGenesisConfigOptions genesisConfigOptions = new StubGenesisConfigOptions();
+    genesisConfigOptions.homesteadBlock(5L);
+    genesisConfigOptions.constantinopleBlock(10L);
+    final ProtocolSchedule protocolSchedule = createProtocolSchedule(genesisConfigOptions);
+
+    final Map<Long, BesuControllerBuilder> schedule = new HashMap<>();
```
siladu, line 54: nit: for clarity, renaming this consensusSchedule would help distinguish from protocolSchedule

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
```diff
@@ -0,0 +1,334 @@
...
+            syncState,
+            ethProtocolManager);
+  }
+
+  @Override
+  protected ProtocolSchedule createProtocolSchedule() {
+    final MutableProtocolSchedule combinedProtocolSchedule =
+        new MutableProtocolSchedule(genesisConfig.getConfigOptions().getChainId());
+
+    for (Entry<Long, BesuControllerBuilder> builderSchedule : besuControllerBuilders.entrySet()) {
```
siladu, line 111: Possibly can't guarantee iteration order with the HashMap?
Even if we changed the map impl to e.g. LinkedHashMap, is it still safer to sort by the keys first?

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
jframe: This from another PR. Would be best to address this in the other PR.

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
jframe: Order we iterate in isn't important as long we can determine what spec is before and after. I've changed to the code in this area quite a bit though so my may not make sense with what was there before.

besu/src/test/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilderTest.java
jframe: done

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
```diff
@@ -105,7 +115,46 @@ protected MiningCoordinator createMiningCoordinator(
...
+                    () -> new TreeSet<>(Comparator.comparing(BftForkSpec::getBlock))));
+
+    final MutableProtocolSchedule combinedProtocolSchedule =
+        new MutableProtocolSchedule(genesisConfig.getConfigOptions().getChainId());
+    for (BftForkSpec<ProtocolSchedule> spec : forkSpecs) {
+      checkState(
+          spec.getConfigOptions() instanceof MutableProtocolSchedule,
+          "Consensus migration requires a MutableProtocolSchedule");
+      final MutableProtocolSchedule protocolSchedule =
+          (MutableProtocolSchedule) spec.getConfigOptions();
```
siladu, line 134: Is getConfigOptions() an appropriate method name now that BftForkSpec<C> is more generic?

siladu (commented):
Have way through looking at it, so just adding a comment for now

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
jframe: No probably not, I had the same thought. But didn't want to do any refactoring of that class in this PR as I was expecting a lot of other files to change from a rename. But I'll try it out locally it might not be that bad.

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
```diff
@@ -105,7 +115,46 @@ protected MiningCoordinator createMiningCoordinator(
...
+
+      final Optional<Long> endBlock =
+          Optional.ofNullable(forkSpecs.higher(spec)).map(BftForkSpec::getBlock);
```
siladu, line 137: The use of NavigableSet.higher is neat!

siladu (approved):
LGTM 👍

lucassaldanha (approved):
LGTM. One suggestion would be to move the logic inside `createProtocolSchedule()` into a separate class (`CombinedProtocolSchedule` or something else). This can make it easier to test and also make the code in `ConsensusScheduleBesuControllerBuilder` easier to understand.

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
jframe: There was an additional 10 files added to PR. So will do that refactoring in another PR

lucassaldanha (approved):

PR #3071, opened 2021-11-16
Making sure that the resolved DNS lists get used in full

gezero:

## PR description

Previously we would always process the list of DNS resolved peers from the beginning. We would
at most only try `maxConnections` valid elements from the list and never try more. So the each
time we might be trying the same 25 peers, we connect with them and then drop them. This PR
firstly shuffles the peers list each time before applying it and then also takes from the list for
as long as there are free peers slots available as opposed to guess the limit upfront and hope
that all the peers connect.

## Changelog

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
@@ -55,6 +55,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
@@ -376,6 +377,7 @@ public class DefaultP2PNetwork implements P2PNetwork {
   public Stream<DiscoveryPeer> streamDiscoveredPeers() {
     List<DiscoveryPeer> peers = dnsPeers.get();
     if (peers != null) {
+      Collections.shuffle(peers);
       return Stream.concat(peerDiscoveryAgent.streamDiscoveredPeers(), peers.stream());
     }
     return peerDiscoveryAgent.streamDiscoveredPeers();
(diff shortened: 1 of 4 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -376,6 +377,7 @@ void attemptPeerConnections() {
   public Stream<DiscoveryPeer> streamDiscoveredPeers() {
     List<DiscoveryPeer> peers = dnsPeers.get();
     if (peers != null) {
+      Collections.shuffle(peers);
       return Stream.concat(peerDiscoveryAgent.streamDiscoveredPeers(), peers.stream());
     }
     return peerDiscoveryAgent.streamDiscoveredPeers();
```
matkt, line 383: Maybe we can also shuffle this part ?

gezero:
Unfortunately streams don't have a convenient method to shuffle them. Let me add a unit test to cover the `takewhile`

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
gezero: Hmm, unfortunately streams don't have a convenient method to shuffle the elements.

gezero:
I am going to add a test covering the `takeWhile` part

gezero:
I added a test to check the takeWhile

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
gezero: Shuffling a stream goes against the "lazy" nature of a stream I would say. To shuffle it you would need to process all elements at least once. So you need to collect, shuffle, and convert result to another stream.

jflo (approved):

matkt (approved):
LGTM

PR #3076, opened 2021-11-17
Stream JSON RPC responses to avoid creating big JSON string in memory…

fab-10:
… for large responses

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -809,4 +820,51 @@ private String buildCorsRegexFromConfig() {
...
+        response.setChunked(true);
+        chunked = true;
+      }
+
+      if (response.writeQueueFull()) {
+        LOG.debug("HttpResponse write queue is full pausing streaming");
+        response.drainHandler(e -> paused.release());
+        try {
+          paused.acquire();
+          LOG.debug("HttpResponse write queue is not accepting more data, resuming streaming");
```
frankisawesome, line 852: Is this meant to be '... is now accepting more data' ?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketRequestHandler.java
```diff
@@ -230,4 +254,60 @@ private JsonRpcResponse errorResponse(final Object id, final JsonRpcError error)
...
+      }
+
+      Buffer buf = Buffer.buffer(len);
+      buf.appendBytes(bbuf, off, len);
+      buffer = buf;
+    }
+
+    private void writeFrame(final Buffer buf, final boolean isFinal) throws IOException {
+      if (response.writeQueueFull()) {
+        LOG.debug("HttpResponse write queue is full pausing streaming");
```
frankisawesome, line 288: This should be 'WebsocketReponse...' right

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketRequestHandler.java
```diff
@@ -230,4 +254,60 @@ private JsonRpcResponse errorResponse(final Object id, final JsonRpcError error)
...
+          LOG.debug("HttpResponse write queue is not accepting more data, resuming streaming");
+        } catch (InterruptedException ex) {
+          Thread.currentThread().interrupt();
+          throw new IOException(
+              "Interrupted while waiting for HttpServerResponse to drain the write queue", ex);
+        }
+      }
+
+      if (firstFrame) {
+        response.writeFrame(WebSocketFrame.textFrame(buf.toString(), isFinal));
```
frankisawesome, line 301: Does this work with binary frames as well?

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketRequestHandlerTest.java
```diff
@@ -57,13 +62,17 @@
   private Vertx vertx;
   private WebSocketRequestHandler handler;
   private JsonRpcMethod jsonRpcMethodMock;
+  private ServerWebSocket websocketMock;
   private final Map<String, JsonRpcMethod> methods = new HashMap<>();
 
   @Before
   public void before(final TestContext context) {
     vertx = Vertx.vertx();
```
frankisawesome, line 70: do we still need the vertx instance?

frankisawesome (commented):
A few small comments as I'm wrapping my head around the changes here, will try and do some performance testing.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
fab-10: yes, this is another benefit of streaming the response with Vert.x, since it is possible to implement a back pressure strategy, to avoid sending more data than the client can receive, and so less memory will be used because the JSON fragments are created only when they can be sent to the client, instead of creating all the JSON string upfront.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketRequestHandler.java
fab-10: Right, good catch!

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketRequestHandler.java
fab-10: To send text messages with Vert.x the first frame must have a string as payload, otherwise it is considered a binary message, for this it necessary to do this test on the first frame.
This way we keep backward compatibility and we do not need to think about the encoding, since websocket text messages are UTF-8 encoded.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/WebSocketRequestHandlerTest.java
fab-10: no sure I full get this, but vertx is used in `handler.handle` method invocations

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -822,4 +833,51 @@ private String buildCorsRegexFromConfig() {
       return stringJoiner.toString();
     }
   }
+
+  static class JsonResponseStreamer extends OutputStream {
```
jflo, line 837: It might be worth separating this inner class into it's own file, as JsonRpcHttpService is already over 800 lines.

jflo (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
fab-10: done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonResponseStreamer.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright ConsenSys AG.
```
jflo, line 2: New files should be Copyright Hyperledger Besu Contributors

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/JsonResponseStreamer.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright ConsenSys AG.
```
jflo, line 2:
```suggestion
 * Copyright Hyperledger Besu contributors
```

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonResponseStreamerTest.java
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright ConsenSys AG.
```
jflo, line 2:
```suggestion
 * Copyright Hyperledger Besu contributors
```

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/websocket/JsonResponseStreamerTest.java
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright ConsenSys AG.
```
jflo, line 2:
```suggestion
 * Copyright Hyperledger Besu contributors
```

jflo (approved):
Some minor copyright changes

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonResponseStreamer.java
fab-10: you are right, too fast cut and paste :)

PR #3082, opened 2021-11-17
Fix admin_nodeInfo when response has jdk8 optional values

lucassaldanha:
## PR description

We were not handling JDK8 (de)serialization properly in our RPC services. This PR is using jackson-jdk8 module to handle Optional (de)serialization for responses.

This will ensure that serialization maps the value of Optional to null, if no value is present, or the wrapped value if present.

It does not make sense for the QBFT config to return null for the validator address and start block, so we are not including it into the json when these values are not present.

## Fixed Issue(s)
N/A

## Changelog


jframe (approved):
LGTM

Issue #2999, opened 2021-11-04
Switchable mining co-ordinator


jframe:
Switchable mining coordinator that will switch between mining co-ordinators based on the consensus mechanism forks

**Acceptance Criteria**
* Can switch between mining coordinators

PR #3098, opened 2021-11-23
Migrating mining coordinator duplicate event

siladu:
Split out from https://github.com/hyperledger/besu/pull/3097 as it's a non-breaking issue. (But still part of https://github.com/hyperledger/besu/issues/2999)

The issue is that both MigratingMiningCoordinator and its delegate, BftMiningCoordinator are registered as BlockAddedObservers when the mining coordinator is started. 
This means a duplicate event is processed:
```
DEBUG | BaseBftController | New chain head detected (block number=1), currently mining on top of 0.
DEBUG | BaseBftController | New chain head detected (block number=1), currently mining on top of 1.
TRACE | BaseBftController | Discarding duplicate NewChainHead event. chainHeight=1 
```

It is correctly discarded so it's not a blocker, but it should still be fixed.

I decided to cover this only using unit tests.
I considered the qbft integration-tests, but they avoid executing the mining coordinator, instead requiring the test code to 'manually' do the coordination.
This fix also seemed too fined grained for an acceptance test.


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/BftMiningCoordinator.java
```diff
@@ -84,19 +85,29 @@ public void start() {
     }
   }
 
+  void removeObserver() {
```
siladu, line 88: Does this make sense to go in the BlockAddedObserver interface?

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/SchedulableMiningCoordinator.java
```diff
@@ -0,0 +1,145 @@
...
+    this.blockchain = blockchain;
+    this.activeMiningCoordinator =
+        this.miningCoordinatorSchedule.getFork(blockchain.getChainHeadBlockNumber()).getValue();
+  }
+
+  @Override
+  public void start() {
+    blockAddedObserverId = blockchain.observeBlockAdded(this);
+    activeMiningCoordinator.start();
+    if (activeMiningCoordinator instanceof BftMiningCoordinator) {
```
jframe, line 59: Will have debug through to understand this I think. But I don't see why the MiningCoordinator stop method that calls `blockchain.removeObserver` isn't sufficient. As we always call stop on the current miningCoordinator before calling start on the new one.

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/MigratingMiningCoordinator.java
```diff
@@ -118,13 +125,13 @@ public void onBlockAdded(final BlockAddedEvent event) {
...
           currentBlock,
           activeMiningCoordinator.getClass().getSimpleName(),
           nextMiningCoordinator.getClass().getSimpleName());
       activeMiningCoordinator.stop();
-      nextMiningCoordinator.start();
       activeMiningCoordinator = nextMiningCoordinator;
+      startActiveMiningCoordinator();
     }
     if (activeMiningCoordinator instanceof BlockAddedObserver) {
       ((BlockAddedObserver) activeMiningCoordinator).onBlockAdded(event);
```
jframe, line 137: From my local testing I don't think we need the trigger the onBlockAdded manually. The delegated miningCoordinators are able to subscribe to the onBlockAdded and appear to be working without this.

So if we remove propagating this event then the need to remove the observer will also not be needed.

jframe: After some discussion on this, this approach of not removing the observer from the delegate mining coordinator and relying on it's on block added subscription has some pitfalls. The one being we don't know which observer will be called first and if there will be a delay between the two. It could be possible the previous mining co-ordinator is able to send a proposal even.

So don't think this is a good idea afterall.

jframe (approved):

Issue #2897, opened 2021-10-13
Post-Interop Merge Epic


garyschulte:
### Description
Epic to track merge topics, post-interop

- [ ] [Kintsugi v2](https://github.com/hyperledger/besu/issues/3006)

Issue #3081, opened 2021-11-17
Kintsugi v2: implement RANDOM


garyschulte:
### Description
Besu added `random` to the blockHeader in amphora, but disabled it due to it being out of scope.  For the kintsugi v2 spec, we fully implement EIP-4399:
 - use repurpose/replace mixHash with random post-TTD
   - in the block hash
   - in the RLP encoding (save on 32 bytes)
   - random should be persisted in the db
     - we should drop the unused blockheader fields from the db-serialization post-TTD, but this is an optimization
 - change to the EVM opcode 0x44(DIFFICULTY):
   - rename to RANDOM
   - return the value of random rather than difficulty(which will be fixed at 0)  

### Acceptance Criteria
* implemented in `merge-branch`
* passing merge-fuzz for v2
* passing evm reference tests

PR #3104, opened 2021-11-24
merge of Blockheader changes to support random field

garyschulte:

## PR description
In support of [EIP-4399](https://github.com/ethereum/EIPs/blob/979cf123558bab61e2278f3c962c9a68d1adc2f4/EIPS/eip-4399.md) this PR adds a field to block headers, `random` which is 32 bytes of randao randomness data.

In accordance with spec, we re-purpose mixHash as random in RLP encoding and decoding, since they should be mutually exclusive -  a block with `random`  will not have a `mixHash` and vice-versa

related to #2897 

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
@@ -26,6 +26,7 @@ import java.util.function.Supplier;
 
 import com.google.common.base.Suppliers;
 import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
 
 /** A mined Ethereum block header. */
 public class BlockHeader extends SealableBlockHeader
@@ -34,8 +35,6 @@ public class BlockHeader extends SealableBlockHeader
   public static final int MAX_EXTRA_DATA_BYTES = 32;
 
   public static final long GENESIS_BLOCK_NUMBER = 0L;
-
-  private final Hash mixHash;
 
   private final long nonce;
 
@@ -60,7 +59,7 @@ public class BlockHeader extends SealableBlockHeader
       final long timestamp,
       final Bytes extraData,
       final Long baseFee,
-      final Hash mixHash,
+      final Bytes32 mixHashOrRandom,
       final long nonce,
       final BlockHeaderFunctions blockHeaderFunctions,
       final Optional<LogsBloomFilter> privateLogsBloom) {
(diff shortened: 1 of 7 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -242,6 +254,7 @@ public static BlockHeader readFrom(
     final Hash mixHash = Hash.wrap(input.readBytes32());
     final long nonce = input.readLong();
     final Long baseFee = !input.isEndOfCurrentList() ? input.readLongScalar() : null;
+    final Bytes32 random = !input.isEndOfCurrentList() ? input.readBytes32() : null;
```
matkt, line 257: Maybe I missed something but there is a mismatch between readFrom and writeTo.
- In WriteTo : we write Random or MixHash + we write it before the nonce
- In ReadTo : MixHash is not optional and we read random after the nonce and the baseFee

matkt (changes requested):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
garyschulte: Good catch.  Prior to amphora we (besu) were just appending random at the end, similar to baseFee.  EIP-4399 now says we need to reuse mixhash in order to save 32 bytes.  It isn't clear to me yet how we can differentiate between random and mixhash when we read blocks off the wire or from the db.

This implementation isn't ever going to read random, so we might as well remove the line until we fully address EIP-4399 in #3081

garyschulte: I think we can repurpose mixHash as a generic Bytes32, and get rid of the explicit `random` constructor parameter.  Both random and mixhash would use the same storage and we differentiate between the two via the accessors.  It is a bit hacky imo, but the spec doesn't leave any other options I can see.

matkt (approved):
the new commits fix the problem that I raised. It can be argued that a field does two different things. I approve but I prefer to have the opinion of an additional reviewer regarding this strategy

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockHeader.java
```diff
@@ -164,8 +163,16 @@
    *
    * @return TheBASEFEE of this header.
    */
-  @Unstable
```
shemnon, line 167: Keep the unstable tag.  Another PR is coming in to change this.

shemnon (approved):
Approved contingent on keeping the. `@Unstable` tag on basefee.

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockHeader.java
garyschulte: done 👍

Issue #474, opened 2020-03-16
Allow configuration of TLS protocol and ciphersuites


magooster:
### Description
As an Enterprise, I want to be able to control TLS protocols and ciphersuites used by a system so that I can meet my enterprise requirements. 

### Acceptance Criteria
* Operator can configure supported TLS protocols, default should be "TLSv1.2"
* Operator can configure supported TLS ciphersuites, default should be empty. (Which means the default for TLSv1.2 from the SSLEngine (JDK))

### Additional Information
Example [implementation](https://github.com/magooster/besu/tree/tls-configuration)

arash009:
Thank you for suggesting this feature. We’ve added it to our backlog but likely won’t be working on it in the near term.
We’ll reach out if and when we decide to implement this, but if you have any questions in the meantime, or would like help from the Besu community to get a PR started, you can reach out [here](https://chat.hyperledger.org/channel/besu)

jframe:
Hi @magooster, Noticed on the example implementation you mentioned you have most of the code implemented for this feature. Do you mind creating a PR for this?

gezero:
I would like to work on this

gezero:
I would like to work on this.

PR #3105, opened 2021-11-24
Tls configuration

gezero:
## PR description

Adding two new options:
* `--rpc-http-tls-protocols` to be able to specify TLS protocol version 
* `--rpc-http-tls-ciphersuites` to be able to specify Java cypher suites

## Fixed Issue(s)

Fixes [#474](https://github.com/hyperledger/besu/issues/474): 

## Changelog


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1974,12 +1996,28 @@ private void checkPrivacyTlsOptionsDependencies() {
           "Known-clients file must be specified or CA clients must be enabled when TLS client authentication is enabled for JSON-RPC HTTP endpoint");
     }
 
+    rpcHttpTlsProtocols.retainAll(getJDKEnabledProtocols());
+    if (rpcHttpTlsProtocols.isEmpty()) {
+      throw new ParameterException(commandLine, "No valid TLS protocols specified");
```
fab-10, line 2001: Could be useful to add the list of allowed protocols to the message

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1974,12 +1996,28 @@ private void checkPrivacyTlsOptionsDependencies() {
...
+    }
+
+    for (String ciphersuite : rpcHttpTlsCipherSuites) {
```
fab-10, line 2004: why the validations of the cipherSuites is implemented differently from the tlsProtocol?

fab-10:
You should add the `documentation` label to this PR and the related ticket, so that the doc team could update the Besu command line documentation

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
gezero: Cool, I will add that

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
gezero: There are only few protocols available (for instance on my jdk ["TLSv1.3", "TLSv1.2", "SSLv2Hello"]), their names are short and clear and you would only only diverge from default in legacy systems I would think. It makes sense to me to just show 1 error telling you that you did not specify any supported protocol I will update the error to show user which protocols are supported.


On the other hand when specifying the suite it requires precision so it makes sense to tell the user about each mistake explicitly. There are 48 enabled suites on my jdk. 

If you think I should remake it some other way, let me know 👍

fab-10: thanks for clarify, now it make sense the different approach

fab-10 (approved):
LGTM

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -638,6 +642,20 @@ void setBannedNodeIds(final List<String> values) {
           "Enable to accept clients certificate signed by a valid CA for client authentication (default: ${DEFAULT-VALUE})")
   private final Boolean isRpcHttpTlsCAClientsEnabled = false;
 
+  @Option(
+      names = {"--rpc-http-tls-protocol", "--rpc-http-tls-protocols"},
+      description = "Comma separated list of TLS protocols to support (default: TLSv1.2)",
```
pinges, line 647: TLS 1.3 should be (at least part of) the default. It gives better performance and is more secure and should be well supported by now. I guess first preference should be 1.3 and if we are worried about 1.3 support we could add 1.2.

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -2321,6 +2322,142 @@ public void rpcHttpTlsClientAuthWithCAClientAndKnownClientFile() {
...
+
+  @Test
+  public void rpcHttpTlsCheckValidProtocolsAndCipherSuites() {
+    final String host = "1.2.3.4";
+    final int port = 1234;
+    final String keystoreFile = "/tmp/test.p12";
+    final String keystorePasswordFile = "/tmp/test.txt";
+    final String protocols = "TLSv1.3,TLSv1.2";
+    final String cipherSuites =
+        "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256";
```
pinges, line 2423: We should add a 1.3 cipher suite here, e.g. TLS_AES_256_GCM_SHA384

pinges (changes requested):

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
gezero: You are right, updated.

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
gezero: Updated

pinges (approved):
LGTM

Issue #2965, opened 2021-10-27
Merge: json-rpc endpoints and parameters


garyschulte:
### Description
Part of an effort to iteratively merge the amphora hackathon feature branch into main

* merge the json-rpc parameters, types and endpoints into main

PR #3107, opened 2021-11-25
Merge: engine_getPayload json-rpc endpoint

garyschulte:
## PR description
Merge json-rpc endpoint for `engine_getPayload` from the merge feature branch.

Addresses part of #2965 

## Changelog

- [ X] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).


ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockHeaderTestFixture.java
```diff
@@ -130,6 +133,11 @@ public BlockHeaderTestFixture baseFeePerGas(final long baseFee) {
     return this;
   }
 
+  public BlockHeaderTestFixture random(final Bytes32 random) {
+    this.random = Optional.of(random);
```
matkt, line 137: Just in case. Are we sure that this field will never be null ?

garyschulte: Yeah, `.ofNullable` would ensure tests that send a null random or tests that clone a blockheader with a null random don't throw.  I'll make it so

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeContext.java
```diff
@@ -0,0 +1,57 @@
...
+
+  MergeContext setTerminalTotalDifficulty(final Difficulty newTerminalTotalDifficulty);
+
+  void setIsPostMerge(final Difficulty totalDifficulty);
+
+  boolean isPostMerge();
+
+  boolean isSyncing();
+
+  void observeNewIsPostMergeState(final NewMergeStateCallback newMergeStateCallback);
```
jflo, line 38: Tiny opinion: I woulda used a Future here, but then again I've been looking at a lot of Vert.x code :)

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/PayloadIdentifier.java
```diff
@@ -0,0 +1,88 @@
...
+  @JsonCreator
+  public PayloadIdentifier(final String payloadId) {
+    this(Long.decode(payloadId));
+  }
+
+  public PayloadIdentifier(final Long payloadId) {
+    this.val = UInt64.valueOf(payloadId);
+  }
+
+  public static PayloadIdentifier random() {
```
jflo, line 40: Seems like this is only used for testing, and could possibly contribute to flakey tests. Any reason not to choose static values for testing instead of generating newly random ones each time?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright ConsenSys AG.
```
jflo, line 2: Hyperledger Besu Contributors

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
```diff
@@ -0,0 +1,90 @@
...
+    VALID,
+    INVALID,
+    SYNCING;
+  }
+
+  public enum ForkChoiceStatus {
+    SUCCESS,
+    SYNCING;
+
+    public boolean equalsIgnoreCase(final String status) {
```
jflo, line 42: nice, very handy

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayload.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright ConsenSys AG.
```
jflo, line 2: Hyperledger Besu Contributors

jflo (changes requested):
Can we strike the static random() method on PayloadIdentifier? it feels really out of place, and seems to be only used by the unit test. Even in that test, it doesn't seem to be necessary, we could accomplish the same thing with an entropy source in the test instead of the class.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/PayloadIdentifier.java
garyschulte: The payload identifier is used by the consensus client to fetch block payloads.  Payloads are not generated in this PR however, that is why it looks like a test concern.  I don't think there is any risk in having a deterministic payload identifier, so I think we can use a payload header hash instead and that should be sufficiently unique.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeContext.java
garyschulte: I could see that in a general sense given the method name.  The only particular usage for this currently is as a callback to enable/disable mining though.  We don't benefit from asynchrony in that case, in fact I think it makes things harder to reason about.  Perhaps this method could be renamed to reflect its very specific one-time use case of disabling mining during the pre/post-merge transition.

jflo (approved):

PR #3108, opened 2021-11-26
Rename BftForksSchedule to ForksSchedule

siladu:
Preparation for https://github.com/hyperledger/besu/pull/3097

* Rename BftForksSchedule to ForkSchedule
* Make ForksSchedule generic beyond BftConfigOptions to allow it to work for MiningCoordinator
* Rename static factory to make it specific to BftConfigOptions
* Move ForksSchedule to common package


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/ForksSchedule.java
```diff
@@ -42,13 +41,14 @@
...
-  public BftForksSchedule(final ForkSpec<C> genesisFork, final Collection<ForkSpec<C>> forks) {
+  public ForksSchedule(final ForkSpec<C> genesisFork, final Collection<ForkSpec<C>> forks) {
     this.forks.add(genesisFork);
     this.forks.addAll(forks);
   }
 
-  public static <T extends BftConfigOptions, U extends BftFork> BftForksSchedule<T> create(
-      final T initial, final List<U> forks, final BftSpecCreator<T, U> specCreator) {
+  public static <T extends BftConfigOptions, U extends BftFork>
+      ForksSchedule<T> createForBftConfigOptions(
```
siladu, line 50: I'm thinking this should either be made generic or be moved out to another class?

jframe: This really quite specific to bft transitions. So think it should be out to a Bft specific class

siladu: done!

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/ForksSchedule.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class ForksSchedule<C> {
+
+  private final NavigableSet<ForkSpec<C>> forks =
+      new TreeSet<>(
+          Comparator.comparing((Function<ForkSpec<C>, Long>) ForkSpec::getBlock).reversed());
+
+  @VisibleForTesting
```
jframe, line 33: Should remove this, it's not just for testing now

siladu: Good spot 👍

jframe (approved):

PR #3111, opened 2021-11-28
Eip 4399

jflo:
## PR description

Implements EIP-4399 by repurposing the DIFFICULTY opcode to provide entropy provided by beacon chain post merge.

This is intended to change the behavior during the merge, and leaving the semantic changes to clarify this behavior till after the merge, when the opcode can be documented as renamed to RANDOM, thought its instruction byte remains the same.

## Changelog


garyschulte (commented):
I was expecting a new RandomOperation with the same opcode and/or changes to DifficultyOperation to[rename it](https://github.com/ethereum/EIPs/blob/979cf123558bab61e2278f3c962c9a68d1adc2f4/EIPS/eip-4399.md#renaming)

I don't have a great suggestion on how to do this renaming on the basis of a non-protocolSpec change.  

I don't think there is a functional problem with not renaming or aliasing the operation, other than that traces might not show the 'correct' operation name.  

If indeed there is no functional difference between the op name DIFFICULTY and RANDOM, we could make the name change part of the EIP-4488 hardfork 🤔

evm/src/main/java/org/hyperledger/besu/evm/operation/DifficultyOperation.java
```diff
@@ -27,8 +29,12 @@ public DifficultyOperation(final GasCalculator gasCalculator) {
   @Override
   public Operation.OperationResult executeFixedCostOperation(
       final MessageFrame frame, final EVM evm) {
-    frame.pushStackItem(frame.getBlockValues().getDifficultyBytes());
-
+    if (frame.getBlockValues().getDifficultyBytes() == null
```
shemnon, line 32: Don't make this operation do two different things.  Make a separate `RandomOperation` and have the hard fork spec drop it in.

shemnon (changes requested):

evm/src/main/java/org/hyperledger/besu/evm/operation/DifficultyOperation.java
jflo: I wanted to do that, but since they are the same byte, didn't think it was possible. If you're saying that dropping it in after Difficulty will override it, then that is much cleaner. Will do.

jflo: Ah now I remember. I didn't do that either because this isn't driven by a protocol schedule. This happens when difficulty becomes zero on the merge. See the definition of `TERMINAL_BLOCK` here https://eips.ethereum.org/EIPS/eip-3675#definitions

jflo: Ok, so what we could do, is drop in RANDOM to overwrite DIFFICULTY for 0x44, at some hardfork block before the merge. The conditional has to stay in RANDOM though, since it shouldn't change behavior till difficulty value disappears from block. Then we clean it up in the first release after the merge.

garyschulte: I think this is the most viable option, given the merge transition mechanism.  If an EIP-4488 hardfork does not happen pre-merge, we could add it to the soft fork that is (was?) slated to happen a couple weeks prior to the merge which adds the target terminal total difficulty.

evm/src/main/java/org/hyperledger/besu/evm/operation/DifficultyOperation.java
```diff
@@ -18,6 +18,8 @@
 import org.hyperledger.besu.evm.frame.MessageFrame;
 import org.hyperledger.besu.evm.gascalculator.GasCalculator;
 
+import org.apache.tuweni.bytes.Bytes;
```
garyschulte, line 21: unused import?

evm/src/main/java/org/hyperledger/besu/evm/operation/RandomOperation.java
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.evm.operation;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.hyperledger.besu.evm.EVM;
+import org.hyperledger.besu.evm.frame.MessageFrame;
+import org.hyperledger.besu.evm.gascalculator.GasCalculator;
+
+public class RandomOperation extends AbstractFixedCostOperation {
```
garyschulte, line 22: 👍

evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
```diff
@@ -367,4 +368,31 @@ public static void registerLondonOperations(
     registerIstanbulOperations(registry, gasCalculator, chainId);
     registry.put(new BaseFeeOperation(gasCalculator));
   }
+
+  public static EVM eip4399(final BigInteger chainId, final EvmConfiguration evmConfiguration) {
```
garyschulte, line 372: LGTM, perhaps once the EIP-4488 hardfork has a name, we can rename this EVM

garyschulte (approved):
LGTM, as long as spotless doesn't complain about the unused import

besu/src/main/java/org/hyperledger/besu/cli/config/NetworkName.java
```diff
@@ -21,6 +21,7 @@
   SEPOLIA,
   GOERLI,
   DEV,
+  PREMERGE,
```
shemnon, line 24: I don't think this is the type of network we should be putting in our named networks.

shemnon (changes requested):
I'm not liking putting a test network in the list of named networks.  Testing would be better served by using the `--genesis-file=preMerge.json` CLI flag.

config/src/main/resources/preMerge.json
```diff
@@ -0,0 +1,35 @@
+{
```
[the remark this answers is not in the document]
jflo: moved this to test

shemnon (approved):
Is philly an official name or a filly-in name? ;)

PR #3112, opened 2021-11-29
Include invalid value in Enode error message

ajsutton:
## PR description
Makes it much easier for users to identify their mistake if the CLI argument after `--bootnodes <some-valid-enode>` has a typo. PicoCLI will treat any unrecognised option as another bootnode so it's very confusing to get an invalid enode message when the enode is fine and it's the next argument that's incorrect.

For example `--bootnodes enode://blah@127.0.0.1:30303 --datapath foo` would otherwise print an error message saying there wasn't an `enode://` prefix when there clearly is.  When the invalid value is shown the user can see that it's actually `--datapath` that's unrecognised and fix the problem.

## Changelog


macfarla (approved):
LGTM except spotless

PR #3118, opened 2021-11-30
Start external services before we start plugins

antonydenyer:
## PR description - start external services before we start plugins

Reasons:

- you may wish to query the current state of your blockchain using rpc before you start processing new blocks

- you may wish to register a websocket e.g to listen for blocks currently, you will probably miss one because we start processing before we start the ws service

To enable this a 'beforeExternalServices' hook was addedd to allow plugins to do things before those services are started. 

Signed-off-by: Antony Denyer <git@antonydenyer.co.uk>

## Changelog


macfarla (commented):
LGTM

besu/src/main/java/org/hyperledger/besu/services/BesuPluginContextImpl.java
```diff
@@ -126,13 +128,43 @@ private void addPluginVersion(final BesuPlugin plugin) {
...
+
+      try {
+        plugin.beforeExternalServices();
+        LOG.debug(
+            "beforeExternalServices called on plugin of type {}.", plugin.getClass().getName());
+      } catch (final Exception e) {
+        LOG.error(
+            "Error calling `beforeExternalServices` on plugin of type "
+                + plugin.getClass().getName()
+                + ", stop will not be called.",
```
macfarla, line 151: what does it mean "stop will not be called"

besu/src/main/java/org/hyperledger/besu/Runner.java
```diff
@@ -103,31 +103,35 @@
             cacher -> new AutoTransactionLogBloomCachingService(blockchain, cacher));
   }
 
-  public void start() {
+  public void startExternalServices() {
+    LOG.info("Starting Ethereum services ... ");
```
macfarla, line 107: do we want to say "Ethereum external services" here

besu/src/main/java/org/hyperledger/besu/services/BesuPluginContextImpl.java
antonydenyer: The plugin doesn't get added to the list of running plugins, so we don't call stop when we shutdown besu. What we're saying is you need to take care of cleaning up your own resources.

macfarla (approved):

PR #3123, opened 2021-12-02
Move discovery options constants to genesis files of predefined networks

### LINKED ISSUE #3113: Move bootnodes and DNS discovery to genesis file of predefined networks
### Description
With the introduction of [discovery options in the genesis file](https://github.com/hyperledger/besu/pull/2944), most of the constants in `org.hyperledger.besu.ethereum.p2p.config.DiscoveryConfiguration` can be moved into its corresponding genesis file.

### Acceptance Criteria
Resolve the genesis content file early in the `org.hyperledger.besu.cli.BesuCommand` based on the `--network` option and then continue execution in the same way than with an external genesis file.

### Additional Information
The `networkId` will still be hardcoded for each predefined network, but that can be addressed in a future pull request.

diega:
## PR description
This PR is a continuation of #2944 by moving the hardcoded discovery values into each of the genesis files.
  * a2b081e01 has the minimal set of changes for the _factory method_ of `org.hyperledger.besu.cli.config.EthNetworkConfig` to resolve the bootnodes and dns from the genesis file. The _factory method_ could have been removed and its behavior integrated directly into the `org.hyperledger.besu.cli.BesuCommand` reusing the logic for external genesis files, but that would have impacted in many tests, making the review harder to perform.
  * efaf0ca39 and c12da44ca are not mandatory for this PR but I think it would be hard to make this changes as part of a bigger and meaningful PR. Can be removed if any reviewer request it.
 
It's worth to mention that the hardcoded discovery values were keep just for testing purposes within that scope.

## Enhacement(s)
closes #3113


atoulme (approved):
LGTM

PR #3130, opened 2021-12-06
Backward sync

gezero:
This PR is a WIP proof of concept to be discussed with @garyschulte to possibly change the direction of the work

Even if the approach would get approved, at least the following still needs to get done before a full PR can be raised:
* add tests that properly describe the requirements - at this moment the tests I have are not worth sharing publicly
* don't ignore the blocks proposed by the consensus layer - at this moment, we only care about the headers and ignore the block body.
* should any of the backward sync state persist between restarts?
* the backward sync should sync towards a block that was sent by the consensus layer. At this moment, we sync towards a known head of the blockchain that we already have. What about reorgs?

Highlights to note:
* The algorithm starts when called `org.hyperledger.besu.consensus.merge.blockcreation.backward.sync.BackwardsSyncContext::syncBackwardsUntil`
* The request returns a completable future that might complete exceptionally in case where the sync was interrupted by a newer request sent later or there was an error during sync. 
* Theoretically, the Backward sync can be merged into master independently. 


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/backward/sync/ForwardSyncStep.java
```diff
@@ -0,0 +1,114 @@
...
+      if (!pivot.equals(context.getCurrentPivot().get())) {
+        result.completeExceptionally(
+            new RuntimeException(
+                "The pivot changed, we should stop current flow, some new flow is ongoing..."));
+        return result;
+      } else {
+        return executeForwardStep(result);
+      }
+    } else {
+      result.completeExceptionally(new RuntimeException("No pivot... that is weird"));
```
garyschulte, line 47: entertaining error messages :)

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/backward/sync/BackwardSyncStep.java
```diff
@@ -0,0 +1,137 @@
...
+
+  // if the previous header is not present yet, we need to go deeper
+  @VisibleForTesting
+  protected CompletableFuture<Void> possiblyMoreBackwardSteps(final BlockHeader blockHeader) {
+    CompletableFuture<Void> completableFuture = new CompletableFuture<>();
+    if (context.getProtocolContext().getBlockchain().contains(blockHeader.getHash())) {
+      LOG.debug(
+          "The backward sync let us to a block that we already know... We will init forward sync...");
+      completableFuture.complete(null); // we finished backward sync
+      return completableFuture;
```
garyschulte, line 126: this is an interesting choice - to hit storage looking for the blockhash rather than comparing it to the head or the latest finalized head.  I think this could be a problem if we got a malicious request to executeBlock.  We could end up on a big reorg fork.

Ideally we would check that the block parent is either our current head or latest finalized block.  

Also we should throw rather than more backwards steps if 
* If we DO have a finalized block, and the current block number is not less than finalized block number
* If we DO NOT have a finalized block, keep recursing until we are within some threshold of the terminal total difficulty block.  

Re: the second bullet, I think we might have a corner case where we reorg back before the terminal total difficulty block, so we need to have some reasonable provision for this, but not try to backwards sync all the way back to genesis.  We should abort a backward sync that exceeds that threshold

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/backward/sync/BackwardsSyncContext.java
```diff
@@ -0,0 +1,88 @@
...
+  private final ProtocolContext protocolContext;
+  private final ProtocolSchedule protocolSchedule;
+  private final BlockValidator blockValidator;
+  private final EthContext ethContext;
+  private final MetricsSystem metricsSystem;
+
+  private final Map<Long, BackwardChain> backwardChainMap = new ConcurrentHashMap<>();
+  AtomicReference<Block> currentPivot = new AtomicReference<>();
+
+  public BackwardsSyncContext(
```
garyschulte, line 40: Ideally we would extend SyncState with a merge-specific syncState and just use the MergeContext for access to protocolContext and protocolSchedule.  We could add metrics and ethcontext into the merge-specific sync state class if we need.

That will allow us to look at just a single place when we need to check if we are syncing - either forward or backward, when we are executing blocks or updating fork choice

garyschulte (commented):
Looks like a great start

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/backward/sync/BackwardSyncStep.java
garyschulte: Also, same applies to the head.  It will be normal to need to sync back to a block with a parent prior to the current head.  Anything that is after the latest finalized block is fair game for a reorg

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/backward/sync/BackwardsSyncContext.java
garyschulte: On second thought, having backwards sync and forward sync be separate works well with the notion of appending new blocks to the end of the backward sync from `engineExecuteBlock`.  The only place we have to check both syncState and backwardsSyncContext is in `engineForkChoiceUpdated`.

PR #3073, opened 2021-11-16
Vert.x upgrade

jflo:
## PR description

Upgrades vert.x dependency. 

Notable areas of impact:

- Getting connections from an HTTP client is in itself asynchronous, so another layer of futures is added.
- The way keypairs are added to the JWTAuthProvider is different, and now they are configured using PEM strings.
- JWT attributes and principal fields are changed, for example exp (expiry) is a user attribute, not a principal property.

## Changelog

- [X] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #3135, opened 2021-12-07
Vert.x upgrade (with tests)

diega:
This is a PR continuing #3073 by @jflo just to see if it passes the _acceptanceTests_


macfarla:
acceptance tests are green! but DCO failed

diega:
DCO was failing b/c I made a _rebase_ instead of _merge_ but I fixed it. Now the **integrationTests** are still failing. I ran the reported `org.hyperledger.besu.enclave.EnclaveTest` locally and it works-on-my-machine™, but there has to be another problem, b/c the task is timing out after 10 minutes, when in `main` it's taking less than 5.

macfarla:
the tests that are failing are EnclaveTest which has been updated in this PR to use TesseraTestHarness (currently needs docker) (was OrionTestHarness). integrationTests step runs on medium executor which does not have docker capability. There are a few options to fix this. Not sure which is the best

macfarla (approved):

jflo (approved):

Issue #3142, opened 2021-12-08
exclude test code from sonar coverage


macfarla:
### Description
As an [Besu dev], I want [code coverage to not be enforced on test code] so that [I can clearly see what production code is not covered]. 

### Acceptance Criteria
* acceptance-test code excluded from "new code" coverage

### Steps to Reproduce (Bug)
1. PR that adds new acceptance-test code eg https://sonarcloud.io/component_measures?id=hyperledger_besu&metric=new_coverage&pullRequest=3118&view=list

**Expected behavior:** [What you expect to happen]
```
besu/src/main/java/org/hyperledger/besu/services/BesuPluginContextImpl.java
85.2%	4	0
```
**Actual behavior:** [What actually happens]
```
acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/ThreadBesuNodeRunner.java
0.0%	11	0
besu/src/main/java/org/hyperledger/besu/services/BesuPluginContextImpl.java
85.2%	4	0
```

PR #3149, opened 2021-12-10
(fix) exclude ATs from sonar cloud

frankisawesome:

## PR description
Excludes ATs from sonar cloud runs

## Fixed Issue(s)
fixes https://github.com/hyperledger/besu/issues/3142


frankisawesome:
Ok so sonar cloud on 8f02b9c flags code in the acceptance-tests folder that doesn't directly run tests as uncovered code.

frankisawesome:
Nothing flagged after adding the config option

frankisawesome:
Note to myself while I remember - we also need a to test that coverage
works for non AT files (don't think that'd be an issue but just in case)

jflo (approved):

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BlockchainModule.java
```diff
@@ -61,6 +61,7 @@ MutableWorldView getMutableWorldView(
       final GenesisState genesisState,
       @Named("KeyValueStorageName") final String keyValueStorageName) {
     if ("memory".equals(keyValueStorageName)) {
+      System.out.println("trigger bad coverage");
```
macfarla, line 64: we want to remove this line before merging

macfarla (dismissed):

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BlockchainModule.java
frankisawesome: Oh yeah sorry I requested your review and then realised there's one last thing I needed to test still😛

frankisawesome:
Ok I've tested:
1. File under acceptance-tests with no tests, coverage tested (not good)
2. File under acceptance-tests with no tests, new config, coverage not tested (what we want)
3. File outside of AT, new config, coverage still tested (good)

PR #3156, opened 2021-12-10
Add retry mechanism when we failed to download a block from peers

matkt:
## PR description

Add retry mechanism when we failed to download a block from peers

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingPeerTask.java
```diff
@@ -142,6 +142,14 @@ private boolean isRetryableError(final Throwable error) {
     return error instanceof TimeoutException || (!assignedPeer.isPresent() && isPeerError);
   }
 
+  public EthContext getEthContext() {
+    return ethContext;
+  }
+
+  public MetricsSystem getMetricsSystem() {
+    return metricsSystem;
+  }
```
garyschulte, line 151: these can be protected right?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/RetryingGetBlockFromPeersTask.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Override
+  protected void handleTaskError(final Throwable error) {
+    if (getRetryCount() < getMaxRetries()) {
+      LOG.info(
+          "Failed to get block with hash {} and number {} retrying later", blockHash, blockNumber);
+    }
+    super.handleTaskError(error);
+  }
```
garyschulte, line 97: What is our recovery path if we hit max retries?  is there any mechanism to periodically retry fetching blocks or are we just stuck?

garyschulte (approved):
I think this is a solid improvement, but I think there should be some sort of periodic retry mechanism if we hit max retries.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -150,27 +149,6 @@ private void onBlockAdded(final BlockAddedEvent blockAddedEvent) {
...
               });
-    } else {
-
-      LOG.trace("Not ready for import blocks found for {}", newBlock.getHeader().getNumber());
-
-      pendingBlocksManager
-          .lowestAnnouncedBlock()
-          .map(ProcessableBlockHeader::getNumber)
-          .ifPresent(
-              minAnnouncedBlockNumber -> {
```
garyschulte, line 161: this, in essence, was a periodic retry mechanism, but it relied on blocks being added.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/RetryingGetBlockFromPeersTask.java
matkt: I voluntarily left a maximum because we may actually have a bad block that has been broadcast and that we will never find. This could be a way to back us up if a peers broadcast us a huge list of n+1  blocks IMO

I think rather than trying stubbornly to find a block. the best way is to change peers at some point if we are no longer moving forward. This could be discussed as a follow-up to this PR in a second time

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
matkt: I'm not even sure that adding a new peer will necessarily trigger the search for a new block currently

This mechanism was added as a retry to fix this same problem we have now but it did not work considering what we have on the nodes

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingPeerTask.java
matkt: yes done

matkt:
As proposed by @fab-10  I retry now with a dynamic list of peers

fab-10 (approved):
it is a good first iteration, possible discussed improvement, like changing peers and longer retries, could be added later if the problem persists

PR #3172, opened 2021-12-14
factor out timer from BesuRunner into TransactionPoolEvictionService

taccatisid:
## PR description

Factor out timer from BesuRunner into TransactionPoolEvictionService so that BesuRunner does not contain inner details of the transaction pool implementation.

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolEvictionService.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2: nit: contributors to Hyperledger Besu

macfarla (approved):

Issue #2756, opened 2021-09-14
Add data-path to error message re chain data mismatch


macfarla:
eg if I start besu with default --data-path and I've already forgotten what the default value is, this will save me looking it up. 

### Description
As a Besu user, I want to see the effective data-path in the error message so that I can delete it if switching between networks. 

### Acceptance Criteria
* Actual/effective data-path is printed out in the error message when genesis block mismatch.

### Steps to Reproduce (Bug)
1. Start besu with --network=rinkeby
2. stop besu
3. start besu with --network=ropsten

**Expected behavior:** [What you expect to happen]
```
Supplied genesis block does not match stored chain data in </path/to/data/path/dir>
Please specify a different data directory with --data-path or specify the original genesis file with --genesis-file.
```
**Actual behavior:** [What actually happens]
```
Supplied genesis block does not match stored chain data.
Please specify a different data directory with --data-path or specify the original genesis file with --genesis-file.
```
**Frequency:** 
every time the genesis block doesn't match

asyashn:
I can work on this issue!

PR #3176, opened 2021-12-15
(fix) log data path on error

frankisawesome:

## PR description
We will now log the data path if the genesis doesn't match network provided for easier debug

## Fixed Issue(s)
fixes https://github.com/hyperledger/besu/issues/2756

## Changelog
Not required


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -530,8 +555,15 @@ private void setGenesis(final Block genesisBlock) {
         throw new IllegalStateException("Blockchain is missing genesis block data.");
       }
       if (!genesisHash.get().equals(genesisBlock.getHash())) {
+        final String firstLine;
+        if (dataDirectory != null) {
+          firstLine =
+              "Supplied genesis block does not match stored chain data in " + dataDirectory + "\n";
+        } else {
+          firstLine = "Supplied genesis block does not match stored chain data.\n";
```
taccatisid, line 563: Even if we do not pass a data directory will still have the implicit one. It would be nice to log the effective dataDirectory in that case as well.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -530,8 +555,15 @@ private void setGenesis(final Block genesisBlock) {
         throw new IllegalStateException("Blockchain is missing genesis block data.");
       }
       if (!genesisHash.get().equals(genesisBlock.getHash())) {
+        final String firstLine;
+        if (dataDirectory != null) {
+          firstLine =
+              "Supplied genesis block does not match stored chain data in " + dataDirectory + "\n";
```
macfarla, line 561: might be cleaner to make firstLine a StringBuffer and append to it, or make that first bit a const

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -530,8 +555,15 @@ private void setGenesis(final Block genesisBlock) {
...
+        } else {
+          firstLine = "Supplied genesis block does not match stored chain data.\n";
+        }
         throw new InvalidConfigurationException(
-            "Supplied genesis block does not match stored chain data.\n"
+            firstLine
                 + "Please specify a different data directory with --data-path or specify the original genesis file with --genesis-file.");
```
macfarla, line 567: I wonder if it's worth changing this to add something like "or --network if you're specifying a testnet or mainnet config"

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
macfarla: YES because that I can never remember

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
frankisawesome: that makes sense! Good idea

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
frankisawesome: Hmm yeah I think dataDirectory will always be not null as long as it's run from an actual command line, it can only be null in certain tests when we run Besu differently and misses the default CLI stuff, I could be wrong though, I'll look into it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
frankisawesome: 1368608

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
frankisawesome: 1368608

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
frankisawesome: 1368608

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchainTest.java
```diff
@@ -89,11 +89,12 @@ public void initializeExistingWithWrongGenesisBlock() {
     createMutableBlockchain(kvStore, genesisBlock);
 
     // Initialize a new blockchain store with same kvStore, but different genesis block
-    assertThatThrownBy(() -> createMutableBlockchain(kvStore, gen.genesisBlock()))
+    assertThatThrownBy(() -> createMutableBlockchain(kvStore, gen.genesisBlock(), "/test/path"))
         .isInstanceOf(IllegalArgumentException.class)
         .hasMessageContaining(
-            "Supplied genesis block does not match stored chain data.\n"
-                + "Please specify a different data directory with --data-path or specify the original genesis file with --genesis-file.");
+            "Supplied genesis block does not match stored chain data in /test/path.\n"
```
macfarla, line 95: nit: chain data stored in

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -70,12 +70,21 @@ private DefaultBlockchain(
...
+      final MetricsSystem metricsSystem,
+      final long reorgLoggingThreshold,
+      final String dataDirectory) {
     checkNotNull(genesisBlock);
     checkNotNull(blockchainStorage);
     checkNotNull(metricsSystem);
 
     this.blockchainStorage = blockchainStorage;
-    genesisBlock.ifPresent(this::setGenesis);
+    genesisBlock.ifPresent(block -> this.setGenesis(block, dataDirectory));
```
macfarla, line 87: it seems weird for the DefaultBlockchain to have the data directory?

frankisawesome: It doesn't, but since we handle the validation of the genesis block at this level we pass the string in as a param

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchainTest.java
frankisawesome: 702a483

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
macfarla: ok so it gets passed in and then discarded

macfarla (approved):

PR #3177, opened 2021-12-15
Make 'to' field optional in eth_call method according to the spec

fab-10:

## PR description

According to the spec, the 'to' field of a transaction is optional, and it is not specified when call `eth_call` to create a contract.

## Fixed Issue(s)

fixes #3

## Changelog


garyschulte (approved):
🚢

Issue #3093, opened 2021-11-22
Validate that blockperiodseconds is set to a positive integer


siladu:
0 or decimal values should yield an error on startup.

blockperiodseconds is configured in the genesis file, e.g.
https://besu.hyperledger.org/en/stable/HowTo/Configure/Consensus-Protocols/IBFT/#genesis-file
https://besu.hyperledger.org/en/21.10.2/Tutorials/Private-Network/Create-IBFT-Network/

lucassaldanha:
@georgep9 is looking into this ticket.

PR #3186, opened 2021-12-16
Validate that blockperiodseconds is set to a positive integer

georgep9:
## PR description

Implemented getPositiveInt function in JsonUtil to validate a JSON positive number value. Using this function, the blockperiodseconds should now be validated whenever retrieved from the genensis config. 

## Fixed Issue(s)

Fixes https://github.com/hyperledger/besu/issues/3093


config/src/test/java/org/hyperledger/besu/config/JsonBftConfigOptionsTest.java
```diff
@@ -70,6 +71,13 @@ public void shouldGetDefaultBlockPeriodFromDefaultConfig() {
         .isEqualTo(EXPECTED_DEFAULT_BLOCK_PERIOD);
   }
 
+  @Test
```
lucassaldanha, line 74: Do we wanna add some extra edge cases like negative numbers?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/operator/GenerateBlockchainConfig.java
```diff
@@ -136,8 +136,8 @@ private void checkPreconditions() {
   /** Generates output directory with all required configuration files. */
   private void generateBlockchainConfig() {
     try {
-      handleOutputDirectory();
       parseConfig();
+      handleOutputDirectory();
```
siladu, line 140: I would postpone adding this to the PR (unless we end up modifying besu operator in this PR as well). This change would be better isolated in it's own PR so we don't have to worry about testing alongside the other stuff.

georgep9: Definitely

config/src/test/java/org/hyperledger/besu/config/JsonBftConfigOptionsTest.java
```diff
@@ -70,6 +71,20 @@ public void shouldGetDefaultBlockPeriodFromDefaultConfig() {
...
+  public void shouldThrowOnDecimalBlockPeriod() {
+    final BftConfigOptions config = fromConfigOptions(singletonMap("BlockPeriodSeconds", 1.5));
+    assertThatThrownBy(() -> config.getBlockPeriodSeconds())
+        .isInstanceOf(IllegalArgumentException.class);
+  }
+
+  @Test
+  public void shouldThrowOnNonPositiveBlockPeriod() {
```
siladu, line 82: I think these would better live in PositiveNumberTest. 
But then you might still have one in the *ConfigOptionsTest classes that's something like "shouldThrowForNonPositiveInteger"

config/src/test/java/org/hyperledger/besu/config/CliqueConfigOptionsTest.java
```diff
@@ -64,6 +65,20 @@ public void shouldGetDefaultBlockPeriodFromDefaultConfig() {
...
+    assertThatThrownBy(() -> config.getBlockPeriodSeconds())
+        .isInstanceOf(IllegalArgumentException.class);
+  }
+
+  @Test
+  public void shouldThrowOnNonPositiveBlockPeriod() {
+    final CliqueConfigOptions config = fromConfigOptions(singletonMap("BlockPeriodSeconds", -1));
+    assertThatThrownBy(() -> config.getBlockPeriodSeconds())
+        .isInstanceOf(IllegalArgumentException.class);
+  }
```
siladu, line 80: I think these would better live in PositiveNumberTest.
But then you might still have one in the *ConfigOptionsTest classes that's something like "shouldThrowForNonPositiveInteger"

util/src/test/java/org/hyperledger/besu/util/number/PositiveNumberTest.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright ConsenSys AG.
```
siladu, line 2: Well spotted for noticing the copyright was required!
I think it changed quite recently (some docs may be out of date?) so it should now read "Copyright Hyperledger Besu Contributors."

config/src/test/java/org/hyperledger/besu/config/JsonBftConfigOptionsTest.java
```diff
@@ -70,6 +71,13 @@ public void shouldGetDefaultBlockPeriodFromDefaultConfig() {
         .isEqualTo(EXPECTED_DEFAULT_BLOCK_PERIOD);
   }
 
+  @Test
+  public void shouldThrowOnNonPositiveBlockPeriod() {
+    final BftConfigOptions config = fromConfigOptions(singletonMap("BlockPeriodSeconds", -1));
```
siladu, line 76: nit: could this be "blockperiodseconds" for consistency? Don't think lowercase matters here but it does in other config-related things so better to keep it lower everywhere I think.

config/src/test/java/org/hyperledger/besu/config/CliqueConfigOptionsTest.java
```diff
@@ -64,6 +65,13 @@ public void shouldGetDefaultBlockPeriodFromDefaultConfig() {
         .isEqualTo(EXPECTED_DEFAULT_BLOCK_PERIOD);
   }
 
+  @Test
+  public void shouldThrowOnNonPositiveBlockPeriod() {
+    final CliqueConfigOptions config = fromConfigOptions(singletonMap("BlockPeriodSeconds", -1));
```
siladu, line 70: nit: suggest "blockperiodseconds" in lowercase

config/src/main/java/org/hyperledger/besu/config/JsonUtil.java
```diff
@@ -291,4 +293,15 @@ private static boolean validateInt(final JsonNode node) {
     }
     return true;
   }
+
+  public static int getPositiveInt(
```
siladu, line 297: nit: Could this method live next to getInt? It's very related to it

util/src/test/java/org/hyperledger/besu/util/number/PositiveNumberTest.java
```diff
@@ -0,0 +1,62 @@
...
+    final String positiveNumberString = "1";
+    final int positiveNumberInt = 1;
+    PositiveNumber positiveNumber = PositiveNumber.fromString(positiveNumberString);
+    assertThat(positiveNumber).isNotNull();
+    assertThat(positiveNumber.getValue()).isEqualTo(positiveNumberInt);
+  }
+
+  @Test
+  public void shouldThrowOnDecimalValueFromString() {
+    assertThatThrownBy(() -> PositiveNumber.fromString("0.1"))
```
siladu, line 35: "1.1" is a better choice here as "0.1" might fail due to being non-positive, rather than a decimal.

siladu (changes requested):
Minor nitpicks and suggestions in the comments.

I am going to be annoying though and ask that you make the same change for `BftFork.getBlockPeriodSeconds` please :)
It is possible to use a 'transition' to change the blockperiodseconds of an existing network so it needs the same validation constraints applied to it.
See: https://besu.hyperledger.org/en/stable/HowTo/Configure/Consensus-Protocols/IBFT/#configure-block-time-on-an-existing-network-deployment

siladu (changes requested):

siladu (commented):
Code looks good 👍 . 
I want to have a play with it locally before I approve which will probably be next year :)

siladu (approved):
LGTM!

PR #3187, opened 2021-12-16
Install solc for Circle CI build

fab-10:

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,10 @@ commands:
       - run:
           name: Install Packages - LibSodium, nssdb
           command: |
+            sudo add-apt-repository ppa:ethereum/ethereum
             sudo apt-get update
             sudo apt-get install -y libsodium23 libsodium-dev apt-transport-https haveged libnss3-tools
+            sudo apt-get install -y solc
             sudo service haveged restart
       - restore_cache:
           name: Restore cached gradle dependencies


acceptance-tests/tests/build.gradle
```diff
@@ -24,6 +24,7 @@ sourceSets.main.solidity.srcDirs = ["$projectDir/contracts"]
 
 solidity {
   resolvePackages = false
+  executable = "/usr/bin/solc"
```
antonydenyer, line 27: I think we should do something like:

`executable = System.getenv('SOLC_BINARY_PATH') ?: null`

then if you don't have solc installed sokt will attempt to fetch it for you (if services are running)

macfarla: for running it locally?

antonydenyer: In an ideal world, I'd be able to clone besu and run `./gradlew check` and not have to have anything installed. 

By adding the path explicitly it means that everyone needs to have solc installed in the same place to be able to build. It also means that you can only build on *nix! 

Another option would be to put some code into the gradle scripts to download solc if it's needed (basically doing what sokt does at the moment)

fab-10: the build already requires some packages to be installed, libsodium and nssdb, so adding another is not changing to much, and running tests on Windows does not work yet, but I agree that these are shortcomings that could be addressed.

PR #3197, opened 2021-12-17
Bugfix/suggested fee recipient

garyschulte:
## PR description
fix a defect found in kintsugi testnet where we produce blocks with the coinbase from miningParams, but use suggestedFeeRecipient for the blockheader.
* consistently use suggestedFeeRecipient and fall back to miningParams
* start MiningCoordinator test coverage with this case
* clean up of unnecessary blockValidator params to MiningCoordinator and BackwardsSyncContext
* ensure only latest (best) blockProposal is available by payloadId from MergeContext

## Changelog

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
@@ -18,7 +18,6 @@ import org.hyperledger.besu.consensus.merge.MergeContext;
 import org.hyperledger.besu.datatypes.Address;
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.datatypes.Wei;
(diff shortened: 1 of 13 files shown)


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -88,7 +84,7 @@ public MergeCoordinator(
                 protocolContext,
                 protocolSchedule,
                 this.miningParameters.getMinTransactionGasPrice(),
-                this.miningParameters.getCoinbase().orElse(Address.ZERO),
+                address.or(miningParameters::getCoinbase).orElse(Address.ZERO),
```
garyschulte, line 87: this was the cause of the bad block proposals on kintsugi.  Consistenly using suggestedFeeRecipient now, with fallback to miningParameters coinbase

jflo (approved):

PR #3200, opened 2021-12-19
Snapshot 21.10.6

shemnon:
## Changelog


atoulme (approved):

macfarla:
I don't think we need this one - we are already on version=22.1.0-RC2-SNAPSHOT

macfarla (approved):

shemnon:
Keeping this there in case there is another log4j high vulnerability.  I'm hoping it's not going to happen but it feels like the 12 days of Christmas.

PR #3202, opened 2021-12-19
When downloading the world state, persist children before parents

gezero:
Several changes are happening at the same time in this PR.

1. The pending requests queue in the world state downloader is now a different data structure. We now use a list of priority queues.
2. The node requests now know about the parent request that was responsible for spawning them.
3. When a node has data available and is asked to persist before its children are persisted, the node will not do anything. Instead, it will wait for all its children to persist first and will persist together with the last child. 

Storing children before parents gives us the following benefits:
* We don't need to store pending requests on disk any more. 
* When restarting download from a new pivot, we do not need to walk and check the whole tree any more. 

And the following drawbacks:
* We now have pending nodes in memory for which we already downloaded data, but we do not store them in the database yet. 

Overall expectations on performance:
We still need to download every single state node at least once, so there is no improvement there. We will save a significant amount of time in case we change pivots. And we save lots of read/writes on filesystem because tasks are not needed to be written to disk any more. 

We want to avoid having too many pending unsaved nodes in memory, not to run out of it. If we were always handling only one request to our peers at the same time, we would not need to be worried, and we would just use a simple depth first search. Because we batch our requests, we might produce too many pending unprocessed nodes in memory if we are not careful about the order of processing requests. That is where the priority on node request comes from. We want to always process nodes lower in the tree before nodes higher in the tree. Preferably, we want to first process children from the same parent so that we can save the current unsaved parent as soon as possible. 

Similarly, because of the order of processing of the nodes, we do not need to store huge number of requests on the disk any more and the whole list fits comfortably into the memory. Without batching, we would not have more than a thousand requests around waiting. Because of the batching, we can see the number of requests occasionally rises all the way up to 300 000, but usually should be under 200 000.

Note that at any time there should not be more pending unsaved blocks than pending requests. Such a situation would be a bug to be reported.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/NodeDataRequest.java
```diff
@@ -132,10 +107,33 @@ public NodeDataRequest setRequiresPersisting(final boolean requiresPersisting) {
   }
 
   public final void persist(final WorldStateStorage.Updater updater) {
+    if (pendingChildren.get() > 0) {
+      if (totalPending.incrementAndGet() > max + 999) {
```
matkt, line 111: maybe we can explain why 999 ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/NodeDataRequest.java
```diff
@@ -132,10 +107,33 @@ public NodeDataRequest setRequiresPersisting(final boolean requiresPersisting) {
...
+        max = totalPending.get();
+        LOG.warn("Maximum pending nodes just reached {}", max);
```
matkt, line 113: what we do in the case ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/NodeDataRequest.java
```diff
@@ -15,24 +15,38 @@
...
+
   private final RequestType requestType;
   private final Hash hash;
   private Bytes data;
   private boolean requiresPersisting = true;
   private final Optional<Bytes> location;
+  private Optional<NodeDataRequest> possibleParent = Optional.empty();
+  private final AtomicInteger pendingChildren = new AtomicInteger(0);
+  private static NodeDataRequest root;
+  private int level = 0;
```
matkt, line 45: it's not important but I might have used depth instead of level

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/CompleteTaskStep.java
```diff
@@ -62,13 +65,15 @@ public void markAsCompleteOrFailed(
...
       final Task<NodeDataRequest> task) {
     if (task.getData().getData() != null) {
-      enqueueChildren(task, header, downloadState);
       completedRequestsCounter.inc();
       task.markCompleted();
       downloadState.checkCompletion(worldStateStorage, header);
     } else {
       retriedRequestsCounter.inc();
       task.markFailed();
+      if (failedCount.incrementAndGet() >= 100_000) {
```
matkt, line 74: I think this can be configurable. In my opinion we already have a flag that we could maybe use

gezero: This was a temporary counter, I think I am going to remove it altogether.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/NodeDataRequest.java
gezero: I am going to remove this log together with removal of the totalPending nodes.

gezero: This is just a statistic that I use to make sure that the amount does not explode to be too high. It should get stable quickly after the fast sync gets initiated. A very high number or an always growing number would be a sign of a bug.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/NodeDataRequest.java
gezero: I am going to remove it. maybe >= 1000 would make a better sense, I just update it when the difference is at least 1000

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+    return internalQueues.get(priority);
+  }
+
+  @Override
+  public synchronized Task<T> remove() {
+    assertNotClosed();
+    final Queue<T> lastNonEmptyQueue = findLastNonEmptyQueue();
+    if (lastNonEmptyQueue.isEmpty()) {
+      return null;
```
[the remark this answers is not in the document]
gezero: I guess I copied what was already there in other implementations. They return null when the collection is empty.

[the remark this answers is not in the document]
gezero: If you think I should update the return value to Optional in this PR let me know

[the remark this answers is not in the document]
matkt: it's fine for me

services/tasks/src/test/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueuesTest.java
```diff
@@ -0,0 +1,175 @@
...
+      return priority;
+    }
+
+    @Override
+    public int getDepth() {
+      return depth;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
atoulme, line 158: this class implements equals and hashCode. Is it expected all impls of `TasksPriorityProvider` implement those methods? I don't see them implemented for all of them.

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
```diff
@@ -0,0 +1,169 @@
...
+
+  public synchronized boolean contains(final T request) {
+    final PriorityQueue<T> queue = findQueue(request.getDepth());
+    return queue.contains(request)
+        || unfinishedOutstandingTasks.stream()
+            .map(InMemoryTask::getData)
+            .anyMatch(data -> data == request);
+  }
+
+  private static class InMemoryTask<T extends TasksPriorityProvider> implements Task<T> {
```
atoulme, line 140: doesn't implement equals or hashCode, is that going to be a problem?

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
```diff
@@ -0,0 +1,169 @@
...
+  private synchronized boolean markTaskCompleted(final InMemoryTask<T> task) {
+    return unfinishedOutstandingTasks.remove(task);
+  }
+
+  public synchronized boolean contains(final T request) {
+    final PriorityQueue<T> queue = findQueue(request.getDepth());
+    return queue.contains(request)
+        || unfinishedOutstandingTasks.stream()
+            .map(InMemoryTask::getData)
+            .anyMatch(data -> data == request);
```
atoulme, line 137: Interesting native `==` use there

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @Override
+  public synchronized boolean allTasksCompleted() {
+    return isEmpty() && unfinishedOutstandingTasks.isEmpty();
+  }
+
+  @Override
+  public synchronized void close() {
+    closed.set(true);
```
atoulme, line 112: Why is close not checking the `closed` bit?

atoulme: also, why is close not handling the `unfinishedOutstandingTasks` set? Do you want to wait for them to complete to return? What's the handling of this?

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.PriorityQueue;
+import java.util.Queue;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.jetbrains.annotations.NotNull;
+
+public class InMemoryTasksPriorityQueues<T extends TasksPriorityProvider>
+    implements TaskCollection<T> {
+  private final List<PriorityQueue<T>> internalQueues = new ArrayList<>(17);
```
atoulme, line 30: should be 16, given the initialization goes to 16

services/tasks/src/test/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueuesTest.java
gezero: I used the equals/hashcode/toString only when creating the testcases. The Interface is expecting to see the same instance of objects at the moment (as you noted with the == used in the contains method)

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
gezero: That should not be a problem when used with the priority queues. The only requirement is for the priority not to change during the time when the entry is stored in the queue.

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
gezero: Do you think I should change it to something else?

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
gezero: I will look into this. I copied over what was used in other implementations, might be worth double-checking what to do correctly here.

gezero: You are right, unfinishedOutstandingTasks should be cleared when close is called.  I am changing it. Also, it makes sense to only clear the collections once.

matkt:
ok on my side, I'm waiting for Antoine's validation

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
atoulme: a bit unclear, but it's a smell in java. It tends to bite down the road when for some reason two objects are expected to be equal but have different pointers.

atoulme (approved):
LGTM

services/tasks/src/main/java/org/hyperledger/besu/services/tasks/InMemoryTasksPriorityQueues.java
gezero: changed to equals

matkt (approved):
LGTM

PR #3203, opened 2021-12-19
Update minimum solc version in all contracts

macfarla:
Recent changes to support latest (0.8.x) version of solc mean that contracts no longer compile with earlier versions. Require at least 0.7.0

Also added SPDX header to some .sol files that were missing it.

## Changelog


mark-terry (approved):
LGTM.

pinges (commented):
LGTM

PR #3224, opened 2022-01-03
3141 terminal blocks

### LINKED ISSUE #3141: Implement merge terminal block overrides
### Description
[Kintsugi v3 version of EIP-3675](https://github.com/ethereum/EIPs/blob/d896145678bd65d3eafd8749690c1b5228875c39/EIPS/eip-3675.md#fork-choice-rule) specifies "terminal block override" behavior.

Besu should implement Optional/empty placeholders for 
* TERMINAL_BLOCKHASH
* TERMINAL_BLOCK_NUMBER

If present post-merge, these configs should behave similar to besu's existing 'required-block' parameter which will force the TERMINAL_TOTAL_DIFFICULTY block to have the specified block number and hash as the terminal PoW block.  These values won't be know until after the transition, so they should be implemented as empty configs initially.

### Acceptance Criteria
* placeholeders for teminal PoW block override configs
* when present, these values should be treated as a required-block for Peers
* `engineExecuteBlock` should return JsonRpcError.INVALID_TERMINAL_BLOCK if the latest valid ancestor is not a descendant of configured TTD block

jflo:
## PR description

Allows for configuration of a terminal block once it has been identified via the difficulty bomb.

## Fixed Issue(s)

fixes #3141 

## Changelog

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
@@ -42,7 +42,7 @@ public class NoDifficultyRule implements AttachedBlockHeaderValidationRule {
     if (protocolContext.getConsensusContext(MergeContext.class).isPostMerge()) {
       if (totalDifficulty
           .get()
-          .greaterOrEqualThan(
+          .greaterThan(
               protocolContext
                   .getConsensusContext(MergeContext.class)
                   .getTerminalTotalDifficulty())) {
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
@@ -39,11 +39,13 @@ public class NoDifficultyRule implements AttachedBlockHeaderValidationRule {
       LOG.warn("unable to get total difficulty, parent {} not found", header.getParentHash());
       return false;
     }
-    if(protocolContext.getConsensusContext(MergeContext.class).isPostMerge()) {
+    if (protocolContext.getConsensusContext(MergeContext.class).isPostMerge()) {
       if (totalDifficulty
-              .get()
-              .greaterOrEqualThan(
-                      protocolContext.getConsensusContext(MergeContext.class).getTerminalTotalDifficulty())) {
+          .get()
+          .greaterOrEqualThan(
+              protocolContext
+                  .getConsensusContext(MergeContext.class)
+                  .getTerminalTotalDifficulty())) {
         return (header.getDifficulty() == null || header.getDifficulty().equals(Difficulty.ZERO));
       } else {
         return true;
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
@@ -37,7 +37,7 @@ public class NoDifficultyRule extends MergConsensusRule {
       LOG.warn("unable to get total difficulty, parent {} not found", header.getParentHash());
       return false;
     }
-    if (super.isProofOfStakeBlock(header, protocolContext)
+    if (super.shouldUsePostMergeRules(header, protocolContext)
         && !super.isTerminalProofOfWorkBlock(header, protocolContext)) {
       return (header.getDifficulty() == null || header.getDifficulty().equals(Difficulty.ZERO));
     } else {
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
@@ -45,8 +45,4 @@ public class NoDifficultyRule extends ProofOfStakeRule {
     }
   }
 
-  @Override
-  public boolean includeInLightValidation() {
-    return super.includeInLightValidation();
-  }
 }
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
@@ -44,5 +44,4 @@ public class NoDifficultyRule extends ProofOfStakeRule {
       return true;
     }
   }
-
 }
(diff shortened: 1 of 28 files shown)


garyschulte (changes requested):
Great tests.  Some nits and minor changes.  

I have yet to grok the reorg test, but wanted to provide timely feedback

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeReorgTest.java
```diff
@@ -0,0 +1,191 @@
...
+  as long as a post-merge PoS block has not been finalized,
+  then you can and should be able to re-org to a different pre-TTD block
+  say there is viable TTD block A and B, then we can have a PoS chain build on A for a while
+      and then see another PoS chain build on B that has a higher fork choice weight and causes a re-org
+  once any post-merge PoS chain is finalied though, you'd never re-org any PoW blocks in the tree ever again */
+
+  @Test
+  public void reorgsAcrossTDDToDifferentTargetsWhenNotFinal() {
+    // Add N blocks to chain from genesis, where total diff is < TTD
+    List<Block> endOfWork = subChain(genesisState.getBlock().getHeader(), 10, Difficulty.of(100L));
```
garyschulte, line 110: I am not sure how much difference this makes, but 10 blocks * 100 diff = 1000 total diff == TTD exactly.  I think the tenth block here is technically the terminal block.

consensus/merge/src/test/resources/mergeAtGenesis.json
```diff
@@ -13,7 +13,7 @@
   "timestamp": "0x0",
   "extraData": "0x11bbe8db4e347b4e8c937c1c8370e4b5ed33adb3db69cbdb7a38e1e50b1b82fa",
   "gasLimit": "0x1fffffffffffff",
-  "difficulty": "0x10000",
+  "difficulty": "0x0",
```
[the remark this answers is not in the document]
jflo: the term "merge at genesis" is breaking my brain. what if we named this posAtGenesis instead? I can revert it, and create a new file to express a genesis that starts at pow and transitions to pos for relevant tests.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyRule.java
```diff
@@ -39,11 +39,17 @@ public boolean validate(
...
     }
-    if (totalDifficulty
-        .get()
-        .greaterOrEqualThan(
-            protocolContext.getConsensusContext(MergeContext.class).getTerminalTotalDifficulty())) {
-      return (header.getDifficulty() == null || header.getDifficulty().equals(Difficulty.ZERO));
+    if (protocolContext.getConsensusContext(MergeContext.class).isPostMerge()) {
+      if (totalDifficulty
+          .get()
+          .greaterThan(
```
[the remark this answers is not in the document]
jflo: yes, i think you are right. This should be saying if the current chain has reached TTD, evaluate passed in block as a PoS block...

Which makes me think we have another problem, and this will break if we ever re-validate a past block after the chain has merged.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyValidationTest.java
```diff
@@ -55,7 +55,7 @@ public void difficultyMustBeZero() {
     final BlockHeader parentHeader = mock(BlockHeader.class);
 
     final BlockHeader invalidHeader = mock(BlockHeader.class);
-    when(invalidHeader.getDifficulty()).thenReturn(Difficulty.MAX_VALUE);
+    when(invalidHeader.getDifficulty()).thenReturn(Difficulty.of(10));
```
jflo, line 58: turns our Difficulty will overflow

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/ProofOfStakeRule.java
```diff
@@ -0,0 +1,80 @@
...
+import org.hyperledger.besu.ethereum.core.BlockHeader;
+import org.hyperledger.besu.ethereum.core.Difficulty;
+import org.hyperledger.besu.ethereum.mainnet.AttachedBlockHeaderValidationRule;
+
+import java.util.Optional;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public abstract class ProofOfStakeRule implements AttachedBlockHeaderValidationRule {
```
garyschulte, line 28: naming feedback, we I'd suggest MergeConsensusRule just to keep with the theme.

garyschulte, line 28: I don't think proof-of-stake should leak into the execution client.  We aren't actually doing proof of stake, we are deferring to an external consensus mechanism.  If besu were being used as an execution client in a rollup, there may or may not be proof-of-stake, it could just as easily be PoA etc

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/ProofOfStakeRule.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final Logger LOG = LogManager.getLogger(ProofOfStakeRule.class);
+
+  protected boolean isProofOfStakeBlock(final BlockHeader header, final ProtocolContext context) {
+
+    Optional<Difficulty> currentChainTotalDifficulty =
+        context.getBlockchain().getTotalDifficultyByHash(header.getParentHash());
```
garyschulte, line 34: how expensive is this call?  is the result cached or are we re-reading for each execution?

garyschulte (approved):
LGTM.  non-blocking feedback

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/ProofOfStakeRule.java
jflo: fair point, easy fix, will adopt

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/headervalidationrules/ProofOfStakeRule.java
jflo: The result is not cached, and this is as expensive as any other blockhash lookup in RocksDB- and it will grow with chain height.  I don't think we can short-circuit due to re-orgs that go back to PoW?

garyschulte (approved):
🚢

shemnon:
Merge had no DCO, so it was rolled back as it was breaking every besu build.

commit is here - https://github.com/shemnon/besu/commit/11b5f95dd5ef51af398f8b343b266debadd6f0b9

Issue #3229, opened 2022-01-04
security vulnerabilities identified in Nexus IQ


mgriffin75:
My Enterprise is at a heightened security awareness with the log4j vulnerabilities as are most. We scanned 21.10.5 in Nexus IQ, the scan results are below. We cannot proceed with security vulnerabilities seven or higher. When can these be remediated? When do you plan to update log4j to 2.17.1? Threat of 10
com.fasterxml.jackson.core : jackson-databind : 2.12.4
com.google.code.gson : gson : 2.8.6
com.squareup.okhttp: okhttp: 2.7.5
com.squareup.okhttp3 : okhttp:4.9.1
io.kubernetes : client-java : 5.0.0
io.netty: netty-all : 4.1.66. Final
io.netty: netty-all: 4.1.66. Final
io.netty: netty-codec : 4.1.65.Final
io.netty: netty-codec : 4.1.65.Final
io.netty: netty-codec : 4.1.65. Final

Threat of 7
com.squareup.okhttp: okhttp: 2.7.5
commons-codec : commons-codec : 1.13
io.grpc: grpc-core: 1.39.0
io.kubernetes: client-java : 5.0.0
io.netty: netty-all : 4.1.66.Final
io.netty: netty-codec-http: 4.1.65.Final
io.netty: netty-handler : 4.1.65.Final
commons-codec: commons-codec: 1.13
io.grpc: grpc-core : 1.39.0
io.kubernetes : client-java:5.0.0
io.netty: netty-all : 4.1.66.Final
io.netty: netty-codec-http : 4.1.65.Final
io.netty: netty-handler: 4.1.65.Final
io.vertx: vertx-web : 3.9.8
org.apache.logging.log4j : log4j-core : 2.17.0
org.bouncycastle : bcprov-ext-jdk15on: 1.61
org.bouncycastle : bcprov-jdk15on: 1.69

macfarla:
@mgriffin75 Release 21.10.6 completed just now: this updates log4j to 2.17.1 https://github.com/hyperledger/besu/releases/tag/21.10.6

mgriffin75:
@macfarla thank you for the update, new release resolves the log4j vulnerability.  However we are unable to move forward in our Enterprise due to the security vulnerabilities (above that are 7 and higher) identified in Nexus IQ.  Any ideas on when these will be addressed?  Thanks for your help

macfarla:
[...]

PR #3236, opened 2022-01-05
Updating kubernetes dependency

macfarla:

    dependency 'io.kubernetes:client-java:6.0.1'

https://github.com/kubernetes-client/java/releases

See #3229 

## Changelog


macfarla:
breaking change in the API io.kubernetes:client-java

gezero:
It seems the missing parameter was unused by us in this context anyway. If used the selector includeUnitialized would change the list of returned services, but we did not specify this (or any other) selector and as such the call returns us all services. I raised a [PR](https://github.com/macfarla/besu/pull/1) to add a commit to remove that parameter.

PR #3242, opened 2022-01-06
DO NOT MERGE - TLS P2P Hacky POC Fix

siladu:
Hacky POC to fix TLS P2P failing to process messages over a certain size (> TLS record limit of 16 KB)

Copied Framer impl into PlainFramer and commented out the encryption-related bits.
This means we still get some of the partial message processing logic.
Essentially, we've added a HEADER containing the expected size of the message and will wait for more data if the message buffer only contains a partial message.

This will become the fix for https://github.com/ConsenSys/protocol-misc/issues/554


siladu:
Replaced by https://github.com/hyperledger/besu/pull/3253

PR #3264, opened 2022-01-11
Release 21.10.7

macfarla:
Update dependencies in preparation for release 21.10.7
Also JWT support for multiple algorithms as this overlapped with the vert.x upgrade

## Changelog


macfarla:
DCO is borked

Issue #2867, opened 2021-10-11
Tessera ATs - run Tessera as a process


pinges:
When we added the Tessera container tests the time to run the acceptance tests increased.

Currently we are starting a docker container for each of the Besu and Tessera nodes for each test that involves Tessera.

1. Reuse the docker instances per test class
2. Make Tessera nodes run in a process/thread like they do in other acceptance tests

Any other ideas?

mark-terry:
Option 1:
https://github.com/testcontainers/testcontainers-java/pull/1781

Container reuse in testcontainers is still WIP and appears to have outstanding work around assigning networks. This does not appear to be viable at this point in time.

MadelineMurray:
@mark-terry - talking about this in planning today.  We're going to do with option 2 and can start the work to make this change (ie, it's not blocked). 

@pinges is going to link his previous PR using this approach. The previous PR is a MVP approach so this work is about implementing that approach with production quality code.

MadelineMurray:
There's going to be two PRs - one in Tessera and one in Besu so leaving this issue in the Besu repo rather than moving to Tessera.

taccatisid:
Going for "Make Tessera nodes run in a process/thread like they do in other acceptance tests"

jflo:
Am interested in this, as the PrivacyReorgTest depends on Orion today, and when I swapped it out for Tessera in this pr https://github.com/hyperledger/besu/pull/3073, I triggered the dependency on testcontainers. That's not a deal breaker per-se, but since it is happening as a unit test instead of an integration/acceptance test, there is no docker daemon running to deploy Tessera to.

macfarla:
This functionality is now in Tessera. However we need a Tessera release 
22.1.1 is current https://github.com/ConsenSys/tessera/releases/tag/tessera-22.1.1
Need 22.1.2 for this

Gabriel-Trintinalia:
Blocked by https://github.com/ConsenSys/tessera/issues/1490

Gabriel-Trintinalia:
Blocked by https://github.com/hyperledger/besu/pull/4532

Tessera needs Java 17.

non-fungible-nelson:
@macfarla - anything related to latest tessera AT changes? Close?

PR #3265, opened 2022-01-11
Release 21.10.7

macfarla:
Update dependencies in preparation for release 21.10.7
Also JWT support for multiple algorithms as this overlapped with the vert.x upgrade

See #3229 

## Changelog


fab-10 (approved):
LGTM

gradle/versions.gradle
```diff
@@ -39,27 +39,27 @@ dependencyManagement {
...
     dependency 'commons-io:commons-io:2.11.0'
 
     dependency 'dnsjava:dnsjava:3.4.0'
 
     dependency 'info.picocli:picocli:4.6.1'
 
-    dependency 'io.grpc:grpc-core:1.39.0'
-    dependency 'io.grpc:grpc-netty:1.39.0'
-    dependency 'io.grpc:grpc-stub:1.39.0'
+    dependency 'io.grpc:grpc-core:1.37.1'
```
shemnon, line 52: I'd like exposition as to why grpc is going back two minor versions.  Also, does this have the same security baseline as 1.39.0?  i.e. are we going back on CVE exposures by doing this?

.circleci/config.yml
```diff
@@ -147,7 +147,7 @@ jobs:
           command: ./gradlew --no-daemon jacocoRootReport sonarqube -Dsonar.login=$SONAR_TOKEN
 
   integrationTests:
-    executor: besu_executor_med
+    executor: xl_machine_executor
```
shemnon, line 150: We should loop back and figure out how to get the integration tests running on smaller machines.  xl vs medium is an expensive jump.

gradle/versions.gradle
macfarla: agree. changing this back to 1.39.0 and asking customer for their scan report

.circleci/config.yml
macfarla: agree. this will be rectified once we can get rid of dependency on container tests ie https://github.com/hyperledger/besu/issues/2867

macfarla:
changed version back to 21.10.7-SNAPSHOT and set to auto-merge so the snapshot version will be updated and scanned

.circleci/config.yml
macfarla: cc @antonydenyer

PR #3270, opened 2022-01-12
Clear previous version fast sync runs data

### LINKED ISSUE #3267: Upgrade of version to the latest main commit during ongoing fastsync breaks the world state database
### Description
As a Besu user, I want to be able to upgrade Besu version while in the middle of fastsync. 

### Acceptance Criteria
* be able to stop and upgrade Besu to new version in the middle of fast sync

### Steps to Reproduce (Bug)
1. Start fastsync on old version of Besu
2. Stop Besu before the fast sync world state download finishes
3. Upgrade to the latest Besu commit
4. Start Besu

* The problem always happens, but may be not visible because missing world state nodes are not in use yet.

gezero:
When picking up the fast sync after previous incomplete run of fast sync, we have to clear the old world state data to be sure children are persisted before parents.

## Fixed Issue(s)

fixes #3267

## Changelog


fab-10 (approved):
LGTM

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastDownloaderFactory.java
```diff
@@ -82,6 +85,13 @@
     }
     if (worldStateStorage instanceof BonsaiWorldStateKeyValueStorage) {
       worldStateStorage.clear();
+    } else {
```
matkt, line 88: why you are not clearing with Bonsai ?

PR #3275, opened 2022-01-13
Remove orion gradle dependency

### LINKED ISSUE #3271: remove orion dependency
We've removed all Orion tests but testutil still has a dependency on orion which should be removed (and removed from versions.gradle)
* there is a transitive dependency that it provides but we should add that directly rather than via orion 
* check no other modules have orion dependency (I think I saw it somewhere else)

macfarla:
* Remove orion gradle dependency
* Add direct dependency in testutil for vertx and jackson-databind
* Fixed dependency version listing for com.fasterxml.jackson.datatype:jackson-datatype-jdk8
Fixes #3271

## Changelog


atoulme (approved):
LGTM

antonydenyer (approved):

Issue #3259, opened 2022-01-10
Add Compression to TLS P2P


siladu:
This is specified here: https://github.com/ethereum/devp2p/blob/master/rlpx.md#version-5-eip-706-september-2017

Our enterprise-only TLS P2P implementation (a modified version of the RLPx spec) does not include encryption, a MAC or compression. Encryption/MAC are replaced with TLS, however compression remains unimplemented. It is still a desirable feature to reduce unnecessary data traffic as explained in https://eips.ethereum.org/EIPS/eip-706

The TLS P2P implementation uses TLS v1.3 which has removed compression due to security flaws, so compression via TLS is not an option.

The recommendation is to use Snappy Compression similar to the original RLPx implementation, preferably reusing the same code for both TLS and non-TLS modes.

PR #3282, opened 2022-01-14
besu-3259 Add Compression to TLS P2P

lucassaldanha:
## PR description

Enabling Snappy compression for DevP2P-over-TLS. I'll follow up with a PR adding some testing around the Netty pipeline, however, this code is already being tested as part of our TLS acceptance tests so it is safe to merge.

## Fixed Issue(s)
see #3259


usmansaleem (approved):
LGTM

jframe (approved):

PR #3284, opened 2022-01-16
revert to using bundled solc

### LINKED ISSUE #3277: Remove build-time dependency on solc
Currently to build besu you need to have solidity installed. 

We did this recently to get around the web3j gradle plugin needing to contact an internal server that was down. Long story. But 
* the server is back up
* the gradle plugin is soon/already updated to avoid contacting this internal server (check if there's an updated version)
so shouldn't be a problem in the future.

We want to go back to using the bundled solc that comes with the plugin. 

Most helpful would be to look at the recent PR that made the change to circle CI config to install solc
https://github.com/hyperledger/besu/pull/3183/files
Two changes that need to be reverted
* don't install solc on CI executor
* don't specify solc path/version in acceptance-tests/tests/build.gradle

See discussion here https://chat.hyperledger.org/channel/besu-contributors?msg=jysJnqTRZxWc77Zoj

macfarla:

Remove install of solc on CI executor and revert to using solc that is bundled with the gradle plugin

Fixes #3277 

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,9 +37,8 @@ commands:
       - run:
           name: Install Packages - LibSodium, nssdb
           command: |
-            sudo add-apt-repository ppa:ethereum/ethereum
             sudo apt-get update
-            sudo apt-get install -y libsodium23 libsodium-dev apt-transport-https haveged libnss3-tools solc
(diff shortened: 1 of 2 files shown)


lucassaldanha (approved):
LGTM

PR #3288, opened 2022-01-17
dependency check - added gradle plugin

macfarla:

Add plugin for gradle dependency check

## Changelog


macfarla:
it seems a better solution is the homebrew install
brew install dependency-check
dependency-check -s build/distributions

PR #3296, opened 2022-01-18
[regression] Reenable cors-origin to support any protocol

diega:

## PR description
Before the vert.x upgrade (#3135), the usage of `*` in the Cors-Origin [allowed any string](https://github.com/vert-x3/vertx-web/blob/3.9.8/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/CorsHandlerImpl.java#L169) to pass. Meanwhile the [current version](https://github.com/vert-x3/vertx-web/blob/4.2.1/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/CorsHandlerImpl.java#L231) performs a series of validations for the same wildcard, which allows only a fixed set of protocols (i.e. ftp, http, https). To return the original semantics of `*` according the documentation, an actual regex must be used.

This causes problems when a browser extension, like [Metamask](https://metamask.io/), tries to connect to a Besu client because they tend to use custom protocols. E.g. Chrome uses `chrome-extension` and Firefox `moz-extension`

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Update to block header validation for IBFT and QBFT to support London fork EIP-1559 [#3251](https://github.com/hyperledger/besu/pull/3251)
 
 ### Bug Fixes
+- Fix regression on cors-origin star value
 
 ## 22.1.0-RC2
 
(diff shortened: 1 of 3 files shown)


macfarla (approved):
I can verify that this test fails without the change

jflo (approved):

PR #3317, opened 2022-01-24
add trivy workflow

macfarla:

trivy workflow

## Changelog

### DIFF
--- a/.github/workflows/trivy-analysis.yml
+++ b/.github/workflows/trivy-analysis.yml
@@ -1,0 +1,40 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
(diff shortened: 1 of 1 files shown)


shemnon:
Looking at the report my concern it is that it is too heavy on the docker issues and not enough on the core dependency issues.  Almost all of the findings call for tweaks in the docker files and hence this tool looks to only be validating a distribution not the core java libraries.

macfarla:
> Looking at the report my concern it is that it is too heavy on the docker issues and not enough on the core dependency issues. Almost all of the findings call for tweaks in the docker files and hence this tool looks to only be validating a distribution not the core java libraries.

Fair point. I want to compare this report to the nightly dockerScan #3346

macfarla:
closing this one. The nightly docker scan is clearer.

PR #3333, opened 2022-01-27
Add ec-curve parameter public key export/export-address subcommands

### LINKED ISSUE #3206: Export subcommand for R1 curve
### Description
As an node operator, I'd like to use R1 curves and the sub command to precreate keys doesn't appear to support this https://besu.hyperledger.org/en/latest/Reference/CLI/CLI-Subcommands/#export_1

In doing the https://besu.hyperledger.org/en/latest/Reference/CLI/CLI-Subcommands/#generate-blockchain-config I can get the validators to have R1 nodekeys prefectly fine and when a node starts up clean, it looks at the genesis and can generate an R1 compatible nodekey but the subcommand needs an extra arg

### Acceptance Criteria
Have an R1 nodekey generated by the `export` subcommand

### Steps to Reproduce (Bug)
1. Create an R1 network
2. Create a nodekey using ` .../bin/besu public-key export --to=public.key` 
3. Use the new nodekey from step 2 for a new node that joins the network in step 1 and it'll fail with Breach_of_protocol errors

Expect that the nodekey would work for the R1 network

Node doesn't join the network and has Breach_of_protocol errors in the logs

100% 

* Software version: [`besu --version`] 21.10.2 +
* Java version: [`java -version`] jdk11
* OS Name & Version: [`cat /etc/*release`] ubuntu 20.04lts
* Kernel Version: [`uname -a`] linux 5.11
* Virtual Machine software & version: [`vmware -v`] n/a
* Docker Version: [`docker version`] n/a same behaviour natively 
* Cloud VM, type, size: [Amazon Web Services I3-large] n/a

### Tasks
* Add curve type to the public key subcommand. Keep 
* Update KeyPairUtil to log what curve we are using when creating a new public key

Hint: Take a look at GenerateBlockchainConfig processEcCurve

lucassaldanha:
## PR description
- Add ec-curve parameter public key export/export-address subcommands
- Defaults to secp256k1
- Support secp256k1 and secp256r1
- Added a log message when creating a key with the selected ec curve name 

Note: due to the static initialization of the SignatureAlgorithmFactory, we need to ensure that its setup happens before any other key-related operation. This is why I have added it as the first instruction in the `run()` method in the subcommands.

## Fixed Issue(s)
fixes #3206 

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 ## 22.1.0
+- Add `--ec-curve` parameter to export/export-address public-key subcommands [#3333](https://github.com/hyperledger/besu/pull/3333)
 
 ### 22.1.0 Breaking Changes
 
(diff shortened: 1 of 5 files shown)


jframe (approved):
LGTM

PR #3344, opened 2022-01-28
Changed almost all test classes to MockitoJunitRunner

### LINKED ISSUE #2951: Use MockitoJUnitRunner for more test classes
### Description
Convert as many test classes as possible to `@RunWith(MockitoJUnitRunner.class)` instead of the default JUnit runner.

This will expose mocked function calls that are not called by the given tests. This mocks should either be deleted (when only mocked in a single test), or prefixed with `lenient()` if the same mock is shared in multiple tests of which only some use it.

This issue should be timeboxed; hard cases should be moved to separate issues.

If tests are identified as ineffective because they do not call a critical mock, this should be flagged (or fixed if easy).

### Acceptance Criteria
* As many test as possible run using `MockitoJUnitRunner`
* Tests classes that do not use `MockitoJUnitRunner` should be marked as such (so that it is easier later on to deal with them)
* Newly failing tests are fixed

TheSwarnim:
Signed-off-by: Swarnim Pratap Singh swarnimpratap132@gmail.com

## PR description
Changes remaining classes to MockitoJunitRunner test classes as much as possible in the main directory in besu/src/test/java/org/hyperledger/besu
 Only class files of ForkIdsTest(running with Parameterized.class) and PrivacyReordTest(Test failed with MockitoJunitRunner) were not converted

## Fixed Issue(s)

fixes #2951

## Changelog

### DIFF
--- a/besu/src/test/java/org/hyperledger/besu/BesuInfoTest.java
+++ b/besu/src/test/java/org/hyperledger/besu/BesuInfoTest.java
@@ -19,7 +19,10 @@ import static org.assertj.core.api.Assertions.assertThat;
 import java.util.Optional;
 
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
 
+@RunWith(MockitoJUnitRunner.class)
 public final class BesuInfoTest {
 
   /**
--- a/besu/src/test/java/org/hyperledger/besu/PrivacyTest.java
+++ b/besu/src/test/java/org/hyperledger/besu/PrivacyTest.java
@@ -55,7 +55,10 @@ import org.junit.After;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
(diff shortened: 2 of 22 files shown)


TheSwarnim:
Hi @macfarla , I've added files related to the mentioned issue. Please review it.

macfarla:
Hi @TheSwarnim please run spotless and then push the changes. I then expect some of these tests to fail because of unnecessary mocking, so you will need to be able to run the tests :) 
From there it might require a bit more digging to figure out what the failing tests were trying to accomplish - so it would be helpful if there are any that fall into the "too hard" bucket, to make a list of those and create a separate issue for them so we can tackle them gradually. (We don't want to just remove all the unused mocking code without considering why it was there.)

TheSwarnim:
Hi @macfarla, thanks for your comment. I actually forget to run the spotlessly but I've now successfully run and pushed the commit. Please review it.

macfarla (approved):
LGTM!

PR #3346, opened 2022-01-28
change dependencies for trivy scan step

macfarla:

small executor does not have sudo
but we don't need libSodium install in order to scan the docker image

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,8 +46,8 @@ commands:
             sudo apt-get update
             sudo apt-get install -y libsodium23 libsodium-dev apt-transport-https haveged libnss3-tools
             sudo service haveged restart
-      - restoreGradleCache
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,6 @@
 ---
 version: 2.1
 orbs:
-  slack: circleci/slack@3.4.2
```
siladu, line 4: How is slack related to this PR? Just checking it's an intentional change :)

macfarla: it's not - a previous version of this CI step was attempting to use slack but it's not setup in hyperledger - so I removed that code - but forgot to remove the orb

macfarla: so yes it's an intentional change :)

siladu (approved):
LGTM

PR #3326, opened 2022-01-25
Refactor to async retrieve blocks, and change peer when retrying to get a block

fab-10:
## PR description

`GetBlockFromPeersTask` is replaced with `RetryingGetBlockFromPeersTask` everywhere, since it is more effective in refreshing the peers while retrying.
`RetryingGetBlockFromPeersTask` has been changed to first try the peer that announce the block if known, and the try a different peer on every retry, and exit if the block has been downloaded by someone else in the meantime.
The task to get block from peers in `BlockPropagationManager` was blocking, and this had effect on the propagation of the `onBlockAdded` event, actually pausing it until the task was done, now the get block from peers task is async, and we keep track of the requests for non announced block, to avoid doing redoundant work.

A lot of debug/trace logs has been added to help the investigation on sync issues.

## Fixed Issue(s)

fixes #3304

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #3353, opened 2022-01-31
Add max allowed peers info to EthPeers so it can be easily accessible

fab-10:
Sometime is useful to know the max allowed number of peers, but this info
was not available in the synchronizer, it make sense to add that info to
the EthPeers class so it is available whenever you are working with peers.

## Changelog


fab-10:
closing since merged with https://github.com/hyperledger/besu/pull/3326

PR #3372, opened 2022-02-03
Persist last finalized block hash

### LINKED ISSUE #2913: Merge: Persist latest finalized block
### Description
Consensus layer will communicate via engineForkChoiceUpdated the current head and latest finalized blocks.  Presently the latest finalized block is only held in memory and will be absent on restart.  Besu needs to persist the latest finalized block when it receives it, and load this value on restart.

### Acceptance Criteria
* persist latest finalized block to underlying storage
* load finalized block from underlying storage on restart

fab-10:

## Fixed Issue(s)

fixes #2913 
## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueStoragePrefixedKeyBlockchainStorage.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueStoragePrefixedKeyBlockchainStorage.java
@@ -45,7 +45,7 @@ public class KeyValueStoragePrefixedKeyBlockchainStorage implements BlockchainSt
   private static final Bytes FINALIZED_BLOCK_HASH_KEY =
       Bytes.wrap("finalizedBlockHash".getBytes(StandardCharsets.UTF_8));
 
-  private static final Bytes CONSTANTS_PREFIX = Bytes.of(1);
+  private static final Bytes VARIABLES_PREFIX = Bytes.of(1);
   static final Bytes BLOCK_HEADER_PREFIX = Bytes.of(2);
   private static final Bytes BLOCK_BODY_PREFIX = Bytes.of(3);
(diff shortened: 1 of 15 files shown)


gezero (commented):
Looks good to me.

gezero (commented):
Looks Ok to me

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueStoragePrefixedKeyBlockchainStorage.java
```diff
@@ -72,6 +74,11 @@ public KeyValueStoragePrefixedKeyBlockchainStorage(
         .orElse(Lists.newArrayList());
   }
 
+  @Override
+  public Optional<Hash> getFinalized() {
+    return get(CONSTANTS_PREFIX, FINALIZED_BLOCK_HASH_KEY).map(this::bytesToHash);
```
garyschulte, line 79: I don't think there is any benefit to re-using the CONSTANTS_PREFIX.  I don't think it will hurt necessarily, but it might be clearer if we added a CONSENSUS_PREFIX.

garyschulte (commented):
LGTM. 
similar comment about CONSTANT_PREFIX.  Finalized hash is mutable, so the name is kinda eh.  It doesn't look like we are doing any kind of partitioning with the prefixes though, so it is mostly semantics.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueStoragePrefixedKeyBlockchainStorage.java
fab-10: I agree that the name CONSTANTS_PREFIX is wrong, because every value in it is actually changing, so I am renaming it to VARIABLES_PREFIX

PR #3384, opened 2022-02-07
refactor MergeCoordinator executeBlock to return validation reason

### LINKED ISSUE #3145: refactor MergeCoordinator executeBlock to return validation reason
### Description
[Kintsugi v3 execution spec](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.5/src/engine/specification.md#response) specifies that we should provide additional detail on a block execution that was marked `INVALID`.  Currently `MergeCoordinator.executeBlock()` simply supplies a boolean for success or failure, leveraging `BlockValidator.validateAndProcessBlock`.  

### Acceptance Criteria
* MergeCoordinator.executeBlock returns ValidationResult or similar Optional to indicate the success or failure of the execution, as well as any failure causes
* EngineExecuteBlock incorporates failure causes into the `validationError` field

fab-10:
## Fixed Issue(s)

fixes #3145 
## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
@@ -23,7 +23,6 @@ import org.hyperledger.besu.ethereum.core.TransactionReceipt;
 import org.hyperledger.besu.ethereum.mainnet.BlockBodyValidator;
 import org.hyperledger.besu.ethereum.mainnet.BlockHeaderValidator;
 import org.hyperledger.besu.ethereum.mainnet.BlockProcessor;
-import org.hyperledger.besu.ethereum.mainnet.BlockProcessor.Result;
 import org.hyperledger.besu.ethereum.mainnet.HeaderValidationMode;
 
 import java.util.ArrayList;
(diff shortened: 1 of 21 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -64,64 +63,70 @@ public MainnetBlockValidator(
...
       badBlockManager.addBadBlock(block);
-      return Optional.empty();
+      return new Result(errMsg);
     }
     final BlockHeader parentHeader = maybeParentHeader.get();
 
     if (!blockHeaderValidator.validateHeader(header, parentHeader, context, headerValidationMode)) {
       badBlockManager.addBadBlock(block);
-      return Optional.empty();
+      return new Result("Invalid header for block " + block.toLogString());
```
fab-10, line 92: @garyschulte is it enough this generic `Invalid header for block` message or we should popup more details from `blockHeaderValidator.validateHeader`?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -64,64 +63,70 @@ public MainnetBlockValidator(
...
+      return new Result(errMsg);
     }
     final MutableWorldState worldState = maybeWorldState.get();
 
-    final Result result = processBlock(context, worldState, block);
+    final BlockProcessor.Result result = processBlock(context, worldState, block);
     if (result.isFailed()) {
       badBlockManager.addBadBlock(block);
-      return Optional.empty();
+      return new Result("Error processing block");
```
fab-10, line 115: @garyschulte is it enough this generic `Error processing block` message or we should popup more details from `processBlock` method?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -64,64 +63,70 @@ public MainnetBlockValidator(
...
     }
 
     List<TransactionReceipt> receipts = result.getReceipts();
     if (!blockBodyValidator.validateBody(
         context, block, receipts, worldState.rootHash(), ommerValidationMode)) {
       badBlockManager.addBadBlock(block);
-      return Optional.empty();
+      return new Result("Block body not valid");
```
fab-10, line 122: @garyschulte same here, is it enough this generic `Block body not valid` message or we should popup more details from `blockBodyValidator.validateBody` method?

garyschulte: "Block body not valid" is good enough for the error message we are returning to the consensus client.  More detail is better but it seems wise to limit the scope of the signature changes.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
garyschulte: good enough IMO.  If we need more specifics we can always implement it, but the api we are trying to support is not user facing nor is the consumer of this api user facing.  The only real motivation is better logging IMO

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
garyschulte: :+1:

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
fab-10: same as above

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
fab-10: same as above

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
fab-10: Ok, then we can stop here for the moment, I have just made un update to extract the failure handling code in a method, so the code is cleaner.

garyschulte (commented):
posthumous 👍

PR #3391, opened 2022-02-08
MERGE: new payload API for Kiln testnet

daniellehrner:
## Changelog


jflo (approved):
So far so good! minor comments.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadTest.java
```diff
@@ -225,10 +240,10 @@ public void shouldRespondWithSyncingDuringBackwardsSync() {
     BlockHeader mockHeader = new BlockHeaderTestFixture().buildHeader();
     var resp = resp(mockPayload(mockHeader, Collections.emptyList()));
 
-    EngineExecutionResult res = fromSuccessResp(resp);
+    EnginePayloadStatusResult res = fromSuccessResp(resp);
```
garyschulte, line 243: we need the backward syncing mock, and also there is a peculiarity about baseFee since it is optional - you have to specify it for the hash to match

```suggestion
    when(mergeCoordinator.latestValidAncestorDescendsFromTerminal(any(BlockHeader.class)))
        .thenReturn(true);
    when(mergeCoordinator.isBackwardSyncing())
        .thenReturn(Boolean.TRUE);
    BlockHeader mockHeader = new BlockHeaderTestFixture().baseFeePerGas(Wei.ONE).buildHeader();
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
```diff
@@ -116,66 +112,75 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
-    }
-
     // TODO: post-merge cleanup
     if (!mergeCoordinator.latestValidAncestorDescendsFromTerminal(newBlockHeader)) {
-      return new JsonRpcErrorResponse(
-          requestContext.getRequest().getId(), JsonRpcError.INVALID_TERMINAL_BLOCK);
+      return respondWith(requestContext.getRequest().getId(), null, INVALID_TERMINAL_BLOCK);
+    }
+
+    if (mergeContext.isSyncing()) {
```
garyschulte, line 146: Unlike kintsugi, kiln spec says we should respond with SYNCING if we are syncing at all (whether backward or forward).  

Backward sync is tracked separately in mergeCoordinator, so we can handle that differently.  We should append the newPayload hash to the backward sync.  Something like:
```suggestion
    if (mergeContext.isSyncing() || mergeCoordinator.ifBackwardSyncing(newBlock)) {
      ...
```
 and in mergeCoordinator something like:
`public boolean ifBackwardsSyncing(Block blockToAppend) {}`

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadTest.java
garyschulte: the mock for isBackwardSyncing would change to be for ifBackwardSyncing(...) if/when you implement the mergeCoordinator suggestion

garyschulte (changes requested):
we need to account for backward syncing.  Otherwise LGTM

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
garyschulte: The other way to implement the append behavior would be to let it flow into `executeBlock` like we did in kintsugi, and rather than returning a validation error, we let the append happen there and return a `Result` from `mergeCoordinator.executeBlock(..)` that indicates we are backward syncing

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadTest.java
daniellehrner: I've adapted the test now

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
daniellehrner: I've for now added a stub method:
```
public boolean isBackwardSyncing(final Block blockToAppend) {
    return isBackwardSyncing();
  }
```

Do you think that is enough for now?

jflo (approved):

garyschulte (approved):
🚢 !

PR #1743, opened 2021-01-04
Don't Reject Local Transactions Below Gas Minimum

RatanRSur:
### Fixed Issue(s)
fixes #1480 

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #3397, opened 2022-02-10
Reject locally-sourced transactions below the minimum gas price when not mining

antonydenyer:
## PR description
Reject locally-sourced frontier transactions below the minimum gas price when not mining.

Brings back some of the functionality that was taken out by https://github.com/hyperledger/besu/pull/1743

The main problem is that if you're not mining and send zero priced frontier transactions to the node it will fill up the transaction pool. 

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -152,6 +156,14 @@ public boolean addTransactionHash(final Hash transactionHash) {
     return validationResult;
   }
 
+  private boolean zeroGasPriceFrontierTransaction(final Transaction transaction) {
```
macfarla, line 159: lessThanMinGasPriceFrontierTransaction?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolTest.java
```diff
@@ -986,6 +989,46 @@ public void shouldRejectGoQuorumTransactionWithNonZeroValue() {
...
+    givenTransactionIsValid(transaction);
+
+    final ValidationResult<TransactionInvalidReason> result =
+        transactionPool.addLocalTransaction(transaction);
+
+    assertThat(result).isEqualTo(ValidationResult.valid());
+  }
+
+  @Test
+  public void shouldRejectZeroGasPriceFrontierTransactions() {
```
macfarla, line 1013: WhenNotMining

ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/RetestethContext.java
```diff
@@ -206,7 +206,7 @@ private boolean buildContext(
             retestethClock,
             metricsSystem,
             syncState,
-            Wei.ZERO,
+            new MiningParameters.Builder().build(),
```
macfarla, line 209: everywhere else you've replaced this with
            new MiningParameters.Builder().minTransactionGasPrice(Wei.ZERO).build(),

macfarla (commented):
couple of renames suggested and some nits

ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/RetestethContext.java
antonydenyer: Ahh yes, was attempting to be explicit (the default is Wei.ZERO)

macfarla:
I think it looks ok but maybe someone with more mainnet/miner chops should review also?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -134,6 +135,11 @@ public boolean addTransactionHash(final Hash transactionHash) {
     final ValidationResult<TransactionInvalidReason> validationResult =
         validateLocalTransaction(transaction);
     if (validationResult.isValid()) {
+      if (!miningParameters.isMiningEnabled()
```
garyschulte, line 138: FYI in the merge branch we use MiningParameters enabled/disabled to start and stop the miner at the transition from PoW to PoS.  I think using miningEnabled for this is fine, it just has unintended consequences post-merge.

garyschulte (changes requested):
Seems safe, but I think we should prevent un-executable transactions on mainnet by making sure we check the transaction meets the bare-minimum baseFee requirement, which IIRC is 7 wei

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -134,6 +135,11 @@ public boolean addTransactionHash(final Hash transactionHash) {
...
+          && transaction.getType() == TransactionType.FRONTIER
+          && gasPriceIsLessThanMinGasPrice(transaction)) {
```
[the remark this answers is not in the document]
antonydenyer: :thinking: interesting didn't know that.
 
I guess the problem we're trying to solve is someone adding a transaction to their transaction pool and it gets stuck forever. I guess at this point in time when you submit the transaction, you can only know some things with certainty. Base fee and type of fee market can not be guaranteed as you don't know what the block height will be when the transaction gets mined. We could add something like (in a separate pr :wink: ) a slippage cli parameter for fees, so when you submit a transaction you need to be within x% of recommended fees.

What do you think?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
antonydenyer: Good point, there's probably a bunch of others. The specific scenario we want to allow is if I'm mining on a PoW network I want to be able to submit 'free' transactions to my pool because I'm going to mine them myself. Maybe we will end up with `isStakingEnabled`?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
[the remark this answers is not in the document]
garyschulte: I get that we can't know the current baseFee at the time of submission, but we do know with certainty that there is a basefee floor below which a transaction can never be executed due to the way the 1559 basefee adjustments work.  On a baseFee network (with london parameters) a transaction with a basefee below 7 wei is unexecutable, ever.  

This outcome isn't mentioned in the EIP, but is just inherent to the parameters of London.  This is the only resource I could find a link to that mentions it: 
https://hackmd.io/@tvanepps/1559-wallets#Variable-Mechanics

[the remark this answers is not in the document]
garyschulte: I see your point about block height however - we don't know that a transaction won't be valid in a future fork.  But I think that is a bit of a stretch to accept transactions that might be executable in a future hardfork 🤔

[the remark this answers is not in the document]
antonydenyer: Yeah, I guess we can reasonably say that if we're currently past that height we are never going back so accept the transaction to the pool (or reject if the fees are too low).

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/TransactionTestFixture.java
```diff
@@ -30,7 +30,7 @@
 
   private long nonce = 0;
 
-  private Wei gasPrice = Wei.of(5);
+  private Wei gasPrice = Wei.of(5000);
```
garyschulte, line 33: I am surprised this didn't cause more test fallout ☘️

garyschulte (approved):
Good test coverage, looks :+1:

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/TransactionTestFixture.java
antonydenyer: Looks like the places that care about it set it explicitly!

Issue #3335, opened 2022-01-27
NullPointerException: Error while creating packet


fab-10:
### Description

```
{
	"timestamp": "2022-01-27T13:44:13,932",
	"level": "ERROR",
	"thread": "vert.x-eventloop-thread-1",
	"class": "PeerDiscoveryController",
	"message": "Error while creating packet",
	"throwable": " java.util.concurrent.CompletionException: java.lang.NullPointerException
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:718)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2073)
	at org.hyperledger.besu.ethereum.p2p.discovery.VertxPeerDiscoveryAgent$VertxAsyncExecutor.lambda$execute$1(VertxPeerDiscoveryAgent.java:262)
	at io.vertx.core.impl.future.FutureImpl$3.onSuccess(FutureImpl.java:141)
	at io.vertx.core.impl.future.FutureBase.lambda$emitSuccess$0(FutureBase.java:54)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:164)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:469)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:503)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:986)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.NullPointerException
	at org.hyperledger.besu.ethereum.p2p.discovery.VertxPeerDiscoveryAgent.sendOutgoingPacket(VertxPeerDiscoveryAgent.java:160)
[...]

PR #3405, opened 2022-02-11
fix NPE, issue 3335

pinges:
https://github.com/hyperledger/besu/issues/3335

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
@@ -279,7 +279,6 @@ public abstract class PeerDiscoveryAgent {
   }
 
   protected void handleIncomingPacket(final Endpoint sourceEndpoint, final Packet packet) {
-    // TODO: check whether
     final int udpPort = sourceEndpoint.getUdpPort();
     final int tcpPort =
         packet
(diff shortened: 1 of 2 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -279,6 +279,7 @@ protected boolean validatePacketSize(final int packetSize) {
   }
 
   protected void handleIncomingPacket(final Endpoint sourceEndpoint, final Packet packet) {
+    // TODO: check whether
```
macfarla, line 282: stray TODO

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/VertxPeerDiscoveryAgent.java
```diff
@@ -148,26 +148,31 @@ protected void handleListenerSetup(
...
-          if (ar.failed()) {
-            result.completeExceptionally(ar.cause());
-          } else {
-            result.complete(null);
-          }
-        });
+    final CompletableFuture<Void> result = new CompletableFuture<>();
+    if (socket == null) {
+      result.completeExceptionally(
+          new RuntimeException("Discovery socket already closed, because Besu is closing down"));
```
frankisawesome, line 162: Is a RuntimeException appropriate here or something more specific?

frankisawesome (changes requested):
LGTM, but I would still like to see a test for this method here

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/VertxPeerDiscoveryAgent.java
pinges: This should only happen when we are closing down (we have called stop in the VertxPeerDiscoveryAgent, which sets the socket to null), but there is still a message in the system that causes  sendOutgoingMessage() to be called.

mark-terry:
Does this specifically address the reported issue, or just add handling around the case where the socket may be null?

frankisawesome (approved):
Discussed on zoom, we will make an issue for missing test file for this class :)

Issue #3268, opened 2022-01-12
Implement post-TTD full sync strategy


garyschulte:
### Description
Currently besu leverages the typical basic forward sync strategies for the merge.  Meaning we use full-sync, fast-sync and snap-sync and will sync up to the latest head block from our peers.  However post-merge, when besu syncs a new instance it should stop at the terminal-total-difficulty block and await direction from a consensus client regarding what target to forward-sync to.

Handing the concurrent sync of execution and consensus clients is still an evolving topic, but to mitigate attacks and zombie mining that might happen post-merge, we can and should implement a different post-TTD sync strategy.  

Implementing this strategy has an added benefit of making bonsai tries more viable for the merge, as we will not have to do an expensive rewind operation to get back to TTD once we start processing forkchoiceUpdated calls from the consensus client.  

### Acceptance Criteria

* incorporatea configured terminal block number and block hash as an initial sync target,
** in the absence of TTD block number/hash, continue to use heaviest chain criteria for which chain to sync
* stop initial sync at terminal difficulty, and mark the chain as sync'd until otherwise directed by consensus client
* taking direction from consensus engine via forkchoiceUpdated for the chain and head to forward sync up to

ajsutton:
Given the way fast and snap sync work, it may make sense to not start sync at all if peers are advertising a total difficulty in excess of TTD until the CL client sends the first `forkChoiceUpdated` notification which would identify the chain to sync.  If you sync'd the TTD block you'd likely download a very old world state and then either have to full sync from there or discard it and start a new fast/snap sync from where the CL starts from.

If the CL is doing a full sync of the beacon chain, that first notice will be for the TTD block  and the CL would then execute each block past that, but when the CL is using checkpoint sync it's starting sync will be close to the chain head which may be very significantly past the TTD block and the CL will not execute any blocks prior to the checkpoint state it starts with.

garyschulte:
marking this issue as complete after changing the acceptance criteria and breaking it down into #3506, #3507, #3512, #3513

PR #3423, opened 2022-02-14
FullSync Future should stop when total terminal difficulty is reached

gezero:

The Full sync future should finish executing once the terminal total difficulty is reached. Note that we will not stop syncing exactly at the TTD but we will finish syncing the current batch of blocks.

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
@@ -119,7 +119,7 @@ public class BlockPropagationManager {
       clearListeners();
       started.set(false);
     } else {
-      LOG.info("Attempted to stop when we are not even running...");
+      LOG.warn("Attempted to stop when we are not even running...");
     }
   }
 
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
@@ -82,6 +82,9 @@ public class BlockPropagationManager {
   private final Set<Long> requestedNonAnnouncedBlocks =
       Collections.newSetFromMap(new ConcurrentHashMap<>());
   private final PendingBlocksManager pendingBlocksManager;
+  private Optional<Long> onBlockAddedSId = Optional.empty();
+  private Optional<Long> newBlockSId;
+  private Optional<Long> newBlockHashesSId;
 
   BlockPropagationManager(
       final SynchronizerConfiguration config,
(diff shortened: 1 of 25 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fullsync/FullSyncTargetManager.java
```diff
@@ -105,6 +109,9 @@ private boolean isSyncTargetReached(final EthPeer peer) {
 
   @Override
   public boolean shouldContinueDownloading() {
-    return true;
+    return terminalTotalDifficulty.isEmpty()
+        || terminalTotalDifficulty
+            .get()
+            .greaterThan(protocolContext.getBlockchain().getChainHead().getTotalDifficulty());
```
garyschulte, line 115: I can't help but want to suggest:

```
    return terminalTotalDifficulty.filter(ttd -> 
        ttd.lessThan(
            protocolContext
                .getBlockchain()
                .getChainHead()
                .getTotalDifficulty()))
        .isPresent();

```

But perhaps yours is more readable

garyschulte (approved):
LGTM.  

Feel free to disregard opinionated suggestion to use streaming Optional.

garyschulte:
One other comment - we need to follow this up with a post-ttd sync method pretty soon, otherwise we are going to be unable to sync kintsugi.  Backward sync alone won't work for networks that have a long post-merge history

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fullsync/FullSyncTargetManager.java
gezero: I am migrating this part to be more flexible. The terminating condition will be possible to change to a target block hash and even change it during the sync.

gezero:
I am only going to merge the changes here, when I will have the follow-up sync ready.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fullsync/FullImportBlockStep.java
```diff
@@ -35,18 +35,25 @@
...
     this.protocolSchedule = protocolSchedule;
     this.protocolContext = protocolContext;
     this.ethContext = ethContext;
+    this.fullSyncTerminationCondition = fullSyncTerminationCondition;
   }
 
   @Override
   public void accept(final Block block) {
+    if (fullSyncTerminationCondition.shouldStopDownload()) {
+      LOG.info("Not importing another block, because terminal condition was reached.");
```
garyschulte, line 54: This guy needs to be DEBUG.  Running on devnet4 the logs are getting spammed for every block post TTD until the pipeline finishes (at the current peers' head).

maybe this could be emitted at info level if `block.number() mod 100 == 0 ` ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fullsync/FullSyncTargetManager.java
```diff
@@ -105,6 +108,6 @@ private boolean isSyncTargetReached(final EthPeer peer) {
 
   @Override
   public boolean shouldContinueDownloading() {
-    return true;
+    return terminationCondition.shouldContinueDownload();
```
garyschulte, line 111: merging this pr onto a devnet4 branch, I am still seeing the pipeline run all the way to the head of peers (with no-op imports).  But when I restart the client, it correctly does not try to sync beyond TTD.  It seems like we are not evaluating this during a running pipeline.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -111,12 +114,37 @@ public void start() {
     }
   }
 
+  public void stop() {
+    if (started.get()) {
+      clearListeners();
+      started.set(false);
+    } else {
+      LOG.info("Attempted to stop when we are not even running...");
```
garyschulte, line 122: this doesn't seem like an INFO level log

garyschulte (changes requested):
I tested with this PR on devnet4 with good, but mixed results regarding the initial full-sync.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fullsync/FullImportBlockStep.java
gezero: Ok I'll make it debug.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fullsync/FullSyncTargetManager.java
gezero: The pipeline should now be stopping properly when the terminal condition is reached.

garyschulte (approved):
This fixes #3268 .  Remaining sync issues have been split into separate issues

Issue #3385, opened 2022-02-07
Implement updated semantics for forkchoiceUpdatedV1 from kiln spec


garyschulte:
### Description
Implement updated execution spec for forkchoiceUpdatedV1, corresponding to execution engine spec 1.6
https://hackmd.io/@n0ble/kiln-spec

### Acceptance Criteria
* implement [forkchoiceUpdatedV1 spec 1.6](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.6/src/engine/specification.md#specification-1)
* mirror spec names for structures:
  * rename ExecutionForkChoiceParameter to EngineForkchoiceStateV1Parameter
  * rename ExecutionPayloadAtttributesParameter to EnginePayloadAttributesV1Parameter

PR #3437, opened 2022-02-15
forkchoiceUpdated implement kiln/1.0-alpha6 execution spec 

garyschulte:
## PR description
* address merge-at-genesis case for terminal-proof-of-work checks
* update forkchoice status field json to reflect kiln spec
* use `INVALID` responses according to spec rather than rpc errors (HTTP 400)
* return a result from forkchoiceUpdated rather than void return with exceptions
* unit tests and mock fixes for existing tests

https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.6/src/engine/specification.md

## Fixed Issue(s)

#3385 

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineUpdateForkChoiceResult.java
```diff
@@ -17,28 +17,27 @@
...
 public class EngineUpdateForkChoiceResult {
-  private final ForkChoiceStatus status;
+  private final EnginePayloadStatusResult payloadStatus;
   private final PayloadIdentifier payloadId;
 
   public EngineUpdateForkChoiceResult(
-      final ForkChoiceStatus status, final PayloadIdentifier payloadId) {
-    this.status = status;
+      final EnginePayloadStatusResult payloadStatus, final PayloadIdentifier payloadId) {
+    this.payloadStatus = payloadStatus;
```
daniellehrner, line 34: As forkchoiceUpdated only allows a subset of `EnginePayloadStatusResult.status`, I think I would verify that nobody uses an invalid status here. I think throwing an exception if the status is `ACCEPTED` or `INVALID_BLOCKHASH` should be sufficient.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinatorTest.java
```diff
@@ -327,7 +333,7 @@ public void ancestorIsValidTerminalProofOfWork() {
 
   @Test
   public void ancestorNotFoundValidTerminalProofOfWork() {
-    final long howDeep = MergeCoordinator.MAX_TTD_SEARCH_DEPTH + 1;
+    final long howDeep = MergeCoordinator.MAX_TTD_SEARCH_DEPTH + 2;
```
daniellehrner, line 336: Just for my understanding: Why do we need + 2 here instead of + 1?

daniellehrner (changes requested):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EnginePayloadStatusResult.java
```diff
@@ -0,0 +1,52 @@
...
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.ExecutionEngineJsonRpcMethod.ExecutionStatus;
+
+import java.util.Optional;
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonPropertyOrder({"status", "latestValidHash", "validationError"})
+public class EnginePayloadStatusResult {
+  ExecutionStatus status;
```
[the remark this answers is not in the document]
garyschulte: Yes, that is in-line with the effort conform to "Engine" rather than the mix of execution and engine in class and enum names 👍

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineUpdateForkChoiceResult.java
garyschulte: I wasn't really keen on combining the two enums, but the execution spec specifically called out that both use the same payloadStatusV1 struct.  I can add some guard rails for those two invalid response status vals 👍

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinatorTest.java
garyschulte: Because we start the search from the parent so we need the chain depth to be at longer than max_depth + 1, so +2 ;)

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -297,10 +299,17 @@ public void updateForkChoice(final Hash headBlockHash, final Hash finalizedBlock
           blockchain.setFinalized(blockHeader.getHash());
           mergeContext.setFinalized(blockHeader);
         });
+
+    return ForkchoiceResult.withResult(newFinalized, Optional.of(newHead));
   }
 
   @Override
   public boolean latestValidAncestorDescendsFromTerminal(final BlockHeader blockHeader) {
+    if (blockHeader.getNumber() <= 1L) {
```
jflo, line 308: i do not trust this, but i believe it. I can't think of a counter to this assumption.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyValidationTest.java
```diff
@@ -56,6 +56,7 @@ public void difficultyMustBeZero() {
 
     final BlockHeader invalidHeader = mock(BlockHeader.class);
     when(invalidHeader.getDifficulty()).thenReturn(Difficulty.of(10));
+    when(invalidHeader.getNumber()).thenReturn(1337L);
```
jflo, line 59: Tell me it's Gary without telling me it's Gary :)

jflo (approved):

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -243,50 +243,52 @@ public Result executeBlock(final Block block) {
...
-    newFinalized
-        .map(Optional::of)
-        .orElse(currentFinalized)
-        .ifPresent(
-            finalized -> {
-              if (!isDescendantOf(finalized, newHead)) {
-                throw new IllegalStateException(
+    Optional<String> descendantError =
+        newFinalized
+            .map(Optional::of)
```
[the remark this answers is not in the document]
garyschulte: This is a cheeky way to combine two optionals.  We are left with the `finalized` lambda parameter being either newFinalized or currentFinalized

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
garyschulte: Yeah, blockNumber could be negative which would be hugely pathological.  We could make this `== 0` instead

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/headervalidationrules/NoDifficultyValidationTest.java
garyschulte: lol

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
garyschulte: whoops scratch that - I forgot this condition is checking for blocknumber in [0,1]

garyschulte: I added a bit of safety to ensure we are past TTD.

daniellehrner (approved):

jflo (approved):

Issue #3003, opened 2021-11-04
Allow BftMiningCoordinator to be shutdown from same thread


jframe:
The BftMiningCoordinator has a shutdown method to stop processing of events by stopping the BftProcessor and BftExecutors. This doesn't work if it is called from the same Thread as the BftMiningCoordinator is running on.

**Acceptance Critieria**
* BftMiningCoordinator can stop regardless of what the calling thread is

**How To Reproduce**

1. Set a single node IBFT to QBFT migration
2. Have a migration to the QBFT
3. The BftMiningCoordinator should attempt to shutdown the IBFT BftMiningCoordinator
4. At this point the we fail to migrate due the shutdown hanging

**Notes**
* The MigratingMiningCoordinator calls the stop method on the BftMiningCoordinators when the migration is reached. This happens in the onBlockAdded method.
* It's important that the stop method remains synchronous

PR #3438, opened 2022-02-15
QBFT Migration: allow active mining coordinator to be shutdown

siladu:
Simple solution with async tasks but not easily testable. Compromised the unit tests in favour of simplicity/readability, given that the migration happy path will also be covered by acceptance tests.

~This draft can be compared with https://github.com/hyperledger/besu/pull/3478 which removes the problem by removing the bftProcessor's shutdownLatch.~ (closed this PR as came to the conclusion this solution had safety problem in the migration scenario)

The is a proposed fix for https://github.com/hyperledger/besu/issues/3003


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/MigratingMiningCoordinator.java
```diff
@@ -127,17 +128,27 @@ public void onBlockAdded(final BlockAddedEvent event) {
...
+      final Runnable startNextCoordinatorTask =
+          () -> {
+            activeMiningCoordinator = nextMiningCoordinator;
+            startActiveMiningCoordinator();
+            if (activeMiningCoordinator instanceof BlockAddedObserver) {
+              ((BlockAddedObserver) activeMiningCoordinator).onBlockAdded(event);
+            }
+          };
+
+      CompletableFuture.runAsync(stopActiveCoordinatorTask).thenRun(startNextCoordinatorTask);
```
lucassaldanha, line 149: What happens if the application tries to get hold of the mining coordinator in between `stopActiveCoordinatorTask` and `startNextCoordinatorTask`? We will end up giving them a mining coordinator instance that is in the process of being stopped.

The `stop()` implementation on `AbstractMiningCoordinator` is synchronized, I wonder if running as an async callback might introduce some side effects that we haven't thought of.

lucassaldanha: Although, it looks like we synchronize it just for the sake of keeping the MiningCoordinator internal state consistent and to avoid concurrent `shutdown()` calls to the executor, etc. So it should be safe 🤷

siladu: > What happens if the application tries to get hold of the mining coordinator in between stopActiveCoordinatorTask and startNextCoordinatorTask? 

We delegate to BftMiningCoordinator.stop() which handles graceful shutdown:
https://github.com/hyperledger/besu/blob/93a599abf272f3214911df15d2b11b80b8bf3014/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/BftMiningCoordinator.java#L86-L100

It waits for the bftProcessor to stop before shutting the executor down.
Not sure if that's what you mean by the application getting hold of mining coordinator?

BftProcessor is run on a single threaded executor so shouldn't need to synchronise (although that's what's causing the unintended consequences I'm attempting to fix). I think AbstractMiningCoordinator tasks are async but would need to double check.

siladu: By the way, AbstractMiningCoordinator isn't used by the BFT implementations.

jframe: Think is fine. It's possible that the some the MiningCoordinator methods could be called whilst in the process of one shutting down but couldn't see any problems with that. Some messages will probably be dropped but as soon as we start this we are done before the original consensus mechanism so think it's ok.

consensus/common/src/test/java/org/hyperledger/besu/consensus/common/MigratingMiningCoordinatorTest.java
```diff
@@ -122,28 +123,16 @@ public void stopShouldUnregisterThisCoordinatorAsObserver() {
...
-            createCoordinatorSchedule(coordinator1, delegateCoordinator), blockchain)
-        .onBlockAdded(blockEvent);
-
-    verify(blockchain).observeBlockAdded(delegateCoordinator);
-    verify(blockchain).removeObserver(2L);
-  }
-
-  @Test
-  public void onBlockAddedShouldMigrateToNextMiningCoordinatorAndDelegate() {
+  public void onBlockAddedShouldNotThrowExceptionWhenMigrating() {
```
jframe, line 126: Don't think this test is worth having as it is

siladu: Removed

jframe (approved):

siladu:
I sign off to the above commit

PR #3443, opened 2022-02-16
QBFT Migration: allow active mining coordinator to be shutdown (with coverage)

siladu:
A better tested but more complex version of the solution from https://github.com/hyperledger/besu/pull/3438.

Pulled code apart to get decent unit test coverage.

https://github.com/hyperledger/besu/issues/3003


siladu:
I don't want to progress with this solution, another alternative solution is being tested: https://github.com/hyperledger/besu/pull/3478

PR #3458, opened 2022-02-18
Update discovery to 22.2.0

ajsutton:
## PR description
Update to the latest version of discovery (22.2.0). Main change is to switch RLP parsing to use tuweni-rlp instead of web3j which avoids needing to create intermediate objects during the parsing process.

## Changelog


macfarla (approved):
LGTM

PR #3467, opened 2022-02-20
Quorum mode ATs - don't run in parallel

macfarla:
Quorum ATs do not run well in parallel. This PR removes the parallel gradle flag. Aiming to reduce flakiness of these tests.

## Changelog


.circleci/config.yml
```diff
@@ -237,15 +245,14 @@ jobs:
...
-    executor: xl_machine_executor
+    executor: quorum_ats_executor_xl
     steps:
       - prepare
       - attach_workspace:
           at: ~/project
       - run:
           name: Quorum Acceptance Tests
-          no_output_timeout: 40m
+          no_output_timeout: 30m
```
frankisawesome, line 255: I'm not sure if I'm understanding this correctly but it seems like we're changing the step from parallel to sequential but reducing the timeout? Shouldn't the timeout be longer (although 30m seems quite excessive)

macfarla: Fair point but this timeout is only used if no output is received for that length of time. This only happens if something goes seriously wrong.

fab-10 (approved):

PR #3472, opened 2022-02-21
Migrate JumpDestCache to CodeCache

shemnon:
## PR description

Migrate the jump dest analysis cache into a more generic code cache that
lives inside the EVM.

This is in preparation for Ethereum Object Formats where a string of
code may be treated differently depending on what EVM version is
executing the code. Newer versions will also have difference analyses to
run that will need different backing data structures.

## Changelog


evm/src/main/java/org/hyperledger/besu/evm/fluent/EVMExecutor.java
```diff
@@ -302,7 +302,10 @@ public EVMExecutor code(final Code code) {
   }
 
   public EVMExecutor code(final Bytes codeBytes, final Hash hash) {
-    this.code = new Code(codeBytes, hash);
+    if (evm == null) {
```
jflo, line 305: Could we add a constructor for the Executor and make the EVM a final parameter on it? Seems like that is a safe thing to make a hard dependency for this class.

jflo (approved):
LGTM, one suggestion with regard to binding the EVM to the Executor a little harder.

PR #3474, opened 2022-02-21
use med executor for quorum ATs

macfarla:

The nightly CI failed overnight https://app.circleci.com/pipelines/github/hyperledger/besu/13408/workflows/e13faa55-846d-40cb-a18a-cb96eeff6d2f/jobs/74709
So this PR is 
* changing the executor back to the machine executor that it was using before
* changing from XL to MED
* removing extra gradle workers config

## Changelog


.circleci/config.yml
```diff
@@ -28,13 +28,11 @@ executors:
     environment:
       GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.parallel=true -Dorg.gradle.workers.max=4
       
-  quorum_ats_executor_xl: # 8cpu, 16G ram
-    docker:
-      - image: cimg/openjdk:11.0
-    resource_class: xlarge
+  quorum_ats_executor_med:
+    machine:
+      image: ubuntu-2004:202101-01
```
frankisawesome, line 33: The image here is changed from openjdk to ubuntu, is this intended?

.circleci/config.yml
```diff
@@ -268,9 +266,8 @@ jobs:
       - capture_test_logs
 
   acceptanceTestsQuorum:
-    executor: quorum_ats_executor_xl
+    executor: quorum_ats_executor_med
     steps:
-      - prepare
```
frankisawesome, line 273: Is there a good reason the prepare step is removed?

frankisawesome (changes requested):

.circleci/config.yml
macfarla: yes. updated the PR desc with extra info

.circleci/config.yml
macfarla: I don't think we need it. this CI just uses the develop docker image which is already built elsewhere

fab-10 (approved):

PR #3476, opened 2022-02-22
Add some thread names for debugging clarity

siladu:
BFT executor threads include IBFT or QBFT to help with the IBFT -> QBFT migration.


jframe (approved):
LGTM

Issue #2889, opened 2021-10-12
implement trace_get API


matkt:

macfarla:
returning a subset of trace_transaction

macfarla:
this one is done apart from `creationMethod` field which is included in besu response but not open ethereum response

macfarla:
Added an example to the docs ticket https://github.com/hyperledger/besu-docs/issues/963

PR #3477, opened 2022-02-22
trace_get 

mark-terry:
Signed-off-by: Mark Terry <mark.terry@consensys.net>

## PR description
Implementation for trace_get API.
* exclude revertReason from result (matches Besu trace_call and openethereum)

Note this implementation matches openethereum behaviour except for:
* we manually added `creationMethod` to test files even though openethereum 3.2.5 does not include this field (this matches Besu trace_transaction implementation which also includes the `creationMethod` field)

See #2889 

## Changelog


macfarla:
The unit tests failures are explained by: the version of openethereum which generated these test files (3.2.5 installed via homebrew) doesn't have this patch https://github.com/openethereum/openethereum/issues/114

matkt:
regarding the revert reason I think it was added voluntarily even if OE does not manage it. We should ask @shemnon  if he has more context on it.

matkt:
if creationMethod is not needed in this case why we are not using this field ? https://github.com/hyperledger/besu/blob/b0de73fe57bfd8c398939b8e7edef577dd773e04/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java#L233

macfarla:
> regarding the revert reason I think it was added voluntarily even if OE does not manage it. We should ask @shemnon if he has more context on it.

for revert reason we have decided to leave as is for now, and made a new ticket #3515 to revisit it.

macfarla:
> if creationMethod is not needed in this case why we are not using this field ?
> 
> https://github.com/hyperledger/besu/blob/b0de73fe57bfd8c398939b8e7edef577dd773e04/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java#L233

we decided to include this extra field based on conversation with @mbaxter who is waiting for these APIs to be implemented.

pinges (approved):
LGTM

pinges (commented):
I think we should fix the code smells from Sonar Cloud if they make sense.

PR #3478, opened 2022-02-22
Remove shutdown latch from BftProcessor

siladu:
We don't want BftMiningCoordinator.stop() to wait for BftProcessor to shutdown before calling shutdown on the BftExecutor.
This prevents BftMiningCoordinator from being able to be shutdown from the same thread.
We want to enable shutdown from the same thread to support migrating from one MiningCoordinator instance to another (IBFT -> QBFT).

This is the relevant discussion as to why the latch was originally added in: https://github.com/hyperledger/besu/pull/104#discussion_r334705699

I believe the latch is unnecessary because BftExecutors.stop() (which calls bftProcessExecutor.shutdownNow) should be sufficient to safely prevent any further tasks from being accepted, whilst BftExecutors.awaitStop() (calling bftProcessExecutor.awaitTermination()) gives the current task(s) a chance to finish.

This PR is most if not all of the solution to https://github.com/hyperledger/besu/issues/3003

### DIFF
--- a/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BftProcessor.java
+++ b/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BftProcessor.java
@@ -17,7 +17,6 @@ package org.hyperledger.besu.consensus.common.bft;
 import org.hyperledger.besu.consensus.common.bft.events.BftEvent;
 
 import java.util.Optional;
-import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


siladu:
Closing this in favour of #3438

Although we probably don't need the latch for the 'shutdown besu' scenario, it doesn't fully solve the QBFT migration scenario. It was only worth risking this change if it was going to help resolve #3003 but I don't think it's possible to safely shutdown the bftProcessor/Executor from its own thread, we need another thread to manage this.

PR #3486, opened 2022-02-23
Handle TTD and mining pre and post merge 

### LINKED ISSUE #2898: Merge: Consistently Handle Terminal Total Difficulty
### Description
When the terminal proof-of-work block is reached ensure we:
 * stop any mining activities
 * disable sending and handling of NEW_BLOCK_HASHES and NEW_BLOCK messages 
 * stop sync process

TTD is part of genesis, but should be override-able via the CLI 

### Acceptance Criteria
* correct behavior at TTD for the above cases
* correct behavior for a re-org below TTD
* test coverage
* create a terminal-total-difficulty CLI override parameter which overrides the TTD from genesis config file

### Additional Information
Much of this behavior is implemented already, but there are corner cases and no test coverage

garyschulte:

## PR description
Correctly and consistently handle mining pre/post merge for PoW and Clique networks.  The majority of this PR is code from the merge branch which needed some 'treatment' in order to not cause regressions in behavior on main.

* for clarity, rename isStoppedAtTerminalDifficulty to hasReachedTerminalDifficulty 
* use AtomicReference for PostMergeContext rather than synchronized
* add mining parameter override for consensus mechanisms that mine regardless of miningParameters
* merge parameter validation from `merge` branch which allows mining parameters without PoW mining (if merge is enabled)
* merge fix for optional in TxPoolEvictionService
* fix for ancestorIsValidTerminalProofOfWork in MergeCoordinator when close to TTD
* address Rayonism tech-debt, use setBlockChoiceRule at TTD rather than disabling reorg behavior entirely pre-merge

## Fixed Issue(s)

fixes #2898 

## Changelog

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PostMergeContext.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PostMergeContext.java
@@ -33,7 +33,7 @@ import com.google.common.collect.EvictingQueue;
 public class PostMergeContext implements MergeContext {
   static final int MAX_BLOCKS_IN_PROGRESS = 12;
 
-  private static PostMergeContext singleton;
+  private static final AtomicReference<PostMergeContext> singleton = new AtomicReference<>();
 
   private final AtomicReference<SyncState> syncState;
   private final AtomicReference<Difficulty> terminalTotalDifficulty;
@@ -57,12 +57,11 @@ public class PostMergeContext implements MergeContext {
     this.syncState = new AtomicReference<>();
   }
 
-  public static synchronized PostMergeContext get() {
-    // TODO: get rid of singleton if possible: https://github.com/hyperledger/besu/issues/2898
-    if (singleton == null) {
-      singleton = new PostMergeContext();
+  public static PostMergeContext get() {
+    if (singleton.get() == null) {
+      singleton.set(new PostMergeContext());
     }
-    return singleton;
+    return singleton.get();
   }
 
   @Override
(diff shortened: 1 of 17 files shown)


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PostMergeContext.java
```diff
@@ -73,35 +71,38 @@ public static synchronized PostMergeContext get() {
...
+      throw new IllegalStateException("cannot set null terminal total difficulty");
+    }
     terminalTotalDifficulty.set(newTerminalTotalDifficulty);
     return this;
   }
 
   @Override
   public void setIsPostMerge(final Difficulty totalDifficulty) {
-    if (Optional.ofNullable(lastFinalized.get()).isPresent()) {
+    if (isPostMerge.get().orElse(Boolean.FALSE)
```
fab-10, line 83: my understanding is that if we have a finalized block, then we are post merge, so this check seems redundant, am I missing something?

fab-10 (approved):
LGTM, just a comment on a possible redundant check

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PostMergeContext.java
garyschulte: Yes the finalized check is unnecessary now.  The post-merge condition is 'stronger' - we will never switch back to the prior consensus mechanism and will instead rely on forkchoiceUpdated to direct any reorg around TTD.  

I will remove the unnecessary finalized check and update the comment 👍

PR #3499, opened 2022-02-28
Improve contract validation errors

shemnon:
## PR description

For each contract validation error return a relevant error instead of
the incorrect "INSUFFICIENT_GAS" error.

## Changelog


diega (approved):
LGTM

Issue #3500, opened 2022-02-28
Merge: Terminal block check fail if missing parent


garyschulte:
### Description
When checking if a block from the consensus client descends from the terminal block, besu can incorrectly report INVALID_TERMINAL_BLOCK if the block has missing ancestor blocks.  

For example, when a consensus client is restarted and and re-syncs the beacon chain, the resulting "gap" in blocks the besu has seen will cause the terminal block check to fail, and besu responds with INVALID_TERMINAL_BLOCK.  

### Acceptance Criteria
* if consensus client sends a block for which besu is missing ancestor blocks, besu should attempt to sync and respond with SYNCING

### Steps to Reproduce (Bug)
1. start besu and teku
2. let them sync
3. stop teku for a few minutes
4. restart teku, teku optimistic sync's

**Expected behavior:** [What you expect to happen]
besu initiates backward sync

**Actual behavior:** [What actually happens]
Besu return INVALID_TERMINAL_BLOCK

**Frequency:** [What percentage of the time does it occur?]
100%

PR #3501, opened 2022-02-28
Bugfix/terminal block check

garyschulte:
## PR description
Ensure besu correctly checks for latest ancestor before checking if a block descends from terminal.  

Also, forkchoiceUpdated calls for finalized blocks which besu does not have should start or append to a backward sync.

## Fixed Issue(s)

#3500

## Changelog


fab-10 (approved):
LGTM

PR #3503, opened 2022-03-01
Add GitHub action to verify gradle-wrapper.jar

ajsutton:
## PR description
Checks that the gradle wrapper jar is from an official release since a maliciously modified jar added as part of a gradle upgrade would be a significant security risk but is hard to verify from the code diff.  This checks the jar comes from an Official Gradle release.

See https://github.com/gradle/wrapper-validation-action for more details.

## Changelog


frankisawesome (approved):
Looks good, our other gh action yml has a SPDX header, should probably have one here as well?

ajsutton:
SPDX seems like a good idea. Added.

PR #3539, opened 2022-03-08
Trace APIs: add revert reason

macfarla:
Add revertReason to trace_transaction, trace_call, trace_callMany, trace_rawTransaction and trace_get

Note revertReason manually added to the relevant test files generated from OE.
Also added a new test file for trace_transaction.

## Changelog


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/AbstractJsonRpcHttpBySpecTest.java
```diff
@@ -281,4 +285,9 @@ private String filterStringTrace(final String expectedResult) {
     final Matcher m = GAS_MATCH_FOR_TRACE.matcher(expectedResult);
     return m.replaceAll("");
   }
+
+  private String filterRevertReason(final String expectedResult) {
```
shemnon, line 289: This does not properly test the revert reason.  Our code could now output random corrupted bytes for the revert reason and it would pass the tests.

My thinking is we add the revert reasons to the test vectors to ensure we have stable and expected output for the reverts.  Since OE is deprecated and being end-of-lifed at the merge we don't need to worry about re-generating test vectors against OE.

macfarla: fair point. (although the generation of revertReason itself is tested in other places in the code). I've modified 4 test files and added a comment to the JSON.

pinges (approved):
LGTM

PR #3543, opened 2022-03-09
Upgrade TLS version to 1.3 in BesuCommand.java

bhaskarvilles:
Upgrade TLS version to 1.3 in BesuCommand.java
Signed-off-by: Bhaskar Ram <bhaskar@parrtosec.in>


macfarla (commented):
looks good to me and DCO is happy
there is a .dccache file that got accidentally added, can you remove it :)

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -3041,7 +3041,7 @@ protected void enableGoQuorumCompatibilityMode() {
 
   public static List<String> getJDKEnabledProtocols() {
     try {
-      SSLContext context = SSLContext.getInstance("TLS");
+      SSLContext context = SSLContext.getInstance("TLS v1.3");
```
macfarla, line 3044: is "v1.3" equivalent to "1.3"?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -3030,7 +3030,7 @@ protected void enableGoQuorumCompatibilityMode() {
 
   public static List<String> getJDKEnabledCypherSuites() {
     try {
-      SSLContext context = SSLContext.getInstance("TLS");
+      SSLContext context = SSLContext.getInstance("TLS 1.3");
```
frankisawesome, line 3033: Looks like you'll need "TLSv1.3" here
https://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#SSLContext

frankisawesome (changes requested):

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
macfarla: is there a test that we can add to ensure it's a valid value?

bhaskarvilles:
> looks good to me and DCO is happy there is a .dccache file that got accidentally added, can you remove it :)

Done !

PR #3546, opened 2022-03-09
Richer return information for Precompiled Contracts

shemnon:
## PR description

Allow precomiled contracts to return richer information instead of
results/fail. System/precompile contracts can now revert, fail, refund
gas, etc. instead of just succeed or fail.

This also had the nice effect of removing some hard-coded names for privacy contracts in the EVM code, which decouples it from privacy and frees up privacy for easier evolution.

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
```diff
@@ -65,6 +66,10 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(PrivacyPrecompiledContract.class);
 
+  static final PrecompileContractResult NO_RESULT =
```
jflo, line 69: What if we defined this as a static on PrecompileContractResult? That would minimize duplication and further divorce from privacy stuff.

evm/src/main/java/org/hyperledger/besu/evm/precompile/PrecompiledContract.java
```diff
@@ -52,5 +54,90 @@
...
+  /**
+   * Executes the pre-compiled contract.
+   *
+   * @param input the input for the pre-compiled contract.
+   * @param messageFrame context for this message
+   * @return the output of the pre-compiled contract.
+   * @deprecated Migrate to use {@link #computePrecompile(Bytes, MessageFrame)}.
+   */
+  @Deprecated(since = "22.1.2")
+  default Bytes compute(final Bytes input, final @Nonnull MessageFrame messageFrame) {
```
jflo, line 77: If an implementing class fails to override compute, won't we get a call loop between these default implementations?

jflo (changes requested):
If there really is a call loop danger in the compute/computePrecompile default methods, I think that needs to be corrected.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
shemnon: This is unique to the privacy precompiles, as they set refund gas to true.

evm/src/main/java/org/hyperledger/besu/evm/precompile/PrecompiledContract.java
shemnon: This is the side effect of backwards compatability.

First, to support pre-existing precompiles we cannot require the new `computePrecompile` method be implemented, so it must have a default method.

Second, we don't want to require new precompiles to stub out a deprecated method useless code.  So for good ergonomics we need a default.

shemnon: both of these defaults call the other, but if we only provided a "real" implementation for one and an error producing one for the other then we still only discover at runtime that you didn't override methods.  Even if we detect the loop (which has performance impacts for well behaving subclasses) you still won't detect the failure except at runtime.  And by "at runtime" I mean in unit test code because everyone unit tests before they do a test network right?

shemnon: So the question is _how_ to correct it.  Both methods must provide defaults.  If we make computePrecomile the error throwing method then that breaks backwards compatibility.  If we make compute the error throwing method it breaks downstream users that still call compute directly.  If we presume that implementors are useful and implement at least one of the methods there is no loop.

So as the reviewer, how do you want me to break things?

jflo: > If we presume that implementors are useful and implement at least one of the methods there is no loop.

Cool, thats all i needed to know.

jflo (approved):

Issue #3528, opened 2022-03-07
Trace RPC methods - consistent approach to logging


macfarla:
Trace APIs - consistent approach to logging.

What level of logging - Is trace level the right level? - eg here https://github.com/hyperledger/besu/pull/3459#discussion_r820347827?
What are we logging? params? Just the name of the called method? Just the most important params?

PR #3553, opened 2022-03-10
added trace logging of parameter RPC calls

macfarla:

LOG.trace of all supplied RPC params

## PR description

Fixes 

## Changelog #3528 


garyschulte (commented):
Can we use `org.hyperledger.besu.util.Slf4jLambdaHelper.traceLambda()` here?  some of these parameters look non-trivial to evaluate

macfarla:
> Can we use `org.hyperledger.besu.util.Slf4jLambdaHelper.traceLambda()` here? some of these parameters look non-trivial to evaluate

I did consider it - but none of them are being evaluated for the log, they are all evaluated earlier and I just log the variable value. Apart from getName() which is trivial to evaluate.

macfarla:
> > Can we use `org.hyperledger.besu.util.Slf4jLambdaHelper.traceLambda()` here? some of these parameters look non-trivial to evaluate
> 
> I did consider it - but none of them are being evaluated for the log, they are all evaluated earlier and I just log the variable value. Apart from getName() which is trivial to evaluate.

added traceLambda where callParams are logged. all other values are already string or long.

garyschulte (approved):
🚢  ƛ

PR #3556, opened 2022-03-10
modified batch request test to avoid async complete issue

macfarla:

Send 2 x the same request as the batch. The request of "" seems to cause different behaviour.

## Changelog


frankisawesome (approved):
Lgtm

PR #3569, opened 2022-03-11
Warn if consensus client doesn't call transition configuration endpoint within 120 seconds

### LINKED ISSUE #3561: Log warnings when CL heartbeat lapses
### Description
When Besu is configured as an Execution Layer client on mainnet, an error/warning should be raised if engine_exchangeTransitionConfigurationV1 is not called at least once every 2 minutes

### Acceptance Criteria
* Logging output has warnings when expected CL is not calling.

### Steps to Reproduce (Bug)
1. Configure Besu as a merge EL client and run it.

After 2 minutes without receiving the call, warnings should be output in the log. Warnings should repeat every 2 minutes.

garyschulte:
## PR description
Adds a QoS timer behavior to `engine_exchangeTransitionConfigurationV1` rpc method 

[implements 120 second QoS warning](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#specification-3)

If the RPC method is not configured, there will be no such warning.  

Example log warning:

`2022-03-10 19:40:30.853-08:00 | vert.x-eventloop-thread-1 | WARN  | EngineExchangeTransitionConfiguration | not called in 120 seconds, consensus client may not be connected`

## Fixed Issue(s)

fixes #3561

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
```diff
@@ -37,10 +39,16 @@
 public class EngineExchangeTransitionConfiguration extends ExecutionEngineJsonRpcMethod {
   private static final Logger LOG =
       LoggerFactory.getLogger(EngineExchangeTransitionConfiguration.class);
+  private final Vertx timerVertx;
+  private static final AtomicLong qosTimerId = new AtomicLong();
+  private static final AtomicLong qosLastCall = new AtomicLong(System.currentTimeMillis());
+  static final long QOS_TIMEOUT = 120000L;
```
macfarla, line 45: should we specify units in the name

macfarla (approved):
LGTM. minor comment on the name

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
```diff
@@ -102,4 +113,29 @@ private JsonRpcResponse respondWith(
...
+    return z -> {
+      if (getLastCallMillis() + qosTimeout < System.currentTimeMillis()) {
+        logQosFailure(qosTimeout);
+      }
+      resetQosHandler(qosTimeout, qosHandler(qosTimeout));
+    };
+  }
+
+  void resetQosHandler(final long qosTimeout, final Handler<Long> qosHandler) {
+    Optional.ofNullable(qosTimerId.get()).ifPresent(timerVertx::cancelTimer);
```
daniellehrner, line 129: `AtomicLong.get()` returns `long`, so it never can be `null` and `cancelTimer` will be called every time. We could init `qosTimerId` with `null` and check directly for that. Not sure if there are other solutions that would be cleaner.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
```diff
@@ -102,4 +113,29 @@ private JsonRpcResponse respondWith(
...
+    qosLastCall.set(System.currentTimeMillis());
+    qosTimerId.set(timerVertx.setTimer(qosTimeout, qosHandler));
+  }
+
+  void logQosFailure(final long qosTimeout) {
+    LOG.warn("not called in {} seconds, consensus client may not be connected", qosTimeout / 1000L);
+  }
+
+  long getLastCallMillis() {
+    return Optional.ofNullable(qosLastCall.get()).orElse(0L);
```
daniellehrner, line 139: Here we have the same behavior as above, qosLastCall.get() cannot be null.

daniellehrner (changes requested):
I tested it with the following config, but did not see the warning in the log:

```
Xmerge-support=true
genesis-file="/Users/daniel/projects/mergemock/genesis.json"
network-id=65535
rpc-http-enabled=true
rpc-http-api=["ETH","NET","ENGINE"]
host-allowlist=["*"]
rpc-http-cors-origins=["*"]
logging="TRACE"
discovery-enabled=false
```

Is the config wrong or is there another problem?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
garyschulte: Good point - I started with AtomicReference, and changed it to AtomicLong.  I will init the AtomicLong with an explicit timerId like Long.MAX_VALUE and just cancel it

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
garyschulte: That config should work, I used the same mergemock config to test :).  Did you wait 2 minutes to see the warning?

```
2022-03-11 04:02:47.029-08:00 | main | INFO  | Runner | Starting Ethereum main loop ... 
..
2022-03-11 04:04:42.218-08:00 | EthScheduler-Timer-0 | INFO  | FullSyncTargetManager | No sync target, waiting for peers. Current peers: 0
2022-03-11 04:04:46.911-08:00 | vert.x-eventloop-thread-0 | WARN  | EngineExchangeTransitionConfiguration | not called in 120 seconds, consensus client may not be connected
...
2022-03-11 04:06:42.388-08:00 | EthScheduler-Timer-0 | INFO  | FullSyncTargetManager | No sync target, waiting for peers. Current peers: 0
2022-03-11 04:06:46.919-08:00 | vert.x-eventloop-thread-0 | WARN  | EngineExchangeTransitionConfiguration | not called in 120 seconds, consensus client may not be connected
...
2022-03-11 04:10:42.723-08:00 | EthScheduler-Timer-0 | INFO  | FullSyncTargetManager | No sync target, waiting for peers. Current peers: 0
2022-03-11 04:10:46.922-08:00 | vert.x-eventloop-thread-0 | WARN  | EngineExchangeTransitionConfiguration | not called in 120 seconds, consensus client may not be connected
```

jflo (approved):

PR #3571, opened 2022-03-11
[MINOR] xl executor for quorum ATs (nightly)

macfarla:

Revert back to the xlarge executor for Quorum ATs

Last time these tests were consistently passing, the config looked like this https://github.com/hyperledger/besu/commit/1c39d4814327b8b095a8801adcb2332a12e19bb6

## Changelog


.circleci/config.yml
```diff
@@ -266,7 +268,8 @@ jobs:
       - capture_test_logs
 
   acceptanceTestsQuorum:
-    executor: quorum_ats_executor_med
+    parallelism: 1
```
mark-terry, line 271: Does this have an effect? Or does it just explicitly turn parallelism off?

macfarla: tbh I'm not 100% sure but I'm reverting back to how it was when this CI was passing. Once it's passing, we can go from there.

jflo (approved):

PR #3578, opened 2022-03-13
Remove -x flag for bonsai CLI options

frankisawesome:

## PR description
Removes -X from the bonsai flags so users feel safer using the options as bonsai have been tested extensively, note forest will still be the default storage format.

## Changelog
Add to breaking changes (should we support the -X flag for a little longer?)


matkt (approved):
LGTM

shemnon (approved):
This is fine.  The `--X` part of the option means that there is no commitment to keep it working or the same between any two releases. Removing it means changing it in a backwards incompatible way or removing it only occurs at quarterly releases.

diega (commented):
Are these options still considered _unstable_? if not, I think we need to change the package and the `unstableDataStorageOptions` field of the `BesuCommand`

matkt:
> Are these options still considered _unstable_? if not, I think we need to change the package and the `unstableDataStorageOptions` field of the `BesuCommand`

no, it's not unstable. good catch we can change that

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/DataStorageOptions.java
```diff
@@ -29,10 +29,10 @@
 
 public class DataStorageOptions implements CLIOptions<DataStorageConfiguration> {
 
-  private static final String DATA_STORAGE_FORMAT = "--Xdata-storage-format";
+  private static final String DATA_STORAGE_FORMAT = "--data-storage-format";
```
macfarla, line 32: as per comments, can the package name here be changed to remove "unstable"

macfarla (changes requested):
just need to remove "unstable" from the package name for DataStorageOptions

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/DataStorageOptions.java
frankisawesome: [8bd3d10](https://github.com/hyperledger/besu/pull/3578/commits/8bd3d10e473dc7fc2984c924161b484ebadd1e48)

macfarla (approved):

PR #3607, opened 2022-03-18
Bugfix/stuck at ttd

### LINKED ISSUE #3609: merge - bugfix for stuck at TTD
### Description
When besu stops at TTD waiting for consensus client, it will in some cases get stuck at TTD and refuse to move.  If the CL sets the TTD block hash as finalized, besu fails to recognize that the TTD descends from a valid terminal block and returns INVALID_TERMINAL_BLOCK to the consensus client.

### Acceptance Criteria
* ensure besu correctly identifies TTD block as descending from TTD

### Steps to Reproduce (Bug)
1. sync besu and teku on kiln 
2. observe besu stuck at TTD

Besu should finalize the TTD block and move forward

Besu gets stuck at TTD

on kiln, with besu and teku combo, 100%

* Software version: 22.1.2+

garyschulte:
## PR description
* fix a corner case where besu does not recognize TTD block descending from TTD
* extend the distance that besu will 'look back' for TTD block in the case of a long period on non-finality post-merge
* added MergeCoordinator.addNewPayloadToSync to append newPayload to backward sync if we short circuit newPayloads early due to missing parent hash

## Fixed Issue(s)

fixes #3609

## Changelog

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
@@ -431,6 +431,11 @@ public class MergeCoordinator implements MergeMiningCoordinator {
   }
 
   @Override
+  public CompletableFuture<Void> appendNewPayloadToSync(final Block newPayload) {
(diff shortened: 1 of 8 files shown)


daniellehrner (approved):
LGTM

PR #3610, opened 2022-03-18
Tune transaction synchronization parameter to adapt to mainnet traffic

### LINKED ISSUE #3524: Sync issues and many "1000 expired transaction messages have been skipped" warning
### Description
It seems there is a correlation when Besu is not able to stay in sync on mainnet and the presence of many warnings in the log, stating that `"1000 expired transaction messages have been skipped.`

for example this graph, shows that Besu if frequently some blocks behind
image
and at the same time the log if full of warnings:
```
{"timestamp":"2022-03-04T09:14:34,355","level":"WARN","thread":"EthScheduler-Transactions-1","class":"TransactionsMessageProcessor","message":"1000 expired transaction messages have been skipped.","throwable":""}
{"timestamp":"2022-03-04T09:14:35,411","level":"WARN","thread":"EthScheduler-Transactions-1","class":"TransactionsMessageProcessor","message":"1000 expired transaction messages have been skipped.","throwable":""}
{"timestamp":"2022-03-04T09:14:35,688","level":"WARN","thread":"EthScheduler-Transactions-0","class":"PendingTransactionsMessageProcessor","message":"1000 expired transaction messages have been skipped.","throwable":""}
{"timestamp":"2022-03-04T09:14:36,161","level":"WARN","thread":"EthScheduler-Transactions-0","class":"TransactionsMessageProcessor","message":"1000 expired transaction messages have been skipped.","throwable":""}
{"timestamp":"2022-03-04T09:14:36,938","level":"WARN","thread":"EthScheduler-Transactions-0","class":"TransactionsMessageProcessor","message":"1000 expired transaction messages have been skipped.","throwable":""}
{"timestamp":"2022-03-04T09:14:37,638","level":"WARN","thread":"EthScheduler-Transactions-1","class":"TransactionsMessageProcessor","message":"1000 expired transaction messages have been skipped.","throwable":""}
```

at the same time the `transactions` executor, has a very long queue
image
and this could be a symptom, that Besu is queuing too much and is not able to process in time these messages.

If queuing too much is the issues, then a limit on the queue or other strategy to discard some messages should be implemented

* Software version: 22.1.0
* Java version: Java11
* Cloud VM, type, size: Amazon t3.xlarge with EBS gp3

fab-10:
## PR description
This PR is composed of the tuning of some internal parameters related to the transaction synchronization.

1. Currently Besu handles hundreds of transaction message per seconds, so having a queue of 1M messages result in a lot of expired messages, make sense to reduce the capacity to drop incoming messages that could not be handled anyway.
2. Currently on mainnet there are many thousand of pending transactions exchanged between peers, but Besu has a short memory of what has been exchanged with a specific peer, with the result that the same transaction is often exchanged back and forth with the same peer, expecially when the peer is another Besu (below a sample of one transaction exchanged many time between peers) so it is necessary to remember more transactions seen with peers.
3. Added detailed information about exchanged transactions at trace log level.

This is a sample extraction from the trace logs of one transactions, peers `0x0b7cbef740d1af50cf`, `0xec1e73d0839ec87abe`, `0x364c56424b81989e57` are other Besus
[tx-sample.pdf](https://github.com/hyperledger/besu/files/8303223/tx-sample.pdf)

Running this patch on a 4 Besu nodes for some hours, showns that CPU and memory usage is significantly reduced
image

## Fixed Issue(s)

fixes #3524 
## Changelog


garyschulte (approved):
🚢

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransactionsMessageProcessor.java
```diff
@@ -96,7 +98,14 @@ void processNewPooledTransactionHashesMessage(
...
+
+      traceLambda(
+          LOG,
+          "Received pooled transaction hashes message from {}, incoming hashes {}, incoming list {}",
+          peer::toString,
+          incomingTransactionHashes::size,
+          incomingTransactionHashes::toString);
 
       transactionTracker.markTransactionsHashesAsSeen(
           peer, transactionsMessage.pendingTransactions());
```
daniellehrner, line 111: Instead of `transactionsMessage.pendingTransactions()` we could use `incomingTransactionHashes` directly.

daniellehrner (approved):
I've left a comment for a very small change, but I am fine to merge it without the change as well.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransactionsMessageProcessor.java
fab-10: right, I will incorporate this in the next change that I am working on to improve eth/66 support

Issue #3540, opened 2022-03-08
Add support for Mac M1 processor/Multi-Platform builds


lunoon:
### Description
As a developer/user I want to use the Docker image to develop and test.  
However there is currently no support for the M1 processor yet and the current workaround is to use —platform=linux/amd64 flag in Docker.

### Acceptance Criteria
* Support of the docker image with ARM processors such as M1 with Multi-Platform builds.

### Versions (Add all that apply)
* Software version: [`besu --version`] all
* Java version: [`java -version`] all
* OS Name & Version: [`cat /etc/*release`] MacOS 12.2
* Kernel Version: [`uname -a`] Darwin 21.3.0
* Docker Version: [`docker version`] all

macfarla:
We are working on getting this working. https://github.com/hyperledger/besu/pull/3481

Issue #3220, opened 2021-12-31
Not supported rocksDBException on AWS


matkt:
### Description

{"timestamp":"2021-12-31T10:44:39,631","level":"INFO","thread":"main","class":"ProtocolScheduleBuilder","message":"Protocol schedule created with milestones: [Petersburg: 0, Istanbul: 1561651, Berlin: 4460644, London: 5062605]","throwable":""}
org.rocksdb.RocksDBException: NotSupported

To display full help:
besu [COMMAND] --help
2021-12-31 10:44:51,219 main INFO Registered Log4j as the java.util.logging.LogManager.
ANTLR Tool version 4.7.1 used for code generation does not match the current runtime version 4.9.2

**Frequency:** [What percentage of the time does it occur?]
100%

### Versions (Add all that apply)
* Software version: ed9d64d9cb8d1b4a82d5379ef513adaafae91806
* Java version: [`java -version`]
* OS Name & Version: 
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
Amazon Linux release 2 (Karoo)

* Kernel Version: [`uname -a`]
Linux ip-10-0-1-196.us-east-2.compute.internal 4.14.232-177.418.amzn2.x86_64 # 1 SMP Tue Jun 15 20:57:50 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

garyschulte:
@matkt is there any more detail on this to help reproduce it?  I bumped rocksdbjni back to 6.27.3 again and haven't been able to reproduce it on an ec2 instance.  What was the JVM?  was there an existing db?  using bonsai or forest?

matkt:
> @matkt is there any more detail on this to help reproduce it? I bumped rocksdbjni back to 6.27.3 again and haven't been able to reproduce it on an ec2 instance. What was the JVM? was there an existing db? using bonsai or forest?

I think it was not an empty DB. Maybe it can help to reproduce

garyschulte:
Was this bonsai?  I have been unable to reproduce this.  I have tried:

* starting from empty with a build using 6.27.3
* starting goerli fast-sync on 22.1.0-RC3, and restarting with a 6.27.3 build
[...]

PR #3634, opened 2022-03-25
update rocksdbjni to v6.29.4 for compatibility with arm64 CPU architecture

jpnovais:
Signed-off-by: Pedro Novais <jpvnovais@gmail.com>

## PR description
Updates rocksdbjni dependency to v6.29.4 for compatibility with arm64 CPU architecture. 
It was tested on Apple M1 Pro 2021, OSx 12.3 

## Related Issue(s)
- #3540 

## Documentation

## Changelog


garyschulte (changes requested):
Tested locally on an M1:
* `./gradlew check` passes ✅ (with one unrelated reflection related test failure)
* trying to reproduce #3220 via syncing/re-syncing/participating in testnets
  * kiln on m1 with native jdk ✅
  * goerli on m1 with a docker image ❌
  
I will add more detail once I have it, just wanted to get this in here before somebody merges this PR

gradle/versions.gradle
```diff
@@ -159,7 +159,7 @@ dependencyManagement {
     dependency 'org.openjdk.jmh:jmh-core:1.34'
     dependency 'org.openjdk.jmh:jmh-generator-annprocess:1.34'
 
-    dependency 'org.rocksdb:rocksdbjni:6.15.2'
+    dependency 'org.rocksdb:rocksdbjni:6.29.4'
```
garyschulte, line 162: At the start of all fast-sync's we clear at a minimum the worldstate.  The problem seems to be that when we call `RocksDbKeyValueStorage.clear(segmentHandle)`, we call 
```
deleteRange(final ColumnFamilyHandle columnFamilyHandle,
      final byte[] beginKey, final byte[] endKey)
```

which is a defaulted overload for:
```
 RocksDB.deleteRange(private native void deleteRange(final long handle, final byte[] beginKey,
      final int beginKeyOffset, final int beginKeyLength, final byte[] endKey,
      final int endKeyOffset, final int endKeyLength, final long cfHandle)
```

which is throwing `NotSupported`.  I am digging around in the native codebase to find out if it is indeed no longer supported and if so what the alternative is.  

It is unfortunate that the jni interface offers up a deprecated signature with no warning or annotation... 🤔

garyschulte (changes requested):
more detail on the 'NotSupported' exception.

gradle/versions.gradle
garyschulte: I am not sure why deleteRange is throwing NotSupported, but on a local branch I have made some changes to `RocksDBColumnarKeyValueStorage.clear()` to just drop and recreate the column family rather than deleting each and every key via iterator.  I am testing this locally and can push a suggested change to this PR.

Anecdotally, I think the drop/recreate strategy might resolve #3682 which is experiencing a long blocking behavior on this call.  🤞

garyschulte (changes requested):
I can't push to your fork, but I'd like to keep the changes in your PR

Would you pull this commit from my repo that fixes the `clear()` problem?  

https://github.com/garyschulte/besu/commit/38a9b09b9e99ba96d0cbce171deedbcbddd9845b

gradle/versions.gradle
garyschulte: this can be resolved by dropping the column family and recreating it instead of iterating over all of the keys in the segmenthandle.  This should resolve the blocking `clear()` operations on #3682 also.

matkt: I don't see how deleting the column will physically delete the data. and if we have to iterate on all the data of the segment, those can be very long or even block Besu. Have you had the opportunity to test this technique on a database filled

matkt: The last time I had tried this technique it had freeze all Besu because of the number of keys

matkt: Are we sure that after everything will be removed ?

garyschulte: anecdotally I see the account db storage getting reclaimed on a restart, hopefully I will catch a re-pivot and see it happening organically.  I will confirm/disconfirm tomorrow.

jpnovais:
@garyschulte I pulled your commit and merged main as well. Thanks for looking into this.

garyschulte:
> @garyschulte I pulled your commit and merged main as well. Thanks for looking into this.

It looks like we need to run spotless on that commit for some indentation diffs.

garyschulte (approved):

shemnon:
:tada:

PR #3636, opened 2022-03-28
delete slice input data code

### LINKED ISSUE #3635: Index out of Bound error was occured when sending concurrent private transaction.
### Description

When I used Flexible privacy groups, For sending concurrent private transaction, I distributed each of transactions using priv_distributeRawTransaction, and returned enclaveKey. After returned enclaveKey, I put enclave Key, in the data column of each raw transactions, finally I send transaction using eth_sendRawTransaction. But I found a problem. If enclaveKey vaule is null, an index size error occurs at the member node connected to tessera node. After this error, because of validator node isn't related to private transaction, validator node still staking up blocks. But the member node can't  longer stack up blocks.

```
Critical Exception Processing Transaction","throwable":" java.lang.IndexOutOfBoundsException: index (0) must be less than size (0)
    at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:1355)
    at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:1337)
    at org.apache.tuweni.bytes.ArrayWrappingBytes.slice(ArrayWrappingBytes.java:73)
    at org.hyperledger.besu.ethereum.mainnet.precompiles.privacy.FlexiblePrivacyPrecompiledContract.compute(FlexiblePrivacyPrecompiledContract.java:110)
    at org.hyperledger.besu.evm.processor.MessageCallProcessor.executePrecompile(MessageCallProcessor.java:153)
    at org.hyperledger.besu.evm.processor.MessageCallProcessor.start(MessageCallProcessor.java:65)
    ...
```

 I think that error was occured at hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/FlexiblePrivacyPrecompiledContract.java 110. 

```
final String key = input.slice(0, 32).toBase64String();
```

This code do not check whether the input value is not empty array. So, I changed this code as follows, I tried to connect the errors  to EnclaveClientException. It works correctly. Is there any problem with this code?

```
final String key = input.toBase64String();
```

### Acceptance Criteria
* In exception handling, Warning occurs that the enclave key is none, and moves on to the next block.

### Steps to Reproduce (Bug)
1. Distribute each of transactions using priv_distributeRawTransaction, and returned enclaveKey.
2. Put null(supposed to enclaveKey), in the data column of each raw transactions.
3. Send transaction using eth_sendRawTransaction

**Frequency:**
always

* Software version: 22.1.1, 22.1.2

yooseonghyuck:
## PR description
This code do not check whether the input value is not empty array. So, I changed this code as follows, I tried to connect the errors to EnclaveClientException.

## Fixed Issue(s)

fixes #3635

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/FlexiblePrivacyPrecompiledContract.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/FlexiblePrivacyPrecompiledContract.java
@@ -107,7 +107,7 @@ public class FlexiblePrivacyPrecompiledContract extends PrivacyPrecompiledContra
 
     final Hash pmtHash = messageFrame.getContextVariable(KEY_TRANSACTION_HASH);
 
-    final String key = input.slice(0, 32).toBase64String();
+    final String key = input.toBase64String();
 
     final ReceiveResponse receiveResponse;
     try {


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/FlexiblePrivacyPrecompiledContract.java
```diff
@@ -107,7 +107,7 @@ public PrecompileContractResult computePrecompile(
 
     final Hash pmtHash = messageFrame.getContextVariable(KEY_TRANSACTION_HASH);
 
-    final String key = input.slice(0, 32).toBase64String();
```
shemnon, line 110: what if the input is greater than 32 bytes?

yooseonghyuck: As a result of testing, An exception was occurs. I think, this exception is caused by that  the tessera node can’t find corresponding enclavekey.

```
“timestamp”:“2022-03-29T05:45:41,018"
“container”:“ip-172-31-57-54.ap-northeast-2.compute.internal”
"level”:“DEBUG”,“thread”:“EthScheduler-Workers-1"
“class”:“FlexiblePrivacyPrecompiledContract”
“message”:“Can not fetch private transaction payload with key kSUnNL+W53+lT7qc4POA380jiArDnj8i5ePSa89zLEurzQ==“
”throwable”:” org.hyperledger.besu.enclave.EnclaveClientException: Message with hash= was not found 
  at org.hyperledger.besu.enclave.Enclave.handleJsonResponse(Enclave.java:160)
  at org.hyperledger.besu.enclave.Enclave.lambda$receive$2(Enclave.java:85)
  ...
```

macfarla: It doesn't make sense for the enclaveKey to be null - you need to figure out why that's happening.

yooseonghyuck: the enclaveKey might be able to null by mistake. An asynchronous problem can be exemplified.

macfarla:
This isn't the right approach to fix the problem. Closing this PR

PR #3655, opened 2022-03-30
Fast sync with merge

### LINKED ISSUE #3506: Post-merge Fast sync behavior
### Description
Post-merge, fast sync should behave differently with regard to selecting a pivot block

When the chain genesis has a terminal total difficulty configured, fast sync should ​wait/listen for a finalized block from a forkchoiceUpdated call by the consensus layer.  If/when a new pivot block needs to be selected, only use a finalized block from the CL

### Acceptance Criteria
* at the start of fast sync, determine whether the chain is pre or post merge by inspecting the total difficulty of the chain from the fast-sync peers
* ensure the fast-sync peers all have the same total difficulty (once chain reaches TTD, the difficulty should be static)
* ensure the pivot block is finalized (ideally it is the most recent finalized block)

### LINKED ISSUE #3507: Post-merge Snap-sync behavior
### Description
This is largely the same as #3506, but for snap-sync
Post-merge, snap sync should behave differently with regard to selecting a pivot block

When the chain genesis has a terminal total difficulty configured, snap sync should ​wait/listen for a finalized block from a forkchoiceUpdated call by the consensus layer. If/when a new pivot block needs to be selected, only use a finalized block from the CL

### Acceptance Criteria
* at the start of snap sync, determine whether the chain is pre or post merge by inspecting the total difficulty of the chain from the snap-sync peers
* ensure the snap-sync peers all have the same total difficulty (once chain reaches TTD, the difficulty should be static)
* ensure the pivot block is finalized (ideally it is the most recent finalized block)

### LINKED ISSUE #3724: Re-pivoting a fast sync across TTD fails
### Description
- This issue was discovered on the lighthouse/besu pair testing mainnet-shadowfork-1
    - Started a fastsync before ttd first pivot was at 14563240 (pre-merge)
    - During the fast sync, Besu re-pivoted to a new block three times (14563804, 14565117, 14567386) each being past TTD.  It incorrectly failed to apply postmerge rules to 14564208, the first block after TTD.
    - World state however was unaffected, and the world state is now ahead of the blockchain import. Any backwards syncs trigger from engine_forkChoiceUpdated would proceed back to the TTD, and then immediately re-try to import 14564208. Now the correct import rules would be applied, the block would be structurally valid, however we don't have the worldstate the block requires, and so the block cannot be executed.

Import rules being applied during fast sync need to be merge aware just like everywhere else. Should probably revisit full sync and snap-sync code as well to make sure the validation logic is re-used across all of them.

fab-10:
## PR description

This PR add the information of an initial sync phase to SyncState. Initial sync is everything that is not Full sync, we need that because when using Fast/Snap sync, we need to complete this phase before Besu can report that it is in sync, when processing CL messages, and before the Backward sync can start.

If TTD is configured, the Fast sync goes in a transition modality, that is, if the best peer has reached TTD, then Fast sync listens for a finalized block, and when it is available choses it as pivot, otherwise if best peer is still before TTD, then the previous PoW strategy is used, selecting the pivot block from peers.

Then `EngineForkchoidUpdated` API has been extended to accept listeners and inform them when a new message, that includes the last finalized block, as arrived, in this way Fast sync can use the finalized block as pivot.

Other improvements avoid to start Full sync and block propagation manager, in case at the start the network is already after TTD, because both will be stop just after the start anyway. 

## Fixed Issue(s)

fixes #3506 
fixes #3507
fixes #3724 

## Documentation

## Changelog

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
@@ -435,7 +435,7 @@ public abstract class BesuControllerBuilder implements MiningParameterOverrides 
 
     if (genesisConfigOptions.getTerminalTotalDifficulty().isPresent()) {
       LOG.info(
-          "TTD difficulty is present, creating initial sync phase with transition to Pos support");
+          "TTD difficulty is present, creating initial sync phase with transition to PoS support");
 
       final MergeContext mergeContext = protocolContext.getConsensusContext(MergeContext.class);
       final FinalizedBlockHashSupplier finalizedBlockHashSupplier =
@@ -451,6 +451,7 @@ public abstract class BesuControllerBuilder implements MiningParameterOverrides 
 
       return new TransitionPivotSelector(
           genesisConfigOptions,
+          finalizedBlockHashSupplier,
           pivotSelectorFromPeers,
           new PivotSelectorFromFinalizedBlock(
               genesisConfigOptions,
--- a/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
@@ -86,6 +86,7 @@ import java.nio.file.Path;
 import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -336,7 +337,8 @@ public abstract class BesuControllerBuilder implements MiningParameterOverrides 
             syncConfig.getComputationParallelism(),
             metricsSystem);
     final EthContext ethContext = new EthContext(ethPeers, ethMessages, snapMessages, scheduler);
-    final boolean fastSyncEnabled = SyncMode.FAST.equals(syncConfig.getSyncMode());
+    final boolean fastSyncEnabled =
+        EnumSet.of(SyncMode.FAST, SyncMode.X_SNAP).contains(syncConfig.getSyncMode());
     final SyncState syncState = new SyncState(blockchain, ethPeers, fastSyncEnabled);
 
     final TransactionPool transactionPool =
--- a/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
@@ -453,7 +453,9 @@ public abstract class BesuControllerBuilder implements MiningParameterOverrides 
           genesisConfigOptions,
           pivotSelectorFromPeers,
           new PivotSelectorFromFinalizedBlock(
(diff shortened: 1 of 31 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/TransitionPivotSelector.java
```diff
@@ -0,0 +1,80 @@
...
+
+    if (bestPeerEstDifficulty.lessThan(totalTerminalDifficulty)) {
+      LOG.info(
+          "Chain has not yet reached TTD, best peer has estimated difficulty {},"
+              + " select pivot from peers",
+          bestPeerEstDifficulty);
+      return pivotSelectorFromPeers.selectNewPivotBlock(peer);
+    }
+
+    LOG.info(
```
gezero, line 68: I have a feeling that this condition should be first. Even if all our peers are before TTD but consensus layer told us already about a finalized block, we should use it.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -418,6 +428,38 @@ public BesuController build() {
...
   }
 
+  private PivotBlockSelector createPivotSelector(final ProtocolContext protocolContext) {
+
+    final PivotSelectorFromPeers pivotSelectorFromPeers = new PivotSelectorFromPeers(syncConfig);
+    final GenesisConfigOptions genesisConfigOptions = genesisConfig.getConfigOptions();
+
+    if (genesisConfigOptions.getTerminalTotalDifficulty().isPresent()) {
+      LOG.info(
+          "TTD difficulty is present, creating initial sync phase with transition to Pos support");
```
gezero, line 438: Minor: Pos ->PoS

gezero (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/TransitionPivotSelector.java
fab-10: make sense, we should follow what CL says, will change it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -273,6 +273,7 @@ public Updater putAccountInfoState(final Hash accountHash, final Bytes accountVa
     @Override
     public WorldStateStorage.Updater saveWorldState(
         final Bytes blockHash, final Bytes32 nodeHash, final Bytes node) {
+      trieBranchStorageTransaction.put(WORLD_ROOT_HASH_KEY, nodeHash.toArrayUnsafe());
```
matkt, line 276: duplicate. we can remove it

matkt (approved):
Looking at the snapsync part. seems to be ok for me

PR #3664, opened 2022-03-31
FlexiblePrivacyPrecompile - Handle invalid Payload ID

macfarla:
Handle the case where payloadId passed in is null or < 32 bytes. 

See #3635 

Added 2 new tests for null and < 32bytes

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/FlexiblePrivacyPrecompiledContract.java
```diff
@@ -104,6 +104,10 @@ public PrecompileContractResult computePrecompile(
     if (skipContractExecution(messageFrame)) {
       return NO_RESULT;
     }
+    if (input == null || input.size() < 32) {
+      LOG.debug("Can not fetch private transaction payload with key {}", input);
```
pinges, line 108: Currently we do only have two valid sizes, 32 and 64 bytes, so I guess we could do:
`if (input == null || (input.size() != 32 && input.size() != 64))`
And I think we should log that there was an invalid input and probably even log this as an error?

pinges (commented):
please check out my comment

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/FlexiblePrivacyPrecompiledContract.java
macfarla: done

pinges (approved):
LGTM

PR #3684, opened 2022-04-04
Update storage for snapsync and fastsync

matkt:

## PR description

Modification of the behavior of bonsai storage  in order to avoid deleting everything between two pivot blocks.
The flat database will be deleted between two pivot blocks. And the trie will be used when data is not available in the flat database. During the next writing of this data it will be put automatically in the flat for a faster reading.

This will cause a reduction of the size of the database. Because we will not have data never called in the flat database.

But this may reduce the reading speed when accessing data for the first time. 

For now it will be only activated for snapsync. To check if it's working well  

## Documentation

## Changelog


garyschulte:
relates to #3694

PR #3695, opened 2022-04-05
Add IPC transport for JSON-RPC APIs

### LINKED ISSUE #535: Support IPC transport for JSON-RPC APIs
So that system administrators can expose JSON-RPC APIs exclusively to localhost without needing firewalls, add a IPC transport for the JSON-RPC APIs.

This would make things better for Vipnode: [https://github.com/vipnode/vipnode/issues/37#issuecomment-505197281]

Acceptance Criteria
 * IPC can be enabled or disabled independently of JSON-RPC

 * Supported API groups (admin, eth, net etc) can be specified separately for IPC and HTTP endpoints

 * IPC implementation is compatible with Web3J IPC implementation

 * IPC implementation is compatible with geth attach

diega:
## PR description
Adds socket interface for JSON-RPC, only supported on BSD (OSX) and Linux. This feature has been developed with the intention to reuse and unify, as much as possible, the behaviour among the other transport layers (i.e. HTTP and WS)

Given this is a big PR to review, I have split it in the following commits (I recommend checking them individually):
- [HEAD~7](https://github.com/diega/besu/commit/jsonrpc_ipc~7): Fixes two tests for the HTTP implementation (more info in the commit message)
- [HEAD~6](https://github.com/diega/besu/commit/jsonrpc_ipc~6): Extracts HTTP authentication to its own handler
- [HEAD~5](https://github.com/diega/besu/commit/jsonrpc_ipc~5): Moves authentication handler first in the pipeline, to fail fast **(*)**
- [HEAD~4](https://github.com/diega/besu/commit/jsonrpc_ipc~4): Replaces _timeout handler_ in GraphQL service to use the one provided by Vert.x **(*)**
- [HEAD~3](https://github.com/diega/besu/commit/jsonrpc_ipc~3): Extracts HTTP json parse to its own handler
- [HEAD~2](https://github.com/diega/besu/commit/jsonrpc_ipc~2): Extracts json-rpc execution logic outside of the HTTP handler
- [HEAD~1](https://github.com/diega/besu/commit/jsonrpc_ipc~1): Refactor WS service to use the same execution logic as the HTTP service **(*)**
- [HEAD](https://github.com/diega/besu/commit/jsonrpc_ipc): Implements IPC transport layer

(*): these commits can be omitted without affecting the implementation of the IPC transport

## Note
This implementation has been tested against `geth attach` and it works fine.

## Fixed Issue(s)
fixes #535

## Documentation

## Changelog


atoulme (approved):
LGTM

shemnon (approved):
I like the idea of it, but it's been so long since I've been in that code I'm not sure what is an improvement and what isn't.  The acceptance tests still pass.

A set of acceptance tests that use IPC to do some testing wold be nice, but doesn't have to be in this PR and can be in a follow on. It's also tricky because it may have OS specific setup.

Generally speaking a quarterly release is the right time to put these changes in, as we've signaled that's when breaking features come in.  While re-writing core parts of JSON aren't breaking, they are higher risk.

Unless @macfarla, @usmansaleem , @mark-terry, or someone else whose had their hands deep in the RPC code chime in I'm cool with @diega pressing the merge button before RC2 gets spun out.

macfarla:
@mark-terry can you review?

gradle/versions.gradle
```diff
@@ -172,9 +174,7 @@ dependencyManagement {
 
     dependency 'org.web3j:abi:4.8.9'
     dependency 'org.web3j:besu:4.8.9'
-    dependency('org.web3j:core:4.8.9') {
-      exclude group: 'com.github.jnr', name: 'jnr-unixsocket'
-    }
+    dependency 'org.web3j:core:4.8.9'
```
macfarla, line 177: we did exclude this to avoid a CVE

diega: Oh, ok. I thought it was because of licensing issues. I can revert that but I'll have to remove the web3j test along with it.

diega: It seems to be [fixed](https://github.com/web3j/web3j/issues/1616) in `org.web3j:core:4.9.1`. I can try submitting another PR upgrading the version and then rebase this one on top of the new `main`, wdyt?

macfarla: sounds good to me - tag @antonydenyer for review

antonydenyer: Given that we only use web3j in tests and we're actually using ipc now the risk is very low. That said I'd attempt an upgrade as part of this pr, looking through the changes in web3j they look minimal.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/ipc/JsonRpcIpcServiceTest.java
```diff
@@ -0,0 +1,242 @@
...
+        .describedAs("Test completed on time")
+        .isTrue();
+    if (testContext.failed()) {
+      throw testContext.causeOfFailure();
+    }
+  }
+
+  @Test
+  void worksWithWeb3j() {
+    final Path socketPath = tempDir.resolve("besu-test.ipc");
```
antonydenyer, line 81: I think this test should be done in the acceptance tests with everything wired without mocks. It means that we isolate the dependency on web3j.

antonydenyer (changes requested):
Really excited to see this implemented. Great stuff.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/ipc/JsonRpcIpcServiceTest.java
diega: Sure. I just did that. I removed this test and I added an acceptance one just checking a proper `net_version` response. I also removed the `org.web3j:core` _testDependency_ that I had added to the `:ethereum:api` module so the usage of this dependency is still isolated in the _acceptance tests_ scope.

gradle/versions.gradle
diega: @antonydenyer I gave a try to upgrade web3j to `4.9.0` (not `4.9.1` b/c there is not such version for the Gradle plugin) in #3752. Beside of some failing tests, this ends up in some classpath cluttering, as far as I can see in IntelliJ, and I would like to take time to investigate.
By removing the unit test and adding an analogous one into the acceptance test scope, the usage of this library is again isolated. So, if this is acceptable, I would suggest to not block this PR b/c of the re-inclusion of the `jnr-unixsocket` dependency and continue with the review.

antonydenyer: Agreed - it's a bit of a mess!

antonydenyer (approved):

macfarla:
@diega this needs a changelog entry :)

mark-terry (approved):
LGTM.

diega:
Sure @matt-nelson-csi, thank you for the heads up

PR #3699, opened 2022-04-06
Enforce trailing peer limits when select pivot block fails

frankisawesome:

## PR description
Rather than waiting 5 seconds and hoping everything goes well the next run we should run the trailing peer limiter task and drop the useless peers.
## Fixed Issue(s)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -69,6 +72,13 @@ public FastSyncActions(
     this.ethContext = ethContext;
     this.syncState = syncState;
     this.metricsSystem = metricsSystem;
+    this.trailingPeerLimiter =
+        new TrailingPeerLimiter(
+            ethContext.getEthPeers(),
+            () ->
+                new TrailingPeerRequirements(
+                    syncState.getLocalChainHeight() + syncConfig.getFastSyncPivotDistance(),
```
matkt, line 80: are we sure about that ? If I'm close to the head (<getFastSyncPivotDistance)

frankisawesome: I changed it to use the min of local + pivot distance or the current best estimate, will do some testing today

frankisawesome:
I've set up two test instances on goerli and the sync finished in 4h8m and 4h15m respectively, quite happy with this.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -177,10 +179,26 @@ private CompletableFuture<FastSyncState> retrySelectPivotBlockAfterDelay() {
     return ethContext
         .getScheduler()
         .scheduleFutureTask(
+            this::limitTrailingPeersAndRetrySelectPivotBlock, Duration.ofSeconds(5));
+  }
+
+  private long conservatitvelyEstimatedPivotBlock() {
```
macfarla, line 185: s/conservatitvely/conservatively

macfarla (approved):
LGTM. added a comment re a typo

PR #3703, opened 2022-04-06
Add flag for snapsync

matkt:

## PR description

We start adding the flags for the snapsync. Most are experimental flags because normally the user doesn't have to change the values. But it can be used to change the values if needed or to tune performance.

## Documentation

## Changelog

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/options/unstable/SynchronizerOptions.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/options/unstable/SynchronizerOptions.java
@@ -17,6 +17,8 @@ package org.hyperledger.besu.cli.options.unstable;
 import org.hyperledger.besu.cli.options.CLIOptions;
 import org.hyperledger.besu.cli.options.OptionParser;
 import org.hyperledger.besu.ethereum.eth.sync.SynchronizerConfiguration;
+import org.hyperledger.besu.ethereum.eth.sync.snapsync.ImmutableSnapSyncConfiguration;
+import org.hyperledger.besu.ethereum.eth.sync.snapsync.SnapSyncConfiguration;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -99,6 +99,34 @@ public DefaultSynchronizer(
...
+              syncConfig,
+              dataDirectory,
+              protocolSchedule,
+              protocolContext,
+              metricsSystem,
+              ethContext,
+              worldStateStorage,
+              syncState,
+              clock);
+    }*/
```
garyschulte, line 128: can we omit this from the PR?

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/SynchronizerOptions.java
```diff
@@ -212,6 +225,53 @@ public void parseBlockPropagationRange(final String arg) {
   private int worldStateTaskCacheSize =
       SynchronizerConfiguration.DEFAULT_WORLD_STATE_TASK_CACHE_SIZE;
 
+  @CommandLine.Option(
+      names = SNAP_PIVOT_BLOCK_WINDOW_VALIDITY_FLAG,
+      hidden = true,
+      defaultValue = "126",
```
garyschulte, line 231: not 128?

garyschulte: ah ok, just saw:
```
// we use 126 and not the max value (128) to avoid sending requests that will be refused
```

garyschulte (changes requested):
LGTM, minor ask that we just omit the DefaultSynchronizer comment block from the PR.  I think that is the only change in that file

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DefaultSynchronizer.java
matkt: I prefer to omit this one and activate this part of the code when all the PRs will be merged. In two PRs maybe not more

garyschulte (approved):

PR #3707, opened 2022-04-07
Feature/bonsai fast sync clear flat db

### LINKED ISSUE #3694: Resume worldstate download on fast-sync re-pivot when using bonsai
### Description
When besu is using bonsai storage and doing a fast-sync, re-pivoting to a new block causes besu to re-start the worldstate download.  

This means a fast-sync bonsai node is in a race to download its entire worldstate before its peers are no longer able to serve it.  

### Acceptance Criteria
* bonsai fast-sync nodes should not restart worldstate download on re-pivot

### Steps to Reproduce (Bug)
1. fast-sync mainnet with besu on bonsai
2. observe the worldstate downlod re-start upon fast-sync re-pivoting

* bonsai fast-sync nodes should continue to download worldstate on a re-pivot
* at the end of fast-sync the worldstate trie should implement a 'heal' process similar to snap-sync

* besu re-starts the woldstate download on re-pivot

predictably with slower hardware, and slower upstream internet connections

* Software version: [`besu --version`] 22.1.x

garyschulte:
## PR description
During fast sync on bonsai, clear only the flat database rather than the entire bonsai worldstate.  

## Fixed Issue(s)

Along with #3684 , fixes #3694 

## Documentation

## Changelog

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/AccountTrieNodeDataRequest.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/AccountTrieNodeDataRequest.java
@@ -43,7 +43,12 @@ class AccountTrieNodeDataRequest extends TrieNodeDataRequest {
 
   @Override
   public Optional<Bytes> getExistingData(final WorldStateStorage worldStateStorage) {
-    return worldStateStorage.getAccountTrieNodeData(getLocation().orElse(Bytes.EMPTY), getHash());
+    return getLocation()
+        .flatMap(
+            location ->
+                worldStateStorage
+                    .getAccountStateTrieNode(location, getHash())
+                    .filter(data -> Hash.hash(data).equals(getHash())));
   }
 
   @Override
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/StorageTrieNodeDataRequest.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/StorageTrieNodeDataRequest.java
@@ -49,9 +49,12 @@ class StorageTrieNodeDataRequest extends TrieNodeDataRequest {
     return getAccountHash()
         .flatMap(
             accountHash ->
-                worldStateStorage
-                    .getAccountStateTrieNode(accountHash, getHash())
-                    .filter(data -> Hash.hash(data).equals(getHash())));
+                getLocation()
+                    .flatMap(
+                        location ->
+                            worldStateStorage
+                                .getAccountStorageTrieNode(accountHash, location, getHash())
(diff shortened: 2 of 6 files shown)


matkt (approved):
LGTM if the test is green

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/StorageTrieNodeDataRequest.java
```diff
@@ -46,8 +46,12 @@ protected void doPersist(final Updater updater) {
 
   @Override
   public Optional<Bytes> getExistingData(final WorldStateStorage worldStateStorage) {
-    return worldStateStorage.getAccountStorageTrieNode(
-        getAccountHash().orElse(Hash.EMPTY), getLocation().orElse(Hash.EMPTY), getHash());
+    return getAccountHash()
+        .flatMap(
+            accountHash ->
+                worldStateStorage
+                    .getAccountStateTrieNode(accountHash, getHash())
```
matkt, line 53: an issue in this call getAccountStateTrieNode instead of getAccountStorageTrieNode

matkt (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/StorageTrieNodeDataRequest.java
garyschulte: Thanks for catching this copypasta!

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/AccountTrieNodeDataRequest.java
```diff
@@ -43,7 +43,12 @@ protected void doPersist(final Updater updater) {
 
   @Override
   public Optional<Bytes> getExistingData(final WorldStateStorage worldStateStorage) {
-    return worldStateStorage.getAccountTrieNodeData(getLocation().orElse(Bytes.EMPTY), getHash());
+    return getLocation()
+        .flatMap(
+            location ->
+                worldStateStorage
+                    .getAccountStateTrieNode(location, getHash())
+                    .filter(data -> Hash.hash(data).equals(getHash())));
```
fab-10, line 51: if I understood well, when the data at that location does not match the hash, that location could be deleted?

fab-10 (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/AccountTrieNodeDataRequest.java
garyschulte: it could, and we could implement lazy writes also.  For now though we simply rely on subsequent writes to clean up a dirty value.

Issue #3736, opened 2022-04-19
Investigate new peer selection strategy


macfarla:
Try different ways to select "best" peer to download from for fast sync. 

Idea: Measuring which peer responded fastest with the pivot block header.

PR #3741, opened 2022-04-20
Use pivot header download time to select sync target

frankisawesome:

## PR description
We previously select the peer with the best chain when selecting a sync target - this makes little sense and this PR uses the pivot block header download time as the comparator instead.

## Fixed Issue(s)
#3736 

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -193,6 +193,13 @@ public Optional<EthPeer> bestPeerWithHeightEstimate() {
         p -> p.isFullyValidated() && p.chainState().hasEstimatedHeight());
   }
 
+  public Optional<EthPeer> bestPeerWithHeightEstimateForFastSync() {
+    return streamAvailablePeers()
+        .filter(p -> p.isFullyValidated() && p.chainState().hasEstimatedHeight())
+        .sorted(BEST_CHAIN)
+        .max(Comparator.comparing(EthPeer::getLastPivotHeaderDownloadTime));
```
macfarla, line 200: wouldn't we want min here?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -193,6 +193,13 @@ public Optional<EthPeer> bestPeerWithHeightEstimate() {
         p -> p.isFullyValidated() && p.chainState().hasEstimatedHeight());
   }
 
+  public Optional<EthPeer> bestPeerWithHeightEstimateForFastSync() {
+    return streamAvailablePeers()
+        .filter(p -> p.isFullyValidated() && p.chainState().hasEstimatedHeight())
+        .sorted(BEST_CHAIN)
```
macfarla, line 199: if we're taking max from downloadTime, we don't need to sort by best chain?

frankisawesome: Some peers may not have a download time recorded yet and the BEST_CHAIN is a fall back for when not enough peers have it

macfarla: what will the max() return in that case... tests would help!

frankisawesome: yep agreed, definitely needs to be min but I'll write tests with different scenarios.

macfarla (approved):
LGTM

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/RetryingGetHeaderFromPeerByNumberTask.java
```diff
@@ -71,6 +71,8 @@ protected CompletableFuture<List<BlockHeader>> executePeerTask(
               if (!peerResult.getResult().isEmpty()) {
                 result.complete(peerResult.getResult());
               }
+              assignedPeer.ifPresent(
+                  ethPeer -> ethPeer.setLastPivotHeaderDownloadTime(getTaskTimeInSec()));
```
ajsutton, line 75: Does second resolution for the timing actually give us a meaningful measure? It might eliminate some very slow peers but I would have thought this would wind up giving most peers the same value.

frankisawesome: I think in the case that there's a small difference between the fastest peer and the rest it'll improve the overall sync time quite a bit since that one peer will take us all the way in most cases in fast sync

ajsutton: I may be misunderstanding the code, but my point is when there's a small difference using the measurement in seconds rather than milliseconds will cause that difference to be lost.

ajsutton: Also, do we have some timings to show this helps? It probably wouldn't hurt but we should be requesting block data from a wide range of peers and requests for the headers we get from the sync target used to be done in parallel to that so small to moderate timing differences would be unlikely to make a difference. The code could have changed a lot since I last looked at it though.

ajsutton: btw, I'm not intending to block this merging, just adding some drive by observations. :)

frankisawesome: Aha I gotcha, that's a great catch actually. For some additional context, I spun some nodes for testing running this pr vs a control group on goerli and I observed pretty much the same sync performance, using millisecond here can probably help..

frankisawesome: The timer is in second but double so we still get all the decimals
Maybe I'm completely missing it here but I thought the process is:
- Request pivot header from a number of peers
- Confirm the header with peers
- Download chain data from one "best peer" till the pivot
- Download world state from one "best peer" till the pivot
Is that not the case here @ajsutton

ajsutton: Nope, that's not how it works, or if it is it's badly broken. :)

* Request block at the pivot block number from multiple peers and ensure there's agreement on it. The resolution when peers disagree has changed around a bit - looks like it keeps trying to find one that all the peers it queries agree on.
* Start an async pipeline (see `FastSyncDownloadPipelineFactory`):
  * Download checkpoint headers from our best peer - these are spaced out and let us check that the headers we get from other peers matches up with our target chain.
  * Fill in the header gaps between those checkpoints from multiple peers
  * Get the bodies for those headers from multiple peers
  * Get the receipts for those headers from multiple peers
  * Import the completed blocks

Importantly, all of those steps are async and trying to keep their output pipeline full, so we're downloading the next checkpoint headers while trying to get the headers/bodies/receipts for previous ones which generally means even if there's a delay from some requests the pipeline has enough buffer to keep things busy the whole time.

This change would potentially improve the speed we can download checkpoint headers but I don't think that's the bottleneck now because it's such a small number of the requests we make.  You probably want to look at the metrics the pipeline publish to track which steps are the bottleneck.

frankisawesome:
@macfarla I'm gonna close this one as it seems like my assumptions are not correct per adrian's comments above :)

macfarla:
yeah we may have been overthinking an edge case :) all good!

PR #3747, opened 2022-04-21
Treat the nonce as unsigned long

### LINKED ISSUE #3702: Nonce must be 64-bit
### Description
The nonce in the [Ethereum yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf) is defined as a 64 bit value (see page 5 bottom). In the current implementation we use `long` for the nonce. The maximum positive value of `long` is `2^63 − 1` which means it is less than the specified `2^64`. We need to use `BigInteger` instead of `long` to fix this.

This issue was discovered by the latest version of the reference tests. The valid nonce in this test causes an overflow in our current implementation: https://github.com/ethereum/tests/blob/develop/src/TransactionTestsFiller/ttNonce/TransactionWithHighNonce64Minus2Filler.json

daniellehrner:

## PR description

We have to use a 64 bit value for nonce in order to be compliant with the Ethereum specifiaction. Currently we use `long` which is 64 bit, but normally only 63 bit are used for the value and one bit is used to indicate if the number is positive or negative. Which means that our nonce currently is only 63 bit long.

There are alternative functions that allow to treat `long` as a 64 bit unsigned integer, by simply interpreting all of the 64 bits as value. This PR intends to do that in order to avoid creating instances of classes like `UInt64` whenever we operate with the nonce.

## Fixed Issue(s)
Example: fixes #3702 

## Documentation

## Changelog


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetProofTest.java
```diff
@@ -176,7 +182,7 @@ private GetProofResult generateWorldState() {
     final Wei balance = Wei.of(1);
     final Hash codeHash =
         Hash.fromHexString("0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
-    final long nonce = 1;
+    final long nonce = Long.parseUnsignedLong("18446744073709551614");
```
macfarla, line 185: is it worth spelling it out with a const name eg MAX_UNSIGNED_LONG_VALUE

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionEncoder.java
```diff
@@ -38,8 +38,8 @@ interface Encoder {
     void encode(Transaction transaction, RLPOutput output);
   }
 
-  private static final ImmutableMap<TransactionType, Encoder> TYPED_TRANSACTION_ENCODERS =
-      ImmutableMap.of(
+  private static final Map<TransactionType, Encoder> TYPED_TRANSACTION_ENCODERS =
+      Map.of(
```
macfarla, line 42: why are we getting rid of ImmutableMap?

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPInput.java
```diff
@@ -145,6 +145,17 @@ public interface RLPInput {
    */
   long readLongScalar();
 
+  /**
+   * Reads a non-negative scalar from the input and return is as a long value which is interpreted
```
macfarla, line 149: s/return is/return it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PrivateTransaction.java
```diff
@@ -565,7 +565,7 @@ private static Bytes32 computeSenderRecoveryHash(
         RLP.encode(
             out -> {
               out.startList();
-              out.writeLongScalar(nonce);
+              out.writeUnsignedLongScalar(nonce);
```
macfarla, line 568: Also L465 in this file
    out.writeLongScalar(t.getNonce());

macfarla (changes requested):
I think there are a couple instances of writeLongScalar(nonce) that have been missed
eg ethereum/privacy/PrivateTransaction L465
AccountRangeMessage L124

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetProofTest.java
daniellehrner: Thank you! I've changed them now and checked again, I think now all occurences for the nonce have been changed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/privacy/PrivateTransaction.java
daniellehrner: Thank you! I've changed them now and checked again, I think now all occurences for the nonce have been changed

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPInput.java
daniellehrner: Thank you! I've changed it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionEncoder.java
daniellehrner: It was a suggestion from SonarCloud. I've reverted it back, because I am not sure why it is suggesting it.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetProofTest.java
daniellehrner: I've change it now to use constants in the tests

PR #3752, opened 2022-04-22
Upgrade Web3J

diega:
~~**ON HOLD**: waiting for `org.web3j:quorum` release (ref: [web3j-quorum#66](https://github.com/web3j/web3j-quorum/issues/66))~~

Update Web3J dependency

This PR also reverts #3789

## Documentation

## Changelog


diega:
@antonydenyer to move this PR to `4.9.3` we are missing `org.web3j:quorum` which is on `4.9.0`. Also the `web3j-gradle-plugin` would be great if can be moved to `4.9.3` (it's on `4.9.2`). Should I fill an issue in the respective projects?
This would be an attempt to check if we can make the _acceptanceTests_ to pass using those versions

antonydenyer:
> @antonydenyer to move this PR to `4.9.3` we are missing `org.web3j:quorum` which is on `4.9.0`. Also the `web3j-gradle-plugin` would be great if can be moved to `4.9.3` (it's on `4.9.2`). Should I fill an issue in the respective projects? This would be an attempt to check if we can make the _acceptanceTests_ to pass using those versions

I think raising an issue in the respective projects would make sense.

I'm happy to have a look at the `acceptanceTests` first, before looking at the version inconsistencies. 

I think we should be able to use everything on to `4.9.3` and have `org.web3j:quorum` and  `web3j-gradle-plugin` forced to use the `4.9.3` dependencies bit life would be a lot easier if upstream libraries were updated.

antonydenyer:
Blocked by https://github.com/web3j/web3j/issues/1747

diega:
While we wait for a release of the web3j components to fix the _acceptanceTests_, I tried to force the dependencies as suggested.
For `org.web3j:quorum` it seems to possible by using [_constraints_](https://docs.gradle.org/current/userguide/dependency_constraints.html), but `web3j-gradle-plugin` forces the inclusion of the `org.web3:core:$projectVersion` dependency programmatically  (see [`o.w.g.p.Web3jPlugin:49`](https://github.com/web3j/web3j-gradle-plugin/blob/v4.9.2/src/main/java/org/web3j/gradle/plugin/Web3jPlugin.java#L49)) so we end up using both versions during the build/test cycle. Fortunately this is not so terrible as both cycles have different classpaths, at build `v4.9.0` will be used just for the plugin, and at runtime `+v4.9.0` will be used for any other project usage.

antonydenyer:
@diega Had to force the dependency in the end! I think this is ready for review now.

acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/BlockUtils.java
```diff
@@ -50,7 +52,7 @@ public static BlockHeader createBlockHeader(
         Bytes.fromHexString(block.getExtraData()),
         null,
         mixHash,
-        block.getNonce().longValue(),
+        new BigInteger(block.getNonceRaw().substring(2), 16).longValue(),
```
antonydenyer, line 55: https://github.com/web3j/web3j/issues/1752

diega:
@antonydenyer nice trick! that way you'll be overriding the dependency added programmatically by the plugin, right?

antonydenyer:
> @antonydenyer nice trick! that way you'll be overriding the dependency added programmatically by the plugin, right?

More like a sledgehammer - it will force everything to use '4.9.4', this could be a problem when you come to upgrade again. You could change the version in `versions.gradle` but it would have no affect. Perhaps we should use @freemanzMrojo approach as it has a little more finesse and may be easier for someone to spot. What do you think?

acceptance-tests/tests/build.gradle
```diff
@@ -72,8 +73,12 @@ dependencies {
   testImplementation 'org.web3j:abi'
   testImplementation 'org.web3j:besu'
   testImplementation 'org.web3j:core'
-  constraints {
-    implementation('jnr-posix:3.1.15') { because '3.0.47 from web3j has issues' }
+
+  implementation('org.web3j:core') {
+    version {
+      strictly('4.9.4')
+      because('web3j plugin 4.9.4 hasn\'t been released yet')
```
shemnon, line 80:
```suggestion
      because('Plugin version is 4.9.2 (latest), but we want it to use web3j libs version 4.9.4 ')
```

container-tests/tests/build.gradle
```diff
@@ -23,7 +23,11 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter'
   testImplementation 'org.testcontainers:testcontainers'
   testImplementation 'org.web3j:core'
-  testImplementation 'org.web3j:quorum'
+  testImplementation('org.web3j:quorum') {
+    constraints {
+      implementation('org.web3j:core:4.9.4') { because 'quorum 4.9.4 hasn\'t been released yet' }
```
shemnon, line 28:
```suggestion
      implementation('org.web3j:core:4.9.4') { 
        because 'Web3J/Quorum only has a 4.9.2 lib, but we want it to use web3j libs version 4.9.4 ')
      }
```

antonydenyer (approved):

freemanzMrojo:
Hi @antonydenyer @shemnon , hope you are doing well, and happy Friday! I was wondering if there is anything else to do in this PR, otherwise maybe you could merge it? I need it for https://github.com/hyperledger/besu/pull/4086, thanks in advance!

shemnon:
I'm holding off for 22.7.0 to ship as this is not essential to the merge.

PR #3759, opened 2022-04-25
Add errorprone to prevent Log4j direct usage

diega:
## PR description
This pull request contains some leftovers from #3285 and, most importantly, adds an _error prone_ check for preventing Log4j usage.
Most of this commits were part of #3425 but they mostly abstract stuff from Log4j and are not Logback specific so I thought it would make sense to introduce them here as well.

Here is the full list of changes:
- [HEAD~6](https://github.com/diega/besu/commit/logging_tweaks~6): replace Log4j2 logger with Slf4j added since #3285
- [HEAD~5](https://github.com/diega/besu/commit/logging_tweaks~5): renames leftover method from #3285
- [HEAD~4](https://github.com/diega/besu/commit/logging_tweaks~4): removes Log4j2 usage dependencies relying on the `Log4j2ConfiguratorUtil` (this is for simplifying the last diff)
- [HEAD~3](https://github.com/diega/besu/commit/logging_tweaks~3): replace Log4j2 `ThreadContext` usage with Slf4j's `MDC`
- [HEAD~2](https://github.com/diega/besu/commit/logging_tweaks~2): this is to explicitly check allowed levels instead of depending on Log4j enum options
- [HEAD~1](https://github.com/diega/besu/commit/logging_tweaks~1): for the same reason than before, the level values are filtered by the received string 
- [HEAD](https://github.com/diega/besu/commit/logging_tweaks): adds error prone check for preventing `o.a.l.log4j.LogManager` usages

## Documentation

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AdminChangeLogLevel.java
```diff
@@ -25,6 +25,7 @@
 
 import java.util.Arrays;
 import java.util.Optional;
+import java.util.Set;
 
 import org.apache.logging.log4j.Level;
```
shemnon, line 30: Can this use of Log4J be removed?  How needed is it?

errorprone-checks/src/main/java/org/hyperledger/errorpronechecks/BannedMethod.java
```diff
@@ -44,7 +44,9 @@ public class BannedMethod extends BugChecker implements MethodInvocationTreeMatc
           staticMethod().onClass("com.google.common.base.Objects").withAnyName(),
           "Do not use com.google.common.base.Objects methods, use java.util.Objects methods instead.",
           staticMethod().onClass("org.junit.Assert"),
-          "Do not use junit assertions. Use assertj assertions instead.");
+          "Do not use junit assertions. Use assertj assertions instead.",
+          staticMethod().onClass("org.apache.logging.log4j.LogManager"),
+          "Do not use org.apache.logging.log4j.LogManager, use org.slf4j.LoggerFactory methods instead.");
```
shemnon, line 49: Should this include the other log4j methods that were removed in this PR? static methods on `org.apache.logging.log4j.core.config.Configurator`, `org.apache.logging.log4j.ThreadContext`, instance methods on import `org.apache.logging.log4j.core.LoggerContext`, `org.apache.logging.log4j.core.config.Configuration`,
`org.apache.logging.log4j.core.config.LoggerConfig`?  Or is there a way to go full Ender-Wiggin on it and have it fail for any class from the log4j packages?

diega: No, we cannot rip off the whole package because we still have hard dependencies on it. The most visible one is `o.h.b.util.Log4j2ConfiguratorUtil` but we also depend on `o.a.l.log4j.Level` for the CLI. I went on banning just `o.a.l.log4j.LogManager` b/c it's the most common entry point to the Log4j framework from the day to day developer perspective

jflo (approved):

PR #3766, opened 2022-04-27
block import optimization

matkt:

## PR description

This PR makes significant changes during the block import (fast/snap sync) in order to speed up the synchronization time of Besu

This is the list of modification :
- A significant number of useless rlp `encode/decode` takes place during synchronization. Here is an example of useless encode/decode
    - Decode RLP from peer to have a Transaction object
    - After encode this transaction to rlp in order to calculate the hash of the transaction
    - Encode again this transaction to rlp in order to calculate the transactionRoot
    - Encode again this transaction to rlp in order to save the trx into the DB
    This is the same for all objects like `BlockHeader`, `BlockBody`, `Transaction`, `transactionReceipt`. The idea is to keep the rlp when decoding to use it if we need rlp again. 
- We do the generation of the `Receiptroot`, `transactionRoot` and `ommerHash` for the same block several times.
We do this at the time of the `GetBodiesFromPeerTask` and the `GetReceiptsFromPeerTask` but we also do it at the time of the `body validation`. So we calculate these values twice there.
The problem is that these are very consuming steps because we have to recreate a trie based on transactions and receipts. The idea is to store it in an optional field header and so if we already have it we don't need to calculate it again
- In the `GetBodiesFromPeerTask` and the `GetReceiptsFromPeerTask` methods we do the validations of the `Receiptroot` and the `transactionRoot` in parallel instead of doing it block by block. this saves time and like that we have all the root for the import part already calculated
- We also force the generation of the transaction Hash in parallel in the `GetBodiesFromPeerTask`. We need this hash during the import step in order to save the location of the each transaction hash in a block. It's better to calculate the hash in parallel during this step
- We validate the `block header` twice. During this step `CheckpointHeaderValidationStep` but also when calling the `fastImportBlock` method. Every time we want to validate a block we check a random to know if we want a full validation or not (1 chance out of 100 to make a full). The problem is even more important in POW because we have more chance for the same block to do a full check becaure we are calling two times this method `headerValidationPolicy.getValidationModeForNextBlock()`. So even more chance to do a full validation of a block on the mainnet
- Too many slow observers that impact the blocki import
  - TransactionPool (useless during block  import)
  - AbstractMiningCoordinator  (useless during block  import)
  - SyncState (useless during block  import)
  - FilterManager (useless during block  import)
  - LogSubscriptionService (useless during block  import)
  
  I keep :
   - NewBlockHeadersSubscriptionService
   - TransactionLogBloomCacher

  After we will have a fullsync with all observers

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/ListReceipts.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/ListReceipts.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright contributors to Hyperledger Besu
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.core;
+
+import org.hyperledger.besu.evm.log.LogsBloomFilter;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import org.jetbrains.annotations.NotNull;
+
+public class ListReceipts extends ArrayList<TransactionReceipt> {
+
+  private final Optional<Bytes> rlp;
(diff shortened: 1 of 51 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/ListReceipts.java
```diff
@@ -0,0 +1,45 @@
...
+package org.hyperledger.besu.ethereum.core;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.jetbrains.annotations.NotNull;
+
+public class ListReceipts extends ArrayList<TransactionReceipt> {
```
shemnon, line 24: Please use a "tuple" object (one field is RLP, one field is the List<TransactionReceipts>) instead of extending a standard collections object.  This ties us deeply to one collection implementation.

matkt: I added new fields to this objects. using the list avoid to change a lot of code but I will check if I can update all the part and creating a new object instead of extending the list

PR #3768, opened 2022-04-27
Set mainnet default gasLimit to 30M. Fixed #3767

### LINKED ISSUE #3767: Set default gasLimit to 30M for mainnet
### Description
Mainnet miners have voted to keep the gasLimit around 30,000,000 since [August 2021](https://etherscan.io/chart/gaslimit). Geth had an 8,000,000 gasLimit default on a [merge testnet which resulted in a drop in the gasLimit](https://github.com/ethereum/go-ethereum/pull/24680#issue-1200385203).

>Post-merge lots of validators will run with default configs, which means the gas limit will move to the default gas limit.
On mainnet-shadow-fork-1 this resulted in a drop from a gaslimit of 30M down to 8M.

Geth has since [updated their default to 30M](https://github.com/ethereum/go-ethereum/commit/c40943a167b603382bbd0f45c92c9b9c2669a080). 

If I an understanding the Besu default mainnet setting correctly the [default gasLimit is still 5000](https://github.com/hyperledger/besu/blob/main/config/src/main/resources/mainnet.json#L40) (The original limit at genesis 😲). I think this should be incresed to 30M in prepeartion for The Merge since there is a chance that validators may not understand/care about miner setting.

jmcpheron:
Update default gasLimit in preparation for The Merge™.


jmcpheron:
This would close Issue #3767

shemnon (changes requested):
This doesn't actually change the default gas for mainnet.  It changes a block value in the genesis block of Ethereum Mainnet, which changes the hash of the genesis block, which breaks validation from the very first block.

This value must be 5000 (`0x1388` hex), as shown here - https://etherscan.io/block/0

matkt:
> This doesn't actually change the default gas for mainnet. It changes a block value in the genesis block of Ethereum Mainnet, which changes the hash of the genesis block, which breaks validation from the very first block.
> 
> This value must be 5000 (`0x1388` hex), as shown here - https://etherscan.io/block/0

idk what do you think @shemnon but for me the gaslimit will increase automatically thanks to this  https://github.com/hyperledger/besu/blob/ed1329cf848a9d7858d039a55add810da3c428b5/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/LondonTargetingGasLimitCalculator.java

jmcpheron:
Thanks for the feedback. The documentation on the [target-gas-limit CLI](https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#target-gas-limit) option states:

>If a value for `target-gas-limit` is not specified, the block gas limit remains at the value specified in the [genesis file](https://besu.hyperledger.org/en/stable/Reference/Config-Items/#genesis-block-parameters).

Does that update in the [LondonTargetingGasLimitCalculator](https://github.com/hyperledger/besu/blob/ed1329cf848a9d7858d039a55add810da3c428b5/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/LondonTargetingGasLimitCalculator.java) change to default gasLimit to whatever the mainnet is voting for?

macfarla:
@garyschulte can we close this one since the gas limit is hardcoded?

Issue #3775, opened 2022-04-29
Missing assertions in TxPoolBesuPendingTransactionsTest


frankisawesome:
Seems to me the first couple of tests are missing any meaning assertions on the result we're getting

PR #3776, opened 2022-04-29
Remove Leading Zeros from TransactionPendingResult

jclapis:
## PR description

Currently, there is an issue with the way `TransactionPendingResult` gets serialized by in the JSON-RPC subsystem. Specifically, when responding to `eth_getTransactionByHash,` the `gasPrice`, `maxFeePerGas`, and `maxPriorityFeePerGas` values should be Quantities without leading zeros like so:

```
    "gas" : "0x11e63",
    "gasPrice" : "0x77359407",
    "maxPriorityFeePerGas" : "0x77359400",
    "maxFeePerGas" : "0xd09dc3000",
```

This is the official behavior defined in the spec and the behavior demonstrated by Geth and Nethermind.

However, in Besu, they are treated like 32-byte arrays and serialized with the leading zeros in place:

```
    "gas" : "0x10c83",
    "gasPrice" : "0x0000000000000000000000000000000000000000000000000000000ab5d04c00",
    "maxFeePerGas" : "0x0000000000000000000000000000000000000000000000000000000ab5d04c00",
```

This is a simple oversight caused by using `Wei.toHexString()` instead of `Wei.toShortHexString()`, which this PR addresses.

@frankisawesome has offered to fix the relevant unit tests that were unable to catch this, and will attach the fixes to this PR when they are ready.

### DIFF
--- a/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TxPoolBesuPendingTransactionsTest.java
+++ b/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TxPoolBesuPendingTransactionsTest.java
@@ -29,8 +29,10 @@ import org.hyperledger.besu.ethereum.eth.transactions.sorter.AbstractPendingTran
 import org.hyperledger.besu.ethereum.eth.transactions.sorter.GasPricePendingTransactionsSorter;
 
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


frankisawesome:
- Force pushed because the original change was 2 lines and fixing DCO seems harder than force pushing
- I think the test file here is missing lots of useful assertions, created a ticket https://github.com/hyperledger/besu/issues/3775

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TxPoolBesuPendingTransactionsTest.java
```diff
@@ -74,6 +75,31 @@ public void shouldReturnPendingTransactions() {
     assertThat(result.size()).isEqualTo(4);
   }
 
+  @Test
+  public void pendingTranasctionGasPricesDoNotHaveLeadingZeroes() {
```
macfarla, line 79: s/pendingTranasctionGasPricesDoNotHaveLeadingZeroes/pendingTransactionGasPricesDoNotHaveLeadingZeroes

frankisawesome: [eb4a045](https://github.com/hyperledger/besu/pull/3776/commits/eb4a045c18e07576082459b7db3de9e473e6a202)

siladu (approved):
LGTM

PR #3788, opened 2022-05-04
[MINOR] fixed one more error from CodeQL

macfarla:
* dereferenced variable is always null in log message

## Documentation

## Changelog


macfarla:
OK CodeQL says: A class that implements Cloneable should override Object.clone. For non-trivial objects, the Cloneable contract requires a deep copy of the object's internal state. A class that does not have a clone method indicates that the class is breaking the contract and will have undesired behavior.
But Sonar says: "clone" should not be overridden

lucassaldanha (approved):

PR #3794, opened 2022-05-06
Fix for node returns5 nodes in4 node network

pinges:
This change needs to be tested against a test net and maybe mainnet before it can be merged into main.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -85,25 +88,48 @@ public EthPeers(
...
-            peerValidators,
-            clock,
-            permissioningProviders);
-    connections.putIfAbsent(peerConnection, peer);
+    final Bytes peerId = peerConnection.getPeer().getId();
+    connections.compute(
+        peerId,
+        (id, existingPeer) -> {
+          if (existingPeer != null) {
+            existingPeer.setConnection(peerConnection); // just replace the connection
```
macfarla, line 96: this seems so simple

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -351,13 +353,20 @@ private CompletableFuture<PeerConnection> initiateOutboundConnection(final Peer
...
       return;
     }
+    final Peer peer = peerConnection.getPeer();
+    // Disconnect if not permitted
+    if (!peerPermissions.allowNewInboundConnectionFrom(peer)) {
+      LOG.debug(
+          "Node is not permitted to connect. Disconnect incoming connection: {}", peerConnection);
+      peerConnection.disconnect(DisconnectReason.UNKNOWN);
+      return;
+    }
```
macfarla, line 369: nice. it makes sense to do this earlier

PR #3812, opened 2022-05-10
cli: Do not warn when using --fast-sync-min-peers with snap sync

### LINKED ISSUE #3801: snapsync/cmdline: --fast-sync-min-peers also used by snapsync
## Usecase
I want to sync my Besu node against a single local (Geth) node, and not connect to the wider internet. To that end, I have been using the "fast-sync-min-peers" option to reduce the default of 5 to 1. If I do not do this, Besu will just wait indefinitely until it can connect to 5 nodes.

For completeness sake, my command line is `./besu --data-storage-format=bonsai --p2p-port=30303 --sync-mode=X_SNAP --fast-sync-min-peers=1 --discovery-enabled=false` 

## Issue
When using the command line above, I get the following warning about the `fast-sync-min-peers` option being unused (which is false, it _is_ used). 

```
2022-05-08 14:53:01.520+02:00 | main | WARN  | Besu | --fast-sync-min-peers has been ignored because --sync-mode was not defined on the command line.
```

Small patch to resolve it (with the "EnumSet" logic copied from elsewhere in the code)

```diff
diff --git a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
index cc699228f..b5d17072f 100644
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -203,6 +203,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collection;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -1858,7 +1859,7 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
         logger,
         commandLine,
         "--sync-mode",
-        !SyncMode.FAST.equals(syncMode),
+        !EnumSet.of(SyncMode.FAST, SyncMode.X_SNAP).contains(syncMode),
         singletonList("--fast-sync-min-peers"));

     if (!securityModuleName.equals(DEFAULT_SECURITY_MODULE)
```

## Side note / somewhat related issues
I would prefer if the message would instead state that sync mode "FAST" or "SNAP" have not been set. For example, if I were to set `--sync-mode FULL`, I would get the same error message about `--sync-mode` not having been passed. It was, it just was not set to the right value. 

Aside from that, the naming of `--fast-sync-min-peers` might be a bit off now, as it's also used by snap sync. But with snap sync also being a "fast" sync mode, I guess it's fine :)

pietjepuk2:
## PR description
The --fast-sync-min-peers option and its value are also relevant for
snap sync. We should therefore not log a warning about the min peers
option being ignored if it is passed.

## Fixed Issue(s)

Fixes #3801

## Documentation

## Changelog


macfarla:
Thanks for your PR @pietjepuk2 - the logic makes sense to me - but your commit is missing signoff. You should be able to follow the instructions here to fix it -> https://github.com/hyperledger/besu/pull/3812/checks?check_run_id=6366283404

matkt:
LGTM when the DCO will be ok

jackvreeken:
My bad, never had to deal with DCO before :)

matkt (approved):

PR #3828, opened 2022-05-12
Return Invalid forkchoice state if fcU contains invalid finalizedBlockHash or safeBlockHash

### LINKED ISSUE #3827: Missing check in engine_forkchoiceUpdatedV1 if finalizedBlockHash and safeBlockHash are valid
### Description
The engine spec says:
> Client software MUST return -38002: Invalid forkchoice state error if the payload referenced by forkchoiceState.headBlockHash is VALID and a payload referenced by either forkchoiceState.finalizedBlockHash or forkchoiceState.safeBlockHash does not belong to the chain defined by forkchoiceState.headBlockHash.

These checks are currently missing and need to be added.

daniellehrner:

When engine_forkchoiceUpdatedV1 is called we need to check that the blocks of finalizedBlockHash and safeBlockHash belong to the chain defined by headBlockHash

## Fixed Issue(s)
fixes #3827 

## Documentation

## Changelog


fab-10 (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
```diff
@@ -60,101 +62,108 @@ public String getName() {
...
+        new EngineUpdateForkchoiceResult(
+            VALID,
+            result.getNewHead().map(BlockHeader::getHash).orElse(null),
+            payloadId.orElse(null),
+            Optional.empty()));
+  }
+
+  private boolean isPartOfCanonicalChain(final Hash blockHash) {
+    final Optional<BlockHeader> maybeBlockHeader =
+        protocolContext.getBlockchain().getBlockHeader(blockHash);
```
garyschulte, line 160: the presence of a header in the blockchain isn't really an indicator it is part of the canonical chain.   It could be a fork for example.   If we want to be sure it is part of the canonical chain we should use `isDescendantOf(finalized, safeBlockHash)`

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
```diff
@@ -60,101 +62,108 @@ public String getName() {
...
+              Optional.of(newHead.get() + " did not descend from terminal block")));
     }
 
-    LOG.info(
-        "Consensus fork-choice-update: head: {}, finalized: {}",
-        forkChoice.getHeadBlockHash(),
-        forkChoice.getFinalizedBlockHash());
+    ForkchoiceResult result =
+        mergeCoordinator.updateForkChoice(
+            forkChoice.getHeadBlockHash(), forkChoice.getFinalizedBlockHash());
```
garyschulte, line 118: we should add safeBlockHash to the mergeContext while we are here.  We will need it for json-rpc eth_getBlockByNumber

garyschulte (changes requested):
👍 on the flow changes.  Couple comments about our treatment of safeBlockHash

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
garyschulte: The spec is actually transitive about this.  It says that safeBlockHash should be an ancestor of headBlockHash.  But really it needs to be somehwere between the inclusive range of [finalized .. head]

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
```diff
@@ -60,101 +62,108 @@ public String getName() {
...
+        forkChoice.getHeadBlockHash(),
+        forkChoice.getFinalizedBlockHash(),
+        forkChoice.getSafeBlockHash());
+
     Optional<Hash> maybeFinalizedHash =
         Optional.ofNullable(forkChoice.getFinalizedBlockHash())
             .filter(finalized -> !Hash.ZERO.equals(finalized));
 
+    if (!isPartOfCanonicalChain(forkChoice.getSafeBlockHash())
+        || (maybeFinalizedHash.isPresent() && !isPartOfCanonicalChain(maybeFinalizedHash.get()))) {
```
garyschulte, line 98: one additional consideration - I asked a clarifying question about safeBlockHash here: 
https://discord.com/channels/595666850260713488/892088344438255616/975915304045649960

I think the spec leaves room for us to get 0x000..000 as safeBlockHash in the same way do for finalized, prior to the first finalized block.

garyschulte (changes requested):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
garyschulte: it looks like we should expect 0x00...00 as safeBlockHash prior to TTD being justified/finalized:

https://discord.com/channels/595666850260713488/892088344438255616/976024691766591508

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
daniellehrner: We save it now in mergeContext and in the blockchain storage

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeContext.java
```diff
@@ -59,7 +63,7 @@ long addNewForkchoiceMessageListener(
 
   Optional<Block> retrieveBlockById(final PayloadIdentifier payloadId);
 
-  void fireNewForkchoiceMessageEvent(
+  void fireNewUnverifiedForkchoiceMessageEvent(
```
daniellehrner, line 66: This has been renamed, because we fire this message even in the case that we deem the forkchoice state as invalid. It is needed in order to make pivot selection, for the initial sync after the merge.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
```diff
@@ -68,93 +71,138 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+        requestId,
+        new EngineUpdateForkchoiceResult(
+            VALID,
+            result.getNewHead().map(BlockHeader::getHash).orElse(null),
+            payloadId.orElse(null),
+            Optional.empty()));
+  }
+
+  private boolean isValidForkchoiceState(
+      final Hash safeBlockHash, final Hash finalizedBlockHash, final BlockHeader newBlock) {
```
garyschulte, line 159: I'd rather see this logic should be in MergeCoordinator.updateForkChoice, where a failure returns a specific ForkchoiceResult that can indicate INVALID_FORK_CHOICE_STATE.  Otherwise the only place we are doing these checks in in a json rpc method, and the public updateForkChoice method is unsafe.

garyschulte (changes requested):
LGTM, just feedback on validation/safety.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
garyschulte: As discussed, we can address this in a subsequent PR to keep the size/scope of this PR more reviewable

garyschulte (approved):
🚢

PR #3835, opened 2022-05-14
Ref tests 10.3

shemnon:
## PR description
Upgrade reference Tests to 10.3

10.3 revision of reference tests has notable changes

* Nonce can be up to 2^64-1, with some opcode and validity interactions 
  specced in eip-2681
* Wei fields can be up to 2^256, tests check for rollover
* VM Tests were removed
* Legacy Tests were removed

Includes changes from PR #3747 
 
## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -150,19 +152,29 @@ public ValidationResult<TransactionInvalidReason> validate(
       }
     }
 
-    // transactionValidationParams.isAllowExceedingBalance() is used on eth_call
-    if (!feeMarket.satisfiesFloorTxCost(transaction)
-        && !transactionValidationParams.isAllowExceedingBalance()) {
+    if (transaction.getNonce() == -1) {
```
macfarla, line 155: how is -1 NONCE_TOO_HIGH

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetProofTest.java
```diff
@@ -176,7 +182,7 @@ private GetProofResult generateWorldState() {
     final Wei balance = Wei.of(1);
     final Hash codeHash =
         Hash.fromHexString("0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
-    final long nonce = 1;
+    final long nonce = Long.parseUnsignedLong("18446744073709551614");
```
macfarla, line 185: this value is used a few times - is it worth having a constant

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/GeneralStateReferenceTestTools.java
```diff
@@ -114,6 +114,16 @@ public static void executeTest(final GeneralStateTestCaseEipSpec spec) {
     final WorldState initialWorldState = spec.getInitialWorldState();
     final Transaction transaction = spec.getTransaction();
 
+    // Sometimes the tests ask us assemble an invalid transaction.  If we have
+    // no valid transaction then there is no test.  GeneralBlockChain tests
+    // will handle the case where we receive the TXs in a serilized form.
```
macfarla, line 119: s/serilized/serialized

macfarla (commented):
couple of comments. would be good for someone else familiar with ref tests to review also.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
shemnon: the max unsigned value is the same as the signed 2's complement -1.  However, I replaced it with MAX_NONCE to make it clear what and why we are testing.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -639,7 +649,12 @@ public Wei getUpfrontGasCost(final Wei gasPrice) {
     if (gasPrice == null || gasPrice.isZero()) {
       return Wei.ZERO;
     }
-    return Wei.of(getGasLimit()).multiply(gasPrice);
+    var cost = BigInteger.valueOf(getGasLimit()).multiply(gasPrice.getAsBigInteger());
+    if (cost.bitLength() > 256) {
```
matkt, line 653: how is it possible ?

shemnon: Multiplication of gas price by gas limit.  By using BigInteger I avoid the 256-bit rollover the `Wei` type enforces.  Detecting fixed-bit rollover for multiplication has too many corner cases, better to just do it in unbounded bit length and check for size.  

The `HighGasPrice` reference tests explicitly trip this condition and require the transaction to be marked as invalid.

PR #3850, opened 2022-05-17
Migrate referenceTests to Junit5

diega:
## PR description
More Junit5 migrations.
This PR moves `referenceTests` classes into its own sourceSet so they can be migrated independently from the `unit` tests. As a consequence, explicit `.class` inclusions/exclusions from test tasks have been removed

## Documentation

## Changelog


ethereum/referencetests/build.gradle
```diff
@@ -13,16 +13,93 @@
...
+    file("src/reference-test/templates/GeneralStateReferenceTest.java.template"),
+    "regressions",
+    "$buildDir/generated/sources/reference-test/$name/java/org/hyperledger/besu/ethereum/vm/generalstate",
+    "GeneralStateRegressionReferenceTest"
+    )
+}
+
 sourceSets {
-  test {
+  referenceTest {
```
shemnon, line 84: What advantage does a new reference test source set provide?  It will bring us further from "default" maven and gradle project setups, so I feel there needs to be a clear advantage to taking on this technical debt.

.gitmodules
```diff
@@ -1,5 +1,5 @@
 [submodule "eth-ref-tests"]
-	path = ethereum/referencetests/src/test/resources
+	path = ethereum/referencetests/src/reference-test/external-resources
```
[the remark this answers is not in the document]
diega: There are to directory changes here.
`resources` -> `external-resources`: I can move it back to be _resources_ and set the not official reference tests to some other directory like _extra-resources_ if it's more suitable.
`referencetests/src/test` -> `referencetests/src/reference-test`: I could have kept _test_, but the `referencetests/src/test` directory is tied to the _test_ task and I would not be able to exclude its execution from the regular `./gradlew test` call.

ethereum/referencetests/build.gradle
diega: I don't actually see this as technical debt here. It's an usual practice within Gradle to set different sourcesets for tests of different nature (e.g. _integrationTests_). But this also relates to what I replied to your other comment, it's to isolate the classpath from the _test_ task execution

shemnon: The reason I view it as technical debt is because source sets are a fairly poorly understood and fluid aspect of Gradle's object definitions (speaking from a 10 year perspective).  They (the gradle company) have a propensity for arbitrarily changing interactions with the source sets across newer versions of gradle, so my goal is to reduce the burden encountered when we need to support new versions of Gradle, and since upgrading Gradle is often required for new JDK versions we cannot easily say "stick with an older version."  This is movement from a "standard" layout to a non-standard layout.  We need to gain something from that for it to be a good idea.

Why do the classpaths even need to be isolated from the task execution?

diega: I'm with you about minimizing complexity within Gradle. The introduction of a different _sourceSet_ is intended to do exactly that. On `main`, the classpath of the **test** and the classpath of the **refereceTest** were hand picked by including/excluding specific `*.class` files, which looked a bit forced to me. Also, what I'm trying to do here with the _sourceSet_ is to remain as minimal as possible. This feature was introduced in version [0.8](https://docs.gradle.org/0.8/userguide/java_plugin.html#N112C9) and hasn't changed since then, so the scenario where they break compatibility seems pretty unlikely to me. I think that the introduction of this new _sourceSet_ would turn visible a preexisting de facto separation of classpaths, clear up the dependencies of the **referenceTests** and integrate seamlessly with IntelliJ. For me, the main point of this PR is that this allowed me to just update _these_ tests, instead of mixing those changes between the update of the `:ethereum:core` _unit_ tests

shemnon: If you are willing to take strong ownership over this I'll allow it.

shemnon:
This patch currently does not work  I downloaded the patch and ran `./gradlew clean; ./gradlew install; ./gradlew referenceTests`.  None of the reference tests from the ethereum submodule ran.  I ran `git submodule update --init --recursive`.  The build finished again.  However none of the submodule tests ran.

* I would expect the build to fail if the reference tests submodule has not been updated or does not exist yet
* I would expect the build to fail if only the tests in the non-submodule were generated. Only the 5 tests used for test testing were generated and executed.

This second issue may have been a gradle cache issue or an intellij issue, however that feeds into my general distrust of moving large directories around to non-standard places.  A full clean and intellij restart fixed it.  If there is a migration process it needs to be socialized to all the maintainers.

ethereum/referencetests/build.gradle
shemnon: If we are going to commit to source sets for reference tests then the tasks that generate the tests (blockchainReferenceTests, generalstateReferenceTests, generalstateRegressionReferenceTests) should use source sets in their task definition instead of manually coded paths.

diega:
@shemnon I changed the name of the task from `referenceTest` to `referenceTests` to match the existing CI. To try your workflow, these were my steps
```shell
$ git clone git@github.com:hyperledger/besu.git
$ cd besu
$ curl -L https://github.com/hyperledger/besu/pull/3850.patch | git apply --index
$ git submodule update --init --recursive
$ ./gradlew clean
$ ./gradlew referenceTests
```
and they ran fine.
At some point I applied the patch without the `--index` flag (important, so git is able to handle the submodule move properly) and when I tried to run `./gradlew referenceTests` I got the expected error for when the reference tests are missing
```
> Task :ethereum:referencetests:referenceTest FAILED

* Where:
Build file '/Users/diegoll/dev/upstream/besu/ethereum/referencetests/build.gradle' line: 111

* What went wrong:
Execution failed for task ':ethereum:referencetests:referenceTest'.
> ethereum/referencetest/src/reference-test/external-resources/README.md missing: please clone submodules (git submodule update --init --recursive)
```
Regarding IntelliJ I didn't have any trouble, but it's good to know that a refresh of the Gradle context solved the situation.

It this was merged the only extra step needed is to run `git submodule update --init --recursive` in the same way it was needed for #3835

shemnon:
Since we are planning to go to Java 17 in 22.7.0 I was wondering if we should batch together all these changes that break developer workflow into one release.

Is it possible to break apart the changes that don't involve moving the submodule from the changes that do require it?  Does it make 2 useful PRs?

ethereum/referencetests/build.gradle
diega: Yes, I can take ownership over this and will take special care when we are willing to upgrade Gradle/JVM versions.
Regarding using _sourceSets_ for the other tasks, I don't see if they would suit fine given the nature of the tasks we are talking here. _sourceSets_ [are meant to be used](https://docs.gradle.org/7.4.2/userguide/building_java_projects.html#sec:java_source_sets) within a compile/build cycle of JVM-based languages, and the tasks here are for generating sources to be used **in** that cycle but for **referenceTest**. Even if we could find a way, I think we are getting enough integration between the new _sourceSet_ and the tasks by linking the `srcDirs` of it with the tasks _outputs_

diega:
I don't think it would be possible to split this PR in two meaningful ones. The first two commits are splitted just for the sake of the history, and the [third one](https://github.com/diega/besu/commit/junit5_referenceTests~1), where I'm moving the submodule, is mandatory for the final migration of just the **referenceTests** to Junit5.
What's the disturbance that you are experiencing with your developer workflow? I can take a look if there is something I can do to improve the PR to avoid that.
Anyway, if it works better for you, I can keep this branch updated until it's time to release `22.7.0`

shemnon:
If we add in a check to ensure that the submodule is present we can bring it in before a quarterly release then.

This sequence would have to fail and the error message would need to have instructions on the git incantations to run (`git submodule update --init --recursive` I presume)
```
$ rm -rf besu
$ git clone git@github.com:hyperledger/besu.git
$ cd besu
$ ./gradlew referenceTests
```

shemnon:
Looks like if I read top down instead of bottom up I would have seen this - https://github.com/hyperledger/besu/pull/3850#issuecomment-1136121619 - which satisfies me.  Let me run it later tonight.  I'll be offline for a bit.

shemnon (approved):

macfarla:
> It this was merged the only extra step needed is to run `git submodule update --init --recursive` in the same way it was needed for #3835

would it make sense to add this to the changelog - is this effectively a breaking change?

shemnon:
It's a developer experience issue, that is documented when the build fails in the error message how to mitigate.  Devs won't be cluless, plus we should announce this in the contributors room when it merges.

So that depends on the audience of the changelog, is it end users or developers or both?

macfarla:
> It's a developer experience issue, that is documented when the build fails in the error message how to mitigate. Devs won't be cluless, plus we should announce this in the contributors room when it merges.
> 
> So that depends on the audience of the changelog, is it end users or developers or both?

A message in the contributors chat is also good - but ephemeral - if I'm away that week I wouldn't see it - whereas the changelog is a long-lived record. Then again, if the build didn't work a dev would just ask hey what's up with this. 

I think what I'm saying is - I'm happy with a message in besu-contributors

PR #3861, opened 2022-05-19
removes verbose memory tracking, and adds jdk to include jcmd and tools

jflo:


siladu (approved):
LGTM

PR #3872, opened 2022-05-19
Add Network Upgrades & Anvil Default Accounts to Dev Genesis

RatanRSur:
Signed-off-by: Ratan (Rai) Sur <ratan.r.sur@gmail.com>

## PR description
My EIP-1559 tx got rejected by the dev network so I added all the new upgrades :)

The next part is more of a feature. I added the default accounts for `anvil` to streamline the local node -> besu pipeline. This way it can be a drop in replacement.

## Documentation

## Changelog


shemnon:
My google-fu is weak.  Can you provide a pointer to anvil?

Also, can you (as part of the comment element) explain where these keys come from so we aren't creating more "revealed" key material?  Such as "Anvil demo user number 1"

I think we made the "new key" mistake for the first three accounts.

RatanRSur:
You're not alone haha. No one can google for anvil sucessfully. https://github.com/foundry-rs/foundry/tree/master/anvil

Sure, I'll update the comments.

shemnon:
I love that the check word for the mnemonic phrase is 'junk'

shemnon:
I would approve, but there is a unit test broken by adding the istanbulBlock field. I'm not picky about how it's resolved.

iamhsk:
This PR should be merged after fixing the broken tests. Updating the dev genesis file story can be found here: #4125

macfarla:
@RatanRSur I'm assuming you've done what you need with this one and moved on. We've created an issue to break the dependency between the privacy tests and the dev genesis file. It's in our backlog but we're working on other priorities for now.

macfarla:
closing this PR. Needs further work to decouple tests from the dev.json as per #4125

PR #3886, opened 2022-05-23
Test LZ4_COMPRESSION on all RocksDB column families

ahamlat:
By default, RocksDB uses Snappy compression algorithm, but from[ RocksDB documentation](https://github.com/facebook/rocksdb/wiki/Compression#configuration), we can read that "LZ4 is almost always better than Snappy".

## Documentation

## Changelog

### DIFF
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
@@ -51,6 +51,7 @@ import org.rocksdb.CompressionType;
 import org.rocksdb.DBOptions;
 import org.rocksdb.Env;
 import org.rocksdb.LRUCache;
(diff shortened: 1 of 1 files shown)


jclapis:
image

Confirmed, Besu ended up using ~13 GB (6.5 GB in RAM, 6.5 GB in swap) with this changeset. Testing the old Snappy behavior had it at ~3 GB in RAM and no swap. I'll test again with LZ4 enabled but none of the other explicit cache settings specified and see how it does.

jclapis:
image

After running LZ4 without any of the other manual settings (i.e. using [this build](https://github.com/garyschulte/besu/commit/f9fc1b11b37e9579647feffd9863e327330e106e)) Besu has been happily syncing on Mainnet with my Pi using 4.5 GB of RAM and no swap space. It's at ~24% after 5 hours. This configuration is quite promising.

ahamlat:
@garyschulte and @jclapis : Sorry for the inconvenience, this is a draft PR, so I used the same PR to do some other tests related to RocksDB configuration, not only LZ4 compression.

ahamlat:
Other new tests are done in this PR #4166

PR #3911, opened 2022-05-27
Do not require a minimum block height when downloading headers or blocks

fab-10:
After The Merge, there will be no more broadcating of the following messages:
* `NewBlock`
* `NewBlockHashes`

and we used these messages to update the status (height and difficulty) of a peer.
The status of a peer is set at the beginning of the connection, and then update using the above messages, but after the switch to PoS, since there messages are no longer sent, the status of a peer will remain the same for the duration of the connection, and this could mean that Besu is not able to query for block headers or blocks anymore, since some of the messages currently enforce that a peer has minimum height before we send them.

This PR address this issue, removing the constraint that a peer must have a minimum height, before sending block header or block requests, since we actually do not known which is the current status of the peer, and also try to parse peer responses for information that could be useful to update the peer status.

Even if the removal of the constraint is mandatory for the Merge, it could also benefit PoW networks, since assuming that the records keep for any peer is always up-to-date, but this could not be the case, since the peer could send some messages only to a subset of its peers, or some messages could not be processed.
This shold not introduce too much overhead, because Besu tries first with the best peers, the one that should be able to respond to our request.

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/RetryingGetHeaderFromPeerByHashTask.java
```diff
@@ -48,7 +47,6 @@
       final MetricsSystem metricsSystem) {
     super(ethContext, 3, List::isEmpty, metricsSystem);
     this.protocolSchedule = protocolSchedule;
-    this.minimumRequiredBlockNumber = minimumRequiredBlockNumber;
```
macfarla, line 51: should minimumRequiredBlockNumber be removed from the constructor also? L47

macfarla (approved):
Looks ok to me. minor comment

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/RetryingGetHeaderFromPeerByHashTask.java
fab-10: good catch, removed

PR #3920, opened 2022-05-30
Clear correctly heal pending task before restart

matkt:

## PR description

Properly clean the pending tasks when changing the pivot block in order to avoid blocking the snapsynnc process

## Documentation

## Changelog


ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/DynamicPivotBlockManagerTest.java
```diff
@@ -119,4 +119,32 @@ public void shouldSwitchToNewPivotBlockWhenNeeded() {
     verify(snapSyncState).setCurrentHeader(pivotBlockHeader);
     verify(fastSyncActions).waitForSuitablePeers(any());
   }
+
+  @Test
+  public void shouldSwitchToNewPivotOnlyOnCe() {
```
fab-10, line 124: s/OnCe/Once/?

fab-10 (approved):
LGTM, please add a CHANGELOG entry

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/DynamicPivotBlockManagerTest.java
garyschulte: nah he meant Only on windows CE ;)

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/DynamicPivotBlockManager.java
```diff
@@ -93,9 +93,9 @@ public void switchToNewPivotBlock(final BiConsumer<BlockHeader, Boolean> onSwitc
           LOG.info(
               "Select new pivot block {} {}", blockHeader.getNumber(), blockHeader.getStateRoot());
           syncState.setCurrentHeader(blockHeader);
+          lastPivotBlockFound = Optional.empty();
```
garyschulte, line 96: Since this is getting set in both `check` and `switchToNewPivotBlock` , perhaps we should use a `final AtomicReference<BlockHeader> instead of a mutable `Optional<BlockHeader>`

IDK if the snap pipeline might be calling these concurrently, but it seems possible.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/DynamicPivotBlockManagerTest.java
matkt: haha. good catch

fab-10:
please add a CHANGELOG entry

PR #3922, opened 2022-05-31
Temporaly disable sonar integration

lucassaldanha:


shemnon:
Not terribly sad to see it go (having failures occur post-submit that you couldn't trigger pre-submit is a feel-bad).  But not the way it deserved to go.

lucassaldanha:
I believe GH Actions has better control over secrets and public PRs (at least more granularity). We could use it to run the Sonarcloud analysis if we want.

macfarla (approved):
LGTM - let's investigate whether we can use GitHub actions to run this in the future

PR #3936, opened 2022-06-03
do not disconnect a peer on re-pivot race

garyschulte:
## PR description
Resolves a race where we query peer for a pivot block, and the pivot changes before the peer responds.  Previously this would disconnect the peer and log a warning.  Now, instead we re-try with the peer with the new pivot block and only disconnect the peer if the hash does not match the original requested pivot.

## Documentation

## Changelog


jflo (approved):
Switching to completed futures and composing them seems more safe.

macfarla (approved):

Issue #3890, opened 2022-05-24
Merge bug: besu producing zero transaction blocks


garyschulte:
### Description
besu/consensus client combinations on mainnet-shadowfork-5 are producing blocks with zero transactions.

metrics indicate that besu has stopped seeing new transactions, possibly since the merge.

### Steps to Reproduce (Bug)
1. log into msnf5 node, observe transaction pool depth of 0

garyschulte:
This appears to be a problem that occurs at merge transition that prevents remote transactions from getting added to besu's transaction pool.  Short term, this has a very simple mitigation: just restart besu after the merge and remote transactions make it into the txpool just fine.

jflo:
We know why Besu has been proposing empty blocks after hitting TTD. It turns out, that incoming transactions cannot be added to the transaction pool unless besu is synced within 100 blocks of its best peer. When one of those peers is a vanilla mainnet peer, and besu is on a shadowfork, that can be a problem.

jflo:
It seems this functionality has caused about a 50% reduction in the number of peers that Besu can maintain on Ropsten.  Reopening to investigate and address.

jflo:
Current case can be seen on Goerli Shadowfork 5.  The Prysm(bonsai) and Teku (bonsai) instances are the only ones that are submitting proposals with transactions, and has txs in their pools.  They both had been restarted at some point more than 13 hours after TTD.

PR #3937, opened 2022-06-05
does not require being in sync to accept remote transactions

jflo:

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/NewPooledTransactionHashesMessageProcessor.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/NewPooledTransactionHashesMessageProcessor.java
@@ -42,7 +42,6 @@ import org.slf4j.LoggerFactory;
 public class NewPooledTransactionHashesMessageProcessor {
 
   private static final int SKIPPED_MESSAGES_LOGGING_THRESHOLD = 1000;
-  private static final long SYNC_TOLERANCE = 100L;
 
   private static final Logger LOG =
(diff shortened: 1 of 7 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/NewPooledTransactionHashesMessageProcessor.java
```diff
@@ -111,26 +108,25 @@ private void processNewPooledTransactionHashesMessage(
           incomingTransactionHashes::size,
           incomingTransactionHashes::toString);
 
-      if (syncState.isInSync(SYNC_TOLERANCE)) {
```
garyschulte, line 114: instead of removing this check entirely, what about having this class subscribe to Synchronizer in sync events?  the first in-sync event would 'open the floodgates':  something like:
```
    syncState.subscribeInSync(inSync -> {
      if (inSync) {
        acceptingRemoteTransactions.set(inSync);
      };
    }, 100000L);
```

garyschulte (approved):
non-blocking suggestion to implement as a one-way check rather than no check.  LGTM

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/NewPooledTransactionHashesMessageProcessor.java
jflo: That will only work if we can achieve sync prior to any PoW clients peering with us.

jflo:
fixes #3890

Issue #3933, opened 2022-06-02
org.rocksdb.RocksDBException: Invalid column family specified in write batch


qntoine:
### Description
Using Besu through Rocket Pool (1.4.1), I am getting the following error logs when trying to sync using snap sync.
The error seems to occur once the initial snap sync is completed and the client switches to fast sync. 

### Logs
```
eth1_1           | java.util.concurrent.CompletionException: org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Invalid column family specified in write batch
eth1_1           |      at java.base/java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:368)
eth1_1           |      at java.base/java.util.concurrent.CompletableFuture.completeRelay(CompletableFuture.java:377)
eth1_1           |      at java.base/java.util.concurrent.CompletableFuture$UniRelay.tryFire(CompletableFuture.java:1097)
eth1_1           |      at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
eth1_1           |      at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
eth1_1           |      at org.hyperledger.besu.services.pipeline.Pipeline.abort(Pipeline.java:180)
eth1_1           |      at org.hyperledger.besu.services.pipeline.Pipeline.lambda$runWithErrorHandling$3(Pipeline.java:159)
eth1_1           |      at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
eth1_1           |      at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
eth1_1           |      at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
eth1_1           |      at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
eth1_1           |      at java.base/java.lang.Thread.run(Thread.java:833)
eth1_1           | Caused by: org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Invalid column family specified in write batch
eth1_1           |      at org.hyperledger.besu.plugin.services.storage.rocksdb.segmented.RocksDBColumnarKeyValueStorage$RocksDbTransaction.commit(RocksDBColumnarKeyValueStorage.java:277)
[...]

Issue #3924, opened 2022-05-31
RocksDBException: Invalid column family specified in write batch


jtraglia:
### Description

Sorry, I don't have many details. Ran into this issue when setting up ropsten nodes last week:

```
2022-05-26 03:53:01.769+00:00 | EthScheduler-Services-85 (importBlock) | ERROR | PipelineChainDownloader | Chain download failed. Restarting after short delay.
java.util.concurrent.CompletionException: org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Invalid column family specified in write batch
	at java.base/java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:368)
	at java.base/java.util.concurrent.CompletableFuture.completeRelay(CompletableFuture.java:377)
	at java.base/java.util.concurrent.CompletableFuture$UniRelay.tryFire(CompletableFuture.java:1097)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2194)
	at org.hyperledger.besu.services.pipeline.Pipeline.abort(Pipeline.java:180)
	at org.hyperledger.besu.services.pipeline.Pipeline.lambda$runWithErrorHandling$3(Pipeline.java:159)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:577)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1585)
Caused by: org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Invalid column family specified in write batch
	at org.hyperledger.besu.plugin.services.storage.rocksdb.segmented.RocksDBColumnarKeyValueStorage$RocksDbTransaction.commit(RocksDBColumnarKeyValueStorage.java:277)
	at org.hyperledger.besu.services.kvstore.SegmentedKeyValueStorageTransactionTransitionValidatorDecorator.commit(SegmentedKeyValueStorageTransactionTransitionValidatorDecorator.java:49)
[...]

PR #3940, opened 2022-06-06
fix Invalid column family specified in write batch

matkt:

## PR description

I analyzed the behavior of this error and it seems to appear every time when we change the pivot block.

When we do this we are in effect deleting the columns from the flat database to recreate  a new one. It seem that this has an impact on the writing of the data because it point to the old columns that we have deleted.

I saw that the rocksdb code allows to ignore writes that point to an inexisting column
https://github.com/facebook/rocksdb/blob/49628c9a83d337ea7b01f7ad3bdffa340e013913/db/write_batch.cc#L1707

Knowing that this column is no longer useful it does not matter to not write on this column anymore but it is important to write on the others and this feature is doing what we want https://javadoc.io/static/org.rocksdb/rocksdbjni/5.13.3/org/rocksdb/WriteOptions.html#ignoreMissingColumnFamilies() 

## Fixed Issue(s)

fixes https://github.com/hyperledger/besu/issues/3933
fixes https://github.com/hyperledger/besu/issues/3924

## Documentation

## Changelog


jflo (approved):

macfarla (approved):
LGTM

macfarla (approved):
LGTM

PR #3976, opened 2022-06-16
adds ttd block to sepolia config

jflo:


daniellehrner:
Unrelated to the TTD change, but there is another bootnode that we could add to the config: https://github.com/goerli/sepolia

PR #3944, opened 2022-06-07
3890 disconnect pow net

jflo:
## PR description

- Takes note of the total difficulty of the final PoW block, pow_d.
- Upon receipt of the second finalized block, disconnect any peers that have a total difficulty over pow_d (likely not many).
- Upon receipt of the second finalized block, disconnect incoming new peers that have a total difficulty over pow_d.
- Upon receipt of the second finalized block, start rejecting connections from peers sending NEW_BLOCK or NEW_BLOCK_HASH messages.
- Upon receipt of the second finalized block, stop all block propagation.

## Fixed Issue(s)

fixes #3890 
fixes #3943 

## Documentation

- [X] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [X] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #3978, opened 2022-06-16
Release 22.4.3

jflo:
reverts PR for disconnecting from the pow network. Updates version number.


macfarla:
Just so it's part of the record - this reverted #3944

PR #3983, opened 2022-06-17
update the docker container versions for Tessera and GoQuorum

pinges:
On my new M1 mac book the Tessera docker wouldn't start. Updating the Tessera docker version fixed that.


macfarla:
only one AT failed so it must be flaky!

macfarla (approved):

PR #3969, opened 2022-06-16
Have consistent warning between CLI, ENV and TOML config

wcgcyx:
## PR description
Add additional check of option in environment variables and config file for detecting ignored dependent config warning.

## Fixed Issue(s)
fixes #2069

PR #3993, opened 2022-06-20
Additional testings on ignored option warning 

wcgcyx:
## PR description
Add tests on options configured through ENV

## Fixed Issue(s)
Follow-up PR to address comments from https://github.com/hyperledger/besu/pull/3969

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
@@ -111,11 +111,11 @@ public class CommandLineUtils {
   }
 
   private static boolean isOptionSet(final CommandLine.Model.OptionSpec option) {
-    CommandLine commandLine = option.command().commandLine();
+    final CommandLine commandLine = option.command().commandLine();
     try {
       return !option.stringValues().isEmpty()
           || !Strings.isNullOrEmpty(commandLine.getDefaultValueProvider().defaultValue(option));
-    } catch (Exception e) {
+    } catch (final Exception e) {
       return false;
     }
   }
--- a/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
@@ -14,13 +14,13 @@
  */
 package org.hyperledger.besu.cli.util;
 
-import com.google.common.base.Strings;
 import org.hyperledger.besu.util.StringUtils;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import com.google.common.base.Strings;
 import org.slf4j.Logger;
 import picocli.CommandLine;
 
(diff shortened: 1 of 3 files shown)


besu/src/main/java/org/hyperledger/besu/cli/util/CommandLineUtils.java
```diff
@@ -99,19 +100,23 @@ public static void checkMultiOptionDependencies(
...
+                Arrays.stream(option.names()).anyMatch(dependentOptionsNames::contains)
+                    && isOptionSet(option))
         .map(option -> option.names()[0])
         .collect(
             Collectors.collectingAndThen(
                 Collectors.toList(), StringUtils.joiningWithLastDelimiter(", ", " and ")));
   }
+
+  private static boolean isOptionSet(final CommandLine.Model.OptionSpec option) {
+    CommandLine commandLine = option.command().commandLine();
```
usmansaleem, line 114: nit: use final keywork with variable names unless they are meant to be assigned another value later on: `final CommandLine commandLine...`. Similarly, `catch(final Exception e) {`

besu/src/test/java/org/hyperledger/besu/cli/CommandLineUtilsTest.java
```diff
@@ -53,9 +63,19 @@ private abstract static class AbstractTestCommand implements Runnable {
 
     final CommandLine commandLine;
 
-    AbstractTestCommand(final Logger logger) {
+    final HashMap<String, String> environment = new HashMap<>();
```
usmansaleem, line 66: nit: `final Map<String, String> environment = new HashMap<>();`

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -1249,6 +1249,33 @@ public void p2pOptionsRequiresServiceToBeEnabledToml() throws IOException {
     assertThat(commandErrorOutput.toString(UTF_8)).isEmpty();
   }
 
+  @Test
```
jframe, line 1252: After looking at the tests in CommandLineUtilsTest, I think we leave these env tests out.

besu/src/test/java/org/hyperledger/besu/cli/CommandLineUtilsTest.java
```diff
@@ -53,9 +64,19 @@ private abstract static class AbstractTestCommand implements Runnable {
...
     final CommandLine commandLine;
 
-    AbstractTestCommand(final Logger logger) {
+    final Map<String, String> environment = new HashMap<>();
+
+    AbstractTestCommand(final Logger logger, final Optional<File> configFile) {
       this.logger = logger;
       commandLine = new CommandLine(this);
+      if (configFile.isPresent()) {
+        commandLine.setDefaultValueProvider(
```
siladu, line 73: Rather than setting the provider here and having to pass the configFile up the class hierarchy, I think it would be better to set it directly in the `multipleMainOptionsToml`, especially since that's the only test than needs it.

You should have access to everything you need via `testCommand`

siladu (commented):
Couple of comments, mainly that CommandLineUtilsTest can be cleaned up a bit to avoid a lot of code changes.

Minor point about splitting the filters up again

jframe (approved):
LGTM

siladu (approved):
LGTM

PR #3994, opened 2022-06-20
When on PoS the head can be only be updated by ForkchoiceUpdate

### LINKED ISSUE #3957: eth_getBlockByNumber should support latest as fcu head
Per the definition of the `latest` tag here:

https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false

eth_getBlockByNumber should support a `latest` tag that responds with whatever block was most recently set as the head via an `engine_forkChoiceUpdated` call from the consensus layer.

Suggest implementing by making the `eth_getBlockByNumber` rpc method register as a forkchoice updated listener with the MergeContext.

fab-10:
When executing a newPayload or build a new block, we do not need to move
the chain head and update the world state, since this will be done by a
following forkchoice update call, but we still need to validate the block
and doing so we can also prepare everything for the future call, so we do
not need to re-execute everything, but only update the pointers, so that the
response to the forkchoice update call is quick.

Moreover on block proposal do only a validation of the block without storing or remembering nothing, to avoid saving data that could be stale, for example the empty block that is always proposed, is useless if another better block is produced later.

## Fixed Issue(s)

fixes #3957
## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiPersistedWorldState.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiPersistedWorldState.java
@@ -16,6 +16,7 @@
 
 package org.hyperledger.besu.ethereum.bonsai;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static org.hyperledger.besu.ethereum.bonsai.BonsaiAccount.fromRLP;
 import static org.hyperledger.besu.ethereum.bonsai.BonsaiWorldStateKeyValueStorage.WORLD_BLOCK_HASH_KEY;
 import static org.hyperledger.besu.ethereum.bonsai.BonsaiWorldStateKeyValueStorage.WORLD_ROOT_HASH_KEY;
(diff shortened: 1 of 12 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -308,15 +323,18 @@ private BlockAddedEvent appendBlockHelper(final BlockWithReceipts blockWithRecei
...
 
-    // Update canonical chain data
-    final BlockAddedEvent blockAddedEvent = updateCanonicalChainData(updater, blockWithReceipts);
-
-    updater.commit();
-    if (blockAddedEvent.isNewCanonicalHead()) {
-      updateCacheForNewCanonicalHead(block, td);
+    final BlockAddedEvent blockAddedEvent;
+    if (storeOnly) {
+      blockAddedEvent = handleStoreOnly(blockWithReceipts);
```
garyschulte, line 328: Does this mean that if/when we "rewind" to this future block in forkchoiceUpdated that we are not going to fire a an event of type `HEAD_ADVANCED` anymore?

garyschulte (commented):
Question about HEAD_ADVANCED event types, but this looks great.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
fab-10: good catch! Indeed this need to be reviewed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiPersistedWorldState.java
```diff
@@ -236,6 +237,35 @@ protected Hash calculateRootHash(
     return Hash.wrap(rootHash);
   }
 
+  @Override
+  public void remember(final BlockHeader blockHeader) {
+    checkArgument(blockHeader != null, "Block header must not be null");
+
+    final BonsaiWorldStateUpdater localUpdater = updater.copy();
```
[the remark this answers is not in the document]
gezero: The copy method will disappear, I will see this as fail of build when I will try to merge.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
fab-10: @garyschulte please check now, this commit
[Introduce forwardToBlock](https://github.com/hyperledger/besu/pull/3994/commits/b02de746dd560d68e227f108eebd8dbde214428b)

It moves head to the block sent by a previous newPayload, and triggers
the advanced head event that will also update the Bonsai world state.

garyschulte (approved):
LGTM 👍

fab-10:
Thanks @garyschulte for the review, I am closing this, since it is now part of this one https://github.com/hyperledger/besu/pull/4013, that contains also other changes to avoid persisting state when preparing new payloads.

PR #3997, opened 2022-06-21
Add process_release metric.

mcdee:
## PR description

Add a `process_release` metric containing release information.  This shows up as following:

```
process_release{version="besu/v22.4.1-dev-fe8ad2d5/linux-x86_64/openjdk-java-11",} 1.0
```

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## 22.7.1
 
 ### Additions and Improvements
+- Version information available in metrics [#3997](https://github.com/hyperledger/besu/pull/3997)
+
 ### Bug Fixes
 
 ## 22.4.3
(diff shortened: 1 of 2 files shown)


jflo (approved):
Thanks for the contrib!

shemnon:
Metrics is the wrong place for this.  Long term it will increase the cardinality of the measures and really isn't a metric of operation.  This is more appropriate for log output, and is, in fact, already present in log outputs.

shemnon (changes requested):
Metrics is the wrong place for this. Long term it will increase the cardinality of the measures and really isn't a metric of operation. This is more appropriate for log output, and is, in fact, already present in log outputs.

mcdee:
@shemnon logs aren't easily available remotely, and commonly this sort of information leaves the logs relatively quickly as they rotate or are truncated.  All beacon nodes provide some form of this information in their metrics, and it has proven useful to be able to obtain information about the deployed version across multiple instances in larger operational environments.

As for the cardinality, we're talking 20-30 per year and this is insignificant compared to the volume of data being gathered elsewhere.

shemnon:
Its a bad devops practice.  What is the measure of the gauge?  The real measure is coming through a label instead of the metric itself, the metric should be what is queried, not the labels.

> All beacon nodes provide some form of this information in their metrics
Can you provide concrete examples?

influ7:
Thanks

PR #4001, opened 2022-06-22
EthPeers unique

pinges:

## PR description

To fix some issues that have been reported this PR makes sure that in EthPeers only has one EthPeer for each individual peer.


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -383,52 +409,83 @@ && getConnectionCount() >= maxConnections) {
...
+
     // Track this new connection, deduplicating existing connection if necessary
     final AtomicBoolean newConnectionAccepted = new AtomicBoolean(false);
-    final RlpxConnection inboundConnection = RlpxConnection.inboundConnection(peerConnection);
+    final Peer peer = peerConnection.getPeer();
+    final RlpxConnection newConnection =
+        isInboundConnection
+            ? RlpxConnection.inboundConnection(peerConnection)
+            : RlpxConnection.outboundConnection(
+                peer, CompletableFuture.completedFuture(peerConnection));
```
pinges, line 427: Create an issue for that

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/rlpx/connections/RlpxConnection.java
```diff
@@ -37,7 +37,10 @@ public static RlpxConnection inboundConnection(final PeerConnection peerConnecti
   }
 
   public static RlpxConnection outboundConnection(
-      final Peer peer, final CompletableFuture<PeerConnection> future) {
+      final Peer peer,
+      final CompletableFuture<PeerConnection>
+          future) { // TODO: we could get rid of the completable future for the
```
pinges, line 42: remove TODO

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeer.java
```diff
@@ -567,6 +522,18 @@ public Optional<BlockHeader> getCheckpointHeader() {
...
 
+  public void replaceConnection(final PeerConnection peerConnection) {
+    connection = peerConnection;
+  }
+
+  public boolean statusHasBeenReceived() {
+    return connection.statusHasBeenReceived();
+  }
+
+  public void setLastRequestTimestamp(final int protocolVersion) {
```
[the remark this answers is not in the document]
macfarla: fixed

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapProtocolManager.java
```diff
@@ -137,7 +137,10 @@ public void processMessage(final Capability cap, final Message message) {
 
   @Override
   public void handleNewConnection(final PeerConnection connection) {
-    ethPeers.registerConnection(connection, peerValidators);
+    // TODO: Talk to Karim about this
```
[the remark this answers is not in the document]
matkt: no I think it's good. but if you can try a snapsync to be sure that would be better

util/src/main/java/org/hyperledger/besu/util/Slf4jLambdaHelper.java
```diff
@@ -44,7 +44,7 @@ public static void infoLambda(
 
   public static void debugLambda(
       final Logger log, final String message, final Supplier<?>... params) {
-    if (log.isDebugEnabled()) {
+    if (log.isInfoEnabled()) {
```
fab-10, line 47: My understanding is that all the log changed from debug to info are for surfacing some more logs needed during the development.
I would like to suggest a different approach, that avoid to make all these changes, that also makes harder to focus on actual changes.
My suggestion is to revert the changes to the log level, and instead use a custom log4j configuration to enable debug level only on the classes you are interested in during the development. This way the change list in the PR will be much smaller, and there is no effort in reverting these changes before merging to main, and avoid the risk of missing to revert some.
I am available to help setting the custom log4j conf, in case you need.

fab-10 (commented):
This is my first review pass, and just made some general comments on the first things that come to mind, for sure I will need to do at least another pass to fully understand the change and give a feedback.

To make it easier for reviewers to understand the change, it will be great if you can write also a description of the major changes your are doing to the main classes, to highlight how the behaviour will change from how it works now.

util/src/main/java/org/hyperledger/besu/util/Slf4jLambdaHelper.java
jflo: Seconded! You can set an environment var much like:  

`LOG4J_CONFIGURATION_FILE=/Users/jflo/src/besu_log4j.xml`

And provide a log4j config xml and set the logging level per class or package as you need to.

jflo (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -85,24 +91,110 @@ public EthPeers(
...
-            peerValidators,
-            clock,
-            permissioningProviders);
-    connections.putIfAbsent(peerConnection, peer);
+    synchronized (this) {
+      final Bytes peerId = peerConnection.getPeer().getId();
+      LOG.info(
+          "Registering a connection with peer {}, connection {}",
+          peerId,
+          System.identityHashCode(peerConnection));
```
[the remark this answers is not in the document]
pinges: I thought that this is derived from a memory address and that there should only be a very small chance of it not being unique. I think it is possible that it is reused once the memory has been freed and the memory is reused by the VM. 
This is only used for logging to identify which of the two possible connections to a peer is used and it enables me to follow the lifecycle of a specific connection.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -383,52 +407,83 @@ && getConnectionCount() >= maxConnections) {
...
+
+    // Disconnect if too many peers
+
+    if (!randomPeerPriority) {
```
matkt, line 429: I have the feeling that all these checks are done when the connection comes from a peer. so do you think if we should check isInboundConnection to know if we need to do this part

Issue #3763, opened 2022-04-26
ibft2 ignores --min-gas-price=0


danillosl:
### Description
I'm running a 4 nodes ibft2 from the besu website but apparently the network is ignoring the `--min-gas-price=0`, even though the result of `eth_gasPrice` is `{"jsonrpc" : "2.0", "id" : 53, "result" : "0x0"}` when i try to deploy a contract with truffle i get:

```
> truffle migrate --reset --network besu


Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.


Starting migrations...
======================
> Network name:    'besu'
> Network id:      1337
> Block gas limit: 9007199254740991 (0x1fffffffffffff)


1_initial_migration.js
======================

   Replacing 'Migrations'
   ----------------------
   > transaction hash:    0x21e0a743a9cc684d9a6b21ad3b9c4d72e83c9e672040f1b1b99727abf1b0e4cb
   > Blocks: 0            Seconds: 0
   > contract address:    0x42699A7612A82f1d9C36148af9C77354759b210b
   > block number:        10
   > block timestamp:     1650988775
   > account:             0xFE3B557E8Fb62b89F4916B721be55cEb828dBd73
   > balance:             199.999438700537858346
   > gas used:            203143 (0x31987)
   > gas price:           2.763075578 gwei
   > value sent:          0 ETH
   > total cost:          0.000561299462141654 ETH

   > Saving migration to chain.
   > Saving artifacts
   -------------------------------------
   > Total cost:     0.000561299462141654 ETH


2_deploy_contracts.js
=====================

   Replacing 'contractA'
   --------------------------
   > transaction hash:    0x7ae277c256cb64f8e3487868cdc2f9884d3955ee7c3fe5e47193db08d456bb8e
   > Blocks: 0            Seconds: 0
   > contract address:    0x9a3DBCa554e9f6b9257aAa24010DA8377C57c17e
   > block number:        12
   > block timestamp:     1650988779
   > account:             0xFE3B557E8Fb62b89F4916B721be55cEb828dBd73
   > balance:             199.98087195961082743
   > gas used:            6826558 (0x682a3e)
   > gas price:           2.70141724 gwei
   > value sent:          0 ETH
   > total cost:          0.01844138147105992 ETH


   Replacing 'contractB'
   -------------------------------
   > transaction hash:    0x930918fb20d3ceadeb67db8c9314772c5ee6847cfbc60fede610ce645898c009
[...]

Issue #3374, opened 2022-02-04
Getting the most up to date EVM version without EIP1559


soragXYZ:
Hi!

I have a private network with a gasPrice null, with the EVM version being set to Berlin

I want to update it, in order to have the most up-to-date version of the EVM, but here comes my issue:

The London update contains the EIP1559 which introduce a small burn for each transaction, but with a private network with a gasPrice null it is an issue because it becomes impossible to perform a single transaction if the issuer of the transaction doesn't have funds

So my question is: how can I update my genesis.json to the latest EVM version (arrowGlacier at the time of writing) but without the EIP1559 ?

Thanks

daniellehrner:
I would recommend you to use IBFT2 and configure the network to run with 0 gas. EIP1599 will not take effect if you do that.

IBFT2: https://besu.hyperledger.org/en/stable/HowTo/Configure/Consensus-Protocols/IBFT/
Free gas networks: https://besu.hyperledger.org/en/stable/HowTo/Configure/FreeGas/

soragXYZ:
Hi @daniellehrner thanks for the reply

That's what I did, I'm using IBFT2 with a gasPrice null but it doesn't work, I can't sign a transaction if I don't have funds because for each transaction there is this small fee introduced by the EIP1559

Is it possible to share a working genesis.json configuration ?

ccamaleon5:
Hello, we have the same problem. A network with IBFT 2.0 consensus and 0 gas price cannot be run on the London version of Ethereum, this is due to EIP 1559. Thus, for a network with 0 gas price, the baseFee parameter of the block header should be configurable. What do we need to do to run a network with a gas price 0?

siladu:
Hi @soragXYZ and @ccamaleon5 I believe this PR should fix this issue: https://github.com/hyperledger/besu/pull/4003

It's available on the main branch and hopefully it'll make it into the next release. 
In addition to the code in the PR, you must also include `"baseFeePerGas": "0x0"` at the top level of the genesis file.
We will update the documentation to reflect this.

Please reopen or create a new issue if the problem persists even with this code.

siladu:
The fix has been released: https://github.com/hyperledger/besu/releases/tag/22.7.0-RC1

PR #4003, opened 2022-06-23
Support free gas networks when using London fee market

siladu:
In order to support [free gas private networks](https://besu.hyperledger.org/en/stable/HowTo/Configure/FreeGas/) in a London/EIP1559 context, I propose that we support setting a zero baseFee.

If `baseFeePerGas` is overridden as 0 wei, then bypass the need for a transaction to have a gas price of >= 7 wei.

The genesis would look something like this:
```
{
  "config": {
    ...
    "londonBlock": 0,
    ...
  },
  "baseFeePerGas": "0x0",
  "nonce": "0x0",
  ...
}
```

One potential downside is that once baseFeePerGas is set to 0, there's no going back for that chain. You can still use `--min-gas-price` to enforce _some_ gas, but the London base fee calculation will be forever opted out of as a consequence of the zero baseFee in the genesis block. This also means no ether would get burned since base fee will always be zero.

This is currently blocking users with free gas networks from upgrading past the Berlin fork.

## Fixed Issue(s)

Fixes https://github.com/hyperledger/besu/issues/3374
Fixes https://github.com/hyperledger/besu/issues/3763

## Documentation

## Changelog


antonydenyer:
It's a fiddly one! We talked about previously over at https://github.com/hyperledger/besu/pull/3397 

Should we tie the minimum to what is set in the genesis?

siladu:
> It's a fiddly one! We talked about previously over at https://github.com/hyperledger/besu/pull/3397

It is indeed! I saw that PR, was planning on asking for your review for this ;) 

Pre-London, configuring a free gas network only involved setting --min-gas-price=0 on the nodes but this fix would involve the user setting baseFee to zero as well, which is a bit annoying but documentable.
 
The reason I've gone with that for now is because zero baseFee is also required to pass this validation: https://github.com/hyperledger/besu/blob/645fdd08ee940cb770d2415a7a958b32d0c3b94c/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java#L134-L141

Another option could be to default baseFee to zero and set it explicitly for public networks.

Was trying to get my head around how the fee gets burned and if there would be any implications for this configuration.

shemnon (approved):

siladu:
Keeping open for @garyschulte to look at.

garyschulte (approved):
0 or >=7 seems like a viable way to support both zero gas and "working" base fee networks :+1:

PR #4008, opened 2022-06-23
try updater protection

matkt:

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
@@ -48,7 +48,7 @@ public class BonsaiWorldStateArchive implements WorldStateArchive {
 
   private final Blockchain blockchain;
 
-  private final BonsaiPersistedWorldState persistedState;
+  private BonsaiPersistedWorldState persistedState;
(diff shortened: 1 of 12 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -255,14 +255,13 @@ public Optional<MutableWorldState> getMutable(final Hash rootHash, final Hash bl
     }
   }
 
-  @Override
-  public MutableWorldState getMutable() {
-    return persistedState;
+  private BonsaiPersistedWorldState persistedState() {
```
shemnon, line 258: i'm not sure if this will improve the scenario where two block imports are working at the same time.  BonsaiPersistedWorldState is a gateway to the database, and having two gateways to the same database could introduce a different style of corruption.

shemnon (commented):
Considering that the #4041 fix is working well I don't think we need to ship this in 22.7.0.

One alternate design may be to fix the database state at the last finalized block (or some configurable tailing epoch in the event the CL starts having finalization issues) and use the layered states to access live data.  That way parallel validations are working on an in-memory layered version of the state and having multiple in-memory copies of the same height will be less disruptive.

This of course would involve a global lock when the DB rolls to a finalized blocks, and we would need to drop layered entries from memory after the finalized roll forward.  We also would want to block read requests while the roll occurs too.  It's a fairly large change that would need to land after the merge, or use a different storage strategy.

PR #4018, opened 2022-06-27
Updates to DNS discovery

### LINKED ISSUE #1707: Performance updates to DNS discovery
### Description
- Modify DNS discovery code to not dump all of the results into discovery immediately but slowly add them in batches of roughly 25 at a time
- Modify to not always traverse the DNS tree in the same order, that puts pressure on “earlier” leaves.  Perhaps adding in a random xor and then using that to sort the depth first traversals.
- Modify trie traversal logic to depth first and emitting nodes as trie leafs are encountered

diega:

## PR description
This PR adds a new implementation for [EIP-1459](https://eips.ethereum.org/EIPS/eip-1459) replacing [DNSDaemon](https://github.com/apache/incubator-tuweni/blob/main/dns-discovery/src/main/kotlin/org/apache/tuweni/discovery/DNSDaemon.kt) from Tuweni with a new one, vert.x based.
This new implementation provides the following benefits:
- Random tree traversal
- Batched discovery
- Periodic traversal continues from last state
- Non blocking implementation, based on vert.x DNS client 

## Fixed Issue(s)
Fixes #1707

## Notes
This is sent as a draft to gather feedback on this implementation. How often should we check the DNS for new peers? How many leaves should be attempted to connect per branch? Should this parameters need to be parameterized through the CLI? Based on the willingness to merge this proposal I would finish the specification by adding support to linked trees

## Documentation

## Changelog


atoulme:
@diega is something wrong with the DNSDaemon from Tuweni? I didn't see feedback or anything on Tuweni, any issues worth reporting?

matt-nelson-csi:
Can we engage on this @diega ? Have you done any work to test the implementation and outcomes?

diega:
Thanks @atoulme for taking the time of looking at it.
There is nothing wrong with Tuweni's implementation, but it currently doesn't have any dependency on vert.x and I wanted to take the chance of this requirement to make use of vert.x DNS client to better integrate its execution within the existing event loop (instead of using a `java.util.Timer`) . Having that said, I coded it in a pretty abstract way so if you think this would be suitable to be part of Tuweni I would be happy to port it to Kotlin and send it as a PR.

diega:
@matt-nelson-csi absolutely.
Beside the unit testing part of this PR, I ran it using `--bootnodes=
--discovery-dns-url=enrtree://AKA3AM6LPBYEUDMVNU3BSVQJ5AD45Y7YPOHJLEF6W26QOE4VTUDPE@all.mainnet.ethdisco.net`, which allowed me to tweak some constants like how many records per round try to connect (I chose 100, but we can change that based on some other experience). It would be difficult to establish a comparison between both implementations regarding peers. Eventually both would attempt to connect to the same set of boot nodes but this implementation does it random and incrementally. A next step would be to only request new nodes from the DNS tree only when we are bellow certain number of peers instead of doing it every 60 seconds.

atoulme:
@diega sounds good. FWIW a PR to Tuweni would be awesome, but if this PR here and now does the job, no problem. Interesting on the use of Vert.x, that makes sense. I think dnsjava has been less mature in ways, I never did try the DNSClient offered by Vert.x.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -325,29 +321,37 @@ public boolean removeMaintainedConnectionPeer(final Peer peer) {
...
-      if (!peers.isEmpty()) {
-        dnsPeers.set(peers);
+      final InetSocketAddress host = tcp.orElseGet(udp::get);
+      final Bytes compressedKey = (Bytes) enr.get(EnrField.PKEY_SECP256K1);
+      // final SignatureAlgorithm signatureAlgorithm =
+      //     SignatureAlgorithmType.DEFAULT_SIGNATURE_ALGORITHM_TYPE.get();
+      // final ECPoint decompressedPoint =
+      //
+      // signatureAlgorithm.publicKeyAsEcPoint(signatureAlgorithm.createPublicKey(compressedKey));
+      final ECPoint decompressedPoint = SECP256K1_CURVE.decodePoint(compressedKey.toArrayUnsafe());
```
atoulme, line 342: you might want to test for the case where the key is missing or invalid, and handle expliclity issues arising in this lambda.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -506,10 +506,24 @@ private P2PNetwork doBuild() {
       peerDiscoveryAgent = peerDiscoveryAgent == null ? createDiscoveryAgent() : peerDiscoveryAgent;
       rlpxAgent = rlpxAgent == null ? createRlpxAgent(localNode, peerPrivileges) : rlpxAgent;
 
+      DNSPeerDiscoveryService dnsStreamer = null;
+      final String discoveryURI = config.getDiscovery().getDNSDiscoveryURL();
+      if (discoveryURI != null) {
+        LOG.info("Creating DNS discovery with URL {}", discoveryURI);
+        final DnsClientOptions dnsOptions = new DnsClientOptions();
+        if (config.getDnsDiscoveryServerOverride().isPresent()) {
+          dnsOptions.setHost(config.getDnsDiscoveryServerOverride().get());
```
atoulme, line 515: nit - use ifPresent to combine isPresent and get

gfukushima:
Hi @diega any reasons we ended up closing this PR instead of moving forward to make it to main?

diega:
@gfukushima I don't really remember, but if you find this useful we can reopen it and try to complete the spec

PR #4026, opened 2022-06-28
Add terminal block hash and number to Ropsten genesis file

fab-10:

## Documentation

## Changelog


besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -604,26 +603,43 @@ private WorldStateArchive createWorldStateArchive(
   protected List<PeerValidator> createPeerValidators(final ProtocolSchedule protocolSchedule) {
     final List<PeerValidator> validators = new ArrayList<>();
 
-    final OptionalLong daoBlock = configOptionsSupplier.get().getDaoForkBlock();
-    if (daoBlock.isPresent()) {
-      // Setup dao validator
-      validators.add(
-          new DaoForkPeerValidator(protocolSchedule, metricsSystem, daoBlock.getAsLong()));
-    }
+    configOptionsSupplier
```
shemnon, line 606: This is less readable than an if block.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -604,26 +603,43 @@ private WorldStateArchive createWorldStateArchive(
...
+                validators.add(
+                    new ClassicForkPeerValidator(protocolSchedule, metricsSystem, classicBlock)));
 
     for (final Map.Entry<Long, Hash> requiredBlock : requiredBlocks.entrySet()) {
       validators.add(
           new RequiredBlocksPeerValidator(
               protocolSchedule, metricsSystem, requiredBlock.getKey(), requiredBlock.getValue()));
     }
 
+    configOptionsSupplier
```
shemnon, line 628: This is not as readable as a couple of `if` blocks.  Two if blocks would make it obvious we are adding a config if two config items are persent.  get()\.get<foo>().filter(...).ifPresent(...) is much too verbose, especially when one of the ifPresent paths is nested.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
fab-10: I usually find the lambda version more effective to describe what to do in case of something is present, and less error prone, but I can easily read both versions, so I will revert.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
fab-10: done

macfarla (approved):
LGTM

macfarla (approved):

PR #4037, opened 2022-07-01
Enforce RocksDB to respect block cache size

ahamlat:
Enforce RocksDB to respect Block cache size, knowing that we can modify this block cache size with --Xplugin-rocksdb-cache-capacity.

We can read from [RocksDB Documentation](https://github.com/facebook/rocksdb/wiki/Block-Cache) that :
> strict_capacity_limit: In rare case, block cache size can go larger than its capacity. This is when ongoing reads or iterations over DB pin blocks in block cache, and the total size of pinned blocks exceeds the capacity. If there are further reads which try to insert blocks into block cache, if strict_capacity_limit=false(default), the cache will fail to respect its capacity limit and allow the insertion. This can create undesired OOM error that crashes the DB if the host don't have enough memory. Setting the option to true will reject further insertion to the cache and fail the read or iteration. The option works on per-shard basis, means it is possible one shard is rejecting insert when it is full, while another shard still have extra unpinned 

## Documentation

## Changelog

### DIFF
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
@@ -148,7 +148,7 @@ public class RocksDBColumnarKeyValueStorage
   }
 
   private BlockBasedTableConfig createBlockBasedTableConfig(final RocksDBConfiguration config) {
-    final LRUCache cache = new LRUCache(config.getCacheCapacity());
+    final LRUCache cache = new LRUCache(config.getCacheCapacity(), -1, true);
     return new BlockBasedTableConfig()
         .setBlockCache(cache)
         .setFormatVersion(5)


plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -148,7 +148,7 @@ public RocksDBColumnarKeyValueStorage(
   }
 
   private BlockBasedTableConfig createBlockBasedTableConfig(final RocksDBConfiguration config) {
-    final LRUCache cache = new LRUCache(config.getCacheCapacity());
+    final LRUCache cache = new LRUCache(config.getCacheCapacity(), -1, true);
```
ajsutton, line 151: Wouldn't we also need to handle the new read failures we could now get if the cache is full? We'd need to have something that provides back-pressure to limit the rate of reads we attempt until the DB has cache available again and can serve them. Otherwise we'll wind up at best spamming the logs with scary exceptions as reads fail and possibly wind up rejecting blocks we should accept and causing a chain fork.

ahamlat: Good point, but my understanding is that when the cache is full and the insertion is rejected, it is handled internally and the KV entry is retrieved from SST files. I think the side effect here is more a performance degradation because of the cache misses and more IO to retrieve these missed entries from disk (SST files).

ahamlat: Some errors happens and the nodes 2 of the four nodes crashed.
By enabling Jemalloc and setting block cache without forcing it resolved the native memory issue, I'm going then to close this one.

Issue #3369, opened 2022-02-03
SPIKE: add metrics for RocksDB


macfarla:
Currently we have no metrics around RocksDB. Investigate what would be involved. What metrics make sense?
 - [ ] useful metrics
 - [ ] not detrimental to performance

fab-10:
@macfarla it seems that RocksDB metrics are already collected, see `RocksDBMetricsFactory` for example, but I am not able to see them exposed by Prometheus, so there could be some issues on exposing these metrics.

ajsutton:
That category may just be disabled by default.  From memory there were some categories disabled because they included timers I think which had a significant performance impact.  Quite possibly we should just remove the timers and enable the rest.

macfarla:
ok thanks @fab-10 and @ajsutton - that gives us somewhere to start!

macfarla:
There are 4 timers, 2 gauges and rollbackCount
  public RocksDBMetrics(
      final OperationTimer readLatency,
      final OperationTimer removeLatency,
      final OperationTimer writeLatency,
      final OperationTimer commitLatency,
      final Counter rollbackCount) {
...
RocksDbMetricsFactory also creates
gauge
        "rocks_db_table_readers_memory_bytes",
        "Estimated memory used for RocksDB index and filter blocks in bytes",
gauge
        "rocks_db_files_size_bytes",
        "Estimated database size in bytes",

And in BesuMetricCategory we explicitly exclude all the RocksDb metrics

    // Why not KVSTORE_ROCKSDB and KVSTORE_ROCKSDB_STATS, KVSTORE_PRIVATE_ROCKSDB_STATS,
    // KVSTORE_PRIVATE_ROCKSDB_STATS?  They hurt performance under load.

macfarla:
@fab-10 would it make sense to remove the timers altogether, or would it make more sense to move them to a separate category that can still be enabled when troubleshooting?

fab-10:
I think we should always have the data about the latency, that is exposed by the timers, but if the comment says that collecting stats has a non negligible side effect on performance, then I am afraid that these stats could not be useful at all, because they could modify the behaviour of Besu. cc: @ahamlat on which is a tollerable overhead for collecting metrics.

At this point I am thinking that a refactor could be required to reduce the overhead of these metrics collection, another option is to leverage the stats that RocksDB dump on its logs at regular intervals, that as last resort could be also delegated to an external process.

macfarla:
[...]

PR #4038, opened 2022-07-01
Enable rocksDb metrics

ahamlat:
Enable RocksDB metrics to evaluate the overhead. 

## Fixed Issue(s)

#3369
## Documentation

## Changelog


ahamlat:
Enabling RocksDB metrics on a Besu node generates some overhead, especially during the sync time. On AWS a t3.xlarge instance, Besu node with RocksDB metrics enabled is 3 hours slower for the total sync time. After sync, we didn’t notice a real difference with this type of AWS instance. It is possible that less powerful VMs may show different results. 
You can check all the metrics [here](https://docs.google.com/document/d/1xi0HQ5s6tiprWpTlsXUYyG-CDTPHzTWIX71SSfDG060/edit?usp=sharing).

Enabling RocksDB metrics doesn't need to modify the JAVA code, we can do it just by specifying two Besu parameters :
metrics-protocol and metrics-category. You can find below and example on how to enable RocksDB metrics.

```--metrics-protocol=PROMETHEUS --metrics-category=BLOCKCHAIN,ETHEREUM,EXECUTORS,NETWORK,PEERS,KVSTORE_ROCKSDB,KVSTORE_ROCKSDB_STATS,PRUNER,RPC,SYNCHRONIZER,TRANSACTION_POOL,STRATUM```

I recommend to enable RocksDB metrics only when these metrics are needed and take in account this overhead. I also recommend to not enable RocksDB metrics on a machine/VM with less than 4CPU or/and 16 GIB RAM.

PR #4041, opened 2022-07-01
fix for bonsai state root mismatch

matkt:

## PR description

The problem comes from concurrent access to Bonsai's storageToUpdate map. These concurrent accesses can lead to the overwriting of certain data or just missing in this map and impact the behavior when importing a block

## Documentation

## Changelog


shemnon (approved):
I like this fix, it matches what we've seen: occasionally incorrect storage reads.

Critically `warmedUpStorage` is a MultiMap that is specifically a [`HashMultiMap`](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/HashMultimap.java#L40) whose class documentation explicitly states 

> This class is not threadsafe when any concurrent operations update the multimap.

Missing values would be the most common failure mode, when a total system lockup doesn't happen (like when you do concurrent updates on an ordinary HashMap)

shemnon:
To add weight to this theory a [this transaction](https://etherscan.io/tx/0x7da2837185906754ed728062bffa356976f4ef89565e4f777c8fcc70d8359a84) has been observed failing and has interacted with (this contract as a internal call)[https://etherscan.io/address/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2] and there were access list entries [here](https://etherscan.io/tx/0xc6add6a711c5a3c7f38a6930a167a09946a95523d84bd8d866ad109926814d6f#accesslist), [here](https://etherscan.io/tx/0x07e5a6608e6282d9c3eed07eccf51391af10a96bb915411efcb8253f90ca3ed7#accesslist), and [here](https://etherscan.io/tx/0x824e947e38ab175468f0e2114deb15d7ac738f19391979e062b841bbeaa653c7#accesslist) in the same block.  All three are mev bots so it makes sense they would be sure the access list saves on gas.

matkt:
thanks for the review @shemnon 

I just compiled the result of the weekend with 5 nodes without the fix and 5 nodes with the fix
- Without the fix each nodes had state root mismatch (total during the weeking 58 state root mismatch)
- With the fix 0 state root mismatch during the weekend

PR #4055, opened 2022-07-05
fcU: Ignore newHead if it is an ancestor of chain head

### LINKED ISSUE #4051: fcU: Ignore if newHead references an ancestor of chain head
### Description
The spec says:
> Client software MAY skip an update of the forkchoice state and MUST NOT begin a payload build process if forkchoiceState.headBlockHash references an ancestor of the head of canonical chain. In the case of such an event, client software MUST return {payloadStatus: {status: VALID, latestValidHash: forkchoiceState.headBlockHash, validationError: null}, payloadId: null}.

Right now we are rolling back the chain if we receive an ancestor of our chain head. We need to check if the newHead is indeed an ancestor of chain head and ignore the update in such a case.

daniellehrner:
## PR description
A forkchoice update will be ignored if `newHead` is an ancestor of chain head.

## Fixed Issue(s)
fixes #4051

## Documentation

## Changelog

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
@@ -289,6 +289,12 @@ public class MergeCoordinator implements MergeMiningCoordinator {
     MutableBlockchain blockchain = protocolContext.getBlockchain();
     final Optional<BlockHeader> newFinalized = blockchain.getBlockHeader(finalizedBlockHash);
 
+    if (newHead.getNumber() < blockchain.getChainHeadBlockNumber()
+        && isDescendantOf(newHead, blockchain.getChainHeadHeader())) {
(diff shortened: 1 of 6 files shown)


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -434,7 +440,7 @@ boolean ancestorIsValidTerminalProofOfWork(final BlockHeader blockheader) {
 
     boolean resp =
         parent.filter(header -> isTerminalProofOfWorkBlock(header, protocolContext)).isPresent();
-    LOG.warn(
+    LOG.debug(
```
daniellehrner, line 443: I've changed this one to debug, because it does not look like anything that should case a warning in the log

garyschulte (approved):
tbh I have misgivings about this behavior, but the hive suite seems to have made this optional behavior mandatory.

PR #4056, opened 2022-07-05
Return the correct latest valid hash in case of bad block

### LINKED ISSUE #3893: Hive: Invalid Ancestor Chain Re-Org tests not passing
### Description
Test **Invalid Ancestor Chain Re-Org, Invalid StateRoot, Invalid P1', Reveal using newPayload** is failing with this error:
```
FAIL (Invalid Ancestor Chain Re-Org, Invalid StateRoot, Invalid P1', Reveal using newPayload): Unexpected LatestValidHash on EngineNewPayloadV1: 0x0000000000000000000000000000000000000000000000000000000000000000, expected=0x29b6378be5c532eb43559a8f85a0760f40d50c99a9ebba7a8f1d7e5fe4c5bfaa
```

fab-10:

## PR description

In case a bad block is passed to `newPayload` or `forkchoiceUpdate` then response must contain the hash of the lastest valid block on the chain of the bad block, while now we just return the zero hash.

## Fixed Issue(s)

fixes #3893 

## Documentation

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
```diff
@@ -85,22 +85,17 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+              mergeCoordinator
+                  .getLatestValidAncestor(forkChoice.getHeadBlockHash())
+                  .orElse(Hash.ZERO),
               null,
               Optional.of(forkChoice.getHeadBlockHash() + " is an invalid block")));
     }
 
     Optional<BlockHeader> newHead =
-        protocolContext.getBlockchain().getBlockHeader(forkChoice.getHeadBlockHash());
+        mergeCoordinator.getOrSyncHeaderByHash(forkChoice.getHeadBlockHash());
```
garyschulte, line 96: :+1:

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
```diff
@@ -160,8 +164,7 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
     final var block =
         new Block(newBlockHeader, new BlockBody(transactions, Collections.emptyList()));
 
-    if (mergeContext.isSyncing()
-        || mergeCoordinator.getOrSyncHeaderByHash(newBlockHeader.getParentHash()).isEmpty()) {
+    if (mergeContext.isSyncing() || parentHeader.isEmpty()) {
```
garyschulte, line 167: This has a side effect of *not* appending the parent header to a backward sync if it is not in the blockchain already, since `parentHeader` is just an optional fetched from the blockchain as opposed to trying to fetch and appends the hash to backward sync if it was not found.  I think we need to keep that behavior  - perhaps we change the lookup on line 152 to use `mergeCoordinator.getOrSyncByHash`?

fab-10:
@garyschulte @gezero actually contrary what I first thought it is not necessary to modify backward sync for this fix, reverted the changes, so this PR is now smaller and the removal of the finalized check could be done in another step

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
garyschulte: as discussed, I am 👌  to rely on the append of the newPayload hash

garyschulte (approved):
👍

PR #4058, opened 2022-07-05
3943 stop blocks on finalized

### LINKED ISSUE #3943: Stop the block synchronization after the first finalized block
### Description
The Merge spec says that an execution should behave like this after the switch to PoS:

> Beginning with receiving the finalized block next to the FIRST_FINALIZED_BLOCK, the networking stack MUST remove the handlers corresponding to the following messages:
> 
> NewBlockHashes (0x01)
> NewBlock (0x07)
> Peers that keep sending these messages after the handlers have been removed SHOULD be disconnected.
> 

currently Besu stop any block synchronization task when the TTD is reached, so it is not following the spec, and doing so is not able to easily recognize if a peer is still on PoW after TTD has been reached, and disconnect those peers.

### Acceptance Criteria

Keep processing and sending 
* NewBlockHashes (0x01)
* NewBlock (0x07)
messages until the first finalized block is received from the CL

* Software version: besu-22.4.3-SNAPSHOT

jflo:
fixed #3943


besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
```diff
@@ -145,13 +151,50 @@ protected PluginServiceFactory createAdditionalPluginServices(
...
+
+    ConsensusContext cc = protocolContext.getConsensusContext(ConsensusContext.class);
+    if (cc instanceof MergeContext) {
+      protocolContext
+          .getConsensusContext(MergeContext.class)
+          .observeNewIsPostMergeState(ethProtocolManager);
+      protocolContext
+          .getConsensusContext(MergeContext.class)
+          .addNewForkchoiceMessageListener(ethProtocolManager);
+    }
```
garyschulte, line 186: This might be better to have in the MergeBesuControllerBuilder since the intent is for all of the Transition* classes to be dropped after the paris fork.  the Transition builder composes the pre-merge and post merge builders, so it should be functionally equivalent to have this in the merge controller builder.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PostMergeContext.java
```diff
@@ -99,7 +99,9 @@ public void setIsPostMerge(final Difficulty totalDifficulty) {
 
     if (oldState.isEmpty() || oldState.get() != newState) {
       newMergeStateCallbackSubscribers.forEach(
-          newMergeStateCallback -> newMergeStateCallback.onNewIsPostMergeState(newState));
+          newMergeStateCallback ->
+              newMergeStateCallback.onCrossingMergeBoundary(
```
garyschulte, line 103: this is a clarifying name change at transition, but post-merge this event fires when the client starts up even if we are long past TTD.  I agree NewIsPostMergeState is not a great name, but CrossingMergeBoundary is a bit misleading also considering the startup behavior.

ethereum/eth/src/main/java/org/hyperledger/besu/consensus/merge/NewForkchoiceMessageListener.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+package org.hyperledger.besu.consensus.merge;
+
+import org.hyperledger.besu.datatypes.Hash;
+
+import java.util.Optional;
+
+public interface NewForkchoiceMessageListener {
```
garyschulte, line 23: I hate having a lot of naming feedback, but wouldn't `ForkchoiceMessageListener` work here?

jflo: yeah, i actually used that in a previous version.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -73,6 +78,10 @@ public class EthProtocolManager implements ProtocolManager, MinedBlockObserver {
   private final List<PeerValidator> peerValidators;
   // The max size of messages (in bytes)
   private final int maxMessageSize;
+  private Optional<Difficulty> powTerminalDifficulty;
+  private final StampedLock powTerminalDifficultyLock = new StampedLock();
+  private Hash lastFinalized = Hash.ZERO;
+  private final AtomicLong numFinalizedSeen = new AtomicLong(0);
```
garyschulte, line 84: I think we are leaking too much of the merge implementation details into protocol manager.  

How about either:
* extracting and encapsulate the merge specific behavior into an Optional "PeerFilterAction" 
* or subclassing a merge-specific EthProtocolManager

besu/src/main/java/org/hyperledger/besu/controller/MergeBesuControllerBuilder.java
```diff
@@ -73,13 +78,49 @@ protected MiningCoordinator createMiningCoordinator(
...
+      protocolContext
+          .getConsensusContext(MergeContext.class)
+          .observeNewIsPostMergeState(ethProtocolManager);
+      protocolContext
+          .getConsensusContext(MergeContext.class)
+          .addNewForkchoiceMessageListener(ethProtocolManager);
+    }
+
+    return ethProtocolManager;
+  }
```
garyschulte, line 116: if we passed in an Optional<PeerFilterAction> or however we encapsulate the merge-specific logic, we wouldn't need to do this kind of introspection on class type.

garyschulte (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -409,4 +439,35 @@ public List<Bytes> getForkIdAsBytesList() {
         ? Collections.emptyList()
         : chainHeadForkId.getForkIdAsBytesList();
   }
+
+  private boolean isFinalized() {
+    return this.numFinalizedSeen.get() > 1;
```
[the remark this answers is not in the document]
jflo: Specs say not to disconnect till the 2nd finalized epoch.

garyschulte (approved):
clean refactor.  one clarification and/or simplification question, non-blocking

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/MergePeerFilter.java
```diff
@@ -0,0 +1,102 @@
...
+      LOG.debug("disconnecting peer for sending new blocks after transition to PoS");
+      peer.disconnect(DisconnectReason.SUBPROTOCOL_TRIGGERED);
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  private boolean isFinalized() {
+    return this.numFinalizedSeen.get() > 1;
```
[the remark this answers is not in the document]
jflo: Spec describes it as waiting till the 2nd finalized block.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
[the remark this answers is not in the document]
macfarla: got it

PR #4066, opened 2022-07-06
After merge add a rule to check that the current block is more recent than its parent

fab-10:

## PR description

In PoS current block timestamp must be greather than its parent block timestamp, so adding a new header validation rule to check that.
There is also a fix in the `shouldUsePostMergeRules` that was not correctly checking for TD >= TTD.

With this the Hive test *Invalid Ancestor Chain Re-Org, Invalid Timestamp, Invalid P9', Reveal using sync* passes.

## Documentation

## Changelog


macfarla (approved):
LGTM

PR #4072, opened 2022-07-08
Loads uncached logs on cache miss.

### LINKED ISSUE #3921: Besu failing to index event log
### Description
On one of the EF teku-besu kiln nodes, the besu instance is failing to return one of the deposit event logs. Since the CL requires all deposit events in order Teku is complaining about this quite a lot.

This is on box 164.92.203.174

JSON RPC request is:
```
{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"topics":["0x649bbc62d0e31342afea4e5cd82d4049e7e1ee912fc0889aa790803be39038c5"],"fromBlock":"0x8547d","toBlock":"0x8547d","address":["0x4242424242424242424242424242424242424242"]}],"id":0}
```

The broken besu instance returns:
```
{
  "jsonrpc" : "2.0",
  "id" : 0,
  "result" : [ ]
}
```

But another teku-besu instance on kiln returns the deposit event correctly:
```
{
  "jsonrpc" : "2.0",
  "id" : 0,
  "result" : [ {
    "logIndex" : "0x0",
    "removed" : false,
    "blockNumber" : "0x8547d",
    "blockHash" : "0xa62e2b2f55447ecbf5b603fa032746222da6039f27851175a3214812029ebc70",
    "transactionHash" : "0x589c0803aef2b3d0ecd48888bd0f9d799fd2421860c46bf91e2e21f0eb82414d",
    "transactionIndex" : "0x3c",
    "address" : "0x4242424242424242424242424242424242424242",
    "data" : "0x00000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000140000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000030933f02fc10608e5b41513035b9f8f08306a0c54fcb2b04a673881987e6eb089b983481f5d888d4f6e08ebffbb41fcc4f000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020005db87e87315c6545e51a5daff02eab6df948cc8c9a49d6d7d03b80d1f18096000000000000000000000000000000000000000000000000000000000000000800405973070000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060aa19ef95787fe491388337f47fdb67936ae0aeb437766b5cc44402c8ecd08d51fe7cca7643d85fca8ec0a1c16f5102210e4b41bb6ef2f84964bdfeb733121d046e0d5cd7eb85ce8e399d0d4e61cc0cb979f58717e925e0c06a734f0311a772a00000000000000000000000000000000000000000000000000000000000000008442d000000000000000000000000000000000000000000000000000000000000",
    "topics" : [ "0x649bbc62d0e31342afea4e5cd82d4049e7e1ee912fc0889aa790803be39038c5" ]
  } ]
}
```

Somehow besu has failed to index this deposit event.

jflo:
## PR description

On a miss of the on-disk logs bloom cache, we check the uncached storage, and return that if available.

## Fixed Issue(s)
fixes #3921 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Bug Fixes
 - Fixed a trie log layer issue on bonsai during reorg [#4069](https://github.com/hyperledger/besu/pull/4069)
+- Fixed an issue getting logs when blooms cache misses [#3921](https://github.com/hyperledger/besu/issues/3921)
 
 ## 22.7.0-RC1
 
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
@@ -666,13 +666,35 @@ public class BlockchainQueries {
                 .map(
                     cacheFile -> {
                       try {
-                        return matchingLogsCached(
-                            thisSegment * BLOCKS_PER_BLOOM_CACHE,
-                            thisStep % BLOCKS_PER_BLOOM_CACHE,
-                            Math.min(toBlockNumber, nextStep - 1) % BLOCKS_PER_BLOOM_CACHE,
-                            query,
-                            cacheFile,
-                            isQueryAlive);
+                        List<LogWithMetadata> maybeLogs =
+                            matchingLogsCached(
+                                thisSegment * BLOCKS_PER_BLOOM_CACHE,
+                                thisStep % BLOCKS_PER_BLOOM_CACHE,
+                                Math.min(toBlockNumber, nextStep - 1) % BLOCKS_PER_BLOOM_CACHE,
+                                query,
+                                cacheFile,
+                                isQueryAlive);
+                        if (maybeLogs.isEmpty()) {
+                          LOG.debug(
+                              "cache miss for block range from {} to ",
+                              fromBlockNumber,
+                              toBlockNumber);
+                          maybeLogs =
+                              matchingLogsUncached(
+                                  thisStep,
+                                  Math.min(toBlockNumber, Math.min(toBlockNumber, nextStep - 1)),
+                                  query,
+                                  isQueryAlive);
+                          if (maybeLogs.isEmpty()) {
+                            LOG.debug(
+                                "uncached miss for block range from {} to ",
+                                fromBlockNumber,
+                                toBlockNumber);
+                          }
+                          return maybeLogs;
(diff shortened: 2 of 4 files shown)


ajsutton:
Just checking - a cache "miss" here just means the cache didn't exist for that block right? If this change means that besu will go through individual blocks if the cache returns 0 hits then CLs are going to absolutely destroy besu when they're catching up with the deposits initially.  In that case we can be querying many thousands of blocks at once (walking through most of the chain history) to search for logs, and the majority of those batches will not contain any deposit events.

The event cache was initially added to fix this performance problem because besu would just timeout constantly.

matkt:
Can you reproduce this problem quite easily? @ajsutton  If  yes I would prefer to try to find why the blocks are missing. I think the latest fixes regarding bonsai may fix the problem. But if we can reproduce with more logs  WE will be know the reason

ajsutton:
> Can you reproduce this problem quite easily? @ajsutton If yes I would prefer to try to find why the blocks are missing. I think the latest fixes regarding bonsai may fix the problem. But if we can reproduce with more logs we will know the reason

I'm not sure I'd say "reproduce" but we get this problem on one of our besu merge nodes every day or two at least. I'll get them upgraded to the recent release with bonsai fixes and see if it stops happening.

matkt:
> > Can you reproduce this problem quite easily? @ajsutton If yes I would prefer to try to find why the blocks are missing. I think the latest fixes regarding bonsai may fix the problem. But if we can reproduce with more logs we will know the reason
> 
> I'm not sure I'd say "reproduce" but we get this problem on one of our besu merge nodes every day or two at least. I'll get them upgraded to the recent release with bonsai fixes and see if it stops happening.

it's good news that it happens so often to debug 👌

jflo:
> Just checking - a cache "miss" here just means the cache didn't exist for that block right? If this change means that besu will go through individual blocks if the cache returns 0 hits then CLs are going to absolutely destroy besu when they're catching up with the deposits initially. In that case we can be querying many thousands of blocks at once (walking through most of the chain history) to search for logs, and the majority of those batches will not contain any deposit events.
> 
> The event cache was initially added to fix this performance problem because besu would just timeout constantly.

You are correct ser, this is a hack and overkill. Closing this PR to put better thought into it.

PR #4086, opened 2022-07-12
Enclave public key length constraint removed for private transactions

### LINKED ISSUE #3819: Remove 44-byte long constraint for --privacy-public-key-file so Besu could support Tessera's EC encryptor
### Description
As a user, I want to remove the 44-byte long constraint for `--privacy-public-key-file` so that I can use [Tessera's EC encryptor configuration](https://docs.tessera.consensys.net/en/stable/HowTo/Configure/Cryptographic-elliptic-curves/#configure-an-alternative-cryptographic-elliptic-curve). 

Tessera currently supports the configuration of uncompressed SECP256r1 public keys in base64 (like `MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE+NYbGgkOxL7LWJh/68eCsHMOvxydvNO4gck/0KFsoCtkyCS3O7TuYfWwk9tr/2WKAsd3/v1a+XGffwnahB1PEw==`). This value is not currently supported by Besu for the attribute `--privacy-public-key-file` since it expects a value that is 44 chars long.

The purpose of this ticket is to get rid of that constraint since potentially Tessera could be configured with any JCA implementation and adapt the acceptance tests for privacy so we can use alternativa encryptors to `NaCl` (this time would be `EC`).

About the **acceptance tests**, we would need to adapt them somehow since web3j's `Base64String` [has the same problem](https://github.com/web3j/web3j/blob/master/eea/src/main/java/org/web3j/utils/Base64String.java#L85). Some of the possible options would be:

1. Create another implementation for privacy of the `org.web3j.tx.TransactionManager` interface within the `acceptance-tests` package in Besu so we do not use `Base64String`. The problem with this option is that we will encounter the same issue when trying to submit a private transaction from a "blockchain service" to Besu. It would solve the problem within the acceptance tests though.

2. Modify `org.web3j.tx.PrivateTransactionManager` so it uses something different to `Base64String` in the constructor. This is not an ideal approach IMO since I do not really know who is already using `PrivateTransactionManager` outside of Besu (within Besu though it looks like it is just used for the acceptance tests).

3. (**MY PREFERRED APPROACH**) Modify `Base64String` so it does not check that length of 44, similarly to what is suggested in this issue for Besu (the logic is the same, sometimes the enclave public key in base64 is longer than 44). All the current methods will behave the same way regarding arguments and method outcomes. [I have suggested this change already in Discord](https://discord.com/channels/779382027614158919/905428373218689054/973578966550052894) following web3j's contribution guidelines. [This is my branch with the change applied](https://github.com/freemanzMrojo/web3j/tree/privacy-pubkey-length-constraint-removed). I have [exported the package into my Besu branch](https://github.com/freemanzMrojo/besu/blob/privacy-pubkey-length-constraint-removed/acceptance-tests/dsl/build.gradle#L47) to test it with good results overall in the acceptance tests locally (still need to check some errors).

### Acceptance Criteria
* It is possible to provide uncompressed SECP256r1 base64 public key as the content for the file passed as `--privacy-public-key-file` attribute.
* The acceptance tests are adapted so they accept Tessera's EC encryptor configuration.

freemanzMrojo:
## PR description

Removes the length constraint of 44 characters for the base64 version of an enclave public key for private transactions via Tessera. More details in the linked issue.

Adapts the acceptance tests so this change could be validated against the target Tessera configuration (EC encryptor).

## Fixed Issue(s)

fixes #3819


gradle/versions.gradle
```diff
@@ -190,6 +190,7 @@ dependencyManagement {
     dependency 'org.web3j:besu:4.8.9'
     dependency 'org.web3j:core:4.8.9'
     dependency 'org.web3j:crypto:4.8.9'
+    dependency 'org.web3j:eea:4.9.3'
     dependency 'org.web3j:quorum:4.8.4'
```
antonydenyer, line 195: Ideally we should aim to keep these consistent as possible. 

```
dependency 'org.web3j:besu:4.9.3'
dependency 'org.web3j:core:4.9.3'
dependency 'org.web3j:crypto:4.9.3'
dependency 'org.web3j:eea:4.9.3'
dependency 'org.web3j:quorum:4.9.0'
```

antonydenyer (changes requested):

gradle/versions.gradle
shemnon: `org.web3j:quorum:4.9.3` does not exist, which is why 4.9.0 is used.

freemanzMrojo: I agree completely with you guys, however there is an issue in web3j for versions >= 4.9.0 (https://github.com/web3j/web3j/issues/1747). I have put a comment in the PR description about why 4.9.3 is just for that package.

Happy to upgrade to a potential 4.9.4 (and 4.9.0 for `quorum` ;) ) once the bugfix for https://github.com/web3j/web3j/issues/1747 is implemented. Not sure if it will happen soon enough though.

shemnon:
I have zero exposure to the enclave code.  Perhaps @macfarla @pinges could give a review?

acceptance-tests/tests/build.gradle
```diff
@@ -12,10 +12,19 @@
...
 
+configurations.all {
+  resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+    if (details.requested.group == 'org.web3j' && details.requested.version == '4.9.2') {
+      details.useVersion '4.9.4'
+      details.because 'web3-gradle-plugin has not been updated yet'
+    }
+  }
+}
+
```
antonydenyer, line 27: Should be solved in 

https://github.com/hyperledger/besu/pull/3752 

don't mind which approach is taken.

freemanzMrojo: Me neither, maybe it makes more sense to merge that other PR first since it covers specifically the web3j upgrade, but I do not mind if the approach currently implemented in my branch is finally taken.

cc: @diega

diega: I like this one! @freemanzMrojo would you push it into #3752? I think it's better to keep the PRs/squashes split if possible

freemanzMrojo: I can do that @diega , thanks!

freemanzMrojo: Will include @shemnon wording suggestions as well @diega

freemanzMrojo: Could you give me permission to your `besu` fork @diega  so I can push my commit? Thanks in advance!

diega: sure! check now

freemanzMrojo: Just pushed the commit, thanks!

antonydenyer (approved):

PR #4097, opened 2022-07-13
Finalised blocks should not prevent reorgs in BWS

gezero:
We should allow reorgs bellow finalized blocks during Backwards Sync.


gezero:
A bug in the logic in the check prevented BWSync to finish even in some valid scenarios. I think it is better to remove it altogether.

garyschulte (approved):
LGTM.

We don't have any limit on backward sync now right?  It can reorg a chain past head, past finalized, etc.

PR #4098, opened 2022-07-13
Delegate all the block checks and validation to the block import phase

fab-10:

## PR description

Remove preventive checks during sync, since they will be performed always performed during the import phase later, where the block will be also flagged as bad block.
With this the last failing Hive test (Invalid Ancestor Chain Re-Org, Invalid Number, Invalid P9', Reveal using sync) passes.

## Documentation

## Changelog


garyschulte (approved):
:+1:

PR #4102, opened 2022-07-14
Ignore 2 tests that assume that the system language is English, if that should not be the case

### LINKED ISSUE #2190: Unit tests fail if system language is not English
### Description
The tests `BesuCommandTest.privEnclaveKeyFileDoesNotExist` and `NetworkingServiceLifecycleTest.startDiscoveryPortInUse` fail if the system language of the operating system is another language than English. In `BesuCommandTest.privEnclaveKeyFileDoesNotExist` the test looks like this:

```
@Test
public void privEnclaveKeyFileDoesNotExist() {
  parseCommand("--privacy-enabled=true", "--privacy-public-key-file", "/non/existent/file");

  assertThat(commandOutput.toString()).isEmpty();
  assertThat(commandErrorOutput.toString()).startsWith("Problem with privacy-public-key-file");
  assertThat(commandErrorOutput.toString()).contains("No such file");
}
```

The last assertion is the one that is failing. `No such file` is the error message coming directly from `IOException` when the file is not found. But the error messages of the exceptions thrown by the Java library are translated according to the system language. For example when the system language is set to Spanish the error looks like this:

```
Problem with privacy-public-key-file: /non/existent/file (No existe el archivo o el directorio)
```

In this case `No such file` is not part of the output string and the assertion fails.

 `NetworkingServiceLifecycleTest.startDiscoveryPortInUse`  fails for the same reason:
 
 ```
catch (final Exception e) {
          assertThat(e).hasCauseExactlyInstanceOf(PeerDiscoveryServiceException.class);
          assertThat(e)
              .hasMessageStartingWith(
                  "org.hyperledger.besu.ethereum.p2p.discovery."
                      + "PeerDiscoveryServiceException: Failed to bind Ethereum UDP discovery listener to 0.0.0.0:");

          assumeThat(System.getProperty("user.language")).isEqualTo("en");
          assertThat(e).hasMessageContaining("Address already in use");
        } 
```

Here again the last assertion fails, because it comes directly from an exception thrown by the Java library.

### Expected behavior
The tests should either be successful or be skipped.

### Possible solution
The tests could be skipped using [AssertJ Assume](https://assertj.github.io/doc/#assertj-core-assumptions) and testing if the system language is English or not. Like that they would not fail for developers with another system language. But they would still be executed in the CI / CD pipeline, to verify that the tested behavior has not been changed.

### Steps to Reproduce (Bug)
1. Checkout `master` branch
2. Change system language to any other than English
3. Execute `./gradlew build`

daniellehrner:

## PR description
2 of our tests that check error messages from the JVM assume that the system language is always English. If that is not the case, the JVM error messages will be localized and the tests will fail. This PR checks the system language before running those 2 tests and ignores them, if it is not English. If the system language is English the tests are executed like before.

## Fixed Issue(s)

fixes #2190

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Add a PoS block header rule to check that the current block is more recent than its parent [#4066](https://github.com/hyperledger/besu/pull/4066)
 - Fixed a trie log layer issue on bonsai during reorg [#4069](https://github.com/hyperledger/besu/pull/4069)
 - Fix transition protocol schedule to return the pre Merge schedule when reorg pre TTD [#4078](https://github.com/hyperledger/besu/pull/4078)
+- The build process runs successfully even though the system language is not English  [#4102](https://github.com/hyperledger/besu/pull/4102)
 
 ## 22.7.0-RC1
 
--- a/besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
+++ b/besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
@@ -17,6 +17,7 @@ package org.hyperledger.besu.cli;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.Matchers.startsWith;
 import static org.hyperledger.besu.cli.config.NetworkName.CLASSIC;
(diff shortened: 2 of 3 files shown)


fab-10:
instead of ignoring, could be an option to force the tests to run with the english locale?

fab-10 (approved):
LGTM, please check my comment on setting the locale for those test if possible

PR #4118, opened 2022-07-16
EIP-1153: Transient storage opcodes

codyborn:
## PR description

- Implement [EIP-1153: Transient storage opcodes](https://eips.ethereum.org/EIPS/eip-1153)
- Create new EIP1153 spec for including this feature post Paris.
- Added a new `ByteCodeBuilder` class to make it easier to construct the E2E test cases.

### TStoreEVMOperationTest.java
- [x]  load arbitrary value is 0 at beginning of transaction: `TLOAD(x)` is `0`
- [x]  loading after storing returns the stored value: `TSTORE(x, y)`, `TLOAD(x)` returns `y`
- [x]  loading any other slot after storing to a slot returns 0: `TSTORE(x, y)`, `TLOAD(x + n)` where `n > 0` returns `0`
- [x]  contracts have separate transient storage—loading a slot in a separate contract after storing returns 0: `TSTORE(x, y)`, `CALL(z, ...)`, `TLOAD(x)` returns `0`
- [x]  reentrant calls access the same transient storage: `TSTORE(x, y)`, `CALL(self, ...)`, `TLOAD(x)` returns `y`
- [x]  reentrant calls can manipulate the same transient storage: `TSTORE(x, y)`, `CALL(self, ...)`, `TSTORE(x, z)`, `TLOAD(x)` returns `z`
- [x]  successfully returned calls do not revert transient storage writes: `TSTORE(x, y)`, `CALL(self, ...)`, `TSTORE(x, z)`, `RETURN`, `TLOAD(x)` returns `z`
- [x]  revert undoes the transient storage write from the failed call: `TSTORE(x, y)`, `CALL(self, ...)`, `TSTORE(x, z)`, `REVERT`, `TLOAD(x)` returns `y`
- [x]  revert undoes all the transient storage writes to the same key from the failed call, i.e. applies the modifications in reverse order: `TSTORE(x, y)`, `CALL(self, ...)`, `TSTORE(x, z)`, `TSTORE(x, z + 1)` `REVERT`, `TLOAD(x)` returns `y`
- [x]  revert undoes transient storage writes from inner calls that successfully returned: `TSTORE(x, y)`, `CALL(self, ...)`, `CALL(self, ...)`, `TSTORE(x, y + 1)`, `RETURN`, `REVERT`, `TLOAD(x)` returns `y`
- [x]  delegatecall manipulates transient storage in the context of the current address: `TSTORE(x, y)`, `DELEGATECALL(a, ...)`, `TSTORE(x, z)`, `RETURN`, `TLOAD(x)` returns `z`
- [x]  delegatecall reads transient storage in the context of the current address: `TSTORE(x, y)`, `DELEGATECALL(a, ...)`, `TLOAD(x)` returns `y`
- [x]  transient storage cannot be manipulated in a static context: `TSTORE(x, y)`, `STATICCALL(a, ...)`, `TSTORE(x, z)` reverts
- [x]  transient storage cannot be manipulated in a static context when calling self: `TSTORE(x, y)`, `STATICCALL(self, ...)`, `TSTORE(x, z)` reverts
- [x]  transient storage cannot be manipulated in a nested static context: `TSTORE(x, y)`, `STATICCALL(self, ...)`, `CALL(self, ...)`, `TSTORE(x, z)` reverts
- [x]  Transient storage can be accessed in a static context when calling self`TSTORE(x, y)`, `STATICCALL(self, ...)`, `CALL(self, ...)`, `TLOAD(x)` returns `y`
- [x] transient storage does not persist beyond a single transaction: TSTORE(x, y), RETURN, (new transaction to same contract), TLOAD(x, y) returns 0 

### TStoreOperationTest.java

- [x] Insufficient gas
- [x] Simple tstore test
- [x] TLoad uninitialized
- [x] TStore/TLoad 
- [x] TStore->TStore->TLoad
- [x] No gas refund

## Documentation

## Changelog

Unclear if I should add a new record to the changelog. Please advise.


siladu:
Thanks for the contribution @codyborn!

Please can you point me to where EIP-1153 is confirmed for Shanghai?
I can't see it in this list: https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md#eips-considered-for-inclusion

Also, you'll need to sign off your commits for the DCO check to pass, see https://wiki.hyperledger.org/display/BESU/DCO

codyborn:
Thanks @siladu! It's not yet confirmed for any hardfork; I just put it in the Merge++ hardfork as a placeholder. Do you have a recommended practice for features like this that aren't yet scheduled?

For the DCO, do I need to recreate the PR entirely with new commits or is there a way to roll all these up into a new commit?

siladu:
Hi @codyborn,

> Do you have a recommended practice for features like this that aren't yet scheduled?

You could put it behind an experimental feature toggle. The toggle code has been removed but something similar was done for EIP-1559, see https://github.com/hyperledger/besu/pull/641

> For the DCO, do I need to recreate the PR entirely with new commits or is there a way to roll all these up into a new commit?

Shouldn't need to recreate the PR. You can either amend your existing commits with the sign-off and force push; or squash, sign-off and force push.

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/ShanghaiGasCalculator.java
```diff
@@ -0,0 +1,34 @@
...
+  public ShanghaiGasCalculator() {}
+
+  // EIP-1153
+  @Override
+  public long getTloadOperationGasCost() {
+    return TLOAD_GAS;
+  }
+  
+  @Override
+  public long getTstoreOperationGasCost() {
```
atoulme, line 31: T-> Transient

evm/src/main/java/org/hyperledger/besu/evm/operation/TLoadOperation.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright ConsenSys AG.
```
atoulme, line 2: It should be Copyright Besu Contributors

evm/src/main/java/org/hyperledger/besu/evm/operation/TStoreOperation.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.OptionalLong;
+
+import org.apache.tuweni.units.bigints.UInt256;
+
+public class TStoreOperation extends AbstractOperation {
+
+  public static final long FRONTIER_MINIMUM = 0L;
+  public static final long EIP_1706_MINIMUM = 2300L;
+
+  protected static final OperationResult ILLEGAL_STATE_CHANGE =
```
atoulme, line 33: Why is this protected?

evm/src/main/java/org/hyperledger/besu/evm/operation/TStoreOperation.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.Optional;
+import java.util.OptionalLong;
+
+import org.apache.tuweni.units.bigints.UInt256;
+
+public class TStoreOperation extends AbstractOperation {
+
+  public static final long FRONTIER_MINIMUM = 0L;
+  public static final long EIP_1706_MINIMUM = 2300L;
```
atoulme, line 31: Can you make those private?

evm/src/main/java/org/hyperledger/besu/evm/operation/TLoadOperation.java
```diff
@@ -0,0 +1,59 @@
...
+import org.hyperledger.besu.evm.internal.FixedStack.OverflowException;
+import org.hyperledger.besu.evm.internal.FixedStack.UnderflowException;
+
+import java.util.Optional;
+import java.util.OptionalLong;
+
+import org.apache.tuweni.bytes.Bytes32;
+import org.apache.tuweni.units.bigints.UInt256;
+
+public class TLoadOperation extends AbstractOperation {
```
atoulme, line 32: TransientLoadOperation

evm/src/main/java/org/hyperledger/besu/evm/worldstate/UpdateTrackingAccount.java
```diff
@@ -74,6 +75,7 @@
 
     this.updatedCode = Bytes.EMPTY;
     this.updatedStorage = new TreeMap<>();
+    this.updatedTransientStorage = new TreeMap<>();
```
atoulme, line 78: Why use a TreeMap here?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/ShanghaiGasCalculator.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright ConsenSys AG.
```
atoulme, line 2: copyright, same

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleAccount.java
```diff
@@ -47,6 +47,7 @@ public class SimpleAccount implements EvmAccount, MutableAccount {
   private Supplier<Hash> codeHash =
       Suppliers.memoize(() -> code == null ? Hash.EMPTY : Hash.hash(code));
   private final Map<UInt256, UInt256> storage = new HashMap<>();
+  private final Map<UInt256, UInt256> transientStorage = new HashMap<>();
```
atoulme, line 50: It's a HashMap here - why is it a TreeMap in the other impl?

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
```diff
@@ -0,0 +1,174 @@
+package org.hyperledger.besu.ethereum.core;
```
atoulme, line 1: needs a copyright header

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
```diff
@@ -0,0 +1,174 @@
...
+
+import kotlin.UInt;
```
atoulme, line 3: using a Kotlin class?

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
```diff
@@ -0,0 +1,174 @@
...
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.units.bigints.UInt256;
+import org.hyperledger.besu.datatypes.Address;
+
+public class ByteCodeBuilder {
```
atoulme, line 8: Where is this class used?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.OptionalLong;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hyperledger.besu.ethereum.core.InMemoryKeyValueStorageProvider.createInMemoryWorldStateArchive;
+import static org.hyperledger.besu.evm.frame.ExceptionalHaltReason.INSUFFICIENT_GAS;
+import static org.mockito.Mockito.mock;
+
+public class TStoreOperationTest {
```
atoulme, line 50: TransientStoreOperationTest

atoulme (commented):
Good code contrib. A few nits and things to clean up. :+1:

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
codyborn: Good catch! It was auto-imported by mistake.

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
codyborn: It's used in the `TStoreEVMOperationTest` tests to construct the contract bytecode for the EVM-level tests. This makes it more self-documenting and easier to maintain than having long strings of bytecode in the test cases.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
codyborn: I chose `TStore...` to be consistent with the other file names (ex. `SStoreOperationTest`). Since `TSTORE` is the name of new operation I think it makes sense to keep it as is. Let me know if you agree!

evm/src/main/java/org/hyperledger/besu/evm/operation/TStoreOperation.java
codyborn: Good question. This was taken from `SStoreOperation.java`. I've kept it as protected but moved this definition to `AbstractOperation` so it can be used by both `SStoreOperation.java` and `TStoreOperation.java`.

evm/src/main/java/org/hyperledger/besu/evm/worldstate/UpdateTrackingAccount.java
codyborn: I was simply basing it off of regular storage, but I think you're right that it's not necessary to use a `TreeMap` since its keys are not iterated through in a particular order.

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleAccount.java
codyborn: <Responded to the other comment>

evm/src/main/java/org/hyperledger/besu/evm/operation/TLoadOperation.java
codyborn: See similar comment on `TStoreOperationTest.java`

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
```diff
@@ -188,4 +189,32 @@ public void noGasRefundFromTransientState() {
...
+    for (int i = 0; i < numOfOps; i++) {
+      frame.pushStackItem(UInt256.ONE);
+      frame.pushStackItem(UInt256.fromHexString("0x01"));
+      tstore.execute(frame, null);
+      frame.pushStackItem(UInt256.fromHexString("0x01"));
+      tload.execute(frame, null);
+      frame.popStackItem();
+    }
+    long endTime = System.nanoTime();
+    assertThat((endTime - startTime) / 1e6).isLessThan(5000);
```
moodysalem, line 218: what's the actual result here? iiuc, it's just doing store 1 in 1, load 1, pop, which should be executable in less than 1s ideally

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
atoulme: Please add some javadoc.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
atoulme: OK, I can live with that. Thanks for explaining.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
atoulme: In general not a good idea to test time in this type of test. Our CI is running on busy hardware. We have benchmarks elsewhere.

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
```diff
@@ -0,0 +1,199 @@
...
+      this.value = value;
+    }
+
+    @Override
+    public String toString() {
+      return value;
+    }
+  }
+
+  StringBuilder byteCode = new StringBuilder("0x");
```
atoulme, line 55: You might want to use a MutableBytes object instead of a String. You can add to it and have it output to String at the end.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardChain.java
```diff
@@ -179,7 +179,7 @@ public synchronized void addNewHash(final Hash newBlockHash) {
   }
 
   public synchronized Optional<Hash> getFirstHashToAppend() {
-    return Optional.ofNullable(hashesToAppend.peek());
+    return Optional.ofNullable(hashesToAppend.poll());
```
atoulme, line 182: Was this change added by mistake?

codyborn: Yes, thanks for catching it. It was accidentally included from a merge upstream.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
codyborn: > what's the actual result here? iiuc, it's just doing store 1 in 1, load 1, pop, which should be executable in less than 1s ideally

It's performing this `numOfOps` times (134,433 based on a `blockgaslimit` of 30M) and ensuring it runs less than 5 secs. It's highly hardware dependent which is why I marked it with `@Ignore` to be ran manually. On my laptop, it runs in < 300 ms.

@atoulme I see precompile benchmark tests. Are these the tests that you're referring to?

moodysalem: yeah sorry i meant the whole thing executable in less than 1s, i was just concerned about the 5s limit being close to the result on your machine :)

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
codyborn: I'm leaning toward leaving as a StringBuilder for the following reasons:
1. Opcodes are easier to manage as hex strings since they're more often documented as such ([ex](https://github.com/crytic/evm-opcodes)). It just saves the additional step of having to convert it.
2. `TestCodeExecutor.executeCode()` accepts a hex string and is used elsewhere in the codebase. 
3. There's no real efficiency gain of using `MutableBytes` over `StringBuilder` (correct me if I'm wrong), esp without knowing the number of operations in advance.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
atoulme: Yes, they'd be under the jmh tests.

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/ByteCodeBuilder.java
atoulme: It's tests, efficiency is moot - just looking at what level of string manipulations you have to do. Anyways, just for testing, so it can go as is.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreOperationTest.java
codyborn: @atoulme I've removed this test and created a new one in the jmh suite. See `TransientStorageOperationBenchmark.java`.

codyborn:
Also, I've ran @snreynolds new [EIP1153-specific Ethereum tests](https://github.com/snreynolds/tests/tree/filler-tests) and they all pass.
image

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/EIP1153GasCalculator.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private static final long TLOAD_GAS = WARM_STORAGE_READ_COST;
+  private static final long TSTORE_GAS = WARM_STORAGE_READ_COST;
+
+  public EIP1153GasCalculator() {}
+
+  // EIP-1153
+  @Override
+  public long getTransientLoadOperationGasCost() {
+    return TLOAD_GAS;
```
shemnon, line 27: Why not always set the cost to TLOAD_GAS?  Why update it with a fresh gas calculaotr?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/BerlinGasCalculator.java
```diff
@@ -33,9 +33,9 @@ public class BerlinGasCalculator extends IstanbulGasCalculator {
   // new constants for EIP-2929
   private static final long COLD_SLOAD_COST = 2100L;
   private static final long COLD_ACCOUNT_ACCESS_COST = 2600L;
-  private static final long WARM_STORAGE_READ_COST = 100L;
   private static final long ACCESS_LIST_ADDRESS_COST = 2400L;
   protected static final long ACCESS_LIST_STORAGE_COST = 1900L;
+  protected static final long WARM_STORAGE_READ_COST = 100L;
```
shemnon, line 38: Extraneous change, please remove.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreEVMOperationTest.java
```diff
@@ -0,0 +1,452 @@
...
+import org.apache.tuweni.units.bigints.UInt256;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+public class TStoreEVMOperationTest {
```
shemnon, line 40: This unit test should live in the EVM module as we are testing the operation itself.  Tests testing the configuration enabling TSTORE should be split out and can remain.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecFactory.java
```diff
@@ -185,6 +185,17 @@ public ProtocolSpecBuilder parisDefinition(final GenesisConfigOptions genesisCon
         evmConfiguration);
   }
 
+  public ProtocolSpecBuilder eip1153Definition(final GenesisConfigOptions genesisConfigOptions) {
```
shemnon, line 188: This should be gated by a CLI feature flag until the EIP is committed for a hard fork.  (This was how EIP-1559 and the merge were handled).  Attempts to configure it w/o the flag should produce a fatal error.

shemnon (dismissed):
Two high level changes needed:
* Forks not tied to a published fork specification need to be gated by a CLI feature flag. (This is how all non-final mainnet forks, EIP-1559, and the merge have
* unit tests teting the proper function og the TSTORE/TLOAD operation itself should be in the EVM module.  It may not be the case for existing opcodes but I have a task to move them over to the EVM module.  So new operations need isolated testing.

shemnon:
Also, considering (a) the merge is about to happen and (b) this EIP has not gone into "Eligible for Inclusion" (mostly due to no ACD call bandwidth) I'm inclined to slow roll this merging until we see the merge successfully land.

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/BerlinGasCalculator.java
codyborn: Thanks, must have been autosorted by style checker.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreEVMOperationTest.java
codyborn: Migrating these tests will create a circular dependency unless I move `test-support` out of `ethereum` into a separate module. I'm happy to do this, but I think that this is better suited for a separate PR as this will introduce a lot of small changes across many files. WDYT?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/EIP1153GasCalculator.java
codyborn: Since it has to be included in a gas calculator, I was thinking this would be renamed to the hardfork that this EIP is ultimately included in. Alternatively I can add it to the `LondonGasCalculator` now. WDYT?

shemnon: TSTORE/TLOAD hasn't been "EFIed" for any hard fork, so it would be inappropriate to put it in a named hardfork. (such as in this list - https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md).

Instead I think we need to activate EIPs by number and make EVM decisions based on that config.  In the config files it might look like this - https://github.com/shemnon/besu/tree/eips_in_config

It still would need to be wired into the EVM config, both as integrated in Besu and as a standalone library.

codyborn:
> Also, considering (a) the merge is about to happen and (b) this EIP has not gone into "Eligible for Inclusion" (mostly due to no ACD call bandwidth) I'm inclined to slow roll this merging until we see the merge successfully land.

Thanks for the review @shemnon! Totally fine with the slow roll.. wouldn't want to slow down the merge in any way :)

codyborn:
Hey @shemnon! Now that the merge has passed (congrats!) and we're looking more seriously at 1153, can we look at getting this PR merged in? I have one open question regarding unit test location. LMK what you think.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/vm/operations/TStoreEVMOperationTest.java
macfarla: I support making that a separate PR

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/EIP1153GasCalculator.java
codyborn: > Instead I think we need to activate EIPs by number and make EVM decisions based on that config.

Is this practice already in place or are you suggesting that's how it should be handled in the future?

codyborn:
Hi @shemnon, I've merged the latest upstream changes into this PR. Let me know if there are any outstanding things to address. Thanks!

shemnon:
@codyborn I updated the codebase to deal with some large changes underneath relating to optimizations.  I also updated the storage to be attached to the MessageFrame instead of the account.  It greatly simplified the implication for the EVM library as well,  Went from 44 files down to 12.

Let me know what you think.

codyborn:
> @codyborn I updated the codebase to deal with some large changes underneath relating to optimizations. I also updated the storage to be attached to the MessageFrame instead of the account. It greatly simplified the implication for the EVM library as well, Went from 44 files down to 12.
> 
> Let me know what you think.

Looks great! Just gave it a pass and agree it's much simpler with the MessageFrame approach. Nice work 👏

shemnon (approved):

PR #4120, opened 2022-07-18
increase the default max message size for p2p messages

pinges:
The current limit means that we are disconnecting useful peers during syncing.

### DIFF
--- a/ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManagerTest.java
+++ b/ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManagerTest.java
@@ -203,27 +203,8 @@ public final class EthProtocolManagerTest {
   }
 
   @Test
-  public void disconnectOnVeryLargeMessage() {
-    try (final EthProtocolManager ethManager =
-        EthProtocolManagerTestUtil.create(
-            blockchain,
-            () -> false,
-            protocolContext.getWorldStateArchive(),
-            transactionPool,
-            EthProtocolConfiguration.defaultConfig())) {
-      final MessageData messageData = mock(MessageData.class);
-      when(messageData.getSize()).thenReturn(EthProtocolConfiguration.DEFAULT_MAX_MESSAGE_SIZE + 1);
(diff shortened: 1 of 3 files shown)


jflo (commented):
I'm surprised this didn't impact any test code, that probably means we don't have this covered as well as we should. How did we end up at doubling the value?

pinges:
Currently there is a 10MB limit for incoming messages in EthProtocolManager. When a message exceeds 10MB we are ignoring the message and disconnect the peer that sent the message. There is no hard limit for messages on the eth layer in the devp2p spec.
This PR removes this check as we found out that some useful peers are sending messages greater than 10MB, resulting in their disconnection.
The message size is still limited to 16.7 MB because that is the maximum message size allowed on the RLPx layer.
We have spun up 2 nodes on Goerli that successfully synced (fast sync) within 6.5 hours with that change in place.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManagerTest.java
```diff
@@ -202,28 +202,9 @@ public void disconnectOnWrongChainId() {
     }
   }
 
-  @Test
-  public void disconnectOnVeryLargeMessage() {
```
macfarla, line 206: do we still have a test for the 16Mb limit? is that at the Rlpx layer?

macfarla:
> Currently there is a 10MB limit for incoming messages in EthProtocolManager. When a message exceeds 10MB we are ignoring the message and disconnect the peer that sent the message. There is no hard limit for messages on the eth layer in the devp2p spec. This PR removes this check as we found out that some useful peers are sending messages greater than 10MB, resulting in their disconnection. The message size is still limited to 16.7 MB because that is the maximum message size allowed on the RLPx layer. We have spun up 2 nodes on Goerli that successfully synced (fast sync) within 6.5 hours with that change in place.

do we still have a test for the 16.7Mb limit? is that at the Rlpx layer?

macfarla (approved):

pinges:
The RLPx message size is tested in test shouldThrowExceptionWhenFramingMessageTooLong in class FramerTest

Issue #4128, opened 2022-07-19
RocksDB performance regression


garyschulte:
### Description
There are several reports of slow block import times with the current rocksdb configurations from #3985. The current mitigation for this is to use `--Xplugin-rocksdb-cache-capacity=134217728` to set the cache capacity to 128mb.

### Acceptance Criteria
* besu should use configuration parameters that minimize memory footprint, while preserving performance, with 22.4.4 as a baseline for block import performance

### Steps to Reproduce (Bug)
1. sync mainnet and observe block import times
2. the import times, especially in docker, are slow enough that besu falls behind head and cannot keep up with the chain

### Versions (Add all that apply)
* Software version: `RC2-SNAPSHOT`
* Java version: [`java -version`]
* OS Name & Version: [`cat /etc/*release`]
* Kernel Version: [`uname -a`]
* Virtual Machine software & version: [`vmware -v`]
* Docker Version: [`docker version`]
* Cloud VM, type, size: [Amazon Web Services I3-large]

PR #3985, opened 2022-06-17
Apply some RocksDB documentation suggestions to reduce memory usage.

ahamlat:
Apply some [RocksDB documentation suggestions](https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB) to reduce memory usage.

- Set format_version to 5
- set optimize_filters_for_memory to true
- Increase block size to 32K
- set cache_index_and_filter_blocks to true 

There is almost 1,5 GiB difference in memory consumption between this version and version 22.4.2 on a 16 GB RAM VM after 1 day of syncing (the green line represents the metrics of this PR). 

Screenshot 2022-06-17 at 10 27 59

Sync time is equivalent between the two versions.

Screenshot 2022-06-17 at 10 35 02

Disk usage, CPU consumption and the number of open file descriptors is quite similar for tho instances.

Screenshot 2022-06-17 at 10 30 44
Screenshot 2022-06-17 at 10 30 59
Screenshot 2022-06-17 at 10 31 18

Signed-off-by: Ameziane H <ameziane.hamlat@consensys.net>

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [ ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #4132, opened 2022-07-19
Update RocksDB default block cache size

ahamlat:
Update RocksDB default block cache size to mitigate a performance regression from #3985.

## PR description

We can see in the dashboard below that we have some blocks behind with the RocksDB block cache default value 8 MB, and Besu with block cache size 128 MB is performing better (no blocks behind).

image

## Fixed Issue(s)

#4128
## Documentation

## Changelog


matkt (approved):
LGTM, waiting for a second approval

matkt:
the tests are failing because of this modification @ahamlat . it seems to need a quick update

PR #4134, opened 2022-07-19
Add support for LevelDB

ajsutton:
## PR description
Fairly quick and dirty implementation of a LevelDB plugin for Besu. It does pass the `AbstractKeyValueStorageTest` tests but is pretty lacking in unit tests otherwise currently.  LevelDB doesn't support columns by default but this adds support via a key prefix - code for that is taken from Teku (and we could pull over the unit tests for it if this is worth pursuing).

It may be better to not simulate columns since I think besu always uses globally unique keys anyway (pruning needs the ability to iterate a single column but pruning isn't exactly useful anyway). That would likely simplify the implementation and may give different performance/space characteristics as well but I haven't tried it.

## Documentation

## Changelog


ajsutton:
@matkt and @ahamlat I'm told you may be interested in this. I'm not likely to find a lot of time to push this forward so feel free to close this PR once its no longer useful to you - the code will always be around should we need it in future anyway.

ajsutton:
Going to close this one to avoid clutter. It's here for the record if anyone finds it useful.

PR #4142, opened 2022-07-21
Peering logging

macfarla:
Minor changes to logging. 
* add a separate class to log "Process message" in EthProtocolManager, so logging can be tuned separately by class

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolLogger.java
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.eth.manager;
+
+import org.hyperledger.besu.ethereum.p2p.rlpx.wire.Capability;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EthProtocolLogger {
```
siladu, line 22: What's the reason for a separate class? Doesn't this lose the context of where it was called from?

macfarla: I made it a separate class so that I can set this one to DEBUG level and EthProtocolManager to TRACE level. It's still pretty obvious from the name where it comes from. There's a similar setup in DiscoveryProtocolLogger

siladu (approved):
LGTM

PR #4149, opened 2022-07-21
Reduce RocksDB space amplification

ahamlat:
Reduce RocksDB space amplification with level_compaction_dynamic_level_bytes option set to true. In this case, the size target of levels is changed dynamically based on the size of the last level.


ahamlat:
After more than one month execution, we notice a difference in space usage of about 4 GiB, i.e the node with level_compaction_dynamic_level_bytes set to true consumed 4 GiB less than the node with the default configuration. This difference is not very significant regarding the overall usage 

image

Persist block time (block propagation time) and system metrics are quite similar :
**Persist block time** 
image
image

**CPU utilization**
image

**Memory footprint (Resident Memory)**
image

Network IO
image

Issue #4071, opened 2022-07-07
Investigate: Invalid RLP from peers: Expected current item to be a list, but it is: BYTE_ELEMENT (at bytes 0-1: [04])


macfarla:
Expected current item to be a list, but it is: BYTE_ELEMENT (at bytes 0-1: [04])

We see this message not infrequently when running besu on testnets. And besu is rejecting peers because of it. It may be legitimate but we should investigate. And improve the logging. 

Maybe move the stack trace to TRACE level

This log message from ropsten.
Latest besu 22.7.0-RC1
```
2022-07-07 20:33:47.899+10:00 | nioEventLoopGroup-3-9 | DEBUG | DeFramer | Invalid incoming message
io.netty.handler.codec.DecoderException: org.hyperledger.besu.ethereum.rlp.RLPException: Expected current item to be a list, but it is: BYTE_ELEMENT (at bytes 0-1: [04])
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:480)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:279)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at org.hyperledger.besu.ethereum.p2p.rlpx.connections.netty.AbstractHandshakeHandler.channelRead0(AbstractHandshakeHandler.java:135)
	at org.hyperledger.besu.ethereum.p2p.rlpx.connections.netty.AbstractHandshakeHandler.channelRead0(AbstractHandshakeHandler.java:44)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
[...]

PR #4150, opened 2022-07-22
fix 1 byte long disconnect reason

pinges:
Fixing https://github.com/hyperledger/besu/issues/4071


matkt (approved):
Lgtm

PR #4164, opened 2022-07-25
Fix logIndex offset bug

### LINKED ISSUE #4114: Besu returns logs with incorrect `logIndex` (different behaviour than go-ethereum)
### Description
As a RPC client consumer, I expect Besu to return logs with the same `logIndex` as go-ethereum. It does not do this.

### Acceptance Criteria
Besu should return logs with the same `logIndex` as go-ethereum.

### Steps to Reproduce (Bug)

(examples are from Sepolia - chain ID 11155111)

### go-ethereum

```curl -v --data '{"method":"eth_getLogs","id":"1","params":[{"blockhash": "0xd34a6973e49210a7ae5e487459f625c7444f9ffca93b2430400677cdaa82c293"}],"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST <ethereum RPC URL> | jq```

A sample of this response is below:
```
...
    {
      "address": "0xed6dd2313fb133e8d7a7697eda6f7e7640af0b85",
      "topics": [
        "0x0109fc6f55cf40689f02fbaad7af7fe7bbac8a3d2186600afc7d3e10cac60271",
        "0x000000000000000000000000000000000000000000000000000000000000000b",
        "0x0000000000000000000000000000000000000000000000000000000000000000"
      ],
      "data": "0x0000000000000000000000000000000000000000000000000000000062c4b107",
      "blockNumber": "0x160f2e",
      "transactionHash": "0x3b5feb43df1b020656385df5ff2e7603231e9b83b41165357f3c8d08f786d152",
      "transactionIndex": "0xa",
      "blockHash": "0xd34a6973e49210a7ae5e487459f625c7444f9ffca93b2430400677cdaa82c293",
      "logIndex": "0x18",
      "removed": false
    },
    {
      "address": "0xed6dd2313fb133e8d7a7697eda6f7e7640af0b85",
      "topics": [
        "0x0559884fd3a460db3073b7fc896cc77986f16e378210ded43186175bf646fc5f",
        "0x0000000000000000000000000000000000000000000000000000000000000005",
        "0x000000000000000000000000000000000000000000000000000000000000000b"
      ],
      "data": "0x0000000000000000000000000000000000000000000000000000000062c4b107",
      "blockNumber": "0x160f2e",
      "transactionHash": "0x3b5feb43df1b020656385df5ff2e7603231e9b83b41165357f3c8d08f786d152",
      "transactionIndex": "0xa",
      "blockHash": "0xd34a6973e49210a7ae5e487459f625c7444f9ffca93b2430400677cdaa82c293",
      "logIndex": "0x19",
      "removed": false
    }
    ...
```

### besu

```curl -v --data '{"method":"eth_getLogs","id":"1","params":[{"blockhash": "0xd34a6973e49210a7ae5e487459f625c7444f9ffca93b2430400677cdaa82c293"}],"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST <besu RPC URL> | jq```

```
    {
      "logIndex": "0x1",
      "removed": false,
      "blockNumber": "0x160f2e",
      "blockHash": "0xd34a6973e49210a7ae5e487459f625c7444f9ffca93b2430400677cdaa82c293",
      "transactionHash": "0x3b5feb43df1b020656385df5ff2e7603231e9b83b41165357f3c8d08f786d152",
      "transactionIndex": "0xa",
      "address": "0xed6dd2313fb133e8d7a7697eda6f7e7640af0b85",
      "data": "0x0000000000000000000000000000000000000000000000000000000062c4b107",
      "topics": [
        "0x0109fc6f55cf40689f02fbaad7af7fe7bbac8a3d2186600afc7d3e10cac60271",
        "0x000000000000000000000000000000000000000000000000000000000000000b",
        "0x0000000000000000000000000000000000000000000000000000000000000000"
      ]
    },
    {
      "logIndex": "0x2",
      "removed": false,
      "blockNumber": "0x160f2e",
      "blockHash": "0xd34a6973e49210a7ae5e487459f625c7444f9ffca93b2430400677cdaa82c293",
      "transactionHash": "0x3b5feb43df1b020656385df5ff2e7603231e9b83b41165357f3c8d08f786d152",
      "transactionIndex": "0xa",
      "address": "0xed6dd2313fb133e8d7a7697eda6f7e7640af0b85",
      "data": "0x0000000000000000000000000000000000000000000000000000000062c4b107",
      "topics": [
        "0x0559884fd3a460db3073b7fc896cc77986f16e378210ded43186175bf646fc5f",
        "0x0000000000000000000000000000000000000000000000000000000000000005",
        "0x000000000000000000000000000000000000000000000000000000000000000b"
      ]
    }
    ...
```

I expect logs to be uniquely indexed by `(blockHash, logIndex)`.

It appears that `logIndex` is only unique within `(blockHash, transactionIndex, logIndex)`. This is different to how go-ethereum works.

100% of the time.

### PR DESCRIPTION
**Pull request by reductionista**
## PR description

The ethereum json-rpc spec requires `logIndex` be a log's position within the block:
https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getfilterchanges

This is also what's described in the comments, but instead the current implementation of logIndex starts counting only from the beginning of the current transaction.  This PR brings public transaction logs into conformance with the spec.

This does not include full support for changing logIndex to work the same way with private transactions, because this appears to require a change to the private state storage format.   Partial support has been included for review, but commented out due to non-existent privateStateStorage.getTransactionReceipts(blockHash) method.

Maybe it's not possible to access other private tx's in the same block?

TODO:  modify expected results for more tests, so that everything passes.

## Fixed Issue(s)

Fixes #4114 

## Documentation

## Changelog

### CODE REVIEW

**Comment by ajsutton** [comment]
Private transactions are an interesting case. Someone more familiar with the way they work will have to chime in but I don't think they can be indexed based on the position in the block.  The main issue is working out what previous logs are counted, for example:

 * Logs from public transactions - seems like should obviously be counted
 * Logs from other private transactions from the same group - could be counted
 * Logs from other private transactions from different groups this node has access to - could be counted but getting weird
 * Logs from other private transactions from groups this node doesn't have access to - can't be counted as they're unknown

Notably logIndex for a private transaction wouldn't be unique within a block with any of these options because the logIndex generated for other public transactions in that block shouldn't include the logs from private transactions - otherwise a log from a public tx may reveal information about private transactions the node has access to.

File: ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/privacy/methods/priv/PrivGetTransactionReceipt.java [hunk 1]
```diff
@@ -101,8 +101,22 @@ public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
 
     LOG.trace("Processed private transaction receipt");
 
+    int logIndexStart = 0;
+    //  TODO:  Fetch all PrivateTransactionReceipt's in block, use to compute logIndex
```
pinges, line 105: We should only use receipts for private transactions for the same privacy group. 
PrivateStateStorage gives access to the "PrivateBlockMetadata" with the method "Optional<PrivateBlockMetadata> getPrivateBlockMetadata(Bytes32 blockHash, Bytes32 privacyGroupId);". It contains the list of transaction hashes for private transactions in this block. Using these hashes we can retrieve the private transaction receipts using "Optional<PrivateTransactionReceipt> getTransactionReceipt(Bytes32 blockHash, Bytes32 txHash);"

pinges (commented):
I've added a comment on what I think should be done with the private receipts

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/privacy/methods/priv/PrivGetTransactionReceipt.java
pinges: Using only the private transaction from the same privacy group makes sure that every member of the privacy group get's the same result.

reductionista: @pinges Thanks for looking at it, this sounds like a good plan!

PR #4166, opened 2022-07-26
Apply RocksDB LZ4 compression and compare post-sync metrics

ahamlat:
Apply RocksDB LZ4 compression and compare the performance/resource metrics after sync.

### DIFF
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
@@ -47,6 +47,7 @@ import org.rocksdb.BlockBasedTableConfig;
 import org.rocksdb.ColumnFamilyDescriptor;
 import org.rocksdb.ColumnFamilyHandle;
 import org.rocksdb.ColumnFamilyOptions;
+import org.rocksdb.CompressionType;
 import org.rocksdb.DBOptions;
 import org.rocksdb.Env;
 import org.rocksdb.LRUCache;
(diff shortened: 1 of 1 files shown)


ahamlat:
The two Besu nodes below are launched on AWS t3.xlarge VM, and both use Checkpoint sync.

In all the screenshots below, **_dev-besu-ohio-mainnet-ameziane-checkpoint-sync-RocksDB-lz4-2_** represents a Besu node with RocksDB Snappy compression (default), and **_dev-besu-ohio-mainnet-ameziane-checkpoint-sync-RocksDB-lz4-4_** represents a Besu node with RocksDB LZ4 compression

### Sync time
With Snappy compression (default)

- World state download : 8 hours and 41 minutes
- Blocks import : 1 day 2 hours and 12 minutes

With LZ4 compression

- World state download : 8 hours 46 minutes
- Blocks import : 1 day and 3 minutes

Besu node with RocksDB LZ4 compression has a better sync time.
image

### Block processing time (post-sync)
Post sync block time processing is quite similar, with a slightly better time for LZ4 Besu node. The median value is around 2 seconds.

**Median value**
image

**99th percentile**
image

### Disk usage
Disk usage looks similar, currently Besu node with LZ4 compression occupies 383 GiB while Besu node with Snappy compression uses 380 GiB, and two hours before LZ4 node was consuming 382 Gib and Snappy node 385 GiB. This is due to the compaction mechanism (merge Ln-1 with Ln, and remove duplicate keys) which is periodic.

image


### Memory consumption
12 hours after sync, Besu LZ4 node consumes less Resident and Virtual memory than the Snappy one.

image

### CPU utilization
Besu LZ4 node was consuming less CPU than Snappy node during sync, about one core less or 25% less, and CPU utilization after sync is very similar. 

image

### Disk IO
Disk activity is very close between the two nodes, we notice a throughput of around 250 iops average reads and 170 iops average writes, with peaks around 1000 iops reads and 340 iops writes during sync. The bandwidth is also very close, around 25 MB/s reads and 42 MB/s writes on average. 

**Snappy node**
image

**LZ4 node**
image

### Summary
12 hours after sync, LZ4 presents a lower overhead on Besu in terms of CPU utilization and memory footprint. This allows Besu to sync Faster with LZ4 compression. Block processing time is also slightly better with LZ4 than Snappy, thus I recommend enabling LZ4 compression on RocksDB when available on the OS.

fab-10 (approved):

PR #4170, opened 2022-07-26
Fixed desired gas limit setting in the reference tests service

### LINKED ISSUE #4167: Reference tests loopMul failure
### Description
I've re-implemented retesteth tests executor for our usecase and the only test failing is `vmPerformance/loopMul`, same as on retesteth's website

I did the analysis of why it happens using debugger and I found the following:
1. Gas limit is [set](https://github.com/ethereum/tests/blob/develop/GeneralStateTests/VMTests/vmPerformance/loopMul.json#L22) to `Long.MAX_VALUE` in the genesis file
2. Transaction's expected gasLimit is [set](https://github.com/ethereum/tests/blob/develop/GeneralStateTests/VMTests/vmPerformance/loopMul.json#L181) to the same value
3. When a new block is mined by `test_mineBlocks` method, gas calculator is trying to get closer to the desired gas limit, which is [hardcoded](https://github.com/hyperledger/besu/blob/main/ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/methods/TestMineBlocks.java#L67) to 10M
4. Since parent's (genesis file) gas limit is higher than the target gas limit, gas limit for the next block is set as `Long.MAX_VALUE` - `AbstractGasLimitSpecification.DEFAULT_MAX_CONSTANT_ADMUSTMENT_INCREMENT` which is less than transaction's gas limit.
5. Transaction is not included in the block, the mined block doesn't contain the expected transaction

### Steps to Reproduce (Bug)
1. Run retesteth with `besu -t GeneralStateTests -- --all --lowcpu -j1`, as specified [here](http://retesteth.ethdevops.io/). More info on retesteth [here](https://github.com/ethereum/retesteth) 

**Expected behavior:** 
`vmPerformance/loopMul` test should pass

**Actual behavior:**
It [fails](http://retesteth.ethdevops.io/results/log//2022-07-25-1658776833-besu.txt) with 
```
[0;31mError: StateTest::FillTest: TR hash not found in mined block! (Check that tr is properly mined and not oog)[33m (vmPerformance/loopMul, fork: Berlin, TrInfo: d: 0, g: 0, v: 0, TrData: `:label add-10M
```

**Proposed solution:**
Instead of using the hardcoded value of 10M we can
1. Use Long.MAX_VALUE instead so gas limit calculation always would took a step upwards which would make the tests more permissive in terms of gas limit
2. Take gas limit from `blockchain.getChainHeadHeader().getGasLimit()` which would make gas limit constant for a test chain from block to block and equal to the value in genesis file. This option makes more sense to me.

**Frequency:**
This issue is not flacky. It happens all the time

* Software version: 22.4.1
* OS Name & Version: Whatever is in Besu's docker image (hyperledger/besu:develop-openjdk-latest)
* Docker Version: 20.10.17

* Besu setup info - reference tests setup
* System info - 4 Cores (8 hyperthreaded) 2.4 hz Intel i7, 32 GB RAM

Filter94:
## Fixed Issue(s)
FIxes #4167 

## Documentation

## Changelog

### DIFF
--- a/ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/methods/TestMineBlocks.java
+++ b/ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/methods/TestMineBlocks.java
@@ -64,7 +64,7 @@ public class TestMineBlocks implements JsonRpcMethod {
     final PoWBlockCreator blockCreator =
         new PoWBlockCreator(
             context.getCoinbase(),
-            () -> Optional.of(10_000_000L),
+            () -> Optional.of(blockchain.getChainHeadHeader().getGasLimit()),
             header -> context.getExtraData(),
             context.getTransactionPool().getPendingTransactions(),
             protocolContext,


shemnon (approved):

shemnon:
DCO must be fixed - https://github.com/hyperledger/besu/runs/7523982974
And main needs to be merged.

Filter94:
@shemnon Thanks for the guidance! All done

PR #4171, opened 2022-07-26
4169 dont recurse

### LINKED ISSUE #4169: StackOverflow on recursive call to MergeCoordinator.isDecendantOf()
Kiln is now old enough (over 200k blocks since TTD) that the recursive nature of this method will blow stack:

> issue on Kiln with ExecutionEngineJsonRpcMethod | failed to exec consensus method engine_forkchoiceUpdatedV1. what could be the cause? the Besu node is connected to a lighthouse beacon. logs in the thread.

>   at org.hyperledger.besu.consensus.merge.blockcreation.MergeCoordinator.isDescendantOf(MergeCoordinator.java:536)                                                                                         ││ 2022-07-22 14:16:23.052+00:00 | vert.x-eventloop-thread-2 | ERROR | ExecutionEngineJsonRpcMethod | failed to exec consensus method engine_forkchoiceUpdatedV1                                                │
│ java.lang.StackOverflowError                                                                                                                                                                                 ││     at org.hyperledger.besu.ethereum.rlp.RLPEncodingHelpers.isSingleRLPByte(RLPEncodingHelpers.java:28)                                                                                                      │
│     at ....
 
Method needs to be refactored into a non-recursive pattern, since the distance from TTD will only increase as long as the network is alive.

jflo:
## PR description

Replaces recursive implementation of isDescendentOf with a loop.

## Fixed Issue(s)
fixes #4169 

## Documentation

## Changelog

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
@@ -534,7 +534,9 @@ public class MergeCoordinator implements MergeMiningCoordinator, BadChainListene
     Optional<BlockHeader> parentOf = Optional.of(newBlock);
 
     while (parentOf.isPresent()
-        && !parentOf.get().getBlockHash().equals(ancestorBlock.getBlockHash())) {
+        && !parentOf.get().getBlockHash().equals(ancestorBlock.getBlockHash())
+        && parentOf.get().getNumber()
+            > ancestorBlock.getNumber()) { // if on a fork, don't go further back than ancestor
       parentOf = protocolContext.getBlockchain().getBlockHeader(parentOf.get().getParentHash());
     }
 
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
@@ -530,17 +530,20 @@ public class MergeCoordinator implements MergeMiningCoordinator, BadChainListene
         newBlock.getNumber(),
         newBlock.getBlockHash());
 
-    if (ancestorBlock.getBlockHash().equals(newBlock.getHash())) {
-      return true;
-    } else if (ancestorBlock.getNumber() < newBlock.getNumber()) {
-      return protocolContext
-          .getBlockchain()
-          .getBlockHeader(newBlock.getParentHash())
-          .map(parent -> isDescendantOf(ancestorBlock, parent))
-          .orElse(Boolean.FALSE);
+    Optional<BlockHeader> parentOf = protocolContext.getBlockchain().getBlockHeader(newBlock.getParentHash());
+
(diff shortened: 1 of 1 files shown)


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -530,17 +530,24 @@ public boolean isDescendantOf(final BlockHeader ancestorBlock, final BlockHeader
         newBlock.getNumber(),
         newBlock.getBlockHash());
 
-    if (ancestorBlock.getBlockHash().equals(newBlock.getHash())) {
+    Optional<BlockHeader> parentOf =
```
garyschulte, line 533: because this is a public method that gets called from fcU, we need to check whether ancestorblock == newBlock, OR we can just start with newBlock instead of its parent.. i.e. parentBlock = Optional.of(newBlock), and then loop

garyschulte (changes requested):
LGTM, but we need to account for ancestor == newBlock

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -530,17 +530,24 @@ public boolean isDescendantOf(final BlockHeader ancestorBlock, final BlockHeader
...
-    if (ancestorBlock.getBlockHash().equals(newBlock.getHash())) {
+    // start with self, because descending from yourself is valid
+    Optional<BlockHeader> parentOf = Optional.of(newBlock);
+
+    while (parentOf.isPresent()
+        && !parentOf.get().getBlockHash().equals(ancestorBlock.getBlockHash())) {
+      parentOf = protocolContext.getBlockchain().getBlockHeader(parentOf.get().getParentHash());
+    }
+
+    if (parentOf.isPresent()
```
garyschulte, line 541: I think we can add one extra comparison here and short circuit unrelated forks: check that parentOf.getNumber() is not < ancestor.  That way we don't run all the way back to genesis if we are comparing hashes of blocks on 2 different forks

garyschulte (changes requested):

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -530,17 +530,24 @@ public boolean isDescendantOf(final BlockHeader ancestorBlock, final BlockHeader
         newBlock.getNumber(),
         newBlock.getBlockHash());
 
-    if (ancestorBlock.getBlockHash().equals(newBlock.getHash())) {
+    // start with self, because descending from yourself is valid
+    Optional<BlockHeader> parentOf = Optional.of(newBlock);
+
+    while (parentOf.isPresent()
+        && !parentOf.get().getBlockHash().equals(ancestorBlock.getBlockHash())) {
```
garyschulte, line 537: I think we can add one extra comparison here and short circuit unrelated forks: check that parentOf.getNumber() is not < ancestor.  That way we don't run all the way back to genesis if we are comparing hashes of blocks on 2 different forks

garyschulte (changes requested):

garyschulte (approved):
🚢

Issue #4061, opened 2022-07-05
Allow private networks already using London fee market to use free gas


siladu:
If you started (or hard-forked) a private network with the `londonBlock` fork without configuring the `baseFeePerGas` to zero at the time, then it's not possible to subsequently zero the base fee, meaning that you can't configure a [free gas network](https://besu.hyperledger.org/en/stable/HowTo/Configure/FreeGas/).

Configuring the baseFeePerGas to zero is not yet documented and was [only recently supported by this fix](https://github.com/hyperledger/besu/pull/4003) which works if you haven't yet upgraded to London (or are starting a new network).

~Note: related to https://github.com/hyperledger/besu/issues/4053~

---

### Suggested solution (following [a spike](https://github.com/hyperledger/besu/issues/4067))

Introduce a `zeroBaseFee` config flag which wires up a `ZeroBaseFeeMarket`, an extension of `LondonFeeMarket` which returns 0 when computing the baseFee.

### Genesis examples

To start a new free gas network on the latest fork:
```
{
  "config": {
    "grayGlacier": 0,
    "zeroBaseFee": true,
    ...
}
```

Note, this replaces the only way to currently do this (although this way will still work for new networks):
```
{
  "config": {
    "grayGlacier": 0,
    ...
    },
    "baseFeePerGas": "0x0"  
}
```

To upgrade an existing network to london (or beyond):
```
{
  "config": {
    "berlinBlock": 0,
    "londonBlock": 999,
    "zeroBaseFee": true,
    ...
}
```

To convert an existing london network to free gas:
```
{
  "config": {
    "londonBlock": 0,
    "zeroBaseFee": true, # added at some block > 0
    ...
}
```

This would be a one-way operation: once you've set `zeroBaseFee`, you can't reset the base fee back to non-zero. There are complications around the block validation rules that lead to this simplification.

`zeroBaseFee` is preferred over `baseFeePerGas` because it is a config value rather than a genesis value. This means it can be set at any time by the node operator - even if london is already activated - thus allowing conversion of a non-zero base fee network into a zero base fee (free gas) network.

Using the `transitions` mechanism was evaluated and ruled out due to requiring significant effort without a clear advantage: we don't know if any users will ever need to flexibly set a non-zero base fee in the future.
This solution doesn't preclude adding a more flexible transition in the future.
Not using transitions does mean converting an existing network to free gas requires a synchronised restart, however I would expect this to be an exceptional scenario.

[...]

PR #4174, opened 2022-07-26
Add ZeroBaseFeeMarket to support post-London free gas networks

siladu:
Details on https://github.com/hyperledger/besu/issues/4061

This allows private networks already using London fee market to start using free gas. Previously, since https://github.com/hyperledger/besu/pull/4003, it was only possible to start a free gas London network from genesis.

It also fixes an issue where the 0 initial base fee can creep back up to non-zero when a block uses enough gas: https://github.com/hyperledger/besu/issues/4218

## Documentation

## Changelog


antonydenyer (approved):

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1408,6 +1408,10 @@ public void run() {
         if (genesisConfigOptions.isQuorum()) {
           enableGoQuorumCompatibilityMode();
         }
+
+        if (genesisConfigOptions.isZeroBaseFee()) {
```
shemnon, line 1412: Is this truly needed?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/BaseFeeMarketBlockHeaderGasPriceValidationRule.java
```diff
@@ -39,6 +40,11 @@ public BaseFeeMarketBlockHeaderGasPriceValidationRule(final BaseFeeMarket baseFe
   public boolean validate(final BlockHeader header, final BlockHeader parent) {
     try {
 
+      if (baseFeeMarket instanceof ZeroBaseFeeMarket) {
```
shemnon, line 43: Design wise this is too tight of coupling.  Instead there should be a method in `BaseFeeMarket` `isValidatesBasefee` that returns true or false based on a market config.

Or `isForkBlock` should be removed and replaced with a `validationMode(long blocknumber)` method that returns an enum with 3 states - `NONE`, `INITIAL`, and `ONGOING`.  Zero market would return `NONE`

shemnon (changes requested):
Design wise there is too tight a coupling with an instanceof check for a specific market instance - https://github.com/hyperledger/besu/pull/4174/files#diff-69d6b308e912a669614fc09ad1224590d582389dfc56698f94cd63a4ad68e33eR43.  Change that to a solution that doesn't require an instanceof check.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
siladu: Not really, it's more of a convenience. I'm happy to remove.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/BaseFeeMarketBlockHeaderGasPriceValidationRule.java
siladu: I like the validationMode idea, went with that 👍

shemnon (approved):

siladu:
Converting to a draft to delay merging this in until after The Merge release

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/BaseFeeMarketBlockHeaderGasPriceValidationRule.java
```diff
@@ -40,16 +40,22 @@ public boolean validate(final BlockHeader header, final BlockHeader parent) {
...
-      if (baseFeeMarket.isForkBlock(header.getNumber())) {
+      if (BaseFeeMarket.ValidationMode.INITIAL.equals(
+          baseFeeMarket.baseFeeValidationMode(header.getNumber()))) {
         return baseFeeMarket
             .getInitialBasefee()
             .equals(header.getBaseFee().orElseThrow(() -> MissingBaseFeeFromBlockHeader()));
+      } else if (BaseFeeMarket.ValidationMode.NONE.equals(
+          baseFeeMarket.baseFeeValidationMode(header.getNumber()))) {
+        // skip validation to allow sync when a zero baseFee chain has historical baseFees
+        return true;
```
garyschulte, line 51: if ZeroBaseFeeMarket overrode getInitialBaseFee and always returned zero, we could leave this conditional as it was?

siladu: The scenario solved by this conditional is when you have some baseFee present on your network's block (enabled by London) and then subsequently set `zeroBaseFee: true`. So your blocks' baseFee goes from empty -> non-zero -> 0

During syncing, this rule ends up trying to compare the expected base fee (zero, due to overridden computeBaseFee) with the actual one (non-zero)

garyschulte (approved):
LGTM, 22.7.3-next 👍

Issue #3955, opened 2022-06-10
Besu stops importing new blocks on mainnet


MrEtherGuy:
Besu has stopped updating blocks on mainnet a couple of times recently. After a manual restart it court up with the head of the chain and contiues as normal. Im running Geth on another computer (different port) as a backup which has had no issues.

Besu v22.4.1
openJDK v11.0.15
Ubuntu server v22.04 (kernal 5.15.37)

i7 12700
32GB ram

[Log01.txt](https://github.com/hyperledger/besu/files/8877164/Log01.txt)
[Log02.txt](https://github.com/hyperledger/besu/files/8877165/Log02.txt)
[besu.txt](https://github.com/hyperledger/besu/files/8877169/besu.txt)

daniellehrner:
This log line is a bit strange:
```
Jun 06 16:16:21 node besu[968]: 2022-06-06 16:16:21.842+01:00 | nioEventLoopGroup-3-1 | INFO  | BlockPropagationManager | Saving announced block 14915630 (0x090c5aefdaf3d7e03a4f4ae3be280e8e5795958d33188baf3de5aa48eeaeab94) for future import
```
We have changed this log message a few months ago, but in your log it still show the old message format. Could be please verify that you are actually running v22.4.1? It appears to be an older version.

MrEtherGuy:
Looks good.

curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"web3_clientVersion","params":[],"id":67}' http://localhost:8545
{
  "jsonrpc" : "2.0",
  "id" : 67,
  "result" : "besu/v22.4.1/linux-x86_64/openjdk-java-11"
}

Do you want more of the log?

beaconalerts:
This occured again a couple of hours ago. I've attated a larger log dump over the outage.
[Log03.txt](https://github.com/hyperledger/besu/files/8905857/Log03.txt)
.

MrEtherGuy:
This occured again a couple of hours ago. I've attated a larger log dump over the outage.
[Log03.txt](https://github.com/hyperledger/besu/files/8905933/Log03.txt)

pinges:
@Gabriel-Trintinalia and I looked at the log file (Log03.txt) and found the following:
[...]

PR #4175, opened 2022-07-27
Add mechanism to retrieve missing blocks 

Gabriel-Trintinalia:

## PR description
Add mechanism to retrieve missing blocks when a pending block is saved

## Fixed Issue(s)
#3955

## Documentation

## Changelog

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
@@ -66,7 +66,6 @@ import org.slf4j.LoggerFactory;
 
 public class BlockPropagationManager {
   private static final Logger LOG = LoggerFactory.getLogger(BlockPropagationManager.class);
-  public static final int DEFAULT_MAX_PENDING_BLOCKS_BEFORE_RETRY = 3;
   private final SynchronizerConfiguration config;
   private final ProtocolSchedule protocolSchedule;
   private final ProtocolContext protocolContext;
@@ -441,10 +440,9 @@ public class BlockPropagationManager {
           LOG.info("Saving announced block {} for future import", block.toLogString());
         }
 
-        // Request parent of the lowest announced block when cache is too big
-        if (shouldRequestLowestPendingBlockParent()) {
-          pendingBlocksManager.lowestAnnouncedBlock().ifPresent(this::requestParentBlock);
-        }
+        // Request parent of the lowest announced block
+        pendingBlocksManager.lowestAnnouncedBlock().ifPresent(this::requestParentBlock);
+
         return CompletableFuture.completedFuture(block);
       }
     }
@@ -526,10 +524,6 @@ public class BlockPropagationManager {
     final Range<Long> importRange = config.getBlockPropagationRange();
     return importRange.contains(distanceFromLocalHead)
         && importRange.contains(distanceFromBestPeer);
-  }
-
-  private boolean shouldRequestLowestPendingBlockParent() {
-    return pendingBlocksManager.size() >= DEFAULT_MAX_PENDING_BLOCKS_BEFORE_RETRY;
   }
 
   private String toLogString(final Collection<NewBlockHash> newBlockHashs) {
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
@@ -66,7 +66,7 @@ import org.slf4j.LoggerFactory;
 
 public class BlockPropagationManager {
   private static final Logger LOG = LoggerFactory.getLogger(BlockPropagationManager.class);
-
+  public static final int DEFAULT_MAX_PENDING_BLOCKS_BEFORE_RETRY = 3;
   private final SynchronizerConfiguration config;
   private final ProtocolSchedule protocolSchedule;
   private final ProtocolContext protocolContext;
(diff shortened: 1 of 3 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/PendingBlocksManager.java
```diff
@@ -127,6 +127,10 @@ public Optional<BlockHeader> lowestAnnouncedBlock() {
         .min(Comparator.comparing(BlockHeader::getNumber));
   }
 
+  public int Size() {
```
garyschulte, line 130: camelcase

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -66,7 +66,7 @@
 
 public class BlockPropagationManager {
   private static final Logger LOG = LoggerFactory.getLogger(BlockPropagationManager.class);
-
+  public static final int DEFAULT_MAX_PENDING_BLOCKS_BEFORE_RETRY = 3;
```
ajsutton, line 69: Why would we wait to fall 3 blocks behind before attempting to retrieve the parent? I can see it makes sense to limit the number of parent blocks we'd try to download for a non-canonical chain before giving up finding where it connects to the existing chain but I'd have thought we'd want to be able to import new blocks as quickly as possible rather than delaying until we've fallen 3 blocks out of sync before starting requests.

garyschulte (approved):
Good test coverage 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
garyschulte: effectively DEFAULT_MAX_PENDING_BLOCKS_BEFORE_RETRY=1 ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/PendingBlocksManager.java
Gabriel-Trintinalia: Fixed

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
Gabriel-Trintinalia: So  blocks happen to arrive out of order and Besu can already handle it saving them for future import in the Pending blocks cache. When this happens, the parent seems to arrive just after it without the need of  requesting the parent to the peers. 

However, when for any reason we miss the parent, then Besu enters the saving blocks state and cannot recover. It seems to happen, for example, when we get an ommer block so that parentHash is different and we cannot process the children. 

The threshold would avoid having to request the parent again every time that a block arrives out of order, giving enough time to the parent to be imported organically. 

Also, in a conversation with Karim, he mentioned that it could potentially avoid this to be used in an eventual attack.

ajsutton: I'd be requesting quite eagerly - it should be reasonably rare for blocks to arrive out of order so the additional overhead of requesting a block we wind up receiving is small, but waiting for more 3 more blocks is going to generally be over 30 seconds that we stay out of sync.

In terms of risks of attacks, I'm not sure I see it - you'd have to send me a block to get me to request a block from another peer (and besu should be requesting it from a peer that has announced it has that block which is likely only the original source for a deliberately constructed attack). So we aren't getting an amplification effect that could launch a DoS.

So yeah I'd advocate for setting the max to 1 (essentially just remove the condition entirely - if we get a block and don't have it's parent we should request the parent).

Gabriel-Trintinalia: Right, I will remove the condition entirely then.

The code is requesting the lowest pending block parent:
`pendingBlocksManager.lowestAnnouncedBlock().ifPresent(this::requestParentBlock);`

It makes sense to request the parent of the block being saved instead of the the lowest block parent. Thoughts?

ajsutton: I think its reasonable to request the lowest block we need as that ensures we're requesting blocks in the order we'll need to import them (generally).  The only catch is if we're only requesting one block at a time there's a risk that the block we choose to request might not be available and we get struck repeatedly requesting it instead of moving on to one from a different fork.

Gabriel-Trintinalia: Right, I have removed the condition. The scenario you mentioned should not happen often right? Also, if that happens, the purge mechanism would remove that stuck block when the PendingBlockCache is full and  Besu would eventually request the right missing parent.

ajsutton: It shouldn't happen often but that's where a potential attack could come in - I could send Besu a block with a non-existent parent and a low block number that would always be the first to be requested.  We shouldn't have to wait for the pending cache to fill up before moving on to try a different fork. I don't think that's a new problem with this PR though so could be solved separately.

fab-10:
It is worth to know that the `BlockPropagationManager` runs concurrently with full sync, and full sync should be able to recover from situations where a parent block was missing, so it could be that the real cause of the issue is related to full sync, in the way the sync target is selected, rather that the `BlockPropagationManager`. 
Anyway this could be a fallback mechanism to workaround full sync issues, and if it works well and since this block propagation is not needed after the merge, putting more effort on fixing full sync could not be worth.

fab-10:
What happens in case more than one parent block is missing?

matkt:
> It is worth to know that the `BlockPropagationManager` runs concurrently with full sync, and full sync should be able to recover from situations where a parent block was missing, so it could be that the real cause of the issue is related to full sync, in the way the sync target is selected, rather that the `BlockPropagationManager`. Anyway this could be a fallback mechanism to workaround full sync issues, and if it works well and since this block propagation is not needed after the merge, putting more effort on fixing full sync could not be worth.

the fullsync only starts when we are too far from the head. so indeed it can be strange that the fullsync does not trigger after a certain time. but at the same time I also prefer this fix because we don't have to wait to be too far to fix the problem

fab-10:
Double checked with @matkt, and full sync and `BlockPropagationManager` always run concurrently.

Gabriel-Trintinalia:
> What happens in case more than one parent block is missing?

So if block +3 arrives and we do not have +1 and +2:

1. we will request +3's parent, which is +2. 
2. Whenever +2 arrives, it would be saved for future and then request +1. 
3. When +1 arrives, it is imported, then its children (+2) and then +2's children (+3)

PR #4177, opened 2022-07-27
Allow to set any value for baseFeePerGas in the genesis file

fab-10:
Also made baseFeeFloor final.

## PR description

Even if not pratical on mainnet or main testnes, in theory in the genesis file, the `baseFeePerGas` field could be set to any Wei value, for example it could be 1 Wei, as done in the Hive test `consensus/checkGasLimit_London`.

## Fixed Issue(s)

fixes  Hive test `consensus/checkGasLimit_London`

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LondonFeeMarket.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LondonFeeMarket.java
@@ -32,7 +32,7 @@ public class LondonFeeMarket implements BaseFeeMarket {
       GenesisConfigFile.BASEFEE_AT_GENESIS_DEFAULT_VALUE;
   static final long DEFAULT_BASEFEE_MAX_CHANGE_DENOMINATOR = 8L;
   static final long DEFAULT_SLACK_COEFFICIENT = 2L;
-  private static final Wei DEFAULT_BASEFEE_FLOOR = Wei.of(7L);
+  private static final Wei DEFAULT_BASEFEE_FLOOR = Wei.of(1L);
   private static final Logger LOG = LoggerFactory.getLogger(LondonFeeMarket.class);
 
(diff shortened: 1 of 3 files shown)


siladu (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LondonFeeMarket.java
```diff
@@ -32,14 +32,13 @@ public class LondonFeeMarket implements BaseFeeMarket {
       GenesisConfigFile.BASEFEE_AT_GENESIS_DEFAULT_VALUE;
   static final long DEFAULT_BASEFEE_MAX_CHANGE_DENOMINATOR = 8L;
   static final long DEFAULT_SLACK_COEFFICIENT = 2L;
-  // required for integer arithmetic to work when baseFee > 0
-  private static final Wei DEFAULT_BASEFEE_FLOOR = Wei.of(7L);
+  private static final Wei DEFAULT_BASEFEE_FLOOR = Wei.of(1L);
```
siladu, line 35: This means we will accept transactions into the pool that are < 7 wei, which I don't think is desirable for the vast majority of cases (where baseFee is permanently > 7).

siladu (changes requested):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/LondonFeeMarket.java
fab-10: I did this change for consistency with the base fee in the genesis file, but I am fine reverting it since it will apply only in a very corner case

fab-10: reverted

siladu (approved):
LGTM

PR #4178, opened 2022-07-27
DNS peers handled the same as boot nodes

pinges:
Peers discovered through the DNSDaemon are now added to the peerTable. These peers are potentially used (16 are randomly chosen from the peer table) in RecursivePeerRefreshStateas as the initial peers for bonding and then for finding neighbours, etc.

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
@@ -72,6 +72,7 @@ public class EthProtocolManager implements ProtocolManager, MinedBlockObserver {
   private final BlockBroadcaster blockBroadcaster;
   private final List<PeerValidator> peerValidators;
   private final Optional<MergePeerFilter> mergePeerFilter;
+  private final int maxMessageSize;
 
   public EthProtocolManager(
       final Blockchain blockchain,
@@ -100,6 +101,8 @@ public class EthProtocolManager implements ProtocolManager, MinedBlockObserver {
     this.ethPeers = ethPeers;
     this.ethMessages = ethMessages;
     this.ethContext = ethContext;
+
+    this.maxMessageSize = ethereumWireProtocolConfiguration.getMaxMessageSize();
 
     this.blockBroadcaster = new BlockBroadcaster(ethContext);
 
@@ -276,6 +279,14 @@ public class EthProtocolManager implements ProtocolManager, MinedBlockObserver {
       LOG.debug("Unsolicited message received, disconnecting from EthPeer: {}", ethPeer);
       ethPeer.disconnect(DisconnectReason.BREACH_OF_PROTOCOL);
       return;
+    }
+
+    if (messageData.getSize() > this.maxMessageSize) {
+      LOG.debug(
+          "Peer {} sent a message with size {}, larger than the max message size {}",
+          ethPeer,
+          messageData.getSize(),
+          this.maxMessageSize);
     }
 
     // This will handle responses
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
@@ -179,6 +179,10 @@ public abstract class PeerDiscoveryAgent {
     }
   }
 
+  public Optional<PeerDiscoveryController> getPeerDiscoveryController() {
+    return controller;
+  }
+
   public void updateNodeRecord() {
     if (!config.isActive()) {
       return;
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
@@ -225,7 +225,6 @@ public class PeerDiscoveryController {
 
     peerPermissions.subscribeUpdate(this::handlePermissionsUpdate);
 
-    LOG.debug("Initial discovery peers {}", initialDiscoveryPeers);
     recursivePeerRefreshState.start(initialDiscoveryPeers, localPeer.getId());
 
     final long refreshTimerId =
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
@@ -205,7 +205,6 @@ public class PeerDiscoveryController {
       throw new IllegalStateException("The peer table had already been started");
     }
 
-    LOG.debug("BOOTSTRAP NODES {}", bootstrapNodes.size());
     final List<DiscoveryPeer> initialDiscoveryPeers =
         bootstrapNodes.stream()
             .filter(peerPermissions::isAllowedInPeerTable)
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
@@ -205,6 +205,7 @@ public class PeerDiscoveryController {
       throw new IllegalStateException("The peer table had already been started");
     }
 
+    LOG.debug("BOOTSTRAP NODES {}", bootstrapNodes.size());
     final List<DiscoveryPeer> initialDiscoveryPeers =
         bootstrapNodes.stream()
             .filter(peerPermissions::isAllowedInPeerTable)
@@ -224,6 +225,7 @@ public class PeerDiscoveryController {
 
     peerPermissions.subscribeUpdate(this::handlePermissionsUpdate);
 
+    LOG.debug("Initial discovery peers {}", initialDiscoveryPeers);
     recursivePeerRefreshState.start(initialDiscoveryPeers, localPeer.getId());
 
     final long refreshTimerId =
@@ -315,7 +317,7 @@ public class PeerDiscoveryController {
             .ifPresent(
                 interaction -> {
                   bondingPeers.invalidate(peer.getId());
-                  addToPeerTable(peer);
+                  addBondedPeerToPeerTable(peer);
                   recursivePeerRefreshState.onBondingComplete(peer);
                 });
         break;
@@ -369,13 +371,8 @@ public class PeerDiscoveryController {
         .collect(Collectors.toList());
   }
 
-  private boolean addToPeerTable(final DiscoveryPeer peer) {
+  private boolean addBondedPeerToPeerTable(final DiscoveryPeer peer) {
     if (!peerPermissions.isAllowedInPeerTable(peer)) {
-      return false;
-    }
-
-    final PeerTable.AddResult result = peerTable.tryAdd(peer);
-    if (result.getOutcome() == PeerTable.AddResult.AddOutcome.SELF) {
       return false;
     }
 
@@ -389,6 +386,16 @@ public class PeerDiscoveryController {
     if (peer.getStatus() != PeerDiscoveryStatus.BONDED) {
       peer.setStatus(PeerDiscoveryStatus.BONDED);
       notifyPeerBonded(peer, now);
+    }
+
+    return addToPeerTable(peer);
+  }
+
+  public boolean addToPeerTable(final DiscoveryPeer peer) {
+    final PeerTable.AddResult result = peerTable.tryAdd(peer);
+
+    if (result.getOutcome() == PeerTable.AddResult.AddOutcome.SELF) {
+      return false;
     }
 
     if (result.getOutcome() == PeerTable.AddResult.AddOutcome.ALREADY_EXISTED) {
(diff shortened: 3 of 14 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -278,6 +281,14 @@ public void processMessage(final Capability cap, final Message message) {
       return;
     }
 
+    if (messageData.getSize() > this.maxMessageSize) {
```
matkt, line 284: not sure why this modification is present in the PR

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -205,6 +205,7 @@ public void start() {
       throw new IllegalStateException("The peer table had already been started");
     }
 
+    LOG.debug("BOOTSTRAP NODES {}", bootstrapNodes.size());
```
matkt, line 208: I think this uppercase log is not necessarily useful IMO

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -347,11 +344,16 @@ DNSDaemonListener createDaemonListener() {
                 .build();
         final DiscoveryPeer peer = DiscoveryPeer.fromEnode(enodeURL);
         peers.add(peer);
-        rlpxAgent.connect(peer);
```
matkt, line 350: why are we no longer trying to connect to the peer?

matkt (commented):
just a few questions to understand the PR

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -278,6 +281,14 @@ public void processMessage(final Capability cap, final Message message) {
...
+      LOG.debug(
+          "Peer {} sent a message with size {}, larger than the max message size {}",
+          ethPeer,
+          messageData.getSize(),
+          this.maxMessageSize);
+    }
+
```
macfarla, line 291: this is useful, although unrelated to this PR

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/RecursivePeerRefreshState.java
```diff
@@ -93,6 +93,7 @@ private boolean reachedMaximumNumberOfRounds() {
   }
 
   private void addInitialPeers(final List<DiscoveryPeer> initialPeers) {
+    LOG.debug("INITIAL PEERS: {}", initialPeers);
```
macfarla, line 96: another log not needed

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
macfarla: you put this back in, right @pinges

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -369,16 +370,11 @@ private List<DiscoveryPeer> getPeersFromNeighborsPacket(final Packet packet) {
         .collect(Collectors.toList());
   }
 
-  private boolean addToPeerTable(final DiscoveryPeer peer) {
+  private boolean addBondedPeerToPeerTable(final DiscoveryPeer peer) {
```
macfarla, line 373: what about addToPeerTableAndBond()

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
pinges: removed :-)

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
pinges: The state of the peer is already bonded or will be changed to bonded in this method

jflo (approved):
Tested and a slight initial speedup in gained peers was observed.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
garyschulte: It looks like it is just moved to the subsequent block on line 357 after adding to the peer table

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
```diff
@@ -182,6 +182,24 @@ private void buildBloomFilter() {
...
 
+  /**
+   * Returns the <code>limit</code> peers (at most) bonded closest to the provided target, based on
+   * the XOR distance between the keccak-256 hash of the ID and the keccak-256 hash of the target.
+   *
+   * @param target The target node ID.
+   * @param limit The amount of results to return.
+   * @return The <code>limit</code> closest peers, at most.
+   */
+  public List<DiscoveryPeer> nearestBondedPeers(final Bytes target, final int limit) {
```
garyschulte, line 193: I presume this is necessary in order to not give remote clients a list of unverified peers from DNSDaemon lookups?

garyschulte (approved):
retroactive LGTM

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -391,6 +386,16 @@ private boolean addToPeerTable(final DiscoveryPeer peer) {
...
 
+    return addToPeerTable(peer);
+  }
+
+  public boolean addToPeerTable(final DiscoveryPeer peer) {
+    final PeerTable.AddResult result = peerTable.tryAdd(peer);
+
+    if (result.getOutcome() == PeerTable.AddResult.AddOutcome.SELF) {
+      return false;
+    }
```
mbaxter, line 397: (bug) This method bypasses `peerPermissions`.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
```diff
@@ -193,7 +211,6 @@ private void buildBloomFilter() {
   public List<DiscoveryPeer> nearestPeers(final Bytes target, final int limit) {
     final Bytes keccak256 = Hash.keccak256(target);
     return streamAllPeers()
-        .filter(p -> p.getStatus() == PeerDiscoveryStatus.BONDED)
```
mbaxter, line 196: Why has this method been modified?  And why did we create the very similar `nearestBondedPeers`?  It looks like the functional effect of this is to make the [periodic peer searches](https://github.com/hyperledger/besu/blob/f885c4671333168089a7b86ffedf985fce81954b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java#L448-L450) run across all peers, not just bonded peers.  Perhaps this is desired, but it should be addressed in a separate PR.  

I suspect the reason this was added is because we're now [directly injecting peers](https://github.com/hyperledger/besu/pull/4178/files#diff-5c9af1f2b004d24055b333ddfc58dc0ee664dcf0649163b140e41d1520df3ce1R356) into the discovery table instead of bonding with them first.  I would suggest reverting these changes, and attempting to bond with the DNS peers rather than directly injecting them.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -228,7 +227,7 @@ public void start() {
                       disco,
                       createDaemonListener(),
                       0L,
-                      60000L,
+                      1800000L,
```
mbaxter, line 230: If the call to the DNS server fails, this introduces a big time lag before we try again.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -347,11 +346,16 @@ DNSDaemonListener createDaemonListener() {
...
       }
-      // only replace dnsPeers if the lookup was successful:
       if (!peers.isEmpty()) {
-        dnsPeers.set(peers);
+        final Optional<PeerDiscoveryController> peerDiscoveryController =
+            peerDiscoveryAgent.getPeerDiscoveryController();
+        if (peerDiscoveryController.isPresent()) {
+          final PeerDiscoveryController controller = peerDiscoveryController.get();
+          LOG.debug("Adding {} DNS peers to PeerTable", peers.size());
+          peers.forEach(controller::addToPeerTable);
```
mbaxter, line 356: As mentioned previously, we should be calling `PeerDiscoveryAgent.bond` here instead of directly injecting peers into the discovery peer table.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -278,6 +281,14 @@ public void processMessage(final Capability cap, final Message message) {
       return;
     }
 
+    if (messageData.getSize() > this.maxMessageSize) {
+      LOG.debug(
+          "Peer {} sent a message with size {}, larger than the max message size {}",
+          ethPeer,
+          messageData.getSize(),
+          this.maxMessageSize);
+    }
```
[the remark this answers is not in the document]
pinges: This is just adding debug info so we can see which clients are sending these messages

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -369,16 +369,11 @@ private List<DiscoveryPeer> getPeersFromNeighborsPacket(final Packet packet) {
...
     if (!peerPermissions.isAllowedInPeerTable(peer)) {
       return false;
     }
 
-    final PeerTable.AddResult result = peerTable.tryAdd(peer);
-    if (result.getOutcome() == PeerTable.AddResult.AddOutcome.SELF) {
-      return false;
-    }
-
     // Reset the last seen timestamp.
```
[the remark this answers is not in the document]
pinges: This message is only called for already bonded nodes

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
pinges: Yes, this was missed. I will create a new PR to add this check.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
pinges: This is using the definition of distance from the devP2P spec to find the _limit_ number of closest peers to the _target_.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
pinges: This would be cleaner, I guess. I will do that in a separate PR.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
pinges: Yes!

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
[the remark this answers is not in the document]
pinges: I've removed that. The only time I saw it in the logs was from an older Besu version!

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -179,6 +179,10 @@ public CompletableFuture<Integer> start(final int tcpPort) {
     }
   }
 
+  public Optional<PeerDiscoveryController> getPeerDiscoveryController() {
+    return controller;
+  }
+
```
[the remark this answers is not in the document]
pinges: Done

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
macfarla: good point. maybe somewhere in between?

PR #4186, opened 2022-07-27
Only stop block propagation, not tx handling.

### LINKED ISSUE #3890: Merge bug: besu producing zero transaction blocks

jflo:
## PR description

Since there is a Subscribers for each type of message handler in EthMessages, and only on handler, they all get a subscriber id of 1.  This changes the unsubscribe logic to require the caller to specify the message code handler being unsubscribed from, and removes it from the map of handlers.

## Fixed Issue(s)
fixes #3890 

## Documentation

## Changelog

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
@@ -144,8 +144,9 @@ public class BlockPropagationManager {
 
   private void clearListeners() {
     onBlockAddedSId.ifPresent(id -> protocolContext.getBlockchain().removeObserver(id));
-    newBlockSId.ifPresent(id -> ethContext.getEthMessages().unsubsribe(id));
-    newBlockHashesSId.ifPresent(id -> ethContext.getEthMessages().unsubsribe(id));
+    newBlockSId.ifPresent(id -> ethContext.getEthMessages().unsubscribe(id, EthPV62.NEW_BLOCK));
+    newBlockHashesSId.ifPresent(
+        id -> ethContext.getEthMessages().unsubscribe(id, EthPV62.NEW_BLOCK_HASHES));
     onBlockAddedSId = Optional.empty();
     newBlockSId = Optional.empty();
     newBlockHashesSId = Optional.empty();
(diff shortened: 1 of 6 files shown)


ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/AbstractBlockPropagationManagerTest.java
```diff
@@ -815,6 +815,8 @@ public void shouldStopWhenTTDReached() {
     blockPropagationManager.start();
     syncState.setReachedTerminalDifficulty(true);
     assertThat(blockPropagationManager.isRunning()).isFalse();
+    assertThat(ethProtocolManager.ethContext().getEthMessages().messageCodesHandled())
+        .doesNotContain(1, 7);
   }
```
macfarla, line 820: maybe worth a comment to say what 1 and 7 are here, or use constants EthPV62.NEW_BLOCK

macfarla (approved):
LGTM. one suggestion on using consts in test

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -144,8 +144,9 @@ private void setupListeners() {
 
   private void clearListeners() {
     onBlockAddedSId.ifPresent(id -> protocolContext.getBlockchain().removeObserver(id));
-    newBlockSId.ifPresent(id -> ethContext.getEthMessages().unsubsribe(id));
-    newBlockHashesSId.ifPresent(id -> ethContext.getEthMessages().unsubsribe(id));
+    newBlockSId.ifPresent(id -> ethContext.getEthMessages().unsubscribe(id, EthPV62.NEW_BLOCK));
+    newBlockHashesSId.ifPresent(
+        id -> ethContext.getEthMessages().unsubscribe(id, EthPV62.NEW_BLOCK_HASHES));
```
garyschulte, line 149: This implies that the subscriber id for BlockPropagationManager was the same as the id for transaction pool right?  

I wonder how many other places we could have this problem (where we expect the subscriber id to be unique across `Subcribers` instances).

garyschulte (approved):
LGTM 👍

Issue #3336, opened 2022-01-27
IllegalArgumentException: Must have at least one header to import


fab-10:
### Description
This exception occurs quite frequently, and it worth to investigate that there is a underlying issue 

```
{
	"timestamp": "2022-01-27T13:55:36,788",
	"level": "ERROR",
	"thread": "EthScheduler-Services-16 (downloadHeaders)",
	"class": "PipelineChainDownloader",
	"message": "Chain download failed. Restarting after short delay.",
	"throwable": " java.util.concurrent.CompletionException: java.lang.RuntimeException: Async operation failed. java.lang.IllegalArgumentException: Must have at least one header to import
	at java.base/java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:367)
	at java.base/java.util.concurrent.CompletableFuture.completeRelay(CompletableFuture.java:376)
	at java.base/java.util.concurrent.CompletableFuture$UniRelay.tryFire(CompletableFuture.java:1019)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at org.hyperledger.besu.services.pipeline.Pipeline.abort(Pipeline.java:180)
	at org.hyperledger.besu.services.pipeline.Pipeline.lambda$runWithErrorHandling$3(Pipeline.java:159)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.RuntimeException: Async operation failed. java.lang.IllegalArgumentException: Must have at least one header to import
	at org.hyperledger.besu.services.pipeline.AsyncOperationProcessor.outputNextCompletedTask(AsyncOperationProcessor.java:86)
	at org.hyperledger.besu.services.pipeline.AsyncOperationProcessor.attemptFinalization(AsyncOperationProcessor.java:70)
	at org.hyperledger.besu.services.pipeline.ProcessingStage.run(ProcessingStage.java:43)
[...]

PR #4189, opened 2022-07-28
Accept empty header set in range headers validation

shemnon:
## PR description

If a response to the get header P2P request only returns the header that
is the start of the range we may need to trim it to an empty response,
this is breaking the validation response. One client has started
returning only the range header start in some circumstances (which is a
valid response per spec). Because we sometimes request an overlapping
header this results in an empty stream once we cut the duplicates.

## Documentation

## Changelog

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/RangeHeaders.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/RangeHeaders.java
@@ -14,16 +14,12 @@
  */
 package org.hyperledger.besu.ethereum.eth.sync.range;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 import org.hyperledger.besu.ethereum.core.BlockHeader;
 
 import java.util.List;
 import java.util.Objects;
 
 import com.google.common.base.MoreObjects;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class RangeHeaders {
 
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/RangeHeaders.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/RangeHeaders.java
@@ -18,6 +18,7 @@ import org.hyperledger.besu.ethereum.core.BlockHeader;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 import com.google.common.base.MoreObjects;
 
@@ -40,8 +41,8 @@ public class RangeHeaders {
     return headersToImport;
   }
 
-  public BlockHeader getFirstHeaderToImport() {
-    return headersToImport.size() > 0 ? headersToImport.get(0) : null;
+  public Optional<BlockHeader> getFirstHeaderToImport() {
+    return headersToImport.size() > 0 ? Optional.of(headersToImport.get(0)) : Optional.empty();
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/RangeHeaders.java
```diff
@@ -50,7 +41,7 @@ public List<BlockHeader> getHeadersToImport() {
   }
 
   public BlockHeader getFirstHeaderToImport() {
-    return headersToImport.get(0);
+    return headersToImport.size() > 0 ? headersToImport.get(0) : null;
```
jflo, line 44: Making this return Optional might be a bit more helpful to the reader, and would lend itself to simpler conversion to an empty stream later on.

shemnon:
> suggested improvement is _optional_

I see what you did there.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/RangeHeaders.java
```diff
@@ -14,29 +14,21 @@
...
 
   private final SyncTargetRange range;
   private final List<BlockHeader> headersToImport;
 
   public RangeHeaders(
       final SyncTargetRange checkpointRange, final List<BlockHeader> headersToImport) {
-    if (headersToImport.isEmpty()) {
-      LOG.debug(String.format("Headers list empty. Range: %s", checkpointRange.toString()));
-    }
-    checkArgument(!headersToImport.isEmpty(), "Must have at least one header to import");
```
mbaxter, line 39: Only question here is if this changes the sync process in any meaningful way - could this cause sync to hang, for example, if we have a peer that is repeatedly returning nothing?  It seems that previously we would throw here, and I'm guessing the sync process would be interrupted because of the exception.

mbaxter: @ajsutton - thoughts?

fab-10:
should fix https://github.com/hyperledger/besu/issues/3336

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/RangeHeaders.java
shemnon: Throwing the exception did not disconnect the peer, so we still had the peer.  

But per spec (a) an empty response was 100% fine and (b) Erigon's response was not empty, we just trimmed off the one value we already had.  So there are two routes we may be passing in an empty resposne down the pipeline and both cases the peer is behaving validly, so disconnecting doesn't seem right.

mbaxter: Not saying we should disconnect, just want to make sure we're not introducing anything that could cause the sync to stall / get stuck.  But we handle useless responses [elsewhere](https://github.com/hyperledger/besu/blob/f885c4671333168089a7b86ffedf985fce81954b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeer.java#L215), so even if we got stuck asking for the same range of headers from a single unresponsive peer, I think we would eventually disconnect and move on ...

shemnon: I haven't seen any pauses.  Rather than breaking the pipe an empty stream is the result, and that would get aggregated the same as a broken step, as nothing gets added to the next set of object for the next step in the pipeline.  I haven't seen any slowdown overnight and I started a fresh sync and it's proceeding the same as before with no observable differences.

Side note: I'm really looking forward to being able to bit-torrent the blocks prior to the merge.  EIP-4444 and friends will be welcomed with open arms.

PR #4190, opened 2022-07-28
Feature/engine api override

### LINKED ISSUE #4172: Implement override to force engine-api in the absence of TTD
### Description
In order to smooth the preparations for the merge, consensus is that we want execution clients to enable the engine API even in the absence of merge parameters.  This is in essence a re-enabling of the `--engine-rpc-enabled` api flag removed in #3875

### Acceptance Criteria
* make the engine-api available in the absence of merge configs
* ensure rational behavior pre-TTD calls to engine api endpoints

garyschulte:
## PR description
Allow besu to have the engine api enabled, when there are not merge configs.  includes:
* un-hiding/un-deprecating `--engine-rpc-enabled`
* remove old isMergeEnabled checks in a few block header validators
* use merge-specific block header validator, rather than appending to MainnetBlockHeaderValidator
* disable all engine api methods except engine_exchangeTransitionConfigurationV1 in the absence of an engine-api-compatible coordinator
* add "placeholder" TTD in the absence of a TTD configuration

Tested with Hive and Kurtosis since this affects block validation.  Both are 🟢 

## Fixed Issue(s)

fixes #4172 

## Documentation

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
@@ -166,10 +166,10 @@ public class EngineNewPayload extends ExecutionEngineJsonRpcMethod {
     final var block =
         new Block(newBlockHeader, new BlockBody(transactions, Collections.emptyList()));
 
-    if (mergeContext.map(c -> c.isSyncing()).orElse(Boolean.TRUE) || parentHeader.isEmpty()) {
+    if (mergeContext.get().isSyncing() || parentHeader.isEmpty()) {
       LOG.debug(
           "isSyncing: {} parentHeaderMissing: {}, adding {} to backwardsync",
-          mergeContext.map(c -> c.isSyncing()).orElse(null),
+          mergeContext.get().isSyncing(),
           parentHeader.isEmpty(),
           block.getHash());
       mergeCoordinator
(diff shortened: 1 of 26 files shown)


daniellehrner:
Just today there was a discussion on Discord that revealed that all CL clients already use a placeholder TTD for mainnet. So I guess the easiest would be to just use that one instead:

https://discord.com/channels/595666850260713488/892088344438255616/1002018160268038164

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
```diff
@@ -166,10 +166,10 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
     final var block =
         new Block(newBlockHeader, new BlockBody(transactions, Collections.emptyList()));
 
-    if (mergeContext.isSyncing() || parentHeader.isEmpty()) {
+    if (mergeContext.map(c -> c.isSyncing()).orElse(Boolean.TRUE) || parentHeader.isEmpty()) {
```
jflo, line 169: This is kinda hard to read. It looks like it evaluates to a no-op, does anyone else find this harder to read?

jflo (approved):
One thing that i found hard to parse, but otherwise this cleans up a ton of conditionals and class casting.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
garyschulte: I can see how it is harder to read, it is a side effect of making mergeContext an Optional.  looking for an approach that is easier on the 👀

garyschulte: Supplier/function reference ftw 👍

Issue #3491, opened 2022-02-23
ConcurrentModificationException in Subscribers class


fab-10:
### Description
Subscribers class thrown this exception

```
java.util.ConcurrentModificationException
	at java.base/java.util.concurrent.CopyOnWriteArrayList.bulkRemove(CopyOnWriteArrayList.java:855)
	at java.base/java.util.concurrent.CopyOnWriteArrayList.bulkRemove(CopyOnWriteArrayList.java:833)
	at java.base/java.util.concurrent.CopyOnWriteArrayList.removeIf(CopyOnWriteArrayList.java:816)
	at org.hyperledger.besu.ethereum.eth.manager.EthPeers.reattemptPendingPeerRequests(EthPeers.java:151)
	at org.hyperledger.besu.ethereum.eth.manager.EthPeers.dispatchMessage(EthPeers.java:141)
	at org.hyperledger.besu.ethereum.eth.manager.EthProtocolManager.processMessage(EthProtocolManager.java:275)
	at org.hyperledger.besu.ethereum.p2p.network.NetworkRunner.lambda$setupHandlers$0(NetworkRunner.java:135)
	at org.hyperledger.besu.ethereum.p2p.rlpx.connections.PeerConnectionEvents.lambda$dispatchMessage$1(PeerConnectionEvents.java:64)
	at org.hyperledger.besu.util.Subscribers.lambda$forEach$0(Subscribers.java:112)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.hyperledger.besu.util.Subscribers.forEach(Subscribers.java:109)
	at org.hyperledger.besu.ethereum.p2p.rlpx.connections.PeerConnectionEvents.dispatchMessage(PeerConnectionEvents.java:64)
	at org.hyperledger.besu.ethereum.p2p.rlpx.connections.netty.ApiHandler.channelRead0(ApiHandler.java:99)
	at org.hyperledger.besu.ethereum.p2p.rlpx.connections.netty.ApiHandler.channelRead0(ApiHandler.java:33)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
[...]

PR #4206, opened 2022-08-02
Fix ConcurrentModificationException on ReattemptPendingPeerRequests

Gabriel-Trintinalia:

## PR description
A ConcurrentModificationExpetion happens when EthPeers reattempts to execute its Pending Requests and one  request causes a peer to disconnect. This happens because RemoveIf on CopyOnWriteArrayList will test the filter for each item first and mark those who need to be removed and then throw the exception if the list was modified by any filter. 

Replacing removeIf with iterator seems to solve the issue.

## Fixed Issue(s)
#3491

## Documentation

## Changelog

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
@@ -26,6 +26,7 @@ import java.time.Clock;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.ConcurrentModificationException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -36,6 +37,7 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -164,9 +166,14 @@ public void dispatchMessage(final EthPeer peer, final EthMessage ethMessage) {
...
-  private void reattemptPendingPeerRequests() {
+  @VisibleForTesting
+  void reattemptPendingPeerRequests() {
     synchronized (this) {
-      pendingRequests.removeIf(PendingPeerRequest::attemptExecution);
+      try {
+        pendingRequests.removeIf(PendingPeerRequest::attemptExecution);
+      } catch (ConcurrentModificationException ex) {
+        LOG.warn("Failed execution of pending requests: {}", ex.getClass().getName());
+      }
```
fab-10, line 176: Is it correct to says that if the ConcurrentModificationException occurs, it stops processing the following pending requests in the list?
In that case could make sense to catched the exception in `PendingPeerRequest.attemptExecution` and return false there, so the processing of following item will be done.

fab-10 (commented):
A question about where to catch the exception

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
Gabriel-Trintinalia: Hey Fabio, not really, CopyOnWriteArrayList will test the filter for each item first and mark those who need to be removed and then throw the exception if the list was modified by any filter. That means that it will try to execute all the requests first and then remove then from the list.  

It is not the request itself that causes the exception, but the fact that it modifies the list during the removeIf. The following snippet replicates it: https://onecompiler.com/java/3ybxwtufc

I will try to modify this logic so we first try the requests and then remove them. It should avoid the expection

Gabriel-Trintinalia: Replacing removeIf with an iterator seems to not throw the exception

mark-terry (approved):
LGTM.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
fab-10: looks good with the iterator!

fab-10 (approved):

PR #4209, opened 2022-08-02
Fix post-merge fast sync regression

### LINKED ISSUE #4208: Bugfix: sync crossing merge boundary failing pre-merge header validation rules
### Description
#4190 introduced a regression where a batch of blocks that cross TTD can fail to validate.  For example, if 200 blocks are fetched to import and 100 of them are pre-merge, and 100 post-merge, the pre-merge block validator will get selected because we have not yet imported a block that crosses the terminal total difficulty.

### Acceptance Criteria
* fast sync block imports should use the correct/corresponding block header validators when importing batches, even if a batch crosses the terminal total difficulty block

### Steps to Reproduce (Bug)
1. fast sync ropsten using the merge commit of #4190

 all fast syncs complete

all fast sync methods fail to import post-merge blocks

99.5% of the time (depending on the batch size and the roll of the dice where the block range boundary is)

garyschulte:
## Fixed Issue(s)

fixes #4208

## Documentation

## Changelog


garyschulte:
dropped in favor #4224

Issue #4197, opened 2022-07-29
Investigate and optimise to avoid duplicate requests when requesting parents


pinges:
### Description
It seems that Besu tries to retrieve the same parent block multiple times.
Version tested: 22.7.0-RC3
Network tested: goerli
Logs: https://app.logz.io/#/goto/2860f315aa951c7afb3bd193995efe39?switchToAccountId=172317
Extract
```
{"timestamp":"2022-07-28T08:58:30,038","level":"INFO","thread":"nioEventLoopGroup-3-3","class":"BlockPropagationManager","message":"Saving announced block 7304399 (0x320eeaf23e6b851f03ec55b3ed15d8fb39a8ae35e8ddf3811104de45f1aa475d) for future import","throwable":""}
{"timestamp":"2022-07-28T08:58:30,039","level":"INFO","thread":"nioEventLoopGroup-3-3","class":"BlockPropagationManager","message":"Retrieving parent 0x9286825c5d7909c73efeaa925b7a6fcd52fe298f923bafb05690e55646b1eff4 of block #7304377 from peers","throwable":""}
{"timestamp":"2022-07-28T08:58:30,198","level":"INFO","thread":"nioEventLoopGroup-3-9","class":"BlockPropagationManager","message":"Retrieving parent 0x9286825c5d7909c73efeaa925b7a6fcd52fe298f923bafb05690e55646b1eff4 of block #7304377 from peers","throwable":""}
{"timestamp":"2022-07-28T08:58:35,896","level":"INFO","thread":"nioEventLoopGroup-3-8","class":"BlockPropagationManager","message":"Saving announced block 7304402 (0x3fd368509ca0e3520e62aaaf5d37de4f51cc957e9f6a4c7b777bdce735acf4d4) for future import","throwable":""}
{"timestamp":"2022-07-28T08:58:35,896","level":"INFO","thread":"nioEventLoopGroup-3-8","class":"BlockPropagationManager","message":"Retrieving parent 0x9286825c5d7909c73efeaa925b7a6fcd52fe298f923bafb05690e55646b1eff4 of block #7304377 from peers","throwable":""}
{"timestamp":"2022-07-28T08:58:36,153","level":"INFO","thread":"nioEventLoopGroup-3-9","class":"BlockPropagationManager","message":"Retrieving parent 0x9286825c5d7909c73efeaa925b7a6fcd52fe298f923bafb05690e55646b1eff4 of block #7304377 from peers","throwable":""}
[...]

PR #4227, opened 2022-08-08
Improve pending blocks retrieval mechanism

Gabriel-Trintinalia:

## PR description
When a block is saved for future import, requesting the lowest announced block parent does not seem to be always the best option. It can be that we cannot retrieve this block and then we keep selecting the lowest block over and over and eventually it does not fix the gap. A better approach for trying to unstick could be to request the parent of the lowest pending ancestor

Changes:
- When saving a block for future import, try to retrieve the parent of the lowest pending block connected to this block.
- When saving a block for future import, do not try to retrieve the parent the lowest announced block.
- Avoid requesting the same parent block several times when saving pending blocks
- Improve logs when saving/importing a block

## Fixed Issue(s)
potential aid for #3955
#4197

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -155,29 +155,49 @@ private void clearListeners() {
...
+    if (!maybeProcessPendingChildrenBlocks(newBlock)) {
+      traceLambda(
+          LOG, "There are no pending blocks ready to import for block {}", newBlock::toLogString);
+      maybeProcessNonAnnouncedBlocks(newBlock);
+    }
+
+    if (blockAddedEvent.getEventType().equals(EventType.HEAD_ADVANCED)) {
+      final long head = protocolContext.getBlockchain().getChainHeadBlockNumber();
+      final long cutoff = head + config.getBlockPropagationRange().lowerEndpoint();
+      pendingBlocksManager.purgeBlocksOlderThan(cutoff);
```
ajsutton, line 175: Seems suspicious that previously access to this was guarded by a `synchronized (pendingBlocksManager)` but this access is not.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -364,21 +376,19 @@ private CompletableFuture<Block> processAnnouncedBlock(
...
   }
 
-  private void requestParentBlock(final BlockHeader blockHeader) {
+  private void requestParentBlock(final Block block) {
+    BlockHeader blockHeader = block.getHeader();
     if (requestedBlocks.add(blockHeader.getParentHash())) {
       retrieveParentBlock(blockHeader);
     } else {
-      LOG.trace("Parent block with hash {} was already requested", blockHeader.getParentHash());
+      LOG.info("Parent block with hash {} is already requested", blockHeader.getParentHash());
```
ajsutton, line 384: This seems like an overly noisy log message. debug seems reasonable, but this isn't something the user is likely to care about so shouldn't be at info level.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/PendingBlocksManager.java
```diff
@@ -127,6 +131,22 @@ public Optional<BlockHeader> lowestAnnouncedBlock() {
...
+   * @param block target block
+   * @return An optional with the lowest ancestor pending block
+   */
+  public Optional<Block> pendingAncestorBlockOf(final Block block) {
+    if (block == null) {
+      return Optional.empty();
+    }
+    if (pendingBlocks.containsKey(block.getHeader().getParentHash())) {
+      return pendingAncestorBlockOf(pendingBlocks.get(block.getHeader().getParentHash()).block());
+    }
```
ajsutton, line 146: It's probably not a great idea to use a recursive algorithm here.  An iterative approach should be just as simple, likely more performant and not risk stack overflow.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/PersistBlockTask.java
```diff
@@ -217,13 +217,12 @@ protected void cleanup() {
       final double timeInS = getTaskTimeInSec();
       LOG.info(
           String.format(
-              "Imported #%,d / %d tx / %d om / %,d (%01.1f%%) gas / (%s) in %01.3fs. Peers: %d",
-              block.getHeader().getNumber(),
+              "Imported %s / %d tx / %d om / %,d (%01.1f%%) gas in %01.3fs. Peers: %d",
+              block.getHeader().toLogString(),
```
ajsutton, line 221: This change is very out of scope for this PR.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/AbstractBlockPropagationManagerTest.java
```diff
@@ -641,22 +641,16 @@ public void shouldRequestLowestAnnouncedPendingBlockParent() {
...
-    // skip first block then create messages from blocklist
-    blocks.stream()
-        .skip(1)
-        .map(this::createNewBlockHashMessage)
-        .forEach(
-            message -> { // Broadcast new block hash message
-              EthProtocolManagerTestUtil.broadcastMessage(ethProtocolManager, peer, message);
-            });
+    EthProtocolManagerTestUtil.broadcastMessage(
+        ethProtocolManager, peer, createNewBlockHashMessage(blocks.get(3)));
```
ajsutton, line 653: I might be missing something, but I'm a bit unclear why this test needed to be changed. It looks like it was setting up a pretty realistic situation of having missed one block from a chain.

It probably does make sense to add new tests here to demonstrate the change in functionality - in particular having pending blocks received from multiple forks and checking parents of both forks are requested.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -333,6 +333,6 @@ public static org.hyperledger.besu.ethereum.core.BlockHeader convertPluginBlockH
   }
 
   public String toLogString() {
-    return getNumber() + " (" + getHash() + ")";
+    return String.format("#%d (%s)", getNumber(), getHash());
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: Totally agree, fixed.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
Gabriel-Trintinalia: Changed to debug

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/PendingBlocksManager.java
Gabriel-Trintinalia: It makes sense. Done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/PersistBlockTask.java
Gabriel-Trintinalia: Agreed. Fixed

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
Gabriel-Trintinalia: This PR does not change that. It was previously accessed outside of the `synchronized` block.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/AbstractBlockPropagationManagerTest.java
Gabriel-Trintinalia: Yep, I will add a new test

fab-10 (changes requested):
This goes in the right directions, implementing a basic backward sync to fill gaps in the block propagation, I have reported just a couple of minor things.

During this code review, I have also reviewed the role of `maybeProcessNonAnnouncedBlocks` that looks forward, so in the future to try to proactively fetch new blocks, that usually should be notified by peers soon after, so not sure how much it is useful, instead this class was missing something that looks backward, that is probably much needed, not sure if the author relied on the fact that full sync eventally fills the gaps, but when full sync stalls (for yet unknown reasons) we are in throuble, and this PR should fix that!

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/PendingBlocksManager.java
```diff
@@ -127,6 +131,23 @@ public Optional<BlockHeader> lowestAnnouncedBlock() {
...
+  /**
+   * Get the lowest pending ancestor block saved for a block
+   *
+   * @param block target block
+   * @return An optional with the lowest ancestor pending block
+   */
+  public Optional<Block> pendingAncestorBlockOf(final Block block) {
+    if (block == null) {
+      return Optional.empty();
+    }
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: Done! 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -480,6 +484,34 @@ CompletableFuture<Block> importOrSavePendingBlock(final Block block, final Bytes
...
+      // Save pending block
+      if (pendingBlocksManager.registerPendingBlock(block, nodeId)) {
+        LOG.info(
+            "Saved announced block for future import {} - {} saved block(s)",
+            block.toLogString(),
+            pendingBlocksManager.size());
+      }
+
+      // Try to get the lowest ancestor pending for this block, so we can import it
+      Optional<Block> lowestPending = pendingBlocksManager.pendingAncestorBlockOf(block);
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: Sure, it makes sense. I created another method to encapsulate the logic to request the pending blocks  and now the only part that is within the `synchronised` block is the actual `registerPendingBlock` method. Thanks!

fab-10 (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
ajsutton: hmm, not related to this PR then but I suspect there's a race condition there because the prune winds up trying to update both `pendingBlocks` and `pendingBlocksByParentHash`. If `PendingBlocksManager.registerPendingBlock` and `deregisterPendingBlock` wind up running at the same time, the purge may run in between register adding to `pendingBlocks` and adding to `pendingBlocksByParentHash`. The purge would then only remove from `pendingBlocks` and we'd wind up with an entry in `pendingBlocksByParentHash` that never gets removed.  It's very unlikely and a very small memory leak with no effect on functionality though.

Plus there may be some other restriction somewhere I'm missing that ensures those two methods never run at the same time for the same block.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/state/PendingBlocksManagerTest.java
```diff
@@ -259,4 +260,62 @@ public void shouldReturnLowestBlockByNumber() {
...
   }
+
+  @Test
+  public void shouldReturnLowestAncestorPendingBlock() {
+    final BlockDataGenerator gen = new BlockDataGenerator();
+    final Block parentBlock = gen.block();
+
+    final Block childBlock = gen.nextBlock(parentBlock);
+    final Block childBlock2 = gen.nextBlock(childBlock);
+    final Block childBlock3 = gen.nextBlock(childBlock2);
```
ajsutton, line 271: I'd be tempted to add a second fork in here so the test is actually ensuring it follows the specific fork. Currently it would pass just by returning the pending block with the lowest height.

ajsutton (approved):
I'm leaning on @fab-10 's review here a fair bit for the finer details of how this works but looks good to my understanding.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -364,21 +376,19 @@ private CompletableFuture<Block> processAnnouncedBlock(
     return getBlockFromPeers(Optional.of(peer), blockHash.number(), Optional.of(blockHash.hash()));
   }
 
-  private void requestParentBlock(final BlockHeader blockHeader) {
+  private void requestParentBlock(final Block block) {
+    BlockHeader blockHeader = block.getHeader();
```
pinges, line 380: final

pinges (approved):
LGTM, you might want to think about getting rid of the second parameter of the pendingAncestorBlockOf() method.
Some finals missing :-)

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/state/PendingBlocksManagerTest.java
Gabriel-Trintinalia: Done!

PR #4130, opened 2022-07-19
Useful response improves reputation

pinges:
Currently there is no way for peers improve their reputation. With this PR peers improve their reputation when they send a useful response.

Signed-off-by: Stefan <stefan.pingel@consensys.net>

PR #4230, opened 2022-08-08
Introduce a cap to reputation score increase

fab-10:

## PR description

#4130 introduced an way to increase reputation in case the peer is doing good, but there is no cap to the reputation score the peer can have.

The log shows peers with very high reputation, for example:
```
FullSyncTargetManager","message":"Caught up to best peer: PeerId 0xb3e02a84270c590f562e2c8b8274930c54b0a00ae001d44bd790b55c3bbc578c70738e6d4a2d32df78b6cf27a071fe54a068f3192c42ac29bd81a6663fe4fa0f, reputation PeerReputation 561, validated? true, disconnected? false, client: Geth/v1.10.21-stable-67109427/linux-amd64/go1.18.5, connection 1312721736, enode enode://b3e02a84270c590f562e2c8b8274930c54b0a00ae001d44bd790b55c3bbc578c70738e6d4a2d32df78b6cf27a071fe54a068f3192c42ac29bd81a6663fe4fa0f@34.92.61.128:29888, chain state: ChainState{estimatedHeight=7369229, bestBlock=BestBlock{totalDifficulty=0x0000000000000000000000000000000000000000000000000000000000a45b44, blockHash=0xb7c2ba396d3f4561fe619796e7ef399080823161f4b831c5b3f48445e7bf13a1, number=7369228}}. Current peers: 3","throwable":""}
```
since there is no cap to how much the reputation can increase.

I think make sense to have an upper bound, a peer can increase its repution until it reaches the default max again, otherwise a peer that is good for a long time, and start to be useless, will take a long time before its reputation goes below other more recent good peers

## Documentation

## Changelog


jflo (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputation.java
```diff
@@ -30,21 +30,32 @@
...
+
+  private final int maxScore;
+
+  public PeerReputation() {
+    this(DEFAULT_MAX_SCORE);
+  }
+
+  public PeerReputation(final int maxScore) {
+    this.maxScore = maxScore;
+    this.score = maxScore;
```
macfarla, line 57: can we set the initial score less than the max? It's useful right now to see if a peer has score = default, then they haven't us sent anything yet

macfarla (changes requested):
having an upper limit makes sense. it would make sense though to have the default score less than the max

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputation.java
fab-10: update to set an initial (default 100) and a max (default 150) score, defaults could be adapted based on metrics and adjustment size

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputationTest.java
```diff
@@ -24,8 +24,14 @@
...
 public class PeerReputationTest {
 
+  private static final int INITIAL_SCORE = 25;
   private static final int MAX_SCORE = 50;
-  private final PeerReputation reputation = new PeerReputation(MAX_SCORE);
+  private final PeerReputation reputation = new PeerReputation(INITIAL_SCORE, MAX_SCORE);
+
+  @Test(expected = java.lang.IllegalArgumentException.class)
+  public void shouldThrowOnInvalidInitialScore() {
+    new PeerReputation(2, 1);
```
macfarla, line 33: this should be 1, 2

macfarla (approved):
LGTM

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputationTest.java
fab-10: the wrong order is intented to test that initial must not be greather than max

PR #4237, opened 2022-08-09
Better management of jemalloc presence/absence

fab-10:

## PR description

If jemalloc is not present, an error message is printed, but the absence of jemalloc is not a real error, so in this PR we try to understand if jemalloc is available and print messages that are easier for the user to understand, and invite him to install jemalloc to get better memory usage

## Documentation

## Changelog


jflo (approved):

macfarla (approved):

besu/src/main/scripts/unixStartScript.txt
```diff
@@ -182,8 +182,18 @@ APP_ARGS=`save "\$@"`
...
-export LD_PRELOAD=libjemalloc.so
+if [ "\$darwin" = "false" -a "\$msys" = "false" ]; then
+  # check if jemalloc is available
+  TEST_JEMALLOC=\$(LD_PRELOAD=libjemalloc.so sh -c true 2>&1)
+
+  # if jemalloc is available the output is empty, otherwise the output has an error line
+  if [ -z "\$TEST_JEMALLOC" ]; then
+    export LD_PRELOAD=libjemalloc.so
+  else
+    echo "INFO: jemalloc not found, you could improve memory usage installing it."
```
ajsutton, line 193: I'm not sure that echoing from the shell script is a good option for logging. It won't integrate with any logging configuration options and so won't necessarily be written to the right place. We could check if the `LD_PRELOAD` env var is unset on the java side to log this warning via the logging system.

fab-10: Good point, I was not totally sure of this output as well, and I agree that moving the log to java side is a stronger solution.

garyschulte: it is worth noting that LD_PRELOAD is primarily a linux thing.  If we move this check into java, we are going to have to derive what our OS is so we don't erroneously report missing lib.  E.g. mac or windows

fab-10: moved the log from the script to Java, and used a dedicated env var when using jemalloc, since `LD_PRELOAD` could be used for other things

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1492,6 +1494,17 @@ private void registerConverters() {
...
   }
 
+  private void detectJemalloc() {
+    // jemalloc is only supported on Linux at the moment
+    if (PlatformDetector.getOSType().equals("linux")) {
+      Optional.ofNullable(environment.get("BESU_USING_JEMALLOC"))
+          .ifPresentOrElse(
+              present -> logger.info("Using jemalloc"),
+              () ->
+                  logger.info("jemalloc not found, you could improve memory usage installing it"));
```
ajsutton, line 1504:
```suggestion
                  logger.info("jemalloc library not found, memory usage may be reduced by installing it"));
```

ajsutton (approved):
LGTM. Just a suggested wording change for the log message - feel free to take or leave it.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
fab-10: Taken thanks

PR #4253, opened 2022-08-11
poc snapshot wo rocksdb

matkt:

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.bonsai;
+
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.ethereum.storage.StorageProvider;
+import org.hyperledger.besu.ethereum.trie.MerklePatriciaTrie;
+import org.hyperledger.besu.ethereum.trie.StoredMerklePatriciaTrie;
+import org.hyperledger.besu.ethereum.trie.StoredNodeFactory;
+import org.hyperledger.besu.ethereum.worldstate.StateTrieAccountValue;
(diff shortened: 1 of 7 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -193,13 +226,27 @@ public void clearFlatDatabase() {
...
+    final int bucketNumber = (int) ((blockNumber / BUCKET_SIZE) % 2);
+    final int cleanBucketCounter = BUCKET_SIZE - (int) (blockNumber % BUCKET_SIZE);
+    if (cleanBucketCounter == 1) {
+      ((bucketNumber == 0)
+              ? snapshotTrieBranchStorage.getSecond()
+              : snapshotTrieBranchStorage.getFirst())
+          .clear();
+    }
+  }
+
```
garyschulte, line 239: If I am reading correctly, we wont be writing to the next bucket until we reach the 128 block boundary, and the current bucket gets wiped at that boundary.  this will leave us with empty buckets at 128 block boundaries won't it?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+    return Optional.empty();
+  }
+
+  @Override
+  public Optional<Bytes> getAccountStateTrieNode(final Bytes location, final Bytes32 nodeHash) {
+    if (nodeHash.equals(MerklePatriciaTrie.EMPTY_TRIE_NODE_HASH)) {
+      return Optional.of(MerklePatriciaTrie.EMPTY_TRIE_NODE);
+    } else {
+      return snapshotTrieBranchStorage
```
garyschulte, line 97: we should be able to use the blocknumber % bucket size ternary here rather than just searching both?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateKeyValueStorage.java
matkt: indeed I was wrong. we need buckets of 256 to be ok

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
matkt: I don't know if we can get this information when we call this method. Because it is called in the trie class but can be . Need to check

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateKeyValueStorage.java
garyschulte: or we could keep it at 128 and write to both buckets simultaneously, and keep the clearing behavior as is.

shemnon: You would still have to calculate which bucket to read from, but it would also remove the requirement to switch buckets on a cache miss.  Doubling write load to cut read load in half.  Depends on which is done more often.

PR #4256, opened 2022-08-12
Refactor and fix retrying get block switching peer

fab-10:
## PR description
RetryingGetBlockFromPeersTask had a problem that prevented to complete
when all the peers fail, and that also had the
consequence to not removing the failed requested block from the internal
caches in BlockPropagationManager, that could cause a stall since that
block will not be tried to be retrieved again.

Made also an abstraction of a retrying peer task that tries a different peer on every execution, so it can used also for other tasks

## Fixed Issue(s)

relates to #3955 

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
```diff
@@ -0,0 +1,113 @@
...
+        assignedPeer
+            .filter(u -> getRetryCount() == 1) // first try with the assigned peer if present
+            .map(Optional::of)
+            .orElseGet(this::selectNextPeer); // otherwise select a new one from the pool
+
+    if (maybePeer.isEmpty()) {
+      traceLambda(
+          LOG,
+          "No peer found to try to execute task at attempt {}, tried peers {}",
+          triedPeers::toString,
```
macfarla, line 69: this param and next one should be swapped

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
```diff
@@ -0,0 +1,113 @@
...
+        this::getRetryCount);
+
+    return executeTaskOnCurrentPeer(peerToUse)
+        .thenApply(
+            peerResult -> {
+              traceLambda(
+                  LOG,
+                  "Got result {} from peer {}, attempt {}",
+                  peerResult::toString,
+                  this::getRetryCount);
```
macfarla, line 91: missing the peer {} param

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/ethtaskutils/RetryingSwitchingPeerMessageTaskTest.java
```diff
@@ -0,0 +1,111 @@
...
+  @Test
+  public void completesWhenBestPeerEmptyAndSecondPeerIsResponsive()
+      throws ExecutionException, InterruptedException {
+    // Setup data to be requested and expected response
+
+    // Setup first unresponsive peer
+    final RespondingEthPeer firstPeer =
+        EthProtocolManagerTestUtil.createPeer(ethProtocolManager, 10);
+
+    // Setup first unresponsive peer
```
macfarla, line 54: comment incorrect

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/ethtaskutils/RetryingSwitchingPeerMessageTaskTest.java
```diff
@@ -0,0 +1,111 @@
...
+  @Test
+  public void completesWhenBestPeerTimeoutsAndSecondPeerIsResponsive()
+      throws ExecutionException, InterruptedException {
+    // Setup data to be requested and expected response
+    peerCountToTimeout.set(1);
+    // Setup first unresponsive peer
+    final RespondingEthPeer firstPeer =
+        EthProtocolManagerTestUtil.createPeer(ethProtocolManager, 10);
+
+    // Setup first unresponsive peer
```
macfarla, line 86: comment incorrect

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/ethtaskutils/RetryingSwitchingPeerMessageTaskTest.java
```diff
@@ -0,0 +1,111 @@
...
+    final RespondingEthPeer secondPeer =
+        EthProtocolManagerTestUtil.createPeer(ethProtocolManager, 9);
+
+    // Execute task and wait for response
+    final T requestedData = generateDataToBeRequested();
+    final EthTask<T> task = createTask(requestedData);
+    final CompletableFuture<T> future = task.run();
+
+    // First peer timeouts
```
macfarla, line 95: I think the peerCountToTimeout.set(1) line could be moved here to make more sense?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/task/RetryingGetBlockFromPeersTaskTest.java
```diff
@@ -0,0 +1,72 @@
...
+        metricsSystem,
+        maxRetries,
+        Optional.of(requestedData.getResult().getHash()),
+        GENESIS_BLOCK_NUMBER);
+  }
+
+  @Test
+  @Override
+  @Ignore("GetBlock could not return partial response")
+  public void failsWhenPeerReturnsPartialResultThenStops() {}
```
macfarla, line 63: is this new? do we plan on fixing these tests?

macfarla (commented):
a few comments. I like the task that tries different peers, I can see that being useful!

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
fab-10: done

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/ethtaskutils/RetryingSwitchingPeerMessageTaskTest.java
fab-10: done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
fab-10: done

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/ethtaskutils/RetryingSwitchingPeerMessageTaskTest.java
fab-10: done

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/task/RetryingGetBlockFromPeersTaskTest.java
fab-10: this is part of the parent class `RetryingMessageTaskTest` , that has tests for use cases that do not always apply to actual derived test classes, in our case we test getting a block from peer, and there is no partial response, either the block is dowloaded or not, so this test is ignored since do no apply. You can find similar ignore in all the other concrete implementation of these abstract test classes.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private void refreshPeers() {
+    final EthPeers peers = getEthContext().getEthPeers();
+    // If we are at max connections, then refresh peers disconnecting one of the failed peers,
+    // or the least useful
+    if (peers.peerCount() >= peers.getMaxPeers()) {
+      failedPeers.stream()
+          .filter(peer -> !peer.isDisconnected())
+          .findAny()
+          .or(() -> peers.streamAvailablePeers().sorted(peers.getBestChainComparator()).findFirst())
```
gezero, line 145: it is interesting that getBestChainComparator should give the worst peer first

gezero (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
fab-10: yes I thought the same :)

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/RetryingGetBlockFromPeersTask.java
```diff
@@ -14,135 +14,101 @@
...
-        .filter(peer -> !triedPeers.contains(peer))
-        .findFirst()
-        .orElseThrow(NoAvailablePeersException::new);
-  }
-
   @Override
   protected boolean isRetryableError(final Throwable error) {
-    return (blockNumber > protocolContext.getBlockchain().getChainHeadBlockNumber())
-        && (super.isRetryableError(error) || error instanceof IncompleteResultsException);
+    return super.isRetryableError(error) || error instanceof IncompleteResultsException;
```
garyschulte, line 92: do we no longer need to check that the blockNumber is greater than chain head?  or rather should we check that the optional blockHash is already present in the chain?

fab-10: I removed that check because as you said checking the block number alone is not correct, and could prevent downloading legit new blocks with a different hash for the same block number.

About the check in GetBlockFromPeerTask, we usually check before fetching from peers if we have that block, so the chance to re-download a block that we already have is very low, to justify the change.

garyschulte (approved):
👍

PR #4265, opened 2022-08-16
Block prop on first final

jflo:
Stops block propagation via devp2p when network finalizes, not at TTD as previous.


besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -100,6 +100,7 @@
 import java.util.OptionalLong;
 import java.util.function.Supplier;
 
+import org.jetbrains.annotations.NotNull;
```
garyschulte, line 103: Did you mean javax.validation.constraints.NotNull ?  the jetbrains annotation only has an effect in the IDE afaik.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/AbstractBlockPropagationManagerTest.java
```diff
@@ -842,25 +845,27 @@ public void shouldThrowErrorWhenNoValidPeerAvailable() {
   }
 
   @Test
-  public void shouldStopWhenTTDReached() {
+  public void shouldStopWhenFinalized() {
```
garyschulte, line 848: non-blocking, but an additional test that asserts block propagation never starts if we are past TTD would be a good measure.  I don't see a DefaultSynchronizerTest class however.

garyschulte (changes requested):
LGTM, minor change and non-blocking ask

garyschulte:
FYI, running hive master against this branch resolves the 4 failing gossip engine-transition tests 👍

fab-10 (approved):

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
garyschulte: removed per convo with @jflo

PR #4269, opened 2022-08-17
Filter out disconnected peers when fetching available peers

fab-10:

## PR description

Currently a peer that is marked as disconnected can be returned when asking `EthPeers` for available peers, but returning a disconnected peer is not useful and potentially a cause of problems since all tasks will fail on it.
Not sure how much this affect peer selection in the real world, but in unit tests it happens to choose a disconnected peer for request.
In any case it seems reasonable to add this check, and we can evaluate if we also need to require that a peer is fully validated, before returning it as available

## Documentation

## Changelog

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
@@ -203,7 +203,9 @@ public class EthPeers {
   }
 
   public Stream<EthPeer> streamAvailablePeers() {
-    return streamAllPeers().filter(EthPeer::readyForRequests);
+    return streamAllPeers()
+        .filter(EthPeer::readyForRequests)
+        .filter(peer -> !peer.isDisconnected());
   }
 
   public Stream<EthPeer> streamBestPeers() {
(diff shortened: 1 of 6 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -203,7 +203,9 @@ public Stream<EthPeer> streamAllPeers() {
   }
 
   public Stream<EthPeer> streamAvailablePeers() {
-    return streamAllPeers().filter(EthPeer::readyForRequests);
+    return streamAllPeers()
+        .filter(EthPeer::readyForRequests)
+        .filter(peer -> !peer.isDisconnected());
```
garyschulte, line 208: seems sensible enough.  This would be additional protection from a race between besu disconnecting a peer and subsequently responding to that disconnect?

fab-10: No sure I got your question, but this part is not affecting the way we process incoming messages

pinges (approved):
LGTM

matkt (approved):
LGTM

macfarla (approved):
LGTM

PR #4271, opened 2022-08-17
3955 logs and retries

fab-10:
## PR description

This PR is built on top of  #4256 so check it first is not yet merged.

Currently even if we use the RetryingGetBlockFromPeers, that tries to download the block checking all the peers in turn, if all peers fail, the retrieval is not repeated, unless the block is announced again.
This PR improve the block propagation manager, to repeat the download of a block, until some conditions apply: it is still not imported and the distance from the local head is within configured range.

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -397,34 +403,110 @@ private CompletableFuture<Block> retrieveParentBlock(final BlockHeader blockHead
...
+      debugLambda(
+          LOG, "No need to retry to get block {} since it is already present", block::toLogString);
+      return CompletableFuture.completedFuture(block);
+    }
+
+    long distanceFromChainHead = blockNumber - blockchain.getChainHeadBlockNumber();
+    if (!config.getBlockPropagationRange().contains(distanceFromChainHead)) {
+      debugLambda(
+          LOG,
+          "Not retrying to get block {} since we are too far from local chain head {}",
```
jflo, line 475: Is this a good place to suggest the user to start over with a checkpoint sync? If we got here during a backwards sync, that might be the only way to recover.

jflo (approved):
one minor non-blocking suggestion, looks good though.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockPropagationManager.java
fab-10: Not sure I got this, backward sync does not run along with BlockPropagationManager

fab-10:
@jflo added a timeout and did some refactoring in the last commit, do you mind to review again?

macfarla (approved):

PR #4283, opened 2022-08-22
interpret an empty string (0x80) as an unsigned byte of 0

### LINKED ISSUE #4279: Accept wit/80 from Nethermind
I have created this issue against Nethermind: https://github.com/NethermindEth/nethermind/issues/4443
Should we allow this encoding of the witness capability in Besu for now, so that we can decode the hello from Nethermind? This is a bug in Nethermind, but other clients probably allow that encoding, otherwise they would have noticed that they cannot connect to other clients …
#4443 Nethermind hello message reports wrong version for witness protocol
Describe the bug  
The hello message sent by Nethermind contains the capability for the witness protocol. It is RLP encoded as:  
c5: list with length 5  
83: String of length 3  
776974: “wit”  
Show more
<https://github.com/NethermindEth/nethermind|NethermindEth/nethermind>NethermindEth/nethermind | Today at 11:05 AM | Added by GitHub

quite possible that the number 0 is encoded as a 0 length byte string.

pinges:

## PR description
This PR addresses Besu failure to decode a devp2p Hello message that contains the witness protocol capability version 0. In that Hello message the version of the witness protocol is encoded as 0x80 (zero length string), which is a valid way of encoding the version number 0.  

## Fixed Issue(s)
fixes #4279

## Changelog


ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/WireMessagesSedesTest.java
```diff
@@ -59,6 +67,8 @@ private static void assertSedesWorks(final byte[] data) {
     // Re-serialize and check that data matches
     final BytesValueRLPOutput out = new BytesValueRLPOutput();
     peerInfo.writeTo(out);
-    assertThat(out.encoded()).isEqualTo(input);
+    if (encEqualsInput) {
```
ajsutton, line 70: Should probably only re-serialise the data if we're going to actually check it matches.

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/WireMessagesSedesTest.java
```diff
@@ -43,11 +43,19 @@ public void deserializeHello() {
...
                 + "3b1e5d41");
 
-    assertSedesWorks(rlp);
+    assertSedesWorks(rlp, true);
+
+    rlp =
+        decodeHexDump(
+            "f8a305b74e65746865726d696e642f76312e31332e342d302d3365353937326332342d32303232303831312f5836342d4c696e75782f362e302e36e4c5836574683ec5836574683fc58365746840c58365746841c58365746842c5837769748082765db84005c95b2618ba1ca53f0f019d1750d12769267705e46b4dbfb77f73998b21d30973161542a2090bcaa5876e6aed99436009f3f646029bb723b8dff75feec27374");
+    // This test conatains a hello message from Nethermind. The Capability version of the wit
+    // capability is encoded as 0x80, which is an empty string
```
macfarla, line 52: this is the actual HELLO from Nethermind, right?

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/WireMessagesSedesTest.java
```diff
@@ -59,6 +67,8 @@ private static void assertSedesWorks(final byte[] data) {
     // Re-serialize and check that data matches
     final BytesValueRLPOutput out = new BytesValueRLPOutput();
     peerInfo.writeTo(out);
-    assertThat(out.encoded()).isEqualTo(input);
+    if (encEqualsInput) {
+      assertThat(out.encoded()).isEqualTo(input);
+    }
```
macfarla, line 72: this is necessary because we don't encode zero in the same way? Can we link to the ticket that will fix this?

macfarla (approved):
LGTM just some clarifying questions

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPInput.java
```diff
@@ -221,6 +221,11 @@ public interface RLPInput {
    *     long.
    */
   default int readUnsignedByte() {
+    if (isZeroLengthString()) {
+      // Decode an empty string (0x80) as an unsigned byte with value 0
+      readBytes();
+      return 0;
+    }
```
ajsutton, line 228: I suspect we should be fixing the serialisation of 0 as well to ensure we write it as an 0 length string. Otherwise we won't be to spec.

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/WireMessagesSedesTest.java
pinges: yes

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/WireMessagesSedesTest.java
pinges: I can add a comment!

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPInput.java
pinges: I've created an issue to look at encoding and decoding of 0. This PR fixes the decoding of the Nethermind hello, which was out priority.

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/WireMessagesSedesTest.java
pinges: I've created an issue to look at encoding and decoding of 0. This will probably be reversed when this ticket has been finished.

PR #4146, opened 2022-07-21
API Method: rollup_createPayload

jpnovais:
## PR description
Adds a new api method `rollup_createPayloadV1` as an extension to [Engine API](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md) in order to allow to use Besu a L2 execution engine for Rollups;

This method allows to create a Payload with a given list of transactions, executed by it's given order. Invalid transactions are returned with error reason. Unprocessed transactions that don't fit within block gasLimit are returned as well.

To enable this method and rollup behaviour, Besu should run with `--engine-rollup-extension-enabled=true` on CLI

#### Request
Arguments:
- `parentBlockHash`: DATA, 32 Bytes - hash of parent block of this new payload;
- `transactions`:  Array of DATA - Array of transaction objects, each object is a byte list (DATA) representing TransactionType || TransactionPayload or LegacyTransaction as defined in [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)
- `prevRandao`: DATA, 32 Bytes - value for the prevRandao field of the new payload
- `feeRecipient`: DATA, 20 Bytes - suggested value for the feeRecipient field of the new payload
- `timestamp`: QUANTITY, 64 Bits - value for the timestamp field of the new payload

```json
{
    "id": "100",
    "jsonrpc": "2.0",
    "method": "rollup_createPayloadV1",
    "params": [
        "0xa326af38b7ad1ba192c00b89675c450ffd8e30d83c128ef37a62da26b9f58b9d",
        [
            "0xf86180820bb88259d894f1a98bb35ff74fb6eb07c145c233434aee4f1f960a8026a0b7b0cc34f06e99da2a0eb3f290be06fa5982e5b7883b9d2a46e09ee03861955ba008972a7505395050eb78e332d811049c824534a46d751b8305e1b3f819533017",
            "0xf8610a820bb88259d894f1a98bb35ff74fb6eb07c145c233434aee4f1f960a8025a09f0f2f41a2a7228d342b81f3937441a0767e5d78c25c86a8e5e5bec7e5fd3b33a02775ee8f94d673774f7f6fd090fa4ccd15b797a47236d59c0ad83549191660e4",
[...]

PR #4289, opened 2022-08-22
Synchronous Vertx as dependency for engine api

jpnovais:
## PR description
Make synchronous vertex EngineJsonRpcMethods a dependency instead of creating a new one. Key reasons for this change:
- allow other methods that require synchronization with `engine_` to share same vertx instance. (e.g `rollup_` extension method in #4146)
- improve testability, before some tests failed with too many files open because vext inside EngineJsonRpcMethods was not being closed on test's `afterEach` hook;

## Documentation

## Changelog


besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -986,6 +987,9 @@ private Map<String, JsonRpcMethod> jsonRpcMethods(
       final Map<String, BesuPlugin> namedPlugins,
       final Path dataDir,
       final RpcEndpointServiceImpl rpcEndpointServiceImpl) {
+    // sync vertex for engine consensus API, to process requests in FIFO order;
+    final Vertx syncVertx = Vertx.vertx(new VertxOptions().setWorkerPoolSize(1));
```
antonydenyer, line 991: `final Vertx consesusEngineServer = Vertx.vertx(new VertxOptions().setWorkerPoolSize(1));`

antonydenyer:
I think it's fine, feels no worse than what is there already - would appreciate a second opinion.

antonydenyer (approved):

PR #4285, opened 2022-08-22
Reduced fast-sync-min-peers default value to 2.

mark-terry:
Signed-off-by: mark-terry <mark.terry@consensys.net>

## PR description
Reduces the default fast-sync minimum peers from 5 to 2 to improve syncing performance in low participation networks

## Fixed Issue(s)

Partially addresses #4202 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #4298, opened 2022-08-23
Reduce fast-sync-min-peers for post merge

matkt:

## PR description

Post merge we no longer need to wait to have 5 peers to validate a pivot block knowing that the pivot block comes from the  consensus client. This PR proposes to reduce the min peers to 1 for post merge network. It will always be possible to change it via the flag if necessary

## Documentation

## Changelog


macfarla:
Does this overlap with https://github.com/hyperledger/besu/pull/4285

matkt:
Reducing to two for all networks seems a bit dangerous imo. This may have to be done on a case-by-case basis depending on the networks. This PR focuses more on the post merge

macfarla:
@mark-terry can you review this one?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -506,9 +507,8 @@ private InetAddress autoDiscoverDefaultIP() {
   @Option(
       names = {"--fast-sync-min-peers"},
       paramLabel = MANDATORY_INTEGER_FORMAT_HELP,
-      description =
-          "Minimum number of peers required before starting fast sync. (default: ${DEFAULT-VALUE})")
-  private final Integer fastSyncMinPeerCount = FAST_SYNC_MIN_PEER_COUNT;
+      description = "Minimum number of peers required before starting fast sync.")
```
fab-10, line 510: is it possible to says which is the default pre and post merge? 
since there is actually a default, it seems not correct to remove the info here

fab-10 (commented):
Looks good, just want to verify if it is possible to restore the default info

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2783,10 +2783,19 @@ private Optional<PkiBlockCreationConfiguration> maybePkiBlockCreationConfigurati
   }
 
   private SynchronizerConfiguration buildSyncConfig() {
+    Integer fastSyncMinPeers = fastSyncMinPeerCount;
+    if (fastSyncMinPeers == null) {
+      if (isMergedNetwork(network)) {
```
mark-terry, line 2788: Is there more logic required here for the case where the sync begins at a pre-merge block? I've had cases recently on Ropsten where using fast sync with <=2 peers resulted in the sync stalling (ended up on a fork perhaps?)

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
matkt: I modified to add the default value again

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
matkt: we cannot start a sync on a premerge block because of the snapsync. We need to use <128 old blocks. So after small amount of time it's not possible

matkt: if the network merged the pivot block need to come from CL so 1 it's enough .

fab-10 (approved):

fab-10 (approved):
Please update the CHANGELOG

mark-terry (approved):

PR #12, opened 2019-09-16
[PAN-3155] Handle discovery peers with updated endpoints

mbaxter:
## PR description
Update discovery logic to consider a peer with an unknown discovery endpoint to be unknown regardless of whether we've encountered a peer with the same node id before.  This makes the discovery logic more forgiving in the face of node restarts.  

If `nodeA` bonds with `nodeB`, then `nodeB` leaves the network and later comes back with a different ip address or listening port, `nodeA` would previously continue trying to communicate with `nodeB` at its original address.  With these changes, `nodeA` will now treat the restarted `nodeB` as a new peer and communicate with it on its updated endpoint.  Additionally, `nodeB`'s information will be updated in the peer table so that neighbors requests return updated information on this node.

PR #4310, opened 2022-08-25
Log imported block info post merge

fab-10:
After the switch to PoS, info of the imported blocks are not more logged, this PR re-add the log when a block is imported via the NewPayload engine RPC call, and also update the info logged to remove ommers and add base fee.

Sample of the message:
`Imported #12,862,033 / 5 tx / 439,018 (1.5%) gas / base fee 7 wei / (0xb1b6d51e6382a1c88b42344f53d5321ee63bec0226fae502876af55e7109f431) in 0.028s.`

## Documentation

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
@@ -42,6 +42,7 @@ import org.hyperledger.besu.ethereum.core.BlockHeaderFunctions;
 import org.hyperledger.besu.ethereum.core.Difficulty;
 import org.hyperledger.besu.ethereum.core.Transaction;
 import org.hyperledger.besu.ethereum.core.encoding.TransactionDecoder;
+import org.hyperledger.besu.ethereum.eth.manager.EthPeers;
 import org.hyperledger.besu.ethereum.mainnet.BodyValidation;
 import org.hyperledger.besu.ethereum.mainnet.MainnetBlockHeaderFunctions;
 import org.hyperledger.besu.ethereum.rlp.RLPException;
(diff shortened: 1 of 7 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
```diff
@@ -252,4 +255,17 @@ JsonRpcResponse respondWithInvalid(
         requestId,
         new EnginePayloadStatusResult(INVALID, latestValidHash, Optional.of(validationError)));
   }
+
+  private void logImportedBlockInfo(final Block block, final double timeInS) {
+    LOG.info(
+        String.format(
+            "Imported #%,d / %d tx / %,d (%01.1f%%) gas / base fee %s / (%s) in %01.3fs.",
```
shemnon, line 262: Nit: could we put basefee between tx and gas?  We are dropping ommers and we could just re-use the position for basefee.

shemnon: Question: is dropping peer count intentional?  Does the ELs no longer gossip?  If there is still gossip I think we want peer counts.

shemnon (commented):
Other than the question about peer count LGTM, will approve once that has a good answer.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
fab-10: 1. Moved the base fee where ommes where
2. I was thinking on logging peers count in another way, but not yet sure how, so it make sense to readd them, since there is still gossip

shemnon (approved):

PR #4325, opened 2022-08-29
refactors to rely on blocks added, and behave like singleton

jflo:
Will not show ready for merge on startup when local stored chain is past ttd.
Uses block observer to determine state change

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
@@ -15,6 +15,8 @@
 package org.hyperledger.besu.controller;
 
 import org.hyperledger.besu.config.GenesisConfigFile;
+import org.hyperledger.besu.config.GenesisConfigOptions;
+import org.hyperledger.besu.consensus.merge.MergeContext;
 import org.hyperledger.besu.consensus.merge.PandaPrinter;
 import org.hyperledger.besu.consensus.merge.PostMergeContext;
 import org.hyperledger.besu.consensus.merge.TransitionBackwardSyncContext;
@@ -29,8 +31,10 @@ import org.hyperledger.besu.ethereum.GasLimitCalculator;
 import org.hyperledger.besu.ethereum.ProtocolContext;
 import org.hyperledger.besu.ethereum.blockcreation.MiningCoordinator;
 import org.hyperledger.besu.ethereum.chain.Blockchain;
+import org.hyperledger.besu.ethereum.core.Difficulty;
 import org.hyperledger.besu.ethereum.core.MiningParameters;
 import org.hyperledger.besu.ethereum.core.PrivacyParameters;
+import org.hyperledger.besu.ethereum.core.Synchronizer;
 import org.hyperledger.besu.ethereum.eth.EthProtocolConfiguration;
 import org.hyperledger.besu.ethereum.eth.manager.EthContext;
 import org.hyperledger.besu.ethereum.eth.manager.EthMessages;
@@ -39,6 +43,8 @@ import org.hyperledger.besu.ethereum.eth.manager.EthProtocolManager;
 import org.hyperledger.besu.ethereum.eth.manager.EthScheduler;
 import org.hyperledger.besu.ethereum.eth.manager.MergePeerFilter;
 import org.hyperledger.besu.ethereum.eth.peervalidation.PeerValidator;
+import org.hyperledger.besu.ethereum.eth.sync.DefaultSynchronizer;
+import org.hyperledger.besu.ethereum.eth.sync.PivotBlockSelector;
 import org.hyperledger.besu.ethereum.eth.sync.SynchronizerConfiguration;
 import org.hyperledger.besu.ethereum.eth.sync.backwardsync.BackwardSyncContext;
 import org.hyperledger.besu.ethereum.eth.sync.state.SyncState;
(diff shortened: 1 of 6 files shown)


besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
```diff
@@ -176,6 +189,50 @@ protected PluginServiceFactory createAdditionalPluginServices(
...
+                pivotBlockSelector);
+    final GenesisConfigOptions maybeForTTD = configOptionsSupplier.get();
+    Optional<Difficulty> currentTotal =
+        protocolContext
+            .getBlockchain()
+            .getTotalDifficultyByHash(protocolContext.getBlockchain().getChainHeadHash());
+
+    PandaPrinter.init(
+        currentTotal,
+        Difficulty.of(configOptionsSupplier.get().getTerminalTotalDifficulty().get()));
```
fab-10, line 222: should this be inside the following if and use `maybeForTTD` var?

besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
```diff
@@ -176,6 +189,50 @@ protected PluginServiceFactory createAdditionalPluginServices(
...
+    if (maybeForTTD.getTerminalTotalDifficulty().isPresent()) {
+      LOG.info(
+          "TTD present, creating DefaultSynchronizer that stops propagating after finalization");
+      protocolContext.getConsensusContext(MergeContext.class).addNewForkchoiceMessageListener(sync);
+    }
+    sync.subscribeInSync(PandaPrinter.getInstance());
+    protocolContext.getBlockchain().observeBlockAdded(PandaPrinter.getInstance());
+    protocolContext
+        .getConsensusContext(MergeContext.class)
+        .addNewForkchoiceMessageListener(PandaPrinter.getInstance());
```
fab-10, line 232: it will be nice to unsubscribe the panda printer after it has done the job, or are you thinking of removing this code after the merge?

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PandaPrinter.java
```diff
@@ -31,18 +34,48 @@
...
+  }
+
+  public static PandaPrinter getInstance() {
+    if (INSTANCE == null) {
+      throw new IllegalStateException("Uninitialized, unknown ttd");
+    }
+    return INSTANCE;
+  }
+
+  public PandaPrinter(final Optional<Difficulty> currentTotal, final Difficulty ttd) {
```
fab-10, line 58: since this is a singleton, should this be private?

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PandaPrinter.java
```diff
@@ -31,18 +34,48 @@
...
+    this.ttd = ttd;
+    if (currentTotal.isPresent() && currentTotal.get().greaterOrEqualThan(ttd)) {
+      this.readyBeenDisplayed.set(true);
+      this.ttdBeenDisplayed.set(true);
+    }
```
fab-10, line 63: should also `finalizedBeenDisplayed` be set to `true`?

fab-10 (commented):
Please update the CHANGELOG too

besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
jflo: I'll just remove it. Unsubscribing is a good idea, but then the PandaPrinter needs to keep a reference to the ConsensusContext so it can unsubscribe after finalization.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PandaPrinter.java
jflo: Yes it should! But since our tests are multi-threaded and singletons become a problem, I made that public to simplify testing.

macfarla (approved):

besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
```diff
@@ -176,6 +189,49 @@ protected PluginServiceFactory createAdditionalPluginServices(
...
+              .getBlockchain()
+              .getTotalDifficultyByHash(protocolContext.getBlockchain().getChainHeadHash());
+      PandaPrinter.init(
+          currentTotal, Difficulty.of(maybeForTTD.getTerminalTotalDifficulty().get()));
+    }
+    sync.subscribeInSync(PandaPrinter.getInstance());
+    protocolContext.getBlockchain().observeBlockAdded(PandaPrinter.getInstance());
+    protocolContext
+        .getConsensusContext(MergeContext.class)
+        .addNewForkchoiceMessageListener(PandaPrinter.getInstance());
```
fab-10, line 231: These should be inside the if as well

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PandaPrinter.java
fab-10: I see, but it could be package private

fab-10 (changes requested):

fab-10 (approved):

PR #4352, opened 2022-09-06
adds means to report causing exception up to RPC layer 

### LINKED ISSUE #4349: engine-api: Catch and handle underlying storage exceptions
### Description
Besu's engine api should separately catch and handle underlying exceptions (such as storage exceptions) and failures that are unrelated to block execution/validation.  Rather than presuming that any exception during block processing translates into an `INVALID` response.

For example, on mnsf-12,  the lighthouse-besu-2 node built and proposed a block, but on the subsequent engine_newPayload execution of the proposal, encountered a RocksDB timeout.  This timeout exception was interpreted as a bad block and we incorrectly responded to CL with `INVALID`

### Acceptance Criteria
* engine-api should correctly discriminate between validation and execution failures rather than underlying exceptions such as rocksdb storage exceptions.

jflo:
so it can decide how to present it to the CL

fixes #4349

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
@@ -110,7 +110,9 @@ public class MainnetBlockValidator implements BlockValidator {
         shouldPersist ? maybeWorldState.get() : maybeWorldState.get().copy();
 
     final BlockProcessor.Result result = processBlock(context, worldState, block);
-    if (result.isFailed()) {
+    if (result.isFailed() && result.causedBy().isPresent()) {
+      return handleAndReportFailure(block, "Error processing block", result);
+    } else if (result.isFailed()) {
       return handleAndReportFailure(block, "Error processing block");
     }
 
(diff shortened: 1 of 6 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -110,7 +110,9 @@ public BlockValidator.Result validateAndProcessBlock(
         shouldPersist ? maybeWorldState.get() : maybeWorldState.get().copy();
 
     final BlockProcessor.Result result = processBlock(context, worldState, block);
-    if (result.isFailed()) {
+    if (result.isFailed() && result.causedBy().isPresent()) {
```
matkt, line 113: why does the fact that there is a causedBy make us pass the condition? shouldn't we check what has a causedBy value?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -144,6 +146,21 @@ private Result handleAndReportFailure(final Block invalidBlock, final String rea
     return new Result(reason);
   }
 
+  private Result handleAndReportFailure(
+      final Block invalidBlock, final String reason, final BlockProcessor.Result result) {
+    if (result.causedBy().isPresent()) {
+      LOG.error(
+          "{}. Block {}, caused by {}", reason, invalidBlock.toLogString(), result.causedBy());
+      // TODO: if it's an internal error, don't add it
```
matkt, line 154: is it for a next PR ?

jflo: Yes. Right now I don't know what different causes we might encounter, or which ones should result in treating the block as bad.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
jflo: Because if we have an exception during block _processing_ we want to return a include that as the cause of our _validation_ result.  If it is not present, we just continue with the old logic. Trying to keep this low impact till we can think it through better.

matkt (approved):
LGTM

Issue #4294, opened 2022-08-23
Losing multiple peers within seconds because of BREACH_OF_PROTOCOL


pinges:
### Description
While syncing a number of nodes on Goerli and MainnetI noticed that most nodes do lose almost all their peers within a few seconds. The following screenshot was taken from a node that was started from a nightly build from around the 18th of August (besu/v22.7.1-dev-f847ead5/linux-x86_64/openjdk-java-11):
![0EA9234B-F693-4A6F-9CE8-99217A57E4FC.jpeg](https://images.zenhubusercontent.com/5e701925567b4e8945ea7800/85b15fbb-60ed-4765-a6a5-c38d8f4f8cdc)

A number of instances on different nodes have been checked and they all seem to follow the same pattern.
The following is from the logfile from one of the nodes:

`[DEBUG] 2022-08-18 15:15:00.125 [nioEventLoopGroup-3-1] RequestManager - Request ID incorrect (BREACH_OF_PROTOCOL), disconnecting peer PeerId 0x3f88068013a17437b7fae2f155894a33dd94beb55b9d82c667123c6431d4967b52a9ce155f0c44dd2dcb405cbef5addcc97ffedc5bc6630e68e4b548d4b01a9e, reputation PeerReputation 149, validated? true, disconnected? false, client: Geth/v1.10.21-stable-55d8eac8/linux-amd64/go1.18, connection 539126146, enode enode://3f88068013a17437b7fae2f155894a33dd94beb55b9d82c667123c6431d4967b52a9ce155f0c44dd2dcb405cbef5addcc97ffedc5bc6630e68e4b548d4b01a9e@44.198.36.250:30303`

This is always followed by multiple lines of:
`[DEBUG] 2022-08-18 15:15:00.126 [nioEventLoopGroup-3-1] EthProtocolManager - Ignoring message received from unknown peer connection: NettyPeerConnection{nodeId=0x3f88068013a17437b7fae2f155894a33dd94beb55b9d82c667123c6431d4967b52a9ce155f0c44dd2dcb405cbef5addcc97ffedc5bc6630e68e4b548d4b01a9e, clientId=Geth/v1.10.21-stable-55d8eac8/linux-amd64/go1.18, caps=eth/66, snap/1}`

And in all cases checked a timeout happened before the BREACH_OF_PROTOCOL:
[...]

PR #4377, opened 2022-09-09
Fix banning (static peers) because of late response in eth/66

### LINKED ISSUE #4320: Do not ban (static) peers for late responses
## Usecase
(same backstory as #3801) I want to sync my Besu node against a single local (Geth) node, and not connect to the wider internet. The Besu node is in the `trusted-nodes.json` file of Geth, and the Geth node is in the `static-nodes.json` of Besu. 

For completeness sake, my command line is `./besu --data-storage-format=bonsai --p2p-port=30303 --sync-mode=FAST --fast-sync-min-peers=1 --discovery-enabled=false` (same behavior with `X_SNAP`)

## Issue
(Possibly related to #4294)
When using the command line above, I eventually get a situation where Geth takes _slightly_ too long to respond, e.g. 5.1 seconds. Besu has already considered the request to be timed out, and severely punishes the peer by disconnecting right away with `BREACH_OF_PROTOCOL`. It also never reconnects to it (which is unlike Geth/Nethermind, at least for static nodes).

The offending bit of code is https://github.com/hyperledger/besu/blob/113bd54c6b0ff748a3ed23f2e89fa31a4cb73795/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/RequestManager.java#L70-L93

I reproduced this issue on Besu/Geth/Nethermind by slightly [modifying Geth ](https://github.com/ethereum/go-ethereum/compare/master...pietjepuk2:go-ethereum:sometimes-delay-response-trusted-nodes) to delay the response to the Nth GetHeaders request.

## Other clients behavior
- Geth doesn't disconnect on late replies. It does requests sequentially it seems, and just waits with sending a new request until it gets a response (waits more than 10 seconds at least). If I delay the response with a minute, Geth just disconnects the peer after ~40-50 seconds before reconnecting to it a few seconds later.
- Nethermind is a bit weird in that it continuously disconnects and reconnects every minute or so. It doesn't punish a late response (>5 seconds), but either way always reconnects.
- (Not entirely sure). Late responses were probably not punished by Besu with Eth65 (i.e. no `RequestId`). I couldn't quite understand the code (or test it), but my gut feeling is that it just skipped processing the late response or handled it anyway.

## Suggested solutions
1. Just `LOG.debug("Request ID incorrect (BREACH_OF_PROTOCOL), disconnecting peer {}", peer);` but don't disconnect the peer. Maybe decrease peer reputation.
2. Keep track of timed-out request ids, and if response is in one of them -> ignore. If it's not -> disconnect.

## Sidenotes
- Even after disconnecting, whether it be "BREACH_OF_PROTOCOL" or because the reputation dropped too low, I would expect Besu to reconnect to static nodes regardless. Other clients seem to do so anyway.
- I have at least one logging of Besu receiving a response _barely_ within 5 seconds (in `RequestManager`), but `AbstractPeerRequestTask` still timing out on it ~200 ms later. The processing or delay on Besu's side surely shouldn't be part of the 5 seconds timeout window?

pietjepuk2:
## PR description

- The RequestManager commit that lessens the penalty should be somewhat straightforward. A late response is typically already punished with a timeout before (too many of which and we disconnect). Instead of immediately considering a late response a breach of protocol, we just consider it useless (again, too many of which and we disconnect).

- The not banning of maintained/static peers was a bit more difficult. I'm not sure if this is the right^TM way to do it, but a Peer object has no notion of it being static/maintained. And static nodes are also not instantiated in a single call, they are combined with bootnodes. So adding a `isStatic` field to EthPeer also seemed more trouble than it was worth. Therefore I settled on this solution, but there might be a better/simpler/more elegant one.

Feedback more than welcome!

## Fixed Issue(s)

Fixes #4320
Related to #4294

## Documentation

## Changelog

## TODO before merging
- [x] Elaborate RequestManager commit message
- [x] Elaborate in PeerDenylistManager commit message
- [x] Fix test cases PeerDenylistManager 
- [x] Add test case (with maintained/static peer) for PeerDenylistManager 
- [x] run spotlessApply to fix formatting issues

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/PeerDenylistManager.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/PeerDenylistManager.java
@@ -14,6 +14,7 @@
  */
 package org.hyperledger.besu.ethereum.p2p.network;
 
+import org.hyperledger.besu.ethereum.p2p.peers.MaintainedPeers;
 import org.hyperledger.besu.ethereum.p2p.permissions.PeerPermissionsDenylist;
 import org.hyperledger.besu.ethereum.p2p.rlpx.DisconnectCallback;
 import org.hyperledger.besu.ethereum.p2p.rlpx.connections.PeerConnection;
@@ -36,8 +37,12 @@ public class PeerDenylistManager implements DisconnectCallback {
 
   private final PeerPermissionsDenylist denylist;
 
-  public PeerDenylistManager(final PeerPermissionsDenylist denylist) {
+  private final MaintainedPeers maintainedPeers;
+
+  public PeerDenylistManager(
+      final PeerPermissionsDenylist denylist, final MaintainedPeers maintainedPeers) {
     this.denylist = denylist;
+    this.maintainedPeers = maintainedPeers;
   }
 
   @Override
@@ -46,8 +51,15 @@ public class PeerDenylistManager implements DisconnectCallback {
       final DisconnectReason reason,
       final boolean initiatedByPeer) {
     if (shouldBlock(reason, initiatedByPeer)) {
-      LOG.debug("Added peer {} to peer denylist for reason {}", connection, reason.name());
-      denylist.add(connection.getPeer());
+      if (maintainedPeers.contains(connection.getPeer())) {
+        LOG.debug(
+            "Skip adding maintained peer {} to peer denylist for reason {}",
+            connection,
+            reason.name());
+      } else {
+        LOG.debug("Added peer {} to peer denylist for reason {}", connection, reason.name());
+        denylist.add(connection.getPeer());
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/PeerDenylistManager.java
```diff
@@ -46,8 +50,12 @@ public void onDisconnect(
       final DisconnectReason reason,
       final boolean initiatedByPeer) {
     if (shouldBlock(reason, initiatedByPeer)) {
-      LOG.debug("Added peer {} to peer denylist for reason {}", connection, reason.name());
-      denylist.add(connection.getPeer());
+      if (maintainedPeers.contains(connection.getPeer())) {
+        LOG.debug("Skip adding maintained peer {} to denylist for reason {}", connection, reason.name());
```
pietjepuk2, line 54: the other log message has "to _peer_ denylist", not "to denylist". Or just different wording in general, as this is a bit awkward.

jflo: I don't see anywhere that maintainedPeers gets called outside of a test class. It is possible this is dead code. You seem to be under the impression that this is where static peers are maintained, which seems reasonable to me... I just don't see any way to prove it.

jflo (approved):
This looks good, and should help non-static peers as well.

macfarla:
@jflo one place maintainedPeers is used is in RlpxAgent - maintained peers are allowed to push the number of connections over maxPeers https://github.com/hyperledger/besu/blob/16922cac83f27327c4979bd8def12679086074a8/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/rlpx/RlpxAgent.java#L239

macfarla:
and @pietjepuk2 you'll need to run spotlessApply to fix some formatting issues :)

PR #4386, opened 2022-09-13
Replace boolean return with BlockImportResult object

### LINKED ISSUE #4350: FIx duplicate log message about importing the same block
### Description
Message from @fab-10:

> Strange thing that I am seeing after the update to 22.7.1, but not sure it was present before, because the log has changed, basically we are importing twice the same block, and this should not happen

```
{"timestamp":"2022-08-26T15:41:41,382","level":"INFO","thread":"EthScheduler-Workers-2","class":"PersistBlockTask","message":"Imported #15,416,217 / 61 tx / 0 om / 5,733,156 (19.1%) gas / (0x1ed30511c454cb73f88b12eb6675e5618a90fde673a770db30caf0bd8d84f60d) in 3.379s. Peers: 25","throwable":""}
{"timestamp":"2022-08-26T15:41:41,382","level":"INFO","thread":"EthScheduler-Workers-3","class":"PersistBlockTask","message":"Imported #15,416,217 / 61 tx / 0 om / 5,733,156 (19.1%) gas / (0x1ed30511c454cb73f88b12eb6675e5618a90fde673a770db30caf0bd8d84f60d) in 1.038s. Peers: 25","throwable":""}
{"timestamp":"2022-08-26T15:41:41,382","level":"INFO","thread":"EthScheduler-Workers-3","class":"BlockPropagationManager","message":"Imported 1 pending blocks: [15416217]","throwable":""}
{"timestamp":"2022-08-26T15:41:41,382","level":"INFO","thread":"EthScheduler-Workers-2","class":"BlockPropagationManager","message":"Imported 1 pending blocks: [15416217]","throwable":""}
```

**Expected behavior:**
There should be a check to see if the block is already present.

**Actual behavior:**
~Block seems to be imported twice.~ 
Maintainers confirmed that blocks don't get imported twice. 
As part of this ticket, we will fix the logs so that it doesn't show duplicate log message regarding importing the same block. 

### Versions
22.7.1

Gabriel-Trintinalia:

## PR description
BlockImporter currently returns a `boolean true` when: 

- A  block is successfully imported or 
- the block is already in the blockchain

If the same block is mistakenly tried to be imported more than once, `PersistBlockTask` will log "Imported #" for each try, ignoring that the block is actually imported only once, resulting in misleading logging. 

This PR changes the return type of `BlockImporter.importBlock` to return an object `BlockImportResult` instead of `boolean`.

`BlockImportResult` implements `isImported`, which has the current behaviour of the importBlock: returns `true` if the block is successfully processed or is already in the chain. In addition, it contains a getStatus that, if needed, can be used to distinguish whether the block was processed or was already in the chain. 

PersistBlockTask will look at the getStatus in order to decide what needs to be logged in the cleanup.  

## Fixed Issue(s)
fixes #4350

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockImportResult.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2: Contributors to Hyperledger Besu

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockImportResult.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.mainnet;
+
+public class BlockImportResult {
+
+  public enum BlockImportStatus {
+    TRUE,
+    FALSE,
```
macfarla, line 21: so TRUE is "freshly imported" and FALSE is "invalid block"? do we need that nuance in the result options?

Gabriel-Trintinalia: Yes, that is correct. Happy to change the names to something more meaningful.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockImportResult.java
Gabriel-Trintinalia: Did I miss anything here?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockImportResult.java
Gabriel-Trintinalia: Replaced the values with more meaningful names

macfarla (approved):
LGTM

PR #4387, opened 2022-09-13
Revise Memory Expansion

shemnon:
## PR description

Move from a perfect fit memory expansion model to an exponential expansion model.

## Documentation

## Changelog


garyschulte (approved):
Locally seeing 24 -> 5.4  :+1:

:ship:

jflo (approved):

shemnon:
@garyschulte count your decimal places.

garyschulte:
`time` included in jvm startup presumably.  😉

holiman:
Shouldn't you have changed this too? 
```
  @Override
  public String toString() {
    return Bytes.wrap(memBytes).toHexString();
  }
```
-- so that it returns only the `activeMemory` area, not the unued-but-allocated data at the end? 
I don't know if this is used anywhere, but if it's e.g. used in the json output for tracing, then it needs to be fixed.

PR #4388, opened 2022-09-13
RLP zero encoding fix.

### LINKED ISSUE #4284: RLP encoding of 0 (zero) as 0x80
### Description
I have created a PR to fix a problem with Nethermind clients (https://github.com/hyperledger/besu/pull/4283), where Besu cannot decode the Hello message sent by Nethermind because it contains a capability for the witness protocol with version 0. This 0 was encoded 0x80, which is the encoding for a zero length string but in this case needs to be interpreted as the integer value 0.
I learned that this is (probably) the right way to encode the value 0.

The following should be done:
- research where 0x80 is the right way to encode the value 0 and, if necessary, fix the RLP encoding in Besu
- fix RLP decoding where necessary

mark-terry:

## PR description
Updates the RLPOutput encoding to encode 0 as 0x80

## Fixed Issue(s)

Fixes #4284 

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Additions and Improvements
 - Allow free gas networks in the London fee market [#4061](https://github.com/hyperledger/besu/issues/4061)
 - Upgrade besu-native to 0.6.0 and use Blake2bf native implementation if available by default [#4264](https://github.com/hyperledger/besu/pull/4264)
+- Improved RLP processing to handle zero-length string as 0x80 [#4388](https://github.com/hyperledger/besu/issues/4388).
```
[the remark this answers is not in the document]
mark-terry: I've updated the changelog line to include both relevant PRs.

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPOutput.java
```diff
@@ -187,7 +188,7 @@ default void writeLong(final long l) {
    *     {@code b < 0} or {@code b > 0xFF}.
    */
   default void writeUnsignedByte(final int b) {
-    writeBytes(Bytes.of(b));
+    processZeroByte(Long.valueOf(b), a -> writeBytes(Bytes.of(b)));
```
pinges, line 191: As b is an integer, could Bytes.of(b) be more than one byte long? With the other writeUnsignedXXX we are making sure that there aren't more bytes than we want (e.g. Bytes.ofUnsignedShort in writeUnsignedShort). Do we have to do something similar here as well?

pinges (commented):
Please have a look at my comment.

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPOutput.java
mark-terry: Within Bytes.of() there's a check to confirm whether the value fits within a byte, which throws an IllegalArgumentException if the value would be truncated.

Issue #4403, opened 2022-09-15
Investigate RocksDB issue


iamhsk:
May cause sync and timeouts

Issue #4402, opened 2022-09-15
Bonsai concurrency issue with Nimbus


iamhsk:

siladu:
Duplicate of https://github.com/hyperledger/besu/issues/4372?

macfarla:
closed as duplicate

Issue #4372, opened 2022-09-08
Concurrency issue with Nimbus/Besu combination


garyschulte:
### Description
The nimbus behavior in 2.8.2 (asking for a proposal on each forkchoice) uncovered a concurrency issue with Besu block proposals:

Report from @jclapis:

Hi folks, I'm using Besu v22.7.1 with Nimbus v22.8.1 on Goerli/Prater and ran into an issue that @garyschulte recommended I provide here, so you can gauge whether or not this PR would fix it.

Yesterday my Besu node ran into a critical error and it stopped importing new blocks:
```
{"log":"2022-09-02 19:04:01.036+00:00 | nioEventLoopGroup-3-5 | INFO  | BackwardSyncContext | Backward sync failed (Did not receive a headers for hash 0xc33bdbbe21f95dad81361ffba139b39d983dff362e591f744aadce9540b7eb29). Current Peers: 66. Retrying in 5000 milliseconds...\n","stream":"stdout","time":"2022-09-02T19:04:01.036719219Z"}
{"log":"2022-09-02 19:04:06.149+00:00 | nioEventLoopGroup-3-10 | INFO  | BackwardSyncContext | Backward sync failed (Did not receive a headers for hash 0x7e086e87786bf4ad5da6398a939a98e417ebf4fc8acfdd3ebfe875771124d145). Current Peers: 66. Retrying in 5000 milliseconds...\n","stream":"stdout","time":"2022-09-02T19:04:06.149689208Z"}
{"log":"2022-09-02 19:04:07.245+00:00 | vert.x-eventloop-thread-7 | WARN  | EngineExchangeTransitionConfiguration | not called in 120 seconds, consensus client may not be connected\n","stream":"stdout","time":"2022-09-02T19:04:07.246771771Z"}
{"log":"2022-09-02 19:04:11.248+00:00 | nioEventLoopGroup-3-10 | INFO  | BackwardSyncContext | Backward sync failed (Did not receive a headers for hash 0x7e086e87786bf4ad5da6398a939a98e417ebf4fc8acfdd3ebfe875771124d145). Current Peers: 66. Retrying in 5000 milliseconds...\n","stream":"stdout","time":"2022-09-02T19:04:11.249239501Z"}
[...]

Issue #4250, opened 2022-08-10
Critical Exception Processing Transaction


CaelorS:
### Description
Hi, 

I have an issue sychronizing besu with the network. I launch an x-snap synchronisation, and here is what I get, after a few day. I'm stuck here... 

Any help would be appreciated. Thank ! 

```août 10 18:21:23 jeremy-NUC10i7FNK besu[42974]: 2022-08-10 18:21:23.149+02:00 | nioEventLoopGroup-3-7 | INFO  | SyncTargetManager | Found common ancestor with peer Peer 0x68dff20f06112b6ee8... PeerReputation 100, validated? true, disconnected? false at block 15295479
août 10 18:21:39 jeremy-NUC10i7FNK besu[42974]: 2022-08-10 18:21:39.922+02:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
août 10 18:21:41 jeremy-NUC10i7FNK besu[42974]: 2022-08-10 18:21:41.343+02:00 | Timer-0 | WARN  | RlpxAgent | Attempt to connect to peer with no listening port: enode://92ae467ca2719a0ca683d103580b78f3d0d0d58be0f8313ebe268da05e94ef0f2d91aa7834b0ec260e4b121a04a4bbf428d31d2b86bd34df3141ee56ccd63b01@xx.xx.xxx.xx:0?discport=30853
août 10 18:21:41 jeremy-NUC10i7FNK besu[42974]: 2022-08-10 18:21:41.737+02:00 | Timer-0 | INFO  | DNSTimerTask | Refreshing DNS records with enrtree://AKA3AM6LPBYEUDMVNU3BSVQJ5AD45Y7YPOHJLEF6W26QOE4VTUDPE@all.mainnet.ethdisco.net
août 10 18:21:43 jeremy-NUC10i7FNK besu[42974]: 2022-08-10 18:21:43.745+02:00 | EthScheduler-Services-5 (importBlock) | ERROR | MainnetTransactionProcessor | Critical Exception Processing Transaction
août 10 18:21:43 jeremy-NUC10i7FNK besu[42974]: org.hyperledger.besu.ethereum.trie.MerkleTrieException: Unable to load trie node value for hash 0xf83f41ea3851008549a6dc38424b5253c2edcf2a3e7135c160b84f614d5e3baf location 0x
[...]

PR #4409, opened 2022-09-16
Bonsai snapshot worldstate 

garyschulte:
## PR description
This PR adds support for using rocksdb snapshot transactions as the basis for a bonsai worldstate.  This gives us the ability to have an isolated copy of the world state for concurrent operations like block production, block validation, transaction validation, historical state queries, etc.

The notion is to merge this capability into main, and start replacing the uses of BonsaiLayeredWorldState and direct uses of BonsaiInMemoryWorldState when we need isolated and/or concurrent operations on the worldstate.

This is a strictly additive PR and does not change existing implementations.  Uses of snapshot worldstates will be useful to resolve issues such as #4372 #4250 #4199 #4151 and other bonsai concurrency issues.

Changes in support of snapshot worldstate include:
* moving (back) to OptimisticTransactionDB, so we can mutate the snapshot transactions
* creating `BonsaiWorldStateKeyValueStorage.BonsaiUpdater` interface extending `WorldStateStorage.Updater`, so we can have a snapshot specific Updater implementation
* Addition of SnappableStorage, SnappedStorage to plugin-api in order to support/expose the snapshot behavior 

## Fixed Issue(s)

relates to #4402 
relates to #4403 

## Documentation

## Changelog

### DIFF
--- a/config/src/main/java/org/hyperledger/besu/config/GenesisAllocation.java
+++ b/config/src/main/java/org/hyperledger/besu/config/GenesisAllocation.java
@@ -16,6 +16,7 @@ package org.hyperledger.besu.config;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
@@ -30,6 +31,10 @@ public class GenesisAllocation {
 
   public String getAddress() {
     return address;
+  }
+
+  public Optional<String> getPrivateKey() {
+    return Optional.ofNullable(JsonUtil.getString(data, "privatekey", null));
   }
 
   public String getBalance() {
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
@@ -4,9 +4,10 @@ import org.hyperledger.besu.ethereum.core.MutableWorldState;
 import org.hyperledger.besu.plugin.services.storage.SnappableKeyValueStorage;
 
 /**
- * This class takes a snapshot of the worldstate as the basis of a mutable worldstate. It is not
- * able to commit or persist however. This is useful for async blockchain opperations like block
- * creation and/or point-in-time queries.
+ * This class takes a snapshot of the worldstate as the basis of a mutable worldstate. It is able
+ * to commit/perist as a trielog layer only. This is useful for async blockchain opperations like block
+ * creation and/or point-in-time queries since the snapshot worldstate is fully isolated from the main
+ * BonsaiPersistedWorldState.
  */
 public class BonsaiSnapshotWorldState extends BonsaiInMemoryWorldState {
   //  private static final Logger LOG = LoggerFactory.getLogger(BonsaiSnapshotWorldState.class);
@@ -32,8 +33,8 @@ public class BonsaiSnapshotWorldState extends BonsaiInMemoryWorldState {
 
   @Override
   public MutableWorldState copy() {
-    // TODO: we are transaction based, so perhaps we return a new transaction based worldstate...
-    //       for now lets just return ourselves
+    // TODO: find out if we can stack transaction based snapshots, so perhaps we return a new transaction based
+    //  worldstate.  For now we just return ourself rather than a copy.
     return this;
   }
 }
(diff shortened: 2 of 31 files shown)


matkt:
Excellent work I will review it on Monday

matkt (commented):
I will continue the review but i already share this

config/src/main/java/org/hyperledger/besu/config/GenesisAllocation.java
```diff
@@ -32,6 +33,10 @@ public String getAddress() {
     return address;
   }
 
+  public Optional<String> getPrivateKey() {
+    return Optional.ofNullable(JsonUtil.getString(data, "privatekey", null));
```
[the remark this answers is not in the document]
garyschulte: Yes, it is for testing so we have the ability to use the genesis allocations for test accounts

matkt (approved):
Seems to be ok. waiting for the next PR related to snapshot

garyschulte:
IMO, this is ready to merge to main to clear the way for subsequent PRs that implement bonsai snapshots.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
```diff
@@ -0,0 +1,92 @@
...
+      final BonsaiWorldStateArchive archive,
+      final BonsaiSnapshotWorldStateKeyValueStorage snapshotWorldStateStorage) {
+    super(archive, snapshotWorldStateStorage);
+    this.accountSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.accountStorage;
+    this.codeSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.codeStorage;
+    this.storageSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.storageStorage;
+    this.trieBranchSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.trieBranchStorage;
+  }
+
+  public static BonsaiSnapshotWorldState create(
```
gezero, line 48: Is there a potential race with the storages being changed while the snapshots are generated?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBSnapshotTransaction.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  @Override
+  public void put(final byte[] key, final byte[] value) {
+    try (final OperationTimer.TimingContext ignored = metrics.getWriteLatency().startTimer()) {
+      snapTx.put(columnFamilyHandle, key, value);
+    } catch (final RocksDBException e) {
+      if (e.getMessage().contains(NO_SPACE_LEFT_ON_DEVICE)) {
+        LOG.error(e.getMessage());
+        System.exit(0);
```
gezero, line 92: System.exit feels too brutal to me. Is there any chance this can corrupt db for instance in case an ongoing update is happening in parallel?  Maybe we don't access db from different threads.

gezero (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
garyschulte: There is some risk of that theoretically, but considering we write to rocksdb in batch commits and the snapshots are essentially atomic, this is probably very very minimal chance in practice.  There is certainly far less concurrency risk than our current strategy where the underlying persisted state can change underneath a BonsaiInMemoryWorldState or BonsaiLayeredWorldState

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBSnapshotTransaction.java
garyschulte: Good question.  This is the same behavior as RocksDBTransaction.  The right course of action isn't really clear since we do not have any disk space to work with.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
garyschulte: It seems reasonable to implement locking in BonsaiPersistedWorldState so that we do not commit a batch write (via calculateRootHash) and take a snapshot concurrently.  It should be a very short lock for getting a snapshot, and I delaying getting a snapshot while waiting on a lock would normally not be significant.

I will add this in one of the subsequent snapshot PRs and add tests :+1

Issue #4400, opened 2022-09-15
Besu Execution timeout/missed attestations 


non-fungible-nelson:
### Description
Basic story, details to come - essentially Besu is constrained in execution time and is hitting CL timeouts that are causing it to miss attestations. This is our working theory, with supporting evidence. The team will update with more below.

ColinCampbell:
> Is this potentially the same issue as https://github.com/hyperledger/besu/issues/4398?

I'm seeing timeouts and missed attestations without any JSON-RPC errors or exceptions, fwiw

non-fungible-nelson:
Anecdotal evidence in Discord of users experiencing this issue with very high speced machines. Likely only tangentially tied to resource usage.

fiah84:
I'm missing attestations, but I see no indication of besu not having enough resources. I have dedicated hardware (6c ryzen, 32gb, 2tb 970 evo plus) that is fast enough

I filtered out the following topics from the debug log to see if I could make sense of the rest:
GetPooledTransactionsFromPeerTask
RlpxAgent
AbstractPeerConnection
EthPeer
RecursivePeerRefreshState
EthProtocolManager
DeFramer
HandshakeHandlerOutbound 

[debug.txt](https://github.com/hyperledger/besu/files/9579054/debug.txt)

torfbolt:
Could this lead to Teku logging empty slots, as mentioned in https://github.com/ConsenSys/teku/issues/6205 ?

sdherr:
I'm not sure I'm in the right place, but I'll paste my logs. Things were executing normally until this happened:

```
Sep 15 17:01:25 node besu[1137]: java.lang.RuntimeException: java.lang.IllegalStateException: Expected to create code, but the code exists.  Address=0x9b873b45b23965cc69b81b5596e4bf4dfe31b6ba
Sep 15 17:01:25 node besu[1137]:         at org.hyperledger.besu.ethereum.bonsai.BonsaiWorldStateArchive.getMutable(BonsaiWorldStateArchive.java:206)
Sep 15 17:01:25 node besu[1137]:         at org.hyperledger.besu.consensus.merge.blockcreation.MergeCoordinator.forwardWorldStateTo(MergeCoordinator.java:362)
Sep 15 17:01:25 node besu[1137]:         at org.hyperledger.besu.consensus.merge.blockcreation.MergeCoordinator.setNewHead(MergeCoordinator.java:351)
[...]

PR #4410, opened 2022-09-17
Fix Besu StacklessClosedChannelException errors and resulted Timeout errors in CL clients

### LINKED ISSUE #4398: JsonRpcExecutorHandler | Error streaming JSON-RPC response
I started receiving the below error after the merge

execution_1           | 2022-09-15 14:03:47.422+00:00 | vert.x-worker-thread-1 | ERROR | JsonRpcExecutorHandler | Error streaming JSON-RPC response
execution_1           | io.netty.channel.StacklessClosedChannelException
execution_1           | 	at io.netty.channel.AbstractChannel$AbstractUnsafe.write(Object, ChannelPromise)(Unknown Source)

Frequency: every few minutes 

besu besu/v22.7.2/linux-x86_64/openjdk-java-11 via eth-docker 
ubuntu 22.04
docker : 20.10.12

ahamlat:

## PR description
StacklessClosedChannelException are thrown when some Eth calls, like eth_syncing or eth_getBlockByHash have to wait for the engine_newPayloadV1 call to finish. As the latter takes sometimes more than 1 second to execute depending on the block size, transactions type and hardware setup, the CL closes the connection [(in the case of Lighthouse)](https://github.com/sigp/lighthouse/pull/3470) because the timeout is set to 1 second on these calls.

In the screenshot below, we can see that 1 eth_getBlockByHash call and 2 eth_syncing calls are waiting for engine_newPayloadV1 call to finish.

image

The Eth calls should't execute sequentially with the other Engine API calls, this fix will make several requests to execute concurrently on port 8551. 
@garyschulte please confirm that the 'Engine" will still execute sequentially even with this new configuration.

You can find lighthouse logs [before](https://gist.github.com/ahamlat/8ccdaf24c6524d377fdfff973fa5307b) and [after](https://gist.github.com/ahamlat/aff44be042e57a72729ee693d446eabf) the fix.
## Fixed Issue(s)

fixes #4398 and #4400 
## Documentation

## Changelog

- [x ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Upgrade besu-native to 0.6.0 and use Blake2bf native implementation if available by default [#4264](https://github.com/hyperledger/besu/pull/4264)
 
 ### Bug Fixes
-
+- Fix StacklessClosedChannelException in Besu and resulted timeout errors in CL clients ([#4398](https://github.com/hyperledger/besu/issues/4398), [#4400](https://github.com/hyperledger/besu/issues/4400))
 
 ## 22.7.2
 ### Besu 22.7.2 is a recommended release for the Merge and Mainnet users. 22.7.1 remains Merge-ready. This release provides additional robustness before the Merge with some fixes and improvements in sync, peering, and logging.
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcService.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcService.java
@@ -444,7 +444,8 @@ public class JsonRpcService {
                       authenticationService.get(),
                       config.getNoAuthRpcApis()),
                   rpcMethods),
-              tracer));
+              tracer),
+          false);
     } else {
       mainRoute.blockingHandler(
           HandlerFactory.jsonRpcExecutor(
@@ -452,7 +453,8 @@ public class JsonRpcService {
                   new TimedJsonRpcProcessor(
                       new TracedJsonRpcProcessor(new BaseJsonRpcProcessor()), requestTimer),
                   rpcMethods),
-              tracer));
+              tracer),
+          false);
     }
 
     if (authenticationService.isPresent()) {


kayagoban:
I installed a Besu patched with this PR and it has had little effect on my missed attestations.  Correctly voted head percent still is floating between 58% and 81%.  Before merge I was at 98%.  But maybe this helps fix other problems.  Good luck!

ahamlat:
@kayagoban thanks for your feedback, what is your hardware setup and your CL client ? do you see see timeout errors in CL side and could you share Besu logs to check block processing time ?

siladu (approved):
LGTM, as long as the message ordering property is maintained for forkChoiceUpdated as per https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#message-ordering

garyschulte (approved):
Confirmed that this fixes the eth_ concurrency issue and keeps the engine_ api calls ordered and synchronous :+1:

As suggeted by @jfo we should do additional websocket engine api testing with nimbus also to ensure that concurrent websocket requests are handled correctly when nimbus is using websockets for the engine api

kayagoban:
> 

I’m running Besu/Prysm.  No timeout errors on CL side.  Maybe I can make a new issue with what logs I do have and my symptoms.

ahamlat:
@kayagoban Yes, please, we think we have different use cases around missing attestations. Please, provide your hardware setup, Besu and Prysm logs.

ibhagwan:
@garyschulte, running this patch on mainnet with nimbus websockets, just missed an attestation after about two hours of running the patched version.

garyschulte:
> @garyschulte, running this patch on mainnet with nimbus websockets, just missed an attestation after about two hours of running the patched version.

Thanks for testing.  Other than continuing to miss attestations did nimbus/besu work normally otherwise? What percentage of attestations were missed before and after?

ibhagwan:
> > @garyschulte, running this patch on mainnet with nimbus websockets, just missed an attestation after about two hours of running the patched version.
> 
> Thanks for testing. Other than continuing to miss attestations did nimbus/besu work normally otherwise? What percentage of attestations were missed before and after?

Yes, other than the missed attestation it’s working normally, I can’t tell if there is a difference yet as prior it was also missing roughly 1-2 attestations/hr.

It will be interesting to see if effectiveness improves, before the patch it would hover between 90-95% due to the occasional sub optimal inclusion distance, would be interesting to see after 12hrs if this improves.

stefanlsd:
I'm receiving missed attestations with a Prysm / Besu configuration.    Relevant from Prysm might be the repeating message

time="2022-09-19 07:44:23" level=info msg="Subscribed to topic" prefix=sync topic="/eth2/4a26c58b/beacon_attestation_2/ssz_snappy"
time="2022-09-19 07:44:33" level=error msg="Could not handle p2p pubsub" error="could not process block: could not validate new payload: timeout from http.Client: received an undefined ee error" prefix=sync topic="/eth2/4a26c58b/beacon_block/ssz_snappy"

ibhagwan:
Sadly I can report that this PR did not fix missed attestations with nimbus, roughly same amount of attestations are missed and effectiveness won’t go beyond 94-95% (was 97-98% pre-merge).

siladu:
Missed attestations despite this fix being tracked here https://github.com/hyperledger/besu/issues/4400

PR #4411, opened 2022-09-19
Less strict engine QoS timer

### LINKED ISSUE #4404: Reduce noise from Prysm & Nimbus warning messages

daniellehrner:

## PR description

Resets engine QoS timer with every call to the engine API instead of only when `ExchangeTransitionConfiguration` is called. `ExchangeTransitionConfiguration` mismatch will only submit a debug log not a warning anymore.

## Fixed Issue(s)
fixes #4404

## Documentation

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
@@ -21,7 +21,6 @@ import org.hyperledger.besu.consensus.merge.MergeContext;
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.ethereum.ProtocolContext;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.JsonRpcRequestContext;
-import org.hyperledger.besu.ethereum.api.jsonrpc.internal.QosTimer;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.ExecutionEngineJsonRpcMethod;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.EngineExchangeTransitionConfigurationParameter;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.JsonRpcResponse;
@@ -46,21 +45,11 @@ public class EngineExchangeTransitionConfiguration extends ExecutionEngineJsonRp
       Difficulty.fromHexString(
           "0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffc00");
 
-  static final long QOS_TIMEOUT_MILLIS = 120000L;
-
-  private final QosTimer qosTimer;
-
   public EngineExchangeTransitionConfiguration(
-      final Vertx vertx, final ProtocolContext protocolContext) {
-    super(vertx, protocolContext);
-    qosTimer =
-        new QosTimer(
-            vertx,
-            QOS_TIMEOUT_MILLIS,
-            lastCall ->
-                LOG.warn(
-                    "not called in {} seconds, consensus client may not be connected",
-                    QOS_TIMEOUT_MILLIS / 1000L));
+      final Vertx vertx,
+      final ProtocolContext protocolContext,
+      final EngineCallListener engineCallListener) {
+    super(vertx, protocolContext, engineCallListener);
   }
 
   @Override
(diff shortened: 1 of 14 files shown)


fab-10 (approved):
LGTM, just a question, 
why have not you done the common `engineCallListener.executionEngineCalled()` call in the parent class? in that wat it could be inherited easily by future engine apis

daniellehrner:
> LGTM, just a question, why have not you done the common `engineCallListener.executionEngineCalled()` call in the parent class? in that wat it could be inherited easily by future engine apis

@fab-10 To make it possible to unit test. There is no acceptance test to test it and not really any other good way except for unit tests. As this timer will most probably be removed in future hard forks, I thought it is a justified tradeoff.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadTest.java
```diff
@@ -116,6 +121,7 @@ public void shouldReturnValid() {
     assertThat(res.getLatestValidHash().get()).isEqualTo(mockHeader.getHash());
     assertThat(res.getStatusAsString()).isEqualTo(VALID.name());
     assertThat(res.getError()).isNull();
+    verify(engineCallListener, times(1)).executionEngineCalled();
```
garyschulte, line 124: 👍

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
```diff
@@ -70,8 +59,7 @@ public String getName() {
 
   @Override
   public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext) {
-    // update our QoS "last call time"
-    getQosTimer().resetTimer();
+    engineCallListener.executionEngineCalled();
```
garyschulte, line 62: Minor point, we can DRY these up by adding this call into `ExecutionEngineJsonRpcMethod.response()` method rather than having it each engine api endpoint.

garyschulte (approved):
Looks great.  There is a non-blocking DRY opportunity, but otherwise 🚢

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineQosTimerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        z -> {
+          try {
+            // once on construction:
+            verify(spyTimer, times(1)).resetTimer();
+          } catch (Exception ex) {
+            ctx.fail(ex);
+          }
+          // should warn
+          verify(spyEngineQosTimer, times(1)).logTimeoutWarning();
+          async.complete();
```
garyschulte, line 93: I know these were already outside the try, but these should move into the try just in case the verify throws for some reason.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineQosTimerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        z -> {
+          try {
+            // once on construction, once on call:
+            verify(spyTimer, times(2)).resetTimer();
+          } catch (Exception ex) {
+            ctx.fail(ex);
+          }
+          // should not warn
+          verify(spyEngineQosTimer, never()).logTimeoutWarning();
+          async.complete();
```
garyschulte, line 68: same, move into try block

garyschulte (changes requested):
The failing unit test for EngineQosTimerTest implies something outside the try block is throwing before the async can complete.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineExchangeTransitionConfiguration.java
daniellehrner: The reason why I did not do that, was because of the unit tests. When I move it there, I am not sure if I can still test the correct behavior with the unit tests. Do you think it would still be possible?

Fabio had the same question:
https://github.com/hyperledger/besu/pull/4411#issuecomment-1251270287

garyschulte:
⛴️

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineQosTimerTest.java
daniellehrner: Done

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineQosTimerTest.java
daniellehrner: Done

garyschulte (approved):
🚢

Issue #4401, opened 2022-09-15
Post-Merge some blocks contain 0 transactions 


non-fungible-nelson:
### Description
Issue stub - more details to come. Post-Merge we are seeing some blocks with 0 included transactions. 

### Acceptance Criteria
*  Blocks contain transactions 

### Steps to Reproduce (Bug)
1. Observe logs post-merge for included transactions

macfarla:
https://discord.com/channels/905194001349627914/938504958909747250/1009795829625786418 
Hello! I have vouch + teku + besu in prater, and I see that sometimes there are no txs in my proposed blocks. Example: https://prater.beaconcha.in/block/3689038 and https://prater.beaconcha.in/block/3649149
Besu logs show nothing interesting about that: https://pastebin.com/ZcTqHatn
Teku logs also nothing related (as I can understand it): https://pastebin.com/vRCGM6uz   (about fee address logs - we set in vouch and it works, we can see it in a new block, dunno why teku complains about that constantly)
How to research this problem?

macfarla:
^ empty blocks on prater do happen normally. 

From discord thread https://discord.com/channels/697535391594446898/1009446584611774535
there are a couple cases where besu might produce empty blocks.

When asked to produce a block, Besu starts with an empty block proposal candidate, just so it has something to respond to the CL with if it is not done building a block by the time the CL requests it.  This historically has happened with nimbus because nimbus was not waiting the prescribed 500 milliseconds between asking for a block to be built and requesting the block.  I don't think prysm has had this issue.

The other case is that it is possible non of the transactions in besu's pool are eligible for inclusion, either because of basefee or because of nonce.  in 22.7.2 besu is changing its transaction eviction policy to ensure that we prioritize future nonce transactions for eviction ahead of lower value but executable transactions.

If you query the pending transactions endpoint, you can spot check some of the transactions in the pool to check whether they are executable or not

macfarla:
https://discord.com/channels/905194001349627914/1009795829625786418/1019880263049822208
The same as above, but in mainnet after the merge
at least 2 empty blocks
https://beaconcha.in/slot/4700074
https://beaconcha.in/slot/4700312
 the second one after restart teku/besu

macfarla:
we are investigating this - working theory is that this is related to resource constraints - suggestion is to try increasing JVM memory as per https://besu.hyperledger.org/en/latest/public-networks/how-to/configure-jvm/manage-memory/

EvilJordan:
[...]

PR #4417, opened 2022-09-21
Make transaction pool limits for sender based on pool size

garyschulte:
## PR description
Limit the portion of the transaction pool that a single sender can occupy to a percentage
* defaults to 0.1%, or default of 5 per sender with default max txpool size
* moves this config out of BesuCommand and into TransactionPoolOptions

This PR will see the default txpool configuration restrict the number of transactions from a single sender in the pool from 64 to 5.  This should decrease the cost of transaction eviction during transaction processing, and yield a transaction pool with a greater percentage of executable transactions for building a block.

## Fixed Issue(s)

relates to #4401 

## Documentation

## Changelog


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1218,16 +1218,6 @@ static class TxPoolOptionGroup {
...
-
-    @Option(
-        names = {"--tx-pool-future-max-by-account"},
-        paramLabel = MANDATORY_INTEGER_FORMAT_HELP,
-        converter = PercentageConverter.class,
-        description =
-            "Maximum per account of currently unexecutable future transactions that can occupy the txpool (default: ${DEFAULT-VALUE})",
-        arity = "1")
-    private final Integer maxFutureTransactionsByAccount =
-        TransactionPoolConfiguration.MAX_FUTURE_TRANSACTION_BY_ACCOUNT;
```
fab-10, line 1230: I think we need to keep this for the moment, otherwise it is a breaking change, what we can do is to give an error if this and the new one are used together, or to say that this has no effect if the new one is used

fab-10 (commented):
Looks good, we just need to found a way to avoid a breaking change

fab-10 (approved):

fab-10:
Please just update the CHANGELOG

garyschulte:
burned in on mainnet along with 4423, 4425

PR #4427, opened 2022-09-22
add engine_preparePayload_debug endpoint

garyschulte:
## PR description

PR to enable debugging block proposals by re-adding an updated version of the deprecated 
`engine_preparePayload` endpoint

*be sure to use the current or a recent parent blockhash or you will DoS bonsai db*

All fields in the parameter are optional, including the parameter itself

example usage:
```
curl --location --request POST 'http://localhost:8551' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2NTkwNDEzMzZ9.4PDdSaG9hFOFR4Th7rEgaKKECsXfz6IPckFRcfSP13o' \
--data-raw '{
    "jsonrpc":"2.0",
    "method":"engine_preparePayload_debug",
    "params":
    [
        {
            "parentHash": "0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3",
            "feeRecipient": "0x0000000000000000000000000000000000000000",
            "timestamp": "0x0",
            "prevRandao": "0x0",
            "withdrawals": []
        }
    ],
    "id":1
}'
```

OR, this usage will use fee recipient 0x00..00, current timestamp, current chain head, "deadbeef" prevrandao, and empty withdrawals:
```
curl --location --request POST 'http://localhost:8551' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2NTkwNDEzMzZ9.4PDdSaG9hFOFR4Th7rEgaKKECsXfz6IPckFRcfSP13o' \
--data-raw '{
    "jsonrpc":"2.0",
    "method":"engine_preparePayload_debug",
    "params":
    [],
    "id":1
}'
```

## Documentation

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
@@ -58,7 +58,7 @@ public enum RpcMethod {
   ENGINE_GET_PAYLOAD_BODIES_BY_HASH_V1("engine_getPayloadBodiesByHashV1"),
   ENGINE_GET_PAYLOAD_BODIES_BY_RANGE_V1("engine_getPayloadBodiesByRangeV1"),
   ENGINE_EXCHANGE_CAPABILITIES("engine_exchangeCapabilities"),
-
+  ENGINE_PREPARE_PAYLOAD_DEBUG("engine_preparePayload_debug"),
   PRIV_CALL("priv_call"),
   PRIV_GET_PRIVATE_TRANSACTION("priv_getPrivateTransaction"),
   PRIV_GET_TRANSACTION_COUNT("priv_getTransactionCount"),
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EnginePreparePayloadDebug.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EnginePreparePayloadDebug.java
@@ -1,5 +1,5 @@
 /*
- * Copyright ConsenSys AG.
+ * Copyright contributors to Hyperledger Besu
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -14,19 +14,23 @@
  */
 package org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.engine;
 
+import static java.util.stream.Collectors.toList;
+
 import org.hyperledger.besu.consensus.merge.blockcreation.MergeMiningCoordinator;
 import org.hyperledger.besu.ethereum.ProtocolContext;
 import org.hyperledger.besu.ethereum.api.jsonrpc.RpcMethod;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.JsonRpcRequestContext;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.ExecutionEngineJsonRpcMethod;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.EnginePreparePayloadParameter;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.WithdrawalParameter;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.JsonRpcError;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.JsonRpcErrorResponse;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.JsonRpcResponse;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.JsonRpcSuccessResponse;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.results.EnginePreparePayloadResult;
+import org.hyperledger.besu.ethereum.core.Withdrawal;
 
-import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 
 import io.vertx.core.Vertx;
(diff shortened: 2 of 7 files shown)


garyschulte:
> Since this is a debug method, would it be simpler to use if it always built on the chainhead? It would be less timing/juggling when it is being used.

yes, yes it would 👍 .  I think we can do the same thing we do with timestamp and default it to the current head if it is otherwise not specified

gfukushima:
Hi Gary, can this be closed since it's been merged as part of another PR?

garyschulte:
> Hi Gary, can this be closed since it's been merged as part of another PR?

this was merged in a speculative RC candidate, and isn't in main yet.

non-fungible-nelson:
can we close or re-status this @garyschulte ?

garyschulte:
IMO we should rebase and merge as it is useful for debugging proposals.  @fab-10 was using a different mechanism to test proposals IIRC, so would be helpful. To get his opinion

fab-10:
> IMO we should rebase and merge as it is useful for debugging proposals. @fab-10 was using a different mechanism to test proposals IIRC, so would be helpful to get his opinion

It make sense to have this endpoint as well, it is an easier and fast method to trigger a payload creation

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EnginePreparePayloadDebug.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright ConsenSys AG.
```
fab-10, line 2: Adjust copyright x3

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/EnginePreparePayloadParameter.java
```diff
@@ -0,0 +1,59 @@
...
+  private final Address feeRecipient;
+  private final Bytes32 prevRandao;
+  private final Optional<Long> timestamp;
+
+  @JsonCreator
+  public EnginePreparePayloadParameter(
+      @JsonProperty("parentHash") final Optional<Hash> parentHash,
+      @JsonProperty("feeRecipient") final Address feeRecipient,
+      @JsonProperty("timestamp") final Optional<UnsignedLongParameter> timestamp,
+      @JsonProperty("prevRandao") final String prevRandao) {
```
fab-10, line 37: Need to be update for withdrawals

fab-10 (changes requested):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/EnginePreparePayloadParameter.java
garyschulte: added :+1:

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
```diff
@@ -58,7 +58,7 @@ public enum RpcMethod {
   ENGINE_GET_PAYLOAD_BODIES_BY_HASH_V1("engine_getPayloadBodiesByHashV1"),
   ENGINE_GET_PAYLOAD_BODIES_BY_RANGE_V1("engine_getPayloadBodiesByRangeV1"),
   ENGINE_EXCHANGE_CAPABILITIES("engine_exchangeCapabilities"),
-
+  ENGINE_PREPARE_PAYLOAD_DEBUG("engine_preparePayload_debug"),
```
gfukushima, line 61: Here is a tricky question, shouldn't this method be in the debug set? Another reason to move that is the way we built the engine_exchange_capabilities might report this method. CL is probably going to ignore it though.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EnginePreparePayloadDebug.java
```diff
@@ -0,0 +1,88 @@
...
+  public String getName() {
+    return RpcMethod.ENGINE_PREPARE_PAYLOAD_DEBUG.getMethodName();
+  }
+
+  @Override
+  public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext) {
+    final EnginePreparePayloadParameter enginePreparePayloadParameter =
+        requestContext.getRequiredParameter(0, EnginePreparePayloadParameter.class);
+
+    // TODO: respond with error if we're syncing
```
macfarla, line 60: is this a new TODO ?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
garyschulte: DEBUG is present even if the ENGINE api is not, so I think this needs to live within engine api.  Good point about the capabilities though 🤔 I will check to see if this causes problems for CLs

fab-10: What about a ENGINE_DEBUG set? So it is separated from ENGINE and can be enabled only if used

garyschulte: I filtered the debug prepare payload from the capabilities list in the same way the capabilities list filters itself 👍

garyschulte: just saw the ENGINE_DEBUG set suggestion.  I think we could add something like that if we have more than one engine debug endpoint in the future 👍 .  Rule of three applies in this case IMO

macfarla (approved):
LGTM

PR #4464, opened 2022-10-01
upgrade tuweni to 2.3.0

atoulme:

## PR description
Release candidate test for Besu.

## Documentation

## Changelog


antonydenyer (approved):

build.gradle
```diff
@@ -119,6 +119,10 @@ allprojects {
       url 'https://splunk.jfrog.io/splunk/ext-releases-local'
       content { includeGroupByRegex('com\\.splunk\\..*') }
     }
+    maven {
```
shemnon, line 122: This appears to be a staging repository.  How long until the full release occurs?  Or will the transition be transparent because the non-staged versions will go to central?

gradle/versions.gradle
```diff
@@ -45,7 +50,7 @@ dependencyManagement {
 
     dependency 'com.google.guava:guava:31.1-jre'
 
-    dependency 'com.graphql-java:graphql-java:18.1'
+    dependency 'com.graphql-java:graphql-java:19.2'
```
shemnon, line 53: Is the graphql version change required as part of the tuweni upgrade? Can it be put in a different PR?

gradle/versions.gradle
```diff
@@ -124,17 +129,17 @@ dependencyManagement {
...
-    dependency 'org.apache.tuweni:tuweni-concurrent:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-crypto:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-devp2p:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-dns-discovery:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-io:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-net:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-rlp:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-toml:2.2.0'
-    dependency 'org.apache.tuweni:tuweni-units:2.2.0'
+    dependency 'org.apache.tuweni:tuweni-bytes:2.3.0'
```
shemnon, line 132: consider converting this to a `dependencySet`

gradle/versions.gradle
atoulme: It is required, but could be done separate. This old version apparently depends on an old version of antlr which has a vulnerability. Tuweni upgraded, and now Besu complains that the new version of antlr passed in transitively is not compatible with the antlr schema used by graphql.

By upgrading graphql, we remove entirely its dependency to antlr.

shemnon: Sounds good, just wanting to make sure there was an intentional reason that was on record.

PR #4465, opened 2022-10-02
BlockCreator returns BlockCreationResult transaction selection results

jpnovais:
## PR description
Block creation returns `BlockCreationResult` which contains the new `Block` and `TransactionSelectionResults`. `TransactionSelectionResults` was extended with a list of `TransactionValidationResult` that contain the invalid transactions that were removed from the TX Pool during the block creation processes and the correspondent validation result.

At the of block creation and transaction selection, we have a list of all the transactions "processed" during the block creation: included transactions + discarded transactions and correspondent validation error (instead of just dropping them from Tx Pool);

Although, these changes surged out the work here in this PR #4146. I believe this would benefit Besu block creator flow.

## Documentation

## Changelog


config/src/main/java/org/hyperledger/besu/config/MergeConfigOptions.java
```diff
@@ -18,6 +18,7 @@
 
 public class MergeConfigOptions {
   private static final AtomicBoolean mergeEnabled = new AtomicBoolean(false);
+  private static final AtomicBoolean rollupExtensionEnabled = new AtomicBoolean(false);
```
jflo, line 21: Extending besu for rollup support likely warrants its own configuration class. You probably don't want to put it here since this should be getting cleaned up soon now that we are post-merge.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockCreator.java
```diff
@@ -22,12 +23,31 @@
 import java.util.Optional;
 
 public interface BlockCreator {
-  Block createBlock(final long timestamp);
+  class BlockCreationResult {
+    private final Block block;
+    private final TransactionSelectionResults transactionSelectionResults;
```
jflo, line 28: i love the overall concept for this addition, can we give this field a better name? if it is always going to be used for removed transactions, lets name it something that indicates those semantics. As is it implies these are the results of the transactions included in the block, which is misleading.

jflo (changes requested):
Love the spirit of this PR, just needs some clarifying improvements.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
```diff
@@ -74,10 +75,49 @@ public class BlockTransactionSelector {
...
+
+    @Override
+    public boolean equals(final Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      TransactionValidationResult that = (TransactionValidationResult) o;
```
fab-10, line 105: if we can assume that a transaction can only appear once in this list, then could be enough to check the tx hash for equality

fab-10 (commented):
LGTM, just a small optimization if possible, and I share the feedback of creating a dedicated config for Rollups

fab-10:
Please update the CHANGELOG

config/src/main/java/org/hyperledger/besu/config/MergeConfigOptions.java
jpnovais: This change was actually committed by mistake, as it does it is not used in the remaining changes. I will revert it.
Will keep in mind this class will be removed in the future and create rollup own config options.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
jpnovais: I am afraid we can't make that assumption. This is being used by 
```java 
// AbstractBlockCreator class 
BlockCreationResult createBlock(
      final Optional<List<Transaction>> maybeTransactions,
      final Optional<List<BlockHeader>> maybeOmmers,
      final Optional<Bytes32> maybePrevRandao,
      final long timestamp,
      boolean rewardCoinbase)
```
`maybeTransactions` can have repeated transactions.
So, I think it would be safer to keep it as it is.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockCreator.java
jpnovais: > As is it implies these are the results of the transactions included in the block, which is misleading.

This field contains the transactions that were included in the block as well. This is a class that already exited in Besu and was extended with `List<TransactionValidationResult> invalidTransactions` field, now containing 

```java
public static class TransactionSelectionResults {
    private final List<Transaction> transactions = Lists.newArrayList();
    private final List<TransactionReceipt> receipts = Lists.newArrayList();
    private final List<TransactionValidationResult> invalidTransactions = Lists.newArrayList();
    private long cumulativeGasUsed = 0;
    // ... 
}
```

Anyway,  if you have a suggestion for a better name happy to change :)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   * Before this change, the `bytes32` type was used for the enclave public keys, just supporting encryptors with public keys of that length (like the default NaCl)
   * For the EC encryptor, the encoded public key length is 91
 - `--tx-pool-hashes-max-size` option removed (deprecated in 22.1.3).
+- Block creation now returns `BlockCreationResult` which contains the new `Block` and `TransactionSelectionResults`.
```
fab-10, line 9: this should not be a braking change, since you are not changing any configuration option, or plugin interface, so it could go under **Additions and Improvements**

jpnovais: Distraction mistake 🤦🏼‍♂️ . Fixed, thanks 👍

fab-10 (approved):

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockCreator.java
jflo: No, I think this is good, it's just a layer deeper than I looked. No problems with that.

jflo (approved):

PR #4501, opened 2022-10-10
hide deprecated option tx-pool-max-hashes

macfarla:

Deprecated since 22.1.3 https://github.com/hyperledger/besu/blob/main/CHANGELOG.md#2213

## Documentation

## Changelog


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1188,10 +1188,11 @@ static class TxPoolOptionGroup {
     private final Integer txPoolMaxSize = TransactionPoolConfiguration.MAX_PENDING_TRANSACTIONS;
 
     @Option(
+        hidden = true,
         names = {"--tx-pool-hashes-max-size"},
```
jframe, line 1192: Can we just remove this config option, or do we need to wait for another release? Doesn't look like it is used in the code at all.

fab-10: Yes we can remove it since we are starting a new RC series, adding a breaking change warning to the release notes.
I have deprecated it since months, and I doubt anyone is actually using this option.

macfarla: done!

fab-10 (approved):

PR #4508, opened 2022-10-10
Reduce the number of runtime exceptions (SecurityModuleException)

ahamlat:

## PR description
During handshake, flip the encrypted message decryption by starting with the new format (EIP-8), and if there is an exception, try the old format. This will reduce the number of exceptions and unnecessary executions.

I noticed by adding some debug logs that on 4400 calls to ECIESHandshaker.handleMessage, there're only 7 cases where the encrypted message has the old format.
image

By changing the order of execution, starting by decoding the new format (eip8) and without resetting the logs, we can notice  there're no more exceptions

image

There is one case I'm not sure about it : if (buf.writerIndex() < size), do we need to try to decode with the old format or we should just throw HandshakeException. With the current PR, HandshakeException will be caught and the old format decryption will be executed.

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
   * For the EC encryptor, the encoded public key length is 91
 
 ### Additions and Improvements
+
+- Reduce the number of runtime exceptions (SecurityModuleException) and unnecessary executions during ECIES handshake [#4508](https://github.com/hyperledger/besu/pull/4508)
```
fab-10, line 11: I would also explain that we are reducing these exceptions try EIP-8 formatted messages first

fab-10 (approved):
LGTM, just a suggestion on the changelog

garyschulte (approved):
LGTM.

Seems like such a low percentage of handshakes are using the pre-EIP-8 format, that it doesn't really matter much whether we try to decode with the old format if the buffer length is wrong.

macfarla (approved):

fab-10 (approved):

PR #4513, opened 2022-10-10
Corrected eth_getLogs default fromBlock value.

### LINKED ISSUE #4123: JSON-RPC method eth_getLogs improperly interprets block arguments
### Description
Besu's implementation of `eth_getLogs` accepts a "blockParameter" for the `fromBlock` and `toBlock` values where the block parameter can be a number or a special string ("latest", "earliest", or "pending").  If a string parameter is provided, it is ignored and  [is instead replaced](https://github.com/hyperledger/besu/blob/32fd3476e3409ec5b10c654b96a00418f24f53a5/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetLogs.java#L61-L63) with the latest block (in the case of the `toBlock` parameter) or the earliest block (in the case of the `fromBlock` parameter).  This is a bug - if I request blocks from "latest" to "latest", I would expect to get logs from the latest block, but I instead initiate a query for all logs across the entire history of the chain. 

Additionally, the [JSON-RPC spec documentation](https://github.com/ethereum/execution-apis/blob/1e2058b3a104ada4adfd3d3b90e9982050541fc4/src/engine/specification.md#underlying-protocol) only mentions numeric values for the `fromBlock` and `toBlock` parameters.  So, the API should most likely be throwing an error if a string value is given.  However, if besu is going to have expanded functionality and accept string parameters, they should be interpreted correctly.

As a general protection, it's probably a good idea to return an error if the range of blocks requested is too large.  An optional upperbound parameter (`--rpc-logs-range-upper-bound`?) could be added to enforce this.  

### Acceptance Criteria
* Besu does not interpret requests to `eth_getLogs` from "latest" to "latest" as a query across the entire range of all blocks

mark-terry:

## PR description
Changes the default ```fromBlock``` value in ```eth_getLogs``` from ```earliest``` to ```latest``` per the spec.

Added tests to verify BlockParameter value interpretation.

## Fixed Issue(s)

Fixes #4123 

## Documentation

## Changelog


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/BlockParameterTest.java
```diff
@@ -0,0 +1,115 @@
...
+    assertThat(blockParameter.isEarliest()).isTrue();
+    assertThat(blockParameter.isFinalized()).isFalse();
+    assertThat(blockParameter.isLatest()).isFalse();
+    assertThat(blockParameter.isNumeric()).isFalse();
+    assertThat(blockParameter.isPending()).isFalse();
+    assertThat(blockParameter.isSafe()).isFalse();
+  }
+
+  @Test
+  public void latestShouldReturnEmpty() {
```
macfarla, line 37: why does latest return empty

macfarla (commented):
for completeness can you add a test for the invalid case - what happens if I supply a non-parseable string

fab-10 (changes requested):
I think we need to first define which is our goal, because I see 2 possibility here:

1. Follow the specs, then we need to reject everything that is not an uint, and return an error to the caller
2. Extend the functionality, accepting with block names "lastest", "earlier" etc... then we need to interpret them correctly

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/BlockParameterTest.java
mark-terry: All string values except "earliest" return an empty `number` value.

mark-terry: Will update test names to clarify.

mark-terry:
> I think we need to first define which is our goal, because I see 2 possibility here:
> 
> 1. Follow the specs, then we need to reject everything that is not an uint, and return an error to the caller
> 2. Extend the functionality, accepting with block names "lastest", "earlier" etc... then we need to interpret them correctly

I think option 2 is preferable; maintaining API consistency by using the `filterOptions` parameter type. "earliest", "latest" and numeric values are already covered; will investigate the others...

fab-10:
> > I think we need to first define which is our goal, because I see 2 possibility here:
> > 
> > 1. Follow the specs, then we need to reject everything that is not an uint, and return an error to the caller
> > 2. Extend the functionality, accepting with block names "lastest", "earlier" etc... then we need to interpret them correctly
> 
> I think option 2 is preferable; maintaining API consistency by using the `filterOptions` parameter type. "earliest", "latest" and numeric values are already covered; will investigate the others...

Ok then take a look at `AbstractBlockParameterMethod` class, probably it can help to how block parameters are handled in other methods

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetLogs.java
```diff
@@ -71,4 +72,26 @@ public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
     return new JsonRpcSuccessResponse(
         requestContext.getRequest().getId(), new LogsResult(matchingLogs));
   }
+
+  private long interpretBlockParam(final BlockParameter block) {
+    if (block.isFinalized()) {
+      return blockchain
+          .finalizedBlockHeader()
+          .orElseThrow(() -> new IllegalArgumentException("Finalized block not found."))
```
macfarla, line 80: IllegalArg? will this return a JsonErrorResponse to the user

mark-terry: Not quite. Will address...

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetLogsTest.java
```diff
@@ -0,0 +1,296 @@
...
+    verify(blockchainQueries).matchingLogs(eq(genesisBlock), eq(latestBlock), any(), any());
+    verify(blockchainQueries, never()).finalizedBlockHeader();
+    verify(blockchainQueries, never()).safeBlockHeader();
+  }
+
+  @Test
+  public void shouldQuerySafeToFinalized() {
+    final long finalizedBlockNumber = 50L;
+    final long safeBlockNumber = 25L;
+    final JsonRpcRequestContext request = buildRequest("safe", "Finalized");
```
fab-10, line 231: is the capitalization of "Finalized" intended? is it case insensitive?

fab-10 (commented):
ok the parameter interpretation, it will be nice to have an abstract class for `FilterParameters` similar to what `AbstractBlockParameterMethod` does for `BlockParameter` to unify all methods that use the filter, but this is probably a refactor that can be done separately.
Have you tested this PR with a Consensus client? since `eth_getLogs` is critial after the merge

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetLogsTest.java
mark-terry: Not intentional, but it gets sanitised in BlockParameter. Good catch. 👍

mark-terry:
> Have you tested this PR with a Consensus client? since `eth_getLogs` is critial after the merge

It's currently running with Teku on Sepolia without any obvious issues. 👍

fab-10 (approved):

CHANGELOG.md
```diff
@@ -36,8 +39,6 @@
 - Avoid a cyclic reference while printing EngineExchangeTransitionConfigurationParameter [#4357](https://github.com/hyperledger/besu/pull/4357)
 - Corrects treating a block as bad on internal error [#4512](https://github.com/hyperledger/besu/issues/4512)
 - In GraphQL update scalar parsing to be variable friendly [#4522](https://github.com/hyperledger/besu/pull/4522)
-- Initiate connection to maintained peers soon after startup. [#4469](https://github.com/hyperledger/besu/pull/4469)
-- Update apache-commons-text to 1.10.0 to address CVE-2022-42889 [#4542](https://github.com/hyperledger/besu/pull/4542)
```
fab-10, line 40: Please readd these 2 entries

mark-terry: https://github.com/hyperledger/besu/pull/4560

Issue #4460, opened 2022-09-30
Besu 22.7.4 created 2 blocks with 5-8% gas utilization


nhlroyrocks:
### Description
As an validator, I want besu to create blocks at least in the ballpark of 50% full so that block space is not wasted and greater transaction fees may be collected. 

### Acceptance Criteria
produce blocks with 50-100% gas utilization

### Steps to Reproduce (Bug)
1. propose a block running besu execution client
    a. [Nearly empty block](https://etherscan.io/block/15637228)
    b. [Another nearly empty block within 12 hours of the first](https://etherscan.io/block/15640450)

**Expected behavior:** [What you expect to happen]
fuller blocks like the ones before and after my blocks.
before:  https://etherscan.io/block/15637227
   after:  https://etherscan.io/block/15637229
before:  https://etherscan.io/block/15640449
   after:  https://etherscan.io/block/15640451

**Actual behavior:** [What actually happens]
 1. [Nearly empty block](https://etherscan.io/block/15637228)
 2. [Another nearly empty block within 12 hours of the first](https://etherscan.io/block/15640450)
 
**Frequency:** [What percentage of the time does it occur?]
100% 2/2

### Versions (Add all that apply)
* Software version: besu/v22.7.4/linux-x86_64/openjdk-java-11

* Java version: 
openjdk version "11.0.16" 2022-07-19
OpenJDK Runtime Environment (build 11.0.16+8-post-Ubuntu-0ubuntu120.04)
OpenJDK 64-Bit Server VM (build 11.0.16+8-post-Ubuntu-0ubuntu120.04, mixed mode, sharing)

* OS Name & Version: 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.4 LTS"
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
[...]

PR #4516, opened 2022-10-11
Try to build better block proposals until timeout or GetPayload is ca…

fab-10:
…lled

## PR description

Take advantage of all the time between a FcU with payload attributes, that start the block creation process, until the GetPayload to repeatedly try to build better blocks.
Currently when the Consensus client asks to Besu to build a block, Besu first create the empty block, then it asynchronously try to build a better block that will eventually replace the empty block. The better block jobs is only done once, but we can repeat it to fit more newly arrived transactions until the CL client retrieve the block proposal, or if this does not happen, after the maximum allowed time for the task, that is configurable using `--Xpos-block-creation-max-time` and has a default of 12 seconds.

## Fixed Issue(s)

mitigates #4460

## Documentation

## Changelog


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -208,75 +216,119 @@ public PayloadIdentifier preparePayload(
...
-          } else {
-            debugLambda(
-                LOG,
-                "Something went wrong creating block for payload id {}, error {}",
-                payloadIdentifier::toShortHexString,
-                () -> logException(throwable));
-          }
-          return CompletableFuture.failedFuture(throwable);
-        });
+  private void newBlockFound(
```
jflo, line 307: This method is named after the situation being reacted to. Can we rename it so it describes what it is going to do instead? Perhaps upgradeIfTimeLeft?

jflo (approved):
Looks good, one minor optional naming improvement suggested.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
fab-10: Indeed its name can be improved, and I am thinking about something like `evaluateNewBlock` since it is not yes sure the block is an upgrade, since it still to be validated and the compared with already found blocks, what do you think?

jflo (approved):
Even better, reduces dependency scope so it doesn't require the whole EthContext. nice.

PR #4517, opened 2022-10-11
Upgrade RocksDB version from 6.29.5 to 7.6.0.

ahamlat:

## PR description
Upgrade RocksDB version from 6.29.5 to 7.6.0.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,13 +16,15 @@
 - `--Xmerge-support` option remove (deprecated in 22.4.2) [#4518](https://github.com/hyperledger/besu/pull/4518)
 
 ### Additions and Improvements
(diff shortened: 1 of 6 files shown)


ahamlat:
Upgrading RocksDB to version 7.6.0 give good results once the node is synced. The node had some GC issues at the beginning of sync but it was able to recover and sync faster. The only significant difference I see in RocksDB metrics is on read from SST files, RocksDB version 7.6.0 has 29 microseconds as 95 percentile against 159 microseconds for version 6.29.5. I will do more tests but it seems promising. I compared all the nodes I started that are currently synced and the one with version 7.6.0 has better performance, without high spec flag enabled. (I just noticed that the node with RocksDB version 6.29.5 has high spec enabled).

image

This seems also to have also an impact if we upgrade the version on a synced node with an older RocksDB version>

image

I will do more tests on different hardware specifications by combining with other RocksDB configurations that gave also good results.

fab-10 (approved):
LGTM, just add a CHANGELOG entry for it

ahamlat:
> LGTM, just add a CHANGELOG entry for it

Done.

PR #4521, opened 2022-10-12
check for self in one more spot

macfarla:

When regularly checking connections to maintained peers (static nodes), ensure connection to self is not initiated.

## Documentation

## Changelog

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
@@ -359,8 +359,10 @@ public class DefaultP2PNetwork implements P2PNetwork {
     if (!localNode.isReady()) {
       return;
     }
+    final EnodeURL localEnodeURL = localNode.getPeer().getEnodeURL();
     maintainedPeers
         .streamPeers()
+        .filter(peer -> !peer.getEnodeURL().getNodeId().equals(localEnodeURL.getNodeId()))
         .filter(p -> !rlpxAgent.getPeerConnection(p).isPresent())
         .forEach(rlpxAgent::connect);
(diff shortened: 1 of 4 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -359,8 +359,10 @@ void checkMaintainedConnectionPeers() {
     if (!localNode.isReady()) {
       return;
     }
+    final EnodeURL localEnodeURL = localNode.getPeer().getEnodeURL();
     maintainedPeers
```
siladu, line 363: Is there a good reason why self exists in the maintainedPeers set? Or is that being looked at separately?

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -359,8 +359,10 @@ void checkMaintainedConnectionPeers() {
...
         .streamPeers()
+        .filter(peer -> !peer.getEnodeURL().getNodeId().equals(localEnodeURL.getNodeId()))
```
gezero, line 365: Is enoudeURL::equals working in this case?

gezero (approved):

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/DefaultP2PNetwork.java
macfarla: the way the code is now, at startup, you don't know your own enode at the time the static nodes are added to the maintainedPeers set, so you can't avoid adding your own enode

PR #4529, opened 2022-10-13
Tune EthScheduler thread pools to avoid to recreate too many threads

fab-10:

## Documentation

## Changelog


jflo (approved):
Prior to this pr, every task would be hitting an empty pool, causing new thread creation, and that new thread would be retained for 0ms after being returned to pool. 

I don't see how this change doesn't result in a big performance improvement.  Nice find.

Issue #4199, opened 2022-07-29
We need access to a worldstate image of the last 128 blocks with bonsai


matkt:
With bonsai we have the worldstate of the head and tries logs to be able to rollback or rollforward the worldstate to an old/recent block. This technique with trielogs works very well in the case of reorgs but has limitations.

In particular with the snapsync as a server we must be able to provide the nodes of the worldstate of the last 128 blocks. If we have to rollback after every request it will be a lot of work for the node and risks impacting the stability of the node.

To avoid this it will be preferable to keep an image of the last 128 blocks and rocksdb's snapshot feature might help to do this

http://rocksdb.org/blog/2015/11/10/use-checkpoints-for-efficient-snapshots.html

This feature can also help make bonsai more stable. If we detect worldstate corruption for some reason (crash or bug). We could go back and take the worldstate of block head-1 and resume synchronization from this one

Example : 

- 1 : the node tries to import block h+1
- 2 : stateroot mismatch or other issue detected
- 3 : besu take the worldstate image of block h-1 and try again to import block h
- 4: the database is fixed thanks to the new import
- 5: we import block h+1 again

garyschulte:
As we discussed, a persistent snapshot seems like a viable source to rollback to whereas a collection of n (e.g. 128)  in-memory snapshot transaction based worldstates is likely a more suitable mechanism for snapsync server.

see https://github.com/garyschulte/besu/tree/feature/bonsai-isolation (re-pushed and rebased off of main) and specifically:

[takeSnapshot()](https://github.com/garyschulte/besu/blob/feature/bonsai-isolation/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java#L182)
[RocksDbColumnarKeyValueSnapshot](https://github.com/garyschulte/besu/blob/feature/bonsai-isolation/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueSnapshot.java)

Issue #4151, opened 2022-07-22
World State Root does not match expected value -> Error processing block/Failed to import block: Invalid block


howardrcc:
### Description
Hi there. I successfully synced to mainnet a few day using snapsync. I am getting Errors now. Here is a snippet of the log (I'm not seeing other usefull messages in TRACE/DEBUG). I'm not sure what is happing. I tried restart and upgrading besu as suggested in other Issues but thats not working for me.

Any suggestions?

``` 
2022-07-22 10:52:28.053+00:00 | EthScheduler-Services-3 (importBlock) | ERROR | AbstractBlockProcessor | failed persisting block
java.lang.RuntimeException: World State Root does not match expected value, header 0xe821a58beda94f8ff6b230b1933235f365c52a90a590e09bfda7a89c95991939 calculated 0xe50fb371f71f4d01e127f68c8d8f99b5cd638d3ac4ffb7cf46c9378c9b49de2d
        at org.hyperledger.besu.ethereum.bonsai.BonsaiPersistedWorldState.persist(BonsaiPersistedWorldState.java:255)
        at org.hyperledger.besu.ethereum.mainnet.AbstractBlockProcessor.processBlock(AbstractBlockProcessor.java:201)
        at org.hyperledger.besu.ethereum.mainnet.AbstractBlockProcessor.processBlock(AbstractBlockProcessor.java:42)
        at org.hyperledger.besu.ethereum.mainnet.BlockProcessor.processBlock(BlockProcessor.java:75)
        at org.hyperledger.besu.ethereum.MainnetBlockValidator.processBlock(MainnetBlockValidator.java:158)
        at org.hyperledger.besu.ethereum.MainnetBlockValidator.validateAndProcessBlock(MainnetBlockValidator.java:112)
        at org.hyperledger.besu.ethereum.MainnetBlockValidator.validateAndProcessBlock(MainnetBlockValidator.java:71)
        at org.hyperledger.besu.ethereum.mainnet.MainnetBlockImporter.importBlock(MainnetBlockImporter.java:44)
        at org.hyperledger.besu.ethereum.core.BlockImporter.importBlock(BlockImporter.java:44)
        at org.hyperledger.besu.ethereum.eth.sync.fullsync.FullImportBlockStep.accept(FullImportBlockStep.java:61)
        at org.hyperledger.besu.ethereum.eth.sync.fullsync.FullImportBlockStep.accept(FullImportBlockStep.java:31)
[...]

PR #4531, opened 2022-10-13
Snapshot based non-persisting MutableWorldState usage

garyschulte:
## PR description

Adds Bonsai snapshots as an experimental option (`--Xbonsai-use-snapshots`) into BonsaiWorldStateArchive for arbitrary mutable non-persisting worldstate.  Also refactors TrieLogManager into a LayeredTrieLogManager and SnapshotTrieLogManager such that cached worldstates use either layered worldstates or snapshot worldstates.

Since snapshot transactions need to release their snapshot in order to allow their logs to be compacted, this PR refactors direct uses of mutable worldstates to be in try-with-resources blocks (to auto-close snapshots after usage):
* MainnetTransactionValidator
* AbstractBlockCreator
* TransactionSimulator
* TransactionPool.validateTransaction

## Fixed Issue(s)

relates to #4372
relates to #4250
relates to #4199
relates to #4151

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/TrieLogManager.java
```diff
@@ -130,7 +131,19 @@ public synchronized void updateLayeredWorldState(
...
+        .stream()
+        .forEach(
+            layer -> {
+              layeredWorldStatesByHash.remove(layer);
+              try {
+                layer.close();
+              } catch (Exception e) {
+                LOG.warn("Error closing bonsai worldstate layer", e);
+              }
+            });
```
garyschulte, line 146: Not sure where this came from.  It isn't in main.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/TrieLogManager.java
```diff
@@ -130,7 +131,19 @@
...
   public synchronized void scrubLayeredCache(final long newMaxHeight) {
     final long waterline = newMaxHeight - RETAINED_LAYERS;
-    layeredWorldStatesByHash.entrySet().removeIf(entry -> entry.getValue().getHeight() < waterline);
+    layeredWorldStatesByHash.values().stream()
+        .filter(layer -> layer.getHeight() < waterline)
+        .collect(Collectors.toList())
+        .stream()
+        .forEach(
+            layer -> {
+              layeredWorldStatesByHash.remove(layer);
```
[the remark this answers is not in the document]
garyschulte: Not sure where this came from.  It isn't in main or history.  Will remove

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -50,31 +55,64 @@ public class BonsaiWorldStateArchive implements WorldStateArchive {
...
   public BonsaiWorldStateArchive(
+      final BonsaiWorldStateKeyValueStorage worldStateStorage,
+      final Blockchain blockchain,
+      final Optional<Long> maxLayersToLoad,
+      final boolean useSnapshots) {
+    this(
+        useSnapshots
+            ? new SnapshotTrieLogManager(
+                blockchain, worldStateStorage, maxLayersToLoad.orElse(RETAINED_LAYERS))
+            : new LayeredTrieLogManager(
```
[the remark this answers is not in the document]
garyschulte: right.  We just have a collection of snapshots, not a layered worldstate.  Trie logs are only used to roll a snapshot to a particular state, we don't navigate through them to reproduce a particular state

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -294,10 +294,16 @@ public long getTransactionCount(final Address address, final long blockNumber) {
...
-        .orElse(0L);
+    try (final var worldState =
+        getWorldState(blockHash)
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        "Missing worldstate for stateroot " + blockHash.toShortHexString()))) {
+      return Optional.ofNullable(worldState.get(address)).map(Account::getNonce).orElse(0L);
+    } catch (Exception ex) {
+      throw new RuntimeException(ex);
```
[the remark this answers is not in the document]
garyschulte: Good point - for now we just re-throw wrapped in a runtime exception.  There is some risk that specific exception handling could get bypassed in this case (like the explicit handling of storage exceptions in the merge api).  It is probably better not to explicitly catch and just have a finally block so we can be sure the worldstate is close()'d

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -252,21 +249,20 @@ private BlockTransactionSelector.TransactionSelectionResults selectTransactions(
...
+            () -> {
+              LOG.info("Unable to create block because world state is not available");
+              return new CancellationException(
+                  "World state not available for block "
+                      + parentHeader.getNumber()
+                      + " with state root "
+                      + parentStateRoot);
+            })
+        // TODO: only copy if this is from layered worldstate
+        .copy();
```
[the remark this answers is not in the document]
garyschulte: I need to re-review this after adding the snapshot TrieLogManager.  My original intent was to push the copy up into `getMutableSnapshot()`.  Right now copy() is safe but potentially unnecessary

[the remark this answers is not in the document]
garyschulte: for the snapshot trie log manager, when we get a mutable non-persisting worldstate, it is giving a copy.  That way we can't have a rogue bit of code accidentally corrupting the snapshots we are keeping for the last 512 states.  

I am inclined to do this for the trie log manager used for the layered worldstate also and remove the copy() here.  
Then for read only uses like blockchain queries we can create a ReadOnlyWorldState so that we only bother to create copies when we are actually going to mutate the worldstate.  Thoughts?  I will probably wait and do this in a subsequent PR since this one is getting a bit unwieldy already.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -94,71 +131,74 @@ public Optional<WorldState> get(final Hash rootHash, final Hash blockHash) {
...
   }
 
-  public Optional<SnapshotMutableWorldState> getMutableSnapshot(final Hash blockHash) {
+  public Optional<MutableWorldState> getMutableSnapshot(final Hash blockHash) {
+    // TODO: decide whether we want to use BonsaiSnapshotWorldState or
+    // BonsaiSnapshotPersistedWorldState
     return rollMutableStateToBlockHash(
-            BonsaiSnapshotWorldState.create(this, worldStateStorage), blockHash)
+            //            BonsaiSnapshotWorldState.create(this, worldStateStorage), blockHash)
+            BonsaiSnapshotPersistedWorldState.create(this, worldStateStorage), blockHash)
```
garyschulte, line 144: both implementations are working, but the 'persisted snapshot' version is currently enabled.  We need to do some performance and resource consumption analysis to see which one is "better" in practice

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -918,8 +934,8 @@ public Optional<Wei> gasPriorityFee() {
   }
 
   private <T> Optional<T> fromWorldState(
-      final Hash blockHash, final Function<WorldState, T> getter) {
-    return getWorldState(blockHash).map(getter);
+      final Hash blockHash, final Function<MutableWorldState, T> getter) {
+    return mapWorldState(blockHash, getter);
```
garyschulte, line 938: this is redundant, just use mapWorldState directly

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueStorageProvider.java
```diff
@@ -92,7 +96,7 @@ public KeyValueStorage getStorageBySegmentIdentifier(final SegmentIdentifier seg
   @Override
   public SnappableKeyValueStorage getSnappableStorageBySegmentIdentifier(
       final SegmentIdentifier segment) {
-    return (SnappableKeyValueStorage) storageInstances.computeIfAbsent(segment, storageCreator);
+    return (SnappableKeyValueStorage) getStorageBySegmentIdentifier(segment);
```
garyschulte, line 99: 👀

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -144,21 +144,31 @@ public Optional<TransactionSimulatorResult> process(
...
 
-    return processWithWorldUpdater(
-        callParams, transactionValidationParams, operationTracer, header, updater);
+  public MutableWorldState getWorldState(final BlockHeader header) {
+    return worldStateArchive
+        .getMutable(header.getStateRoot(), header.getHash(), false)
+        .orElseThrow(
+            () ->
+                new IllegalArgumentException(
+                    "Public world state not available for block " + header.getNumber()));
```
garyschulte, line 171: this should be private and the impl of traceCallMany should get pushed into this class

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -303,20 +313,8 @@ private Optional<Transaction> buildTransaction(
...
-
-    return getEffectiveWorldStateUpdater(header, publicWorldState);
-  }
-
   // return combined private/public world state updater if GoQuorum mode, otherwise the public state
-  private WorldUpdater getEffectiveWorldStateUpdater(
+  public WorldUpdater getEffectiveWorldStateUpdater(
       final BlockHeader header, final MutableWorldState publicWorldState) {
 
     if (maybePrivacyParameters.isPresent()
```
garyschulte, line 320: also make private

gezero (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -94,71 +131,71 @@ public Optional<WorldState> get(final Hash rootHash, final Hash blockHash) {
...
         .map(SnapshotMutableWorldState.class::cast);
   }
 
   @Override
   public Optional<MutableWorldState> getMutable(
-      final long blockNumber, final boolean isPersistingState) {
-    final Optional<Hash> blockHashByNumber = blockchain.getBlockHashByNumber(blockNumber);
-    if (blockHashByNumber.isPresent()) {
-      return getMutable(null, blockHashByNumber.get(), isPersistingState);
+      final Hash rootHash, final Hash blockHash, final boolean shouldPersistState) {
```
jflo, line 147: rootHash is only needed if shouldPersistState == true, and in that case it is not passed any further along. Could be deleted.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
```diff
@@ -0,0 +1,115 @@
...
+      final BonsaiWorldStateArchive worldStateArchive) {
+
+    debugLambda(
+        LOG,
+        "adding snapshot world state for block {}, state root hash {}",
+        blockHeader::toLogString,
+        worldStateRootHash::toHexString);
+    worldStateArchive
+        .getMutableSnapshot(worldStateRootHash)
+        .map(BonsaiSnapshotWorldState.class::cast)
```
jflo, line 66: nice. TIL there is a functional interface for casting

jflo (approved):
just a couple of minor optional notes, but this is pretty awesome.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/processor/BlockReplay.java
```diff
@@ -145,14 +145,18 @@ private <T> Optional<T> performActionWithBlock(
     if (previous == null) {
       return Optional.empty();
     }
-    final MutableWorldState mutableWorldState =
+    try (final MutableWorldState mutableWorldState =
         worldStateArchive
             .getMutable(previous.getStateRoot(), previous.getHash(), false)
-            .orElse(null);
-    if (mutableWorldState == null) {
+            .orElseThrow(
```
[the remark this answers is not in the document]
garyschulte: We actually won't throw here, we will catch the runtime exception in the catch block on :157

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -144,21 +144,31 @@ public Optional<TransactionSimulatorResult> process(
...
+  }
 
-    return processWithWorldUpdater(
-        callParams, transactionValidationParams, operationTracer, header, updater);
+  private MutableWorldState getWorldState(final BlockHeader header) {
+    return worldStateArchive
+        .getMutable(header.getStateRoot(), header.getHash(), false)
+        .orElseThrow(
+            () ->
+                new IllegalArgumentException(
```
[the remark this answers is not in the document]
garyschulte: changed to header.toLogString() 👍

PR #4551, opened 2022-10-22
Refactoring of calculate root hash

gezero:
Small refactoring of the calculate root hash. 

The second commit fixes a tiny bug where we leave a few insignificant nodes in the trie branch storage.


macfarla:
this is already merged in as part of #4568  so we can close this one, right @gezero ?

PR #4555, opened 2022-10-25
List.of like utility class for BlockBodies

gezero:
This is a QoL refactor that should not have an effect of behaviour. 

When working on Withdrawals I noticed code similar to some of these:
```
  new BlockBody(emptyList(),emptyList());
  new BlockBody(List.of(transaction1),emptyList());
  new BlockBody(emptyList(),List.of(ommerHeader));
```

This PR adds a utility class BlockBodies that should work similarly to List.of(...) there is a BlockBodies.empty()
there is a utility method for BlockBodies.of(transaction) there is a utility method for BlockBodies.of(ommerHeader)

And there will be probably similar one when you will need a Block body only with a withdrawal.

The BlockBody.empty() was moved into this utility class as well...


macfarla (commented):
it is slightly confusing to me that the methods on BlockBodies (plural) only ever return a single block - would it make sense to have the of() util methods just added to the BlockBody class itself?

garyschulte (approved):
Pleasant parallel to java Collections

non-fungible-nelson:
Hi Jiri - closing as stale, please feel free to reopen if you are able to rebase!

PR #4562, opened 2022-10-26
Migrate storage APIs to Bytes32

shemnon:
## PR description

The current internal storage APIs implement UInt256 rather than Bytes32, which is an accident of prior EVM design.  This migrates all internal handling of storage addresses and values to Bytes32.

The main performance gain is that UInt256 stores the data as an array of ints, whereas Bytes32 keeps a sliced or wrapped array of bytes.  Since database APIs are all byte array based this should remove unneeded conversions when going to/from the DB.

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
@@ -28,7 +28,6 @@ import org.hyperledger.besu.evm.worldstate.UpdateTrackingAccount;
 import org.hyperledger.besu.evm.worldstate.WrappedEvmAccount;
 
 import java.util.Collection;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -38,11 +37,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Function;
 
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
-import org.apache.tuweni.units.bigints.UInt256;
 
 public class BonsaiWorldStateUpdater extends AbstractWorldUpdater<BonsaiWorldView, BonsaiAccount>
     implements BonsaiWorldView {
@@ -54,7 +51,7 @@ public class BonsaiWorldStateUpdater extends AbstractWorldUpdater<BonsaiWorldVie
   // storage sub mapped by _hashed_ key.  This is because in self_destruct calls we need to
   // enumerate the old storage and delete it.  Those are trie stored by hashed key by spec and the
   // alternative was to keep a giant pre-image cache of the entire trie.
-  private Map<Address, Map<Hash, BonsaiValue<UInt256>>> storageToUpdate = new ConcurrentHashMap<>();
+  private Map<Address, Map<Hash, BonsaiValue<Bytes32>>> storageToUpdate = new ConcurrentHashMap<>();
 
   BonsaiWorldStateUpdater(final BonsaiWorldView world) {
     super(world);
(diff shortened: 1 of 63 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -201,7 +182,7 @@ public void commit() {
                 (keyHash, entryValue) -> {
                   final Hash slotHash = Hash.wrap(keyHash);
                   if (!deletedStorageUpdates.containsKey(slotHash)) {
-                    final UInt256 value = UInt256.fromBytes(RLP.decodeOne(entryValue));
+                    final Bytes32 value = Bytes32.leftPad(Bytes.wrap(RLP.decodeOne(entryValue)));
```
pinges, line 185: Why do we need the Bytes.wrap here?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/privacy/PrivateStateGenesisAllocatorTest.java
```diff
@@ -146,8 +146,8 @@ private void assertGenesisAccountApplied() {
   private void assertManagementContractApplied() {
     Account managementProxy = worldState.get(FLEXIBLE_PRIVACY_PROXY);
     assertThat(managementProxy.getCode()).isEqualTo(PROXY_RUNTIME_BYTECODE);
-    assertThat(managementProxy.getStorageValue(UInt256.ZERO))
-        .isEqualTo(UInt256.fromBytes(DEFAULT_FLEXIBLE_PRIVACY_MANAGEMENT));
+    assertThat(managementProxy.getStorageValue(Bytes32.ZERO))
+        .isEqualTo(Bytes32.leftPad(Bytes.wrap(DEFAULT_FLEXIBLE_PRIVACY_MANAGEMENT)));
```
pinges, line 150: DEFAULT_FLEXIBLE_PRIVACY_MANAGEMENT is of type Address, which implements Bytes. Why do we need the wrap?

pinges (commented):
Please have a look at my comments.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
shemnon: We don't.  Large scale change missed it.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/privacy/PrivateStateGenesisAllocatorTest.java
shemnon: We don't, removing it.

shemnon:
I looked for all instances where `Bytes.wrap(Bytes ... bytes)` was called with a single argument and removed them all.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/TrieLogLayer.java
```diff
@@ -136,13 +136,13 @@ public static TrieLogLayer readFrom(final RLPInput input) {
...
-        final Map<Hash, BonsaiValue<UInt256>> storageChanges = new TreeMap<>();
+        final Map<Hash, BonsaiValue<Bytes32>> storageChanges = new TreeMap<>();
         input.enterList();
         while (!input.isEndOfCurrentList()) {
           input.enterList();
           final Hash slotHash = Hash.wrap(input.readBytes32());
-          final UInt256 oldValue = nullOrValue(input, RLPInput::readUInt256Scalar);
-          final UInt256 newValue = nullOrValue(input, RLPInput::readUInt256Scalar);
+          final Bytes32 oldValue = nullOrValue(input, RLPInput::readUInt256Scalar);
+          final Bytes32 newValue = nullOrValue(input, RLPInput::readUInt256Scalar);
```
fab-10, line 145: is reading uint256 correct here, or should it read bytes32?

fab-10 (commented):
LGTM, just checking if a reading operation is correct.
Have you any result on how much this can improve performance?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/TrieLogLayer.java
shemnon: `readBytes32Scalar` wasn't in the RLPInput interface, so I pulled it up.  It also exposed that the proofs were still on `UInt256`s

shemnon:
Benchmarks I have are not for public consumption.

This was mostly about pulling the `UInt256` out of APIs it's not needed (i.e. no math calls).  Each UInt256 requires a round trip from a primitive byte array into a primitive int array and then back to a bytes array for serialization.  So this is more about GC pressure (where the byte array can be re-used and passed around as slices, instead of two primitive conversions minimum) that it was about wall clock time.

fab-10 (approved):

PR #4565, opened 2022-10-27
Add port conflict exception

### LINKED ISSUE #4176: Better error when unable to start due to port conflict for json-rpc services
### Description
when starting besu, if the configured rpc port is not available, besu will exit with a not very helpful error.  This applies to the rpc http port, the engine port, and web service port

### Acceptance Criteria
* besu should log a better error when it is unable to start json rpc service(s) on the configured port

### Steps to Reproduce (Bug)
1. start 2 besu instances with the same configured rpc port
2. the second node will fail to start with a less than useful error message, e.g. :
```
2022-07-26 18:53:12.207-07:00 | main | INFO  | JsonRpcService | Starting JSON-RPC service on 127.0.0.1:8551
Service engineJsonRpc failed to start

To display full help:
besu [COMMAND] --help
2022-07-26 18:53:12.297-07:00 | BesuCommand-Shutdown-Hook | ERROR | Besu | Failed to stop Besu
```

gfukushima:
## PR description
Added methods to check if port is available and call before starting websockets and rpc services.

## Fixed Issue(s)

Fixes #4176
## Documentation

## Changelog


util/src/main/java/org/hyperledger/besu/util/NetworkUtility.java
```diff
@@ -98,4 +104,23 @@ public static void checkPort(final int port, final String portTypeName) {
...
+      serverSocket.setReuseAddress(true);
+      serverSocket.bind(new InetSocketAddress(port));
+      return true;
+    } catch (IOException ex) {
+      LOG.trace(String.format("Failed to open port %d for TCP", port), ex);
+    }
+    return false;
+  }
+
+  public static void checkPortsAvailable(final int tcpPort) {
```
fab-10, line 119: then method name says `Ports` so suggests that more than one port as parameter like a var arg, otherwise use the singular

fab-10 (changes requested):
just a comment on the method name, then a general question, this covers listening ports of the RPC server, are the other listening ports already covered?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -219,6 +219,8 @@ public CompletableFuture<?> start() {
 
     final CompletableFuture<?> resultFuture = new CompletableFuture<>();
     try {
+      NetworkUtility.checkIfPortIsAvailable(config.getPort());
```
jframe, line 222: Think this check would be better to happen sooner in BesuCommand otherwise this will need to be added in to multiple services. Perhaps near where we do the checkPortClash() call in BesuCommand configure method?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcService.java
```diff
@@ -219,6 +219,7 @@ public CompletableFuture<Void> start() {
 
     final CompletableFuture<Void> resultFuture = new CompletableFuture<>();
     try {
+      NetworkUtility.checkIfPortIsAvailable(config.getPort());
```
jframe, line 222: We also need this check to apply to engineJsonRpc service and probably also the graphQLHttp service.

jframe (dismissed):

util/src/main/java/org/hyperledger/besu/util/NetworkUtility.java
gfukushima: Changed the name of the method, thanks for raising that @fab-10

gfukushima: Changed the approach to cover the ports we will require to initialise all the services defined in the config. So we can have a more general solution that loops through the ports and check if they are available.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
gfukushima: Thanks for the feedback, I've changed the implementation after that.

fab-10 (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcService.java
gfukushima: N/A with the new approach I guess

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -3126,6 +3127,25 @@ private void checkPortClash() {
...
+        .forEach(
+            port -> {
+              if (!NetworkUtility.isPortAvailable(port)) {
+                unavailablePorts.add(port);
+              }
+            });
+    if (!unavailablePorts.isEmpty()) {
+      throw new InvalidConfigurationException(
+          "Port(s) '"
+              + unavailablePorts.toString()
```
jframe, line 3144: nit: toString on unavailablePorts isn't necessary

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -3126,6 +3127,25 @@ private void checkPortClash() {
             });
   }
 
+  private void checkIfRequiredPortsAreAvailable() {
+    List<Integer> unavailablePorts = new ArrayList<>();
```
jframe, line 3131: nit: final

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -219,6 +219,7 @@ public CompletableFuture<?> start() {
 
     final CompletableFuture<?> resultFuture = new CompletableFuture<>();
     try {
+
```
jframe, line 222: any reason to add the extra empty line?

util/src/main/java/org/hyperledger/besu/util/NetworkUtility.java
```diff
@@ -98,4 +105,31 @@ public static void checkPort(final int port, final String portTypeName) {
...
+      datagramSocket.bind(new InetSocketAddress(port));
+      return true;
+    } catch (IOException ex) {
+      LOG.trace(String.format("failed to open port %d for UDP", port), ex);
+    }
+    return false;
+  }
+
+  public static boolean isPortAvailable(final int port) {
+    if (isPortAvailableForTcp(port) && isPortAvailableForUdp(port)) return true;
```
jframe, line 132: Could simplify this to
`return isPortAvailableForTcp(port) && isPortAvailableForUdp(port)`

util/src/main/java/org/hyperledger/besu/util/NetworkUtility.java
```diff
@@ -98,4 +105,31 @@ public static void checkPort(final int port, final String portTypeName) {
...
+      datagramSocket.setReuseAddress(true);
+      datagramSocket.bind(new InetSocketAddress(port));
+      return true;
+    } catch (IOException ex) {
+      LOG.trace(String.format("failed to open port %d for UDP", port), ex);
+    }
+    return false;
+  }
+
+  public static boolean isPortAvailable(final int port) {
```
jframe, line 131: Worth adding a unit test for this method?

util/src/main/java/org/hyperledger/besu/util/NetworkUtility.java
```diff
@@ -98,4 +105,31 @@ public static void checkPort(final int port, final String portTypeName) {
               "%s port requires a value between 1 and 65535. Got %d.", portTypeName, port));
     }
   }
+
+  public static boolean isPortAvailableForTcp(final int port) {
```
siladu, line 109: Could be private?

util/src/main/java/org/hyperledger/besu/util/NetworkUtility.java
```diff
@@ -98,4 +105,31 @@ public static void checkPort(final int port, final String portTypeName) {
...
+      serverSocket.setReuseAddress(true);
+      serverSocket.bind(new InetSocketAddress(port));
+      return true;
+    } catch (IOException ex) {
+      LOG.trace(String.format("Failed to open port %d for TCP", port), ex);
+    }
+    return false;
+  }
+
+  public static boolean isPortAvailableForUdp(final int port) {
```
siladu, line 120: could be private?

util/src/main/java/org/hyperledger/besu/util/NetworkUtility.java
jframe, line 132: There could be a case where this reports a false positive if for example an existing service is using the udp port but we only require the tcp port. Would it make sense to just do the tpc or udp port that is necessary for the service rather than both?

gfukushima: Done, thanks!

gfukushima: Did some changes to only check p2p service will check both UDP and TCP. The rest of the services check TCP port only.

jframe (approved):

PR #4567, opened 2022-10-27
DRY-up *PendingTransactionSorter

fab-10:
## PR description

`GasPricePendingTransactionsSorter` and `BaseFeePendingTransactionsSorter` only differ in the way the prioritize pending transactions, so all common duplicated functionalities have been move up into `AbstractPendingTransactionsSorter`, to avoid code duplication and simplify future development.
Moreover `BaseFeePendingTransactionsSorter` now picks the lowest value transaction for eviction, and along with removing it, all the following transactions for the same sender are evict as well, to prevent nonce gaps.

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsSorter.java
```diff
@@ -443,7 +491,7 @@ public List<Transaction> signalInvalidAndGetDependentTransactions(final Transact
               pendingTx ->
                   traceLambda(
                       LOG,
-                      "Transaction {} piked for removal since there is a lowest invalid nonce {} for the sender",
+                      "Transaction {} piked since there is a lowest invalid nonce {} for the sender",
```
macfarla, line 494: Transaction {} invalid since there is a lower invalid nonce {} for the sender

?

macfarla (approved):
non-blocking suggestion on log message

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsSorter.java
fab-10: suggestion applied, tnx

PR #1784, opened 2021-01-11
Updated made to handle block parameter as an object

davemec:
## PR description
Updated made to handle block parameter as an object

## Fixed Issue(s)
#1733 

## Changelog

- [X] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #4582, opened 2022-11-01
RPC methods that lookup block by hash will now return error if no block found

macfarla:
Previously were returning null in this case eg ethGetCode

New behavior matches geth 1.10.25

Do we consider this a breaking change?

## Documentation

## Changelog


pinges (approved):
LGTM

shemnon (dismissed):
I argue this (and geth) are out of spec

https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbyhash

> **Returns**
>
>     Object - A block object, or null when no block was found:
>     [... description of fields ...]

shemnon:
Furthermore this touches on endpoints that use the "default block parameter."  Querying by Hash is not a valid option.  The data type is "BlockNumberOrTag" - https://github.com/ethereum/execution-apis/blob/main/src/eth/state.yaml#L60

And the type does not accept hashes - https://github.com/ethereum/execution-apis/blob/main/src/schemas/block.yaml#L92-L108

So if there was movement towards the spec conformance direction that step would be to not return answers when looking up by hash even when one is found.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugAccountAt.java
```diff
@@ -81,6 +81,7 @@ protected Object resultByBlockHash(
           requestContext.getRequest().getId(), JsonRpcError.BLOCK_NOT_FOUND);
     }
 
+    // transactions.isEmpty
```
Gabriel-Trintinalia, line 84: Is this line commented on purpose or was it just for testing?

Gabriel-Trintinalia (approved):
LGTM!

macfarla:
> I argue this (and geth) are out of spec
> 
> https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbyhash
> 
> > **Returns**
> > ```
> > Object - A block object, or null when no block was found:
> > [... description of fields ...]
> > ```
This change was made here https://github.com/hyperledger/besu/pull/1784 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1898.md

macfarla:
There has been further conversation here https://discord.com/channels/595666850260713488/595701319793377299/1041758135821533266 indicating that it should be part of the spec. 

EIP-1898 - changing status from stagnant to draft https://github.com/ethereum/EIPs/pull/5980 and adding relevant parts to the spec https://github.com/ethereum/execution-apis/pull/326

macfarla:
Update on this 
* EIP 1898 has moved (from stagnant) back to draft https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1898.md
* EIP 1898 requires EIP 1474 - PR here to move it to draft also https://github.com/ethereum/EIPs/pull/5986 - waiting on review
* changes to spec to allow block hash param also waiting on review https://github.com/ethereum/execution-apis/pull/326 | raised in Eth R&D discord https://discord.com/channels/595666850260713488/804019759934078987

macfarla:
This has been merged https://github.com/ethereum/execution-apis/pull/326 so next step is to verify that (with this PR) besu matches the spec

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugAccountAt.java
macfarla: removed

shemnon (approved):
The official execution spec and legacy JSON-RPC spec is disturbingly void of standard error handing.  I'll approve this on the presumption that this is not at odds with how other clients handle such an exception.

macfarla:
> The official execution spec is disturbingly void of standard error handing. I'll approve this on the presumption that this is not at odds with how other clients handle such an exception.

Thanks @shemnon - this PR actually arose out of an effort to standardise error handling across clients https://github.com/ethereum/execution-apis/issues/286

PR #4587, opened 2022-11-01
Update SLF4J version 

diega:
## PR description
Update [SLF4J](https://www.slf4j.org/) version to `2.0.6` matching the corresponding bindings for Log4j2. Also, I removed the [`o.h.b.u.Slf4jLambdaHelper`](https://github.com/hyperledger/besu/blob/22.7.7/util/src/main/java/org/hyperledger/besu/util/Slf4jLambdaHelper.java) using new idiomatics for dealing with lambda expressions as logging arguments.

This is medium sized PR, but it's composed of syntactic changes mostly performed using automated tools.

## Documentation

## Changelog

Signed-off-by: Diego López León <dieguitoll@gmail.com>


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
```diff
@@ -64,12 +62,12 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext request) {
...
-    traceLambda(
-        LOG,
-        "{} parameters: start block number {} count {}",
-        () -> getName(),
-        () -> startBlockNumber,
-        () -> count);
+    LOG.atTrace()
+        .setMessage("{} parameters: start block number {} count {}")
+        .addArgument(() -> getName())
+        .addArgument(() -> startBlockNumber)
```
shemnon, line 68: Are all the lamdas of variables, parameters, and fields needed?  (follow on, is there a good mechanical way to migrate it?)
```suggestion
        .addArgument(startBlockNumber)
```

shemnon (approved):
LGTM.  The movement to non-lambda fields/parameters/variables is optional.

gradle/verification-metadata.xml
```diff
@@ -1615,6 +1615,11 @@
             <sha256 value="24150c7dbdc8a5bb987cbeb4898ee1515d2893d3f6224acfc563da1a7fd5c283" origin="Generated by Gradle"/>
          </artifact>
       </component>
+      <component group="io.fabric8" name="kubernetes-client-bom" version="5.12.2">
```
siladu, line 1618: I'm surprised at all the changes in this file - are they are relevant for this PR?

siladu (approved):
Awesome change!

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
diega: What I did initially was to implement these in the `Slf4jLambdaHelper`, then inline those with IntelliJ and then unroll the loops. Another option would have been to replace the variable arguments methods with fixed one (I would say there are no more than 5 or 6 var args ever), implement those, and then inline. 
I'll take a look to remove the lambdas when not necessary, thanks for noticing it.

gradle/verification-metadata.xml
diega: yes, that surprised me too. Those are added by `org.apache.logging.log4j:log4j:2.20.0` 😞 ([here](https://repo1.maven.org/maven2/org/apache/logging/log4j/log4j/2.20.0/log4j-2.20.0.pom)'s the `log4j-2.20.0.pom`)

PR #4602, opened 2022-11-04
Explain and improve price validation for London transactions during s…

fab-10:
…election

## PR description

Explain and improve price validation for London transactions during block proposal selection.

For EIP1559 transactions, the price is dynamic and depends on network conditions, so we can
only calculate at this time the current minimum price the transaction is willing to pay
and if it is above the minimum accepted by the node.
If below we do not delete the transaction, since when we added the transaction to the pool,
we assured sure that the maxFeePerGas is >= of the minimum price accepted by the node
and so the price of the transaction could satisfy this rule in the future.

Also if a EIP1559 fails validation because its maxFeePerGas is below current baseFee, this should be considered a specific transient error, and not wrongly report INVALID_TRANSACTION_FORMAT.

Moreover this is applied only to remoteTransaction because local transactions are allowed to have gas price 
below the configured minimum, so we need to track local senders and do not enforce the rule for them.

Also fixed an edge case when a local transaction is readded to the pool due to a block reorg, and it was wrongly 
considered remote with the risk of being rejected due to low gas price.

## Documentation

## Changelog


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
```diff
@@ -342,14 +331,44 @@ private TransactionSelectionResult evaluateTransaction(
...
+      // and if it is above the minimum accepted by the node.
+      // If below we do not delete the transaction, since when we added the transaction to the pool,
+      // we assured sure that the maxFeePerGas is >= of the minimum price accepted by the node
+      // and so the price of the transaction could satisfy this rule in the future
+      final Wei currentMinTransactionGasPriceInBlock =
+          feeMarket
+              .getTransactionPriceCalculator()
+              .price(transaction, processableBlockHeader.getBaseFee());
+      if (minTransactionGasPrice.compareTo(currentMinTransactionGasPriceInBlock) > 0) {
+        LOG.debug(
```
shemnon, line 351: Volume could be very high, perhaps LOG.trace?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
```diff
@@ -358,10 +377,15 @@ private TransactionSelectionResult transactionSelectionResultForInvalidResult(
...
         "Delete invalid transaction {}, reason {}",
         transaction::toTraceLog,
-        invalidReasonValidationResult::getInvalidReason);
+        invalidReason::toString);
     return TransactionSelectionResult.DELETE_TRANSACTION_AND_CONTINUE;
   }
 
+  private boolean isTransientValidationError(final TransactionInvalidReason invalidReason) {
+    return invalidReason.equals(TransactionInvalidReason.GAS_PRICE_BELOW_CURRENT_BASE_FEE)
+        || invalidReason.equals(TransactionInvalidReason.INCORRECT_NONCE);
```
shemnon, line 386: Should we distinguish between nonce too low and nonce too high?  nonce too low will almost never become a valid transaction, especially if the nonce that used it's value is behind a finalized block.

shemnon (approved):
LGTM

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
fab-10: Good point, I have checked the `MainnetTransactionValidator` and found that INCORRECT_NONCE actually only means future nonce, since the check for nonce too low is done before and reported as NONCE_TOO_LOW, so nonce too low is not considered transient and will be removed, so we are aready good.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
fab-10: done

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
shemnon: can we rename incorrect nonce to NONCE_TOO_HIGH make it apparent on it's face that's what it's covering?  I want to remove the "secret knowledge" that incorrect_nonce only happens when nonce is to high.

fab-10: Make sense, and did the renaming, and since NONCE_TOO_HIGH was used when the nonce is above the max value, I had to renamve it to NONCE_OVERFLOW as well

shemnon (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -228,10 +229,26 @@ public void onBlockAdded(final BlockAddedEvent event) {
...
+                      tx -> pendingTransactions.isLocalSender(tx.getSender())));
+      var reAddLocalTxs = txsByOrigin.get(true);
+      var reAddRemoteTxs = txsByOrigin.get(false);
+      if (!reAddLocalTxs.isEmpty()) {
+        LOG.trace("Re-adding {} local transactions from a block event", reAddLocalTxs.size());
+        reAddLocalTxs.forEach(this::addLocalTransaction);
+      }
+      if (!reAddRemoteTxs.isEmpty()) {
+        LOG.trace("Re-adding {} remote transactions from a block event", reAddRemoteTxs.size());
+        addRemoteTransactions(reAddTransactions);
```
macfarla, line 250: why not addRemoteTransactions(reAddRemoteTxs)

fab-10: good catch! fixed

macfarla (approved):
LGTM

PR #4633, opened 2022-11-08
Remove SHL, SHR and SAR from default EVM

diega:
## PR description
This PR removes the opcodes `SHL`, `SHR` and `SAR` from the base EVM b/c those were added in [EIP-145](https://eips.ethereum.org/EIPS/eip-145) and shouldn't be available since Genesis.

## Fixed Issue(s)
Spotted by @winsvega during the execution of LegacyTests. E.g.
```shell
$ ./retesteth -t LegacyTests/Constantinople/BCGeneralStateTests/stShift -- --singletest shr11_d0g0v0 --clients besu --singlenet Byzantium --nodes 127.0.0.1:8545
```

## Documentation

## Changelog

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/EVM.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/EVM.java
@@ -43,9 +43,6 @@ import org.hyperledger.besu.evm.operation.PushOperation;
 import org.hyperledger.besu.evm.operation.SGtOperation;
 import org.hyperledger.besu.evm.operation.SLtOperation;
 import org.hyperledger.besu.evm.operation.SModOperation;
-import org.hyperledger.besu.evm.operation.SarOperation;
-import org.hyperledger.besu.evm.operation.ShlOperation;
-import org.hyperledger.besu.evm.operation.ShrOperation;
 import org.hyperledger.besu.evm.operation.SignExtendOperation;
 import org.hyperledger.besu.evm.operation.StopOperation;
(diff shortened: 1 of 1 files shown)


shemnon (approved):
Good Catch, we should remove this from the switch or place a guard around the result.

Can we keep the static operation methods in the operation classes?  When EVM lands I will want to re-add it with a guard such as seen here for [push0](https://github.com/shemnon/besu/blob/EOF/evm/src/main/java/org/hyperledger/besu/evm/EVM.java#L218-L222)

diega:
Yes, hopefully we could run the whole suite from the `retesteth` sometime soon.

I reintroduced the static methods, but what do you mean by "remove this from the switch or place a guard around the result"? Didn't I remove those from the proper switch?

winsvega:
Besu runs almost all tests. It appears to be that some failures are due to the geth t8n version. Minor theoretical issues.

shemnon:
The linked code shows how I handle activation of Push0 to keep it in the switch.  So they would look something like

```
          case 0x1b: // SHL
            result =
                enableConstantinople
                    ? ShlOperation.staticOperation(frame)
                    : InvalidOperation.INVALID_RESULT;
```

But some of the framework supporting that is tied up in my EOF branch.

diega:
@shemnon gotcha. Probably this is a discussion for the other PR about EOF, but I think I like the current resolution of operations from a registry and to keep it agnostic of fork names

shemnon (approved):

Issue #4549, opened 2022-10-20
Besu block import an order of magnitude slower than Geth (side by side comparison)


kayagoban:
### Description
As a validator, I would like Besu to match Geth's performance

### Acceptance Criteria
Block import times should be comparable to Geth

### Steps to Reproduce (Bug)
1. Install geth and besu, each with its own prysm instance as CL.  Everything is installed on a single host.  Fully sync all instances.
2. View syslog and compare block import times

**Expected behavior:**
Besu import time is similiar to geth

**Actual behavior:** 
Besu block import is an order of magnitude slower than geth

**Frequency:** 
100%

### Versions (Add all that apply)
* Software version: 22.7.7
* Java version:
openjdk 17.0.4 2022-07-19
OpenJDK Runtime Environment (build 17.0.4+8-Debian-1deb11u1)
OpenJDK 64-Bit Server VM (build 17.0.4+8-Debian-1deb11u1, mixed mode, sharing)
* OS Name & Version: Debian GNU/Linux 11
* Kernel Version:   Linux besu 5.18.0-0.deb11.3-amd64 #1 SMP PREEMPT_DYNAMIC Debian 5.18.14-1~bpo11+1 (2022-07-28) x86_64 GNU/Linux
* Virtual Machine software & version: ESXi 6.5

### Additional Information (Add any of the following or anything else that may be relevant)
* Besu options:
[Service]

Environment="BESU_OPTS=-Xmx8g"

ExecStart=/opt/app/besu-22.7.7/bin/besu \
	--data-path=/opt/besu-data \
	--network=MAINNET \
	--rpc-http-host=127.0.0.1  \
	--rpc-http-apis="ADMIN,ETH,NET,WEB3" \
	--rpc-http-enabled=true \
	--host-allowlist="*" \
	--engine-jwt-secret=/opt/jwtsecret \
	--engine-rpc-enabled=true \
	--engine-rpc-http-port=8552 \
	--data-storage-format=BONSAI \
	--sync-mode=X_CHECKPOINT \
	--p2p-host=xxx.xxx.xxx.xxx \
	--max-peers=50 \
	--p2p-port=xxxxx \
	--Xplugin-rocksdb-high-spec-enabled=true \
	--nat-method=UPNP

* Colocated 1U Server, 50mbit combined; 10cpu Xeon + 64GB + 2.6TB SSD+RAID6

--------
[...]

PR #4635, opened 2022-11-09
Optimize performance of WorldStateUpdater commit method

ahamlat:

## PR description
WorldStateUpdater commit method is one of the most consuming methods during block processing (engine_newPayloadV1 call). This PR will focus on parallelizing some parts of this method to make it faster.

image

The first results after 2 hours of execution

image

The results after 12 hours
image

We can see with CPU profiling that the commit methods takes less time and most of the execution is done in parallel in ForkjoinPool.commonPool().

image

image

## Fixed Issue(s)

Mitigate #4549

## Documentation

## Changelog


shemnon (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Additions and Improvements
 - Explain and improve price validation for London and local transactions during block proposal selection [#4602](https://github.com/hyperledger/besu/pull/4602)
 - Support for ephemeral testnet Shandong.  EIPs are still in flux, besu does not fully sync yet, and the network is subject to restarts. [#//FIXME](https://github.com/hyperledger/besu/pull///FIXME)
+- Improve performance of block processing during the "commit" step [#4635](https://github.com/hyperledger/besu/pull/4635)
```
shemnon, line 10: Should get the word parallel... "Parallelized commit step during block processing to improve performance."

macfarla (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -219,76 +217,82 @@ public void commit() {
...
+              if (tracked.getStorageWasCleared()) {
+                storageToClear.add(updatedAddress);
+                pendingStorageUpdates.clear();
+              }
+
+              final TreeSet<Map.Entry<UInt256, UInt256>> entries =
+                  new TreeSet<>(Map.Entry.comparingByKey());
+              entries.addAll(updatedAccount.getUpdatedStorage().entrySet());
+
+              entries.forEach(
```
garyschulte, line 270: why not `entries.parallelStream().forEach(` ?  can we not parallelize the storage update also  or do we think there are diminishing returns here?

garyschulte (approved):
LGTM, rebase, feedback and 🚢

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
ahamlat: This was one of the tests we did, you can find it if you take a look to the first commit in this PR. We had with the database corrupted with this test.

CHANGELOG.md
ahamlat: Done.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
daniellehrner: In this case could we maybe add a comment why we are not using `parallelStream`? I think somebody could see this code and try it again. With a comment warning that it will not work, we should save time.

PR #4640, opened 2022-11-09
♻️ Refactor EthProtocol to externalise EthVersion into its own class

Gabriel-Trintinalia:
## PR description
Refactor EthProtocol to externalise EthVersion into its own class to improve readability

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright ConsenSys AG.
+ * Copyright contributors to Hyperledger Besu
```
macfarla, line 2: nit: we don't need to change the copyright for existing files. (not sure it's worth changing back though)

macfarla (approved):
LGTM. non-blocking comment re copyright

PR #4654, opened 2022-11-11
Adding option to debug the potential problems with closing Transaction

gezero:
If we are adding finalize, at least add some way to investigate a potential not closed object. 


garyschulte (approved):
small change to use debugLambda, :ship:

PR #4657, opened 2022-11-11
Ignore .idea directory in spotless checks

daniellehrner:

## PR description

InteliJ allows to define templates for different files like Java classes, interfaces,... If those templates are saved in the project they are saved as `.java` files in a subdirectory of `.idea`.

Such a template looks for example like this:
```
#parse("Copyright Besu.java")
#if (${PACKAGE_NAME} && ${PACKAGE_NAME} != "")package ${PACKAGE_NAME};#end
public class ${NAME} {
}
```
Spotless currently tries to check this templates, because it is configured to analyze all Java files. This check will fail, because the templates are not valid Java code.

This PR excludes all the files in the `.idea` directory from the spotless check to avoid those false positives.

## Documentation

## Changelog


fab-10 (approved):

build.gradle
```diff
@@ -129,7 +129,7 @@ allprojects {
     java {
       // This path needs to be relative to each project
       target '**/*.java'
-      targetExclude '**/src/reference-test/**', '**/src/main/generated/**', '**/src/test/generated/**', '**/src/jmh/generated/**'
+      targetExclude '**/src/reference-test/**', '**/src/main/generated/**', '**/src/test/generated/**', '**/src/jmh/generated/**', '.idea/**'
```
shemnon, line 132: Rather than one-by-one exclusions we should have a more focused inclusion.
```suggestion
      target 'src/**/*.java'
      targetExclude 'src/reference-test/**', 'src/main/generated/**', 'src/test/generated/**', 'src/jmh/generated/**'
```

shemnon (changes requested):

build.gradle
daniellehrner: You are right. That is much better. I've changed it.

shemnon (approved):

Issue #4658, opened 2022-11-11
Implement EIP 3855 PUSH0 instruction


lukelee-sl:
https://eips.ethereum.org/EIPS/eip-3855

### Description
See referenced EIP

### Acceptance Criteria
Pass acceptance tests that exercise the PUSH0 operation

shemnon:
Reference tests are sitting in this PR at the moment - https://github.com/ethereum/tests/pull/1033

PR #4660, opened 2022-11-11
Implementation of  PUSH0

lukelee-sl:
Signed-off-by: lukelee-sl <luke.lee@swirldslabs.com>

## PR description
Implement PUSH0 operation as described in https://eips.ethereum.org/EIPS/eip-3855

## Fixed Issue(s)
#4658

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/EVM.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/EVM.java
@@ -203,7 +203,8 @@ public class EVM {
           case 0x50: // POP
             result = PopOperation.staticOperation(frame);
             break;
-          case 0x60: // PUSH1-32
+          case 0x5f: // PUSH0-32
+          case 0x60:
           case 0x61:
           case 0x62:
           case 0x63:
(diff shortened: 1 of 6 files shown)


evm/src/main/java/org/hyperledger/besu/evm/EVM.java
```diff
@@ -203,7 +203,8 @@ public void runToHalt(final MessageFrame frame, final OperationTracer tracing) {
           case 0x50: // POP
             result = PopOperation.staticOperation(frame);
             break;
-          case 0x60: // PUSH1-32
+          case 0x5f: // PUSH0-32
```
shemnon, line 206: We need to make sure Push0 is not run pre-Shanghai.  So this optimization needs to go or be guarded, see https://github.com/hyperledger/besu/pull/4633

evm/src/main/java/org/hyperledger/besu/evm/operation/PushOperation.java
```diff
@@ -49,6 +50,13 @@ public Operation.OperationResult executeFixedCostOperation(
 
   public static OperationResult staticOperation(
       final MessageFrame frame, final byte[] code, final int pc, final int pushSize) {
+    // Handle the PUSH0 special case
+    if(pushSize == 0) {
```
shemnon, line 54: Push is the most used instruction in the EVM, hands down, hence this is super hot.  Can this be refactored into a separate Push0Operation?

shemnon (changes requested):

shemnon (approved):

PR #4661, opened 2022-11-11
22.10.1 alpha

garyschulte:
## PR description

alpha build for burn in ahead of 22.10.1 release.  

This PR:
* merges main from e7074c0ecb582b50055c1928da85a2e79e9f97ba into release-22.10.x 
* sets gradle.properties version to 22.10.1-alpha


jflo (approved):
godspeed ser

shemnon:
To be clear, the policy of releasing patches off of main hasn't changed, right?

garyschulte:
> To be clear, the policy of releasing double-dot releases off of main hasn't changed, right?

Not yet.  I think we should move to a release process that is better tested, but doing these burn-ins at least can help us gain confidence ahead of a release from main.

shemnon:
We used to cut the release, keep it quiet, run out syncs and burn-ins, then announce or squelch the release based on results.

PR #4669, opened 2022-11-14
Some cleanup in Discovery

pinges:
Some cleanup: The PeerBondedObserver was only ever used to call the connect() method on the RlpxAgent


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -266,8 +266,8 @@ private PeerDiscoveryController createController(final DiscoveryPeer localNode)
         .workerExecutor(createWorkerExecutor())
         .peerRequirement(PeerRequirement.combine(peerRequirements))
         .peerPermissions(peerPermissions)
-        .peerBondedObservers(peerBondedObservers)
         .metricsSystem(metricsSystem)
+        .rlpxAgent(rlpxAgent)
```
macfarla, line 270: it's not polluting the layers to have the RlpxAgent passed in here?

macfarla (commented):
I think it's ok - the only question I have is whether we want the PeerDiscoveryAgent to have access to the RlpxAgent - maybe get a second opinion

macfarla (approved):
LGTM

PR #4677, opened 2022-11-15
bugfix for 22.10.1 alpha

garyschulte:
## PR description
merge bugfix commit from #4468

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
@@ -258,20 +258,39 @@ public class BonsaiLayeredWorldState implements MutableWorldState, BonsaiWorldVi
 
   @Override
   public MutableWorldState copy() {
-    // TODO:  this is a stopgap measure, revisit with
-    // https://github.com/hyperledger/besu/issues/4641
-    // rip off the band-aid.  you know you want to:
+    // TODO: this is unsafe because the underlying persisted worldstate will likely have mutated if
+    // this
+    //  layer is not the current head
+    //  revisit with https://github.com/hyperledger/besu/issues/4641
+    final BonsaiPersistedWorldState bonsaiPersistedWorldState =
+        ((BonsaiPersistedWorldState) archive.getMutable());
+    BonsaiInMemoryWorldStateKeyValueStorage bonsaiInMemoryWorldStateKeyValueStorage =
+        new BonsaiInMemoryWorldStateKeyValueStorage(
+            bonsaiPersistedWorldState.getWorldStateStorage().accountStorage,
+            bonsaiPersistedWorldState.getWorldStateStorage().codeStorage,
+            bonsaiPersistedWorldState.getWorldStateStorage().storageStorage,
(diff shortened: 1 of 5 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -258,20 +258,39 @@ public Stream<StreamableAccount> streamAccounts(final Bytes32 startKeyHash, fina
 
   @Override
   public MutableWorldState copy() {
-    // TODO:  this is a stopgap measure, revisit with
-    // https://github.com/hyperledger/besu/issues/4641
-    // rip off the band-aid.  you know you want to:
+    // TODO: this is unsafe because the underlying persisted worldstate will likely have mutated if
```
siladu, line 261: Can you contextualise 'unsafe'? Or at least explain why this is safe to merge into alpha?

siladu (commented):
Commit looks good comparing to https://github.com/hyperledger/besu/pull/4648/commits/1cdf6c4e4799e616a25baeced8dda99fd7ce8c0e

Just wondering about the 'unsafe' comment

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
garyschulte: Lots of back story about why this is unsafe.  TL/DR It is a problem for reorgs and concurrent operations on the worldstate like block creation. 

To illustrate, a real scenario we have seen frequently:

chain is at block A, worldstate at a
newPayload comes in for block B, mainnet validator processes and persists worldstate b
forkchoiceUpdate comes in that sets head to block B, worldstate to b
forkchoiceUpdate comes in that sets head back to block A, which moves worldstate to a and asks for a block proposal B'

Then, concurrently:
 block proposal B' starts building and executing on parent worldstate a
 concurrently a forkchoice update re-sets head to B and updates the persisted worldstate to b
 boom, stateroot mismatch.  Maybe for the block proposal thread, maybe for processing the  subsequent block C
 
The behavior prior to this PR was:
 1. get a mutable worldstate (this rolls the single persisted worldstate to that blockhash)
 2. if we were not persisting (like just validating) then get an in-memory worldstate that defers to the persisted worldstate for changes that are not yet in-memory. 

Any changes to the persisted worldstate after that in-memory worldstate was created are going to cause a stateroot mismatch for operations on the in-memory worldstate.
  
This PR keeps that unsafe behavior for non-snapshot bonsai, but moves the persisted state rolling into the `copy` method of the layered worldstate rather than having MainnetBlockValidator mutate the persisted worldstate and then call copy on the result.  It is just as unsafe as before, but it doesn't leak this behavior into the bonsai snapshots implementation.  

There isn't a simple or safe way to get an in-memory worldstate that is based on a concurrently mutating underlying database (other than using snapshots). Since snapshots are the way forward, I just want to retain the existing unsafe behavior until we fully remove it and snapshots are more battle tested.

siladu: Great explanation, thanks!

siladu (approved):
LGTM

PR #4682, opened 2022-11-15
Enable RocksDB bloomFilters

ahamlat:
Signed-off-by: Ameziane H <ameziane.hamlat@consensys.net

## PR description
Enable RocksDB full bloom filters, with 10 bits per key.
The benefit from enabling Bloom filters is to improve read performance, more details [here](https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#usage-of-new-bloom-filter). I believe though this will work better if Besu is synced from the beginning with this configuration, because Bloom filters are created when new SST files are created.

Here are the results when we compare median (50th percentile) and 95th percentile values on newPayload calls.

image

We can see a [small overhead in CPU consumption](https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#usage-of-new-bloom-filter), but this doesn't affect block processing time as we still have a room for CPU usage.  

With Bloom Filters
image

Without Bloom filters
Screenshot 2022-11-18 at 09 01 27

Memory usage is quite similar, in the screenshot below we can see a little difference but not sure it is related to bloom filters

image

## Documentation

## Changelog


plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -68,7 +69,6 @@
   private static final Logger LOG = LoggerFactory.getLogger(RocksDBColumnarKeyValueStorage.class);
   private static final String DEFAULT_COLUMN = "default";
   private static final String NO_SPACE_LEFT_ON_DEVICE = "No space left on device";
-  private static final int ROCKSDB_FORMAT_VERSION = 5;
```
fab-10, line 71: why this is removed?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -175,9 +175,9 @@ private BlockBasedTableConfig createBlockBasedTableConfigHighSpec() {
     final LRUCache cache = new LRUCache(ROCKSDB_BLOCKCACHE_SIZE_HIGH_SPEC);
     return new BlockBasedTableConfig()
         .setBlockCache(cache)
-        .setFormatVersion(ROCKSDB_FORMAT_VERSION)
-        .setOptimizeFiltersForMemory(true)
-        .setCacheIndexAndFilterBlocks(true)
```
fab-10, line 180: are these other changes required for bloom filters or are they uncorrelated?

ahamlat: Yes, these are related to bloom filters,   setCacheIndexAndFilterBlocks is by default set to false to avoid putting in cache indexes and filters, this will have an impact on cache hit ratio for other cached data (real data). I did tests with setOptimizeFiltersForMemory(true) and without and didn't see any difference in terms of memory consumption, as I wanted to keep the configuration simple, I delete it.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
ahamlat: I think this is done by error, we should keep it. I will update this part, thanks.

fab-10 (approved):

Issue #4476, opened 2022-10-04
Bonsai lightweight mode to prune old chain data


jframe:
Bonsai lightweight mode to prune old blocks, header and transaction receipts as we only require near-head at least 256 blocks of data.

Acceptance Criteria
* Ensure the pruning doesn't affect performance of Besu. This is lower priority than block processing/import
* Storage is maintained at a small size by pruning old transaction receipts by only keeping HEAD/finalised -x of the transaction receipts.
* Ensure that block headers, block body and transaction receipts are deleted atomically i.e. using rocksdb transactions
* Can survive a chain reorg i.e. can handle forks
* Ensure that syncing continues to work when pivot block is behind the pruning block

Tasks
* Hidden feature flag for bonsai pruning and configuration for number of blocks to prune
* When the threshold is reach for historical data then prune old transaction receipts below threshold
* Create pruning scheduler so that it does not affect the performance of Besu particularly block import/block execution
* Add remove method to KeyValueStoragePrefixedKeyBlockchainStorage for transaction receipts
* Test syncing for checkpoint
* Measure eth_call performance using eth_gen to ensure is not negatively affected by pruning
* Acceptance test/integration test for pruning

Question
* Should this be a seperate process or built-in to the block processing?
  * Decided on using seperate process so more control on when the pruning happens for performance and fulfilling acceptance criteria

Notes
* Lower priority process than block import. If possible prune when idle
* Existing pruning is for Forest mode world state only and it's approach of traversing the trie is not applicable for the bonsai blockchain data pruning we need to do.

PR #4686, opened 2022-11-16
Add chain data pruning experimental feature

wcgcyx:
## PR description
This PR introduces a new feature that allows besu to prune old chain data. Three experimental CLI options are introduced as well: `--Xchain-pruning-enabled` to turn on/off pruning, `--Xchain-pruning-blocks-retained` to specify the number of recent blocks to retain and `--Xchain-pruning-frequency` to specify how often to perform chain data pruning.

## Fixed Issue(s)

Linked to #4476 

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Shanghai implementation of EIP-3540 and EIP-3670 Ethereum Object Format and Code Validation [#4644](https://github.com/hyperledger/besu/pull/4644)
 - Remove some log statements that are keeping some objects live in heap for a long time, to reduce the amount of memory required during initial sync [#4705](https://github.com/hyperledger/besu/pull/4705)
 - Add field `type` to Transaction receipt object (eth_getTransactionReceipt) [#4505](https://github.com/hyperledger/besu/issues/4505)
+- Add chain data pruning feature with three experimental CLI options: `--Xchain-data-pruning-blocks-retained`, `--Xchain-data-pruning-blocks-retained` and `--Xchain-data-pruning-frequency` [#4686](https://github.com/hyperledger/besu/pull/4686)
```
usmansaleem, line 22: nit: `--Xchain-data-pruning-blocks-retained` repeated twice. Should be `enabled` ?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1894,6 +1899,20 @@ public void validateRpcOptionsParams() {
     }
   }
 
+  public void validateChainDataPruningParams() {
+    if (Boolean.TRUE.equals(unstableChainDataPruningOptions.getChainDataPruningEnabled())) {
+      if (unstableChainDataPruningOptions.getChainDataPruningBlocksRetained()
+          < ChainDataPruningOptions.DEFAULT_CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED) {
+        throw new ParameterException(
+            this.commandLine, "--Xchain-data-pruning-blocks-retained must be >= 50400");
```
usmansaleem, line 1907: nit: Append `ChainDataPruningOptions.DEFAULT_CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED` to exception message instead of literal `50400`.

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ChainDataPruningOptions.java
```diff
@@ -0,0 +1,62 @@
...
+      names = {"--Xchain-data-pruning-enabled"},
+      description =
+          "Enable the chain pruner to actively prune old chain data (default: ${DEFAULT-VALUE})")
+  private final Boolean chainDataPruningEnabled = Boolean.FALSE;
+
+  @CommandLine.Option(
+      hidden = true,
+      names = {"--Xchain-data-pruning-blocks-retained"},
+      description =
+          "The number of recent blocks for which to keep the chain data. Must be >= 50400 (default: ${DEFAULT-VALUE})")
```
usmansaleem, line 36: nit: instead of using literal `>= 50400`, append static constant to description?

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -300,6 +320,23 @@ public BesuController build() {
             reorgLoggingThreshold,
             dataDirectory.toString());
 
+    if (isChainDataPruningEnabled) {
+      ChainDataPruner.enablePruning();
+      ChainDataPruner chainDataPruner =
```
usmansaleem, line 325: nit: final

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,186 @@
...
+    pruningEnabled = true;
+  }
+
+  public static boolean isPruningEnabled() {
+    return pruningEnabled;
+  }
+
+  @Override
+  public void onBlockAdded(final BlockAddedEvent event) {
+    LOG.debug("New block added event: " + event);
```
usmansaleem, line 73: should we check for pruningEnabled and early return if false?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueSegmentIdentifier.java
```diff
@@ -35,7 +35,8 @@ public enum KeyValueSegmentIdentifier implements SegmentIdentifier {
   BACKWARD_SYNC_BLOCKS(new byte[] {14}),
   BACKWARD_SYNC_CHAIN(new byte[] {15}),
   SNAPSYNC_MISSING_ACCOUNT_RANGE(new byte[] {16}),
-  SNAPSYNC_ACCOUNT_TO_FIX(new byte[] {17});
+  SNAPSYNC_ACCOUNT_TO_FIX(new byte[] {17}),
+  CHAIN_PRUNER_STATE(new byte[] {18});
```
usmansaleem, line 39: will this make database backward incompatible?

CHANGELOG.md
wcgcyx: yep!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: pruningEnabled is only used as a temporary workaround to skip the check before https://github.com/hyperledger/besu/pull/4703 is merged. I'll leave a TODO in the code to remind us to do a cleanup later.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueSegmentIdentifier.java
wcgcyx: Yea, I think so. I'm just adding a new segment so it should be backward-compatible.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  private void remove(
+      final KeyValueStorageTransaction transaction, final Bytes prefix, final Bytes key) {
+    transaction.remove(Bytes.concatenate(prefix, key).toArrayUnsafe());
+  }
+
+  private Hash bytesToHash(final Bytes bytes) {
+    return Hash.wrap(Bytes32.wrap(bytes, 0));
+  }
```
jframe, line 187: Would be cleaner to move the pruning storage out into it's own class

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,188 @@
...
+        && blockNumber - blocksToRetain - pruningMark >= pruningFrequency) {
+      while (blockNumber - pruningMark >= blocksToRetain) {
+        LOG.debug("Pruning chain data at pruning mark: " + pruningMark);
+        // Get a collection of old fork blocks that need to be pruned.
+        Collection<Hash> oldForkBlocks = getForkBlocks(pruningMark);
+        BlockchainStorage.Updater updater = blockchainStorage.updater();
+        for (Hash toPrune : oldForkBlocks) {
+          Optional<BlockBody> maybeBody = blockchainStorage.getBlockBody(toPrune);
+          if (maybeBody.isEmpty()) {
+            continue;
```
jframe, line 112: It could be possible, though maybe not likely that there is a block header, transaction receipts etc. might still need to be pruned. A safer approach could be to just not remove transaction locations if we can't get the body as don't have the detail to find them.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,188 @@
...
+    // Append block into fork blocks.
+    KeyValueStorageTransaction tx = prunerStorage.startTransaction();
+    Collection<Hash> forkBlocks = getForkBlocks(blockNumber);
+    forkBlocks.add(event.getBlock().getHash());
+    setForkBlocks(tx, blockNumber, forkBlocks);
+    // If a block is a new canonical head, start pruning.
+    if (event.isNewCanonicalHead()
+        && blockNumber - blocksToRetain - pruningMark >= pruningFrequency) {
+      while (blockNumber - pruningMark >= blocksToRetain) {
+        LOG.debug("Pruning chain data at pruning mark: " + pruningMark);
```
jframe, line 105: Perhaps move the body of the while loop out to make this function a bit smaller

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+    // Append block into fork blocks.
+    KeyValueStorageTransaction tx = prunerStorage.startTransaction();
+    Collection<Hash> forkBlocks = getForkBlocks(blockNumber);
+    forkBlocks.add(event.getBlock().getHash());
+    setForkBlocks(tx, blockNumber, forkBlocks);
+    // If a block is a new canonical head, start pruning.
+    if (event.isNewCanonicalHead()
+        && blockNumber - blocksToRetain - pruningMark >= pruningFrequency) {
+      while (blockNumber - pruningMark >= blocksToRetain) {
```
jframe, line 104: Wondering if we should do the pruning on another thread as this could affect the performance of the syncing and engine APIs.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1894,6 +1899,22 @@ public void validateRpcOptionsParams() {
     }
   }
 
+  public void validateChainDataPruningParams() {
+    if (Boolean.TRUE.equals(unstableChainDataPruningOptions.getChainDataPruningEnabled())) {
```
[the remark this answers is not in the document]
wcgcyx: updated

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1894,6 +1899,22 @@ public void validateRpcOptionsParams() {
...
+      if (unstableChainDataPruningOptions.getChainDataPruningBlocksRetained()
+          < ChainDataPruningOptions.DEFAULT_CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED) {
+        throw new ParameterException(
+            this.commandLine,
+            "--Xchain-data-pruning-blocks-retained must be >= "
+                + ChainDataPruningOptions.DEFAULT_CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED);
+      }
+      if (unstableChainDataPruningOptions.getChainDataPruningBlocksFrequency() < 0) {
+        throw new ParameterException(
```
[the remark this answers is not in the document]
wcgcyx: updated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,188 @@
...
+import org.apache.tuweni.units.bigints.UInt256;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ChainDataPruner implements BlockAddedObserver {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ChainDataPruner.class);
+
+  private static final Bytes PRUNING_MARK_KEY =
+      Bytes.wrap("blockNumberTail".getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
wcgcyx: updated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,188 @@
...
+  public void onBlockAdded(final BlockAddedEvent event) {
+    LOG.debug("New block added event: " + event);
+    // Get pruning mark
+    long blockNumber = event.getBlock().getHeader().getNumber();
+    Optional<Long> maybePruningMark = getPruningMark();
+    if (maybePruningMark.isEmpty()) {
+      // Set initial pruning mark
+      maybePruningMark = Optional.of(blockNumber);
+    }
+    long pruningMark = maybePruningMark.get();
```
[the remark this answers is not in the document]
wcgcyx: updated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: updated so that tx location gets pruned the last

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: updated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: I need to think about this. The only concern I have right now is that during initial sync, blocks are added at a very fast rate so it is possible that multiple threads could perform pruning at the same time and pruning operations may not be in order. Not sure if this would corrupt the pruner state.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: Added a separate class for the storage

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
```diff
@@ -116,7 +117,8 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
     }
 
     // TODO: post-merge cleanup, this should be unnecessary after merge
-    if (!mergeCoordinator.latestValidAncestorDescendsFromTerminal(newHead.get())) {
+    if (!mergeCoordinator.latestValidAncestorDescendsFromTerminal(newHead.get())
+        && !ChainDataPruner.isPruningEnabled()) {
```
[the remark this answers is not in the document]
wcgcyx: Yea, I tried that a bit but it seems like that would involve bringing besu command into the protocol context. I consider it to be a temporary solution until #4703 is merged. Then, we could just do a cleanup to get rid of this global field completely.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -300,6 +320,23 @@ public BesuController build() {
             reorgLoggingThreshold,
             dataDirectory.toString());
 
+    if (isChainDataPruningEnabled) {
+      ChainDataPruner.enablePruning();
```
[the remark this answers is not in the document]
wcgcyx: See below

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
```diff
@@ -194,7 +195,8 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
     }
 
     // TODO: post-merge cleanup
-    if (!mergeCoordinator.latestValidAncestorDescendsFromTerminal(newBlockHeader)) {
+    if (!mergeCoordinator.latestValidAncestorDescendsFromTerminal(newBlockHeader)
```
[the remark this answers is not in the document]
wcgcyx: I don't think so. From my understanding, this is just to make sure the transition from pow to pos works fine. It seems like a duplicate check to me post-merge because just above this code, we've checked if the new block connects to the current chain so when the code reaches here, we can be sure the new block must decent from the terminal block.

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ChainDataPruningOptions.java
```diff
@@ -0,0 +1,64 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+package org.hyperledger.besu.cli.options.unstable;
+
+import picocli.CommandLine;
+
+public class ChainDataPruningOptions {
+
+  public static final long DEFAULT_CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED = 50400;
```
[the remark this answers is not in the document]
wcgcyx: Yes, ideally if the pruner has the syncing context to know the current pivot block and if the syncing has finished, then it can avoid pruning beyond the pivot block before syncing is completed. Do you think we should do it in a separate ticket?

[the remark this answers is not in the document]
jframe: I think we should do this in another ticket.

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ChainDataPruningOptions.java
```diff
@@ -0,0 +1,67 @@
...
+  public static final int DEFAULT_CHAIN_DATA_PRUNING_FREQUENCY = 256;
+
+  @CommandLine.Option(
+      hidden = true,
+      names = {"--Xchain-data-pruning-enabled"},
```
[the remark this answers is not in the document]
wcgcyx: Yea, the reason for having the data here is to emphasize that all data related to the chain will be pruned. For example, besu stores transaction location as part of a block import, which isn't actually part of the chain but is a piece of data related to the chain.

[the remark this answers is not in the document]
siladu: I'd argue transaction location is an implementation detail we shouldn't encode into this name and so data is still redundant.
 
I think there's an important distinction between this pruning vs world state pruning and I'm not sure the average user would understand that from "chain-data". One may assume world state is part of the "chain".

What are your thoughts on "block-data"?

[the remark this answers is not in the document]
wcgcyx: That's true also. I'd say it is better to rename the existing pruning to something like "state-pruning" once this becomes a visible option to not confuse users. 

I would prefer "chain-pruning"/"chain-data-pruning" over "block-data-pruning" as we are pruning the chain by removing old blocks from it. The latter may create the impression that we are pruning some data from blocks. 

Wondering about the thoughts of others regarding the naming @jframe @usmansaleem @gfukushima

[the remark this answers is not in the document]
siladu: That's true. Maybe "block-pruning" is more accurate than "block-data-pruning". 
It hides the details about pruning other data associated with the block but I think that's fine.

I'd also be fine with "chain-pruning". 
Confusion with state pruning probably isn't that big a deal, especially since the users of this will be on bonsai so shouldn't be configuring state pruning (although some people still try!)

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1912,6 +1917,18 @@ public void validateRpcOptionsParams() {
     }
   }
 
+  public void validateChainDataPruningParams() {
```
siladu, line 1920: Should we also check that they're not trying to use it with FOREST mode? (or only allow it for BONSAI)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueSegmentIdentifier.java
wcgcyx: Did manual testing earlier, besu with this new column has no problem mounting to a database created by the old besu.

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ChainDataPruningOptions.java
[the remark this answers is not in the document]
wcgcyx: Cool, just updated

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
wcgcyx: Technically it is fine to use with FOREST

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: I've just got the PR updated so pruning is done in a separate thread.

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ChainDataPruningOptions.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+package org.hyperledger.besu.cli.options.unstable;
+
+import org.hyperledger.besu.util.number.PositiveNumber;
+
+import picocli.CommandLine;
+
+public class ChainDataPruningOptions {
+
+  public static final long DEFAULT_CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED = 50400;
```
jframe, line 24: Rename to MAX_CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED? This not just the default but also the maximum number of blocks that will be pruned.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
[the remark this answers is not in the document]
jframe: We shouldn't need to bring the entire Besu command into the protocol context. At most, the chain pruning config values would need to be passed through. A compromise could be to make the ChainPruner a singleton instead.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  public static boolean isPruningEnabled() {
+    return pruningEnabled;
+  }
+
+  @Override
+  public void onBlockAdded(final BlockAddedEvent event) {
+    LOG.debug("New block added event: " + event);
+    new Thread(
```
jframe, line 61: This could possibly end up creating an unlimited number of threads. Suggest that we use an ExecutorService instead. The `MonitoredExecutors` could be useful for creating an executor that has metrics.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,126 @@
...
+
+  private final long pruningFrequency;
+
+  public ChainDataPruner(
+      final BlockchainStorage blockchainStorage,
+      final KeyValueStorage storage,
+      final long blocksToRetain,
+      final long pruningFrequency) {
+    this.blockchainStorage = blockchainStorage;
+    this.prunerStorage = new ChainDataPrunerStorage(storage);
```
jframe, line 45: Looks like we don't need the KeyValueStorage other than to construct the ChainDataPrunerStorage. Can a fully created ChainDataPrunerStorage just be passed in instead?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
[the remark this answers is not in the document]
wcgcyx: Actually, It seems cleaner to add a flag in merge context instead, following what @gfukushima did in #4735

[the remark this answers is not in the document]
wcgcyx: updated to have a temporary flag in protocol context instead. Global variable gets removed.

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ChainDataPruningOptions.java
wcgcyx: This is actually the minimum value for the blocks to retain. I've also changed it to be much smaller.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: Just tried to use MonitoredExecutors but this introduces a circular dependency between :ethereum:core and :ethereum:eth modules

wcgcyx: Updated & switched to ThreadPoolExecutor to limit the number of threads.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: updated

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -140,6 +142,9 @@ public abstract class BesuControllerBuilder implements MiningParameterOverrides
       Collections.emptyList();
   protected EvmConfiguration evmConfiguration;
   protected int maxPeers;
+  protected boolean isChainDataPruningEnabled;
+  protected long chainDataPruningBlocksRetained;
+  protected long chainDataPruningFrequency;
```
siladu, line 147: Could maybe wrap in a ChainDataPruningConfiguration object?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+            1,
+            1,
+            60L,
+            TimeUnit.SECONDS,
+            new ArrayBlockingQueue<>(MAX_PRUNING_WORKER),
+            new ThreadPoolExecutor.DiscardPolicy());
+  }
+
+  @Override
+  public void onBlockAdded(final BlockAddedEvent event) {
```
siladu, line 59: If this errors or doesn't fire for some reason, e.g. node is shutdown between block import and this method, then what happens? Will two blocks be pruned next time it fires?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+          // Get pruning mark
+          final long blockNumber = event.getBlock().getHeader().getNumber();
+          long pruningMark = prunerStorage.getPruningMark().orElse(blockNumber);
+          if (blockNumber < pruningMark) {
+            // Ignore and warn if block number < pruning mark, this normally indicates the
+            // blocksToRetain
+            // is too small.
+            LOG.warn(
+                "Block added event: "
+                    + event
```
siladu, line 72: Is there anything useful in the event other than the number - maybe the hash?

siladu: "this normally indicates the blockToRetain is too small" is a useful message to put into the log itself if this is caused by a user misconfiguration

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+            return;
+          }
+          // Append block into fork blocks.
+          final KeyValueStorageTransaction tx = prunerStorage.startTransaction();
+          final Collection<Hash> forkBlocks = prunerStorage.getForkBlocks(blockNumber);
+          forkBlocks.add(event.getBlock().getHash());
+          prunerStorage.setForkBlocks(tx, blockNumber, forkBlocks);
+          // If a block is a new canonical head, start pruning.
+          if (event.isNewCanonicalHead()
+              && blockNumber - blocksToRetain - pruningMark >= pruningFrequency) {
```
siladu, line 86: Finding this hard to comprehend. Can the subtractions be extracted into a variable and given a name?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+            while (blockNumber - pruningMark >= blocksToRetain) {
```
siladu, line 87: Maybe same for this subtraction too.

IIUC, could it be `numberOfBlocksEligibleForPruning`?
If so, should this be > rather than >=?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+              LOG.debug("Pruning chain data at pruning mark: " + pruningMark);
```
siladu, line 88: Is it "at" pruning mark or "before" pruning mark? Maybe not so straight forward to name if there's forks involved :)

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/ChainDataPrunerTest.java
```diff
@@ -0,0 +1,117 @@
...
+    final ChainDataPruner chainDataPruner =
+        new ChainDataPruner(
+            blockchainStorage, new ChainDataPrunerStorage(new InMemoryKeyValueStorage()), 512, 0);
+    Block genesisBlock = gen.genesisBlock();
+    final MutableBlockchain blockchain =
+        DefaultBlockchain.createMutable(
+            genesisBlock, blockchainStorage, new NoOpMetricsSystem(), 0);
+    blockchain.observeBlockAdded(chainDataPruner);
+
+    // Generate & Import 1000 blocks
```
siladu, line 50: The comments in here are useful 👍

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+            new ArrayBlockingQueue<>(MAX_PRUNING_WORKER),
+            new ThreadPoolExecutor.DiscardPolicy());
+  }
+
+  @Override
+  public void onBlockAdded(final BlockAddedEvent event) {
+    LOG.debug("New block added event: " + event);
+    pruningExecutor.submit(
+        () -> {
+          // Get pruning mark
```
siladu, line 63: I believe the comments in this class are redundant, the code is clear enough to read and understand (except where I've specifically mentioned otherwise)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/ProtocolContext.java
```diff
@@ -67,4 +68,14 @@ public <C extends ConsensusContext> Optional<C> safeConsensusContext(final Class
         .filter(c -> klass.isAssignableFrom(c.getClass()))
         .map(klass::cast);
   }
+
+  // TODO: cleanup - isChainPruningEnabled will not be required after
+  // https://github.com/hyperledger/besu/pull/4703 is merged.
+  public void setIsChainPruningEnabled(final boolean isChainPruningEnabled) {
```
jframe, line 74: Don't think this is a good place for the chain pruning enabled flag, it is very high level. Maybe part of the Merge context would be ok, as we only need this check for the Merge RPCs.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+            1,
+            60L,
+            TimeUnit.SECONDS,
+            new ArrayBlockingQueue<>(MAX_PRUNING_WORKER),
+            new ThreadPoolExecutor.DiscardPolicy());
+  }
+
+  @Override
+  public void onBlockAdded(final BlockAddedEvent event) {
+    LOG.debug("New block added event: " + event);
```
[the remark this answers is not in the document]
wcgcyx: Yep, the logging here is unnecessary. Removed.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: Updated. It should be clearer now

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: yes, should be > rather than >=. Updated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: Updated the logging. Should be clear that pruning chain data at a specific height

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: Updated to remove unnecessary comments

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -41,50 +46,55 @@ public ChainDataPruner(
...
     this.blocksToRetain = blocksToRetain;
     this.pruningFrequency = pruningFrequency;
+    this.pruningExecutor =
+        new ThreadPoolExecutor(
+            1,
+            1,
+            60L,
+            TimeUnit.SECONDS,
+            new ArrayBlockingQueue<>(MAX_PRUNING_WORKER),
+            new ThreadPoolExecutor.DiscardPolicy());
```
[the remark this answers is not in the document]
wcgcyx: I've changed to using monitored executors with metrics. 
The rationale is to have only one thread to do the pruning at the same time. During initial sync, multiple threads could trigger a pruning operation, and having them running at the same time could corrupt the pruner state.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+      final BlockchainStorage blockchainStorage,
+      final ChainDataPrunerStorage prunerStorage,
+      final long blocksToRetain,
+      final long pruningFrequency) {
+    this.blockchainStorage = blockchainStorage;
+    this.prunerStorage = prunerStorage;
+    this.blocksToRetain = blocksToRetain;
+    this.pruningFrequency = pruningFrequency;
+    this.pruningExecutor =
+        new ThreadPoolExecutor(
```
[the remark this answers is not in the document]
wcgcyx: Updated to use monitored executors.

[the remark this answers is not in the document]
wcgcyx: Updated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,123 @@
...
+    LOG.debug("New block added event: " + event);
+    pruningExecutor.submit(
+        () -> {
+          // Get pruning mark
+          final long blockNumber = event.getBlock().getHeader().getNumber();
+          long pruningMark = prunerStorage.getPruningMark().orElse(blockNumber);
+          if (blockNumber < pruningMark) {
+            // Ignore and warn if block number < pruning mark, this normally indicates the
+            // blocksToRetain
+            // is too small.
```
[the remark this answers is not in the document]
wcgcyx: Updated, I've removed unnecessary comments. This comment goes into the logging to show the user that the blocks to retain are too small (probably misconfiguration).

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2175,7 +2192,12 @@ public BesuControllerBuilder getControllerBuilder() {
         .reorgLoggingThreshold(reorgLoggingThreshold)
         .evmConfiguration(unstableEvmOptions.toDomainObject())
         .dataStorageConfiguration(dataStorageOptions.toDomainObject())
-        .maxPeers(p2PDiscoveryOptionGroup.maxPeers);
+        .maxPeers(p2PDiscoveryOptionGroup.maxPeers)
+        .isChainDataPruningEnabled(unstableChainDataPruningOptions.getChainDataPruningEnabled())
+        .chainDataPruningBlocksRetained(
+            unstableChainDataPruningOptions.getChainDataPruningBlocksRetained())
+        .chainDataPruningFrequency(
+            unstableChainDataPruningOptions.getChainDataPruningBlocksFrequency());
```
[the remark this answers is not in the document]
wcgcyx: Updated

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
wcgcyx: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/ProtocolContext.java
wcgcyx: Updated. Ended up placing a flag in merge context since it involves very little change.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,110 @@
...
+          "Block added event: "
+              + event
+              + " has a block number of "
+              + blockNumber
+              + " < pruning mark "
+              + storedPruningMark
+              + " which normally indicates the blocksToRetain is too small");
+      return;
+    }
+    final KeyValueStorageTransaction tx1 = prunerStorage.startTransaction();
```
siladu, line 64: nit: do we need tx1 and tx2, any reason the same variable can't be reused?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,110 @@
...
+    final long storedPruningMark = prunerStorage.getPruningMark().orElse(blockNumber);
+    if (blockNumber < storedPruningMark) {
+      LOG.warn(
+          "Block added event: "
+              + event
+              + " has a block number of "
+              + blockNumber
+              + " < pruning mark "
+              + storedPruningMark
+              + " which normally indicates the blocksToRetain is too small");
```
siladu, line 61: This is warning for the user so better to use the cli option name IMO: `chain-data-pruning-blocks-retained` (excluding --X to future-proof for when it is no longer experimental)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
siladu: If we do need two, then I think a more descriptive name would help readability

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,110 @@
...
+      return;
+    }
+    final KeyValueStorageTransaction tx1 = prunerStorage.startTransaction();
+    final Collection<Hash> forkBlocks = prunerStorage.getForkBlocks(blockNumber);
+    forkBlocks.add(event.getBlock().getHash());
+    prunerStorage.setForkBlocks(tx1, blockNumber, forkBlocks);
+    tx1.commit();
+
+    pruningExecutor.submit(
```
siladu, line 70: Can another thread update the forkBlocks in the DB in between these transactions?
If so, then could that be a problem? 

...I'm thinking not because it's pulled out of storage again in the method below, but would like you to confirm it's been considered :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,110 @@
...
+          long currentPruningMark = storedPruningMark;
+          final long newPruningMark = blockNumber - blocksToRetain;
+          if (event.isNewCanonicalHead()
+              && newPruningMark - currentPruningMark >= pruningFrequency) {
+            long currentRetainedBlock = blockNumber - currentPruningMark + 1;
+            while (currentRetainedBlock > blocksToRetain) {
+              LOG.debug("Pruning chain data with block height of " + currentPruningMark);
+              pruneChainDataAtBlock(tx2, currentPruningMark);
+              currentPruningMark++;
+              currentRetainedBlock = blockNumber - currentPruningMark;
```
siladu, line 82: thanks for updating the variable, this is easier to understand now 👍

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -345,7 +346,13 @@ public BesuController build() {
                   storageProvider.getStorageBySegmentIdentifier(
                       KeyValueSegmentIdentifier.CHAIN_PRUNER_STATE)),
               chainDataPruningBlocksRetained,
-              chainDataPruningFrequency);
+              chainDataPruningFrequency,
+              MonitoredExecutors.newBoundedThreadPool(
+                  ChainDataPruner.class.getSimpleName(),
+                  1,
+                  1,
+                  ChainDataPruner.MAX_PRUNING_WORKER,
```
siladu, line 343: Isn't this queue size rather than number of worker threads?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: I was trying to use one variable actually but tx must be final in the lambda expression so couldn't be reused. Is there a way to get around that?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: Yes, it is possible but it won't be a problem. Additions to forkBlocks are happening in order and the pruning thread is to be created after any additions to forkBlocks. Therefore, when a pruning operation is happening in a thread for a given blockNumber, we can be sure that no additions would be performed on forkBlocks < blockNumber, only deletion will be performed. Deletions don't need to happen in order, so this should be fine.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: Updated.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
wcgcyx: Updated.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
siladu: > we can be sure that no additions would be performed on forkBlocks < blockNumber
might need to chat to get my head around this...I'm wondering how the case where we might import an 'old' block might be handled, e.g. we built on a fork then had a reorg and CL had to tell us about old blocks

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
siladu: Gotcha, you could probably work around it with a class instead of lambda but it's not worth it for this. Renaming would be an improvement IMO but I wouldn't block the PR on it :)

siladu (approved):
Approved! Just the tx1, tx2 renaming suggestion but not a blocker for me

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,111 @@
...
+    forkBlocks.add(event.getBlock().getHash());
+    prunerStorage.setForkBlocks(recordBlockHashesTransaction, blockNumber, forkBlocks);
+    recordBlockHashesTransaction.commit();
+
+    pruningExecutor.submit(
+        () -> {
+          final KeyValueStorageTransaction pruningTransaction = prunerStorage.startTransaction();
+          long currentPruningMark = storedPruningMark;
+          final long newPruningMark = blockNumber - blocksToRetain;
+          if (event.isNewCanonicalHead()
```
jframe, line 76: I think this check for pruning frequency could be done before submitting it to the executor so that less work is done in case nothing needs to happen.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,111 @@
...
+          }
+          prunerStorage.setPruningMark(pruningTransaction, currentPruningMark);
+          pruningTransaction.commit();
+        });
+  }
+
+  private void pruneChainDataAtBlock(final KeyValueStorageTransaction tx, final long blockNumber) {
+    final Collection<Hash> oldForkBlocks = prunerStorage.getForkBlocks(blockNumber);
+    final BlockchainStorage.Updater updater = blockchainStorage.updater();
+    for (Hash toPrune : oldForkBlocks) {
```
jframe, line 94: nit: toPrune can be final

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainPrunerConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public ChainPrunerConfiguration(
+      final boolean chainPruningEnabled,
+      final long chainPruningBlocksRetained,
+      final long chainPruningBlocksFrequency) {
+    this.chainPruningEnabled = chainPruningEnabled;
+    this.chainPruningBlocksRetained = chainPruningBlocksRetained;
+    this.chainPruningBlocksFrequency = chainPruningBlocksFrequency;
+  }
+
```
jframe, line 33: nit: chainPruning prefix could removed from the getters and fields are this is all on the ChainPrunerConfiguration class

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Increase the speed of modexp gas execution and execution. [#4780](https://github.com/hyperledger/besu/pull/4780)
 - Added experimental CLI options `--Xeth-capability-max` and `--Xeth-capability-min` to specify a range of capabilities to be supported by the Eth protocol. [#4752](https://github.com/hyperledger/besu/pull/4752)
 - Set the default curve in the EVMTool, like is done in production operations [#4790](https://github.com/hyperledger/besu/pull/4790)
+- Add chain data pruning feature with three experimental CLI options: `--Xchain-data-pruning-enabled`, `--Xchain-data-pruning-blocks-retained` and `--Xchain-data-pruning-frequency` [#4686](https://github.com/hyperledger/besu/pull/4686)
```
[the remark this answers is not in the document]
wcgcyx: updated

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/ChainDataPruningOptions.java
```diff
@@ -0,0 +1,90 @@
...
+import org.hyperledger.besu.cli.options.CLIOptions;
+import org.hyperledger.besu.ethereum.chain.ChainPrunerConfiguration;
+import org.hyperledger.besu.util.number.PositiveNumber;
+
+import java.util.Arrays;
+import java.util.List;
+
+import picocli.CommandLine;
+
+public class ChainDataPruningOptions implements CLIOptions<ChainPrunerConfiguration> {
```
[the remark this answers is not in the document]
wcgcyx: updated

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -315,6 +326,30 @@ public BesuController build() {
             blockchain, worldStateArchive, protocolSchedule, this::createConsensusContext);
     validateContext(protocolContext);
 
+    if (chainPrunerConfiguration.getChainPruningEnabled()) {
+      protocolContext.getConsensusContext(MergeContext.class).setIsChainPruningEnabled(true);
```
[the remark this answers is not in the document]
wcgcyx: updated

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -315,6 +326,30 @@ public BesuController build() {
...
+                  1,
+                  1,
+                  ChainDataPruner.MAX_PRUNING_THREAD_QUEUE_SIZE,
+                  metricsSystem));
+      blockchain.observeBlockAdded(chainDataPruner);
+      LOG.info(
+          "Chain data pruning enabled with recent blocks retained to be: "
+              + chainPrunerConfiguration.getChainPruningBlocksRetained()
+              + " and frequency to be: "
+              + chainPrunerConfiguration.getChainPruningBlocksFrequency());
```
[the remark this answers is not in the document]
wcgcyx: updated

ethereum/core/build.gradle
```diff
@@ -56,6 +56,7 @@ dependencies {
   implementation 'org.apache.tuweni:tuweni-rlp'
   implementation 'org.hyperledger.besu:bls12-381'
   implementation 'org.immutables:value-annotations'
+  implementation 'org.awaitility:awaitility'
```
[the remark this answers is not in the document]
wcgcyx: yep, that's true, updated.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -0,0 +1,111 @@
...
+    prunerStorage.setForkBlocks(recordBlockHashesTransaction, blockNumber, forkBlocks);
+    recordBlockHashesTransaction.commit();
+
+    pruningExecutor.submit(
+        () -> {
+          final KeyValueStorageTransaction pruningTransaction = prunerStorage.startTransaction();
+          long currentPruningMark = storedPruningMark;
+          final long newPruningMark = blockNumber - blocksToRetain;
+          if (event.isNewCanonicalHead()
+              && newPruningMark - currentPruningMark >= pruningFrequency) {
```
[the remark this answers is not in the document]
wcgcyx: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: updated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
wcgcyx: I need to think about this one.

jframe (approved):

PR #4697, opened 2022-11-17
Spike Spring Boot

gezero:
What if we would use a Spring Dependency Injection instead of our Builders?
What if we would only depend on mainnet parts of Besu project?

These and other questions might be answered in this PR. 

This is a first runabout and as such - It does not work yet. 
But it will...
Hopefully...

This PR creates a new subdirectory of besu where we depend on besu.
The Spring does not leak into besu code at the moment.


shemnon:
I think you should do a temperature check with the other maintainers before going too far down this rabbit hole.  Similar prior investigations into items such as Lombok resulted in late stage rejection.

I for one do not want to see any spring incursions into the EVM codebase as that reduces the utility as a library, and Hedera has a vested interest in such utility.

gezero:
@shemnon Thank you for your feedback. I agree 100% that spring or any other DI framework should not leak into the code that can live standalone. 

The main motivation of this PR is that the current state of the Besu orchestration feels tragic to me. I am convinced that you will appreciate how noninvasive this approach is. There is no need to leak any Spring into the current Besu codebase for the orchestration.

PR #4700, opened 2022-11-18
Use snapshots for in-memory copies of layered worldstate

garyschulte:
## PR description

In `BonsaiLayeredWorldState.copy()`, use a snapshot worldstate as the basis for a BonsaiInMemoryWorldstate copy

Prevents stateroot mismatches that arise from the persisted worldstate being different than the worldstate assoicated with a given layered worldstate.

Also moves isolation test setup into an abstract class, and adds a new BonsaiInMemoryIsolationTest

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
@@ -67,6 +67,14 @@ public class BonsaiSnapshotWorldStateKeyValueStorage extends BonsaiWorldStateKey
         trieLogStorage);
   }
 
+  @Override
+  public void close() throws Exception {
+    accountStorage.close();
+    codeStorage.close();
+    storageStorage.close();
+    trieBranchStorage.close();
+  }
+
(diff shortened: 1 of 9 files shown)


garyschulte:
This has been successfully burning in on a goerli node that was previously quite troubled.  In-Memory worldstate performance has been significantly faster than just using snapshots alone:

Screenshot from 2022-11-18 08-38-54

Heap consumption of the in-memory bonsai has been consistently higher, but the total of memory consumed of both heap and are quite close

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
```diff
@@ -67,6 +67,14 @@ public BonsaiUpdater updater() {
         trieLogStorage);
   }
 
+  @Override
+  public void close() throws Exception {
+    accountStorage.close();
```
matkt, line 72: we don't close trieLogStorage ?

garyschulte: trie log storage is not based on a snapshot, so we don't have to worry about closing it.

matkt (approved):
LGTM

PR #4703, opened 2022-11-18
Post Merge cleanup: remove latestValidAncestorDescendsFromTerminal

fab-10:

## Documentation

## Changelog


fab-10:
Indeed this was one of the thing that I was recalling, not sure if these testnets are a blocker for the cleanup, since their bootstrap is quite controlled and limited to a small number of cooperating nodes.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -519,46 +520,7 @@ private void forwardWorldStateTo(final BlockHeader newHead) {
...
-      final Optional<BlockHeader> maybeFinalized = mergeContext.getFinalized();
-      if (maybeFinalized.isPresent()) {
-        return isDescendantOf(maybeFinalized.get(), blockHeader);
-      } else {
-        Optional<BlockHeader> terminalBlockHeader = mergeContext.getTerminalPoWBlock();
-        if (terminalBlockHeader.isPresent()) {
-          return isDescendantOf(terminalBlockHeader.get(), blockHeader);
-        } else {
-          if (isTerminalProofOfWorkBlock(blockHeader, protocolContext)
-              || ancestorIsValidTerminalProofOfWork(blockHeader)) {
```
garyschulte, line 548: FYI ancestorIsValidTerminalProofOfWork is dead code if we drop latestValidAncestorDescendsFromTerminal

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayload.java
```diff
@@ -201,17 +201,6 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
-    // TODO: post-merge cleanup
-    if (!mergeCoordinator.latestValidAncestorDescendsFromTerminal(newBlockHeader)) {
-      mergeCoordinator.addBadBlock(block);
-      return respondWithInvalid(
-          reqId,
-          blockParam,
-          Hash.ZERO,
-          INVALID,
-          newBlockHeader.getHash() + " did not descend from terminal block");
-    }
```
garyschulte, line 213: just a comment - this reflects the perspective that we do whatever the CL tells us to do and do not second guess it.  I think doing this is consistent with that view

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java
```diff
@@ -115,18 +115,6 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
-    if (!mergeCoordinator.latestValidAncestorDescendsFromTerminal(newHead.get())) {
-      logForkchoiceUpdatedCall(INVALID, forkChoice);
-      return new JsonRpcSuccessResponse(
-          requestId,
-          new EngineUpdateForkchoiceResult(
-              INVALID,
-              Hash.ZERO,
-              null,
-              Optional.of(newHead.get() + " did not descend from terminal block")));
-    }
```
garyschulte, line 128: same comment as on newPayload regarding second guessing the CL.

garyschulte (commented):
In general I agree we should remove these checks, since they are only there to ensure a correct merge transition.  I'd like to see besu successfully work in a multi-client merge network with these removed.  

Last thought - we will certainly fail hive merge tests after this PR.  Not sure if we care though

siladu:
> AFAIK some devnets for Shanghai are still planned to start with PoW and transition to PoS. @gezero Should be able to confirm that. But if that is the case I think we still have to wait until we can remove it.

This is true, at least teku and prysm devnets require the merge to happen

siladu:
> _Update_ Teku has had to implement a workaround for this to be able to interop with geth, not sure if it's permanent though. Prysm still requires it.

Another update. The goal for the next withdrawals testnet is to have all clients starting post-merge, so that blocker will hopefully go away soon.
I hear Prysm should have support soon.
Besu + Teku still seems to require running through the merge, so that requires investigation.

ajsutton:
For the record in https://github.com/ConsenSys/teku/blob/a49f9a8a279a2260c57308fbb313162dfa7e2127/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/MergedGenesisAcceptanceTest.java#L28 we have an AT that starts a Teku+Besu node with an already merged genesis state.

jflo (approved):
buh bye TTD!

PR #4705, opened 2022-11-18
Remove log statements that are keeping references to objects for too long

fab-10:

## PR description

Improvement in terms of reducing java heap used,
since the logs were keeping reference to blocks sent by newPayload, that causes high memory consumption during initial sync, and could be one of the causes that prevent to complete snap sync on low spec machines. Exceptions are also logged by the backward sync, so there is no loss of information.

## Documentation

## Changelog


garyschulte (commented):
do your bwsync PRs add some logging in backward sync that will inform us of failures since we are dropping these?

fab-10:
Yes errors are already logged in backward sync

shemnon (approved):

garyschulte (approved):
:ship: I have been burning in the prior version of this PR overnight and noticed significantly less memory pressure on low spec machine

PR #4717, opened 2022-11-22
[2702] Add check for required option on cli

### LINKED ISSUE #2702: When using config toml and speicifying CommandLine.Option(required=true) besu fails to start
### Description

When you implement a command-line option with `CommandLine.Option(required=true)` in either Besu (ie in `BesuCommand`) or in a plugin Besu will fail to start if you do not specify it explicitly on the command line. 

### Steps to Reproduce (Bug)

Let say we had a config option that we required the user to explicitly set:

```java
  @Option(
      names = {"--accept-terms-and-conditions"},
      description = "You must explicitly accept terms and conditions"),
      require = true
  private final Boolean acceptTermsAndConditions;
```

You would need to run besu accordingly

`besu --accept-terms-and-conditions=true` 

this works as expected.

However, Besu will not start when using `besu --config-file=config.toml`

with the paramer in a config.toml
```toml
accept-terms-and-conditions=true
```

**Actual behavior:** 
Besu should fails to start with the error `Missing required option: ` even though one is specified in a config.toml

**Frequency:**
All the time

### Additional Information
Besu is currently using `info.picocli:picocli:4.6.1`. The way that toml and environment variables are parsed by picocli has been deprecated. 

https://github.com/hyperledger/besu/issues/2701 should be solved first.

wcgcyx:
Signed-off-by: wcgcyx <wcgcyx@gmail.com>

## PR description
Add a check for required options so that besu will fail to start if required options are not explicitly set on the command line. 

## Fixed Issue(s)

fixes #2702 

## Documentation

## Changelog


jframe (approved):
LGTM

PR #4718, opened 2022-11-22
Restore updating chain head and finalized block during backward sync

fab-10:
## PR description

Updating chain head and updating the finalized block during backward sync was wrongly removed during the removal of the finalized block rule, this PR restore it and also complete the removal of the remaining dead code that was meant to manage the finalized rule, plus some renaming to improve semantic and log improvements.
The update of the finalized block, is now done by the `MergeCoordinator` after the backward sync is done, because it has access to correctly set it into the `MergeContext`, while before it was only setting the finalized in the blockchain store. This means that the hashs of the finalized blocks are seen during the sync are kept in memory until the backwards is done, but given the small size of the hashes, this is not a concern in terms of memory usage, while setting the finalized block is a huge benefit in terms of perfomance since the check that a block is descendant of finalized is quicker when we have a fresh finalized block set instead of an old one.

## Documentation

## Changelog


macfarla (approved):
looks ok to me - I'm assuming you have tested it out on a mainnet node?

fab-10:
Tested on testnets, mainnet node testing is ongoing since it takes longer, and will only merge after I get all the results

fab-10:
> looks ok to me - I'm assuming you have tested it out on a mainnet node?

Tests on mainnet went fine, just found and commited an optimization for some edge cases

PR #4720, opened 2022-11-22
Do not send new payloads to backward sync if initial sync is in progress

fab-10:

## PR description

Do not send new payloads to backward sync if initial sync is in progress, since repivoting the initial sync will be very close to the chain head at the end, and so storing new payloads in the backward sync chain has no advantages, and only consumes resources, since most of them will be ignored when the backwars sync starts after the initial sync.

## Documentation

## Changelog


macfarla (approved):
LGTM

PR #4724, opened 2022-11-23
Calculate Transaction Size

### LINKED ISSUE #4723: Transaction size calculation (Required by eth/68)
### Description
The eth/68 protocol intends to pass the transaction size along with the transaction announcement.

Gabriel-Trintinalia:
Signed-off-by: Gabriel Trintinalia <gabriel.trintinalia@gmail.com>

## PR description
The eth/68 protocol intends to pass the transaction size along with the transaction announcement.

## Fixed Issue(s)
fixes #4723

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
@@ -926,6 +926,14 @@ public class Transaction
     return Optional.empty();
   }
 
+  private Bytes toRlp() {
+    return RLP.encode(this::writeTo);
+  }
+
+  public int calculateSize() {
+    return toRlp().size();
+  }
+
   public static class Builder {
 
     protected TransactionType transactionType;
--- a/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoderTest.java
+++ b/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoderTest.java
@@ -26,10 +26,13 @@ import org.hyperledger.besu.ethereum.rlp.RLPException;
 import org.hyperledger.besu.ethereum.rlp.RLPInput;
 
 import java.math.BigInteger;
-import java.util.List;
+import java.util.Arrays;
+import java.util.Collection;
 
 import org.apache.tuweni.bytes.Bytes;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
 class TransactionDecoderTest {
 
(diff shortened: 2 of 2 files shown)


ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoderTest.java
```diff
@@ -93,4 +94,23 @@ void shouldDecodeWithHighNonce() {
...
+
+  @Test
+  void shouldCalculateCorrectTransactionSize() {
+    for (final String rlp_tx :
+        List.of(
+            FRONTIER_TX_RLP,
+            EIP1559_TX_RLP,
+            GOQUORUM_PRIVATE_TX_RLP,
+            NONCE_64_BIT_MAX_MINUS_2_TX_RLP)) {
+      System.out.println(rlp_tx);
```
macfarla, line 106: use a logger or remove the System.out.println

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoderTest.java
```diff
@@ -93,4 +94,23 @@ void shouldDecodeWithHighNonce() {
     assertThat(transaction).isNotNull();
     assertThat(transaction.getNonce()).isEqualTo(MAX_NONCE - 1);
   }
+
+  @Test
+  void shouldCalculateCorrectTransactionSize() {
+    for (final String rlp_tx :
```
macfarla, line 100: you could use a parameterised test for this

macfarla: see WordsTest for an example - then if the first one fails, the rest will still run because they are effectively separate tests

macfarla (approved):
LGTM

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoderTest.java
Gabriel-Trintinalia: removed

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoderTest.java
Gabriel-Trintinalia: Done!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -926,6 +926,14 @@ public Optional<Address> contractAddress() {
...
   }
 
+  private Bytes toRlp() {
+    return RLP.encode(this::writeTo);
+  }
+
+  public int calculateSize() {
+    return toRlp().size();
+  }
+
```
fab-10, line 936: Do you think it makes sense to cache the size on first call, since encoding the transaction every time we want the size could be expensive?

Gabriel-Trintinalia: Hey Fabio, it totally makes sense. It will never change anyway, right?

fab-10: yes all the fields of the transactions are final, and while checking that, I found that this could be another optimization, the `hash` of the transactions is calculated and cached in the `getHash` method that already does the serialization, and so it could be used to sent the size, or the other way around if the `getSize` is called first.

fab-10: @Gabriel-Trintinalia I have implemented the memoization of the size here https://github.com/hyperledger/besu/pull/4812 if you want to review it

PR #4730, opened 2022-11-24
Implement eth/68 (EIP-5793)

### LINKED ISSUE #4715: Implement eth/68 (EIP-5793) - Add tx type and size to tx announcement
Implement eth/68

- https://eips.ethereum.org/EIPS/eip-5793

> The NewPooledTransactionHashes message announces transaction hashes, allowing the peer to selectively fetch transactions it does not yet have.
> 
> EIP-4844 introduces a new transaction type for blob transactions. Since these blob transactions are large, naively broadcasting them to sqrt(peers) could significantly increase bandwidth requirements. Adding the transaction type and the size to the announcement message will allow nodes to select which transactions they want to fetch and also allow them to load balance or throttle peers based on past behavior.
> 
> The added metadata fields will also enable future - upgradeless - protocol tweaks to prevent certain transaction type (e.g. blob transactions) or certain transaction sizes (e.g. 128KB+) from being blindly broadcast to many peers. Enforcing announcements only and retrieval on demand would ensure a much more predictable networking behavior, limiting the amplification effect of transaction propagation DoS attack.

Gabriel-Trintinalia:
## PR description
Implement eth/68

- https://eips.ethereum.org/EIPS/eip-5793

> The NewPooledTransactionHashes message announces transaction hashes, allowing the peer to selectively fetch transactions it does not yet have.
> 
> EIP-4844 introduces a new transaction type for blob transactions. Since these blob transactions are large, naively broadcasting them to sqrt(peers) could significantly increase bandwidth requirements. Adding the transaction type and the size to the announcement message will allow nodes to select which transactions they want to fetch and also allow them to load balance or throttle peers based on past behavior.
> 
> The added metadata fields will also enable future - upgradeless - protocol tweaks to prevent certain transaction type (e.g. blob transactions) or certain transaction sizes (e.g. 128KB+) from being blindly broadcast to many peers. Enforcing announcements only and retrieval on demand would ensure a much more predictable networking behavior, limiting the amplification effect of transaction propagation DoS attack.

Modify the NewPooledTransactionHashes (0x08) message:

(eth/67): [hash_0: B_32, hash_1: B_32, ...]
(eth/68): [[type_0: B_1, type_1: B_1, ...], [size_0: B_4, size_1: B_4, ...], [hash_0: B_32, hash_1: B_32, ...]]

## Fixed Issue(s)
fixes #4715
## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionAnnouncement.java
```diff
@@ -0,0 +1,86 @@
...
+    return type;
+  }
+
+  public Optional<Integer> getSize() {
+    return size;
+  }
+
+  public static List<TransactionAnnouncement> create(
+      final List<TransactionType> types, final List<Integer> sizes, final List<Hash> hashes) {
+    final List<TransactionAnnouncement> transactions = new ArrayList<>();
```
fab-10, line 61: little optimization, you can set the size of the array

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementDecoder.java
```diff
@@ -0,0 +1,82 @@
...
+    final List<TransactionType> types =
+        input.readList(
+            rlp -> {
+              final int type = rlp.readByte() & 0xff;
+              return type == 0 ? TransactionType.FRONTIER : TransactionType.of(type);
+            });
+    final List<Integer> sizes = input.readList(RLPInput::readInt);
+    final List<Hash> hashes = input.readList(rlp -> Hash.wrap(rlp.readBytes32()));
+    input.leaveList();
+    return TransactionAnnouncement.create(types, sizes, hashes);
```
fab-10, line 80: somewhere we should put a check that the lists have the same size or raise a parse error

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementDecoder.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  private static List<TransactionAnnouncement> decodeForEth68(final RLPInput input) {
+    input.enterList();
+    final List<TransactionType> types =
+        input.readList(
+            rlp -> {
+              final int type = rlp.readByte() & 0xff;
+              return type == 0 ? TransactionType.FRONTIER : TransactionType.of(type);
+            });
+    final List<Integer> sizes = input.readList(RLPInput::readInt);
```
fab-10, line 77: I could not find in the spec where the type of the size is defined, I also think int is reasonable, but what if a peer sends a long here?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionAnnouncement.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public TransactionAnnouncement(final Transaction transaction) {
+    this(transaction.getHash(), transaction.getType(), transaction.calculateSize());
+  }
+
+  public TransactionAnnouncement(final Hash hash, final TransactionType type, final Integer size) {
+    this.hash = hash;
+    this.type = Optional.ofNullable(type);
+    this.size = Optional.ofNullable(size);
```
fab-10, line 44: Why allowing nulls here? 
`type` and `size` should never be null at this point, otherwise it means there message was malformed, and its processing should have stopped before.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementEncoder.java
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * <p>format: [[type_0: B_1, type_1: B_1, ...], [size_0: B_4, size_1: B_4, ...], ...]
+   *
+   * @param transactions the list to encode
+   * @return the encoded value. The message data will contain hashes, types and sizes.
+   */
+  private static Bytes encodeForEth68(final List<Transaction> transactions) {
+    final List<Integer> sizes = new ArrayList<>();
+    final List<TransactionType> types = new ArrayList<>();
+    final List<Hash> hashes = new ArrayList<>();
```
fab-10, line 80: little optimization: set the size of the arrays

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementEncoder.java
```diff
@@ -0,0 +1,103 @@
...
+          types.add(transaction.getType());
+          sizes.add(transaction.calculateSize());
+          hashes.add(transaction.getHash());
+        });
+
+    return encodeForEth68(types, sizes, hashes);
+  }
+
+  @VisibleForTesting
+  public static Bytes encodeForEth68(
```
fab-10, line 92: package private should be enough

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManagerTest.java
```diff
@@ -1124,9 +1124,9 @@ public void forkIdForChainHeadMayBeNull() {
...
   @Test
   public void shouldUseRightCapabilityDependingOnSyncMode() {
-    assertHighestCapability(SyncMode.X_SNAP, EthProtocol.ETH67);
-    assertHighestCapability(SyncMode.FULL, EthProtocol.ETH67);
-    assertHighestCapability(SyncMode.X_CHECKPOINT, EthProtocol.ETH67);
+    assertHighestCapability(SyncMode.X_SNAP, EthProtocol.ETH68);
+    assertHighestCapability(SyncMode.FULL, EthProtocol.ETH68);
+    assertHighestCapability(SyncMode.X_CHECKPOINT, EthProtocol.ETH68);
     /* Eth67 does not support fast sync, see EIP-4938 */
     assertHighestCapability(SyncMode.FAST, EthProtocol.ETH66);
```
fab-10, line 1131: instead of replacing eth/67 with eth/68 could you test both?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/NewPooledTransactionHashesMessageProcessorTest.java
```diff
@@ -170,17 +189,181 @@ public void shouldNotScheduleGetPooledTransactionsTaskTwice() {
...
+
+    assertThat(announcementList.size()).isEqualTo(list.size());
+
+    for (final Transaction transaction : list) {
+      final TransactionAnnouncement announcement = announcementList.get(list.indexOf(transaction));
+      assertThat(announcement.getHash()).isEqualTo(transaction.getHash());
+      assertThat(announcement.getType()).hasValue(transaction.getType());
+      assertThat(announcement.getSize()).hasValue(transaction.calculateSize());
+    }
+  }
```
fab-10, line 368: I suggest to add tests that stress the parsing with malformed messages, like arrays not of the same type, missing elements and wrong types, to make sure them are intecepted and handled

fab-10 (changes requested):
overall the implementation is good, please read the inline comments for questions, optimizations and suggestions to improve the handling of malformed messages.

Then there is another part of this implementation that should cover our new strategy on how we broadcast transactions taking in account the new `type` and `size` parameters, it is fine to do it in another PR to keep them small, have you already planned to do it?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionAnnouncement.java
Gabriel-Trintinalia: done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementEncoder.java
Gabriel-Trintinalia: done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/NewPooledTransactionHashesMessage.java
```diff
@@ -14,37 +14,48 @@
...
-
+public class NewPooledTransactionHashesMessage extends AbstractMessageData {
   private static final int MESSAGE_CODE = EthPV65.NEW_POOLED_TRANSACTION_HASHES;
-  private List<Hash> pendingTransactions;
+  private List<TransactionAnnouncement> pendingTransactions;
+  private final Capability capability;
 
-  NewPooledTransactionHashesMessage(final Bytes rlp) {
+  @VisibleForTesting
+  public NewPooledTransactionHashesMessage(final Bytes rlp, final Capability capability) {
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: There are tests using this method in another package. That is why I made it public and annotated it with `@VisibleForTesting`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementEncoder.java
Gabriel-Trintinalia: There are tests using this method in another package. That is why I made it public and annotated it with `@VisibleForTesting`

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManagerTest.java
Gabriel-Trintinalia: So that test asserts the highest capability available for a sync mode. If 68 is enabled, then it will always be 68. Does it make sense? Each Sync mode will have only one possible highest capability

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/NewPooledTransactionHashesMessageProcessorTest.java
Gabriel-Trintinalia: will do

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementDecoder.java
Gabriel-Trintinalia: Yeah agreed, Geth uses uint64, I will do the change

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionAnnouncement.java
Gabriel-Trintinalia: Hum, if the transaction is announced by a protocol version older than Eth/68, only the hash is announced

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementDecoder.java
Gabriel-Trintinalia: done

Gabriel-Trintinalia:
@fab-10 Thanks for the review.

> Then there is another part of this implementation that should cover our new strategy on how we broadcast transactions taking in account the new `type` and `size` parameters, it is fine to do it in another PR to keep them small, have you already planned to do it?

Yeah, that makes sense. This PR only implements the protocol specs, implementing the ability to send and receive NewPooledTransactionHashes with their sizes and types when using eth68 ensuring that we are compatible with older protocol versions. 

What the client should do with the type and size and how it impacts broadcast transaction strategy are not part of the scope. I believe it would be better to handle it in a separate PR.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementDecoder.java
Gabriel-Trintinalia: done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionAnnouncement.java
fab-10: if the transaction is announced by a protocol < Eth/68 then the constructor `public TransactionAnnouncement(final Hash hash)` is used if I am not wrong

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManagerTest.java
fab-10: ok, I missed the _highest capability_ part, probably

fab-10 (approved):
LGTM, just a comment on the nullable arguments

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionAnnouncement.java
Gabriel-Trintinalia: I understood your point now. I changed the constructors to not accept `null` and throw instead. Also added a unit test for that. Thanks again!

fab-10 (approved):

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ https://hyperledger.jfrog.io/hyperledger/besu-binaries/besu/22.10.2/besu-22.10.2
 - Added new RPC endpoints `debug_setHead` & `debug_replayBlock  [4580](https://github.com/hyperledger/besu/pull/4580)
 - Upgrade OpenTelemetry to version 1.19.0 [#3675](https://github.com/hyperledger/besu/pull/3675)
 - Implement Eth/67 sub-protocol [#4596](https://github.com/hyperledger/besu/issues/4596)
+- Implement Eth/68 sub-protocol [#4715](https://github.com/hyperledger/besu/issues/4715)
```
macfarla, line 32: you'll need to move this to 22.10.3

macfarla (approved):
LGTM just need to move the changelog entry

PR #4737, opened 2022-11-25
 Preload state trie node

matkt:
Co-authored-by: Ameziane H <ameziane.hamlat@consensys.net>

## PR description
The idea behind this PR is to preload asynchronously account nodes and storage nodes from the database during the transaction processing to use these nodes during the calculate root hash step.
We've created two caches, one for account nodes and one for storage nodes. The size of these caches is 100k for accounts and 200k for storage. We've tested other values but this configuration is the one that works better.
We also exporter cache metrics as Prometheus metrics to check cache efficiency.

We didn't see any impact on GC activity even on 4 GiB Heaps (-Xmx4g).

### The results
We did our tests on different AWS EC instances, here're the results. 

**M6a.xlarge** (4 vCPU, 16 GiB)
Block processing time is 34% better for median values and 41% better for 95th percentile.

image

**M5.xlarge** (4 vCPU, 16 GiB)
Block processing time is 28% better for median values and 95th percentile.

image

**I3.2xlarge** (8 vCPU, 61 GiB)
Block processing time is 21% better for median values and 95th percentile.

image

**Cache Efficiency**
We can see in the screenshots below that these two caches are very efficient (>99%) and increasing storage cache size more than 200k is not necessary.

Account cache size = 100k and Storage cache size = 200k
image

Account cache size = 100k and Storage cache size = 1 million
image

  

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -283,6 +283,6 @@ public void persist(final BlockHeader blockHeader) {
 
   @Override
   public WorldUpdater updater() {
-    return new BonsaiWorldStateUpdater(this);
+    return new BonsaiWorldStateUpdater(this, (__, ___) -> {}, (__, ___) -> {});
```
gezero, line 286: 🤔 Why do we have to pass these as arguments anywhere? If I look at this code in the future, how do I know we did not forget to provide here something? I suggest overloading a constructor and set the NO OP values as defaults on your fields

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -90,20 +101,23 @@ public BonsaiWorldStateArchive(
...
       final Blockchain blockchain,
-      final boolean useSnapshots) {
+      final boolean useSnapshots,
+      final ObservableMetricsSystem metricsSystem) {
     this.trieLogManager = trieLogManager;
     this.blockchain = blockchain;
     this.worldStateStorage = worldStateStorage;
     this.persistedState = new BonsaiPersistedWorldState(this, worldStateStorage);
     this.useSnapshots = useSnapshots;
+    this.optimizedMerkleTrieLoader = new OptimizedMerkleTrieLoader(metricsSystem);
```
gezero, line 120: I would suggest avoiding using new in constructor. This creates a tight, hidden coupling between classes. Inject your collaborators, throw constructor.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -687,4 +700,48 @@ public boolean isDirty() {
         && storageToClear.isEmpty()
         && codeToUpdate.isEmpty());
   }
+
+  public static class AccountConsumingMap<T> extends ConcurrentHashMap<Address, T> {
```
gezero, line 704: I would strongly advise not to extend standard library classes unless you are very sure about what you are doing. This can create unforeseen consequences where you might break invariants of the parent class when you are not aware you are doing it. This goes double when dealing with concurrent collections. You can achieve similar behaviour using delegation instead of inheritance.

Even if you do not break any invariants now, it's an easy thing for the next developer making some simple change that break an invariant.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
```diff
@@ -0,0 +1,136 @@
...
+
+public class OptimizedMerkleTrieLoader {
+
+  private static final int ACCOUNT_CACHE_SIZE = 100_000;
+  private static final int STORAGE_CACHE_SIZE = 200_000;
+  private final Cache<Bytes, Bytes> accountsNodes;
+  private final Cache<Bytes, Bytes> storageNodes;
+
+  public OptimizedMerkleTrieLoader(final ObservableMetricsSystem metricsSystem) {
+    accountsNodes = CacheBuilder.newBuilder().recordStats().maximumSize(ACCOUNT_CACHE_SIZE).build();
```
gezero, line 44: Please avoid running complicated logic in a constructor. Maybe extract it into a factory method.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
```diff
@@ -0,0 +1,136 @@
...
+    if (metricsSystem instanceof PrometheusMetricsSystem)
+      ((PrometheusMetricsSystem) metricsSystem)
+          .addCollector(BesuMetricCategory.BLOCKCHAIN, () -> cacheMetrics);
+  }
+
+  public void preLoadAccount(
+      final BonsaiWorldStateKeyValueStorage worldStateStorage,
+      final Hash worldStateRootHash,
+      final Address account) {
+    CompletableFuture.runAsync(
```
gezero, line 59: It feels to me weird that nobody handles the future when it is done. Maybe you can have the Optional<Node> that you were preloading to be a return value of the future and when the future completes, update your accountsNodes. That would also prevent the need of modifying a field in the middle of a function. Functions in functional programming should preferably not have side effects, so this might help.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
```diff
@@ -0,0 +1,136 @@
...
+import com.google.common.cache.CacheBuilder;
+import io.prometheus.client.guava.cache.CacheMetricsCollector;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+
+public class OptimizedMerkleTrieLoader {
+
+  private static final int ACCOUNT_CACHE_SIZE = 100_000;
+  private static final int STORAGE_CACHE_SIZE = 200_000;
+  private final Cache<Bytes, Bytes> accountsNodes;
```
gezero, line 40: Two plurals, maybe this should be accountNodes?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
```diff
@@ -0,0 +1,136 @@
...
+                    return node;
+                  },
+                  worldStateRootHash,
+                  Function.identity(),
+                  Function.identity());
+          accountTrie.get(Hash.hash(account));
+        });
+  }
+
+  public void preLoadStorage(
```
gezero, line 76: Rename to preLoadStorageSlot maybe?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
```diff
@@ -0,0 +1,136 @@
...
+  public Optional<Bytes> getAccountStorageTrieNode(
+      final BonsaiWorldStateKeyValueStorage worldStateKeyValueStorage,
+      final Hash accountHash,
+      final Bytes location,
+      final Bytes32 nodeHash) {
+    if (nodeHash.equals(MerklePatriciaTrie.EMPTY_TRIE_NODE_HASH)) {
+      return Optional.of(MerklePatriciaTrie.EMPTY_TRIE_NODE);
+    } else {
+      Optional<Bytes> node = Optional.ofNullable(storageNodes.getIfPresent(nodeHash));
+      if (node.isPresent()) {
```
gezero, line 129: These code feels like it could be more functional.

matkt: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
matkt: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
matkt: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
matkt: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
matkt: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
```diff
@@ -61,12 +61,13 @@ protected Hash calculateRootHash(final BonsaiWorldStateUpdater worldStateUpdater
...
 
-    // for manicured tries and composting, trim and compost here
-
     // next walk the account trie
     final StoredMerklePatriciaTrie<Bytes, Bytes> accountTrie =
         new StoredMerklePatriciaTrie<>(
-            this::getAccountStateTrieNode,
+            (location, hash) ->
+                archive
+                    .getOptimizedMerkleTrieLoader()
```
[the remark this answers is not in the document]
matkt: changed with cached

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/OptimizedMerkleTrieLoader.java
ahamlat: Good point, the idea was to execute asynchronously as much work as possible to fill the cache. These completable futures doesn't return any result, but I totally agree that it should be more clean to control the execution of the completable future. 
This is a first version, and this can be improved in a future PR. The control of completable futures' executions will introduce a lot of changes that can be done in the future PR. We can for example stop all the futures when we start reading from the cache, this is not done in this current PR.

ahamlat (approved):
LGTM

gezero (commented):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiPersistedWorldState.java
```diff
@@ -116,7 +126,10 @@ protected Hash calculateRootHash(
     // next walk the account trie
     final StoredMerklePatriciaTrie<Bytes, Bytes> accountTrie =
         new StoredMerklePatriciaTrie<>(
-            this::getAccountStateTrieNode,
+            (location, hash) ->
```
garyschulte, line 129: it appears that `Optional<Bytes> getAccountStateTrieNode()` is no longer used anywhere

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -300,8 +301,10 @@ public BesuController build() {
             reorgLoggingThreshold,
             dataDirectory.toString());
 
+    final CachedMerkleTrieLoader cachedMerkleTrieLoader = new CachedMerkleTrieLoader(metricsSystem);
```
garyschulte, line 304: this feels like a worldstateArchive concern that is leaking.  Non-blocking, but I think we could benefit from a builder and/or pass metricsSystem instead

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -39,30 +38,44 @@
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.google.common.collect.ForwardingMap;
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
 import org.apache.tuweni.units.bigints.UInt256;
+import org.jetbrains.annotations.NotNull;
```
garyschulte, line 45: I think only intellij is going to use this annotation.  Findbugs uses NonNull and would probably be a better choice.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/CachedMerkleTrieLoader.java
```diff
@@ -0,0 +1,151 @@
...
+  public void preLoadAccount(
+      final BonsaiWorldStateKeyValueStorage worldStateStorage,
+      final Hash worldStateRootHash,
+      final Address account) {
+    CompletableFuture.runAsync(
+        () -> cacheAccountNodes(worldStateStorage, worldStateRootHash, account));
+  }
+
+  @VisibleForTesting
+  public void cacheAccountNodes(
```
garyschulte, line 66: can be default visibility rather than public

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/CachedMerkleTrieLoader.java
```diff
@@ -0,0 +1,151 @@
...
+              worldStateRootHash,
+              Function.identity(),
+              Function.identity());
+      accountTrie.get(Hash.hash(account));
+    } catch (MerkleTrieException e) {
+      // ignore exception for the cache
+    }
+  }
+
+  public void preLoadStorageSlot(
```
garyschulte, line 87: can be default viz

garyschulte (approved):
Non-blocking feedback, can be addressed in a subsequent PR.

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/TrieGenerator.java
```diff
@@ -16,46 +16,60 @@
...
+import java.util.stream.IntStream;
 
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
 import org.apache.tuweni.units.bigints.UInt256;
 
 public class TrieGenerator {
 
-  public static MerklePatriciaTrie<Bytes32, Bytes> generateTrie(
+  public static MerklePatriciaTrie<Bytes, Bytes> generateTrie(
```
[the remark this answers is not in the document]
matkt: because the keys are not necessarily only hashes. storage flat db has longer keys

matkt:
I merge and I will fix in a next PR. Thank you for your reviews

PR #4752, opened 2022-11-30
Hidden option to enforce maximum or minimum eth capability 

Gabriel-Trintinalia:
## PR description
Create hidden flags to enforce a range of Eth capabilities. With future versions of the protocol being implemented, we may need an easy way to compare capabilities without having to change the code every time.  Another use case would be to downgrade your node if an issue was found with a later protocol version.
 
This PR implements:

- Flag `--Xeth-capability-max` that enforces a maximum capability for the Eth protocol. Use: `--Xeth-capability-max=67`
- Flag `--Xeth-capability-min` that enforces a minimum capability for the Eth protocol. Use: `--Xeth-capability-min=65`

## Documentation

## Changelog


macfarla (commented):
I think it makes sense - another use case would be to downgrade your node if an issue was found with a later protocol version, right?

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/EthProtocolOptions.java
```diff
@@ -34,6 +34,8 @@ public class EthProtocolOptions implements CLIOptions<EthProtocolConfiguration>
   private static final String LEGACY_ETH_64_FORK_ID_ENABLED =
       "--compatibility-eth64-forkid-enabled";
 
+  private static final String MAX_CAPABILITY_ENABLED = "--Xeth-max-capability-enabled";
```
macfarla, line 37: re the name - generally the "enabled" suffix indicates a boolean 
what about
--X-eth-capability-max

Gabriel-Trintinalia: Done

Gabriel-Trintinalia:
> I think it makes sense - another use case would be to downgrade your node if an issue was found with a later protocol version, right?

That is another good example

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/network/NetworkRunner.java
```diff
@@ -78,6 +78,12 @@ public void start() {
       LOG.info("Starting Network.");
       setupHandlers();
       network.start();
+
+      LOG.info(
+          "Max supported protocol: {}",
```
macfarla, line 83: this is logging all supported, not just max?

macfarla (approved):
LGTM

PR #2433, opened 2021-06-15
Include custom errorprone checks as external dependency

usmansaleem:
## PR description
Include custom errorprone checks as external dependency. Custom errorprone checks has been refactored to its own repo https://github.com/Consensys/errorprone-checks. 

This change will allow to use custom errorprone checks in other projects as well.

Signed-off-by: Usman Saleem <usman@usmans.info>

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #4754, opened 2022-11-30
Replace errorprone-checks with consensys/errorprone-checks

jtraglia:
## PR description

This PR replaces the internal errorprone-checks with the external [consensys/errorprone-checks](https://github.com/ConsenSys/errorprone-checks) that Teku & Discovery use. It contains all of the same checks plus a few new ones. I think it's a good idea to deduplicate this stuff. In subsequent PRs, I'll review & enable some of those new checks.

## Documentation

## Changelog


shemnon (changes requested):
This is at it's core a repost of PR#2433, so I will re-state [my concerns](https://github.com/hyperledger/besu/pull/2433#pullrequestreview-683549503).  

Taking code that was once hyperledger, and migrating it to another companies repository, then making the original project depdenant on that library raises concerns, of which no answers were given in Jun of 2021.

There was a proposal to create a new besu-errorprone repository, but the original reporter closed the issue instead.

shemnon:
What was curious was when I raised my concern they closed the issue without comment.

jtraglia:
Yeah that is strange. I'm not part of either organization, but I'm not opposed to this being transferred to Hyperledger. Let's see if we can make this happen. I will open an issue in the ConsenSys repo about this.

ajsutton:
The original issue wasn't closed without comment - there was originally agreement to move to a hyperledger repo but it wasn't a priority so eventually it was closed with "Closing as not active and very low in priority."

From the Teku side, it doesn't really work to have these checks in a hyperledger repo and controlled by the besu maintainers as the majority of Teku devs aren't Besu maintainers and we'd be unable to effectively push the changes forward and do releases etc. So we'd wind up maintaining the checks we need in a ConsenSys repo anyway, at which point there's no real point in splitting the besu checks out of besu's main repo.  The ConsenSys repo is open source and Apache licensed so Besu is welcome to either use it or just pull in any changes from it as needed.

jtraglia:
@shemnon So it's unlikely that repo will be transferred and I'm not sure what you think is the correct thing to do here. I'll fix the failing checks if there's any chance of this PR getting merged. Otherwise, can I close this?

shemnon:
Since there is no "parent" repository to put it in I don't see this PR reaching a mutually satisfactory resolution.

ajsutton:
@shemnon Do you think it's worth just pulling in a copy of the additional rules/changes from the ConsenSys version so they can be applied to Besu?  I'm also not against putting a copy into a separate hyperledger/errorprone-checks repo if that makes it easier to pull changes between the two versions, but I suspect it just adds overhead.

shemnon:
I'm willing to do that, but I'm still 2 (minor) security patches and Big EOF behind, so it likely won't be until the new year

ajsutton:
I wasn't really expecting you to actually do it, just wanted to check the idea with you. :)

PR #4757, opened 2022-11-30
Use value comparisons instead of reference comparisons in some places

jtraglia:
## PR description

Change three reference comparisons that I believe should be value comparisons.

* Comparisons with `Bytes.EMPTY` will not work as expected.
  * For example, `Bytes.of() == Bytes.EMPTY` is false.
* While the two implementations are enums, `MetricCategory` is actually an interface.
  * In practice this is fine, but it's safer for this to be a value comparison.

## Documentation

## Changelog


Gabriel-Trintinalia (approved):
lgtm

macfarla (approved):

Issue #4765, opened 2022-12-02
Besu crash on mainnet


PacoBits:
This started with upgrade to version 22.10.1
It  happends each time we restart. Some times take more time, but always crash
We have tried with param       --Xbonsai-use-snapshots=false but also crash

I attach the trace.
```
Description

 A fatal error has been detected by the Java Runtime Environment:

  SIGSEGV (0xb) at pc=0x00007fb738ace941, pid=1, tid=238

 JRE version: OpenJDK Runtime Environment (11.0.17+8) (build 11.0.17+8-post-Ubuntu-1ubuntu220.04)
 Java VM: OpenJDK 64-Bit Server VM (11.0.17+8-post-Ubuntu-1ubuntu220.04, mixed mode, sharing, tiered, compressed oops, g1 gc, linux-amd64)
 Problematic frame:
 C  [libc.so.6+0x22941]  abort+0x213

Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /opt/besu/core.1)

 If you would like to submit a bug report, please visit:
   https://bugs.launchpad.net/ubuntu/+source/openjdk-lts
 The crash happened outside the Java Virtual Machine in native code.
 See problematic frame for where to report the bug.
```

[besu_02122022.log](https://github.com/hyperledger/besu/files/10140765/besu_02122022.log)

garyschulte:
@PacoBits would you upgrade to the latest 22.10.2 release and confirm/disconfirm that it has fixed this issue for you

ziobox:
> @PacoBits would you upgrade to the latest 22.10.2 release and confirm/disconfirm that it has fixed this issue for you

with latest 22.10.2 inside a containerized rocketpool, I have the same issue

```
rocketpool_eth1  | ####################################################################################################
rocketpool_eth1  | #                                                                                                  #
rocketpool_eth1  | # Besu 22.10.2                                                                                     #
rocketpool_eth1  | #                                                                                                  #
rocketpool_eth1  | # Configuration:                                                                                   #
rocketpool_eth1  | # Network: Mainnet                                                                                 #
rocketpool_eth1  | # Data storage: Bonsai                                                                             #
rocketpool_eth1  | # Sync mode: Checkpoint                                                                            #
rocketpool_eth1  | # RPC HTTP APIs: ETH,NET,WEB3                                                                      #
[...]

PR #4767, opened 2022-12-02
Ensure that Snashot worldstates remain open for async operations

ahamlat:

## PR description
in 22.10.1, two new operations pre-load account and storage from their respective worldstate asynchronously.  If the worldstate is based on a snapshot, there is a race where the snapshot might be closed ahead of the async operations completing, and subsequently result in a segfault when trying to access a closed rocksdb snapshot.  

This PR adds a subscriber implementation for BonsaiWorldStateKeyValueStorage which ensures that the storage will not be closed while subscribers are still operating on it.

## Fixed Issue(s)

#4765 
## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
@@ -15,8 +15,8 @@
  */
 package org.hyperledger.besu.ethereum.bonsai;
 
-import org.apache.tuweni.bytes.Bytes;
-import org.apache.tuweni.bytes.Bytes32;
+import static org.hyperledger.besu.util.Slf4jLambdaHelper.warnLambda;
+
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.ethereum.storage.StorageProvider;
 import org.hyperledger.besu.ethereum.storage.keyvalue.KeyValueSegmentIdentifier;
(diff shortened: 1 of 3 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
```diff
@@ -67,12 +70,19 @@ public BonsaiUpdater updater() {
...
-    accountStorage.close();
-    codeStorage.close();
-    storageStorage.close();
-    trieBranchStorage.close();
+    if (!isClosed.getAndSet(true)) {
+      accountStorage.close();
+      codeStorage.close();
+      storageStorage.close();
+      trieBranchStorage.close();
+    }
```
garyschulte, line 85: Since we are loading the cache async, I think we can still end up trying to read from a closed db.  I think a better pattern would be to prevent _actual_ closing the storage while there is still a consumer running an operation.  

Something along the lines of passing a consumer of the worldstate into a `boolean safeExecute(Consumer<KeyValueStorage> x)` function that will ensure the worldstate does not close while the lambda is running.  if the storage is closed, the function should return false to indicate the storage is closed.

We shouldn't block calls to `close()` on that function, just have close() mark the actual closing of the storages.  both `close` and `safeExecute` can call `private void doClose()`at the end of their function to close if both the keyvaluestorage has been marked "to_close" , and no outstanding safeExecute operations are running.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
```diff
@@ -68,11 +81,33 @@ public BonsaiUpdater updater() {
...
+  }
+
+  @Override
+  public synchronized void close() throws Exception {
+    isClosed.getAndSet(true);
+    tryClose();
+  }
+
+  protected void tryClose() throws Exception {
+    if (!isClosed.get() && subscribers.getSubscriberCount() < 1) {
```
matkt, line 105:
```suggestion
    if (isClosed.get() && subscribers.getSubscriberCount() < 1) {
```

garyschulte (approved):

Issue #4631, opened 2022-11-07
Implement EIP-4844 


non-fungible-nelson:
- [x] Transaction Broadcast #4821 
- [x] Excess Data Gas Block headers #4824 
- [x] DataGas accounting #4825 
- [x] Revert SSZ to RLP #5484
- [x] Merge eip-4844-interop into Main
- [x] add proofs to blobs bundle #5398
- [x] align blob return values #5399
- [x] Engine Apis V3 #4966 
- [x] KZG precompile #4822 
- [x] DATAHASH opcode #4823 
- [x] Shard blob tx encoding #4783 
- [x] Implement eth/68 #4730 
- [x] Ban Zero Blob Transactions #5401
- [x] Decouple Blobs #5402
- [x] KZG point eval to big endian #5485
- [x] ensure payload id is unique #5490
- [x] Limit blob related fields #5488 - pending approval
- [x] Validate versioned hashes #5487
- [x] Include datagas details in receipts vie RPC #5489 - pending approval
- [x] add dataGasUsed to EnginePayloadParam and blockHeader #5545
- [x] Increase Blob Throughput (3/6) #5590
- [x] Add VersionedHashes to TransactionResult #5604

gezero:
I think this is the list so far of things that should be done:
For accepting blocks:

- Parsing transactions requires new SSZ encoding
- - We need to refactor the current Transaction encoder that is tightly coupled with the current way of encoding and make it more generic
- - We need to find out if we can reuse what Teku has in Teku codebase
- Integrating with c-kzg
- - stefan.bratanov has a [repo](https://github.com/ConsenSys/jc-kzg-4844)
- - Integration with the c kzg library to check commitments  for a block to validate it. One aggregate commitment over all transactions in block must be calculated for this to be done.
- EVM has access to the commitments (not blobs) so that has to be handled
For creating new blocks:
- We must work on new transaction gossip. The eth/68 I think specifies that the new transaction will be only advertised but not pushed to peers. We need logic to retrieve these transactions with blobs. There is a PR for eth/68 https://github.com/hyperledger/besu/pull/4730
- Transaction pool needs updates (maybe new transaction pool must be created) to handle new types of fees for blobs
- Transaction commitments must be checked per transaction
- We need a short-term storage mechanism for blobs. When a new block gets produced, EL is responsible for storing the blobs until CL asks for them
- New APIs to provide blobs to CL

gezero:
[...]

PR #4778, opened 2022-12-05
Add support for another Transaction Encoder/Decoder

gezero:
Transaction Encoder/Decoder automatically implied RLP encoding. The 4844 Blob Transaction should be encoded with SSZ encoding.

This PR is a suggestion how to refactor the Encoder to support multiple different types of encodings.


gezero:
Implementing SSZ encoding of the transactions might be a first step towards 4844 ( #4631 )

PR #4781, opened 2022-12-05
EIP-4750 EOF Functions

shemnon:
## PR description

Implement CALLF, RETF, and JUMPF operations.  This introduces the 
notion of "code sections" into the code.  For Legacy code section zero
consists of all of the code and is the only code section.  Also 
implements EIP specified code validation to ensure validated EOF code 
won't use CALLF or JUMPF to a non-existent section.

## Documentation

## Changelog


shemnon:
There is the chance the container format will change next week, so we don't need to rush to merge this.

evm/src/main/java/org/hyperledger/besu/evm/EVM.java
```diff
@@ -277,6 +275,18 @@ public void runToHalt(final MessageFrame frame, final OperationTracer tracing) {
           case 0x9f:
             result = SwapOperation.staticOperation(frame, opcode - SWAP_BASE);
             break;
+          case 0xb0: // CALLF
+          case 0xb1: // RETF
+          case 0xb2: // JUMPF
+            // Function operations reset code
+            if (enableShanghai) {
```
diega, line 282: I've been thinking about how to implement this for #4760 and I have some doubts about this mechanism. Do you think this will be enough in a not so distant future when/if `JUMP`/`JUMPI` gets deprecated (or `PC`)? Will we have only two sets of operations, let's say _legacy_ and _oef_, "forever"? Probably we can use it like this so far, but maybe we could revisit this after Shangai

evm/src/main/java/org/hyperledger/besu/evm/code/OpcodesV1.java
```diff
@@ -292,36 +298,57 @@ private OpcodesV1() {
...
+        .orElse(null);
+  }
+
+  /**
+   * validates the code section
+   *
+   * @param code the code section code
+   * @return null if valid, otherwise a string containing an error reason.
+   */
+  static String validateCode(final Bytes code, final int sectionCount) {
```
diega, line 316: Could you create a commit with just these changes so I can cherry-pick it for #4760 and work on top of it? this would help to prevent some conflicts no matter which one gets merged first.
Also, given that we are not using the string response, can we return a boolean and log the failed condition?

diega: On a second thought, I think we can keep the strings so we can use those during testing

shemnon: I'm expecting EIP-4200 (#4760) will merge first, then EIP-4750 (#4781), then EIP-5450 (#4805 ), in immediate sequence when it hits mainline.  I've already merged those two into 5450, so you may be able to cherry pick out the changes into 4760.

shemnon: Ideally most RJumpX changes go in 4760, most function changes go into 5450, and all remaining validation goes in 5450.  Since I expect stack and code validation to be two separate passes for the first implementation I think all code based validation can go in the feature PRs.  The stack validation does have some interactions with rjump and callf, but no interactions on the actual code execution parts, but it depends on stack height and such, not stuff like are we rjumping out of bounds.

evm/src/main/java/org/hyperledger/besu/evm/EVM.java
shemnon: Two loops may have positive performance impacts.  However it also introduces risk if there are cross-version opcode changes.

The advantage of EOF is we can have one "catch all" loop for all of EOF and rely on code validation from permitting bad opcodes from being marked as valid EVM code.

jflo (approved):
Looks good, the only improvement I see would be a little readability around the 16bit conversion. I'm assuming we don't define a method for it for optimization reasons, but perhaps a quick inline comment would aid future readers.

evm/src/main/java/org/hyperledger/besu/evm/EVM.java
```diff
@@ -307,17 +317,8 @@ public void runToHalt(final MessageFrame frame, final OperationTracer tracing) {
...
-    }
-
-    final byte opcode = bytecode.get(offset);
-    final Operation operation = operations.get(opcode);
-    return Objects.requireNonNullElseGet(operation, () -> new InvalidOperation(opcode, null));
+  public Operation[] getOperationsUnsafe() {
+    return operations.getOperations();
   }
 
   public Code getCode(final Hash codeHash, final Bytes codeBytes) {
```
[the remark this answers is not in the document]
shemnon: The intent is that code that failed validation will live in the CodeCache, as an `InvalidCode` instance with the reason and other values (such as returning `false` to `isValid`) to indicate it's invalidity.  The typical mainnet case is an EOA continually re-trying a particular EOF initcode.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeFactory.java
```diff
@@ -64,12 +62,11 @@ public static Code createCode(
...
-        final long[] jumpMap =
-            OpcodesV1.validateAndCalculateJumpDests(layout.getSections()[EOFLayout.SECTION_CODE]);
-        if (jumpMap != null) {
-          return new CodeV1(codeHash, layout, jumpMap);
+        final String validationError = OpcodesV1.validateCode(layout);
+        if (validationError == null) {
+          return new CodeV1(codeHash, layout);
         } else {
-          return new CodeInvalid(codeHash, bytes, "Opcode Validation Failed");
+          return new CodeInvalid(codeHash, bytes, "EOF Code Invalid : " + validationError);
```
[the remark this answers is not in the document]
shemnon: Traces in the core loop has been performance problems.  That overhead to decide if to trace or not and the overhead to load up the objects (even if not converted to a string) is significant enough in a tight loop it is to be avoided.

CodeFactory would be drug into this via any of the Call series operations.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeSection.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.tuweni.bytes.Bytes;
+
+//// java17 convert to record
+public final class CodeSection {
+  final Bytes code;
+  final int inputs;
+  final int outputs;
+  final int maxStackHeight;
+
+  public CodeSection(
```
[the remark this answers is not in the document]
shemnon: done

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
```diff
@@ -58,8 +59,19 @@ public int getSize() {
   }
 
   @Override
-  public Bytes getCodeBytes() {
-    return code;
+  public Bytes getCodeBytes(final int section) {
```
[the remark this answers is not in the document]
shemnon: If `Optional<Code>` wasn't a demonstrated performance problem I would have used that there.  I'm using `null` to indicate that the code doesn't exist and that the API consumer should validate it.

diega (approved):

PR #4786, opened 2022-12-07
Bugfix snapshot transaction segfaults after storage truncation

garyschulte:
## PR description

This PR addresses a race where snap and checkpoint syncs can cause a segfault when starting a worldstate download.  The segfault occurs when rocksdb snapshot transactions are read/written after a storage truncation.  To prevent this and to ensure future uses of clear() and clearFlatDatabase() cannot cause segfaults, this PR:

* adds BonsaiStorageSubscriber type to handle storage events
* removes addCachedLayer from TrieLogManager, moves to AbstractTrieLogManager
* TrieLogManager takes a snapshot immediately in addCachedLayer rather than deferring it 
* notifies subscribers on events that can affect the worldstate, specifically:
  * clear (truncate)
  * clearFlatDatabase (truncating a subset of storage)
  * close (closing the worldstate storage)

Additionally, only drop cached layers by distance from head when the worldstate archive has more than the configured number of retained layers.

## Fixed Issue(s)

addresses #4765 

## Documentation

## Changelog

### DIFF
--- a/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
+++ b/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
@@ -256,16 +256,8 @@ public abstract class AbstractBlockCreator implements AsyncBlockCreator {
               if (ws.isPersistable()) {
                 return ws;
               } else {
-                var wsCopy = ws.copy();
-                try {
-                  ws.close();
-                } catch (Exception ex) {
-                  LOG.error(
-                      "unexpected error closing non-peristable worldstate + "
-                          + parentHeader.toLogString(),
-                      ex);
-                }
-                return wsCopy;
+                // non-persistable worldstates should return a copy which is persistable:
+                return ws.copy();
               }
             })
         .orElseThrow(
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
@@ -22,6 +22,8 @@ import org.hyperledger.besu.ethereum.core.SnapshotMutableWorldState;
 import org.hyperledger.besu.plugin.services.storage.SnappableKeyValueStorage;
 import org.hyperledger.besu.plugin.services.storage.SnappedKeyValueStorage;
 
+import java.util.concurrent.atomic.AtomicLong;
+
 /**
  * This class extends BonsaiPersistedWorldstate directly such that it commits/perists directly to
  * the transaction state. A SnapshotMutableWorldState is used to accumulate changes to a
@@ -35,7 +37,7 @@ public class BonsaiSnapshotWorldState extends BonsaiPersistedWorldState
   private final SnappedKeyValueStorage storageSnap;
   private final SnappedKeyValueStorage trieBranchSnap;
   private final BonsaiWorldStateKeyValueStorage parentWorldStateStorage;
-  private final Long parentWorldStateSubscriberId;
+  private final AtomicLong parentWorldStateSubscriberId = new AtomicLong(Long.MAX_VALUE);
 
   private BonsaiSnapshotWorldState(
       final BonsaiWorldStateArchive archive,
@@ -47,21 +49,27 @@ public class BonsaiSnapshotWorldState extends BonsaiPersistedWorldState
     this.storageSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.storageStorage;
     this.trieBranchSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.trieBranchStorage;
     this.parentWorldStateStorage = parentWorldStateStorage;
-    parentWorldStateSubscriberId = parentWorldStateStorage.subscribe(this);
   }
 
   public static BonsaiSnapshotWorldState create(
       final BonsaiWorldStateArchive archive,
       final BonsaiWorldStateKeyValueStorage parentWorldStateStorage) {
     return new BonsaiSnapshotWorldState(
-        archive,
-        new BonsaiSnapshotWorldStateKeyValueStorage(
-            ((SnappableKeyValueStorage) parentWorldStateStorage.accountStorage).takeSnapshot(),
-            ((SnappableKeyValueStorage) parentWorldStateStorage.codeStorage).takeSnapshot(),
-            ((SnappableKeyValueStorage) parentWorldStateStorage.storageStorage).takeSnapshot(),
-            ((SnappableKeyValueStorage) parentWorldStateStorage.trieBranchStorage).takeSnapshot(),
-            parentWorldStateStorage.trieLogStorage),
-        parentWorldStateStorage);
+            archive,
+            new BonsaiSnapshotWorldStateKeyValueStorage(
+                ((SnappableKeyValueStorage) parentWorldStateStorage.accountStorage).takeSnapshot(),
+                ((SnappableKeyValueStorage) parentWorldStateStorage.codeStorage).takeSnapshot(),
+                ((SnappableKeyValueStorage) parentWorldStateStorage.storageStorage).takeSnapshot(),
+                ((SnappableKeyValueStorage) parentWorldStateStorage.trieBranchStorage)
+                    .takeSnapshot(),
+                parentWorldStateStorage.trieLogStorage),
+            parentWorldStateStorage)
+        .subscribeToParent();
+  }
+
+  protected BonsaiSnapshotWorldState subscribeToParent() {
+   parentWorldStateSubscriberId.set(parentWorldStateStorage.subscribe(this));
+    return this;
   }
 
   @Override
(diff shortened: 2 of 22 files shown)


matkt:
we must check the impact on performance at the time of persistence because we are creating the snapshot at this time

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
```diff
@@ -27,21 +28,26 @@
...
-      final BonsaiSnapshotWorldStateKeyValueStorage snapshotWorldStateStorage) {
+      final BonsaiSnapshotWorldStateKeyValueStorage snapshotWorldStateStorage,
+      final BonsaiWorldStateKeyValueStorage parentWorldStateStorage) {
     super(archive, snapshotWorldStateStorage);
     this.accountSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.accountStorage;
     this.codeSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.codeStorage;
     this.storageSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.storageStorage;
     this.trieBranchSnap = (SnappedKeyValueStorage) snapshotWorldStateStorage.trieBranchStorage;
+    this.parentWorldStateStorage = parentWorldStateStorage;
+    parentWorldStateSubscriberId = parentWorldStateStorage.subscribe(this);
```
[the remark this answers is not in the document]
matkt: and are we sure we manage all the case when the subscribe throw a runtime exception ? I'm not sure there are a lot of case

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
```diff
@@ -62,51 +63,75 @@ public void addCachedLayer(
...
+  @Override
+  public synchronized void onClearFlatDatabase() {
+    dropArchive();
   }
 
-  public static class CachedSnapshotWorldState implements CachedWorldState {
+  private void dropArchive() {
+    // drop all cached snapshot worldstates, they are unsafe when the db has been truncated
+    LOG.info("Key-value storage truncated, dropping cached worldstates");
+    cachedWorldStatesByHash.keySet().forEach(cachedWorldStatesByHash::remove);
```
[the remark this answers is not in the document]
matkt: why not ?  cachedWorldStatesByHash.clear();

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
```diff
@@ -84,5 +92,24 @@ public void close() throws Exception {
...
     trieBranchSnap.close();
+    parentWorldStateStorage.unSubscribe(parentWorldStateSubscriberId);
+  }
+
+  @Override
+  public void onClear() {
+    try {
+      close();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
matkt, line 103: how we  manage this case  in the other level of the code ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -439,4 +435,15 @@ public void rollback() {
       trieLogStorageTransaction.rollback();
     }
   }
+
+  interface BonsaiStorageSubscriber {
+    default void onClear() {}
+    ;
```
matkt, line 441: we can clean this class

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
```diff
@@ -18,41 +18,42 @@
...
       final long maxLayersToLoad,
-      final Map<Bytes32, CachedSnapshotWorldState> cachedWorldStatesByHash) {
+      final Map<Bytes32, CachedWorldState<BonsaiSnapshotWorldState>> cachedWorldStatesByHash) {
     super(blockchain, worldStateStorage, maxLayersToLoad, cachedWorldStatesByHash);
+    worldStateStorage.subscribe(this);
   }
 
   @Override
-  public void addCachedLayer(
+  protected synchronized void addCachedLayer(
```
matkt, line 56: why we need to synchronized this one ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
```diff
@@ -118,8 +143,11 @@ public TrieLogLayer getTrieLog() {
     }
 
     @Override
-    public MutableWorldState getMutableWorldState() {
-      return snapshot.get();
+    public synchronized BonsaiSnapshotWorldState getMutableWorldState() {
```
matkt, line 146: is it easy to use an optional here ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
```diff
@@ -62,51 +63,75 @@ public void addCachedLayer(
...
     final long height;
+    final AtomicBoolean isClosed = new AtomicBoolean(false);
 
     public CachedSnapshotWorldState(
-        final Supplier<BonsaiSnapshotWorldState> snapshotSupplier,
-        final TrieLogLayer trieLog,
-        final long height) {
-      this.snapshot = Suppliers.memoize(snapshotSupplier::get);
+        final BonsaiSnapshotWorldState snapshot, final TrieLogLayer trieLog, final long height) {
+      this.worldStateSubscriberId = snapshot.getWorldStateStorage().subscribe(this);
```
matkt, line 118: we are not already subscribe here ? https://github.com/hyperledger/besu/pull/4786/files#diff-8208cce046b9c78ec905315f4bbbc0a4813abcacd299489acc59833bbce7d35aR50

matkt (dismissed):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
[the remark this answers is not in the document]
garyschulte: since it uses a concurrentHashMap it won't throw, but yeah .clear() is a better option here :+1:

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
garyschulte: 1. SnapshotTrieLogManager - subscribes so it can drop all cached layers after a truncate event
2. BonsaiSnapshotWorldState - subscribes so it can prevent read/write through the snapshot transaction after a truncate event.

If we just had #1, a reference to a snapshot that has already been fetched from the trie log manager can segfault
if we just had #2, then we would keep a bunch of cached worldstates that will refuse to read/write

We need both

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/SnapshotTrieLogManager.java
garyschulte: I considered it, but it was a little messy - I think due to generics IIRC.  I will double check

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
garyschulte: all of the retrievals of SnapshotWorldStates are in a try-with-resources block, so there is implicitly a catch()

garyschulte: this would result in an exception from the caller trying to clear/truncate.  Currently that only happens at the beginning of a fresh snap sync.  I will force it to happen to ensure that snap sync would restart. Really clear() is quite risky with snapshots.  It might be worth overriding it with an iterator-delete based approach if we are using snapshots.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldState.java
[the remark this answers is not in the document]
garyschulte: I moved the subscribe into a fluent member and added it to the static create call :+1:

Re: runtime exceptions, I don't see much opportunity for that to happen within .subscribe(), but I wrapped it in a try-catch just in case

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -256,16 +256,8 @@ private MutableWorldState duplicateWorldStateAtParent() {
               if (ws.isPersistable()) {
                 return ws;
               } else {
-                var wsCopy = ws.copy();
-                try {
-                  ws.close();
```
garyschulte, line 261: @matkt , as discussed - this close() is unnecessary since a non-persistable worldstate does not have resources that need to be cleaned up.  Only layered worldstates are non-persistable

ahamlat (approved):
Approve performance overhead. This PR does not have an effect on block processing performance

PR #4792, opened 2022-12-08
Allow starting PoS networks from genesis

### LINKED ISSUE #4740: MergeProtocolSchedule overrides evmBuilder for post merge specs
### Description
Shandong testnet is PoS from the begining but the transition mechanism is still fired, particularly overriding the _evmBuilder_ with the Paris definition.

As a workaround for syncing Shandong these lines can be removed

https://github.com/hyperledger/besu/blob/2cafa8a25d27ee302a4e42a9769eada9f800e7d9/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeProtocolSchedule.java#L65-L68

but this cannot be applied in general because it will break ETH-Mainnet consensus

diega:
## PR description
This PR uses the _vanilla_ `o.h.b.c.MergeBesuControllerBuilder` when the Total Terminal Difficulty is set to **zero** in the genesis.
The main change is in the HEAD commit (25a870a665b429365281dca244723023d0e7fa03), but for making it readable I had to refactor how the `o.h.b.c.BesuControllerBuilder` was constructed, that's in the previous commit (4369a1c3d48f419d5476ad0cfefc24f488f6d0c6). During these changes, I decided to keep an ignored parameter for not adding complexity to the reviewing process, b/c otherwise I would have to change a lot of tests.

Note: Judging by the comment in [`o.h.b.c.BesuController:216`](https://github.com/hyperledger/besu/blob/22.10.2/besu/src/main/java/org/hyperledger/besu/controller/BesuController.java#L216) it sounds to me that there is a more comprehensive change waiting to be done, but I'm not familiar enough with the merge architecture to implement it in that way yet and maybe this one is good enough for now.

## Fixed Issue(s)
Fixes #4740 

## Documentation

## Changelog


besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -170,59 +169,44 @@ public PluginServiceFactory getAdditionalPluginServices() {
...
+        final GenesisConfigOptions configOptions, final EthNetworkConfig ignoredNetworkConfig) {
+      // wrap with TransitionBesuControllerBuilder if we have a terminal total difficulty:
+      final BesuControllerBuilder builder = internalFromConfigOptions(configOptions);
+      if (configOptions.getTerminalTotalDifficulty().isPresent()) {
+        if (configOptions.getTerminalTotalDifficulty().get().equals(UInt256.ZERO)) {
+          return new MergeBesuControllerBuilder();
+        } else {
+          // TODO this should be changed to vanilla MergeBesuControllerBuilder and the Transition*
+          // series of classes removed after we successfully transition to PoS
+          // https://github.com/hyperledger/besu/issues/2897
```
garyschulte, line 182: Sadly I don't think this is ever going away.  There seems to be an appetite for having merge functionality on testnets

garyschulte (changes requested):
have you tested this PR to see if it has any effect on the test networks you are trying to join?  AFAICT this breaks merge functionality with non-PoW initial chains (like goerli).  Merge logic is unfortunately quite perilous and making a change like this (removing transition* classes) should be very carefully considered.

ajsutton:
> Have we had asks to have a merge-at-genesis network with a TTD of 0? All of the merge-at-genesis test networks I have seen typically have a non-zero TTD that is consumed entirely by the genesis block. I can see the appeal of not needing Transition* classes involved however.

For the record, Teku ATs currently start besu with a TTD of 0 to do merged-genesis networks (https://github.com/ConsenSys/teku/blob/a49f9a8a279a2260c57308fbb313162dfa7e2127/acceptance-tests/src/testFixtures/resources/besu/mergedGenesis.json#L1). It's not hard to change, it just never occurred to me that there'd be any other way of configuring a pre-merged genesis.

diega:
@garyschulte thank you for reviewing it.
The context of this PR was to support the now retired Shandong testnet. [Here](https://github.com/ethereumjs/consensus-deployment-ansible/blob/4220cdb694a0443d345c93b548fbd3a296694c34/shandong-testnet/custom_config_data/genesis.json#L16)'s the genesis used back then.
I confused the transition with the real issue that I was having when I tried to start a new network with a new evm config (ref: #4740) so this would not even solve that case.

PR #4797, opened 2022-12-09
gradle modifications to support java 19

lukelee-sl:
Signed-off-by: lukelee-sl <luke.lee@swirldslabs.com>

Updates to gradle to support java 19.


macfarla (approved):
LGTM

PR #4798, opened 2022-12-09
Runner test refactor

macfarla:
Refactor to make it easier to find the cause when this tests fails. 
* Deduplicated code. 
* Added a call to peerCount to expose peering issues vs syncing issues
* Because of the async wait() you don't get the exact line where the assertion fails. Added comments for common spots where it fails to make this easier. 
* Also 2 minutes is plenty. 

## Documentation

## Changelog


besu/src/test/java/org/hyperledger/besu/RunnerTest.java
```diff
@@ -393,6 +345,24 @@ private void syncFromGenesis(final SyncMode mode, final GenesisConfigFile genesi
...
   }
 
+  private int getNumber(final Response response) throws IOException {
+    final int currentBlock =
+        UInt256.fromHexString(new JsonObject(response.body().string()).getString("result"))
+            .intValue();
+    return currentBlock;
+  }
+
+  @NotNull
```
garyschulte, line 355: what is @NotNull accomplishing in this case?  it is a jetbrains annotation, and won't be checked by findbugs or CI AFAIK.  Also this is on a test method that returns the result of a builder.  

I ask everytime I see the jetbrains import.

besu/src/test/java/org/hyperledger/besu/RunnerTest.java
```diff
@@ -301,61 +261,53 @@ private void syncFromGenesis(final SyncMode mode, final GenesisConfigFile genesi
...
-                                        "{\"jsonrpc\":\"2.0\",\"id\":"
-                                            + Json.encode(7)
-                                            + ",\"method\":\"eth_blockNumber\"}"))
-                                .url(baseUrl)
-                                .build())
-                        .execute()) {
-
-                  assertThat(resp.code()).isEqualTo(200);
+                try (final Response blockNumberResp =
+                    client.newCall(getRequest("eth_blockNumber", baseUrl)).execute()) {
```
garyschulte, line 269: :raised_hands:

garyschulte (approved):
nice refactor.  just a question about the purpose of the annotation

besu/src/test/java/org/hyperledger/besu/RunnerTest.java
macfarla: 👍 good point. removed

PR #4805, opened 2022-12-12
EIP-5450 - EOF Stack Validation

shemnon:
## PR description

EOF stack validation and omnibus validation changes

Builds off of work for EIP-4200 (#4760) and EIP-4750 (#4781) in addition to already committed "small-EOF" changes EOF-3540 + 3670 (#4644).  Changes to the above PRs will be merged in as updates are committed.

## Documentation

## Changelog


evm/src/test/java/org/hyperledger/besu/evm/code/CodeV1Test.java
```diff
@@ -367,4 +374,380 @@ private static Stream<Arguments> immediateContainsOpcodeArguments() {
...
+
+  /**
+   * Vectors from an early <a
+   * href="https://github.com/ipsilon/eof/commit/c1a2422ddbb72db48bacd2406ed7dad28567b403#diff-ce64373581560ddf02962cb731dfa06457d5ff8615fe12b63c881ab97432f1cf">Ipsilon
+   * Prototype</a>
+   *
+   * @return parameterized test vectors
+   */
+  static Stream<Arguments> stackEmpty() {
+    return Stream.of(Arguments.of("Empty", null, 0, List.of(List.of("00", 0, 0, 0))));
```
diega, line 419: Just for future reference, there is an alternative way to name parameterized tests in Junit5. That would be something like the following, adding static imports for `org.junit.jupiter.params.provider.Arguments#arguments` and `org.junit.jupiter.api.Named#named`
```suggestion
    return Stream.of(arguments(named("Empty", null, 0, List.of(List.of("00", 0, 0, 0)))));
```
And you won't need the `ignoredName` field in the test.
More about this in the [Customizing Display Names](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-display-names) section of the documentation.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
```diff
@@ -38,6 +573,216 @@ public class CodeV1 implements Code {
...
+
+        byte[] stackInfo = opcodeStackValidation[thisOp];
+        int stackInputs;
+        int stackOutputs;
+        int pcAdvance = stackInfo[2];
+        if (pcAdvance == 0) {
+          return String.format("Invalid opcode 0x%02X", thisOp);
+        }
+        if (thisOp == CallFOperation.OPCODE || thisOp == JumpFOperation.OPCODE) {
+          int section = ((code[currentPC + 1] & 0xff) << 8) | (code[currentPC + 2] & 0xff);
```
jflo, line 729: Can use the shiny new Words.readBigEndianU16() here.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
```diff
@@ -38,6 +573,216 @@ public class CodeV1 implements Code {
...
+
+    if (maxStackHeight != codeSections[codeSectionToValidate].maxStackHeight) {
+      return String.format(
+          "Calculated max stack height (%d) does not match reported stack height (%d)",
+          maxStackHeight, codeSections[codeSectionToValidate].maxStackHeight);
+    }
+    return null;
+  }
+
+  private static int signedShortAt(final byte[] code, final int currentPC) {
```
jflo, line 782: Could migrate this to Words with his U16 friend. Also, had to read this twice, as it is not a signedShort at the PC, but rather right after it. Maybe `signedShortAfter` is a better name?

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
```diff
@@ -38,6 +573,216 @@ public class CodeV1 implements Code {
...
+    int codeLength = code.length;
+    int[] stackHeights = new int[codeLength];
+    Arrays.fill(stackHeights, -1);
+
+    int thisWork = 0;
+    int maxWork = 1;
+    int[][] workList = new int[codeLength][2];
+
+    int initialStackHeight = codeSections[codeSectionToValidate].getInputs();
+    int maxStackHeight = initialStackHeight;
```
jflo, line 694: is `stackHighWater` a better name, since this isn't really a limit, but rather a running max of the stack depth we update over time?

jflo (approved):
Looks good! Some aesthetic suggestions to improve readability, none blocking.

diega (approved):
This looks really great. Full-EOF here we go!

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
```diff
@@ -38,6 +573,216 @@ public class CodeV1 implements Code {
...
+        .filter(Objects::nonNull)
+        .findAny()
+        .orElse(null);
+  }
+
+  /**
+   * validates the code section
+   *
+   * @param code the code section code
+   * @return null if valid, otherwise a string containing an error reason.
```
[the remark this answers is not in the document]
shemnon: Committing in docs to a "first" error corners us from refactoring the code to do validations in a different or parallel order.  Especially when analyzing independent code segments and the inherent out-of-order processing of stack validation errors.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
shemnon: Yea, this was written before the callf features merged.  will migrate.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
shemnon: Reference code in the EIP uses `max_stack_height`, so I think the intent is clear enough.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV1.java
shemnon: Done, used I16 variant (U16 is unsigned)

evm/src/test/java/org/hyperledger/besu/evm/code/CodeV1Test.java
shemnon: named only takes one "payload" type and does not destructure  the list.  I'll revisit this when we move to Java 17.

PR #4807, opened 2022-12-12
Non blocking RPC endpoint

daniellehrner:
## PR description

This PR converts the standard RPC endpoint in a non blocking one.

## Documentation

## Changelog


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceTestBase.java
```diff
@@ -68,29 +68,37 @@ public class JsonRpcHttpServiceTestBase {
   @ClassRule public static final TemporaryFolder folder = new TemporaryFolder();
   protected final JsonRpcTestHelper testHelper = new JsonRpcTestHelper();
 
-  private static final Vertx vertx = Vertx.vertx();
```
daniellehrner, line 71: The static variables caused test failures, because the tests are running in parallel and they were interfering with each other. The change to non-static variables resolves this issue.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceParameterizedTest.java
```diff
@@ -39,17 +38,11 @@ public JsonRpcHttpServiceParameterizedTest(final String requestJson) {
     this.requestJson = requestJson;
   }
 
-  @BeforeClass
-  public static void setup() throws Exception {
+  @Before
```
daniellehrner, line 41: The tests were interfering with each other, by using the same setup for all of them. Now every test has their own isolated setup.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceTest.java
```diff
@@ -464,6 +457,9 @@ public void ethGetUncleCountByBlockNumberPending() throws Exception {
 
   @Test
   public void ethGetUncleCountByBlockNumberPendingNoData() throws Exception {
+    when(blockchainQueries.headBlockNumber()).thenReturn(0L);
```
daniellehrner, line 460: The test had a hidden dependency on another test. It relied on one of the tests before it to set up the mocking. After the change to non-static variables this dependency became visible and the tests was failing. It was necessary to set up its own mocking configuration.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceTest.java
```diff
@@ -2018,6 +2015,12 @@ public void ethGetStorageAtBlockNumber() throws Exception {
 
   @Test
   public void ethGetStorageAtInvalidParameterStorageIndex() throws Exception {
+    when(blockchainQueries.getBlockchain()).thenReturn(blockchain);
```
daniellehrner, line 2018: Same as in the [comment above](https://github.com/hyperledger/besu/pull/4807/files#r1091801204).

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/JsonRpcExecutorVerticleTest.java
```diff
@@ -0,0 +1,218 @@
...
+            });
+
+    assertThat(testContext.awaitCompletion(5, TimeUnit.SECONDS)).isTrue();
+    if (testContext.failed()) {
+      throw testContext.causeOfFailure();
+    }
+  }
+
+  @Test
+  void exceedingMaxBatchSizeShouldCauseFailure(
```
siladu, line 98: What dictates maxBatch in a production scenario? 
IOW what must happen in a real-world scenario in order to trigger this error?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/HandlerFactory.java
```diff
@@ -51,6 +54,11 @@ public static Handler<RoutingContext> jsonRpcExecutor(
       final JsonRpcExecutor jsonRpcExecutor,
       final Tracer tracer,
       final JsonRpcConfiguration jsonRpcConfiguration) {
-    return JsonRpcExecutorHandler.handler(jsonRpcExecutor, tracer, jsonRpcConfiguration);
+    return BlockingJsonRpcExecutorHandler.handler(jsonRpcExecutor, tracer, jsonRpcConfiguration);
+  }
+
+  public static Handler<RoutingContext> jsonRpcExecutor(
```
siladu, line 60: name this nonBlockingJsonRpcExecutor and the other one blockingJsonRpcExecutor?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/NonBlockingJsonRpcExecutorHandler.java
```diff
@@ -0,0 +1,326 @@
...
+import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.auth.User;
+import io.vertx.ext.web.RoutingContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NonBlockingJsonRpcExecutorHandler implements Handler<RoutingContext> {
```
siladu, line 59: Seems like quite a lot of shared logic between this and the Blocking version. Is there a plan to deduplicate some of this?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/EventBusAddress.java
```diff
@@ -0,0 +1,30 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.api.jsonrpc;
+
+public enum EventBusAddress {
+  RPC_EXECUTE_OBJECT("ethereum.api.json.rpc.executor.object"),
+  RPC_EXECUTE_ARRAY("ethereum.api.json.rpc.executor.array");
```
siladu, line 19: Why do we need to process object and array separately?

siladu (commented):
How does this PR interact with other vertx blocking/non-blocking contructs? e.g. https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java#L67

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/JsonRpcExecutorArrayRequest.java
```diff
@@ -0,0 +1,40 @@
...
+      final Supplier<Boolean> alive,
+      final JsonArray jsonArray) {
+    super(optionalUser, spanContext, alive);
+    this.jsonArray = jsonArray;
+  }
+
+  public JsonArray getJsonArray() {
+    return jsonArray;
+  }
+}
```
macfarla, line 40: do we want a toString method here too?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/JsonRpcExecutorVerticle.java
```diff
@@ -0,0 +1,117 @@
...
+
+    requestMessage.reply(jsonRpcResponse);
+  }
+
+  private void handleJsonRpcExecutorArrayRequest(
+      final Message<JsonRpcExecutorArrayRequest> requestMessage) {
+    final JsonRpcExecutorArrayRequest request = requestMessage.body();
+    final JsonArray jsonArray = request.getJsonArray();
+
+    if (jsonRpcConfiguration.getMaxBatchSize() > 0
```
macfarla, line 84: this would mean if I set maxBatchSize = 0 it would be effectively unlimited, right?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/JsonRpcExecutorVerticle.java
```diff
@@ -0,0 +1,117 @@
...
+        && jsonArray.size() > jsonRpcConfiguration.getMaxBatchSize()) {
+      requestMessage.fail(
+          EXCEEDS_RPC_MAX_BATCH_SIZE.getCode(), EXCEEDS_RPC_MAX_BATCH_SIZE.getMessage());
+      return;
+    }
+
+    JsonRpcResponse[] jsonRpcBatchResponses = new JsonRpcResponse[jsonArray.size()];
```
macfarla, line 91: can this be final

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/JsonRpcError.java
```diff
@@ -263,4 +263,14 @@ public static JsonRpcError fromJson(
...
   }
+
+  public static JsonRpcError fromCodeAndMessage(final int code, final String message) {
+    for (final JsonRpcError error : JsonRpcError.values()) {
+      if (error.code == code && error.message.equals(message)) {
+        return error;
+      }
+    }
+
+    return null;
```
macfarla, line 274: is this a "this should never happen" return null?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/JsonRpcExecutorVerticle.java
macfarla: I think this is the same as the current behaviour - 
-1 is unlimited
>=1 sets a limit
0 doesn't really make sense so maybe we should just doc that 0 is unlimited, same as -1

daniellehrner:
This diagram shows the message flow

Non_Blocking_RPC

macfarla:
@daniellehrner will this work with the pipeline solution #5131

daniellehrner:
There is a breaking change in this PR: [here](https://github.com/hyperledger/besu/blob/91a629c76ee57a22b21b9787d6e369e21999999c/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/NonBlockingJsonRpcExecutorHandler.java#L275)

Basically it means any RPC calls to any of the endpoints of eea_  and priv_  can not be processed anymore within a batch request. As far as I have seen, batch requests are not documented, but could still used in several occasions.

The problem is that those two namespaces use the Vertx web client to make calls to Tessera and in combination with non blocking calls the calls by the web client never return. I tried to debug this for a long time, but could not find the root cause and the only way forward was to exclude them from batch requests.

In the mean time instead of the non blocking calls we have added BesuPipeline to increase the performance of the tracing endpoints. So I am not sure anymore if this tradeoff in this PR is worth it.

PR #4810, opened 2022-12-13
Docker release task for latest tags

jframe:
## PR description
Expands the github release workflow to support adding the latest tags for all the docker variants and architectures.

* Replace the use of git sha tags with existing release tags to avoid creating a lot of unneeded tags. When creating the latest tags for all the variants, arch and manifests 15 tags are needed in total
* Added latest tags for variants as was done for previous releases
* Added latest tags for all the architectures
* Added latest tags for multi-arch images

## Documentation

## Changelog

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,11 +6,12 @@ jobs:
   dockerPromoteX64:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
       - name: Login to DockerHub
         run: echo ${{ secrets.DOCKER_PASSWORD_RW }} | docker login -u ${{ secrets.DOCKER_USER_RW }} --password-stdin
-      - name: pull tagged image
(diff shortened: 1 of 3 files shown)


jflo:
This is a different approach, but certainly viable. Using this release action, we would still require a build to be run and the release tag to be applied so that it can be promoted. The previous approach allowed for promoting a snapshot. 

Both approaches allow us to iterate on an image and defer application of the `latest` tag till when the release is cut.

jframe:
> This is a different approach, but certainly viable. Using this release action, we would still require a build to be run and the release tag to be applied so that it can be promoted. The previous approach allowed for promoting a snapshot.
> 
> Both approaches allow us to iterate on an image and defer application of the `latest` tag till when the release is cut.

This approach doesn't require a build to be run. It does a docker pull to retrieve the image on the docker registry before applying the tags and pushing them.

But it wouldn't work on promoting a snapshot we can't identity a specific build for the snapshot. So using the sha would be a benefit if wanted to keep that ability.

jflo (approved):

PR #4814, opened 2022-12-14
Shanghai timestamp refactored

siladu:
Corrected version of https://github.com/hyperledger/besu/pull/4743


siladu:
Corrected the commits on https://github.com/hyperledger/besu/pull/4743 so this is no longer needed

PR #4819, opened 2022-12-14
Use safe block as pivot block during snapsync

matkt:

## PR description

I found that if we only use finalized block during snapsync is not a good choice because we will have 60 blocks windows to download and heal. It can impact the performance of the sync because we will not have enough time to download a lof of leafs and heal etc. I found
A finalized block is one which has been accepted as canonical by more than 2/3 of validators. To create a conflicting block, an attacker would have to burn at least 1/3 of the total stake.
A safe head block is one which, under normal network conditions, is expected to be included in the canonical chain. Assuming network delays of less than 4 seconds, an honest majority of validators and no attacks on the fork-choice rule, the safe head will never be orphaned.
I think we can use safe block during a sync

## Documentation

## Changelog


fab-10 (approved):

garyschulte (approved):
🚢

PR #4828, opened 2022-12-15
Allow -Pdocker-platform during distDocker

gezero:
As an example of an use case - this should make it easier to build linux/amd64 images on M1.


macfarla (approved):
LGTM

Issue #2681, opened 2021-08-26
Update JavaDoc so we can build with Java 16+


shemnon:
Currently Hypedlerger Besu won't fully build when using Java 16 as the JVM that is running the gradle build. One reason is that some of javadoc rules in have changed.

There are two main items that are hitting us
* Not all public classes have comments. Fixing these can be tedious, but it appears our warning setting lets us get away with it.
* Some places we use out-of-order &lt;H3&gt; headings and the javadoc linter really hates that.

If we can get away without adding javadoc on all public methods that would be great.  If we do have to do thousands of javadocs we should consider the bare-minimum needed, without blatently saying "TODO" on all of them. This can be fixed possibly with CLI flags in the build?

The other lint options should be fixed.  The heading order ones in particular.  But if there are syntax and reference bugs those should be fixed.

joyita07:
Can I work on this?

macfarla:
👋 Hi @joyita07 are you still interested in working on this?

PR #4834, opened 2022-12-16
Fix javadocs to allow doc lint to pass in JDK 16 and above.

usmansaleem:
## PR description
Adding missing javadocs so that javadoc doclint passes against JDK 16+ (invoke by Besu gradle build).

Exclude following packages from javadoc lint:
- org.hyperledger.besu.privacy.contracts.generated
- org.hyperledger.besu.tests.acceptance.*

Temporarily exclude ethereum and evm submodule for doc lint checks.

Important diffs:
- https://github.com/hyperledger/besu/pull/4834/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7
- https://github.com/hyperledger/besu/pull/4834/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47
- https://github.com/hyperledger/besu/pull/4834/files#diff-3d103fc7c312a3e136f88e81cef592424b8af2464c468116545c4d22d6edcf19

Fixed Issue(s)
https://github.com/hyperledger/besu/issues/2681

## Documentation

## Changelog


jflo (approved):
Massive PR, I did not read every single javadoc, but the ones I spot checked are all reasonable. We don't expect these to be in-depth descriptions, but we can improve that over time.

Uzi you absolute legend, thanks for doing this thankless task.

garyschulte:
> Uzi you absolute legend, thanks for doing this thankless task.

seconded. 👍

siladu:
Kudos on this mega PR @usmansaleem! I've been playing trying to break the build by adding an un-javadoc'd method... 

> invoke by Besu gradle build

If you meant `./gradlew build` then this isn't what I observed, I had to specifically run `./gradlew javadoc` on java 16+ (I used java 17) to see errors.

It seems to be configured to run in circle CI (as part of the integration tests for some reason): https://github.com/hyperledger/besu/blob/main/.circleci/config.yml#L200-L203
however this won't break since circle ci is java 11.

So there's currently nothing to stop people committing non-compliant code onto main AFAICT.

Is there a way to force javadoc to throw errors even on java 11 so it will fail the build before we move to java 17?

usmansaleem:
@siladu 
> Is there a way to force javadoc to throw errors even on java 11 so it will fail the build before we move to java 17?

It seems to me a javadoc/doclint has a bug in Java 11 which seems to be fixed post java 15. I was trying to see if there are any defaults which were disabled in 11 but enabled in Java 15+, but it doesn't seem so. What seems to be the plan to move to CircleCI build to Java 17, the current LTS?

siladu:
> What seems to be the plan to move to CircleCI build to Java 17, the current LTS?

Last I heard we were planning to stay on CircleCI, rather than moving to github actions (which would have supported java 17). 
I'm not sure of the java 17 status in general, I've asked on Discord. Hopefully it's just a case of updating CircleCI as you say 🤞

PR #4836, opened 2022-12-18
Optimize Sstore operation execution time

ahamlat:

## PR description
This PR aims to improve SStore operation execution time by getting the original value and current value only once during gas cost and refund calculation. Currently, we make these two calls twice, and even with RocksDB cache, this has an impact on SStore performance as we can see in the CPU profiling flame graph below.
image

After this PR, we can see that getSorageValue (get current value) and getOriginalStorageValue are called only once.
image

## Documentation

## Changelog


shemnon (changes requested):
The CodeQL warnings all appear legitimage, so please address them.

As this is changing an API on an interface it will be a breaking change and will need to land in 23.1.0-beta and not be cherry picked into 22.10.4.
Also, add a changelog entry under breaking changes.

evm/src/main/java/org/hyperledger/besu/evm/operation/SStoreOperation.java
```diff
@@ -46,7 +48,7 @@ public long getMinimumGasRemaining() {
   public OperationResult execute(final MessageFrame frame, final EVM evm) {
 
     final UInt256 key = UInt256.fromBytes(frame.popStackItem());
-    final UInt256 value = UInt256.fromBytes(frame.popStackItem());
+    final UInt256 newValue = UInt256.fromBytes(frame.popStackItem());
```
garyschulte, line 51: is this rename intentional or leftover from development?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/LondonGasCalculator.java
```diff
@@ -40,14 +41,19 @@ public long getSelfDestructRefundAmount() {
...
+      final Supplier<UInt256> originalValue) {
+
+    UInt256 localCurrentValue = currentValue.get();
+    if (localCurrentValue.equals(newValue)) {
       return 0L;
     } else {
-      final UInt256 originalValue = account.getOriginalStorageValue(key);
-      if (originalValue.equals(currentValue)) {
-        if (originalValue.isZero()) {
+      UInt256 localOriginalValue = originalValue.get();
```
garyschulte, line 54: effectively final, can be final

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/IstanbulGasCalculator.java
```diff
@@ -74,15 +79,19 @@ public long calculateStorageCost(
...
+      final Supplier<UInt256> originalValue) {
+
+    UInt256 localCurrentValue = currentValue.get();
+    if (localCurrentValue.equals(newValue)) {
       return 0L;
     } else {
-      final UInt256 originalValue = account.getOriginalStorageValue(key);
-      if (originalValue.equals(currentValue)) {
-        if (originalValue.isZero()) {
+      UInt256 localOriginalValue = originalValue.get();
```
garyschulte, line 92: final again

garyschulte (approved):
:+1:  just one nitpick about omitting final for local variables in gas calculators. Perhaps you had a reason to omit final?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/LondonGasCalculator.java
ahamlat: done

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/IstanbulGasCalculator.java
ahamlat: done

shemnon (approved):

PR #4841, opened 2022-12-20
EIP-6122 Timestamp based fork IDs

jframe:
## PR description
Timestamp based fork IDs implementing [EIP-6122](https://github.com/ethereum/EIPs/pull/6122)

## Documentation

## Changelog


config/src/main/java/org/hyperledger/besu/config/GenesisConfigOptions.java
```diff
@@ -104,7 +104,9 @@ default boolean isConsensusMigration() {
 
   Optional<Hash> getTerminalBlockHash();
 
-  List<Long> getForks();
+  List<Long> getForkBlockNumbers();
+
+  List<Long> getForkTimestamps();
```
siladu, line 109: nit: feel like getForkNumbers or getForkBlockTimestamps would be more consistent since they both represent blocks, but happy to be convinced otherwise :)

config/src/main/java/org/hyperledger/besu/config/JsonGenesisConfigOptions.java
```diff
@@ -590,4 +590,17 @@ public List<Long> getForks() {
         .sorted()
         .collect(Collectors.toList());
   }
+
+  @Override
+  public List<Long> getForkTimestamps() {
+    Stream<OptionalLong> forkBlockTimestamps = Stream.of(getShanghaiTime(), getCancunTime());
+    // when adding forks add an entry to ${REPO_ROOT}/config/src/test/resources/all_forks.json
```
siladu, line 597: useful comment 👍 

nit: out of scope of your changes but wondering if/why we can't use the already built list from the Schedules?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/forkid/ForkIdManager.java
```diff
@@ -34,52 +36,81 @@
 public class ForkIdManager {
 
   private final Hash genesisHash;
-  private final List<ForkId> forkIds;
+  private final List<ForkId> blockNumbersForkIds;
+  private final List<ForkId> timestampsForkIds;
```
siladu, line 40: keep reading this as Timestamps For Kids 😆

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/forkid/ForkIdManager.java
```diff
@@ -162,21 +193,34 @@ private long createForkIds() {
     final CRC32 crc = new CRC32();
     crc.update(genesisHash.toArray());
     genesisHashCrc = getCurrentCrcHash(crc);
-    final List<Bytes> forkHashes = new ArrayList<>(List.of(genesisHashCrc));
-    forkBlockNumbers.forEach(
+    final List<Bytes> numberForkHashes = new ArrayList<>(List.of(genesisHashCrc));
```
siladu, line 196: Bit confused by what "number" means here as opposed to just forkHashes?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/forkid/ForkIdManager.java
```diff
@@ -162,21 +193,34 @@ private long createForkIds() {
...
+          new ForkId(numberForkHashes.get(blockNumberForks.size() + i), timestampForks.get(i)));
     }
     long forkNext = 0;
-    if (!forkBlockNumbers.isEmpty()) {
-      forkNext = forkIds.get(forkIds.size() - 1).getNext();
-      forkIds.add(new ForkId(forkHashes.get(forkHashes.size() - 1), 0));
+    if (!timestampForks.isEmpty()) {
+      forkNext = timestampForks.get(timestampForks.size() - 1);
+      timestampsForkIds.add(new ForkId(numberForkHashes.get(numberForkHashes.size() - 1), 0));
+    } else if (!blockNumberForks.isEmpty()) {
```
siladu, line 221: Why is this else if and not just if?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/forkid/ForkIdTest.java
```diff
@@ -667,18 +743,276 @@ public static Collection<Object[]> data() {
...
             empty(),
             empty(),
             wantPeerCheck("0xafec6b27", 0L, false)
-          }
+          },
+
+          // Fork ID test cases with block number and timestamp based forks
+          // Withdrawals test cases
+          {
+            "Withdrawals // Unsynced",
```
siladu, line 758: Withdrawals test cases worthy of its own test class?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/forkid/ForkIdTestUtil.java
```diff
@@ -79,6 +84,13 @@ public static class Forks {
     public static final List<Long> SHANDONG =
         Arrays.asList(0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L);
     public static final List<Long> PRIVATE = Arrays.asList(0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L);
+
+    public static final List<Long> WITHDRAWALS_BLOCKS =
```
siladu, line 88: s/WITHDRAWALS/SHANGHAI ?

config/src/main/java/org/hyperledger/besu/config/JsonGenesisConfigOptions.java
jframe: It's done deliberately to catch any mismatches between what we would have in the config file with mixed case and parsing as lower case. There is a pretty good comment about this in the GenesisConfigFileTest https://github.com/hyperledger/besu/blob/main/config/src/test/java/org/hyperledger/besu/config/GenesisConfigFileTest.java#L534.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/forkid/ForkIdManager.java
jframe: The next forkId should be added to only one of the timestamp forks or the blockNumber forks. Using an if instead of the else if would break in the case where we had a non-empty blockNumberForks and non-empty timestampForks.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/forkid/ForkIdManager.java
jframe: Agree, forkHashes makes more sense. Renamed it.

config/src/main/java/org/hyperledger/besu/config/GenesisConfigOptions.java
jframe: I like getForkBlockTimestamps. Changing it to that.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/forkid/ForkIdTest.java
jframe: Doesn't seem like quite the right split. But this test class is getting quite large. Maybe should split the test data out into test files? Although I suggest another PR to do that.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/forkid/ForkIdTestUtil.java
jframe: Think WITHDRAWALS makes more sense as the rest are are network names not fork names

siladu: Why is withdrawals more like a network than shanghai? I think it's just one of several features within shanghai 🤷

jframe: Because withdrawals in this case represents the withdrawals testnet. Shanghai is the fork so don't want to use that. But could just sidestep this and put these into the mainnet configuration instead.

siladu (approved):
LGTM!

PR #4842, opened 2022-12-20
Ignore extra RocksDB column families for experimental features

### LINKED ISSUE #4831: Ability to downgrade besu with chain pruning experimental feature
### Description
Currently with chain pruning feature merged into main, a new column family is introduced which prevents user from downgrading. 

### Acceptance Criteria
Ability to downgrade besu from chain pruning experimental feature.

wcgcyx:
Signed-off-by: Zhenyang Shi <wcgcyx@gmail.com>

## PR description
This PR adds the support to ignore certain column family to support having experimental feature store data in rocksdb. 

## Fixed Issue(s)

fixes #4831 

## Documentation

## Changelog


plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -93,10 +96,33 @@ public RocksDBColumnarKeyValueStorage(
...
+      final List<SegmentIdentifier> trimmedSegments = new ArrayList<>(segments);
+      final List<byte[]> existingColumnFamilies =
+          RocksDB.listColumnFamilies(new Options(), configuration.getDatabaseDir().toString());
+      ignorableSegments.forEach(
+          ignore -> {
+            trimmedSegments.remove(ignore);
+            existingColumnFamilies.forEach(
+                existed -> {
+                  if (Arrays.equals(existed, ignore.getId())) {
+                    trimmedSegments.add(ignore);
```
jframe, line 120: If we have a new database with no existing column families can the ignorable segment be used?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -93,10 +96,33 @@ public RocksDBColumnarKeyValueStorage(
...
+                  }
+                });
```
jframe, line 122: Can you add a test for this

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -93,10 +96,33 @@ public RocksDBColumnarKeyValueStorage(
...
+  public RocksDBColumnarKeyValueStorage(
+      final RocksDBConfiguration configuration,
+      final List<SegmentIdentifier> segments,
+      final List<SegmentIdentifier> ignorableSegments,
+      final MetricsSystem metricsSystem,
+      final RocksDBMetricsFactory rocksDBMetricsFactory)
+      throws StorageException {
 
     try (final ColumnFamilyOptions columnFamilyOptions = new ColumnFamilyOptions()) {
+      final List<SegmentIdentifier> trimmedSegments = new ArrayList<>(segments);
```
jframe, line 111: Will this approach still mean it's not possible to move back to an old version of Besu e.g. 22.10.3 that doesn't have this ignore option? So we would still need to add a changelog entry having the pruning option as a breaking change for backwards compatibility?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBPlugin.java
```diff
@@ -45,6 +47,10 @@ public RocksDBPlugin() {
     this.options = RocksDBCLIOptions.create();
   }
 
+  public void addIgnorableSegmentIdentifier(final SegmentIdentifier ignorable) {
```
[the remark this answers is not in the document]
wcgcyx: I was actually trying to avoid changing the storage API by having this only applies to the internal rocksDB plugin for now.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
wcgcyx: yes, the logic is that only ignore given column family if 1, it is marked to be ignored (for example, chain pruning is disabled) and 2, it does not exist in the database. This means if besu starts from a new database and chain pruning is disabled, it will ignore and create/load a db with one less column family.

wcgcyx: There could be three cases in total:
1. There is chain pruning segment in the db already and regardless of whether chain pruning is enabled, the extra column family can't be ignored otherwise besu will fail to start.
2. There isn't chain pruning segment in the db and chain pruning is disabled by default, besu will ignore the extra column family and the db will be backward compatible. (This will be de default behaviour as chain pruning is hidden) 
3. There isn't chain pruning segment in the db but chain pruning is enabled, besu will permanently add the extra column family and the db will be backward incompatible.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
wcgcyx: yep, let me try to add a test for this

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
wcgcyx: See above, as long as the chain pruning is not enabled, the db will be backward compatible (which is the default behaviour).

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -93,10 +96,33 @@ public RocksDBColumnarKeyValueStorage(
...
+      final List<byte[]> existingColumnFamilies =
+          RocksDB.listColumnFamilies(new Options(), configuration.getDatabaseDir().toString());
+      ignorableSegments.forEach(
+          ignore -> {
```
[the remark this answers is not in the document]
wcgcyx: Will do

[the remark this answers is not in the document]
wcgcyx: Updated

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
wcgcyx: Tests added

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -93,10 +96,35 @@ public RocksDBColumnarKeyValueStorage(
...
+            trimmedSegments.remove(ignorableSegment);
+            existingColumnFamilies.forEach(
+                existed -> {
+                  if (Arrays.equals(existed, ignorableSegment.getId())) {
+                    // Add back the segment because column family can not be ignored if it exists
+                    // currently.
+                    trimmedSegments.add(ignorableSegment);
+                  }
+                });
+          });
```
jframe, line 125: minor: Is there a more functional way of writing this with streams?

jframe (approved):

Issue #4794, opened 2022-12-08
Improve "No sync target, checking current peers for usefulness" logging


mark-terry:
### Description
The "No sync target, checking current peers for usefulness" could be improved:

- It has a tendency of being spammy
- The impact of the logged behaviour isn't clear

### Acceptance Criteria

- Reduce spam - fixed interval logging?
- Provide more clarity on what the message means - add more detail, appropriate level etc.

PR #4853, opened 2022-12-21
'No sync target' logging improvements

mark-terry:

## PR description
"No sync target, checking current peers for usefulness" logging made less spammy, with improved detail.
- DEBUG level log message rate reduced to once every 15 seconds, with pivot block target added to logging message.
- INFO level log message added at 2 minute intervals to provide user feedback while import is paused

## Fixed Issue(s)

Addresses #4794

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncTargetManager.java
```diff
@@ -149,4 +167,16 @@ public boolean shouldContinueDownloading() {
         || !worldStateStorage.isWorldStateAvailable(
             pivotBlockHeader.getStateRoot(), pivotBlockHeader.getBlockHash());
   }
+
+  private void startLogDelay(final AtomicBoolean shouldLog, final int logRepeatDelay) {
+    new Timer()
```
siladu, line 172: I think ScheduledThreadPoolExecutor is the natural successor to Timer.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncTargetManager.java
```diff
@@ -67,10 +74,21 @@ public FastSyncTargetManager(
   protected CompletableFuture<Optional<EthPeer>> selectBestAvailableSyncTarget() {
     final BlockHeader pivotBlockHeader = fastSyncState.getPivotBlockHeader().get();
     final Optional<EthPeer> maybeBestPeer = ethContext.getEthPeers().bestPeerWithHeightEstimate();
-    if (!maybeBestPeer.isPresent()) {
-      LOG.debug(
-          "No sync target, checking current peers for usefulness: {}",
-          ethContext.getEthPeers().peerCount());
+    if (maybeBestPeer.isEmpty()) {
+      if (shouldLogDebug.compareAndSet(true, false)) {
```
siladu, line 78: I know log throttling is done in at least one other place: https://github.com/hyperledger/besu/blob/7d5988d4b5aa9db12827f2256fb575ddd3ea6246/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdated.java#L287-L296

Would be good to have a single way to do it in a logging util. I can think of several places we could use it.

diega: Has the usage of a [BurstFilter](https://logging.apache.org/log4j/2.x/manual/filters.html#BurstFilter) been analyzed for this use case?

siladu: I think @fab-10 may have looked into that or some similar feature of log4j. I agree that would be a neater approach and can't see a reason why we wouldn't want to use it for all logs

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/util/LogUtil.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class LogUtil {
+  public static void throttledLog(
+      final Consumer<String> logger,
+      final String logMessage,
+      final AtomicBoolean shouldLog,
+      final int logRepeatDelay) {
+    if (shouldLog.compareAndSet(true, false)) {
+      logger.accept(logMessage);
+      final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
siladu, line 31: If we end up calling `throttledLog` from multiple places, could we end up with performance issues with too many threads?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncTargetManager.java
fab-10: I haven't yet looked at how to throttle this kind of logs, what I did was about summarizing the progress of the sync every tot seconds, this seems different, and I agree that something like the BurstFilter make sense here, since it could also be easily configurable.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/util/LogUtil.java
fab-10: I agree creating the executor on every invocation is overkill in terms of performance, you can use the `ForkJoinPool.commonPool` here

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/util/LogUtil.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright ConsenSys AG.
```
fab-10, line 2: change to Hyperledger Besu contributors

fab-10 (changes requested):
Change request to avoid creating a thread on very log call. But if possible evaluate the usage of a BurstFilter for throttling logs

mark-terry:
Will review BurstFilter. 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncTargetManager.java
mark-terry: From my investigation, BurstFilter doesn't appear to be specific enough to address this use case. My understanding is that it throttles the entire log level, whereas this functionality allows a specific (potentially spammy) log message to be throttled. Is my understanding of BurstFilter correct, or is it able to be targeted to a class/method level?

A logging improvement via log4j configuration would likely also require the user to configure it.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/util/LogUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.function.Consumer;
+
+public class LogUtil {
+  public static void throttledLog(
+      final Consumer<String> logger,
+      final String logMessage,
+      final AtomicBoolean shouldLog,
+      final int logRepeatDelay) {
+
+    final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
fab-10, line 30: Make this a static field of the class, so it only create once and there reused on any call

fab-10 (changes requested):

fab-10 (approved):

PR #4854, opened 2022-12-21
not save invalid trie log

matkt:

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
```diff
@@ -144,6 +144,13 @@ private void updateAccountStorage(
   public void persist(final BlockHeader blockHeader) {
     final BonsaiWorldStateUpdater localUpdater = updater.copy();
     final Hash newWorldStateRootHash = rootHash(localUpdater);
+    if (!newWorldStateRootHash.equals(blockHeader.getStateRoot())) {
+      throw new RuntimeException(
```
garyschulte, line 148: I think this is the right behavior and should prevent creating trie logs that can corrupt the persisted state.  

AFAICT there isn't a way to differentiate an incorrect header stateroot versus a storage inconsistency.  If we throw a StorageException instead, engine_newPayload should not report INVALID but rather returns an error response.  I am not sure which behavior we would want in this case, but a unit test for engine_getPayload would be a good idea just so we can be explicit about the behavior we expect.

Issue #4784, opened 2022-12-07
Unable to copy layered worldstate on 22.10.2 - Snapshots=False


non-fungible-nelson:
### Description
Besu user experiencing the following bug:

022-12-07 08:38:20.491-07:00 | vert.x-worker-thread-0 | INFO  | MainnetBlockValidator | Optional[Unable to copy Layered Worldstate for 0x3c2a249675dca59a4477c9a3ffdcc51e203617266f1a0f6434fa1a1dc5fec3f0]. Block 16130345 (0xa8b549ea57d83df60aec71c6fb40ab8391b0f640d7a2d2a013fce3a296800ad0), caused by org.hyperledger.besu.plugin.services.exception.StorageException: Unable to copy Layered Worldstate for 0x3c2a249675dca59a4477c9a3ffdcc51e203617266f1a0f6434fa1a1dc5fec3f0...

Directly after a restart and their node is stuck. Restart does not fix. Initially the user experienced the error below, then upon restart, the error above. 

2022-12-07 08:09:44.904-07:00 | vert.x-worker-thread-0 | WARN  | EngineNewPayload | Invalid new payload: number: 16130345, hash: 0xa8b549ea57d83df60aec71c6fb40ab8391b0f640d7a2d2a013fce3a296800ad0, parentHash: 0x3c2a249675dca59a4477c9a3ffdcc51e203617266f1a0f6434fa1a1dc5fec3f0, latestValidHash: 0x3c2a249675dca59a4477c9a3ffdcc51e203617266f1a0f6434fa1a1dc5fec3f0, status: INVALID, validationError: Unable to process block because parent world state 0x3b34cb5ab01b1b590a73ac002acbf60790360952ed1679bd6d9f9b608856b29a is not available

User does not have snapshots enabled. User is on 22.10.2. 

[Discord context](https://discord.com/channels/697535391594446898/888098626264502302/1049027346226757682)

### Acceptance Criteria
* Bonsai is resilient to this concurrency bug 

### Steps to Reproduce (Bug)
1. Run Besu
2. Halt or restart
3. Experience error (specific DB state)

**Expected behavior:** [What you expect to happen]
Continues normally. 

**Actual behavior:** [What actually happens]
Bonsai resumes as normal

**Frequency:** [What percentage of the time does it occur?]
Infrequent

### Versions (Add all that apply)
* Software version: 22.10.2

### Additional Information (Add any of the following or anything else that may be relevant)
* Besu setup info - snapshots false
[...]

PR #4862, opened 2022-12-23
Bugfix potential chain head and worldstate inconsistency

garyschulte:
## PR description
Ensure we do not leave the bonsai database in an inconsistent state where chain head and worldstate are at different block hashes

## Fixed Issue(s)

relates to #4784

## Documentation

## Changelog

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
@@ -527,7 +527,10 @@ public class MergeCoordinator implements MergeMiningCoordinator, BadChainListene
         // move chain head forward:
         return blockchain.forwardToBlock(newHead);
       } else {
-        // failed to move head forward:
+        debugLambda(
+            LOG,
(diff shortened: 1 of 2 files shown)


shemnon (approved):
LGTM

AegeanDad:
When do you expect to release this patch to stakers? As it stands, I am down.

garyschulte:
> When do you expect to release this patch to stakers? As it stands, I am down.

This PR does not purport to fix the underlying cause of #4784, but just mitigates the database damage from a failed fork-choice update if/when it occurs.  For now the only recourse if you encounter the defect in 4784 is to resync besu.

AegeanDad:
Given that this data issue pertains to a specific block (0x3bc82334b7902026f2e9ed64122ccf224899f014dc25823a8b30568e588f20cd, block 16243159) in my case, would you be able to recommend a quicker data fix that targets the specific block in the database?

garyschulte:
> Given that this data issue pertains to a specific block (0x3bc82334b7902026f2e9ed64122ccf224899f014dc25823a8b30568e588f20cd, block 16243159) in my case, would you be able to recommend a quicker data fix that targets the specific block in the database?

sent a private DM on discord.  It is non-trivial to recover a database from this state.

PR #4867, opened 2022-12-29
Attempt to fix CPU spikes issue

### LINKED ISSUE #4838: CPU usage spikes on a synced node
### Description
Some users reported random CPU spikes on a Besu synced nodes.
Below some of the graphs shared by users :
image

image

Two different users shared graphs that show a link between these CPU peaks and the increase of number of pooled transactions.
Analysing a thread dump during this CPU peak usage, I noticed that one of the threads is running this part of the code, and holding a lock making 3 other threads waiting for this lock. This could mean that this thread was running for some time.

```
java.lang.Thread.State: RUNNABLE
at java.util.stream.AbstractPipeline.evaluate(java.base@17.0.5/AbstractPipeline.java:234)
at java.util.stream.IntPipeline.reduce(java.base@17.0.5/IntPipeline.java:515)
at java.util.stream.IntPipeline.sum(java.base@17.0.5/IntPipeline.java:473)
at org.hyperledger.besu.ethereum.eth.manager.EthPeer.outstandingRequests(EthPeer.java:524)
at org.hyperledger.besu.ethereum.eth.manager.EthPeer.hasAvailableRequestCapacity(EthPeer.java:532)
```
This part of the code calculates (sum) the number of outstanding requests and it takes into account those from pooled transactions.

image

### Steps to Reproduce (Bug)
* It is hard to reproduce, it could be part of the resolution of this issue.

* Besu has a stable CPU consumption on synced node.

* Besu has some random spikes in CPU usage

This issue is not frequent but some users seem to reproduce more than others, may be related to the use case. 

* Software version: 22.10.3
* Java version: OpenJDK 64-Bit Server VM (17.0.5+8-Ubuntu-2ubuntu122.04 mixed mode, sharing)
* OS Name & Version: ubuntu 22.04

ahamlat:

## PR description
This PR is an attempt to fix CPU spikes issue (#4838).
A CPU profiling on a synced node showed that most of the time is spent on calculating peer outstanding requests. This is triggered by attemptExecution called for each pending request by reattemptPendingPeerRequests and triggered by either EthPeers.dispatchMessage or EthPeers.disconnect.

EthPeers.dispatchMessage
&nbsp;&nbsp;&nbsp;&nbsp;reattemptPendingPeerRequests
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;attemptExecution
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;hasAvailableRequestCapacity
       
EthPeers.registerDisconnect
&nbsp;&nbsp;&nbsp;&nbsp;reattemptPendingPeerRequests
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;attemptExecution
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;hasAvailableRequestCapacity

So each time, a message is dispatched or a peer is disconnected, Besu will iterate over all the pending requests and try to find an available peer for each single request. 
 
image

By looking to the size of pending requests, we can see that this can be huge (more than 200k), and there is a clear correlation with the CPU spikes. My understanding here is that Besu node is spammed sometimes by huge amount of requests which results of a lot of pending requests.

image

The idea behind this PR is to avoid unnecessary computation and iterate over available peers instead of iterating over pending requests.

## Fixed Issue(s)

Should fix #4838

## Documentation

## Changelog


macfarla (commented):
makes sense to me. Have you tried it out - does it improve the CPU spike issue?

ahamlat:
I have been running two nodes with this PR and didn't see a CPU from the installation of the last commit, but it is very hard to see if CPU spikes would happen without this PR.
Analysing the heap dump on one node that had this issue, we found that the number of pending request is very close to the number of GetPooledTransactionsFromPeerTask instances, and it by checking GC roots, it seems to be related to the way BufferedGetPooledTransactionsFromPeerFetcher works currently.
@matkt Made a modification in the last commit to reuse the same instance of BufferedGetPooledTransactionsFromPeerFetcher, I would like to have @fab-10 opinion on that

ahamlat:
The class histogram of one of the nodes facing this issue with more than 42k pending requests
image

fab-10 (approved):
LGTM, just evaluate to change `PendingPeerRequest::getPeerToUse` to not filter by minimum block height

ahamlat:
> LGTM, just evaluate to change `PendingPeerRequest::getPeerToUse` to not filter by minimum block height

I suggest to do this in another PR as it in not related to CPU spikes.

fab-10:
> > LGTM, just evaluate to change `PendingPeerRequest::getPeerToUse` to not filter by minimum block height
> 
> I suggest to do this in another PR as it in not related to CPU spikes.

Fine for me

PR #4868, opened 2022-12-30
Enable EOF in Shanghai

shemnon:
## PR description

Enable EOF into shanghai by wiring it into protocol specs.

## Documentation

## Changelog


evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
```diff
@@ -468,4 +469,15 @@ public static void registerShandongOperations(
     registry.put(new RetFOperation(gasCalculator));
     registry.put(new JumpFOperation(gasCalculator));
   }
+
+  public static EVM shandong(final BigInteger chainId, final EvmConfiguration evmConfiguration) {
```
diega, line 473: Do we need to keep Shandong's EVM and spec?

diega (approved):

evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
shemnon: I have another PR where I nuke it from orbit.  I wanted to keep them separate in case we need to roll this one out.

Issue #4852, opened 2022-12-21
Besu integration tests not running


jframe:
The Besu gradle integrationTest task does not seem to be running the integration tests.

In a recent circle CI job there are no test results https://app.circleci.com/pipelines/github/hyperledger/besu/19158/workflows/e2071db0-6c17-47c1-b54e-569752237cd7/jobs/116688.

Locally it appears to be just be compiling and running any of the integration tests.

siladu:
Looks like this was caused by migrating to junit 5 https://github.com/hyperledger/besu/commit/eb5e604bcddd2504489447ab5549208870f052b9

diega:
@siladu was that commit causing some sort of intermittency in `acceptanceTests` across CI runs? Because I'm pretty sure those were running when I made that change

siladu:
@diega nothing to do with acceptanceTests AFAIK, I'm just trying to fix the `integrationTest` task. It runs 'successfully' but doesn't actually find/run any tests. Problem is not confined to the CI, it happens when you run gradle locally as well or even when you run an individual test in IDE.

PR #4870, opened 2022-12-31
Revert junit5 integrationTests

siladu:
## Fixed Issue(s)

Quick fix for https://github.com/hyperledger/besu/issues/4852 (revert original commit: https://github.com/hyperledger/besu/commit/eb5e604bcddd2504489447ab5549208870f052b9)

## Documentation

## Changelog

### DIFF
--- a/consensus/ibft/build.gradle
+++ b/consensus/ibft/build.gradle
@@ -64,12 +64,9 @@ dependencies {
   integrationTestImplementation project(':metrics:core')
   integrationTestImplementation project(':testutil')
 
+  integrationTestImplementation 'junit:junit'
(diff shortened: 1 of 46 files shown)


shemnon (changes requested):
Consider #4871, what was missing was a gradle config, we don't need to roll it all back.

siladu:
> Consider https://github.com/hyperledger/besu/pull/4871, what was missing was a gradle config, we don't need to roll it all back.

Totally agree. Would rather fix the integration test config. I wasn't sure what the fix was, so I thought this might be a quick revert so we can get the integration tests working again ASAP, but they are already quite broken :)

PR #4874, opened 2023-01-03
Cache empty slots

ahamlat:

## PR description
This will add a cache for empty slots. This would help to improve SLOAD and SSTORE performance. It would improve especially SSTORE during cost gas and refund amount calculation because some original and current values used during this calculation are empty. 

### CPU Profiling before and after this PR
**Before this PR**
Block Processing CPU profiling
image

SSTORE CPU Profiling
image

**After this PR**
Block Processing CPU profiling
Screenshot 2023-01-06 at 10 20 15

SSTORE CPU Profiling
image

### The results
As we can see in the screenshot below, there is an improvement on block processing time on 50th percentile (mean) and 95th percentile.

image

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -714,7 +725,8 @@ public boolean isDirty() {
         && deletedAccounts.isEmpty()
         && storageToUpdate.isEmpty()
         && storageToClear.isEmpty()
-        && codeToUpdate.isEmpty());
+        && codeToUpdate.isEmpty()
+        && emptySlot.isEmpty());
```
garyschulte, line 729: we don't need this since we do not persist anything about emptySlot collection.  Even if it isn't empty, it does not make this updater dirty.  If it isn't empty, but nothing else has changed we might unnecessarily trigger a roothash calc.

garyschulte (changes requested):
Looks good, minor feedback about isDirty()

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
ahamlat: Good point

matkt (approved):
LGTM

PR #4875, opened 2023-01-03
Worldstate-only resync behavior

garyschulte:
## PR description
Implements a  debug rpc endpoint and a behavior to resync just the worldstate.  

## Documentation

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
@@ -34,6 +34,7 @@ public enum RpcMethod {
   CLIQUE_GET_SIGNER_METRICS("clique_getSignerMetrics"),
   DEBUG_ACCOUNT_AT("debug_accountAt"),
   DEBUG_METRICS("debug_metrics"),
+  DEBUG_RESYNC_WORLDSTATE("debug_resyncWorldState"),
   DEBUG_SET_HEAD("debug_setHead"),
   DEBUG_REPLAY_BLOCK("debug_replayBlock"),
   DEBUG_STORAGE_RANGE_AT("debug_storageRangeAt"),
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockManager.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockManager.java
@@ -49,6 +49,12 @@ public class BadBlockManager {
     }
   }
 
+  public void reset() {
+    this.badBlocks.invalidateAll();
+    this.badHeaders.invalidateAll();
+    this.latestValidHashes.invalidateAll();
+  }
+
   /**
    * Return all invalid blocks
    *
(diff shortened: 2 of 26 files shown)


garyschulte:
@shemnon as far as UX for a resync feature, 'touching' a file in the data dir is rather simple to implement and accessible.  A debug rpc endpoint may or may not be a good idea for this kind of functionality.  What do you think about file vs rpc, file and rpc, or some other command pattern to trigger this behavior?

shemnon:
I'm not a devops maven, I'd think we should get the UX feedback from someone who manages fleets of nodes.  Touching a file feels hacky to me but I've discovered lots of DevOps "best practices" feel hacky to me.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
```diff
@@ -34,6 +34,7 @@ public enum RpcMethod {
   CLIQUE_GET_SIGNER_METRICS("clique_getSignerMetrics"),
   DEBUG_ACCOUNT_AT("debug_accountAt"),
   DEBUG_METRICS("debug_metrics"),
+  DEBUG_RESYNC_WORLDSTATE("debug_resyncWorldState"),
```
matkt, line 37: I will prefer to have this king of RPC in the admin namespace

matkt (changes requested):

matkt:
> I'm not a devops maven, I'd think we should get the UX feedback from someone who manages fleets of nodes. Touching a file feels hacky to me but I've discovered lots of DevOps "best practices" feel hacky to me.

I'm no devops but I personally prefer an RPC call or a flag to a file that feels DIY.
I would just say that the RPC call must be admin because it seems too critical to me to simply be in debug

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockManager.java
```diff
@@ -49,6 +49,12 @@ public void addBadBlock(final Block badBlock, final Optional<Throwable> cause) {
     }
   }
 
+  public void reset() {
+    this.badBlocks.invalidateAll();
+    this.badHeaders.invalidateAll();
+    this.latestValidHashes.invalidateAll();
+  }
+
```
fab-10, line 57: Where is this called?

fab-10 (approved):
LGTM, just a method that seems unused

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockManager.java
garyschulte: https://github.com/hyperledger/besu/pull/4875/files#diff-cc598aa8b256f675dfcbd331db0955fd273916856708992c92364aa373405f12R49

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java
garyschulte: as discussed, we will keep this in debug namespace

PR #4883, opened 2023-01-05
Calculate block value

### LINKED ISSUE #4861: Calculate blockValue for engine_getPayloadV2
blockValue stub of 0 was created in https://github.com/hyperledger/besu/pull/4833

Now need to calculate the blockValue, which is still being discussed on https://github.com/ethereum/execution-apis/issues/307

wcgcyx:
Signed-off-by: Zhenyang Shi <wcgcyx@gmail.com>

## PR description
This PR adds the block value calculation to engine_getPayloadV2. 

Block value is calculated as the sum of the block's priority fee as per the current [spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-2). 

However, the calculated block value isn't strictly equal to the balance diff of the fee recipient as it doesn't capture balance gain through transaction execution (balance transfer, contract logic...). For example, tx1 may have less priority fee than tx2 but tx1 involves a direct transfer to fee recipient thus overall makes tx1 better than tx2.

Using the sum of the block's priority fee is the most straight-forward way to calculate the block value as a starting point as it only involves very minimum change. A future improvement is to have block processor report the balance diff of the fee recipient either during block creation or block validation and use that as the actual block value because re-evaluating the block value in get payload call will consume a lot of time. 

## Fixed Issue(s)

fixes #4861 

## Documentation

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
```diff
@@ -95,19 +99,29 @@ public EngineGetPayloadResultV1 payloadTransactionCompleteV1(final Block block)
...
+        blockWithReceipts.getHeader(), txs, Quantity.create(blockValue));
   }
 
-  private long calculateBlockValue(final List<String> ignored) {
-    return 0L;
+  private long calculateBlockValue(final BlockWithReceipts blockWithReceipts) {
+    final Block block = blockWithReceipts.getBlock();
+    final List<Transaction> txs = block.getBody().getTransactions();
+    final List<TransactionReceipt> receipts = blockWithReceipts.getReceipts();
+    Wei totalFee = Wei.of(0);
```
siladu, line 119:
```suggestion
    Wei totalFee = Wei.ZERO;
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
```diff
@@ -95,19 +99,29 @@ public EngineGetPayloadResultV1 payloadTransactionCompleteV1(final Block block)
...
+    for (int i = 0; i < txs.size(); i++) {
```
siladu, line 120: Using a for each would be more idiomatic.

```
for (Transaction tx : txs) {
...
}
```

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadV2Test.java
```diff
@@ -63,6 +80,70 @@ public void shouldReturnBlockForKnownPayloadId() {
     verify(engineCallListener, times(1)).executionEngineCalled();
   }
 
+  @Test
+  public void shouldReturnBlockWithCorrectValue() {
```
siladu, line 84: Would prefer if this was tested at the BlockResultFactory level, is there a reason you've done it here?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
```diff
@@ -95,19 +99,29 @@ public EngineGetPayloadResultV1 payloadTransactionCompleteV1(final Block block)
...
-    final long blockValue = calculateBlockValue(txs);
-    return new EngineGetPayloadResultV2(block.getHeader(), txs, Quantity.create(blockValue));
+    final long blockValue = calculateBlockValue(blockWithReceipts);
+    return new EngineGetPayloadResultV2(
+        blockWithReceipts.getHeader(), txs, Quantity.create(blockValue));
   }
 
-  private long calculateBlockValue(final List<String> ignored) {
-    return 0L;
+  private long calculateBlockValue(final BlockWithReceipts blockWithReceipts) {
```
siladu, line 115: I'm undecided, but wondering if we should do the calculation closer to the block/transaction processing, rather than at the API level. Would that be a trivial change? 
Perhaps it will only ever be used in this API though 🤷

siladu (commented):
> For example, tx1 may have less priority fee than tx2 but tx1 involves a direct transfer to fee recipient thus overall makes tx1 better than tx2.

On the face of it this sounds like an edge case but do you know if there are common scenarios where the transaction sender pays the block builder? I'm wondering if some MEV setups might use this.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
siladu: We discussed moving this method to org.hyperledger.besu.ethereum.core.BlockValueCalculator

wcgcyx: Updated

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadV2Test.java
wcgcyx: Updated

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
wcgcyx: Need to traverse through transactions as well as the receipts, could potentially use two iterators and a while loop to do this, what do you think?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
wcgcyx: Done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
siladu: Didn't spot the receipts access as well, what you have is good then.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/BlockValueCalculatorTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.hyperledger.besu.plugin.data.TransactionType;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes32;
+import org.junit.Test;
+
+public class BlockValueCalculatorTest {
```
siladu, line 31: Maybe a test to ensure 0 txs -> 0 blockValue?

wcgcyx: Updated

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinatorTest.java
```diff
@@ -494,13 +512,16 @@ public void shouldNotStartAnotherBlockCreationJobIfCalledAgainWithTheSamePayload
...
+    verify(mergeContext, times(retries.intValue()))
+        .putPayloadById(eq(payloadId1), blockWithReceipts.capture());
 
-    assertThat(block.getAllValues().size()).isEqualTo(retries.intValue());
+    assertThat(blockWithReceipts.getAllValues().size()).isEqualTo(retries.intValue());
     for (int i = 0; i < retries.intValue(); i++) {
-      assertThat(block.getAllValues().get(i).getBody().getTransactions()).hasSize(i);
+      assertThat(blockWithReceipts.getAllValues().get(i).getBlock().getBody().getTransactions())
+          .hasSize(i);
     }
```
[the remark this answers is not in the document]
wcgcyx: Updated

siladu (approved):
LGTM 👍

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.hyperledger.besu.ethereum.core;
+
+import org.hyperledger.besu.datatypes.Wei;
+
+import java.util.List;
+
+public class BlockValueCalculator {
+
+  public static long calculateBlockValue(final BlockWithReceipts blockWithReceipts) {
```
jframe, line 23: nit: would prefer this wasn't static

jframe (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2: nit: Contributors to Hyperledger Besu

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/BlockValueCalculatorTest.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2: nit: Contributors to Hyperledger Besu

macfarla (commented):
looks sensible to me. couple of nits re copyright statement

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
wcgcyx: Thanks for picking that up. Updated

PR #4888, opened 2023-01-06
Peering - disconnect worst peer

macfarla:
Disconnect "worst" peer in two cases
* AbstractRetryingSwitchingPeerTask - when retrying task with multiple peers and all peers have been tried and found wanting
* FastSyncTargetManager - if best peer is behind the pivot block, disconnect worst peer 
Also add some extra context info to "Best peer" log message

## Documentation

## Changelog


jflo (approved):
wish this had some test coverage, but I can't really think of how to do that...

PR #4876, opened 2023-01-03
Create skeleton structure for Engine API V2 methods

siladu:
## PR description

First step to implementing https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md

Create AbstractEngineNewPayload and AbstractEngineForkchoiceUpdated, extending with V1 and V2 versions.
(AbstractEngineGetPayload and V2 was already introduced in https://github.com/hyperledger/besu/commit/a483f79cc8b7741cb9fb0145d92307b6d5cfdc62)

This will be the basis for parallelising getting a productionised version of the Withdrawals draft onto main. 

## Fixed Issue(s)

Part of https://github.com/hyperledger/besu/issues/4776

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

Issue #4776, opened 2022-12-05
Productionise Withdrawals Engine API Changes


siladu:
Full implementation of https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md

The WIP is rather large and hard to review: https://github.com/hyperledger/besu/pull/4552

Ideally this should be split into PRs to get this into main behind the shanghaiTimestamp fork.

~Therefore this story is blocked on the [shanghaiTimestamp PR](https://github.com/hyperledger/besu/pull/4743).~

Note, this is strongly related to https://github.com/hyperledger/besu/issues/4777 and it might make sense to do unit tests and refactor per PR split.

One suggested split is:
* ~EVM logic  - [WithdrawalsProcessor](https://github.com/hyperledger/besu/pull/4552/files#diff-9bdad13766a20d4c64d5a6d62f3a3a5d26256edf3f8ab23b51e205251f270a5d)~ Covered by #4878 
* Supporting new withdrawals and withdrawalsRoot fields in domain layer, most wiring.
* API [EngineForkchoiceUpdatedV2](https://github.com/hyperledger/besu/pull/4552/files#diff-cd2a726c6924cedeb54ae9eabf94d49e09bb5494f83a319dce2297bd772c69e1), [EngineNewPayloadV2](https://github.com/hyperledger/besu/pull/4552/files#diff-3a23cfe6545f9f51df31b8c64033da79d5712b0902603305a78d16043cfa4b54), [EngineGetPayloadV2](https://github.com/hyperledger/besu/pull/4552/files#diff-c6b22c4aaf5b5035bc8acb1410ae852c2a429a6ca488f0f205c388f7c6dfe888), etc
  * There is also a small piece of validation to add for engine_newPayloadV2 (null/non-null withdrawals when pre/post-shanghai)

siladu:
Covered by other stories

siladu:
Closed the wrong issue

PR #4890, opened 2023-01-08
Implement engine_forkchoiceUpdated Withdrawals validation

siladu:
## PR description

Spec: https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md

In order to avoid maintaining two versions of forkchoiceUpdated as much as possible, the validation works for both V1 and V2.

The JSON deserialization allows for withdrawals to be omitted which supports V1 requests, however if shanghai is enabled then we will expect withdrawals in a V1 request. I think it would be a bug for a CL to send V1 with null withdrawals if the payload attributes had a post-shanghai timestamp.

## Fixed Issue(s)

Part of #4776 

Follows on from https://github.com/hyperledger/besu/pull/4876


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
```diff
@@ -171,6 +192,26 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+    // TODO Withdrawals move this into mergeCoordinator.updateForkChoice with the other
+    // payloadAttributes validation?
+    final List<Withdrawal> withdrawals =
+        Optional.ofNullable(payloadAttributes.getWithdrawals())
+            .map(ws -> ws.stream().map(WithdrawalParameter::toWithdrawal).collect(toList()))
+            .orElse(null);
+    return timestampSchedule
+        .getByTimestamp(payloadAttributes.getTimestamp())
+        .map(
+            protocolSpec -> protocolSpec.getWithdrawalsValidator().validateWithdrawals(withdrawals))
```
jframe, line 205: Is it necessary to have this on the protocol schedule? Do V2 APIs need to work before the shanghai?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
```diff
@@ -171,6 +192,26 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
             Optional.empty()));
   }
 
+  private boolean validateWithdrawals(final EnginePayloadAttributesParameter payloadAttributes) {
+    // TODO Withdrawals move this into mergeCoordinator.updateForkChoice with the other
+    // payloadAttributes validation?
+    final List<Withdrawal> withdrawals =
```
jframe, line 198: Why do we need to convert this to a domain object if we are just checking that there are withdrawals present? Is this for later when withdrawals will be used?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/JsonRpcMethodsFactory.java
```diff
@@ -104,7 +105,12 @@ public Map<String, JsonRpcMethod> methods(
               new EeaJsonRpcMethods(
                   blockchainQueries, protocolSchedule, transactionPool, privacyParameters),
               new ExecutionEngineJsonRpcMethods(
-                  miningCoordinator, protocolContext, ethPeers, consensusEngineServer),
+                  miningCoordinator,
+                  MergeProtocolSchedule.createTimestamp(
```
jframe, line 109: Why does another protocol schedule need to be created? Can the one from the controller be used?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
```diff
@@ -171,6 +192,26 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
             Optional.empty()));
   }
 
+  private boolean validateWithdrawals(final EnginePayloadAttributesParameter payloadAttributes) {
+    // TODO Withdrawals move this into mergeCoordinator.updateForkChoice with the other
```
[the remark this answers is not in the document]
siladu: I might spike it out next to see if it makes sense. I don't like that the validation is spread across this class and MergeCoordinator. 

Ideally, I'd like all api validation to be in this class but I think it's a significant refactor and I don't think before shanghai is the right time to do it. Second best might be to perform all validation in one place, hence the TODO.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdatedV2.java
```diff
@@ -17,17 +17,21 @@
 import org.hyperledger.besu.consensus.merge.blockcreation.MergeMiningCoordinator;
 import org.hyperledger.besu.ethereum.ProtocolContext;
 import org.hyperledger.besu.ethereum.api.jsonrpc.RpcMethod;
+import org.hyperledger.besu.ethereum.mainnet.TimestampSchedule;
 
 import io.vertx.core.Vertx;
 
+// TODO Withdrawals use composition instead? Want to make it more obvious that there is no
```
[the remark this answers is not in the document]
siladu: Yes, I actually added that for this PR. Forgot to call it out myself, but thought it might be a discussion point for the review.

Currently thinking we inheritance may lead people in the future to add V2+ specific changes in the subclass which I'm currently trying to avoid in order to keep the V1/V2 differences as small as possible. Instead, favour making the code backwards compatible.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
siladu: Yes, some CLs e.g. lighthouse always send V2

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
siladu: The problem here is that I want shanghaiDefinition to wire in a validator class, but it can't be in the api module due to a circular dependency with core.

So WithdrawalsValidator can't accept WithdrawalParameter

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
jframe: Could the V2 APIs use the allowed withdrawals validator and the V1 APIs use the not allowed validator instead of driving this from the protocol schedule? Doesn't look like ProtocolSpec needs a getWithdrawalsValidator method.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdatedV2.java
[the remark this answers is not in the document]
siladu: If we decide on composition over inheritance then I'd like to do that in a separate PR as it will affect three engine API methods, assuming we want them to follow a consistent approach.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
siladu: I believe the correct interpretation of the spec is the validation is driven by Shanghai, not by the method name. 

Currently, driving it from V2 would break lighthouse because we'd invalidate pre-shanghai blocks, see https://discord.com/channels/595666850260713488/892088344438255616/1055785287885783040. 

Once shanghai has passed, we'd also be relying on the CLs to send V1 for pre-shanghai blocks (e.g. in syncing scenarios) which implies clients need to keep all versions around forever, hence the aim of backwards compatibility.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdatedV2.java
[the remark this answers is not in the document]
jframe: Think it's ok using inheritance for now

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
jframe: Thinking about this more I think this is ok. The Withdrawals will be needed to prepare a payload anyway in the FCU. And the MainnetBlockValidator will also need the WithdrawalsValidator.

siladu:
Moving to draft while I work on spiking this out...https://github.com/hyperledger/besu/pull/4890#discussion_r1064321476

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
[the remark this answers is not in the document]
siladu: @jframe Turns out it wasn't so bad to move all the payload validation from MergeCoordinator into FCU

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdatedTest.java
```diff
@@ -442,7 +451,196 @@ public void shouldIgnoreUpdateToOldHeadAndNotPreparePayload() {
     verify(engineCallListener, times(1)).executionEngineCalled();
   }
 
-  private EngineUpdateForkchoiceResult assertSuccessWithPayloadForForkchoiceResult(
+  @Test
+  public void shouldReturnInvalidIfPayloadTimestampIsInvalid() {
```
jframe, line 455: In this case, how is the timestamp invalid? Perhaps more descriptive test name to make it clearer why it's invalid. Is this testing the case where ExecutionPayloadV1 is used when timestamp < shanghai?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpecBuilder.java
```diff
@@ -316,6 +323,11 @@ public ProtocolSpec build(final HeaderBasedProtocolSchedule protocolSchedule) {
               privacyParameters.getPrivateStateGenesisAllocator());
     }
 
+    WithdrawalsValidator withdrawalsValidator = new WithdrawalsValidator.ProhibitedWithdrawals();
```
jframe, line 326: NIcer to default to prohibited withdrawals instead of using a null check and doing it here

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpecBuilder.java
```diff
@@ -443,6 +456,10 @@ BlockValidator apply(
         Optional<GoQuorumPrivacyParameters> goQuorumPrivacyParameters);
   }
 
+  public interface WithdrawalsValidatorBuilder {
```
jframe, line 459: Why do need to use a builder? Could just set the value directly on the ProtocolSpecBuilder

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/WithdrawalsValidator.java
```diff
@@ -0,0 +1,57 @@
...
+
+  class AllowedWithdrawals implements WithdrawalsValidator {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AllowedWithdrawals.class);
+
+    @Override
+    public boolean validateWithdrawals(final List<Withdrawal> withdrawals) {
+      final boolean isValid = withdrawals != null;
+      if (!isValid) {
+        LOG.warn("withdrawals should not be null when Withdrawals are activated");
```
jframe, line 52: replace should with must

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/WithdrawalsValidator.java
```diff
@@ -0,0 +1,57 @@
...
+  class ProhibitedWithdrawals implements WithdrawalsValidator {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ProhibitedWithdrawals.class);
+
+    @Override
+    public boolean validateWithdrawals(final List<Withdrawal> withdrawals) {
+      final boolean isValid = withdrawals == null;
+      if (!isValid) {
+        LOG.warn(
+            "withdrawals should be null when Withdrawals are prohibited but were: {}", withdrawals);
```
jframe, line 38: replace should with must

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdatedTest.java
siladu: yeh makes sense, it's the existing fcu validation, not new for withdrawals, i.e. whether or not the timestamp is in the future w.r.t head

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpecBuilder.java
siladu: I guess Jiri was following suit with how it works in the rest of the class

siladu: I'll try direct, don't think it matters since there's no arguments to WithdrawalsValidator

jframe (approved):

PR #4897, opened 2023-01-09
use bonsai updater cache to avoid calling database several times for the same account

matkt:
Co-authored-by: Ameziane H <ameziane.hamlat@consensys.net>

## PR description
This PR will use WorldStateBonsaiUpdater accounts' cache each time we try to get an account from storage. Currently, we use this cache only in some code paths.

The profiling before and after this PR shows a real difference in CPU processing, as we use more a Java cache and get less data from RocksDB.

**Before this PR**

image

**After this PR**
image

If we zoom in one of the calls, we can see that with this PR, in some cases (and it is the case with this 5 minutes CPU profiling), there is no access to the database to get the account

**Before this PR**
image

**After this PR**
image

**Block processing time**
We notice around 7 % improvement with this PR 

image

## Documentation

## Changelog


ahamlat (approved):
LGTM.
I added performances metrics that confirms the performance impovement

PR #4898, opened 2023-01-09
EOF Optimization - don't split code sections

shemnon:
## PR description

Optimize the EOF Functions implementation by not splitting code sections into separate Bytes object.  Instead the EOF evaluation occurs on a single container space.  When a function is called the PC (relative to container start, not section start) is moved, and no section re-loading needs to occur.

EOF lacks PC introspection and absolute jumps, so where "PC=0" starts is not observable to the code.

## Documentation

## Changelog


shemnon:
(a) memory usage.  We don't have to slice out code sections and make copies of the sections.  Also a bit of speed on the parse
(b) complexity.  You will notice that CALLF and RETF no longer need special handling in the EVM loop.  While they can be moved there if it is optimal they no longer need to reset the code in that loop as we are all off of one code loop.

shemnon:
(c) there is also a runtime performance improvement, as we don't have to swap code sections in and out at function calls, but instead it's just a jump with more rules.

PR #4839, opened 2022-12-19
Deprecation notice for GoQuorum-compatible privacy and IBFT 1.0

macfarla:
Signed-off-by: Sally MacFarlane <macfarla.github@gmail.com>

Besu currently has 3 privacy modes. The privacy mode that provides compatibility with GoQuorum privacy is intertwined with mainnet code at the transaction and block processing level. We also have not seen demand for this privacy mode within the Besu community.
We plan to deprecate the GoQuorum-compatible privacy feature, along with IBFT 1.0, in the coming quarterly release 23.1, and then remove it in the following quarterly release 23.4.
Removing the code associated with this feature will reduce the complexity of the transaction and block processing code, reducing friction for mainnet-focused development.

This will not affect the remaining 2 privacy modes (EEA privacy and flexible privacy groups).
This will not affect IBFT 2.0 or QBFT.

Reach out to us on Discord with any questions.

Also remove quorum ATs nightly task - this nightly task is currently consistently failing. Since we are deprecating the GoQuorum-compatible privacy feature in 23.1 so we won't be investing time fixing these tests.

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #4905, opened 2023-01-11
Ibft1 deprecation warning

### LINKED ISSUE #4851: Deprecate IBFT1
Deprecate IBFT1 by adding warning messages that the IBFT1 is being deprecated.

A changelog entry has already been added https://github.com/hyperledger/besu/pull/4839/files

gfukushima:
## PR description
Add IBFT1 deprecation warning.

## Fixed Issue(s)

fix #4851
related to #4839

## Documentation

## Changelog

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
+++ b/besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
@@ -217,9 +217,10 @@ public class BesuController implements java.io.Closeable {
         throw new IllegalArgumentException("Unknown consensus mechanism defined");
       }
 
-      // TODO this warning can be removed once ibft1 has been deprecated
+      // TODO remove this warning once ibft1 has been deprecated
       if (builder instanceof IbftLegacyBesuControllerBuilder) {
-        LOG.warn("IBFT1 is being deprecated. Consider using IBFT2 or QBFT.");
+        LOG.warn(
+            "IBFT1 is being deprecated and will be removed in a future release. Consider using QBFT instead of IBFT1 or using GoQuorum instead of Besu if you need to use IBFT1");
       }
 
       // wrap with TransitionBesuControllerBuilder if we have a terminal total difficulty:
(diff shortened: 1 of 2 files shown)


besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -241,6 +247,9 @@ private BesuControllerBuilder createConsensusScheduleBesuControllerBuilder(
       if (configOptions.isIbft2()) {
         originalControllerBuilder = new IbftBesuControllerBuilder();
       } else if (configOptions.isIbftLegacy()) {
+        // TODO remove this warning once ibf1 has been deprecated
```
siladu, line 250: nit: typo s/ibf1/ibft1...was thinking if anyone tries to search for "ibft1" they wouldn't find this, but probably not worth the extra CI credits to change since it's close to the IbftLegacy.. class anyway!

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -241,6 +247,9 @@ private BesuControllerBuilder createConsensusScheduleBesuControllerBuilder(
...
+        LOG.warn(
+            "IBFT1 is being deprecated and will be removed in a future release. Consider using QBFT instead of IBFT1 or using GoQuorum instead of Besu if you need to use IBFT1");
```
jframe, line 252: Putting this in the IbftLegacyBesuControllerBuilder constructor would be a better place. Would also remove the duplication.

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
siladu: If you put in the constructor as Jason mentioned, then probably don't need the comment since the whole class will presumably be removed

jframe (approved):

PR #4912, opened 2023-01-12
Increase cache performance post snapshot fix

matkt:
## PR description

We detected a regression in the cache after the fix of the snapshot segfault issue. This is due to the fact that the cache tried to use a worldstate storage that was unavailable too soon (not subscribe because shouldClose is true). The cache did not have time to preload enough node and so the calculate root hash is less efficient

the issue is 

- In the validateAndProcessBlock we are doing a copy of the worldState
```java
      try (final var worldState =
        context
            .getWorldStateArchive()
            .getMutable(parentHeader.getStateRoot(), parentHeader.getBlockHash(), shouldPersist)
            .map(
                ws -> {
                  if (!ws.isPersistable()) {
                    return ws.copy();
                  }
                  return ws;
                })
            .orElse(null)) {
```

In our case it was BonsaiLayeredWorldstate

```java
public MutableWorldState copy() {
    // return an in-memory worldstate that is based on a persisted snapshot for this blockhash.
    try (SnapshotMutableWorldState snapshot =
        archive
            .getMutableSnapshot(this.blockHash())
            .map(SnapshotMutableWorldState.class::cast)
            .orElseThrow(
                () ->
                    new StorageException(
                        "Unable to copy Layered Worldstate for " + blockHash().toHexString()))) {
      return new BonsaiInMemoryWorldState(archive, snapshot.getWorldStateStorage());
    } catch (Exception ex) {
      throw new RuntimeException(ex);
    }
  }
```

In this code we can see that we are getting a snapshot and passing the worldstate storage of this snapshot to the BonsaiInMemoryWorldState class. It is a try -with-resources statement so the snapshot will be close at the end of the copy

The close of the snapshot will call the close method of the worldstate storage but this one will not be close because BonsaiInMemoryWorldState subscribed to it . The close snapshot will not close the storage but will set the shouldClose to true. And this is the problem.

when the cache try to subscribe to worldstate storage an exception is raised because this shouldClose is true even if the worldstate is still open 

```java
public synchronized long subscribe(final BonsaiStorageSubscriber sub) {
    if (shouldClose.get()) {
      throw new RuntimeException("Storage is marked to close or has already closed");
    }
    return super.subscribe(sub);
  }
```

The idea is to use isClosed instead of shouldClose during the subscribe 

Here come CPU profiling flame graphs that show the difference before and after the regression 
**Before**

- Calculate Root Hash (Pink part)

image

- Cache Preload

image

**After**

- Calculate Root Hash (Pink part)

image

- Cache Preload
image

Zoom on the code snippet that triggers the runtime exception
image

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiSnapshotWorldStateKeyValueStorage.java
@@ -102,7 +102,7 @@ public class BonsaiSnapshotWorldStateKeyValueStorage extends BonsaiWorldStateKey
 
   @Override
   public synchronized long subscribe(final BonsaiStorageSubscriber sub) {
-    if (shouldClose.get()) {
+    if (isClosed.get()) {
       throw new RuntimeException("Storage is marked to close or has already closed");
     }
     return super.subscribe(sub);


ahamlat (approved):
LGTM, I added some flame graphs to illustrate this use case.

garyschulte (approved):
🚢   great find

PR #4917, opened 2023-01-13
Withdrawals processor

jframe:
## PR description
Process withdrawals when creating a block and when processing a block. Withdrawals processing is only enabled for the shanghai fork and beyond.

## Documentation

## Changelog


macfarla (commented):
looks sensible to me

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreatorTest.java
```diff
@@ -0,0 +1,159 @@
...
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.units.bigints.UInt64;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class AbstractBlockCreatorTest {
```
siladu, line 61: Nice! Surprised this didn't have a test before!

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreatorTest.java
```diff
@@ -52,9 +52,9 @@
 
 import com.google.common.collect.Lists;
 import org.apache.tuweni.bytes.Bytes;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
-public class PoWBlockCreatorTest {
+public class PoWBlockCreatorTest extends AbstractBlockCreatorTest {
```
siladu, line 57: If AbstractBlockCreatorTest isn't abstract, do we need to extend it here?

siladu, line 57: Do you think we need a MergeBlockCreatorTest as well?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/WithdrawalsProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+  class DefaultWithdrawalsProcessor implements WithdrawalsProcessor {
+
+    @Override
+    public void processWithdrawals(
+        final List<Withdrawal> withdrawals, final WorldUpdater worldUpdater) {
+      for (final Withdrawal withdrawal : withdrawals) {
+        final EvmAccount account = worldUpdater.getOrCreate(withdrawal.getAddress());
+        account.getMutable().incrementBalance(withdrawal.getAmount());
+      }
+      worldUpdater.commit();
```
siladu, line 43: Presumably it's possible for this to throw exceptions? Wonder what happens when it does.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/WithdrawalsProcessorTest.java
```diff
@@ -0,0 +1,142 @@
...
+                Wei.of(200)));
+    defaultWithdrawalsProcessor.processWithdrawals(withdrawals, updater);
+
+    assertThat(worldState.get(Address.fromHexString("0x1")).getBalance()).isEqualTo(Wei.of(101));
+    assertThat(worldState.get(Address.fromHexString("0x2")).getBalance()).isEqualTo(Wei.of(202));
+    assertThat(worldState.get(Address.fromHexString("0x3")).getBalance()).isEqualTo(Wei.of(3));
+  }
+
+  @Test
+  void defaultProcessor_shouldProcessWithdrawalsUpdatingEmptyAccountsBalance() {
```
siladu, line 109: good test, this was a bug that caught us out before!

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreatorTest.java
jframe: Hmm.. good point mostly wanted the name to match the test class as I put common logic in AbstractBlockCreator. Rename it to BaseBlockCreatorTest maybe?

jframe: We should have one, the only thing different about the MergeBlockCreator is the block header it creates, but it would be good to have a test for that. Do it this PR or another?

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinatorTest.java
```diff
@@ -213,7 +213,8 @@ public void coinbaseShouldMatchSuggestedFeeRecipient() {
             genesisState.getBlock().getHeader(),
             System.currentTimeMillis() / 1000,
             Bytes32.ZERO,
-            suggestedFeeRecipient);
+            suggestedFeeRecipient,
+            Optional.empty());
```
siladu, line 217: nit: could extract a variable to make this clearer it's emptyWithdrawals

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/WithdrawalsProcessor.java
jframe: It could throw a StorageException, this would be caught by the MainnetBlockValidator and logged. Do you think it would be worth adding some specific log errors for withdrawals?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -170,6 +174,13 @@ protected BlockCreationResult createBlock(
       final ProtocolSpec newProtocolSpec =
           protocolSchedule.getByBlockHeader(processableBlockHeader);
 
+      final WithdrawalsProcessor withdrawalsProcessor = newProtocolSpec.getWithdrawalsProcessor();
+      maybeWithdrawals.ifPresent(
+          withdrawals ->
+              withdrawalsProcessor.processWithdrawals(withdrawals, disposableWorldState.updater()));
+
+      throwIfStopped();
```
siladu, line 182: Interesting, I see this has moved to after withdrawals (compared to the prototype) - can you explain the significance please?

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreatorTest.java
siladu: Just thinking that PoW will rerun AbstractBlockCreatorTest's tests as well as the parent I think - is that the intention?

Potentially could make AbstractBlockCreatorTest abstract so they run for subclasses but not the parent

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -134,6 +139,11 @@ public BlockProcessingResult processBlock(
       receipts.add(transactionReceipt);
     }
 
+    final WithdrawalsProcessor withdrawalsProcessor =
+        protocolSchedule.getByBlockHeader(blockHeader).getWithdrawalsProcessor();
+    maybeWithdrawals.ifPresent(
+        withdrawals -> withdrawalsProcessor.processWithdrawals(withdrawals, worldState.updater()));
```
siladu, line 145: Just want to confirm my understanding: if we were pre-shanghai, but had withdrawals, we'd run the NoOpWithdrawalsProcessor and not error?

I think the api validation would prevent this, but wondering if we maybe log a warning in the NoOp just in case?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PrivacyBlockProcessor.java
```diff
@@ -100,7 +102,13 @@ public BlockProcessingResult processBlock(
 
     final BlockProcessingResult result =
         blockProcessor.processBlock(
-            blockchain, worldState, blockHeader, transactions, ommers, metadataUpdater);
+            blockchain,
+            worldState,
+            blockHeader,
+            transactions,
+            ommers,
+            withdrawals,
```
siladu, line 110: To play 'mainnet devil's advocate', what's the reasoning for changing PrivacyBlockProcessor as well?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/WithdrawalsProcessor.java
siladu: As long as we end up with an invalid block and handle that the right way, not sure if there's value in it being withdrawals specific

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
jframe: There are multiple throwIfStopped calls. No significance in removing it,  think it makes sense to have it after as well, added that back in.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinatorTest.java
jframe: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
jframe: That's right, we would run the NoOpWithdrawalsProcessor.

Changing code around this area to return an Optional<WithdrawalProcessor> so there won't be a NoOpWithdrawalsProcessor at all. So will make that change first.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PrivacyBlockProcessor.java
jframe: That it wouldn't compile without this change as I updated the full arg processBlock method to include Withdrawals. Would prefer not to touch this since it's going to be removed soon. But it's also a trivial change.

jframe:
> Just thinking that PoW will rerun AbstractBlockCreatorTest's tests as well as the parent I think - is that the intention?
>
> Potentially could make AbstractBlockCreatorTest abstract so they run for subclasses but not the parent
Good point would be better to make this abstract

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
```diff
@@ -162,7 +162,8 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
                     newHead,
                     payloadAttributes.getTimestamp(),
                     payloadAttributes.getPrevRandao(),
-                    payloadAttributes.getSuggestedFeeRecipient()));
+                    payloadAttributes.getSuggestedFeeRecipient(),
+                    Optional.empty()));
```
siladu, line 166: nit: Could pass in withdrawals at this point, or could keep in separate PR?

After moving the validation into this class, there's no refactor necessary, should just be a case of mapping the payloadAttributes.withdrawals -> List<Withdraw>...e.g. WithdrawalParameter.toDomain()

siladu (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
jframe: done

jframe: done

PR #4918, opened 2023-01-13
improve preload cache

matkt:
## PR description
This PR aims to use Preload cache  in SLOAD operations.

There is no performance improvement on block processing time when we look to the last 3 hours.

image

But we can see that cache hit ratio is better but not that much to have a big impact on block processing time

image

## Documentation

## Changelog


ahamlat:
I did more tests and analysis, the results on the long run are even worse than before, I'm going to close this PR.

PR #4928, opened 2023-01-14
fixes test that won't pass if local hostname unresolvable via reverse…

jflo:
… dns


siladu (approved):
LGTM

macfarla (approved):

PR #4931, opened 2023-01-15
4844 new tx type

jflo:
Adds the new enumeration for the 4844 blob data transaction type, and fixes many tests which surprisingly broke due to that.

### DIFF
--- a/ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
+++ b/ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
@@ -327,7 +327,7 @@ public class BlockDataGenerator {
       defaultTxs.add(transaction(options.getTransactionTypes()));
     }
 
-    return new BlockBody(options.getTransactions(defaultTxs), ommers, Optional.empty());
+    return new BlockBody(options.getTransactions(defaultTxs), ommers);
   }
 
   private BlockHeader ommer() {
--- a/ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
+++ b/ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
@@ -337,7 +337,7 @@ public class BlockDataGenerator {
   private TransactionType transactionType() {
     // TODO: when TransactionType.EIP4844 is fully supported, revert this.
     return transactionType(
-        TransactionType.FRONTIER, TransactionType.ACCESS_LIST, TransactionType.EIP1559 );
+        TransactionType.FRONTIER, TransactionType.ACCESS_LIST, TransactionType.EIP1559);
   }
 
   private TransactionType transactionType(final TransactionType... transactionTypes) {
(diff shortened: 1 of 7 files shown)


ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
```diff
@@ -627,7 +637,9 @@ public static class BlockOptions {
     private Optional<Long> timestamp = Optional.empty();
     private boolean hasOmmers = true;
     private boolean hasTransactions = true;
-    private TransactionType[] transactionTypes = TransactionType.values();
+    private TransactionType[] transactionTypes = {
```
jflo, line 640: Fun fact, you can actually break a few tests just by changing the enum ordering here.

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
```diff
@@ -335,7 +335,9 @@ private BlockHeader ommer() {
   }
 
   private TransactionType transactionType() {
-    return transactionType(TransactionType.values());
+    // TODO: when TransactionType.EIP4844 is fully supported, revert this.
```
fab-10, line 338: Idk how this work, but just wondering if is it possible to change the return value based on the fact that Cancun is enabled or not

gradle.properties
```diff
@@ -1,4 +1,4 @@
 version=23.1.0-SNAPSHOT
 
 org.gradle.welcome=never
-org.gradle.jvmargs=-Xmx1g
+org.gradle.jvmargs=-Xmx2g
```
fab-10, line 4: is this intended to be in this PR?

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/TransactionType.java
```diff
@@ -21,10 +21,11 @@
 public enum TransactionType {
   FRONTIER(0xf8 /* this is serialized as 0x0 in TransactionCompleteResult */),
   ACCESS_LIST(0x01),
-  EIP1559(0x02);
+  EIP1559(0x02),
+  EIP4844(0x03);
 
   private static final Set<TransactionType> ACCESS_LIST_SUPPORTED_TRANSACTION_TYPES =
-      Set.of(ACCESS_LIST, EIP1559);
+      Set.of(ACCESS_LIST, EIP1559, EIP4844);
```
fab-10, line 28: Please sync with @gezero that in his PR also set this with a different name https://github.com/hyperledger/besu/pull/4919/files#diff-f4664e850b91579cf65137f12d71d2269b6dc4d3f1c59ee5f65ae0782947809a

fab-10 (commented):
LGTM, just a couple of comments, and evaluate a way to avoid forgetting to reset the code after Cancun

gradle.properties
jflo: removed, is handled in https://github.com/hyperledger/besu/pull/4932/files

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
jflo: it's probably possible but since this is in test-support I don't think it is worth it.

macfarla (approved):
LGTM

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/BlockDataGenerator.java
fab-10: fair, the important thing is that we do not forget about it :)

PR #4933, opened 2023-01-15
Datahash opcode

jflo:
Don't merge till https://github.com/hyperledger/besu/pull/4931/files has been merged.


evm/src/main/java/org/hyperledger/besu/evm/operation/DataHashOperation.java
```diff
@@ -0,0 +1,57 @@
...
+    int blobIndex = frame.popStackItem().toInt();
+    final Optional<List<Hash>> maybeHashes = frame.getVersionedHashes();
+    if (frame.getVersionedHashes().isPresent()) {
+
+      List<Hash> versionedHashes = maybeHashes.get();
+      if (blobIndex < versionedHashes.size()) {
+        Hash requested = versionedHashes.get(blobIndex);
+        frame.pushStackItem(requested);
+      } else {
+        return new OperationResult(3, ExceptionalHaltReason.INVALID_OPERATION);
```
fab-10, line 45: should this be `0`?

from the spec:

> returns tx.message.blob_versioned_hashes[index] if index < len(tx.message.blob_versioned_hashes), and otherwise zero

fab-10 (commented):
looks good, please check my comments/questions

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -753,6 +768,7 @@ private static Bytes32 computeSenderRecoveryHash(
         preimage = frontierPreimage(nonce, gasPrice, gasLimit, to, value, payload, chainId);
         break;
       case EIP1559:
+      case EIP4844:
```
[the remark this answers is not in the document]
jflo: I'm also unclear, but i think this will make more sense when we merge in the SSZ serialization

evm/src/main/java/org/hyperledger/besu/evm/operation/DataHashOperation.java
shemnon: So if it's a cancun block and no versioned hashes are passed in then there will be a versioned hashes field with zero length, correct?  May be worth a comment to clarify.

shemnon (commented):
One thing is unclear to me, if the chain is post-cancun, and they pass in a legacy or 1559 transaction should the datahash opcode be available? i.e. should it act as though there were zero hashes and always return zero or should it throw an invalid opcode if a 1559 tx calls a contract with the opcode?

Currently it looks like we are coded for the latter.  This doesn't seem like the valid set of opcodes should be a function of the transaction format.

jflo:
> One thing is unclear to me, if the chain is post-cancun, and they pass in a legacy or 1559 transaction should the datahash opcode be available? i.e. should it act as though there were zero hashes and always return zero or should it throw an invalid opcode if a 1559 tx calls a contract with the opcode?
> 
> Currently it looks like we are coded for the latter. This doesn't seem like the valid set of opcodes should be a function of the transaction format.

Yes! I struggled with this as well while implementing it, and in hindsight i think I like returning zero instead.  That way at least contract authors using DATAHASH can choose to accept or not pre-4844 transactions without a halt.  I will change this and bring it up for discussion and see if test cases to cover this are planned.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 ### Breaking Changes
 
 ### Additions and Improvements
+- Support for new DATAHASH opcode as part of EIP-4844
```
fab-10, line 17: missing the link

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -753,6 +768,7 @@ private static Bytes32 computeSenderRecoveryHash(
         preimage = frontierPreimage(nonce, gasPrice, gasLimit, to, value, payload, chainId);
         break;
       case EIP1559:
+      case BLOB:
```
fab-10, line 771: Here we need to implement a dedicated `blobPreimage` method that takes in consideration the new field, as I am doing for the `max_fee_per_data_gas` field.
If you want I can it in my PR for both fields.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -378,6 +379,10 @@ public TransactionProcessingResult processTransaction(
                 .address(contractAddress)
                 .contract(contractAddress)
                 .inputData(Bytes.EMPTY)
+                .versionedHashes(
+                    transaction.getType() == TransactionType.BLOB
```
fab-10, line 383: is the check redundant? should the field be empty by default if not set?

gradle.properties
```diff
@@ -1,4 +1,4 @@
 version=23.1.0-SNAPSHOT
 
 org.gradle.welcome=never
-org.gradle.jvmargs=-Xmx1g
+org.gradle.jvmargs=-Xmx3g
```
fab-10, line 4: this is in another PR right?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
jflo: perfect, thanks.

gradle.properties
jflo: it is in another pr at 2g, which still wasn't enough for me. i'll revert it here, and we can increment it in another pr if 2g is really insufficient for most devs.

fab-10 (approved):

evm/src/main/java/org/hyperledger/besu/evm/operation/DataHashOperation.java
```diff
@@ -0,0 +1,58 @@
...
+      if (blobIndex < versionedHashes.size()) {
+        Hash requested = versionedHashes.get(blobIndex);
+        frame.pushStackItem(requested);
+      } else {
+        frame.pushStackItem(Bytes.EMPTY);
+      }
+    } else {
+      frame.pushStackItem(Bytes.EMPTY);
+    }
+    return new OperationResult(3, ExceptionalHaltReason.NONE);
```
shemnon, line 51:
```suggestion
    return new OperationResult(3, null);
```
All the exceptional halt checks need null for normal termination.  `NONE` will halt the frame.

evm/src/test/java/org/hyperledger/besu/evm/operations/DataHashOperationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    Hash version0Hash = Hash.fromHexStringLenient("0xcafebabeb0b0facedeadbeef");
+    List<Hash> versionedHashes = Arrays.asList(version0Hash);
+    DataHashOperation getHash = new DataHashOperation(new LondonGasCalculator());
+    MessageFrame frame = mock(MessageFrame.class);
+    when(frame.popStackItem()).thenReturn(Bytes.of(0));
+    when(frame.getVersionedHashes()).thenReturn(Optional.of(versionedHashes));
+    EVM fakeEVM = mock(EVM.class);
+    Operation.OperationResult r = getHash.execute(frame, fakeEVM);
+    assertThat(r.getGasCost()).isEqualTo(3);
+    assertThat(r.getHaltReason()).isEqualTo(ExceptionalHaltReason.NONE);
```
shemnon, line 53:
```suggestion
    assertThat(r.getHaltReason()).isEqualTo(null);
```

evm/src/test/java/org/hyperledger/besu/evm/operations/DataHashOperationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    verify(frame).pushStackItem(version0Hash);
+  }
+
+  @Test
+  public void failsOnBloblessTx() {
```
shemnon, line 58:
```suggestion
  public void pushesZeroOnBloblessTx() {
```

evm/src/test/java/org/hyperledger/besu/evm/operations/DataHashOperationTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    when(frame.popStackItem()).thenReturn(Bytes.of(0));
+    when(frame.getVersionedHashes()).thenReturn(Optional.of(new ArrayList<>()));
+    Operation.OperationResult failed2 = getHash.execute(frame, fakeEVM);
+    assertThat(failed2.getGasCost()).isEqualTo(3);
+    verify(frame, times(2)).pushStackItem(Bytes.EMPTY);
+  }
+
+  @Test
+  public void failsOnVersionIndexOutOFBounds() {
```
shemnon, line 78: `fail` in tests indicates to me an exception raised, rather than a zero push.

```
  public void pushesZeroOnVersionInexOutOfBounds() {
```

shemnon (changes requested):
Return need to be null exceptional halt reason, and some minor test changes (perhaps we should add the null exceptional halt to the last two unit tests for the data hash)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
diega, line 771: Doesn't this new type requiere to calculate the pre-image in a different way than 1559?

diega (commented):
Doesn't `o.h.b.e.m.MainnetProtocolSpecs#cancunDefinition` need to set the _transactionValidatorBuilder_ to add support for the new `BLOB` tx?

For the future, maybe we could came up with a different design for the Transaction/MessageFrame that doesn't carry so many `Optional`s all around.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
jflo: correct it does, that will be provided by another pr. defining at least the case was necessary for unit tests to pass

shemnon (approved):

diega (approved):

PR #4944, opened 2023-01-17
Add optional list of withdrawals to block body

jframe:

## PR description
Add optional list of withdrawals to block body. Also update the equals, hashcode methods. This does not wire withdrawals into the block body it's just to reduce the noise as many uses especially tests will use Optional.empty for withdrawals. Changes to wire in the list of withdrawals will be done in a subsequent PR.

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
@@ -60,9 +60,7 @@ public class BlockBody implements org.hyperledger.besu.plugin.data.BlockBody {
     return ommers;
   }
 
-  /**
-   * @return The list of transactions of the block.
-   */
+  /** Returns the list of withdrawals of the block. */
   public Optional<List<Withdrawal>> getWithdrawals() {
     return withdrawals;
   }
@@ -95,7 +93,7 @@ public class BlockBody implements org.hyperledger.besu.plugin.data.BlockBody {
   }
 
   @Override
-  public boolean equals(Object o) {
+  public boolean equals(final Object o) {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     BlockBody blockBody = (BlockBody) o;
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
@@ -20,18 +20,24 @@ import org.hyperledger.besu.ethereum.rlp.RLPOutput;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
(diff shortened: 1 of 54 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
```diff
@@ -75,34 +90,36 @@ public static BlockBody readFrom(
...
-    sb.append("transactions=").append(transactions).append(", ");
-    sb.append("ommers=").append(ommers);
-    return sb.append("}").toString();
+    return "BlockBody{"
+        + "transactions="
+        + transactions
+        + ", ommers="
+        + ommers
+        + ", withdrawals="
+        + withdrawals
```
siladu, line 122: Will this just return ", withdrawals=Optional.empty" ?

siladu (approved):

pinges (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
jframe: Yeah that's what it will return

garyschulte:
IMO overloading the BlockBody constructor would have been a better choice rather than changing all of the usages of blockbody to take an optional list...

siladu:
> IMO overloading the BlockBody constructor would have been a better choice rather than changing all of the usages of blockbody to take an optional list...

@garyschulte what's your thinking around this?
I think a benefit of doing it this way is you're potentially avoiding subtle bugs by surfacing missing wiring in the compiler.

garyschulte:
> > IMO overloading the BlockBody constructor would have been a better choice rather than changing all of the usages of blockbody to take an optional list...
> 
> @garyschulte what's your thinking around this?
> I think a benefit of doing it this way is you're potentially avoiding subtle bugs by surfacing missing wiring in the compiler.

Withdrawals are an Ethereum mainnet only consideration.  Also now we have a bunch of Optional.empty() params everywhere, whereas an overloaded constructor could just do that in one place.

NBD, it just struck me as odd to require an optional parameter

jframe:
> IMO overloading the BlockBody constructor would have been a better choice rather than changing all of the usages of blockbody to take an optional list...
> 
> @garyschulte what's your thinking around this?
> I think a benefit of doing it this way is you're potentially avoiding subtle bugs by surfacing missing wiring in the compiler.
> 
> Withdrawals are an Ethereum mainnet only consideration. Also now we have a bunch of Optional.empty() params everywhere, whereas an overloaded constructor could just do that in one place.
> 
> NBD, it just struck me as odd to require an optional parameter

You have a good point about this being mainnet specific and this forces everyone to consider whether to use withdrawals. And it would make tests cleaner. The production code though would need to determine what constructor to call so the main code may necessarily not be any cleaner. Will have a quick go-see what the code would be like using an overloaded constructor approach.

PR #4947, opened 2023-01-17
Fix: Hive engine api FCU regression

gfukushima:
Swap lines order to ensure we commit FCU even if payload timestamp is before head timestamp

Signed-off-by: Gabriel Fukushima <gabrielfukushima@gmail.com>

## PR description
This PR fixes the regression introduced recently.

## Fixed Issue(s)

Makes Besu compliant with ForkChoiceUpdate_V1 - item 7 of the Engine API specs for Paris. (https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md)

## Documentation

## Changelog


siladu (approved):
LGTM

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
```diff
@@ -143,6 +143,10 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
                         ws ->
                             ws.stream().map(WithdrawalParameter::toWithdrawal).collect(toList())));
 
+    ForkchoiceResult result =
```
jframe, line 146: Can we add this case to unit test as well

PR #4955, opened 2023-01-18
Change BlockBody to use an overloaded constructor with withdrawals

jframe:
## PR description
Change BlockBody to use an overloaded constructor with withdrawals instead having a mandatory `Optional<List<Withdrawals>>` as withdrawals are only required for ethereum mainnet.

## Documentation

## Changelog

### DIFF
--- a/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
+++ b/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
@@ -214,9 +214,7 @@ public abstract class AbstractBlockCreator implements AsyncBlockCreator {
       final BlockHeader blockHeader = createFinalBlockHeader(sealableBlockHeader);
 
       final Block block =
-          new Block(
-              blockHeader,
-              new BlockBody(transactionResults.getTransactions(), ommers, Optional.empty()));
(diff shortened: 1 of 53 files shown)


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -214,9 +214,7 @@ protected BlockCreationResult createBlock(
       final BlockHeader blockHeader = createFinalBlockHeader(sealableBlockHeader);
 
       final Block block =
-          new Block(
-              blockHeader,
-              new BlockBody(transactionResults.getTransactions(), ommers, Optional.empty()));
+          new Block(blockHeader, new BlockBody(transactionResults.getTransactions(), ommers));
```
siladu, line 217: Isn't this the one we need withdrawals in - or will that be wired up later?

siladu (approved):
LGTM

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
jframe: Yeah, we definitely will need to do that but wanted to just keep this PR about the existing state which doesn't pass in the withdrawals. I will add in the next PR that includes all the wiring changes.

PR #4958, opened 2023-01-18
New excess_data_gas block header field

fab-10:
## Fixed Issue(s)

Add `excess_data_gas` field to block header as for EIP-4844 specification

## Documentation

## Changelog


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -155,6 +155,7 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
             blockParam.getBaseFeePerGas(),
             blockParam.getPrevRandao(),
             0,
+            blockParam.getExcessDataGas().orElse(null),
```
jframe, line 158: Could including this cause an issue for EngineNewPayloadV2 creating blocks that have the excess data gas field in them?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -225,6 +230,9 @@ public void writeTo(final RLPOutput out) {
     if (baseFee != null) {
       out.writeUInt256Scalar(baseFee);
     }
+    if (excessDataGas != null) {
```
jframe, line 233: I think the next expected field is the withdrawalsRoot followed by the excessDataGas. So could merge this in after my PR so this is the last field or just skip the next field.

siladu (approved):
nit: maybe you intend to do this in another PR but excessDataGas would need adding to the ExecutionPayloadResult as well, which engine_getPayloadV3 returns

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
siladu: I ran a test using V2 endpoints on main vs this branch to confirm no V2 regression is introduced by when excessDataGas is null.
If excessDataGas is included in newPayload however, it would indeed change the block hash. I don't see this as an immediate issue but once shanghai is live, a CL's buggy 4844 code could cause besu to miscalculate the root hash.
The same problem exists at the moment for V1 -> V2 migration w.r.t the new withdrawals fields. 

I can't tell from the spec what the EL behaviour should be. Maybe we should throw an error when V2 is called with excessDataGas, need to confirm the spec or understand what other ELs are doing

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
fab-10: yes, merged your PR and moved after `withdrawalsRoot`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -305,6 +315,7 @@ public String toString() {
     sb.append("extraData=").append(extraData).append(", ");
     sb.append("baseFee=").append(baseFee).append(", ");
     sb.append("mixHashOrPrevRandao=").append(mixHashOrPrevRandao).append(", ");
+    sb.append("excessDataGas=").append(excessDataGas).append(", ");
```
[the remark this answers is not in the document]
fab-10: yes and done

siladu:
> @siladu @jframe I have removed the changes to Engine API from this PR, since they were not complete, and @jflo is going to work on that part. Please review for a second time if everything it ok.

Still looks good to me. I would like @jframe to approve this as well

jframe (approved):
LGTM

jframe:
> > @siladu @jframe I have removed the changes to Engine API from this PR, since they were not complete, and @jflo is going to work on that part. Please review for a second time if everything it ok.
> 
> Still looks good to me. I would like @jframe to approve this as well

Looks fine to me too.

PR #4964, opened 2023-01-19
Fix transaction pool issue

matkt:

## PR description

There has been a regression on Besu regarding the transaction pool. Indeed it seems that the transaction pool validation part does not make a copy of the worldstate during the GetMutable and this means that we can have a version which can change and which is not safe. A copy mechanism has been set up which allows to use the snapshot and to protect against worldstate modifications during validation. it was used elsewhere except in transaction pool validation and block replay.

This caused us to invalidate too many transactions

## Documentation

## Changelog

### DIFF
--- a/besu/src/test/java/org/hyperledger/besu/services/BesuEventsImplTest.java
+++ b/besu/src/test/java/org/hyperledger/besu/services/BesuEventsImplTest.java
@@ -132,6 +132,7 @@ public class BesuEventsImplTest {
         .thenReturn(ValidationResult.valid());
     when(mockTransactionValidator.validateForSender(any(), any(), any()))
         .thenReturn(ValidationResult.valid());
+    when(mockWorldState.copy()).thenReturn(mockWorldState);
     when(mockWorldStateArchive.getMutable(any(), any(), anyBoolean()))
         .thenReturn(Optional.of(mockWorldState));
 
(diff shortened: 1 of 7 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchive.java
```diff
@@ -160,28 +160,37 @@ public Optional<MutableWorldState> getMutableSnapshot(final Hash blockHash) {
...
+                                  LOG.warn(
+                                      "Exceeded the limit of back layers that can be loaded ({})",
+                                      trieLogManager.getMaxLayersToLoad());
+                                  return false;
+                                }
+                                return true;
+                              })
+                          .flatMap(header -> snapshotOrLayeredWorldState(blockHash, header)));
     }
+    return mutableWorldstate.map(
```
garyschulte, line 187: 👍

matkt: I will do as before. I cannot see easy way to manage that . I wanted to create an OptionalMutableWorldstate with only a ifPresent that close the ressource automatically but seems to be a big and complex modification

I will also reduce the scope to the transaction pool and we will see another PR for the blockReplay part

garyschulte (approved):
:+1:

Issue #4808, opened 2022-12-12
Add withdrawals details to eth_getBlock methods


jframe:
Return withdrawals details from the eth_getBlock methods when withdrawals are enabled.

https://github.com/ethereum/execution-apis/pull/334

Should probably also add to debug_getBadBlock - are there other debug/trace methods that might need withdrawals as well?

**Acceptance Criteria**
* When withdrawals are enabled return the withdrawalsRoot and withdrawals in the returned block
* Do not include the withdrawalsRoot or withdrawals when withdrawals are not enabled

siladu:
Implemented a quick and dirty solution for the testnet https://github.com/hyperledger/besu/pull/4818/commits/161b70c74b06f94a4722b983aded45811364332a

siladu:
Consider reviewing and updating all of the APIs that return a Block e.g. debug_*

non-fungible-nelson:
Pending integration tests @siladu

PR #4969, opened 2023-01-20
Add withdrawals details to eth_getBlock methods

siladu:
## PR description

This is the critical piece that will allow Withdrawals interop to work.

Integration tests and any other RPCs that return a block will be added in a subsequent PR.

## Fixed Issue(s)

Critical piece of #4808


jframe (approved):
LGTM

Issue #4785, opened 2022-12-07
Catch-All Critical Transaction Exception / Unable to load trie node 


non-fungible-nelson:
### Description
Catch-All - Critical transaction exception / unable to load trie node

Reference - #4779, #4768, #4434 

Maybe related #4379 

Critical Exception processing transaction 
            - Error on new payload call with submitted transaction, snapshot concurrency 

Instability around trie node unable to load:
        - [Discord Context 1](https://discord.com/channels/697535391594446898/888098626264502302/1049531467983761449) 
        - [Discord Context 2](https://discord.com/channels/697535391594446898/888098626264502302/1048975986445791242)
        - [Discord Context 3](https://discord.com/channels/697535391594446898/888098626264502302/1049531467983761449)

### Acceptance Criteria
* Removal of concurrency 

### Steps to Reproduce (Bug)
1. Process certain transactions with an incomplete world state

**Expected behavior:** [What you expect to happen]
No error for valid blocks

**Actual behavior:** [What actually happens]
Infrequent errors when submitting user tx and just processing blocks

**Frequency:** [What percentage of the time does it occur?]
Infrequent 

### Versions (Add all that apply)
* Software version: 22.10.2

Issue #4768, opened 2022-12-02
Critical Exception Processing Transaction


MysticRyuujin:
### Description
My Besu node seems to have become corrupt, possibly after a reorg?

I think this is the second time in a couple weeks this happened. I just re-sync a few weeks ago because of a similar or if not the same bug (I didn't look into it that hard, I figured it was fluke)

```
2022-12-02 18:57:24.300+00:00 | vert.x-worker-thread-0 | INFO  | EngineNewPayload | Imported #16,098,404 / 156 tx / base fee 14.19 gwei / 19,826,119 (66.1%) gas / (0xabc875990fc719c8363fc652c68b84465514a4be0faf57b69e78da016bb1bc2d) in 2.467s. Peers: 17
2022-12-02 18:57:31.013+00:00 | vert.x-eventloop-thread-11 | ERROR | EthStatsService | Failed to reach the ethstats server Failed to resolve 'ethstats.net' after 2 queries 
2022-12-02 18:57:42.724+00:00 | vert.x-eventloop-thread-13 | ERROR | EthStatsService | Failed to reach the ethstats server Failed to resolve 'ethstats.net' after 2 queries 
2022-12-02 18:57:53.796+00:00 | vert.x-eventloop-thread-15 | ERROR | EthStatsService | Failed to reach the ethstats server Failed to resolve 'ethstats.net' after 2 queries 
2022-12-02 18:57:55.394+00:00 | vert.x-worker-thread-0 | WARN  | DefaultBlockchain | Chain Reorganization +9 new / -0 old
       Old - hash: 0xfd2febdf386ce960c55e6dfbbb36027da157e0b78bcbae677cd51c1b4e0cf5e6, height: 16098394
       New - hash: 0x6371917fba3322ba5341b58c20ea6b122fd6071eb9dbd57ad0d28ca5f44b626f, height: 16098403
  Ancestor - hash: 0xfd2febdf386ce960c55e6dfbbb36027da157e0b78bcbae677cd51c1b4e0cf5e6, height: 16098394
[...]

Issue #4379, opened 2022-09-10
Error: Unable to load trie node value


estensen:
Mainnet
v22.7.2
Checkpoint sync
Bonsai
Paired with Lighthouse

```
2022-09-10 07:28:42.488+02:00 | EthScheduler-Workers-0 | INFO  | PersistBlockTask | Imported #15,507,058 / 177 tx / 0 om / 12,357,960 (41.2%) gas / (0xeb57cfb3a467b8d2fac4755ea307e0170cb9b7958c8e2e3867325a39d464aea3) in 0.810s. Peers: 37
2022-09-10 07:28:44.492+02:00 | EthScheduler-Transactions-0 | ERROR | Besu | Uncaught exception in thread "EthScheduler-Transactions-0"
org.hyperledger.besu.ethereum.trie.MerkleTrieException: Unable to load trie node value for hash 0xfa7c13692550ff6de7f9df083cf91b0064a88728009472fd9d1e24f3bcaf7244 location 0x
	at org.hyperledger.besu.ethereum.trie.StoredNode.lambda$load$0(StoredNode.java:133)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.hyperledger.besu.ethereum.trie.StoredNode.load(StoredNode.java:131)
	at org.hyperledger.besu.ethereum.trie.StoredNode.accept(StoredNode.java:63)
	at org.hyperledger.besu.ethereum.trie.StoredMerklePatriciaTrie.get(StoredMerklePatriciaTrie.java:119)
	at org.hyperledger.besu.ethereum.bonsai.BonsaiWorldStateKeyValueStorage.getAccount(BonsaiWorldStateKeyValueStorage.java:110)
	at org.hyperledger.besu.ethereum.bonsai.BonsaiPersistedWorldState.get(BonsaiPersistedWorldState.java:340)
	at org.hyperledger.besu.ethereum.bonsai.BonsaiLayeredWorldState.get(BonsaiLayeredWorldState.java:232)
	at org.hyperledger.besu.ethereum.eth.transactions.TransactionPool.lambda$validateTransaction$2(TransactionPool.java:282)
	at java.base/java.util.Optional.map(Optional.java:260)
	at org.hyperledger.besu.ethereum.eth.transactions.TransactionPool.validateTransaction(TransactionPool.java:280)
	at org.hyperledger.besu.ethereum.eth.transactions.TransactionPool.validateRemoteTransaction(TransactionPool.java:213)
[...]

PR #4972, opened 2023-01-22
[POC] heal the worldstate

matkt:

## PR description

This PR add a heal  mechanism fo the worldtstate in case of inconsistency (unable to trie node). To fix this, we start a quick heal of the worldstate automatically and once the fix is done we restart the block import.

After the detection of an invalid path
- we delete this path to force the healing of this part.
- then we delete the trielogs. 
- we select a pivot block (before of after the heal)
- we move the blokchain to this pivot block (rewind or download the missing blocks) 
- then we launch a worldstate heal. 

This feature can also heal a node that has been inconsistent for a long time, but it will take longer because there will be more nodes to heal.  With this PR the healing will be done as soon as the problem is detected so there will not a lot to heal and it will be fast

## Performed tests
 - Trigger multiple inconsistencies to fix  (passed)
 - Fixed a node that has been inconsistent for a long time (passed)
 - Run a snapsync from scratch on goerli (passed)
 - Run a checkpoint sync from scratch on goelri (passed)
 - Run a checkpoint sync from scratch on main **(in progress)**
 - Run a validator teku+besu on goerli (passed)
 - Profile performance on this node to avoid perf regression (passed)

## Fixed Issue(s)

#4379 
#4785 
#4768

## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/SyncState.java
```diff
@@ -311,13 +316,29 @@ public Optional<Checkpoint> getCheckpoint() {
...
+
   public void markInitialSyncPhaseAsDone() {
     isInitialSyncPhaseDone = true;
     completionListenerSubscribers.forEach(InitialSyncCompletionListener::onInitialSyncCompleted);
   }
 
-  public boolean isInitialSyncPhaseDone() {
-    return isInitialSyncPhaseDone;
+  public boolean isResyncNeeded() {
+    return isResyncNeeded;
```
garyschulte, line 329: we never reset this value afaict.  once we need a resync, we always need a resync ;)

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/checkpointsync/CheckpointDownloaderFactory.java
```diff
@@ -81,17 +81,16 @@ public static Optional<FastSyncDownloader<?>> createCheckpointDownloader(
...
-        && fastSyncState.getPivotBlockHeader().isEmpty()
+      syncState
+          .getAccountToRepair()
+          .ifPresent(address -> snapContext.addInconsistentAccount(Hash.hash(address)));
+    } else if (fastSyncState.getPivotBlockHeader().isEmpty()
         && protocolContext.getBlockchain().getChainHeadBlockNumber()
             != BlockHeader.GENESIS_BLOCK_NUMBER) {
       LOG.info(
-          "Checkpoint sync was requested, but cannot be enabled because the local blockchain is not empty.");
+          "Snap sync was requested, but cannot be enabled because the local blockchain is not empty.");
```
garyschulte, line 93: nit: should be "checkpoint"

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldStateKeyValueStorage.java
```diff
@@ -35,15 +32,8 @@ public BonsaiInMemoryWorldStateKeyValueStorage(
...
-      final Optional<PeerTrieNodeFinder> fallbackNodeFinder) {
-    super(
-        accountStorage,
-        codeStorage,
-        storageStorage,
-        trieBranchStorage,
-        trieLogStorage,
-        fallbackNodeFinder);
+      final KeyValueStorage trieLogStorage) {
+    super(accountStorage, codeStorage, storageStorage, trieBranchStorage, trieLogStorage);
```
garyschulte, line 36: 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/FinalBlockConfirmation.java
```diff
@@ -44,7 +44,8 @@ static FinalBlockConfirmation genesisConfirmation(final MutableBlockchain blockc
   }
 
   static FinalBlockConfirmation ancestorConfirmation(final MutableBlockchain blockchain) {
-    return firstHeader -> blockchain.contains(firstHeader.getParentHash());
+    return firstHeader ->
+        blockchain.getChainHeadHeader().getHash().equals(firstHeader.getParentHash());
```
garyschulte, line 48: This might this break backward sync for a reorg.    e.g. blockhain.contains() will return true if we have block A' but it isn't the chain head.  This check requires it to be the chain head...

garyschulte (commented):
Solid, some of the backward sync changes are concerning.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/FinalBlockConfirmation.java
matkt: I'm using the blck number. Should be fine now but. if you can verify

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
```diff
@@ -84,13 +86,19 @@ protected Hash calculateRootHash(final BonsaiWorldStateUpdater worldStateUpdater
...
+        if (updatedAccount == null) {
+          final Hash addressHash = Hash.hash(accountKey);
+          accountTrie.remove(addressHash);
+        } else {
+          final Hash addressHash = updatedAccount.getAddressHash();
+          final Bytes accountValue = updatedAccount.serializeAccount();
+          accountTrie.put(addressHash, accountValue);
+        }
+      } catch (MerkleTrieException e) {
+        // need to throw to trigger the heal
```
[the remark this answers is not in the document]
siladu: ah, I see it would have throw before, so here you're catching, setting some values (to do what?) and then rethrowing?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
```diff
@@ -85,6 +85,7 @@ protected CompletableFuture<FastSyncState> start(final FastSyncState fastSyncSta
     if (worldStateStorage instanceof BonsaiWorldStateKeyValueStorage) {
       LOG.info("Clearing bonsai flat account db");
       worldStateStorage.clearFlatDatabase();
+      worldStateStorage.clearTrieLog();
```
siladu, line 88: why do we need to do this?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncTargetManager.java
```diff
@@ -167,8 +167,16 @@ private boolean peerHasDifferentPivotBlock(final List<BlockHeader> result) {
...
   public boolean shouldContinueDownloading() {
     final BlockHeader pivotBlockHeader = fastSyncState.getPivotBlockHeader().get();
-    return !protocolContext.getBlockchain().getChainHeadHash().equals(pivotBlockHeader.getHash())
-        || !worldStateStorage.isWorldStateAvailable(
-            pivotBlockHeader.getStateRoot(), pivotBlockHeader.getBlockHash());
+    boolean isValidChainHead =
+        protocolContext.getBlockchain().getChainHeadHash().equals(pivotBlockHeader.getHash());
+    if (!isValidChainHead) {
+      if (protocolContext.getBlockchain().contains(pivotBlockHeader.getHash())) {
+        protocolContext.getBlockchain().rewindToBlock(pivotBlockHeader.getHash());
```
siladu, line 174: Could the pivot block be > the max allowable bonsai trie layers away? If so, could this rewind take an unfeasible amount of time or maybe error? If so, then what state is the user left in and is it recoverable?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/FastDownloaderFactory.java
```diff
@@ -82,9 +81,8 @@ public static Optional<FastSyncDownloader<?>> create(
     final FastSyncState fastSyncState =
         fastSyncStateStorage.loadState(ScheduleBasedBlockHeaderFunctions.create(protocolSchedule));
 
-    if (isResync) {
-      worldStateStorage.clear();
```
siladu, line 86: why no need to clear the worldStateStorage now?

siladu (commented):
Should [ethereum/referencetests/src/reference-test/external-resources](https://github.com/hyperledger/besu/pull/4972/files#diff-f57e5bf63cf366775133a769c5d718b3c321fbd4b0ded98908d3370fb92403b5)
be changing as part of this PR?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/ProtocolContext.java
```diff
@@ -30,13 +31,16 @@ public class ProtocolContext {
   private final WorldStateArchive worldStateArchive;
   private final ConsensusContext consensusContext;
 
+  private Optional<Synchronizer> synchronizer;
```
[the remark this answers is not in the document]
matkt: yes, this is indeed a problem. I tried a lot to do otherwise but this was the cleanest I found with the current code structure

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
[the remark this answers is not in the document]
matkt: I add this value for the pruning and also to add some log indication . Address , slot is important to know what to prune before the heal and the trie don't have this information (only yhe wolrstate (update, inmemory etc)

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
matkt: we will not be able to go back before this pivot block. In addition, we could have bad trielog so it is preferable to heal them too. Finally keeping old trielog could make a false positive for the isWorldStateAvailable.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncTargetManager.java
matkt: this part rewind only the blockchain part not the worldstate part. 
If the pivot block is before the current head, we do a rewind otherwise we do nothing. In any case the worldstate will be healed with the new pivot block without doing a rollback or rollforward because we cannot do it if it is corrupted

it is important to choose a new pivot block and not to heal the current head. Because we are not sure that block will remain on the canonical chain

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/FastDownloaderFactory.java
matkt: For a resync there is already a clear in the Worldstate downloader part https://github.com/hyperledger/besu/pull/4972/files#diff-28d8c7a62dddded52099a727bf58aa864d58b055795b95ebfeee6ac8e2a8cea2R173
So it's clearly useless to do it again @garyschulte

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -167,7 +167,13 @@ public BlockProcessingResult validateAndProcessBlock(
         return new BlockProcessingResult(
             Optional.of(new BlockProcessingOutputs(worldState, receipts)));
       }
-    } catch (StorageException | MerkleTrieException ex) {
+    } catch (MerkleTrieException ex) {
+      context
+          .getSynchronizer()
+          .ifPresent(
```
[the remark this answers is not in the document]
matkt: yes good idea

matkt:
> Should [ethereum/referencetests/src/reference-test/external-resources](https://github.com/hyperledger/besu/pull/4972/files#diff-f57e5bf63cf366775133a769c5d718b3c321fbd4b0ded98908d3370fb92403b5) be changing as part of this PR?

yes good catch I will fix that

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/checkpointsync/CheckpointDownloaderFactory.java
matkt: done thanks

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/SyncState.java
matkt: fixed I switch this flag on `markInitialSyncPhaseAsDone`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/FinalBlockConfirmation.java
garyschulte: when I pull this branch here I see:
 `    return firstHeader -> blockchain.getChainHeadBlockNumber() + 1 >= firstHeader.getNumber();
`
I might be missing something, but I read the firstAncestor block as being the earliest block fetched by backward sync.  and finalBlockConfirmation.ancestorHeaderReached as the termination condition for fetching further back in BackwardSyncAlgorithm.pickNextStep.

If we are backward syncing because of a fork from 5 or 10 blocks behind head, I think this condition could terminate early.  it is almost logically identical to the conditional above it:
`    return firstHeader -> blockchain.getChainHeadBlockNumber() + 1 >= firstHeader.getNumber();
`

The first condition makes sense to me b/c we really just need to ensure this backward path resolves to something we have, rather than a particular height.

matkt: Ok I think I understand your remark. On the other hand, I have to keep this check because it allows me to have the backardsync works after the heal in case we répond. Because we will have the blocks but I still want to go back to import them again . So I think I should add the initial condition + this one. In the case of a REORG the first condition will make sure that it does not stop too soon

matkt:
Because of some release difficulties with this PR , I created another one @garyschulte @siladu ->  https://github.com/hyperledger/besu/pull/5059

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/FinalBlockConfirmation.java
matkt: I proposed another fix on this PR https://github.com/hyperledger/besu/pull/5059/files#diff-0361d2ff12cd43f507bf88a3cb4d781a537ca765329e0811f60a9507527cbe13R47
Thanks

PR #4909, opened 2023-01-12
Add engine get payload body

wcgcyx:
Signed-off-by: Zhenyang Shi <wcgcyx@gmail.com>

## PR description
This PR adds two new engine APIs: engine_getPayloadBodiesByHashV1 and engine_getPayloadBodiesByRangeV1.

## Fixed Issue(s)

fixes #4787 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #4980, opened 2023-01-23
Add getPayloadBodiesByRangeV1 and getPayloadBodiesByHash engine methods

### LINKED ISSUE #4787: Add getPayloadBodies and getPayloadBodiesByRangeV1 Methods
~https://github.com/ethereum/execution-apis/pull/218~
~https://github.com/ethereum/execution-apis/pull/146~

This is merged into main now: 
https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_getpayloadbodiesbyhashv1
https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_getpayloadbodiesbyrangev1

gfukushima:
## PR description
This PR adds the methods getPayloadBodiesByRangeV1 and getPayloadBodiesByHash to the engine API.

Implementation of https://github.com/ethereum/execution-apis/pull/352.

## Fixed Issue(s)

Fixes #4787

## Documentation

## Changelog


siladu (commented):
Still reviewing but will add the comments so far...

Gabriel-Trintinalia (commented):
A test to cover the condition: Client software MUST NOT return trailing null values if the request extends past the current latest known block would be good IMO

macfarla:
does this supersede #4909 ?

gfukushima:
Yes it does @macfarla

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1Test.java
```diff
@@ -146,9 +150,43 @@ public void shouldReturnNullForUnknownNumberAndPayloadForKnownNumber() {
...
+        new Withdrawal(UInt64.ONE, UInt64.ONE, Address.fromHexString("0x1"), GWei.ONE);
+    final BlockBody preShanghaiBlockBody =
+        new BlockBody(
+            List.of(
+                new TransactionTestFixture().createTransaction(sig.generateKeyPair()),
+                new TransactionTestFixture().createTransaction(sig.generateKeyPair()),
+                new TransactionTestFixture().createTransaction(sig.generateKeyPair())),
+            Collections.emptyList());
+
+    final BlockBody ShanghaiBlockBody =
```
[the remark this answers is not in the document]
gfukushima: Fixed, ty

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1Test.java
```diff
@@ -146,9 +150,43 @@ public void shouldReturnNullForUnknownNumberAndPayloadForKnownNumber() {
...
     assertThat(result.getPayloadBodies().size()).isEqualTo(3);
-    assertThat(result.getPayloadBodies().get(0).size()).isEqualTo(1);
+    assertThat(result.getPayloadBodies().get(0).getTransactions().size()).isEqualTo(1);
     assertThat(result.getPayloadBodies().get(1)).isNull();
-    assertThat(result.getPayloadBodies().get(2).size()).isEqualTo(3);
+    assertThat(result.getPayloadBodies().get(2).getTransactions().size()).isEqualTo(3);
+  }
+
+  @Test
+  public void shouldReturnWithdrawalNullWhenBlockIsPreShanghai() {
```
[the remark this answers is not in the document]
gfukushima: Separated them in 2

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
```diff
@@ -0,0 +1,84 @@
...
+  @Override
+  public JsonRpcResponse syncResponse(final JsonRpcRequestContext request) {
+    engineCallListener.executionEngineCalled();
+
+    final long startBlockNumber = request.getRequiredParameter(0, Long.class);
+    final long count = request.getRequiredParameter(1, Long.class);
+
+    traceLambda(
+        LOG,
+        "EngineGetPayloadBodiesByRangeV1 parameters: start block number {} count {}",
```
macfarla, line 62: would it make sense to use getName() here

gfukushima: Done!

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
```diff
@@ -64,10 +64,15 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext request) {
         () -> count);
 
     final Blockchain blockchain = protocolContext.getBlockchain();
+
+    final long latestKnownBlockNumber = blockchain.getChainHeadBlockNumber();
+    final long upperBound = startBlockNumber + count;
+    final long finalBlockNumber = latestKnownBlockNumber < upperBound ? latestKnownBlockNumber + 1 : upperBound;
```
siladu: suggestion: s/finalBlockNumber/endExclusiveBlockNumber (to match LongStream.range's naming)

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
```diff
@@ -0,0 +1,85 @@
...
+
+    return new JsonRpcSuccessResponse(
+        request.getRequest().getId(),
+        blockResultFactory.payloadBodiesCompleteV1(
+            LongStream.range(startBlockNumber, finalBlockNumber)
+                .mapToObj(
+                    blockNumber ->
+                        blockchain
+                            .getBlockHashByNumber(blockNumber)
+                            .flatMap(blockchain::getBlockBody))
```
siladu, line 82: @matkt can/should this be optimised?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlockResultFactory.java
```diff
@@ -115,6 +118,15 @@ public EngineGetPayloadResultV2 payloadTransactionCompleteV2(
         Quantity.create(blockValue));
   }
 
+  public EngineGetPayloadBodiesResultV1 payloadBodiesCompleteV1(
+      final List<Optional<BlockBody>> blockBodies) {
+    final List<PayloadBody> payloadBodies =
+        blockBodies.stream()
+            .map(maybeBody -> maybeBody.map(blockBody -> new PayloadBody(blockBody)).orElse(null))
```
siladu, line 125: Why are we returning null here?

I would expect that any Optional.empty() blockBodies would be skipped over by the stream, i.e. only the Optional.of() values will be collected into a list...is there a reason to keep the nulls around?

gfukushima: Spec says: "3. Client software MUST place null in the response array for unavailable blocks which numbers are lower than a number of the latest known block. " That's why we can't skip the empty blockBodies

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
gfukushima: Done

siladu:
> Client software MUST support count values of at least 32 blocks. The call MUST return -38004: Too large request error if the requested range is too large.

Looks like this isn't implemented?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
siladu: We spoke and we cannot save a db call since we don't key the database by blockNumber. Also, deserializing txs when we retrieve the blockBody is expensive, but required by the API in this case.

siladu:
> Client software MAY NOT respond to requests for finalized blocks by hash.

Not sure what this means...any idea?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
gfukushima: There was some discussion around introducing limits for these requests (github) when the methods were been scoped. Maybe we should consider that as well?

gfukushima:
I've just realized I was looking at the fork of the shanghai spec. @siladu the error code hasn't been implemented, I'll add that. Moving PR back to draft until then

siladu:
> > Client software MAY NOT respond to requests for finalized blocks by hash.
> 
> Not sure what this means...any idea? My guess would be that there is no need for CL to request beyond the finalized blocks so can choose not to return them.
> 
> Need to confirm if it means return a null for that entry....so e.g. would a range of 1000 finalized blocks return 1000 nulls?

Confirmed should return null.

Also, for payloadBodiesByRange, the spec will be changed to enforce CL to respect start >= 1 and count >= 1

jflo:
See updates on issue #4787

siladu:
> See updates on issue #4787

Thanks @jflo, just updated the description. I reviewed against merged shanghai spec.

siladu:
https://github.com/ethereum/execution-apis/pull/370

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByHashV1.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import io.vertx.core.Vertx;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EngineGetPayloadBodiesByHashV1 extends ExecutionEngineJsonRpcMethod {
+  // TODO benchmark this number
+  protected static final long MAX_BLOCKS_ALLOWED = 64L;
```
siladu, line 42: Wonder if we should make it configurable? Might be useful during the testing period even if we remove it before the fork.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByHashV1Test.java
```diff
@@ -0,0 +1,266 @@
...
+    final var result = fromSuccessResp(resp);
+    assertThat(result.getPayloadBodies().size()).isEqualTo(2);
+    assertThat(result.getPayloadBodies().get(0).getTransactions().size()).isEqualTo(3);
+    assertThat(result.getPayloadBodies().get(0).getWithdrawals().size()).isEqualTo(1);
+    assertThat(result.getPayloadBodies().get(1).getTransactions().size()).isEqualTo(1);
+    assertThat(result.getPayloadBodies().get(1).getWithdrawals().size()).isEqualTo(1);
+  }
+
+  @Test
+  public void ShouldReturnErrorWhenRequestExceedsPermittedNumberOfBlocks() {
```
siladu, line 228: nit: lower case should

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.stream.LongStream;
+
+import io.vertx.core.Vertx;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EngineGetPayloadBodiesByRangeV1 extends ExecutionEngineJsonRpcMethod {
+  private static final Logger LOG = LoggerFactory.getLogger(EngineGetPayloadBodiesByRangeV1.class);
+  // TODO benchmark this number
+  protected static final long MAX_BLOCKS_ALLOWED = 64L;
```
siladu, line 41: this could be int?

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByHashV1Test.java
```diff
@@ -0,0 +1,266 @@
...
+    final long apiLimit = MAX_BLOCKS_ALLOWED;
+    final int overLimit = (int) apiLimit + 10;
+    final Hash[] arrayHash = new Hash[overLimit];
+
+    // TODO something more efficient than this must exist
```
siladu, line 233: allow limit to be set by the test, set to 1 and request 2 :)

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
siladu: Seems 1024 is a better number after reading around

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByHashV1.java
gfukushima: There's currently nothing that allows us to easily pass a config specifically to the engine API only. We could potentially introduce if it's really worth. Think for now the const would do specially having that defined by the spec.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByHashV1Test.java
gfukushima: ty

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
gfukushima: ty

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByHashV1.java
```diff
@@ -0,0 +1,90 @@
...
+    traceLambda(LOG, "{} parameters: blockHashes {}", () -> getName(), () -> blockHashes);
+
+    if (blockHashes.size() > getMaxRequestBlocks()) {
+      return new JsonRpcErrorResponse(reqId, JsonRpcError.INVALID_RANGE_REQUEST_TOO_LARGE);
+    }
+
+    final Blockchain blockchain = protocolContext.getBlockchain();
+
+    final EngineGetPayloadBodiesResultV1 engineGetPayloadBodiesResultV1 =
+        blockResultFactory.payloadBodiesCompleteV1(
```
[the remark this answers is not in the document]
siladu: It's 'may not' not 'must not'

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByRangeV1.java
```diff
@@ -0,0 +1,111 @@
...
+    return RpcMethod.ENGINE_GET_PAYLOAD_BODIES_BY_RANGE_V1.getMethodName();
+  }
+
+  @Override
+  public JsonRpcResponse syncResponse(final JsonRpcRequestContext request) {
+    engineCallListener.executionEngineCalled();
+
+    final long startBlockNumber = request.getRequiredParameter(0, Long.class);
+    final long count = request.getRequiredParameter(1, Long.class);
+    final Object reqId = request.getRequest().getId();
```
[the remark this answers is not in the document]
gfukushima: Done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadBodiesByHashV1.java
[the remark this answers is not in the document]
gfukushima: My understanding was the same of what @siladu said. I'm considering that everything that isn't specified with MUST is not a item that we're obliged to comply with. Plus, it's likely to add extra work for Besu to check if the blocks are finalized since we currently only keep track of the latest finalized block

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadBodiesResultV1.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import com.fasterxml.jackson.annotation.JsonValue;
+import org.apache.tuweni.bytes.Bytes;
+
+@JsonPropertyOrder({"payloadBodies"})
+public class EngineGetPayloadBodiesResultV1 {
```
siladu, line 31: nit: Not sure it makes sense for this outer result type to be V1...the type in the spec that is versioned is ExecutionPayloadBodyV1 which you have below as PayloadBody...if anything then that should be versioned, but I'm thinking avoid versioning unless we have to.

The only engine api result class that's versioned is EngineGetPayloadResultV1/V2 are versioned because they are different data shapes and we need to distinguish. I am having second thoughts about those names now though.

siladu (approved):
LGTM

jframe (approved):
LGTM

PR #4982, opened 2023-01-23
Don't cache protocol spec in block creator

shemnon:
## PR description

Don't cache the protocol spec in the block creator.  With the new shanghaiTimestamp the correct spec may be a function of the timestamp not just the block number.  So every time we are asked to build a block re-query the spec.

## Documentation

## Changelog


jframe (approved):
LGTM

siladu (commented):
Was thinking we should have a test case (maybe in MergeBlockCreatorTest?) that proves Shanghai spec gets picked up, what do you think?

PR #4983, opened 2023-01-23
address case of block values exceeding long

garyschulte:

## PR description
address corner case where long does not fit the calculated block value included in engine_getPayloadV2.

Change BlockValueCalculator to use Wei rather than unnecessarily converting to long (just to later be converted again into a Quantity)

## Documentation

## Changelog


siladu (approved):
LGTM

PR #4984, opened 2023-01-23
Improve MainnetBlockBodyValidator logging

siladu:
* Associate block number and hash with the various invalid block warnings, in the format:
`Invalid block {blockNumber} ({blockHash}): ... mismatch (expected=..., actual=...)`

Instead of 
```
WARN  | MainnetBlockBodyValidator | Invalid block: receipts root mismatch (expected=0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421, actual=0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421) 
```

we now have 
```
WARN  | MainnetBlockBodyValidator | Invalid block 1 (0xd2ce6fb30db1ab80b0a5c137f3c8a83d12052e5a40366eba4c15b229c6a863e6): receipts root mismatch (expected=0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421, actual=0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421) 
```

---

* Print a less obscure message when there's no error message:

Instead of 
```
INFO  | MainnetBlockValidator | Optional.empty. Block 11 (0x8edd340dd4718e884ffe327bab2d3e3874bac2799cde83a085e5cae66d53c3b5)
```

we now have

```
INFO  | MainnetBlockValidator | Invalid block 11 (0x8edd340dd4718e884ffe327bab2d3e3874bac2799cde83a085e5cae66d53c3b5)
```

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
@@ -189,9 +189,7 @@ public class MainnetBlockValidator implements BlockValidator {
       if (result.errorMessage.isPresent()) {
         LOG.info("Invalid block {}: {}", invalidBlock.toLogString(), result.errorMessage);
       } else {
-        LOG.info(
-            "Invalid block {}: Block import failed with no error message",
-            invalidBlock.toLogString());
+        LOG.info("Invalid block {}", invalidBlock.toLogString());
       }
     }
     badBlockManager.addBadBlock(invalidBlock, result.causedBy());
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
@@ -180,16 +180,18 @@ public class MainnetBlockValidator implements BlockValidator {
       final Block invalidBlock, final BlockValidationResult result) {
     if (result.causedBy().isPresent()) {
       LOG.info(
-          "{}. Block {}, caused by {}",
-          result.errorMessage,
+          "Invalid block {}: {}, caused by {}",
           invalidBlock.toLogString(),
+          result.errorMessage,
           result.causedBy().get());
       LOG.debug("with stack", result.causedBy().get());
     } else {
       if (result.errorMessage.isPresent()) {
-        LOG.info("{}. Block {}", result.errorMessage, invalidBlock.toLogString());
+        LOG.info("Invalid block {}: {}", invalidBlock.toLogString(), result.errorMessage);
       } else {
-        LOG.info("Block import failed with no error message. Block {}", invalidBlock.toLogString());
+        LOG.info(
+            "Invalid block {}: Block import failed with no error message",
+            invalidBlock.toLogString());
       }
     }
     badBlockManager.addBadBlock(invalidBlock, result.causedBy());
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
@@ -186,7 +186,11 @@ public class MainnetBlockValidator implements BlockValidator {
           result.causedBy().get());
       LOG.debug("with stack", result.causedBy().get());
     } else {
-      LOG.info("{}. Block {}", result.errorMessage, invalidBlock.toLogString());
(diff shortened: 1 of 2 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -186,7 +186,11 @@ private void handleAndLogImportFailure(
           result.causedBy().get());
       LOG.debug("with stack", result.causedBy().get());
     } else {
-      LOG.info("{}. Block {}", result.errorMessage, invalidBlock.toLogString());
+      if (result.errorMessage.isPresent()) {
+        LOG.info("{}. Block {}", result.errorMessage, invalidBlock.toLogString());
+      } else {
+        LOG.info("Block import failed with no error message. Block {}", invalidBlock.toLogString());
```
garyschulte, line 192: suggest "Block {} import failed..."

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
```diff
@@ -106,48 +107,75 @@ public boolean validateBodyLight(
     return true;
   }
 
-  private static boolean validateTransactionsRoot(final Bytes32 expected, final Bytes32 actual) {
+  private static boolean validateTransactionsRoot(
+      final BlockHeader header, final Bytes32 expected, final Bytes32 actual) {
+    LOG.info(
+        "Invalid block {} ({}): transaction root mismatch (expected={}, actual={})",
```
garyschulte, line 113: as discussed, this already exists as toLogString()

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
```diff
@@ -162,7 +190,7 @@ private boolean validateEthHash(
     final BlockBody body = block.getBody();
 
     final Bytes32 ommerHash = BodyValidation.ommersHash(body.getOmmers());
-    if (!validateOmmersHash(header.getOmmersHash(), ommerHash)) {
+    if (!validateOmmersHash(header, header.getOmmersHash(), ommerHash)) {
```
garyschulte, line 193: why pass the object and one of its attributes?  we can just pass header right?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
shemnon, line 113: (a) log line belongs inside the if block.
(b) Could this be done with just the header?

garyschulte (commented):
just waiting on your next push

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
```diff
@@ -106,48 +107,75 @@ public boolean validateBodyLight(
...
       return false;
     }
 
     return true;
   }
 
-  private static boolean validateReceiptsRoot(final Bytes32 expected, final Bytes32 actual) {
+  private static boolean validateReceiptsRoot(
+      final BlockHeader header, final Bytes32 expected, final Bytes32 actual) {
+    LOG.warn(
```
shemnon, line 157: log line belongs in the if block.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
siladu: (a) good catch! There was a second one too.
(b) yes, realised I can use header.toLogString() here as well

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
siladu: I consider this, kind of breaks the expect, actual naming convention in the helper methods...nbd either way IMO. Could potentially just pass in the log string instead of header

shemnon (approved):

garyschulte (approved):
lgtm

PR #4985, opened 2023-01-23
RJumpV overflow

shemnon:
## PR description

Address an RJumpV overflow discovered by reference tests.

## Documentation

## Changelog


macfarla (approved):
LGTM. javadoc warning is from KZG precompile so unrelated
-  changelog entry?

evm/src/main/java/org/hyperledger/besu/evm/operation/RelativeJumpVectorOperation.java
```diff
@@ -41,7 +41,12 @@ public RelativeJumpVectorOperation(final GasCalculator gasCalculator) {
   @Override
   protected OperationResult executeFixedCostOperation(final MessageFrame frame, final EVM evm) {
     final Bytes code = frame.getCode().getBytes();
-    final int offsetCase = frame.popStackItem().toInt();
+    int offsetCase;
+    try {
+      offsetCase = frame.popStackItem().toInt() & 0xff;
```
jframe, line 46: This looks like it is treating the offset as an unsigned value. Are negative offsets allowed? From my quick reading of the spec it looked like they are.

shemnon: Yea, this was a quick attempt at optimizing out negatives.  Did a different approach.

PR #4986, opened 2023-01-23
Fix create block using wrong protocol schedule when parent is genesis

jframe:

## PR description
Fix create block using the wrong protocol schedule when the parent is genesis.

## Documentation

## Changelog


siladu (approved):
LGTM

shemnon (approved):

PR #4992, opened 2023-01-24
Gas accounting for EIP-4844

fab-10:

## PR description

This PR introduces the gas accounting for EIP-4844, specifically it introduce a data gas price and a data gas cost.

There is a new fee market named after Cancun that can calculate the data price, a new gas calculator to compute the gas
used by a blob transaction and a new gas limit calculator to track the max data gas per block.
There are changes also to the transactions selection during the block creation to take in account the data gas used and check it against the limits.

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java
```diff
@@ -83,6 +84,9 @@ public abstract class MainnetProtocolSpecs {
   public static final int SPURIOUS_DRAGON_CONTRACT_SIZE_LIMIT = 24576;
   public static final int SHANGHAI_INIT_CODE_SIZE_LIMIT = 2 * SPURIOUS_DRAGON_CONTRACT_SIZE_LIMIT;
 
+  private static final int CANCUN_MIN_DATA_GAS_PRICE = 1;
```
matkt, line 87: Maybe we can move that to CancunGasCalculator

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -328,7 +334,14 @@ public TransactionProcessingResult processTransaction(
...
+      final long dataGas;
+      if (feeMarket.implementsDataFee()) {
+        dataGas = gasCalculator.dataGasCost(transaction.getVersionedHashes().orElseThrow().size());
+      } else {
+        dataGas = 0L;
+      }
+
+      final long gasAvailable = transaction.getGasLimit() - intrinsicGas - accessListGas - dataGas;
       LOG.trace(
           "Gas available for execution {} = {} - {} (limit - intrinsic)",
```
matkt, line 346: can we add dataGas to the log ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -134,10 +128,48 @@ public ValidationResult<TransactionInvalidReason> validate(
...
+
+    if (transaction.isContractCreation() && transaction.getPayload().size() > maxInitcodeSize) {
+      return ValidationResult.invalid(
+          TransactionInvalidReason.INITCODE_TOO_LARGE,
+          String.format(
+              "Initcode size of %d exceeds maximum size of %s",
+              transaction.getPayload().size(), maxInitcodeSize));
+    }
+
+    if (transaction.getType().supportsBlob()) {
```
matkt, line 144: if we want to push all the cost verification in the same method maybe we can also move this one ?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/CancunGasCalculator.java
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.evm.gascalculator;
+
+import org.hyperledger.besu.datatypes.DataGas;
+
+/** The Cancun gas calculator. */
+public class CancunGasCalculator extends LondonGasCalculator {
+
+  public static final int CANCUN_DATA_GAS_PER_BLOB = 131072; // 2^17
```
matkt, line 22: maybe we can link the EIP in a comment for future dev

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -328,7 +334,14 @@ public TransactionProcessingResult processTransaction(
               transaction.getPayload(), transaction.isContractCreation());
       final long accessListGas =
           gasCalculator.accessListGasCost(accessListEntries.size(), accessListStorageCount);
-      final long gasAvailable = transaction.getGasLimit() - intrinsicGas - accessListGas;
+      final long dataGas;
+      if (feeMarket.implementsDataFee()) {
+        dataGas = gasCalculator.dataGasCost(transaction.getVersionedHashes().orElseThrow().size());
```
matkt, line 339: we can use blobcount for this one or remove this one because we are already calculating this value in 306 line

fab-10: good catch, fixed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
fab-10: sure, will add also accessListGas that is missing too

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/CancunGasCalculator.java
fab-10: done in the class doc

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java
fab-10: done

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/CancunGasCalculator.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.evm.gascalculator;
+
+/** The Cancun gas calculator as defined in EIP-4844 */
+public class CancunGasCalculator extends LondonGasCalculator {
+
+  public static final long CANCUN_DATA_GAS_PER_BLOB = 131072; // 2^17
+  public static final long CANCUN_TARGET_DATA_GAS_PER_BLOCK = 262144; // 2^18
```
shemnon, line 21:
```suggestion
  public static final long CANCUN_DATA_GAS_PER_BLOB = 1 << 17;
  public static final long CANCUN_TARGET_DATA_GAS_PER_BLOCK = 1 << 18;
```

fab-10: included

GesusBitcoin (commented):
Available as a browser extension and as a mobile app, MetaMask equips you with a key vault, secure login, token wallet, and token exchange—everything you need to manage your digital assets.
[toddishugeexpertof....txt](https://github.com/hyperledger/besu/files/10542667/toddishugeexpertof.txt)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -424,35 +439,36 @@ public TransactionProcessingResult processTransaction(
       // after the other so that if it is the same account somehow, we end up with the right result)
       final long selfDestructRefund =
           gasCalculator.getSelfDestructRefundAmount() * initialFrame.getSelfDestructs().size();
-      final long refundGas = initialFrame.getGasRefund() + selfDestructRefund;
-      final long refunded = refunded(transaction, initialFrame.getRemainingGas(), refundGas);
-      final Wei refundedWei = transactionGasPrice.multiply(refunded);
+      final long baseRefundGas = initialFrame.getGasRefund() + selfDestructRefund;
```
jflo, line 442: much better names

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionEIP1559Test.java
```diff
@@ -1,79 +0,0 @@
-/*
```
jflo, line 1: Why should we remove this test?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/StateTestSubCommand.java
```diff
@@ -218,12 +220,14 @@ private void traceTestSpecs(final String test, final List<GeneralStateTestCaseEi
...
         }
 
-        ProtocolSpec protocolSpec = protocolSchedule.getByBlockHeader(blockHeader);
+        final ProtocolSpec protocolSpec = protocolSchedule.getByBlockHeader(blockHeader);
         final MainnetTransactionProcessor processor = protocolSpec.getTransactionProcessor();
         final WorldUpdater worldStateUpdater = worldState.updater();
         final ReferenceTestBlockchain blockchain =
             new ReferenceTestBlockchain(blockHeader.getNumber());
         final Stopwatch timer = Stopwatch.createStarted();
+        // Todo: EIP-4844 use the excessDataGas of the parent instead of DataGas.ZERO
```
jflo, line 229: Alternately, perhaps get it from genesis?

jflo (approved):
Looks good, and I have an optional idea for an improvement.  There are a ton of places where we have to pass around the calculated dataGasPrice, and most (all?) of them also are passing around the BlockHeader.  it feels to me that the dataGasPrice for a block could be a transient field on the BlockHeader, that is computed once at construction.

this would also make it so we don’t need to pass the protocol spec around as much, since we don’t need the fee market anymore.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionEIP1559Test.java
fab-10: because it is not an unit test, there are not assert and it is like a _main_ to run some code

fab-10:
> Looks good, and I have an optional idea for an improvement. There are a ton of places where we have to pass around the calculated dataGasPrice, and most (all?) of them also are passing around the BlockHeader. it feels to me that the dataGasPrice for a block could be a transient field on the BlockHeader, that is computed once at construction.
> 
> this would also make it so we don’t need to pass the protocol spec around as much, since we don’t need the fee market anymore.

I have evaluated the transient field approach, and if it can simplify the code in some places, it has some drawbacks in other parts of the code, since if we do not store it, then when getting a block from the db, then we need to calculate the field just after the deserialization, and for that we need the parent header and the protocol spec at that time, or we store that field too, but probably it is better to store it separately from the block header to avoid side effects. I think your idea could work but, it is not trivial, and requires more analysis to make is a general approach for this kind of fields, so for the moment I prefer to keep this as is.

PR #4993, opened 2023-01-24
Change to optional orElseGet

helderjnpinto:
Fix do not evaluate second expression if the value is present.

Signed-off-by: ħþ <12281088+helderjnpinto@users.noreply.github.com>

## PR description

This is causing some issues when we got a earlier version of besu after this PR https://github.com/hyperledger/besu/pull/3065, the problem is caused by gasPrice existing in transaction but is too big to fit in Long types.
Evaluating the second expression is not needed if the gasPrice is present also causing extra processing.


macfarla (approved):
LGTM

PR #5009, opened 2023-01-26
Simplify protocolSchedule query

siladu:
Related to referencetests/evmtool, @shemnon did this: https://github.com/hyperledger/besu/commit/a080bffcbab3d4cd9f9a264285f6b0c39b53a54b

If we revert this, withdrawals referencetests still pass so I think it's the right thing to do.


siladu:
Already implemented on main via another PR

PR #5001, opened 2023-01-25
Revert "parallelization for unit tests (#5000)"

macfarla:
This reverts commit b22a52a88e42a5c11f0564f1c9603bff55687e9e.

Signed-off-by: Sally MacFarlane <macfarla.github@gmail.com>

It seems not all tests are running. Needs more investigation.

See #4921 

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [ ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #5000, opened 2023-01-24
parallelization for unit tests

macfarla:
Signed-off-by: Sally MacFarlane <macfarla.github@gmail.com>

Modify CI unitTest task to utilize test parallelization
* follow what we do for acceptanceTest task
* except we can't split on test name, so split on directories that contain src/test/java 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #5012, opened 2023-01-26
run parallel tests and full unit tests side by side for comparison

macfarla:

## Documentation

## Changelog


macfarla:
abandoning this for now. Issues
* CI reporting only 7608 tests running - should be twice that ie 12439
* can't get junit reports copied out
* full (non-parallelized) run only takes ~ 12 min so it's not that long

macfarla:
ref original #5000 and the revert PR #5001

PR #5023, opened 2023-01-30
Fix data gas calculation during block import

fab-10:
## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -288,16 +288,17 @@ public TransactionProcessingResult processTransaction(
...
-          feeMarket.getTransactionPriceCalculator().price(transaction, blockHeader.getBaseFee());
-      final Wei dataGasPrice =
-          feeMarket.getTransactionPriceCalculator().dataPrice(transaction, blockHeader);
       LOG.trace(
           "Incremented sender {} nonce ({} -> {})",
           senderAddress,
           previousNonce,
           sender.getNonce());
 
+      final Wei transactionGasPrice =
```
shemnon, line 297: Did moving this to the other side of the trace really have an impact?  Can it be rolled back for clarity?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -202,11 +203,6 @@ private ValidationResult<TransactionInvalidReason> validateCostAndFee(
       }
     }
 
-    if (transaction.getNonce() == MAX_NONCE) {
```
shemnon, line 205: This needs to stay.

shemnon (changes requested):
`[MainnetTransactionValidator.java](https://github.com/hyperledger/besu/pull/5023/files#diff-b687d725d48c7858ace60a62773adcf6bf06bc9ec42c61063c59cbb0746726d1)` nonce check needs to be restored.  

Other than that LGTM.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
fab-10: removed since it is a duplication, the check is present at line 156

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
fab-10: My intent was to move the log of the nonce just after the statement that changed it, no behavior change of course

fab-10: My intent was to move the log of the nonce just after the statement that changed it, no behavior change of course

shemnon (approved):

PR #5026, opened 2023-01-31
Improve withdrawals processing performance

ahamlat:

## PR description
During withdrawals processing on devnet, profiling the execution showed two possible little improvements regarding performance.
image

The flamegraph at the block processing level

image

With this PR, Withdrawals processing doesn't appear in the profiling flame graph (see below) which means it is faster than before. We can check also that the withdrawals are processed for each block with this log (see 16 ws for 16 withdrawals).
```
2023-01-31 14:21:24.413+00:00 | vert.x-worker-thread-0 | INFO  | AbstractEngineNewPayload | Imported #28,191 / 0 tx / 16 ws / base fee 7 wei / 0 (0.0%) gas / (0x6a9b06fb18ffdae6377bd0d60d21a665df02c587393402e73cb2502eeda13cae) in 0.001s. Peers: 19
2023-01-31 14:21:37.092+00:00 | vert.x-worker-thread-0 | INFO  | AbstractEngineNewPayload | Imported #28,192 / 0 tx / 16 ws / base fee 7 wei / 0 (0.0%) gas / (0x7c8060685283b172ed030cf8e7d7d609be61b16f58c2e49c493abceea80e572d) in 0.001s. Peers: 14
2023-01-31 14:21:48.651+00:00 | vert.x-worker-thread-0 | INFO  | AbstractEngineNewPayload | Imported #28,193 / 0 tx / 16 ws / base fee 7 wei / 0 (0.0%) gas / (0x96494ca7da0cdc097a19f829f9decd16ed18b5d7f9d2c3920ca609a7936d09c3) in 0.001s. Peers: 14
2023-01-31 14:22:00.572+00:00 | vert.x-worker-thread-0 | INFO  | AbstractEngineNewPayload | Imported #28,194 / 0 tx / 16 ws / base fee 7 wei / 0 (0.0%) gas / (0xb2f8c474e6f5d348b49269068cc0bd06da5affadd51411fb2eec56d5e59a58be) in 0.001s. Peers: 16
```

image

The flamegraph at the block processing level

image

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -284,47 +284,49 @@ public void commit() {
                             new BonsaiValue<>(wrappedWorldView().getCode(addr).orElse(null), null));
                 pendingCode.setUpdated(updatedAccount.getCode());
               }
+              if (!updatedAccount.getUpdatedStorage().isEmpty()) {
```
usmansaleem, line 287: nitpick: if we flip the condition and return early, we might avoid the indentions. i.e. `if (updatedAccount.getUpdatedStorage.isEmpty() {return;}`

usmansaleem (approved):
LGTM, one minor nit pick.

datatypes/src/main/java/org/hyperledger/besu/datatypes/GWei.java
```diff
@@ -111,7 +111,7 @@ public static GWei fromHexString(final String str) {
    * @return Wei
    */
   public Wei getAsWei() {
-    return Wei.of(getAsBigInteger().multiply(BigInteger.TEN.pow(9)));
+    return Wei.of(getAsBigInteger().multiply(BigInteger.valueOf(1_000_000_000L)));
```
usmansaleem, line 114: nitpick: We might even bring `BigInteger.valueOf(1_000_000_000L)` out as a constant.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
siladu: It's not clear to me why this improves withdrawals performance, can you please explain?

Also, it seems to be a change that will affect things other than withdrawals, so what has been done to ensure we haven't regressed anything else?

Finally, it's bonsai specific - what if we process withdrawals with FOREST?

jframe: This looks like different optimisation from the one mentioned initially. I thought the optimization was to avoid the cost of the parallel streams overhead. This looks like it avoids doing unnecessary work if there is no storage to update. Which is a good thing though.

Are you planning to add an optimization to avoid the parallel stream for withdrawal balance updates?

ahamlat: @siladu You're right, the PR description is not detailed enough. During block processing, we have 3 big consumers in terms of time processing : Transactions execution, commit the changes of each transaction to the updater and persist all the changes (in memory or on disk). During withdrawals processing, we also commit the changes of all the withdrawals to the worldstate updater but we use the same method (commit) that is used to manage all the cases (accounts deletion, storage update, storage deletion, ..etc). In the case of withdrawals, we only modify the balances, so we should just focus on accounts updates.
 @jframe good point, you're totally right. I wanted to start with this first optimisation to see if it is sufficient, and avoid creating another specific method for withdrawals to keep the code simple.

ahamlat: I added the flamegraph at the block processing level to better see that withdrawals processing doesn't appear as it takes less time than before

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Add implementation for eth_createAccessList RPC method [#4942](https://github.com/hyperledger/besu/pull/4942)
 - Updated reference tests to v11.3 [#4996](https://github.com/hyperledger/besu/pull/4996)
 - Add DebugGetRawBlock and DebugGetRawHeader RPC methods [#5011](https://github.com/hyperledger/besu/pull/5011)  
+- Improve withdrawals processing performance [#5026](https://github.com/hyperledger/besu/pull/5026)
```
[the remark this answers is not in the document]
ahamlat: Okay, I'm going to delete it than

datatypes/src/main/java/org/hyperledger/besu/datatypes/GWei.java
ahamlat: Done

jframe (approved):

Issue #4934, opened 2023-01-15
Run withdrawals execution test


jframe:
Run the execution spec tests for withdrawals.

Tests have also had Withdrawals added to them: https://github.com/ethereum/execution-spec-tests

garyschulte:
Will close this one after adding execution spec tests to ci

siladu:
https://github.com/garyschulte/besu/tree/feature/execution-spec-tests

PR #5030, opened 2023-02-01
execution-spec-tests tooling support

shemnon:
## PR description

Add needed tooling support for execution-spec-tests

Adding the `t8n` and `b11r` tool allows Hyperledger besu to produce reference tests from the https://github.com/ethereum/execution-spec-tests repository.

* Add t8n tool, or transition tool.  Calculates the state changes from a set of transactions against a given parent context
* Add b11r tool, or block-builder tool.  Given block parts it can generate a RLP block.
* Changes to some tracing CLI flags to work better with retesteth

## Documentation

## Changelog


holiman:
Please put in the PR description what CLI flags are changed. Apparently `--nomemory` is gone, but it now says it has 
```
      --trace.stack        Enable showing the operand stack in tracing for each
                             op.
```
However, it seems to show me the stack even if I don't explicitly ask for it, just `--json state-test foo`

shemnon:
I'll restore --nomemory.

shemnon:
`--nomemory` added.  the other `--trace.` flaggs are all boolean flags now.  Defautls are a mix of true and false but `--trace.no<foo>` and `--trace.<foo>` will fix them to on or off.

Unit test fixed, I was juggling rewards from both the protocol and CLI and flubbed it.  In other news unit tests did their job.

Ready for another round of reviews.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/B11rSubCommand.java
```diff
@@ -0,0 +1,245 @@
...
+      description = "The ommers for the block")
+  private final Path ommers = stdinPath;
+
+  @Option(
+      names = {"--seal.clique"},
+      paramLabel = "full path",
+      description = "The clique seal/signature for the block")
+  private final Path sealClique = stdinPath;
+
+  @SuppressWarnings("UnusedVariable") // FIXME do ethash sealing
```
siladu, line 89: Is this for a future PR?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/B11rSubCommand.java
```diff
@@ -0,0 +1,245 @@
...
+    } else {
+      rlpOut.startList();
+      rlpOut.endList();
+    }
+    // ommers
+    rlpOut.startList();
+    rlpOut.endList();
+
+    // withdrawals
+    // TODO
```
siladu, line 209: Still TODO? are you blocked on something related to withdrawals?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -143,10 +149,27 @@ void setBytes(final String optionValue) {
   final Boolean showJsonAlloc = false;
 
   @Option(
-      names = {"--nomemory"},
-      description = "Disable showing the full memory output for each op.",
-      scope = INHERIT)
-  final Boolean noMemory = false;
+      names = {"--memory", "--trace.memory"},
```
siladu, line 152: From the PR conversation, sounds like this should still be --nomemory? Or is it the comment that's wrong? https://github.com/hyperledger/besu/pull/5030/#issuecomment-1435324768

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -163,23 +186,46 @@ void setBytes(final String optionValue) {
...
+    // add dagger-injected options
     commandLine.addMixin("Dagger Options", daggerOptions);
 
     // add sub commands here
     commandLine.registerConverter(Address.class, Address::fromHexString);
     commandLine.registerConverter(Bytes.class, Bytes::fromHexString);
     commandLine.registerConverter(Wei.class, arg -> Wei.of(Long.parseUnsignedLong(arg)));
 
-    commandLine.setExecutionStrategy(resultHandler).execute(args);
+    // change negation regexp so --nomemory works
```
siladu, line 218: This might be wrong depending on if we're going for --memory over --nomemory

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -163,23 +186,46 @@ void setBytes(final String optionValue) {
...
+    commandLine.setNegatableOptionTransformer(
+        new CommandLine.RegexTransformer.Builder()
+            .addPattern("^--no(\\w(-|\\w)*)$", "--$1", "--[no]$1")
+            .addPattern("^--trace.no(\\w(-|\\w)*)$", "--trace.$1", "--trace.[no]$1")
+            .addPattern("^--(\\w(-|\\w)*)$", "--no$1", "--[no]$1")
+            .addPattern("^--trace.(\\w(-|\\w)*)$", "--trace.no$1", "--trace.[no]$1")
```
siladu, line 224: I have no idea what's going on here, maybe a more detailed comment would help

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
```diff
@@ -0,0 +1,557 @@
+/*
+ * Copyright ConsenSys AG.
```
siladu, line 2: Old copyright

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
```diff
@@ -0,0 +1,557 @@
...
+    List<TransactionReceipt> receipts = new ArrayList<>();
+    List<RejectedTransaction> invalidTransactions = new ArrayList<>();
+    List<Transaction> validTransactions = new ArrayList<>();
+    ArrayNode receiptsArray = objectMapper.createArrayNode();
+    long gasUsed = 0;
+    for (int i = 0; i < transactions.size(); i++) {
+      Transaction transaction = transactions.get(i);
+
+      final Stopwatch timer = Stopwatch.createStarted();
+      final OperationTracer tracer; // You should have picked Mercy.
```
siladu, line 281: ?? 😆

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
```diff
@@ -0,0 +1,557 @@
...
+
+    // block reward
+    if (!validTransactions.isEmpty()) {
+      worldStateUpdater
+          .getOrCreateSenderAccount(blockHeader.getCoinbase())
+          .getMutable()
+          .incrementBalance((reward == null) ? protocolSpec.getBlockReward() : reward);
+    }
+
+    // deposit CL withdrawals
```
siladu, line 384: that sentence is going to be confusing when we add in-protocol validator deposits 😅

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
```diff
@@ -0,0 +1,557 @@
...
+    try {
+      protocolSpec
+          .getWithdrawalsProcessor()
+          .ifPresent(
+              p -> p.processWithdrawals(referenceTestEnv.getWithdrawals(), worldStateUpdater));
```
siladu, line 389: I might be misunderstanding the point of this tool, but aren't we duplicating and potentially creating diverging behaviour from how mainnet besu behaves?

For example, in AbstractBlockProcessor, we only process withdrawals if both the processor is present and also if withdrawals are present, see: https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java#L143-L149

Same applies to other operations conditionally invoked by this tool. At a glance, it seems like behaviour of mainnet vs this tool may diverge further over time too, so doesn't that invalidate the test?

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/ReferenceTestProtocolSchedules.java
```diff
@@ -42,48 +42,46 @@ public class ReferenceTestProtocolSchedules {
...
+
+    // the following schedules activate EIP-1559, but may have non-default
+    if (genesisStub.getBaseFeePerGas().isEmpty()) {
+      genesisStub.baseFeePerGas(0x0a);
+    }
+    builder.put("London", createSchedule(genesisStub.clone().londonBlock(0)));
+    builder.put("ArrowGlacier", createSchedule(genesisStub.clone().arrowGlacierBlock(0)));
+    builder.put("GrayGlacier", createSchedule(genesisStub.clone().grayGlacierBlock(0)));
+    builder.put("Merge", createSchedule(genesisStub.clone().mergeNetSplitBlock(0)));
+    builder.put("Merge+3545+3670", createTimestampSchedule(genesisStub.clone().shanghaiTime(0)));
```
siladu, line 80: I guess these are EVM EIPS - I can't find 3545 though?

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/ReferenceTestProtocolSchedules.java
```diff
@@ -42,48 +42,46 @@ public class ReferenceTestProtocolSchedules {
...
+    builder.put("Merge+3860", createTimestampSchedule(genesisStub.clone().shanghaiTime(0)));
+    builder.put("Merge+3855", createTimestampSchedule(genesisStub.clone().shanghaiTime(0)));
```
siladu, line 82: Why do we need PUSH0 separately now that it's part of Shanghai?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -163,23 +186,46 @@ void setBytes(final String optionValue) {
...
+
   static final Joiner STORAGE_JOINER = Joiner.on(",\n");
   private final EvmToolCommandOptionsModule daggerOptions = new EvmToolCommandOptionsModule();
-  private PrintWriter out =
+  PrintWriter out =
       new PrintWriter(new BufferedWriter(new OutputStreamWriter(System.out, UTF_8)), true);
+  InputStream in = new ByteArrayInputStream(new byte[0]);
 
-  void parse(final CommandLine.IExecutionStrategy resultHandler, final String[] args) {
+  void execute(final String... args) {
```
siladu, line 201: Good rename 👍

siladu (commented):
Awesome work! Mostly just have a some questions.

I got confused with what you were going for w.r.t --nomemory vs --memory. I think you went for --memory but some of the code and comments seemed to suggest otherwise.

Also, it sounds like this PR might make this CI work that @garyschulte started and paused redundant:
https://github.com/hyperledger/besu/issues/4934#issuecomment-1406090671
IIRC he was using a gradle task to pull down a execution-spec-tests release artifact and add them to the existing reference test executor so they would run as part of the `referenceTests` task. 

...Will this PR result in the execution-spec-tests running in our CI?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/B11rSubCommand.java
shemnon: t8n tool specifies CLI flags for ethash sealing.  I'm not sure if the retesteth tooling needs it or not, but we don't wire it in.  Much like clique seal we ignore it currently as no meaningful value is passed in by execution-spec-tests, and differential testing shows no diffs relating to extradata in the block (which is where clique seals live).

When I loop back to retesteth support this will get some additional support to pass differential data

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/B11rSubCommand.java
shemnon: Waiting on the `execution-spec-tests` to pass in withdrawals information to the `b11r` tool.  When that support comes this is where it lands.  clarified todo.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
shemnon: Line 156 added the "negatable" flag to this option and I added the proper regexp support at line 219.  This boolean flag is now `--nomemory` or ``--memory`, or `--trace.nomemory` or `--trace.memory` and the boolen is set based on the particular flag.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
shemnon: I've added unit tests to verify it works correctly, per the [Beyoncé rule](https://abseil.io/resources/swe-book/html/ch11.html#the_beyonceacutesemicolon_rule)

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
shemnon: They should be the same.  But the difficulty arises in how the test needs to instrument it's output, in addition to other presumptions about how to handle invalid transactions (abort whole block in production vs. discard and produce some block output in test).

This tool also, as much as is exposed, delegates to other existing components to calculate values.

This tool in itself is not the test, but generates the testing vectors.  Because of that it needs to create some output from testing inputs that would ordinarily be fatal to regular processing.  That's why simply sending it off to a block processor won't work because it's not enough to know it fails, but to produce certain vectors for specific invalid calls.

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/ReferenceTestProtocolSchedules.java
shemnon: These specific values were extracted from a particular set of `develop` branch tests from the reference test.  Since we cannot easily handle "fork+EIP" they are encoded here to provide testing for EOFv1 tests until they are formally sunset.

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/ReferenceTestProtocolSchedules.java
shemnon: When this file was updated these were generated these were the correct values - https://github.com/ethereum/tests/blob/c896d1eabd9719c6bd80b979567573c7ec111429/GeneralStateTests/EIPTests/stEIP3860/create2InitCodeSizeLimit.json#L50

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
shemnon: If ConsenSys were to migrate these to the new format we could have spotless verify and then magically apply them to new files.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
shemnon: Internet joke.  https://www.youtube.com/watch?v=d0RmRJsgP28
It's also here - https://github.com/shemnon/besu/blob/19818121e73fa47a99ad72d8871e3ba824dcb62c/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java#L311

t8n tool copied a lot of the execution framework from the main evm tool tracer.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
shemnon: rewritten.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
shemnon: It is indeed an incantation, and any deeper explanation than "it makes --nomemory work" involves explaining PicoCLI magic.  I added a link to the PicoCLI docs where they explain that magic.

besu/src/main/java/org/hyperledger/besu/BesuInfo.java
```diff
@@ -31,7 +31,16 @@ public final class BesuInfo {
   private BesuInfo() {}
 
   /**
-   * Generate Besu version
+   * Generate version-only Besu version
+   *
+   * @return Besu version in format such as besu/v22.10.3/linux-aarch_64/openjdk-java-11
```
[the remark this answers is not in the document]
shemnon: Yea, I also added multiple examples.

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/ReferenceTestProtocolSchedules.java
shemnon: I've removed them in the interim.  But when they show back up in the reference tests EIP directories similar handling will be needed.

shemnon:
> I got confused with what you were going for w.r.t --nomemory vs --memory. I think you went for --memory but some of the code and comments seemed to suggest otherwise.

I want both to work, so the default can change arbitrarily.  If you explicitly want memory or not use the `--memory` or `--nomemory` and the flag will reflect that choice.  No CLI means the current default will be used. There is some PicoCLI config needed to accomplish this, as the default wants to do `--no-memory`, but the standard APIs from geth don't use the dash for their toggles.

> Also, it sounds like this PR might make this CI work that @garyschulte started and paused redundant:
https://github.com/hyperledger/besu/issues/4934#issuecomment-1406090671
IIRC he was using a gradle task to pull down a execution-spec-tests release artifact and add them to the existing reference test executor so they would run as part of the referenceTests task.

No, this is complementary.  The resulting .json files are generated with the t8n tool, but they are different steps.  Javac vs Java.  So Gary's work will still be needed.

> ...Will this PR result in the execution-spec-tests running in our CI?

No.  This PR adds features to the evm tool so that the execution-spec-tests can use Besu to generate the final test vectors.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
siladu: ah ok, that makes sense now thanks.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
siladu: That certainly helps, I didn't clock it was pico magic first time round.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
siladu: Fair point

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
siladu: Thanks for the explanation. Wonder if there's a way to instrument in line in the code? Maybe not worth the readability/complexity cost of doing so though.

siladu (approved):
Looks good. 

> This PR adds features to the evm tool so that the execution-spec-tests can use Besu to generate the final test vectors.

I'm missing some context - is besu the only client used to generate these or are multiple clients used, if so why?
My assumption would be that test vectors can be reused across all clients, but maybe not :)

Do the besu docs want updating as well?

shemnon:
Geth is the only other client to implement t8n at the moment.  Besu and geth implement the retesteth endpoints.  There is pressure on Nethermind to implement one of the testing APIs and the one they are pointing them to is t8n, so it looks like t8n support will be needed.  The long term client diversity goal is to have every client capable of "filling" (or compiling) the reference tests.

The goal of having all clients implement the testing APIs is then new EIPs can be first implemented on any of the clients, and then the other two can work on implementation and differential testing from there.

As for documentation, typically these tools are only used by automation. The human audience is about 10 people, max, and highly enfranchised into the ethereum ecosystem.  The "spec" I wrote this off of is here https://ethereum-tests.readthedocs.io/en/latest/t8ntool.html#command-line-parameters.  For the b11r tool nothing more than a more verbose restating of the --help options would be needed.  Users who aren't sure of the exact implication are comfortable diving into clients codebases to see what is really going on.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nSubCommand.java
shemnon: And we would have to factor out block production, which is currently baked into the AbstractBlockProcessor.

One alternative is a static method that is public and factored out where you bring all the pieces needed to process a block, and have both tools call that at the appropriate time and place.

siladu:
> The goal of having all clients implement the testing APIs is then new EIPs can be first implemented on any of the clients, and then the other two can work on implementation and differential testing from there. 

Nice, makes sense.

> As for documentation, typically these tools are only used by automation. 

Only mentioned updating docs because they already exist: https://besu.hyperledger.org/en/stable/public-networks/reference/evm-tool/#nomemory

winsvega:
Can you add the docs reference how to use it (compile with t8n) ?

PR #5037, opened 2023-02-01
Params should be single item of array type, not outer array of strings

siladu:
See https://github.com/ethereum/execution-apis/issues/373


gfukushima (approved):
LTGM

PR #5039, opened 2023-02-02
Keep Worldstate Storage open for Bonsai archive latest layer

### LINKED ISSUE #5038: eth_getBalance returning 0x0 for latest block
### Description
eth_getBalance endpoint often erroneously returns 0 balance for accounts when querying using 'latest' block identifier.

### Acceptance Criteria
* eth_getBalance should reliably return the balance of an account

### Steps to Reproduce (Bug)
1. sync besu to head using bonsai storage format
2. query eth_getBalance endpoint for any address, using `latest` as the block hash

correct balance returned

0x0 balance returned vast majority of the time

* Software version: 23.1.0-RC1

garyschulte:
## PR description
Fixes an issue where the latest block in BonsaiWorldStateArchive had a closed worldstate.  Persisting BonsaiInMemoryWorldState instances were creating entries in the archive with a snapshot worldstate that would subsequently be closed.  The net result was that queries using the worldstate for the `latest` block would frequently return Optional.empty() when accessing the closed worldstate.

this PR:
* uses a dedicated copy of the Bonsai in memory worldstate when creating a trielog
* adds a close() override to Bonsai layered worldstate to ensure the worldstate resources are cleaned up
* makes minor log change 

## Fixed Issue(s)

fixes #5038 

## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
@@ -150,7 +150,12 @@ public class BonsaiInMemoryWorldState extends BonsaiPersistedWorldState
     final Hash newWorldStateRootHash = rootHash(localUpdater);
     archive
         .getTrieLogManager()
-        .saveTrieLog(archive, localUpdater, newWorldStateRootHash, blockHeader, this);
+        .saveTrieLog(
+            archive,
+            localUpdater,
+            newWorldStateRootHash,
+            blockHeader,
+            (BonsaiPersistedWorldState) this.copy());
     worldStateRootHash = newWorldStateRootHash;
     worldStateBlockHash = blockHeader.getBlockHash();
     isPersisted = true;
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
@@ -41,7 +41,8 @@ import org.apache.tuweni.bytes.Bytes32;
 import org.apache.tuweni.units.bigints.UInt256;
 
 /** A World State backed first by trie log layer and then by another world state. */
-public class BonsaiLayeredWorldState implements MutableWorldState, BonsaiWorldView, WorldState, BonsaiStorageSubscriber {
+public class BonsaiLayeredWorldState
+    implements MutableWorldState, BonsaiWorldView, WorldState, BonsaiStorageSubscriber {
   private Optional<BonsaiWorldView> nextWorldView;
   private final AtomicLong worldStateSubscription;
   protected final long height;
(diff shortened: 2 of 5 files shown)


macfarla (approved):
makes sense to me. changelog entry?

ahamlat:
I tested eth_getBalance several times with this PR and the result matches what I can see in Etherscan or by executing eth_getBalance on 22.10.3 (which does not have the regression).
I don't see any performance regression on block processing and FCU (from prometheus metric and CPU profiling).

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
```diff
@@ -150,7 +150,12 @@ public void persist(final BlockHeader blockHeader) {
     final Hash newWorldStateRootHash = rootHash(localUpdater);
     archive
         .getTrieLogManager()
-        .saveTrieLog(archive, localUpdater, newWorldStateRootHash, blockHeader, this);
+        .saveTrieLog(
+            archive,
+            localUpdater,
+            newWorldStateRootHash,
+            blockHeader,
+            (BonsaiPersistedWorldState) this.copy());
```
matkt, line 158: is this copy can impact the performance ? I see some synchronized method , call to DB . if @ahamlat  can confirm that it is safe

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -75,9 +75,29 @@ public Optional<BonsaiWorldView> getNextWorldView() {
   }
 
   public void setNextWorldView(final Optional<BonsaiWorldView> nextWorldView) {
+    maybeUnSubscribe();
```
matkt, line 78: If we close the nextview, how will the layeredworldstate be able to ask it to retrieve the different values? Is it going to recreate it when needed?

matkt, line 78: If we close the nextview, how will the layeredworldstate be able to ask it to retrieve the different values? Is it going to recreate it when needed?

matkt (commented):
just asked some questions to understand

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
garyschulte: this is to close/unsubscribe from the worldstate this layer was initially created with, so we do not leak any snapshots.  since this is a mutator, this is the last opportunity we will have to unsubscribe, typically when the BonsaiWorldStateArchive is responding to a BlockAddedEvent, and changes the nextWorldView.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiInMemoryWorldState.java
garyschulte: Agreed, an alternate implementation is to directly subscribe and unsubscribe to the underlying worldstate storage.  This is cleaner code-wise, but if there is a performance impact we can directly sub/unsub to the storage itself.

PR #5044, opened 2023-02-03
Cancel older block creation tasks upon receiving a new one

siladu:
To avoid parallel block building on different threads, cancel any previous block building threads before starting a new one.


siladu:
Prysm may sometimes trigger this situation, which may up the priority of this story a bit

https://discord.com/channels/595666850260713488/892088344438255616/1082646887536001146
> Hi, what does the EL do if I send twice the same FCU with the same payload attributes for the next slot, does it restart block production or does it ignore the second call? Is there any penalty in calling twice?
> There are a few edge cases where I sort of can't avoid calling twice with the same attributes

geth does this:
> If the attributes are exactly the same, we don't start a new task. If you change them up a bit, we will start a new building task
Having two tasks building at the same time costs cpu

fab-10:
Closing this as superseded by https://github.com/hyperledger/besu/pull/5263

PR #4873, opened 2023-01-03
Add `Future` and `Experimental` forks and networks

shemnon:
## PR description

Add two new development networks and forks, `Future` for EIPs not yet scheduled for a fork and `Experimental` for EIPs and other work not yet accepted to a public network.  Two new fork-named networks are added like the `dev` network to facilitate easy local development. These replace the deprecated `Shandong` network and fork.

Signed-off-by: Danno Ferrin <danno.ferrin@swirldslabs.com>

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [ ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #5055, opened 2023-02-05
EIP-6110: Handle validator deposits in EL

naviechan:
## PR description

Appends validator deposits to the Execution Layer block structure (EIP-6110)[https://eips.ethereum.org/EIPS/eip-6110]

## Fixed Issue(s)
TODO

## Documentation

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/DepositParameter.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/DepositParameter.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright Hyperledger Besu Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.vertx.core.json.JsonObject;
+import org.apache.tuweni.units.bigints.UInt64;
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.datatypes.GWei;
+import org.hyperledger.besu.ethereum.core.Withdrawal;
+import org.hyperledger.besu.ethereum.core.Deposit;
+
+import java.util.Objects;
+
+public class DepositParameter {
+
+  private final String pubKey;
+
+  private final String withdrawalCredentials;
+  private final String amount;
+
+  private final String signature;
+  private final String index;
+
+  @JsonCreator
+  public DepositParameter(
+      @JsonProperty("pubKey") final String pubKey,
+      @JsonProperty("withdrawalCredentials") final String withdrawalCredentials,
+      @JsonProperty("amount") final String amount,
+      @JsonProperty("signature") final String signature,
+      @JsonProperty("index") final String index) {
+        this.pubKey = pubKey;
+        this.withdrawalCredentials = withdrawalCredentials;
+        this.amount = amount;
+        this.signature = signature;
+        this.index = index;
+  }
+
+  public static DepositParameter fromDeposit(final Deposit deposit) {
+    return new DepositParameter(
+        deposit.getPubKey().toString(), //TODO
+        deposit.getWithdrawalCredentials().toString(),
+        deposit.getAmount().toShortHexString(),
+        deposit.getSignature().toString(),
+        deposit.getIndex().toBytes().toQuantityHexString()
+    );
+  }
+
+  //TODO
+  public Deposit toDeposit() {
+    return new Deposit(
+        pubKey,
+        withdrawalCredentials,
+        GWei.fromHexString(amount),
+        signature,
+        UInt64.fromHexString(index));
+  }
+
+  public JsonObject asJsonObject() {
+    return new JsonObject()
+        .put("pubKey", pubKey)
+        .put("withdrawalCredentials", withdrawalCredentials)
+        .put("amount", amount)
+        .put("signature", signature)
+        .put("index", index);
+  }
+
+  @JsonGetter
+  public String getPubKey() {
+    return pubKey;
+  }
+
+  @JsonGetter
+  public String getWithdrawalCredentials() {
+    return withdrawalCredentials;
+  }
+
+  @JsonGetter
+  public String getAmount() {
+    return amount;
+  }
+
+  @JsonGetter
+  public String getSignature() {
+    return signature;
+  }
+
+  @JsonGetter
+  public String getIndex() {
+    return index;
+  }
+
+  @Override
+  public boolean equals(final Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    final DepositParameter that = (DepositParameter) o;
+    return Objects.equals(pubKey, that.pubKey)
+        && Objects.equals(withdrawalCredentials, that.withdrawalCredentials)
+        && Objects.equals(amount, that.amount)
+        && Objects.equals(signature, that.signature)
+        && Objects.equals(index, that.index);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(pubKey, withdrawalCredentials, amount, signature, index);
+  }
+
+  @Override
+  public String toString() {
+    return "DepositParameter{" +
+            "pubKey='" + pubKey + '\'' +
+            ", withdrawalCredentials='" + withdrawalCredentials + '\'' +
+            ", amount='" + amount + '\'' +
+            ", signature='" + signature + '\'' +
+            ", index='" + index + '\'' +
+            '}';
+  }
+
+}
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadResultV6110.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadResultV6110.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright Hyperledger Besu Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.api.jsonrpc.internal.results;
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import org.apache.tuweni.bytes.Bytes32;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.DepositParameter;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.WithdrawalParameter;
+import org.hyperledger.besu.ethereum.core.BlockHeader;
+import org.hyperledger.besu.ethereum.core.Deposit;
+import org.hyperledger.besu.ethereum.core.Withdrawal;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+@JsonPropertyOrder({
+  "executionPayload",
+  "blockValue",
+})
+public class EngineGetPayloadResultV6110 {
+  protected final PayloadResult executionPayload;
+  private final String blockValue;
+
+  public EngineGetPayloadResultV6110(
+      final BlockHeader header,
+      final List<String> transactions,
+      final Optional<List<Withdrawal>> withdrawals,
+      final Optional<List<Deposit>> deposits,
+      final String blockValue) {
+    this.executionPayload = new PayloadResult(header, transactions, withdrawals, deposits);
+    this.blockValue = blockValue;
+  }
+
+  @JsonGetter(value = "executionPayload")
+  public PayloadResult getExecutionPayload() {
+    return executionPayload;
+  }
+
+  @JsonGetter(value = "blockValue")
+  public String getBlockValue() {
+    return blockValue;
+  }
+
+  public static class PayloadResult {
+
+    protected final String blockHash;
+    private final String parentHash;
+    private final String feeRecipient;
+    private final String stateRoot;
+    private final String receiptsRoot;
+    private final String logsBloom;
+    private final String prevRandao;
+    private final String blockNumber;
+    private final String gasLimit;
+    private final String gasUsed;
+    private final String timestamp;
+    private final String extraData;
+    private final String baseFeePerGas;
+    protected final List<String> transactions;
+    private final List<WithdrawalParameter> withdrawals;
+
+    private final List<DepositParameter> deposits;
+
+    public PayloadResult(
+            final BlockHeader header,
+            final List<String> transactions,
+            final Optional<List<Withdrawal>> withdrawals,
+            final Optional<List<Deposit>> deposits) {
+      this.blockNumber = Quantity.create(header.getNumber());
+      this.blockHash = header.getHash().toString();
+      this.parentHash = header.getParentHash().toString();
+      this.logsBloom = header.getLogsBloom().toString();
+      this.stateRoot = header.getStateRoot().toString();
+      this.receiptsRoot = header.getReceiptsRoot().toString();
+      this.extraData = header.getExtraData().toString();
+      this.baseFeePerGas = header.getBaseFee().map(Quantity::create).orElse(null);
+      this.gasLimit = Quantity.create(header.getGasLimit());
+      this.gasUsed = Quantity.create(header.getGasUsed());
+      this.timestamp = Quantity.create(header.getTimestamp());
+      this.transactions = transactions;
+      this.feeRecipient = header.getCoinbase().toString();
+      this.prevRandao = header.getPrevRandao().map(Bytes32::toHexString).orElse(null);
+      this.withdrawals =
+          withdrawals
+              .map(
+                  ws ->
+                      ws.stream()
+                          .map(WithdrawalParameter::fromWithdrawal)
+                          .collect(Collectors.toList()))
+              .orElse(null);
+      this.deposits =
+          deposits
+              .map(
+                  ws ->
+                      ws.stream()
+                          .map(DepositParameter::fromDeposit)
+                          .collect(Collectors.toList()))
+              .orElse(null);
+    }
+
+    @JsonGetter(value = "blockNumber")
+    public String getNumber() {
+      return blockNumber;
+    }
+
+    @JsonGetter(value = "blockHash")
+    public String getHash() {
+      return blockHash;
+    }
+
+    @JsonGetter(value = "parentHash")
+    public String getParentHash() {
+      return parentHash;
+    }
+
+    @JsonGetter(value = "logsBloom")
+    public String getLogsBloom() {
+      return logsBloom;
+    }
+
+    @JsonGetter(value = "prevRandao")
+    public String getPrevRandao() {
+      return prevRandao;
+    }
+
+    @JsonGetter(value = "stateRoot")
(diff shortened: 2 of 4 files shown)


siladu:
Hey @naviechan check out this page about DCO https://wiki.hyperledger.org/display/BESU/DCO

All commits need to be signed off to pass this check. You can amend your commit messages and force push (or squash the commits and do it once).

I've just approved the CI to run the rest of the checks - you might need to prod us if you want to run these at any point (or you can just run gradle commands locally to check unit tests etc)

Spotless is failing, to fix run `./gradlew spotlessApply` and commit the changes.

Javadoc is also failing, run `./gradlew javadoc` locally to and fix up the warnings, certain packages require javadoc to be added to public methods.

Finally, assemble is passing. I would recommend at least running `./gradlew assemble` locally before pushing to CI.

naviechan:
> Hey @naviechan check out this page about DCO https://wiki.hyperledger.org/display/BESU/DCO
> 
> All commits need to be signed off to pass this check. You can amend your commit messages and force push (or squash the commits and do it once).
> 
> I've just approved the CI to run the rest of the checks - you might need to prod us if you want to run these at any point (or you can just run gradle commands locally to check unit tests etc)
> 
> Spotless is failing, to fix run `./gradlew spotlessApply` and commit the changes.
> 
> Javadoc is also failing, run `./gradlew javadoc` locally and fix up the warnings, certain packages require javadoc to be added to public methods.
> 
> Finally, assemble is failing. I would recommend at least running `./gradlew assemble` locally before pushing to CI. Not a bad idea to always run spotlessApply too, you can script both this and the dco so you don't forget (I use a precommit hook)

Thanks so much this is very helpful! I think I ran into some kind of error for `:acceptance-tests:tests:compileSolidity`. The server returns `503` not sure what I can do on my end.

Also sorry for this huge PR. Was trying to limit the scope of this PR to just add `Deposit` to block header and body. Turns out I had to modify a lot of the classes 😂

I will open another PR for json-rpc related changes

siladu:
> Thanks so much this is very helpful! I think I ran into some kind of error for `:acceptance-tests:tests:compileSolidity`. The server returns `503` not sure what I can do on my end.

This was a transient error, I gave it a kick and now you have a genuine test compilation error.
 
> Also sorry for this huge PR. Was trying to limit the scope of this PR to just adding `Deposit` to block header and body. Turns out I had to modify a lot of the classes 😂

Large PRs are hard to review for sure, but not impossible. One approach that seems to work well is something like...
- PR to add feature toggle
- PR to add new types
- PR per feature, behind the toggle where necessary (should be production grade, including tests)

You can always use one giant draft PR as a prototype before splitting into smaller ones, but I think getting some code onto main ASAP is the least painful approach.

By the way, I see you added `eip6110time`. That'll work fine and it may prove useful to have a separate flag, but note we already have `experimentaleipstime` for this purpose: https://github.com/hyperledger/besu/pull/4873

siladu:
@naviechan would you like someone to do an early review or would you prefer to wait until you mark it as ready?

naviechan:
> @

Yes please do an early review

naviechan:
> > Thanks so much this is very helpful! I think I ran into some kind of error for `:acceptance-tests:tests:compileSolidity`. The server returns `503` not sure what I can do on my end.
> 
> 
> 
> This was a transient error, I gave it a kick and now you have a genuine test compilation error.
> 
>  
> 
> > Also sorry for this huge PR. Was trying to limit the scope of this PR to just adding `Deposit` to block header and body. Turns out I had to modify a lot of the classes 😂
> 
> 
> 
> Large PRs are hard to review for sure, but not impossible. One approach that seems to work well is something like...
> 
> - PR to add feature toggle
> 
> - PR to add new types
> 
> - PR per feature, behind the toggle where necessary (should be production grade, including tests)
> 
> 
> 
> You can always use one giant draft PR as a prototype before splitting into smaller ones, but I think getting some code onto main ASAP is the least painful approach.
> 
> 
> 
> By the way, I see you added `eip6110time`. That'll work fine and it may prove useful to have a separate flag, but note we already have `experimentaleipstime` for this purpose: https://github.com/hyperledger/besu/pull/4873
> 
> 
> 
> 

Thanks for the feedback. I will remove the EIP6110Time in favour of `experimentaleipstime`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
```diff
@@ -304,4 +308,19 @@ private boolean validateWithdrawals(final Block block) {
 
     return true;
   }
+
+  private boolean validateDeposits(final Block block) {
```
siladu, line 312: Perhaps we need this as well, but the "Block Validity" section in the EIP is quite a different algorithm to this as it parses the transactions.

siladu (commented):
Did a first pass and have a few comments.

Watch out for DCO too :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Deposit.java
```diff
@@ -0,0 +1,123 @@
...
+      final BLSSignature signature,
+      final UInt64 index) {
+    this.pubKey = pubKey;
+    this.withdrawalCredentials = withdrawalCredentials;
+    this.amount = amount;
+    this.signature = signature;
+    this.index = index;
+  }
+
+  public static Deposit readFrom(final Bytes rlpBytes) {
```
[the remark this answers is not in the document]
naviechan: Yea Mikhail and I are aware of the change. Our conclusion is to implement this EIP as is, and will change `Deposit` to SSZ when implementing EIP-6465. I think it's simpler this way.

datatypes/src/main/java/org/hyperledger/besu/datatypes/BLSPublicKey.java
```diff
@@ -0,0 +1,83 @@
...
+import org.hyperledger.besu.ethereum.rlp.RLPException;
+import org.hyperledger.besu.ethereum.rlp.RLPInput;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.DelegatingBytes;
+
+/** A BLS public key. */
+public class BLSPublicKey extends DelegatingBytes
+    implements org.hyperledger.besu.plugin.data.PublicKey {
```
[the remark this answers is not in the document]
naviechan: Yea I was debating whether to put it under `datatypes` or `crypto`. My line of thought is `crypto` is more dedicated to classes that need crypto-related operations such as encryption and such. `BLSPublicKey` is mainly a CL thing, all we are doing in EL is just view this as a 48-byte field and we don't perform any crypto operation for `BLSPublicKey` hence putting it under `datatypes`. I am happy to move it to `crypto` if you think otherwise.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -194,6 +203,18 @@ protected BlockCreationResult createBlock(
 
       throwIfStopped();
 
+      final Optional<DepositsProcessor> maybeDepositsProcessor =
```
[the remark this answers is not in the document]
naviechan: You are right, I will remove `DepositsProcessor`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
naviechan: Yes the detail of the deposit validation will be included in the next PR. For this PR I will remove any non-trivial validation on the deposit to avoid confusion

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -200,7 +201,9 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
 
     final var block =
         new Block(
-            newBlockHeader, new BlockBody(transactions, Collections.emptyList(), maybeWithdrawals));
+            newBlockHeader,
+            new BlockBody(
+                transactions, Collections.emptyList(), maybeWithdrawals, Optional.empty()));
```
siladu, line 206: I was expecting deposits from newPayload's ExecutionPayloadV6110 to be wired in here, is there a reason it's not?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
```diff
@@ -177,7 +177,8 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
                     payloadAttributes.getTimestamp(),
                     payloadAttributes.getPrevRandao(),
                     payloadAttributes.getSuggestedFeeRecipient(),
-                    withdrawals));
+                    withdrawals,
+                    Optional.empty()));
```
siladu, line 181: Do we need to pass deposits in here? If we're preparing a payload (creating a block) then won't we parse the transactions and create the deposit objects ourselves later on in the block creation code?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadResultV6110.java
```diff
@@ -0,0 +1,195 @@
...
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import org.apache.tuweni.bytes.Bytes32;
+
+@JsonPropertyOrder({
+  "executionPayload",
+  "blockValue",
+})
+public class EngineGetPayloadResultV6110 {
```
siladu, line 36: We might not need a new class for this since you could technically use EngineGetPayloadResultV2 and add optional deposits to that (which would always be empty for shanghai fork for example).

The problem then is EngineGetPayloadResultV2 is a bad name since it really represents EngineGetPayloadResultV2+. I need to think about this :)

The only reason we needed EngineGetPayloadResultV2 to distinguish from EngineGetPayloadResultV1 was due to the change in top-level json structure with the addition of blockValue.

siladu (commented):
Assuming you are leaving the deposit validation and deposit creation (when creating a block) to another PR, then this is close to being ready. Feel free to mark as ready for review once you have all the CI checks passing.

You will also need to add at least some unit tests to cover the new logic, such as DepositDecoder/Encoder

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadResultV6110.java
naviechan: Yea I was thinking about this as well. The `execution-apis` repo says the name of the payload will be `ExecutionPayloadV6110`: https://github.com/mkalinin/execution-apis/blob/eip6110/src/engine/experimental/eip6110.md#executionpayloadv6110. Of course from the engineering perspective we don’t have to follow the spec 100%

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
naviechan: Json-rpc is not really in the scope of this PR and I am trying to keep the changes in `api` as minimal as possible. Will wire it in the following PR

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadResultV6110.java
siladu: Indeed, ExecutionPayload and EngineGetPayloadResult are not quite equivalent since it's the outer wrapper, but I can understand the association. 
I would have gone with a similar name if I wasn't already thinking about how to remove the version names from these classes entirely :)
This name is fine for now anyway.

siladu (commented):
Thanks for adding the extra tests 👍 

The main things to resolve for me are:
1. the new class hierachy for the Deposit fields, I will ask around and try to get a second opinion
2. The question about including the RLP changes to the Block domain classes - I would recommend we go with the testing approach.

I can help deploy and regression test this on some real nodes if you want, it's probably a good idea regardless since the changes are quite extensive.

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockBody.java
```diff
@@ -46,4 +46,11 @@ public interface BlockBody {
    * @return The list of withdrawals of the block.
    */
   Optional<? extends List<? extends Withdrawal>> getWithdrawals();
+
+  /**
+   * Returns the list of deposits of the block.
+   *
+   * @return The list of deposits of the block.
+   */
+  Optional<? extends List<? extends Deposit>> getDeposits();
```
shemnon, line 55: Until this EIP is shipped on mainnet we should add the `@Unstable` annotation, implying it may change or go away without warning.  For all new fields in the plugin-api package.

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/Deposit.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright Hyperledger Besu Contributors.
```
shemnon, line 2: New files should be `Copyright contributors to Hyperledger Besu`

https://wiki.hyperledger.org/display/TSC/Copyright+and+License+Policy

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
```diff
@@ -62,6 +62,7 @@ public void writeTo(final RLPOutput out) {
     out.writeList(body.getTransactions(), Transaction::writeTo);
     out.writeList(body.getOmmers(), BlockHeader::writeTo);
     body.getWithdrawals().ifPresent(withdrawals -> out.writeList(withdrawals, Withdrawal::writeTo));
+    body.getDeposits().ifPresent(deposits -> out.writeList(deposits, Deposit::writeTo));
```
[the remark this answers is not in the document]
naviechan: I am a bit new to testing so I would be great if I can get some help from you and to utilize Besu's infrastructure! I am leaning towards having the depositing wiring in this PR but I can see how it's easier to test everything at once by having it in the next PR instead

datatypes/src/main/java/org/hyperledger/besu/datatypes/DepositWithdrawalCredential.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Creates an withdrawal credential from the given RLP-encoded input.
+   *
+   * @param input The input to read from
+   * @return the input's corresponding withdrawal credential
+   */
+  public static DepositWithdrawalCredential readFrom(final RLPInput input) {
+    final Bytes bytes = input.readBytes();
```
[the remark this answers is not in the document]
naviechan: Yes `DepositEncoderTest.shouldEncodeDeposit()` should cover the RLP operations for each field in `Deposit`.

datatypes/src/main/java/org/hyperledger/besu/datatypes/BLSPublicKey.java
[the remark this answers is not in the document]
jframe: The datatypes package does seem like a more appropriate place than crypto as it is more focused on operations.

However, the intent of the datatype module is for broad-based types. So I don't think the `DepositWithdrawalCredential` would qualify and should probably just be a Bytes32 since we only care about bytes anyway or move it closer to where it is needed rather than in the top-level module.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/DepositParameter.java
```diff
@@ -0,0 +1,144 @@
...
+  private final String withdrawalCredentials;
+  private final String amount;
+
+  private final String signature;
+  private final String index;
+
+  @JsonCreator
+  public DepositParameter(
+      @JsonProperty("pubKey") final String pubKey,
+      @JsonProperty("withdrawalCredentials") final String withdrawalCredentials,
```
jframe, line 44: In the spec for [DepostiV1]( https://github.com/mkalinin/execution-apis/blob/eip6110/src/engine/experimental/eip6110.md#depositv1) the field is named "withdrawal_credentials"

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -237,6 +241,9 @@ public void writeTo(final RLPOutput out) {
     if (withdrawalsRoot != null) {
       out.writeBytes(withdrawalsRoot);
     }
+    if (depositsRoot != null) {
```
jframe, line 244: This should be appended to the end of the list

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -264,6 +271,8 @@ public static BlockHeader readFrom(
     final Wei baseFee = !input.isEndOfCurrentList() ? Wei.of(input.readUInt256Scalar()) : null;
     final Hash withdrawalHashRoot =
         !input.isEndOfCurrentList() ? Hash.wrap(input.readBytes32()) : null;
+    final Hash depositHashRoot =
```
jframe, line 274: Similar to above we need to read this as the last item in the header RLP list

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeBlockCreator.java
```diff
@@ -82,17 +83,20 @@ public MergeBlockCreator(
...
+   * @param deposits optional list of deposits
    * @return the block creation result
    */
   public BlockCreationResult createBlock(
       final Optional<List<Transaction>> maybeTransactions,
       final Bytes32 random,
       final long timestamp,
-      final Optional<List<Withdrawal>> withdrawals) {
+      final Optional<List<Withdrawal>> withdrawals,
+      final Optional<List<Deposit>> deposits) {
```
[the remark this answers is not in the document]
naviechan: If I am not wrong I think in `AbstractBlockCreator.createBlock()` I will get the deposits from `transactionResults` and stuff them into the block header and body. But this involves ABI decoding which will be included in the next PR so I am gonna leave the deposits to be empty and add a TODO here for now

datatypes/src/main/java/org/hyperledger/besu/datatypes/BLSPublicKey.java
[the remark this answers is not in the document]
naviechan: > The datatypes package does seem like a more appropriate place than crypto as it is more focused on operations.
> 
> However, the intent of the datatype module is for broad-based types. So I don't think the `DepositWithdrawalCredential` would qualify and should probably just be a Bytes32 since we only care about bytes anyway or move it closer to where it is needed rather than in the top-level module.

I agree we should just remove `DepositWithdrawalCredential` and make the field as Bytes32. What do you think about `BLSPublicKey` and `BLSSignature`? AFAIK we are not doing anything special for these 3 fields as we only care about the bytes. Having dedicated data classes are kind of like syntactic sugar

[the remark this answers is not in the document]
siladu: I think BLSPublicKey and BLSSignature are OK as new types since they are quite a general concept and could imagine them being reused in the future. 
Whereas DepositWithdrawalsCredential is quite specific Bytes32 is better IMO...it's not clear to me from the EIP what format this credential takes exactly, maybe there's an existing data type that is appropriate?

Possibly BLSPublicKey and BLSSignature shouldn't live in data types though since they are not broadly used. Maybe should live alongside Deposit.java for now?

siladu (commented):
Thanks for your small commits and clear commit messages :)

I am pretty happy with this, only blocker is the BodyValidator test which I'll see if I can help you with.

Will give @jframe a chance to re-review as well.

datatypes/src/main/java/org/hyperledger/besu/datatypes/BLSPublicKey.java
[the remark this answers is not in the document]
naviechan: > I think BLSPublicKey and BLSSignature are OK as new types since they are quite a general concept and could imagine them being reused in the future. Whereas DepositWithdrawalsCredential is quite specific Bytes32 is better IMO...it's not clear to me from the EIP what format this credential takes exactly, maybe there's an existing data type that is appropriate?
The format of withdrawal credential is one of its kind and is rather specific. To summarize what it contains: there are two types one with 0x00 and one with 0x01 prefix. The first type is a hash of public key with the first byte being replaced by 0. The second type is simply a 20-byte eth1 address. I think there are some discussions in CL on coming up with new prefixes and new encoding for the withdrawal credential. So in short, I don't think we can use existing data type for it.
> 
> Possibly BLSPublicKey and BLSSignature shouldn't live in data types though since they are not broadly used. Maybe should live alongside Deposit.java for now?
Yea I agree. We can move them to Deposit.

jframe (commented):
Happy with the PR now too. Agree with @siladu the only thing missing is a test for the depositRoot in the BodyValidationTest

siladu:
As a regression test, I span up 1x sepolia (snap sync), 1x goerli and 2x mainnet nodes which all got into sync and have stayed in sync for a number of days

Screenshot 2023-03-28 at 5 13 41 pm

PR #5068, opened 2023-02-08
Fix manifest docker not skipping interim builds for RCs

jframe:

## PR description
Docker manifest build was only considering snapshots to skip the major number tags e.g. 23.1. It should have also been including RCs to align with the docker upload task.

## Documentation

## Changelog


siladu (commented):
Can we extract this into its own function so it's one list?

jframe:
> Can we extract this into its own function so it's one list?

Probably, it should be possible somehow. Will give it ago

siladu (approved):
LGTM

PR #5071, opened 2023-02-08
4844: read blobs as raw bytes not as unit256 and fix blob versioned hashes validation

fab-10:

## Documentation

## Changelog


evm/src/main/java/org/hyperledger/besu/evm/precompile/KZGPointEvalPrecompiledContract.java
```diff
@@ -49,53 +52,55 @@ public KZGPointEvalPrecompiledContract() {
...
+      if (pathToTrustedSetup.isPresent()) {
+        Path pathToSetup = pathToTrustedSetup.get();
+        absolutePathToSetup = pathToSetup.toAbsolutePath().toString();
       } else {
-        throw new IllegalArgumentException("provided file not a setup for either 4 or 4096 bits");
+        InputStream is =
+            KZGPointEvalPrecompiledContract.class.getResourceAsStream(
+                "mainnet_kzg_trusted_setup_4096.txt");
+        try {
+          File jniWillLoadFrom = File.createTempFile("kzgTrustedSetup", "txt");
```
jflo, line 66:
```suggestion
          File jniWillLoadFrom = File.createTempFile("kzgTrustedSetup", ".txt");
```

jflo (approved):
minor suggestion, looks good otherwise.

jflo:
Unit tests are failing and it is due to a coredump happening when the new test tries to load the new trusted setup file. If I try to run that test with a known good setup file, it does not coredump.

jflo (changes requested):
Unit tests are failing and it is due to a coredump happening when the new test tries to load the new trusted setup file. If I try to run that test with a known good setup file, it does not coredump.

gezero (approved):

fab-10:
> Unit tests are failing and it is due to a coredump happening when the new test tries to load the new trusted setup file. If I try to run that test with a known good setup file, it does not coredump.

I did some debug and found that the c-kzg lib does not support mixing `MINIMAL` and `MAINNET` preset in the same process, since each one of them as a standalone native lib, so the test loads first the minimal lib, and when we try to load a mainnet trusted setup used by `aggregateKzgProof` it fails.
I do not know if there is a workaround for this, so for now I am putting ignore on this test

PR #5076, opened 2023-02-09
Fix Layered World State issue

ahamlat:

## PR description

We found the problem :

During a remember block we added a copy of the worldstate during the saveTrieLog in order to fix the get balance == 0. By making a copy we have a BonsaiInMemoryWorldstate which does not persist the trielog in rocksdb. We saw the log that tells us that we can’t commit. 
The problem is that during a FCU when we persist a block if we have a reorg we finally save the trielog after the different Rollback / Forward. Because of that we do not save a valid log. Instead of saving a Trielog for 1 to 2 bis we have a Trielog which is a mix of 2-1 + 1-2bis . This happens because it is not normal  to not already have the trielog in base during a reorg.

Because of that we can't rollback/rollfoward anymore with this invalid trielog so we have the layered copy worldstate error

The worldstate copy is not really needed and we decided to not create a copy and to subscribe to the worldstateStorage directly in the BonsaiLayeredWorldstate.

This fixed was not fixing yet the problem because we had another problem. During a getAccount we are getting the BonsaiLayeredWorldstate and we are closing this one after the getAccount. Because of that we are unsubscribing the worldstateStorage and losing the instance. to fix that we decided to use the same code we re using for the validate/transaction pool etc and to create a copy of the layeredWorldstate instead 

## Documentation

## Changelog


ahamlat:
I can confirm that it is working, I tested getBalance on latest several times and with a specific block. We also managed to run  two nodes, one with the fix and other one with [commit e715010](http://e7150102ea208165ac29bd10986fe26bc9e176d0/). After several hours, the reorg happened, the node with the fix (this PR) was able to handle it, while the other node crashed " Unable to copy Layered Worldstate for 0x58c372dc871df277ad86bdfdc3aa623b75d12c6d04718e6835b7c0e4fd4854cc" error.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiLayeredWorldState.java
```diff
@@ -76,7 +81,34 @@ public Optional<BonsaiWorldView> getNextWorldView() {
   }
 
   public void setNextWorldView(final Optional<BonsaiWorldView> nextWorldView) {
+    maybeUnSubscribe(); // unsubscribe the old view
     this.nextWorldView = nextWorldView;
+    maybeSubscribe(); // subscribe the next view
```
garyschulte, line 86: 👍

garyschulte (approved):
🙏

Issue #5061, opened 2023-02-07
Add check for block height when not doing PoS


pinges:
### Description
PR #3911 removed the check, because in a PoS environment the height of a peer is not updated any more. The removal of the check can lead to request that can not be satisfied by the peer, because they have not reached the necessary block height to deliver the requested header/block, which results in an empty response from that peer. This can lead to disconnection from that peer, when 5 useless responses are recorded in within 1 minute (see class PeerReputation).

The checks for the block height should be reinstated for PoW and PoA networks.

PR #5083, opened 2023-02-10
reintroduce checking of block height for certain tasks when we are not PoS (Revert PR#3911)

pinges:
fixes issue #5061


fab-10 (commented):
Since I think that info if the network is running on PoS should be in ProtocolSpec, I did a [PR](https://github.com/hyperledger/besu/pull/5085) for that, note that it is not yet complete, since the revert of some minimunRequiredBlockNumber has still to be done, but it should give an idea.
Many files are modified only to pass new constructor params, so the main ones to check are: 
EthPeers, ProtocolSpec and MainnetProtocolSpecs

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncStep.java
```diff
@@ -69,13 +69,20 @@ protected CompletableFuture<List<BlockHeader>> requestHeaders(final Hash hash) {
...
+            .getFinalized()
+            .flatMap(context.getProtocolContext().getBlockchain()::getBlockHeader);
 
     final RetryingGetHeadersEndingAtFromPeerByHashTask
         retryingGetHeadersEndingAtFromPeerByHashTask =
             RetryingGetHeadersEndingAtFromPeerByHashTask.endingAtHash(
                 context.getProtocolSchedule(),
                 context.getEthContext(),
                 hash,
+                maybeFinalizedHeader.map(BlockHeader::getNumber).orElse(0L),
```
fab-10, line 85: No need to revert this part, since backward sync is always on PoS, so you can statically set 0 here

fab-10 (changes requested):
looks good, only some code to remove since no more useful

macfarla (approved):
LGTM

PR #5084, opened 2023-02-10
Initialize KZG native lib at startup if Cancun is enabled

fab-10:

## PR description

Move c-kzg native lib initialization at startup, when all the other native lib are configured, and also because it is not only used by the precompile, but also for transaction validation.
Added the option `--kzg-trusted-setup` to pass a custom setup file for custom networks or to override the default one for named networks

## Documentation

## Changelog


evm/src/main/java/org/hyperledger/besu/evm/precompile/KZGPointEvalPrecompiledContract.java
```diff
@@ -37,52 +31,53 @@
 public class KZGPointEvalPrecompiledContract implements PrecompiledContract {
   private static final AtomicBoolean loaded = new AtomicBoolean(false);
 
-  private final Bytes successResult;
+  private static final Bytes successResult;
+
+  static {
+    CKZG4844JNI.loadNativeLibrary(CKZG4844JNI.Preset.MAINNET);
```
shemnon, line 37: Please move this initialization to a lazy or on-demand initialization.  

This will incur a large startup penalty on _every_ CLI call that touches the EVM.  This imposes a debilitating penalty on the EVM Tool and will make using Besu in reference tests impractical.

shemnon (changes requested):
static class initialization of the mainnet KZG is a hard stop for me.  It should be loaded lazily or on-demand.

evm/src/main/java/org/hyperledger/besu/evm/precompile/KZGPointEvalPrecompiledContract.java
fab-10: make sense, will move it in the `init` methods

shemnon (approved):
Confirmed that this eliminates the ~2s startup hit I've seen.

PR #5088, opened 2023-02-10
Add shanghaiTime to sepolia

siladu:
Announced here: https://github.com/ethereum/execution-specs/pull/716

Move timestamp forks from getForkBlockNumbers to getForkBlockTimestamps in JsonGenesisConfigOptions - this ultimately gets used to popoulate the ForkIdManager which handles lists of blocks and timestamps the same way so this hasn't changed any actual behaviour, but rather supports the test fixes.

Implement TransitionProtocolSchedule.streamMilestoneBlocks as a concatenation of blockNumbers++blockTimestamps. This may have been a latent bug since it's used to update the node record when a fork transition occurs.


Gabriel-Trintinalia (approved):

jframe (approved):
LGTM

jframe (approved):

PR #5096, opened 2023-02-15
updates to latest version of kzg wrapper

jflo:
Sadly this does not get us any finer grained error reasons, and that seems intentional.


gradle/verification-metadata.xml
```diff
@@ -5614,6 +5614,17 @@
             <sha256 value="8314ff61776036dacab235bf0a225d4ef901e1555a86f91977e162780d3bbe54" origin="Generated by Gradle"/>
          </artifact>
       </component>
+      <component group="tech.pegasys" name="jc-kzg-4844" version="0.3.0">
```
macfarla, line 5617: do we need a process with this verification metadata generation to remove the old metadata eg
`       <component group="tech.pegasys" name="jc-kzg-4844" version="0.1.0">`

macfarla (approved):
LGTM

evm/src/main/java/org/hyperledger/besu/evm/precompile/KZGPointEvalPrecompiledContract.java
```diff
@@ -134,8 +135,16 @@ public PrecompileContractResult computePrecompile(
                 Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
       }
       return result;
+    } catch (CKZGException ckzgException) {
```
fab-10, line 138: This seems like debug code that should be reverted right? also remove the System.out.println

fab-10 (commented):
I think there is some debug code to be removed, and since I need to update the lib again with a new release soon, we can skip this update.

PR #5100, opened 2023-02-16
Fix block value calculation

### LINKED ISSUE #5040: Block value calculation may be incorrect in getPayloadV2
Some WIP hive tests have picked up an issue...

> Besu is returning value, e.g. in one of the blocks, of 54089155500000000, but the expected value by the test (and returned by other clients) is 6224737500000000, which 8 times lower, and this is because, during the calculation, the "cumulative gas used" of the transaction is used to get the fee, instead of the "gas used"
> If i swap the "gas used" by "cumulative gas used" while computing the expected value in the test, I get the exact same value as Besu

gfukushima:
## PR description
This PR fix the block value calculation mechanism by using gasUsed instead of the cumulativeGas to calculate the block value.

This change has been tested against the hive test(GetPayloadV2 Block Value) for this feature.

## Fixed Issue(s)

Fixes #5040
## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
@@ -28,7 +28,7 @@ public class BlockValueCalculator {
     for (int i = 0; i < txs.size(); i++) {
       final Wei minerFee = txs.get(i).getEffectivePriorityFeePerGas(block.getHeader().getBaseFee());
       // we don't store gasUsed and need to calculate that on the fly
-      // receipts are fetched in ascendant sorted by cumulativeGasUsed
+      // receipts are fetched in ascending sorted by cumulativeGasUsed
       long gasUsed = receipts.get(i).getCumulativeGasUsed();
       if (i > 0) {
         gasUsed = gasUsed - receipts.get(i - 1).getCumulativeGasUsed();
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
@@ -27,7 +27,13 @@ public class BlockValueCalculator {
     Wei totalFee = Wei.ZERO;
     for (int i = 0; i < txs.size(); i++) {
       final Wei minerFee = txs.get(i).getEffectivePriorityFeePerGas(block.getHeader().getBaseFee());
-      totalFee = totalFee.add(minerFee.multiply(receipts.get(i).getCumulativeGasUsed()));
+      // we don't store gasUsed and need to calculate that on the fly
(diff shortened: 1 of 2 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
```diff
@@ -27,7 +27,13 @@ public Wei calculateBlockValue(final BlockWithReceipts blockWithReceipts) {
     Wei totalFee = Wei.ZERO;
     for (int i = 0; i < txs.size(); i++) {
       final Wei minerFee = txs.get(i).getEffectivePriorityFeePerGas(block.getHeader().getBaseFee());
-      totalFee = totalFee.add(minerFee.multiply(receipts.get(i).getCumulativeGasUsed()));
+      // we don't store gasUsed and need to calculate that on the fly
+      // receipts are fetched in ascendant sorted by cumulativeGasUsed
```
siladu, line 31:
```suggestion
      // we don't store gasUsed per transaction and need to calculate that on the fly
      // receipts are fetched in ascending sorted by cumulativeGasUsed
```

siladu (approved):
LGTM!

I was wondering if we could/should store gasUsed in an object somewhere, but I don't think it's worth it given that getPayload is called relatively rarely.

jframe (dismissed):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
gfukushima: Had the same thought, did a bit of digging into the code, and even the BlockchainQuery does this same calculation to create instances of TransactionReceiptWithMetadata

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockValueCalculator.java
```diff
@@ -27,7 +27,13 @@ public Wei calculateBlockValue(final BlockWithReceipts blockWithReceipts) {
     Wei totalFee = Wei.ZERO;
     for (int i = 0; i < txs.size(); i++) {
       final Wei minerFee = txs.get(i).getEffectivePriorityFeePerGas(block.getHeader().getBaseFee());
-      totalFee = totalFee.add(minerFee.multiply(receipts.get(i).getCumulativeGasUsed()));
+      // we don't store gasUsed and need to calculate that on the fly
```
jframe, line 30: Can the TransactionWithMetadata type be used here? If we can use that then there is getGasUsed method that can be used instead to avoid duplicating this logic. Should be able to use the BlockchainQueries.blockByHash from the EngineGetPayloadV2 method I think.

jframe: After some discussion, the BlockchainQueries can't be used as these blocks aren't appended yet. So this approach isn't possible

jframe (approved):
LGTM

PR #5103, opened 2023-02-16
trigger heal in case of bad code

matkt:
## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateUpdater.java
```diff
@@ -356,7 +356,12 @@ public void commit() {
   public Optional<Bytes> getCode(final Address address, final Hash codeHash) {
     final BonsaiValue<Bytes> localCode = codeToUpdate.get(address);
     if (localCode == null) {
-      return wrappedWorldView().getCode(address, codeHash);
+      final Optional<Bytes> code = wrappedWorldView().getCode(address, codeHash);
+      if (code.isEmpty() && !codeHash.equals(Hash.EMPTY)) {
```
garyschulte, line 360: should this instead hash code and compare it to codeHash rather than just checking empty?  or is the hashing prohibitively expensive

matkt: Yes this is enough because the hash check is already done at the storage level. And if it is not good we will have an Empty. So I prefer to avoid adding an unnecessary extra shot for this check

garyschulte (approved):
🚢

PR #4540, opened 2022-10-16
Unroll Untraced Execution

shemnon:
## PR description

Add a second run-to-halt path used when there is no tracing.  This path calls static final versions of select operations with the aim of avoiding the megamorphic dispatch of the original loop.

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #5116, opened 2023-02-21
Fixes different gas consumption by Besu when compared to Geth

### LINKED ISSUE #4244: Wrong estimated value when using estimateGas API
### Description
When we execute `estimateGas` API on a smart contract call, it seems that Besu returns a wrong estimated value.

### Steps to Reproduce (Bug)
1. Download the following file and untar it:
[test_gas_estimation.tar.gz](https://github.com/hyperledger/besu/files/9295689/test_gas_estimation.tar.gz)
2. Execute Besu binary with the following flags:
`besu --data-path test_gas_estimation/data-besu --genesis-file test_gas_estimation/besu.json --rpc-http-port 8549 --rpc-http-enabled --rpc-http-api ETH --network-id 15 --discovery-enabled false`
3. Execute the test case file: `node test1.js`

**Expected behavior:**
All other client implementations return 34165.

**Actual behavior:** 
Only Besu returns 36465.

**Frequency:**
Always

* Software version: besu/v22.4.4/linux-x86_64/openjdk-java-16
* OS Name & Version: Ubuntu 20.04.4 LTS
* Node.js / web3.js versions: v16.15.0 / v1.7.3

### Smart contract information (If you're reporting an issue arising from deploying or calling a smart contract, please supply related information)
* Solidity version: v0.8.13
* The contract code is provided in the tar.gz file above.
* The contract is deployed at the following address: 0xAe7405782dcf9Fc759020045E7Acc0795d372bc8

NickSneo:
## PR description
In PR #4540 some Static OPCODEs were commented out, which led to the wrong gas calculation/consumption by Besu in comparison to other Ethereum Clients like Geth.

In this PR I have un-commented those OPCODEs and tested the build, now the actual tx gas consumption is same as that by Geth. Even eth_estimateGas returns the correct values as expected i.e actual gas consumption + 2300 (gas stipend and  63/64 rule).

## Fixed Issue(s)
Fixes #4244 

## Documentation

## Changelog


shemnon:
These particular opcodes were commented out for EVM speed on an experimental basis (A/B testing with and without them in place). So they are commented out for a purpose.  However I do not expect they would impact gas calculations.

I am curious how those would have resulted in differing gas prices.  Is this just for eth_estimateGas?  Are there any other test cases than the one outline in #4244?

NickSneo:
> 

Correct me if I am wrong, commenting those OPCODE will impact gas calculations. I am taking an example of static OPCODE "SUB" which was commented. So in the switch case if current operation is SUB but since it is commented it will go to default case, which is just for running dynamic OPCODE in which it call the function execute on current operation and that is not available on SUB (function available here is executeFixedCostOperation).

It is not just for eth_estimateGas but for actual gas consumption too. 
I have tested it on 3 smart contracts using remix, Besu 22.10.3, Besu 22.7.7, Besu fixed, and Geth 1.11.1

1. Empty Contract
2. HelloWorld smart Contract 
3. ERC20 smart contract

My findings - 

Geth -
Empty deploy - 67066
HelloWorld deploy - 309861
Set function HelloWorld - 45007
ERC20 deploy - 875914
mint function on ERC20 - 1st time - 68353 2nd time - 34153

Besu 22.10.3 -
Empty deploy - 71538
HelloWorld deploy - 367425
Set function HelloWorld - 43679
ERC20 deploy - 1085338
mint function on ERC20 - 1st time - 64813 2nd time - 34813

Besu 22.7.7 -
Empty deploy - 71538
HelloWorld deploy - 367425
Set function HelloWorld - 43679
ERC20 deploy - 1085338
mint function on ERC20 - 1st time - 64813 2nd time - 34813

Besu Modified (with fix) -
Empty deploy - 67066
HelloWorld deploy - 309861
Set function HelloWorld - 45007
ERC20 deploy - 875914
mint function on ERC20 - 1st time - 68353 2nd time - 34153

shemnon:
What changes is that the opcodes are evaluated out of line, not inline.  The default case is on line 305/312 and that goes into an array of objects into a megaporphic virtual call instead of an unrolled loop.

Can you post something like a refernce test that exhibits this behavior?  Something not in eth_estimateGas, which is (a) an estimate and (b) iterative and (c) only gives a "close" answer by design.  possibly eth_call or a block that is breaking consensus?  I think what is being observed is performance changes to the eth_estimateGas done for performance.

shemnon (changes requested):
We need a test case a part of this PR.

NickSneo:
> I guess my main question is how do you measure "actual gas consumption" so we are instrumented the same way.

The actual gas consumption value is recovered from transaction receipt. 
I have created some test scripts which you can run on Besu test dev network and Geth test dev network. The script will run and display estimated gas and actual gas consumed for deploying ERC20 contract and calling the mint function.

[scripts.zip](https://github.com/hyperledger/besu/files/10837457/scripts.zip)

Start the Besu/Geth test dev network
Extract the zip file, install web3js and run the script using -
`node <besu-or-geth>-script.js`

NickSneo:
> We need a test case a part of this PR.

Adding a test case for this.

NickSneo:
Hey @shemnon ,

I added an integration test and checked the results from both original and modified Besu code and they gave me the correct gas consumption values.

Seems the Besu version 22.10.3 had this issue of wrong gas consumption which got fixed in version 23.1.0.
And as you said commented or uncommented OPCODEs doesn't have any impact on gas consumption.

We can close this PR, as it is not required.

PR #5123, opened 2023-02-22
Bonsai cleaning and robustness

matkt:
## PR description

The idea of this refactor is just to make the code more readable and less dangerous to maintain and update. 
#### Current status

Problems :

its number of Worldstate (InMemory, Layered, Snapshot, Persisted). This made the code very unreadable and complicated to update.

the worldstate copy mechanism which meant that we had to use this code everywhere and we had to make copies of copies which could get us lost and prevent us from knowing what we had at the end (InMemory, Snapshot, Persisted? Layered )?
```
try (final var worldState =
            worldStateArchive
                    .getMutable(blockHeader.getStateRoot(), blockHeader.getHash(), false)
                    .map(
                            ws -> {
                              if (!ws.isPersistable()) {
                                return ws.copy();
                              }
                              return ws;
                            })
                    .orElse(null)) {
 if (ws != null) {
   return Optional.ofNullable(mapper.apply(ws));
   if (worldState != null) {
     return Optional.ofNullable(mapper.apply(worldState));
   }
 }
```
 the multiple layers of subscribers. Because of these copies, we could have a copy that subscribes to another copy that itself subscribes to a snapshot in the cache etc. It became complicated to find the logic to properly unsubscribe paved the way for many regressions when modifying codes

the fact of creating a snapshot on a persistent state that may be in a transient state

Class naming 

#### Proposition

The proposal is to delete all the different worldstates to have only one worldstate. The difference will reside in the type of storage that we have
- Snapshot storage
- Layered storage
- Persisted storage

	a worldstate can be frozen or not, which means that it will not try to modify its storage. If it is frozen it will be faster to calculate the root hash because it will have fewer steps

> Layered storage for block 3 -> Layered storage for block 2 -> Snapshot of block 1

We will also have the same logic for all the worldstates and we will not need to duplicate the code or try to understand what a worldstate does or what a worldstate does not do. they all do the same thing. What changes is the storage we use and is it frozen or not.

>BonsaiWorldstate(new BonsaiWorldStateKeyValueStorage()) -> Persisted state
>BonsaiWorldstate(new BonsaiSnapshotWorldStateKeyValueStorage()) -> Snapshot state

Another improvement is the simplification of the BonsaiWorldStateArchive which no longer needs to be complex. It will also be renamed to BonsaiWorldStateProvider to better understand its role.

The provider will just do some simple step : 

Do we want to persist our modifications ? 
- If Yes create a BonsaiWorldState with the persisted storage and rollback/forward if needed
- If No take a wrapped snapshot of the storage from the cache ), create a BonsaiWorldstate with it and rollback/forward if needed , add this new wrapped snapshot (called layered) in the cache and freeze the state

The snapshot of the persisted state is always added before it is needed, at the time of the fcu or at the start of Besu in order to never create a snapshot of a transient state

```
@Override
  public Optional<MutableWorldState> getMutable(
      final BlockHeader blockHeader, final boolean shouldPersistState) {
    if (shouldPersistState) {
      return getMutable(blockHeader.getStateRoot(), blockHeader.getHash());
    } else {
      final BlockHeader chainHeadBlockHeader = blockchain.getChainHeadHeader();
      if (chainHeadBlockHeader.getNumber() - blockHeader.getNumber()
          >= trieLogManager.getMaxLayersToLoad()) {
        LOG.warn(
            "Exceeded the limit of back layers that can be loaded ({})",
            trieLogManager.getMaxLayersToLoad());
        return Optional.empty();
      }
      return trieLogManager
          .getWorldState(blockHeader.getHash())
          .or(() -> trieLogManager.getNearestWorldState(blockHeader))
          .or(() -> trieLogManager.getHeadWorldState(blockchain::getBlockHeader))
          .flatMap(
              bonsaiWorldState ->
                  rollMutableStateToBlockHash(bonsaiWorldState, blockHeader.getHash()))
    }
  }
```
We no longer have an InMemory. The layered will be used for all cases where we do not want to change the persisted state. The trielogs will just be used for rollback or rollfoward but no more to find the status of an account or a storage

We will no longer need to copy etc. We store snapshot storage in the cache and when we need it we wrap it in a layered storage it and create a worldstate from that. This makes it possible to greatly simplify the subscribe mechanism to its minimum and no longer need to copy worldtstates.

Finally, the bonsai updater is renamed to accumulator in order to better explain its role

### Performance metrics
**Block Processing and FCU times**
Block performance and FCU times are at least similar to 23.1.0 release or even better on a 32 GiB VM.

On a 16 GiB/4 vCPU AWS VM (m6a.xlarge)
image

On a 32 GiB/8 vCPU AWS VM (m6a.2xlarge)
image

**CPU usage**
This PR fixes an issue we had on versions 23.1.0 and 23.1.1, which is the short CPU spikes. We can see on both 4vCPU and 8 vCPU, Besu doesn't have anymore the short CPU spikes we can see on version 23.1.0. 

On a 16 GiB/4 vCPU AWS VM (m6a.xlarge)
image

On a 32 GiB/8 vCPU AWS VM (m6a.2xlarge)
image

**Memory management**
Memory management is better with this PR than 23.1.0. We can see in the screenshots below that Besu is able to release more memory with PR during Garbage Collection and thus consume less memory at the process level (RSS memory). GC activity is pretty similar on this PR compared to 23.1.0.
I didn't notice any OutOfMemory error or OOM killer event both configurations (16 and 32 GiB VMs).

On a 16 GiB/4 vCPU AWS VM (m6a.xlarge) (-Xmx5g)
image

image

GC activity (PR #5123)
image

GC activity (version 23.1.0)
image

On a 32 GiB/8 vCPU AWS VM (m6a.2xlarge) (-Xmx8g)
image

image

GC activity (PR #5123)
image

GC activity (version 23.1.0)
image

**Disk and network IO**
Disk and Network IO is very close on both configurations

On a 16 GiB/4 vCPU AWS VM (m6a.xlarge) (-Xmx5g)
image

On a 32 GiB/8 vCPU AWS VM (m6a.2xlarge) (-Xmx8g)
image

**Performance of some ETH RPC Calls (2 active users, during 5 minutes), the time displayed below is in ms**

image

## Documentation

## Changelog


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/CachedSnapshotWorldstateManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.tuweni.bytes.Bytes32;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CachedSnapshotWorldstateManager extends AbstractTrieLogManager<BonsaiWorldState>
+    implements BonsaiWorldStateKeyValueStorage.BonsaiStorageSubscriber {
```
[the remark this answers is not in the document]
garyschulte: pushed a commit with subscriber

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
```diff
@@ -600,6 +623,8 @@ private void rollAccountChange(
       }
       if (replacementValue == null) {
         if (accountValue.getPrior() == null) {
+          // TODO: should we remove from the parent accumulated change also?  only if it is a
```
garyschulte, line 626: @matkt I believe this is a "no" in the current implementation and we can delete this comment.  Do you concur?

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/storage/LayeredKeyValueStorage.java
```diff
@@ -0,0 +1,141 @@
...
+          .filter(entry -> entry.getValue().length > 0)
+          .map(bytesEntry -> bytesEntry.getKey().toArrayUnsafe());
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  @Override
+  public boolean tryDelete(final byte[] key) {
+    // TODO: can we rely on zero byte array to indicate deletion?
```
garyschulte, line 110: @matkt I believe this is "yes", are there cases where we would legitimately want to add and store a zero byte array in key value storage rather than deleting it?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/cache/CachedWorldStorageManager.java
```diff
@@ -151,7 +151,8 @@ public Optional<BonsaiWorldState> getHeadWorldState(
               addCachedLayer(
                   blockHeader,
                   blockHeader.getStateRoot(),
-                  new BonsaiWorldState(archive, rootWorldStateStorage));
+                  new BonsaiWorldState(archive,
+                      new BonsaiSnapshotWorldStateKeyValueStorage(rootWorldStateStorage)));
```
garyschulte: @matkt IMO we should not add the persisted state to the cache directly, since it is mutable.  testing now with this tweak to ensure it does not re-introduce any leaks

garyschulte: no new snapshot leakage found during testing with this commit

ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/debug/debug_accountAt.json
```diff
@@ -6,16 +6,16 @@
     "params": [
       "0xc8df1f061abb4d0c107b2b1a794ade8780b3120e681f723fe55a7be586d95ba6",
       0,
-      "0xbcde5374fce5edbc8e2a8697c15331677e6ebf0b"
```
garyschulte, line 9: changed this test fixture temporarily to access an account available in the updater.

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/storage/LayeredKeyValueStorage.java
matkt: yes because otherwise we don't keep the info and we will get it from the parent that still have it. we must mark it as delete to avoid this

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
matkt: we don't have parent accumulator in the last implementation, we have only parent key value storage . So no

ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/debug/debug_accountAt.json
garyschulte: reverted

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiSnapshotWorldStateKeyValueStorage.java
```diff
@@ -0,0 +1,217 @@
...
+    if (!isClosed.get()) {
+      // alert any subscribers we are closing:
+      subscribers.forEach(BonsaiStorageSubscriber::onCloseStorage);
+
+      // close all of the SnappedKeyValueStorages:
+      accountStorage.close();
+      codeStorage.close();
+      storageStorage.close();
+      trieBranchStorage.close();
+
```
garyschulte, line 205: We are missing closing trieLogStorage here.  snapshots did not close trieLogStorage

matkt: Trielog is never a copy but it is always the real column so if we close we will close the real column. On the other hand we forgot it here  https://github.com/hyperledger/besu/pull/5123/files/f96ca374c433439980b64e3383901daffb1e98fe#diff-fa9b23a353ef5038c992c17a8cfe5e6210cc41369dada665044bc7633e136ab0R457

garyschulte: sorry - meant for this comment to be on BonsaiWorldStateKeyValueStorage.doClose() not the snapshot doClose()

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -414,7 +412,55 @@ public void rollback() {
...
+    if (!isClosed.get()) {
+      // alert any subscribers we are closing:
+      subscribers.forEach(BonsaiStorageSubscriber::onCloseStorage);
+
+      // close all of the KeyValueStorages:
+      accountStorage.close();
+      codeStorage.close();
+      storageStorage.close();
+      trieBranchStorage.close();
+
```
garyschulte, line 457: needs `trieLogStorage.close();`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -73,149 +74,107 @@ public BonsaiWorldStateArchive(
...
+              }
+            })
+        .map(WorldState.class::cast);
   }
 
   @Override
   public boolean isWorldStateAvailable(final Hash rootHash, final Hash blockHash) {
-    return trieLogManager.getBonsaiCachedWorldState(blockHash).isPresent()
+    return trieLogManager.containWorldStateStorage(blockHash)
         || persistedState.blockHash().equals(blockHash)
```
jflo, line 141: The difference between persistedState and worldStateStorage is still cryptic. would persistedState be better called "workingState" since it is the one "main" state being worked on and updated?

garyschulte: `persistedState` here is the one and only persisted BonsaiWorldState and `worldStateStorage` is its storage.  Since the worldstate storage is accessible via BonsaiWorldState, it is a bit redundant to have both as member variables.  it would be a longer conditional, but this could be rewritten as:

```
return trieLogManager.containWorldStateStorage(blockHash)
        || persistedState.blockHash().equals(blockHash)
        || persistedState.getWorldStateStorage().isWorldStateAvailable(rootHash, blockHash);
```

one could also argue that the second and third conditions are functionally equivalent if we want simplicity/clarity.

jflo (commented):
review still in progress

jflo (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -73,149 +74,107 @@ public BonsaiWorldStateArchive(
...
-        blockchain,
-        useSnapshots,
-        cachedMerkleTrieLoader);
+    // TODO: de-dup constructors
+    this.trieLogManager =
+        new CachedWorldStorageManager(
+            this, blockchain, worldStateStorage, maxLayersToLoad.orElse(RETAINED_LAYERS));
+    this.blockchain = blockchain;
+    this.worldStateStorage = worldStateStorage;
+    this.persistedState = new BonsaiWorldState(this, worldStateStorage);
```
siladu, line 92: Having a Java brain-fart moment...how can we reference `this` inside BonsaiWorldStateProvider constructor before BonsaiWorldStateProvider instance is fully initialised?

e.g. what if BonsaiWorldState constructor called `archive.getCachedMerkleTrieLoader()`?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
```diff
@@ -65,35 +70,36 @@ public class BonsaiWorldStateUpdater extends AbstractWorldUpdater<BonsaiWorldVie
   private final Map<Address, StorageConsumingMap<BonsaiValue<UInt256>>> storageToUpdate =
       new ConcurrentHashMap<>();
 
-  BonsaiWorldStateUpdater(final BonsaiWorldView world) {
-    this(world, (__, ___) -> {}, (__, ___) -> {});
```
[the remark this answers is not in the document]
siladu: I enjoyed reading that, reminds me of my scala days 😆

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
```diff
@@ -65,35 +70,36 @@ public class BonsaiWorldStateUpdater extends AbstractWorldUpdater<BonsaiWorldVie
...
-  }
+  private boolean isAccumulatorStateChanged;
```
siladu, line 73: Is this thread-safe?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueSnapshot.java
```diff
@@ -109,21 +116,26 @@ public void clear() {
...
-    snapTx.close();
+    if (closed.compareAndSet(false, true)) {
+      snapTx.close();
+    }
   }
 
-  @Override
-  public KeyValueStorageTransaction getSnapshotTransaction() {
-    return snapTx;
+  private void throwIfClosed() {
```
siladu, line 130: Erroring early, nice

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/InMemoryKeyValueStorage.java
siladu, line 1: I'm surprised this class hasn't been removed or renamed...is it tied to FOREST as well?

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
siladu, line 1: Why doesn't this live in the bonsai package?

siladu (commented):
**Love** the fact this deletes more code than it adds: `+1,951 −3,009`

I see that quite a few unit tests are modified and redundant test code deleted which is great. I may have missed it but I was expecting to see some new tests that covers the new functionality.

I ran coverage for `bonsai` package and it seems quite low, which concerns me, particularly for the potential of future regressions.

Screenshot 2023-03-17 at 4 15 10 pm

matkt:
> **Love** the fact this deletes more code than it adds: `+1,951 −3,009`
> 
> I see that quite a few unit tests are modified and redundant test code deleted which is great. I may have missed it but I was expecting to see some new tests that covers the new functionality.
> 
> I ran coverage for `bonsai` package and it seems quite low, which concerns me, particularly for the potential of future regressions.
> 
> Screenshot 2023-03-17 at 4 15 10 pm

For this it is planned to make a pass to think about testing bonsai correctly. We have planned a next step to build tests that would be good quality tests and not just for a  good coverage. We chatted with @garyschulte and @non-fungible-nelson about that. I don't think that just unit tests will be enough to avoid regressions, more complex tests are needed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -264,47 +223,54 @@ Optional<MutableWorldState> rollMutableStateToBlockHash(
...
+                  + mutableState.getWorldStateStorage().getClass().getSimpleName()
+                  + " for block hash "
+                  + blockHash,
+              e);
+
           return Optional.empty();
         }
       } catch (final RuntimeException re) {
-        LOG.trace("Archive rolling failed for block hash " + blockHash, re);
+        LOG.info("Archive rolling failed for block hash " + blockHash, re);
```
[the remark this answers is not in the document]
matkt: Indeed I think that Error would be more appropriate

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
matkt: It is not modified in a concurrent way so it's good

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
matkt: Because this class is generic and could even be theoretically used for Forest

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
garyschulte: It is a theoretical risk, but not currently one because the archive is the only user of that object until the archive is fully constructed. 

However,  I have never been fond of having an archive reference in worldstate.  In a future PR IMO we could/should create functional interface and/or function references for the bits of the archive functionality that a worldstate needs so we can break the circular archive<->worldstate dependency.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/cache/CachedBonsaiWorldView.java
```diff
@@ -0,0 +1,76 @@
...
+  public synchronized void close() {
+    worldStateStorage.unSubscribe(this.worldViewSubscriberId);
+    try {
+      worldStateStorage.close();
+    } catch (final Exception e) {
+      LOG.warn("Failed to close worldstate storage for block " + blockHeader.toLogString(), e);
+    }
+  }
+
+  public synchronized void updateWorldStateStorage(
```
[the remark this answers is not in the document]
garyschulte: Reason is/was because there is an implied chain of "parent" storages.  If we replaced the object in the cache we would have a "ripple" effect where we have to update all cache entries that represent descendents.  We wanted to minimize the need for locking and potential for concurrency issues.  

If we just replace the storage itself, there is no ripple and it can be a fairly cheap synchronized operation.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/AbstractTrieLogManager.java
```diff
@@ -86,52 +80,28 @@ public synchronized void saveTrieLog(
...
+    final TrieLogLayer trieLog = localUpdater.generateTrieLog(blockHeader.getBlockHash());
     trieLog.freeze();
-    addCachedLayer(
-        forBlockHeader, forWorldStateRootHash, trieLog, worldStateArchive, forWorldState);
-    scrubCachedLayers(forBlockHeader.getNumber());
     return trieLog;
   }
 
-  synchronized void scrubCachedLayers(final long newMaxHeight) {
+  public synchronized void scrubCachedLayers(final long newMaxHeight) {
```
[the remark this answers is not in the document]
garyschulte: will add a docs topic for evicting worldstates from cache :+1:

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldState.java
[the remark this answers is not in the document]
garyschulte: IDK the git history for it, but it is an amalgam of the now deleted Persisted, Snapshot and InMemory worldstate behaviors.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -387,4 +353,13 @@ public Optional<WorldStateProof> getAccountProof(
     // FIXME we can do proofs for layered tries and the persisted trie
     return Optional.empty();
   }
+
+  @Override
+  public void close() {
+    try {
+      worldStateStorage.close();
+    } catch (Exception e) {
+      // no op
```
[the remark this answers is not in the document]
garyschulte: As it stands presently, this close() won't throw unless there is a system level failure which we wouldn't be able to handle (such as an exception from rocksdb)

Karim and I have discussed a follow-on PR that implements a more robust shutdown behavior on exit.  The shutdown PR will modify this behavior and in the case of an exception on closing, we can/will at least log the cause of the failure.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
siladu: Might be nice to use the `@ThreadSafe` annotation to show it's been considered.

Issue #5064, opened 2023-02-08
Recent getBlockByNumber changes caused Goerli sync issue


siladu:
This commit caused an issue when syncing fresh goerli nodes: https://github.com/hyperledger/besu/commit/9ceebc4a57f648ed412ed6dcbc12a42699fffe54

`
{"@timestamp":"2023-02-07T21:49:39,513","level":"INFO","thread":"EthScheduler-Workers-0","class":"CoinbaseHeaderValidationRule","message":"Invalid block header: No clique in/out voting may occur on epoch blocks (7470000)","throwable":""}
{"@timestamp":"2023-02-07T21:49:44,311","level":"WARN","thread":"EthScheduler-Services-800 (downloadHeaders)","class":"PipelineChainDownloader","message":"Invalid block detected (BREACH_OF_PROTOCOL). Disconnecting from sync target. Header failed validation.: Invalid block at #7470000 (0xaaea1696007267a518ee5c6550742f58c81333e71cdee6c38dc2403b628816e1)","throwable":""}
`

The issue was experienced during 23.1.0-RC2 burn-in, specifically on four of these nodes after needing to resync due to another bug: https://github.com/ConsenSys/protocols-aws-inventories/blob/master/clc-goerli.yml

These were all BONSAI, mostly X_SNAP but one FAST.
I then span up a further three nodes to confirm and got the same error with FOREST (with X_SNAP).

The issue was **not experienced while syncing sepolia**: https://github.com/ConsenSys/protocols-aws-inventories/blob/master/clc-sepolia.yml

I am yet to confirm mainnet is unaffected as nodes are still syncing.

jframe:
This was commit reverted on main https://github.com/hyperledger/besu/pull/5063

siladu:
To save a full resync, I already have a node with a database that is synced to just before the issue will occur: 
*dev-besu-clc-teku-ohio-goerli-23.1.0-RC2-snap*

`{"@timestamp":"2023-02-08T04:04:07,623","level":"INFO","thread":"EthScheduler-Services-29 (importBlock)","class":"FastImportBlocksStep","message":"Block import progress: 7329800 of 8452736 (86%)","throwable":""}`

Issue will occur at block #7470000

iamhsk:
Related to #4789

PR #5020, opened 2023-01-30
Replace getByBlockNumber by getByBlockHeader

gfukushima:
## PR description
First batch that replaces getByBlockNumber by getByBlockHeader. These batch includes trivial replacements with none or minimal changes to the logic of the code base.

## Fixed Issue(s)

Partially fixes #4789

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

Issue #4789, opened 2022-12-08
Change getByBlockNumber to be getByBlockHeader


siladu:
Following https://github.com/hyperledger/besu/pull/4743

We should update all the usages we can to be getByBlockHeader and phase out getByBlockNumber.

getByBlockNumber can cause issues in a post-shanghai (TimestampSchedule-enabled) chain.

This has been intentionally separated out of #4743 so it doesn't make the PR huge, see this spike for an idea of what the changes might look like: https://github.com/hyperledger/besu/pull/4687

Another related item for clean-up:
https://github.com/hyperledger/besu/pull/4743#discussion_r1045390099

Tasks
- [x] Simple replacements of getByBlockNumber to getByBlockHeader
- [x] Create getForNextBlock (used in AbstractBlockCreator, PoWMinerExecutor and likely in BFT)
- [x] Investigate FlexiblePrivacyPrecompiledContract usage in RunnerBuilder.createPrivateTransactionObserver()
- [x] Handle Dao fork and classic fork in ProtocolScheduleBuilder
- [x] BFT sequence number usage
- [x] Change streamMilestoneBlocks to a findSchedule(Predicate) [need to fix usage in TransactionPoolFactory, RunnerBuilder]
- [x] Consider deleting getByBlockNumber usage in CombinedProtocolScheduleFactory
- [x] Fix tests to getByBlockHeader or create a test ProtocolSchedule that has getByBlockNumber inc streamMilestoneBlocks for checking milestones
- [x] ScheduleBasedBlockHeaderFunctions getBlockHeaderFunctions usage of getByBlockNumber
- [x] CheckpointSyncDownloadPipelineFactory getBlockHeaderFunctions usage of getByBlockNumber
- [x] Replace HeaderBasedProtocolSchedule with ProtocolSchedule

gfukushima:
Partial work of the trivial replaces done here: https://github.com/gfukushima/besu/tree/replace-getByBlockNumber-by-getByBlockHeader

siladu:
Bug found during goerli sync: https://github.com/hyperledger/besu/issues/5064

PR #5127, opened 2023-02-23
Replace getByBlockNumber by getByBlockHeader

gfukushima:
## PR description
This PR brings the changes introduced in #5020 + the amend to the issue spotted on #5064.
Replacing getByBlockNumber by getByBlockHeader brought issue to sync because TransitionProtocolSchedule getByBlockHeader does not have the same behaviour as getByBlockNumber.
This PR Introduced a private TransitionProtocolSchedule getByBlockHeaderForBws to restore the original behaviour of getByBlockHeader.
getByBlockHeader follows the same approach getByBlockNumber used to have which uses the dispatchFunctionAccordingToMergeState to return the pre or post merge protocolSchedule.

Syncing test done:

- [x] - Checkpoint Sepolia 
- [x] - Checkpoint Goerli
- [x] - Checkpoint Mainnet

## Fixed Issue(s)

Part of [#4789](https://github.com/hyperledger/besu/issues/4789)

## Documentation

## Changelog


consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
```diff
@@ -159,52 +155,45 @@ public void getByBlockHeader_returnsTimestampScheduleIfPresent() {
   }
 
   @Test
-  public void getByBlockNumber_returnsTimestampScheduleIfPresent() {
```
siladu, line 162: Why are the getByBlockNumber tests being removed and migrated to cover getByBlockHeaderForBws?

This PR isn't removing getByBlockNumber so I think the tests should remain, we may need to duplicate some for the new method.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
```diff
@@ -159,52 +155,45 @@ public void getByBlockHeader_returnsTimestampScheduleIfPresent() {
...
 
-    transitionProtocolSchedule.getByBlockNumber(BLOCK_NUMBER);
+    transitionProtocolSchedule.getByBlockHeaderForBws(blockHeader);
 
     verifyPostMergeProtocolScheduleReturned();
   }
 
   @Test
-  public void getByBlockNumber_delegatesToPostMergeScheduleWhenTimestampScheduleDoesNotExist() {
+  public void getByBlockHeader_delegatesToPostMergeScheduleWhenTimestampScheduleDoesNotExist() {
```
siladu, line 179: Test names should be updated from getByBlockHeader_... -> getByBlockHeaderForBws_...

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
```diff
@@ -159,52 +155,45 @@ public void getByBlockHeader_returnsTimestampScheduleIfPresent() {
...
-    transitionProtocolSchedule.getByBlockNumber(BLOCK_NUMBER);
+    transitionProtocolSchedule.getByBlockHeaderForBws(blockHeader);
 
     verifyPreMergeProtocolScheduleReturned();
   }
 
   @Test
-  public void getByBlockNumber_delegatesToPostMergeScheduleWhenBlockNotFound() {
-    when(blockchain.getBlockHeader(BLOCK_NUMBER)).thenReturn(Optional.empty());
+  public void getByBlockHeader_delegatesToPostMergeScheduleWhenBlockNotFound() {
```
siladu, line 167: Not shown in the diff, but I think getByBlockHeader() is now only being covered by one unit test, for when the timestamp schedule is found. It needs at least the other branch of logic tested, which is similar to this unit test.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/TransitionProtocolSchedule.java
```diff
@@ -105,13 +105,28 @@ public ProtocolSchedule getPostMergeSchedule() {
...
+                    protocolSchedule -> protocolSchedule.getByBlockHeader(blockHeader)));
+  }
+
+  /**
+   * Gets the protocol spec by block header, with some additional logic used by backwards sync (BWS)
+   *
+   * @param blockHeader the block header
+   * @return the ProtocolSpec to be used by the provided block
+   */
+  public ProtocolSpec getByBlockHeaderForBws(final ProcessableBlockHeader blockHeader) {
```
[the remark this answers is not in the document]
siladu: After reading around a bit, wonder if `getByBlockHeaderAndHandleReorgBelowTTD` makes more sense.

This appears to be the PR that introduced the extra logic: https://github.com/hyperledger/besu/pull/3696
Seems super-specific and edge-casey...wonder if syncing goerli always hits this edge case though 🤔

siladu:
I think there's still some cases where getByBlockNumber has not been migrated across (even though the header) is available so would like to understand those a bit more.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/TransitionProtocolSchedule.java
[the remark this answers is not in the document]
jframe: Why was using this causing the Goerli sync to fail?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContext.java
```diff
@@ -260,12 +260,12 @@ public ProtocolContext getProtocolContext() {
     return protocolContext;
   }
 
-  public BlockValidator getBlockValidator(final long blockNumber) {
-    return protocolSchedule.getByBlockNumber(blockNumber).getBlockValidator();
+  public BlockValidator getBlockValidator(final BlockHeader blockHeader) {
+    return protocolSchedule.getByBlockHeader(blockHeader).getBlockValidator();
```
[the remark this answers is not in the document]
jframe: This is only needed for backward syncing when using the transition protocol schedule. So it is only needed on the `TransitionBackwardSyncContext` not on this class.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
jframe: Changed the getByBlockHeader tests to use getByBlockHeader instead of getByBlockHeaderForBws. The other tests cover the backward sync conditions.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
jframe: Changed these tests to use getByBlockHeader so we have test coverage for that method

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
jframe: Added back getByBlockNumber tests which are almost duplicates of the getByBlockHeader tests since that method still exists and is being used.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
```diff
@@ -195,16 +179,64 @@ public void getByBlockNumber_delegatesToPostMergeScheduleWhenTimestampScheduleDo
...
+  public void getByBlockHeader_delegatesToPreMergeScheduleWhenBlockNotFound() {
+    when(mergeContext.isPostMerge()).thenReturn(false);
+
+    transitionProtocolSchedule.getByBlockHeader(blockHeader);
+
+    verifyPreMergeProtocolScheduleReturnedUsingBlockHeader();
+  }
+
+  @Test
+  public void getByBlockHeader_delegatesToPostMergeScheduleWhenBlockNotFound() {
```
siladu, line 206: How is this distinguished from the test below, `getByBlockHeader_delegatesToPostMergeScheduleWhenTimestampScheduleDoesNotExist`?

Before, it was testing 'block not found' with this...
`when(blockchain.getBlockHeader(BLOCK_NUMBER)).thenReturn(Optional.empty());`

jframe: Fixed the tests for the block not found cases for `getByBlockNumber`. And have removed tests for `getByBlockHeader` for the block not found cases. For `getByBlockHeader` we don't need to do a `blockchain.getBlockHeader()` call so we can never get the case where the block or block header doesn't exist.

siladu (approved):
LGTM

PR #5129, opened 2023-02-23
[stratum] add proxy subscription + cleanup + Netty integration

diega:
## PR description

The main goal of this PR is to add support for connecting an [iPollo](https://ipollo.com/) miner to Besu using the native stratum protocol.
iPollo expects periodic plain text _work_ messages through the same socket that performed the `eth_submitLogin` operation. As the source code for this miner is not publicacly available, this behaviour was inferred by porting the protocol from a pool softaware (i.e. [open-etc-pool](https://github.com/etclabscore/open-etc-pool)). Also, by adding this, the `stratum1+tcp` protocol from [ethminer](https://github.com/ethereum-mining/ethminer) now works as expected by obtaining new jobs to update its current one (mentioned briefly in [this](https://github.com/hyperledger/besu/pull/140#issuecomment-552074971) comment).

Several things have been adjusted and/or refactored in this PR, but the biggest one is in the last commit. This commit replaces the existing mechanism for supporting the multi protocol nature of the service, with one implemented on top of Netty classes. This can be sent in a separate PR if requested.

It's recommended to review this PR commit by commit.

## Documentation

The documentation for `eth_getWork` needs to be adjusted [here](https://github.com/hyperledger/besu-docs/blob/23.1.0/docs/public-networks/reference/api/index.md#eth_getwork) by removing the fourth return value

## Changelog


shemnon:
Looks reasonable to me. I'll approve next week if none of the other devs have issues.  It seems isolated but I want to give time to review.

Does any of this negatively impact any other stratum users? (I'm not interested in plugging in miners to the CI system, so a hand-cranked set of run throughs will satisfy me). 

Some of the changes in the last commit looked hard wired for specific cases (such as prefixing "POST / HTTP" instead of doing a full HTTP hander), does all of that fade away with netty handling?

siladu (commented):
Similar to Danno, it looks generally fine and no danger of regressing anything else. I will leave for someone else to approve as I no nothing about stratum

macfarla:
Agree with above comments - changes look to be isolated. I also know nothing about stratum.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetWorkTest.java
```diff
@@ -69,8 +69,7 @@ public void shouldReturnCorrectResultOnGenesisDAG() {
     final String[] expectedValue = {
       "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
       "0x0000000000000000000000000000000000000000000000000000000000000000",
-      "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
-      "0x0"
+      "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
```
atoulme, line 72: this is a meaningful change. The block number is no longer sent on eth_getWork ?

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumConnection.java
```diff
@@ -32,103 +28,52 @@
...
-        close();
-        return;
-      }
-
-      LOG.trace("Dispatching HTTP message {}", messagesString);
-      handleMessage(messagesString);
+  void handleBuffer(final Buffer message, final Consumer<String> sender) {
+    final Buffer trimmedMessage;
+    if (message.getByte(message.length() - 1)
+        == System.lineSeparator().getBytes(StandardCharsets.UTF_8)[0]) {
```
atoulme, line 48: That's weird. Is `\r` a valid separator?

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumConnectionTest.java
```diff
@@ -168,46 +174,9 @@ public void testStratum1SubmitHashrate() {
...
                 + "    \"0x03\",\"0x02\" "
                 + "  ]"
-                + "}\n"));
-    assertThat(called.get()).isFalse();
-
-    assertThat(message.get()).isEqualTo("{\"jsonrpc\":\"2.0\",\"id\":23,\"result\":true}\n");
-  }
-
-  @Test
-  public void testHttpMessage() {
```
atoulme, line 178: I'd like to see the exact same test work before and after to make sure there's no regression. Is that possible?

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumConnectionTest.java
```diff
@@ -87,20 +89,21 @@ public void testStratum1Matches() {
...
                 + "  ]"
-                + "}\n"));
+                + "}\n"),
+        message::set);
     assertThat(called.get()).isFalse();
 
     assertThat(message.get())
         .isEqualTo(
-            "{\"jsonrpc\":\"2.0\",\"id\":23,\"result\":[[\"mining.notify\",\"abcd\",\"EthereumStratum/1.0.0\"],\"\"]}\n");
+            "{\"jsonrpc\":\"2.0\",\"id\":23,\"result\":[[\"mining.notify\",\"abcd\",\"EthereumStratum/1.0.0\"],\"\"]}");
```
atoulme, line 98: each message must end with a \n character.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumServer.java
```diff
@@ -90,65 +111,150 @@ public StratumServer(
...
+            protocols,
+            () ->
+                socket
+                    .close()
+                    .onSuccess(v -> logger.debug("Socket closed by server"))
+                    .onFailure(ex -> logger.error("Error closing socket", ex)),
+            response -> sendLineBasedResponse(internalSocket, response));
+    internalSocket
+        .channelHandlerContext()
+        .pipeline()
```
atoulme, line 144: please add an exception handler for the pipeline to recover.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetWorkTest.java
atoulme: this should be a separate PR to avoid mixing too many things and making sure folks are notified of the change in the release notes.

atoulme (changes requested):
There are changes to how line breaks are handled in this PR that show me that we will have regressions if we adopt this change. I also think the changes to eth_getWork should be in a separate PR.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetWorkTest.java
diega: Sure, I can send it separatedly.
I changed this when I realized that the _iPollo_ wasn't accepting the `eth_getWork` responses properly, then I checked the [spec](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getwork) I realized that it only specify three return elements. I would have expected that no parser would comply for having extra elements but it doesn't seems to be the case.
Do you think that being out of spec is enough for chaging this return value? otherwise I would need to implement a different protocol flavor.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumConnection.java
diega: No, you're correct. I missed the case of `line.separator` being `\r\n` for Windows systems. I'll fix it.

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumConnectionTest.java
diega: I moved the messages split one layer up by adding [`io.netty.handler.codec.LineBasedFrameDecoder`](https://netty.io/4.0/api/io/netty/handler/codec/LineBasedFrameDecoder.html) to the channel pipeline. That's why now the `StratumConnection` only serves as a bridge between the message and the _protocols_ and doesn't need to handle line endings anymore.

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumConnectionTest.java
diega: This should be covered by `StratumServerTest#handleHttpPost`, I can add the `Connection` and `Accept-Encoding` headers though.

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumConnectionTest.java
atoulme: The decoder decodes, but how about sending messages?

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetWorkTest.java
atoulme: I guess it's fine to remove the fourth parameter. I am not completely sure where that came from, might be a misinterpretation of the spec for stratum1.

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumServerTest.java
```diff
@@ -47,16 +72,194 @@ public void runJobSetZeroDifficultyShouldNotThrow() {
...
+                              .setDefaultHost(host)
+                              .setDefaultPort(server.actualPort()));
+                  http.post("/")
+                      .sendBuffer(DEFAULT_TEST_MESSAGE)
+                      .onComplete(
+                          testContext.succeeding(
+                              resp -> {
+                                testContext.verify(
+                                    () -> {
+                                      assertThat(resp.statusCode()).isEqualTo(200);
```
atoulme, line 247: Can we assert on the response body here?

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumServerTest.java
```diff
@@ -47,16 +72,194 @@ public void runJobSetZeroDifficultyShouldNotThrow() {
...
+                                              .setDefaultHost(host)
+                                              .setDefaultPort(server.actualPort()));
+                                  http.post("/")
+                                      .sendStream(
+                                          file,
+                                          testContext.succeeding(
+                                              resp -> {
+                                                testContext.verify(
+                                                    () -> {
+                                                      assertThat(resp.statusCode()).isEqualTo(200);
```
atoulme, line 103: what is the response body value?

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumServerTest.java
```diff
@@ -47,16 +72,194 @@ public void runJobSetZeroDifficultyShouldNotThrow() {
...
+                                                        .onComplete(
+                                                            testContext.succeedingThenComplete());
+                                                  }))
+                                      .write(
+                                          DEFAULT_TEST_MESSAGE.appendString(System.lineSeparator()))
+                                      .onComplete(
+                                          writeResult ->
+                                              testContext.verify(
+                                                  () ->
+                                                      assertThat(writeResult.failed())
```
atoulme, line 176: please check what is being returned? Same as other comments, check the body matches expectations

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetWorkTest.java
diega: I opened #5184 for this, but having the fourth value in the HTTP response was a requirement for a mining pool, so it cannot be removed. I reworked it to expose the specified three elements so it can be used in the stratum protocol but leave the HTTP response as is.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/StratumConnection.java
diega: I fixed this and moved this code to the line based pipeline b/c this trim up doesn't make sense for HTTP

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumConnectionTest.java
diega: I added both headers to the mentioned test

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumServerTest.java
diega: I added the assert on the response body expecting the same value used in `StratumConnectionTest#testHttpMessage`

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumServerTest.java
diega: I added an assert on the body response

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumServerTest.java
diega: I added the assertion on the body here too

ethereum/stratum/src/test/java/org/hyperledger/besu/ethereum/stratum/StratumConnectionTest.java
diega: You were correct, the original PR used `System.lineSeparator()` as the delimiter for responses which wasn't suitable for defining a protocol as this was OS dependant. I changed it to always be delimited by `\n` as it is in main. You can see this in `StratumServer#sendLineBasedResponse`.
Also, there is a test covering this type of responses in `StratumServerTest#handlePlainTextMultiple`.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/GetWorkProtocol.java
```diff
@@ -37,50 +39,34 @@
...
-    JsonNode methodNode = message.get("method");
-    if (methodNode != null) {
-      String method = methodNode.textValue();
+    String method = message.getString("method");
+    if (method != null) {
       if ("eth_getWork".equals(method) || "eth_submitWork".equals(method)) {
-        JsonNode idNode = message.get("id");
-        boolean canHandle = idNode != null && idNode.isInt();
-        handle(conn, initialMessage);
+        Integer idNode = message.getInteger("id");
```
atoulme, line 67: this code is not safe. See https://vertx.io/docs/apidocs/io/vertx/core/json/JsonObject.html#getInteger-java.lang.String- - you will get a `ClassCastException` if the value is not an integer.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/GetWorkProtocol.java
```diff
@@ -91,59 +77,47 @@ public boolean maybeHandle(final String initialMessage, final StratumConnection
...
-                + CRLF
-                + responseMessage);
+        sender.accept(responseMessage);
       } catch (JsonProcessingException e) {
         LOG.warn("Error sending work", e);
         conn.close();
       }
     } else if ("eth_submitWork".equals(method)) {
-      JsonNode paramsNode = jsonrpcMessage.get("params");
+      JsonArray paramsNode = jsonrpcMessage.getJsonArray("params");
```
atoulme, line 105: same again, in case if it's not an array you get a ClassCastException:
https://vertx.io/docs/apidocs/io/vertx/core/json/JsonObject.html#getJsonArray-java.lang.String-

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/Stratum1EthProxyProtocol.java
```diff
@@ -93,66 +85,50 @@ public boolean maybeHandle(final String initialMessage, final StratumConnection
...
+        case "eth_submitHashrate" -> rpcResponse = ethSubmitHashRate.response(reqContext);
+        default -> {
+          LOG.debug("Invalid method: {}", req.getMethod());
+          conn.close();
+          return;
+        }
       }
-    } catch (IllegalArgumentException | IOException e) {
+      sender.accept(mapper.writeValueAsString(rpcResponse));
+    } catch (IllegalArgumentException | IOException | NullPointerException e) {
```
atoulme, line 111: NPE?

atoulme:
getting closer - just need a few more changes.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/GetWorkProtocol.java
diega: I changed it now so it always tries to execute the `handle` method.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/GetWorkProtocol.java
```diff
@@ -91,59 +77,47 @@ public boolean maybeHandle(final String initialMessage, final StratumConnection
...
     if (jsonrpcMessage == null) {
       LOG.warn("Invalid message {}", message);
       conn.close();
       return;
     }
-    JsonNode methodNode = jsonrpcMessage.get("method");
-    JsonNode idNode = jsonrpcMessage.get("id");
-    if (methodNode == null || idNode == null) {
+    String method = jsonrpcMessage.getString("method");
+    Integer id = jsonrpcMessage.getInteger("id");
```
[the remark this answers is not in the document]
diega: Yes, you're correct. Thank you for noticing it
Now I think I'm handling those errors case by case and handling each exception at the server level, to respond according the protocol.

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/GetWorkProtocol.java
diega: Handled now

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/Stratum1EthProxyProtocol.java
diega: Removed

atoulme (approved):
OK, approving the PR.

macfarla:
@diega is this one good to merge?

diega:
@macfarla yes it is

non-fungible-nelson:
Hey @diega - should we try to merge this for the 23.4.0 release candidate next week? If so, can you look to get the conflicts resolved today?

Issue #3786, opened 2022-05-04
OutOfMemoryError on JSON RPC method trace_replayBlockTransactions


ahamlat:
### Description
As an Besu, I want to get block transactions traces when I call JSON RPC method trace_replayBlockTransactions on blocks with big transactions (> 20M Gas)

### Acceptance Criteria
* trace_replayBlockTransactions JSON RPC call displays block transactions traces

### Steps to Reproduce (Bug)
1. Create a private network of Besu nodes, set for each node these java options -Xms2G -Xmx2G (BESU_OPTS)
2. Send a huge transaction 25M Gas 
3. Call JSON RPC method trace_replayBlockTransactions with the  block number of the transaction sent in 2. as parameter

**Expected behavior:** 
Return the block transactions traces, we can see below the json result template
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result":[ 
      { 
        "output":"0x",
        "vmTrace":{
          "code":"0x7f3940be4289e4c3587d88c1856cc95352461992db0a584c281226faefe560b3016000527f14c4d2c102bdeb2354bfc3dc96a95e4512cf3a8461e0560e2272dbf884ef3905601052600851",
          "ops":[ 
            { 
              "cost":3,
              "ex":{ 
                "mem":null,
                "push":[ 
                  "0x8"
                ],
                "store":null,
                "used":16756175
              },
              "pc":72,
              "sub":null
            },
            ...
          ]
        },
        "trace":[ 
          { 
            "action":{ 
              "callType":"call",
              "from":"0xfe3b557e8fb62b89f4916b721be55ceb828dbd73",
              "gas":"0xffadea",
              "input":"0x",
              "to":"0x0100000000000000000000000000000000000000",
              "value":"0x0"
            },
            "result":{ 
              "gasUsed":"0x1e",
              "output":"0x"
            },
            "subtraces":0,
            "traceAddress":[      
            ],
            "type":"call"
          }
        ],
        "stateDiff":{
          "0xfe3b557e8fb62b89f4916b721be55ceb828dbd73":{ 
            "balance":{ 
              "*":{ 
                "from":"0xffffffffffffffffffffffffffffffffc3e12a20b",
                "to":"0xffffffffffffffffffffffffffffffffc3dc5f091"
              }
            },
            "code":"=",
            "nonce":{ 
              "*":{ 
                "from":"0x14",
                "to":"0x15"
              }
            },
            "storage":{              
            }
          }
        },
[...]

PR #5029, opened 2023-01-31
Fix for #3786 OOM on trace_block and trace_replayBlockTransactions RPC methods

dhallatmaine:
## PR description
This change addresses a situation where all TraceFrame's for all Transactions in a block are loaded into memory, when only the TraceFrame's for a single Transaction are ever needed.

## Fixed Issue(s)
fixes #3786 

## Documentation
- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog
- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #5131, opened 2023-02-23
Use Besu pipeline in trace_block, trace_filter and trace_replayBlockTransactions

ahamlat:
## PR description
Besu pipeline is a sequence of stages and each stage, gets input data from the previous stage, can be executed in parallel to other stages, and put the executions results (output) into the input of the next stage.

We have for some tracing endpoints Out Of Memory errors because Besu keeps in memory all the traces. The pipeline will help in this case to break down the whole process into steps and keep in memory the objects needed for each step. This is one way to stream the work.

For example, during block tracing, instead of keeping in memory all the traces, with this PR we'll keep in memory only the traces of few transactions (the number of transactions is defined in the pipeline and may change in the future), but we still need to  keep in memory, for now, the object representation of the json file which consumes much less memory. Other optimisations in the future might improve this part.

Below the current implementation of the pipeline, this implementation may change in the future to improve performance or better manage java memory footprint.

image

EDIT : Regarding the illustration above, the implementation changed to stream TransactionTrace instead of Transaction.

You can find below some results on a 16 GiB/4 vCPU VM running Besu with Prysm as a CL

1. trace_replayBlockTransactions

image

2. trace_block

image

3. trace_filter

image

## Fixed Issue(s)

Fixes https://github.com/hyperledger/besu/issues/3786
## Documentation

## Changelog


Gabriel-Trintinalia:
#5029

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java
```diff
@@ -88,31 +98,75 @@ protected ArrayNodeWrapper traceBlock(
...
+
+              try {
+                if (getBlockchainQueries().getEthScheduler().isPresent()) {
+                  getBlockchainQueries()
+                      .getEthScheduler()
+                      .get()
+                      .startPipeline(traceBlockPipeline)
+                      .get();
+                } else {
+                  EthScheduler ethScheduler = new EthScheduler(1, 1, 1, 1, new NoOpMetricsSystem());
```
garyschulte, line 158: nit: since getEthScheduler returns optional, this could be an Optional.else() instead of a an if/else with duplicate startPipeline and get

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecuteTransactionStep.java
```diff
@@ -0,0 +1,104 @@
...
+      final Blockchain blockchain,
+      final DebugOperationTracer tracer,
+      final ProtocolSpec protocolSpec) {
+    this(chainUpdater, transactionProcessor, blockchain, tracer, protocolSpec, null);
+  }
+
+  @Override
+  public TransactionTrace apply(final TransactionTrace transactionTrace) {
+    Block block = this.block;
+    if (block == null) block = transactionTrace.getBlock().get();
```
garyschulte, line 71: it doesn't look like there is any guarantee of transactionTrace.getBlock() being present, is throwing NoSuchElementException the desired behavior when it is not?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java
```diff
@@ -27,31 +28,40 @@
...
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class TraceBlock extends AbstractBlockParameterMethod {
   private static final Logger LOG = LoggerFactory.getLogger(TraceBlock.class);
-
   private static final ObjectMapper MAPPER = new ObjectMapper();
-  private final Supplier<BlockTracer> blockTracerSupplier;
+  // private final Supplier<BlockTracer> blockTracerSupplier;
```
garyschulte, line 58: commented code

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java
```diff
@@ -27,31 +28,40 @@
...
   protected final ProtocolSchedule protocolSchedule;
+  // Either the initial block state or the state of the prior TX, including miner rewards.
```
garyschulte, line 60: what does this comment mean?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java
```diff
@@ -27,31 +28,40 @@
...
 
-  public TraceBlock(
-      final Supplier<BlockTracer> blockTracerSupplier,
-      final ProtocolSchedule protocolSchedule,
-      final BlockchainQueries queries) {
+  public TraceBlock(final ProtocolSchedule protocolSchedule, final BlockchainQueries queries) {
     super(queries);
-    this.blockTracerSupplier = blockTracerSupplier;
+    //    this.blockTracerSupplier = blockTracerSupplier;
```
garyschulte, line 64: commented code

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceFilter.java
```diff
@@ -48,7 +71,7 @@ public TraceFilter(
       final Supplier<BlockTracer> blockTracerSupplier,
```
garyschulte, line 71: should drop blockTracerSupplier if we are not using it

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceFilterSource.java
```diff
@@ -0,0 +1,92 @@
...
+  public boolean hasNext() {
+    if (resultArrayNode.isFull()) return false;
+    if (transactionTraceIterator == null) {
+      return false;
+    }
+    if (transactionTraceIterator.hasNext()) {
+      return true;
+    }
+    transactionTraceIterator = getNextTransactionIterator();
+    return hasNext();
```
garyschulte, line 78: if we start with an empty blockIterator, this can infinitely recurse no?  we should assert that the blockList and therefore blockIterator is not empty in the constructor, or change the flow to make sure we cannot get into trouble.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceFilterSource.java
```diff
@@ -0,0 +1,92 @@
...
+  @Override
+  public TransactionTrace next() {
+    if (transactionTraceIterator == null) {
+      return null;
+    }
+    if (transactionTraceIterator.hasNext()) {
+      return transactionTraceIterator.next();
+    }
+    transactionTraceIterator = getNextTransactionIterator();
+    return next();
```
garyschulte, line 90: same concern

garyschulte (changes requested):
Use of pipeline is very cool.  I tested this with a few mainnet blocks and it worked quite well.

I have some ignorable nit-picks about style and commented code and such, but a couple bits in TraceFilterSource made me want to hold off on merging this until you could address them, but mostly I just ran out of time today for a proper review before the release.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceFilterSource.java
garyschulte: Nevermind - it is getting late and I read the first condition as != null.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java
ahamlat: This comment is not in the right place anymore, it is more related this part of the code https://github.com/ahamlat/besu/blob/block-tracing-with-a-pipeline/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java#L205
If we have no prior updater, this means that we're executing the first transaction, so in this case use the block's initial state.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceFilterSource.java
ahamlat: There is no infinite recurse as we have this condition at the beginning   
```
if (!blockIterator.hasNext()) {
      return null;
    } 
```
There're actually unit tests that test this specific use case, a bloc with no transactions. In the unit tests, block 0 doesn't exist and block 1 is a block with 0 transactions.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceFilterSource.java
ahamlat: In this case getNextTransactionIterator() will return null in the constructor, then hasNext will return false and next() is never called

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java
ahamlat: Done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceBlock.java
ahamlat: Comment deleted

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecuteTransactionStep.java
ahamlat: This case shouldn't happen but I agree, need more control in this case. We can keep default NoSuchElementException but I added a Runtime Exception to be more specific with the message.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/BuildArrayNodeCompleterStep.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2:
```suggestion
 * Copyright contributors to Hyperledger Besu.
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecuteTransactionStep.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public TransactionTrace apply(final TransactionTrace transactionTrace) {
+    Block block = this.block;
+    // case where transactionTrace is created only to trace a block reward
+    if (block == null) {
+      block =
+          transactionTrace
+              .getBlock()
+              .orElseThrow(() -> new RuntimeException("Block should not be empty in this case"));
```
macfarla, line 76:
```suggestion
              .orElseThrow(() -> new RuntimeException("Expecting reward block to be in transactionTrace but was empty"));
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceReplayBlockTransactions.java
```diff
@@ -104,122 +97,86 @@ protected Object resultByBlockNumber(
...
+              final MainnetTransactionProcessor transactionProcessor =
+                  protocolSpec.getTransactionProcessor();
+              final ChainUpdater chainUpdater = new ChainUpdater(traceableState);
+
+              final TransactionSource transactionSource = new TransactionSource(block);
+              final LabelledMetric<Counter> outputCounter =
+                  new PrometheusMetricsSystem(BesuMetricCategory.DEFAULT_METRIC_CATEGORIES, false)
+                      .createLabelledCounter(
+                          BesuMetricCategory.BLOCKCHAIN,
+                          "transactions_traceblock_pipeline_processed_total",
```
macfarla, line 125: should the name for this metric be different from the one in TraceBlock

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/debug/InputDataManager.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(InputDataManager.class);
+
+  private final List<Bytes> inputDataCollection = Collections.synchronizedList(new ArrayList<>());
+
+  public Bytes getInputData(final int index) {
+    return inputDataCollection.get(index);
+  }
+
+  public int addInputData(final Bytes bytes) {
```
macfarla, line 35: where does this get used?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceReplayBlockTransactions.java
ahamlat: I changed the name, thanks.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/debug/InputDataManager.java
ahamlat: This is not used anymore, I deleted the InputDataManager.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceReplayBlockTransactions.java
```diff
@@ -104,122 +97,86 @@ protected Object resultByBlockNumber(
...
+              final MainnetTransactionProcessor transactionProcessor =
+                  protocolSpec.getTransactionProcessor();
+              final ChainUpdater chainUpdater = new ChainUpdater(traceableState);
+
+              final TransactionSource transactionSource = new TransactionSource(block);
+              final LabelledMetric<Counter> outputCounter =
+                  new PrometheusMetricsSystem(BesuMetricCategory.DEFAULT_METRIC_CATEGORIES, false)
+                      .createLabelledCounter(
+                          BesuMetricCategory.BLOCKCHAIN,
+                          "transactions_traceReplayblock_pipeline_processed_total",
```
macfarla, line 125:
```suggestion
                          "transactions_tracereplayblock_pipeline_processed_total",
```

macfarla: the smallest of nits now!

macfarla (approved):
LGTM - one remaining nit

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceReplayBlockTransactions.java
Gabriel-Trintinalia: Done!

PR #5133, opened 2023-02-24
Fix transaction selection result regression

fab-10:
The regression was introduced in commit 28f2bd4a1ad618130763ea1163e0c72f3bf30330, where transactions are kept by type, but that changes the order in which they were added, that is required when building the block, the solution is to also keep a transaction list in the order they are added.

## Documentation

## Changelog


shemnon (approved):
LGTM

PR #5143, opened 2023-02-28
Add Withdrawls execution tests to reference test

usmansaleem:
## PR description
Add Ethereum execution spec test to be generated and executed by reference tests gradle task. Add `https://github.com/ethereum/execution-spec-tests/releases` as gradle repository to easily download and cache the test specs. Update the reference test gradle file to generate the test Java classes.

## Fixed Issue(s)

#4934 
## Documentation

## Changelog


ethereum/referencetests/build.gradle
```diff
@@ -29,6 +29,37 @@ def blockchainReferenceTests = tasks.register("blockchainReferenceTests")  {
...
+configurations {
+  // we need this because referenceTestImplementation is defined as 'canBeResolved=false'.
+  tarConfig.extendsFrom referenceTestImplementation
+  tarConfig {
+    canBeResolved = true
+    canBeConsumed = false
+  }
+}
+
+def execSpecTests = tasks.register("execSpecTests") {
```
siladu, line 41:
```suggestion
def execSpecTests = tasks.register("executionSpecTests") {
```

I think matching the repo name will make searching slightly easier.

ethereum/referencetests/build.gradle
```diff
@@ -29,6 +29,37 @@ def blockchainReferenceTests = tasks.register("blockchainReferenceTests")  {
...
+  final referenceTestsPath = "$buildDir/tmp/execution-spec-tests/"
+  final generatedTestsPath = "$buildDir/generated/sources/reference-test/$name/java"
```
siladu, line 43: Wondering what the right level is to group these...i.e. should they be alongside the "reference-test" or underneath it as you have here.

What's your thinking on this?

ethereum/referencetests/build.gradle
```diff
@@ -100,6 +133,7 @@ dependencies {
   referenceTestImplementation project(path: ':evm')
   referenceTestImplementation project(path: ':testutil')
   referenceTestImplementation project(path: ':util')
+  referenceTestImplementation 'ethereum:execution-spec-tests:0.2.3:fixtures@tar.gz'
```
jframe, line 136: Can the version be included in the versions.gradle file?

build.gradle
```diff
@@ -120,6 +120,21 @@ allprojects {
       content { includeGroupByRegex('com\\.splunk\\..*') }
     }
     mavenCentral()
+
+    // ethereum execution spec tests fixtures. Exclusively for ethereum submodule to run ref tests
+    def ethExecSpecTestsRepo = ivy {
```
jframe, line 125: Why is this repository created in a function? The example in https://gist.github.com/ysb33r/9f95bab338b912c45986 just adds it to the repositories.

usmansaleem: @jframe We are assigning the repository definition to a variable so that it can be used in next section i.e. `exclusiveContent`.

ethereum/referencetests/build.gradle
```diff
@@ -66,13 +97,15 @@ sourceSets {
...
+        execSpecTests,
         generalstateReferenceTests,
         generalstateRegressionReferenceTests
     }
     resources {
       srcDirs 'src/reference-test/resources',
         'src/reference-test/external-resources',
-        'src/reference-test/templates'
+        'src/reference-test/templates',
+        'build/tmp/execution-spec-tests/'
```
siladu, line 108: why do we need this extra directory?

siladu: I see the files exist twice locally...
```
find . -name withdrawals_newly_created_contract.json
./ethereum/referencetests/build/resources/referenceTest/fixtures/withdrawals/withdrawals/withdrawals_newly_created_contract.json
./ethereum/referencetests/build/tmp/execution-spec-tests/fixtures/withdrawals/withdrawals/withdrawals_newly_created_contract.json
```

siladu (commented):
thanks for putting code comments in!

ethereum/referencetests/build.gradle
usmansaleem: unfortunately not ... I did try it, but Spring's dependency management doesn't seem to work with the `revision]/[classifier].[ext]` pattern.

ethereum/referencetests/build.gradle
usmansaleem: I just followed existing definitions....these tasks are basically transformation tasks. they are applying the template to transform the json files to Java files and registering the output paths as `src` for the referenceTest.

ethereum/referencetests/build.gradle
usmansaleem: I suspect that that the resources folder gets 'copied' from src at a later stage ... because if I don't include the build/tmp path, the relavant json files doesn't get visible.

ethereum/referencetests/build.gradle
```diff
@@ -29,6 +29,37 @@ def blockchainReferenceTests = tasks.register("blockchainReferenceTests")  {
     )
 }
 
+configurations {
+  // we need this because referenceTestImplementation is defined as 'canBeResolved=false'.
```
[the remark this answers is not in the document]
usmansaleem: referenceTestImplementation is probably gradle builtin configuration, not declared in our build files. If we try to find the resolved tar (or any other resource), gradle fails and actually suggest to extend the configuration with canBeResolved true.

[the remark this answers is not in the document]
usmansaleem: https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:what-are-dependency-configurations

jframe (approved):

ethereum/referencetests/build.gradle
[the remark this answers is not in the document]
siladu: > referenceTestImplementation is probably gradle builtin configuration
reference tests are specific to ethereum I think, or do you mean they are _using_ built-in configuration?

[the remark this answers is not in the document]
siladu:
```suggestion
  // we need this because referenceTestImplementation defaults to 'canBeResolved=false'.
```

ethereum/referencetests/build.gradle
siladu: I guess the question is "is there any value in keeping execution-spec-test separate to reference-test?" 
I suspect not, esp since they are all running under the `referenceTest` command

build.gradle
```diff
@@ -120,6 +120,21 @@ allprojects {
...
+      url 'https://github.com'
```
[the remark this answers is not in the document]
usmansaleem: `ivy` repo definition needs a url (or local path).

ethereum/referencetests/build.gradle
usmansaleem: Gary has "extraction" as a seperate task ... I avoided it by using `copy` task within definition of `executionSpecTests`.

build.gradle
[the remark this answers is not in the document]
siladu: So why can't we use `url 'https://github.com/ethereum/execution-spec-tests'` and drop the filter?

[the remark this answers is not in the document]
usmansaleem: because of this: `referenceTestImplementation 'ethereum:execution-spec-tests:0.2.3:fixtures@tar.gz'`. We are taking advantage of organization, module, version since thats how the dependencies are declared and using them as variables to be used in the custom ivy module.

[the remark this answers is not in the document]
usmansaleem: I've added additional comment in referencetest build.gradle as well.

siladu (approved):
LGTM

PR #5145, opened 2023-03-02
[MINOR] Add slf4j api so that acceptance tests work with using acctests.runBesuAsProcess = false

siladu:


jframe:
Confirmed acceptance tests now pass when using the thread runner

jframe (approved):
LGTM

PR #5146, opened 2023-03-02
Add an acceptance test to ensure eth_getBlockByNumber and eth_getBlockByHash return withdrawals correctly for shanghai

siladu:
Fixes https://github.com/hyperledger/besu/issues/4808

I spent a good amount of time trying to add a test similar to https://github.com/hyperledger/besu/blob/main/ethereum/api/src/integration-test/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/fork/frontier/EthGetBlockByNumberIntegrationTest.java

...but was ultimately thwarted by some post-merge genesis protocol schedule issues. 

I think this AT-style is actually easier to understand and work with, and also tests the JSON serialization as a bonus.


acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/02_execute_payload_1.json
```diff
@@ -0,0 +1,35 @@
+{
```
jframe, line 1: Why does the first test start from 02?

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/04_update_forkchoice.json
```diff
@@ -0,0 +1,28 @@
+{
```
jframe, line 1: This test also starts with 04. I expected the prefixes would be unique like the other rpc tests

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/06_get_last_paris_block.json
```diff
@@ -0,0 +1,33 @@
+{
+  "request": {
+    "jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x1", false],"id":1
+  },
+  "response": {
+    "jsonrpc": "2.0",
+    "id": 1,
+    "result" : {
+      "number" : "0x1",
```
jframe, line 9: Would be a bit clearer if the block number for the last Paris block was 9. But since blocks, 2-8 are skipped it is actually the last one, just not what I expected going by the genesis config.

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/07_get_first_shanghai_block.json
```diff
@@ -0,0 +1,35 @@
...
+      "extraData" : "0x",
+      "baseFeePerGas" : "0x7",
+      "size" : "0x21c",
+      "gasLimit" : "0x1c9c380",
+      "gasUsed" : "0x0",
+      "timestamp" : "0x10",
+      "uncles" : [ ],
+      "transactions" : [ ],
+      "withdrawalsRoot" : "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
+      "withdrawals": [ ]
```
jframe, line 31: Can we also test this block with the empty withdrawals case using eth_getBlockByHash

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/02_execute_payload_1.json
siladu: haha, I knew you'd pick up on that ;)

I copied the test cases over from replayBlock ATs and I was too lazy to rename all the files after deleting the unnecessary first 01_prepare_payload.
I will rename :p

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/04_update_forkchoice.json
siladu: good catch, I missed that after inserting it...looks like I need to update the test names afterall :)

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/06_get_last_paris_block.json
siladu: Why 9?

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/07_get_first_shanghai_block.json
siladu: yeh sure. I know the wiring is the same code underneath so took a short cut, but maybe I should go whole hog and test last paris block too?

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/06_get_last_paris_block.json
jframe: Aah, I mixed up block number and block time reading this. Yeah, block 2 is fine after all.

acceptance-tests/tests/src/test/resources/jsonrpc/eth/getBlockBy/test-cases/07_get_first_shanghai_block.json
jframe: don't we need to test the last Paris block, just testing the shanghai block would be enough

jframe (approved):
LGTM

PR #5173, opened 2023-03-06
Remove getByBlockNumber for CheckpointSync

### LINKED ISSUE #5167: CheckpointSyncDownloadPipelineFactory getBlockHeaderFunctions usage of getByBlockNumber

siladu:
Pull up and simplify data passed to CheckpointSource, which then allows checkpointBlockHeader to be used by protocolSchedule.getByBlockHeader.

Fixes #5167 

Alternative approach explored: https://github.com/siladu/besu/commit/bbaa80dcd88c517fb0d7a07a6fcd9f8a027bf5a5
but I didn't like the way ScheduleBasedHeaderBlockHeaderFunctions requires an explicit delegation override (otherwise it implicitly gets the wrong value).

CheckpointSync testing:
- [x] Sepolia (prysm)
- [x] Sepolia (lighthouse)
- [x] Goerli (lighthouse)
  - got in sync but required a restarted to transition from initial sync to backwards sync 
  - resync completed without intervention 
- [x] Mainnet (teku)


gfukushima (approved):
ltgm

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/checkpointsync/CheckpointSource.java
```diff
@@ -35,12 +33,10 @@ public class CheckpointSource implements Iterator<Hash> {
   private final AtomicBoolean isDownloading = new AtomicBoolean(false);
 
   public CheckpointSource(
-      final SyncState syncState,
-      final EthPeer ethPeer,
-      final BlockHeaderFunctions blockHeaderFunctions) {
+      final SyncState syncState, final BlockHeader blockHeader, final int checkpointWindowSize) {
```
jframe, line 36: nit: would be nice if this arg name matched the field name

jframe (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/checkpointsync/CheckpointSource.java
siladu: @jframe do you think nbBlocks or checkpointWindowSize makes more sense here? i.e. rename the field or rename the arg name?

PR #5174, opened 2023-03-06
Fix empty body concept after shanghai

### LINKED ISSUE #4976: Withdrawals to handle filling empty blocks when syncing
The empty block check is missing checking the withdrawals in the CompleteBlocksTask hasEmptyBody method. Not sure if this is necessary, we need to understand more about what this class is doing but we need to consider withdrawals here as well.

gfukushima:
## PR description
This PR adds withdrawals to the method that checks for empty block bodies in CompleteBlocksTask.java
Adding withdrawals allows Besu to be able to distinguish empty blocks after shanghai skiping the block body download when body is empty.
Prior to this PR Besu would understand blocks like (0txs, 0 ommers, x ws) as an empty block, during the sync process. (x > 0)

## Fixed Issue(s)

Fixes #4976 

This PR has performed the following tests:
- [x] Recreated and tested fix on zhejiang-testnet using FAST sync
- [x] Successfully synced Sepolia (starting from checkpoint and transitioning to Shanghai)
- [x] Successfully synced Goerli 
- [ ] Successfully synced Mainnet
 
## Documentation

## Changelog


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTask.java
```diff
@@ -79,7 +79,11 @@ private CompleteBlocksTask(
 
   private boolean hasEmptyBody(final BlockHeader header) {
     return header.getOmmersHash().equals(Hash.EMPTY_LIST_HASH)
-        && header.getTransactionsRoot().equals(Hash.EMPTY_TRIE_HASH);
+        && header.getTransactionsRoot().equals(Hash.EMPTY_TRIE_HASH)
+        && header
+            .getWithdrawalsRoot()
```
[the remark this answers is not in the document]
garyschulte: does it though?  if withdrawals are not enabled wouldn't we get an empty Optional?  This looks correct to me.  

I presume we already have test coverage that asserts we will fail a block that tries to execute withdrawals before shanghai.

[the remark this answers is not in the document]
siladu: We discussed on a call. Decided it wasn't needed due to later validation as you say, but we do need to check the protocolSchedule for if withdrawals is enabled to be able to create a withdrawals-aware empty block. That changes and unit tests incoming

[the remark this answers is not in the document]
jframe: Yeah it didn't look right when I first looked at it. But we had a closer look at your right we don't need to do that check here as the validation will take care of that. It would be moving some of the validation into the class which isn't right.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTask.java
```diff
@@ -79,7 +79,11 @@ private CompleteBlocksTask(
```
jframe, line 79: The constructor needs to correct a block body with an empty list of withdrawals when there is an empty block Shanghai. This always creates a block body with an empty optional which is incorrect post-shanghai.

gfukushima: Yes, we've changed the implementation to take protocolSchedule into account for the empty body creation Will push that change soon. Thanks!

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
```diff
@@ -79,6 +95,115 @@ public void shouldCompleteWithoutPeersWhenAllBlocksAreEmpty() {
...
+            mockProtocolSchedule,
+            ethContext,
+            List.of(header1, header2),
+            maxRetries,
+            new NoOpMetricsSystem());
+    assertThat(task.run()).isCompletedWithValue(expectedBlocks);
+  }
+
+  @Test
+  public void shouldCompleteWhenBlocksAreEmptyExceptForWithdrawals() {
```
jframe, line 138: Is this testing that when blocks are empty but have withdrawals they are still downloaded? i.e. that they are considered empty blocks. Was a bit confused from the test name

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
```diff
@@ -79,6 +95,115 @@ public void shouldCompleteWithoutPeersWhenAllBlocksAreEmpty() {
...
+            secondWithdrawalsHeader,
+            new BlockBody(
+                Collections.emptyList(),
+                Collections.emptyList(),
+                Optional.of(Collections.emptyList())));
+    final List<Block> expected = asList(block1, block2, block3);
+
+    final RespondingEthPeer respondingPeer =
+        EthProtocolManagerTestUtil.createPeer(ethProtocolManager, 1000);
+    final RespondingEthPeer.Responder responder = responderForFakeBlocks(expected);
```
jframe, line 172: Why does the responder need all the blocks? It should only need to download block2 as the others are empty

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
```diff
@@ -79,6 +95,115 @@ public void shouldCompleteWithoutPeersWhenAllBlocksAreEmpty() {
...
+    final List<Withdrawal> withdrawals = List.of(withdrawal);
+    final Hash withdrawalsRoot = BodyValidation.withdrawalsRoot(withdrawals);
+
+    final BlockHeader preWithdrawalsHeader = new BlockHeaderTestFixture().number(1).buildHeader();
+    final BlockHeader firstWithdrawalsHeader =
+        new BlockHeaderTestFixture().number(2).withdrawalsRoot(withdrawalsRoot).buildHeader();
+    final BlockHeader secondWithdrawalsHeader =
+        new BlockHeaderTestFixture().number(3).withdrawalsRoot(Hash.EMPTY_TRIE_HASH).buildHeader();
+
+    final Block block1 = new Block(preWithdrawalsHeader, BlockBody.empty());
```
jframe, line 155: nit: rename to emptyBlock

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
```diff
@@ -79,6 +95,115 @@ public void shouldCompleteWithoutPeersWhenAllBlocksAreEmpty() {
...
+    final Block block2 =
+        new Block(
+            firstWithdrawalsHeader,
+            new BlockBody(
+                Collections.emptyList(), Collections.emptyList(), Optional.of(withdrawals)));
+    final Block block3 =
```
jframe, line 161: nit: rename to empyWithdrawalsBlock?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
```diff
@@ -79,6 +95,115 @@ public void shouldCompleteWithoutPeersWhenAllBlocksAreEmpty() {
...
+            List.of(preWithdrawalsHeader, firstWithdrawalsHeader, secondWithdrawalsHeader),
+            maxRetries,
+            new NoOpMetricsSystem());
+
+    final CompletableFuture<List<Block>> runningTask = task.run();
+
+    assertThat(runningTask).isNotDone();
+    respondingPeer.respond(responder);
+    assertThat(runningTask).isCompletedWithValue(expected);
+    //    assertThat(task.run()).isCompletedWithValue(expected);
```
jframe, line 195: Left over from testing?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
siladu: it doesn't we thought it was a good test - e.g. let's say the code under test had a bug where it requested 3 blocks instead of 1, but we only ever returned one block in the responser, we wouldn't catch that.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
siladu: hmmm, wonder if @gfukushima didn't push the latest commit, I thought we removed that

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
siladu: arguably some value in keeping header1, block1, blockNumber = 1 all in sync

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
siladu: > Is this testing that when blocks are empty but have withdrawals they are still downloaded?
yes, i.e. that they are _not_ considered empty blocks like they were being before.

Any name suggestion? This was an modification of the existing test's name: `shouldCompleteWithoutPeersWhenAllBlocksAreEmpty`. Agree the new one could be better

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
gfukushima: Good catch, thanks

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTask.java
```diff
@@ -74,12 +75,32 @@ private CompleteBlocksTask(
...
+                    BlockHeader::getNumber,
+                    header ->
+                        new Block(
+                            header,
+                            new BlockBody(
+                                Collections.emptyList(),
+                                Collections.emptyList(),
+                                isWithdrawalsEnabled(protocolSchedule, header)
+                                    ? Optional.of(Collections.emptyList())
+                                    : Optional.empty()))));
```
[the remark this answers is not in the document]
gfukushima: Done

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
gfukushima: It does not, you're right, good catch!

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
```diff
@@ -79,6 +95,112 @@ public void shouldCompleteWithoutPeersWhenAllBlocksAreEmpty() {
     assertThat(task.run()).isCompletedWithValue(blocks);
   }
 
+  @Test
+  public void shouldCreateCorrectEmptyBlockWhenWithdrawalsAreEnabled() {
```
macfarla, line 99:
```suggestion
  public void whenWithdrawalsAreEnabled_shouldCreateEmptyBlockWithEmptyWithdrawalsList() {
```

macfarla (approved):
LGTM

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteBlocksTaskTest.java
siladu: Renamed both tests to be (hopefully) clearer and more consistent

jframe (approved):

PR #851, opened 2020-05-06
Add current block number to eth_getWork response

abdelhamidbakhta:
## PR description
- Updated `eth_getWork` response to add hex encoded current block number.
- Updated unit tests accordingly.

#### Request
```bash
curl --data '{"method":"eth_getWork","params":[],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
```
#### Response
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": [
    "0xce5e32ca59cb86799a1879e90150b2c3b882852173e59865e9e79abb67a9d636",
    "0x0000000000000000000000000000000000000000000000000000000000000000",
    "0x00a3d70a3d70a3d70a3d70a3d70a3d70a3d70a3d70a3d70a3d70a3d70a3d70a3",
    "0x42"
  ]
}
```

## Fixed Issue(s)

#849 

Signed-off-by: Abdelhamid Bakhta <abdelhamid.bakhta@consensys.net>

PR #5184, opened 2023-03-07
Remove the fourth response element in eth_getWork

diega:
## PR description
The [spec](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getwork)  for `eth_getWork` only defines three return values and certain parsers rejects a longer response array.

## Documentation

## Changelog


shemnon (approved):
@diega is this causing any problems in any miners?  Either by it's presence or absence?  

This was added in #851, with no explanation for why it was added.  So I am OK with it disappearing with no explanation, but if we can add context that would be better.

diega:
@shemnon through HTTP I couldn't detect any issue. This PR was meant to amend a deviation from the spec b/c for #5129 I must to send `eth_getWork` responses as part of the _stratum_ protocol and I wanted to use the same objects for that but, in that context, clients seems to be more strict and they reject an extra element. I'll solve it differently in the mentioned PR and I'll let this behavior as is.
Thanks for digging for a reason to this.

PR #5189, opened 2023-03-08
Isolate log4j

shemnon:
## PR description

Remove some of the tendrils of Log4J that have snuck into various code locations. Isolate them into a single class.

This is preparatory for GraalVM compilation of EVMTool as it doesn't play well with log4j and Besu needs to be able to operate with log4j removed from the classpath.

## Documentation

- [X] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Acceptance Tests (Non Mainnet)

- [X] I have considered running `./gradlew acceptanceTestNonMainnet` locally if my PR affects non-mainnet modules.

## Changelog

- [X] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #5192, opened 2023-03-08
Evmtool graalvm support

shemnon:
## PR description

Changes and config to support using GraalVM AOT to compile and
execute evmTool.  While not as long-term performant the startup time
makes filling reference tests with Besu reasonable.

Builds off of #5189 

## Documentation

## Changelog


crypto/src/main/java/org/hyperledger/besu/crypto/AbstractSECP256.java
```diff
@@ -44,13 +45,6 @@
 /** The Abstract secp256. */
 public abstract class AbstractSECP256 implements SignatureAlgorithm {
 
-  /** The constant PRIVATE_KEY_BYTE_LENGTH. */
-  protected static final int PRIVATE_KEY_BYTE_LENGTH = 32;
-  /** The constant PUBLIC_KEY_BYTE_LENGTH. */
-  protected static final int PUBLIC_KEY_BYTE_LENGTH = 64;
-  /** The constant SIGNATURE_BYTE_LENGTH. */
-  protected static final int SIGNATURE_BYTE_LENGTH = 65;
```
atoulme, line 52: are those constants no longer used? Does this relate to this PR or is it just cleanup?

crypto/src/main/java/org/hyperledger/besu/crypto/AbstractSECP256.java
```diff
@@ -82,7 +76,7 @@ protected AbstractSECP256(final String curveName, final BigInteger prime) {
     curveOrder = curve.getN();
     halfCurveOrder = curveOrder.shiftRight(1);
     try {
-      keyPairGenerator = KeyPairGenerator.getInstance(ALGORITHM, PROVIDER);
+      keyPairGenerator = new ECDSA();
```
atoulme, line 79: Is this change necessary?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/GenesisFileModule.java
```diff
@@ -101,20 +97,11 @@ static GenesisFileModule createGenesisModule(final File genesisFile) throws IOEx
...
-      return new IBFTGenesisFileModule(genesisConfig);
-    } else if (config.containsKey("clique")) {
-      return new CliqueGenesisFileModule(genesisConfig);
-    } else if (config.containsKey("qbft")) {
-      return new QBFTGenesisFileModule(genesisConfig);
-    } else {
-      // default is mainnet
-      return new MainnetGenesisFileModule(genesisConfig);
+    if (config.containsKey("ibft") || config.containsKey("clique") || config.containsKey("qbft")) {
+      throw new RuntimeException("Only Ethash and Merge configs accepted as genesis files");
```
atoulme, line 103: is this related to this PR?

crypto/src/main/java/org/hyperledger/besu/crypto/AbstractSECP256.java
shemnon: They show up as unused in IntelliJ analysis.

crypto/src/main/java/org/hyperledger/besu/crypto/AbstractSECP256.java
shemnon: Yes.  JCA and Bouncy Castle are very difficult to get working with GraalVM.  There is a lot of introspection involved and a lot of pointless overhead to get to what this change does.  We have it configured in the parameters (per ALGORITHM and PROVIDER) so that the referenced class is the only thing that can be returned.  Were this a system where we change crypto configs on the fly or if ethereum accepted multiple key lengths or algos then it would be worth it.  This is just overhead that makes AOT compilation impossible.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/GenesisFileModule.java
shemnon: This unused feature (the ability to calculate the fork from a genesis file) has a long cascading requirement that ultimately brings in all the RocksDB classes, and requires significantly more reflection config.  Nobody uses the feature because it is fragile and buggy.

atoulme:
OK, I'm not sure if you'd like me to review this, as so many changes seem to be picked up along the way. I am happy to review if it helps.

gradle/verification-metadata.xml
```diff
@@ -264,6 +274,17 @@
             <sha256 value="c2fd4366b1647233c99c70c17a2e90c89bf1ea60843749cca99d794525df9cd2" origin="Generated by Gradle"/>
          </artifact>
       </component>
+      <component group="com.fasterxml.jackson.core" name="jackson-annotations" version="2.13.3">
+         <artifact name="jackson-annotations-2.13.3.jar">
```
macfarla, line 278: we already have 2.13.4 of these jackson dependencies - is there a reason we need this specific version?

shemnon: One of the dependent libraries uses it.  Verification is based on the total closure of all possible dependency references, not on the set of actually used references.  You will also note that some SWT jars are also validated.

shemnon:
@atoulme The current diff, as it stands, is the set of changes I would like to commit to get graalvm working.  These changes do not appear in any other change list.  Please review it as-is.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/GenesisFileModule.java
shemnon: In short, these need to be removed to use GraalVM for evmtool.  I assert that no one will miss them.

crypto/src/main/java/org/hyperledger/besu/crypto/AbstractSECP256.java
atoulme: I'm so sorry, I'm thoroughly confused because I assumed that it was possible to configure the curve used. I'll take a deeper look.

macfarla (approved):
looks ok to me

crypto/src/main/java/org/hyperledger/besu/crypto/AbstractSECP256.java
shemnon: The curve kind (ECDSA vs Edwards) is the same across all current configurations.  The curve parameters (secp256k1 or secp256r1) are applied to the key generator after this construction.

Also, this is not for the whole of the public key encryption used in the system, just the key pair generators.  I updated the use of imports to clarify that.

PR #5198, opened 2023-03-09
Replace production usages of ProtocolSchedule.streamMilestoneBlocks

### LINKED ISSUE #5163: Change streamMilestoneBlocks to a findSchedule
Change the ProtocolSchedule streamMilestonBlocks to a findSchedule method using a predicate. Will also need to fix usage in TransactionPoolFactory the RunnerBuilder.

Part of #4789

siladu:
Replace usages of streamMilestoneBlocks on production code paths: RunnerBuilder and TransactionPoolFactory.
These usages were also using getBlockByNumber which is ultimately what we're trying to remove.


siladu:
In testing the `TransactionPoolFactory`, it did not work as I expected, but I think this code may have always had this bug:
https://github.com/hyperledger/besu/blob/main/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolFactory.java#L174-L199

If the network was currently on Berlin, I would expect the `TransactionPoolFactory` to return `GasPricePendingTransactionsSorter`, however, since London exists at all in the genesis, it always returns `BaseFeePendingTransitionsSorter`.

Is this a bug or intended? cc @garyschulte @fab-10 @matkt

fab-10:
> If the network was currently on Berlin, I would expect the `TransactionPoolFactory` to return `GasPricePendingTransactionsSorter`, however, if London exists at all in the genesis, even if it is a future block, it always returns `BaseFeePendingTransitionsSorter`.
> 
> Is this a bug or intended? 

I think it is intended, even if I am not the origin writer of that code, since `BaseFeePendingTransitionsSorter` is backward compatible so even if London is not yet activated it works, and since the txpool is created at startup there is no way to change it at runtime when London activates, so the only way to support the fork is to create it when London is present in the genesis

siladu:
> `BaseFeePendingTransitionsSorter` is backward compatible

Yeh I did wonder if that was the case. But then, why do we need the GasPrice one at all - can't we always just use the BaseFee one if it's compatible pre-London?

fab-10:
> Yeh I did wonder if that was the case. But then, why do we need the GasPrice one at all - can't we always just use the BaseFee one if it's compatible pre-London? 

The GasPrice is more simple and performant for non London networks

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MutableProtocolSchedule.java
```diff
@@ -74,37 +77,51 @@ public ProtocolSpec getByBlockNumber(final long number) {
...
   }
 
-  public List<ScheduledProtocolSpec> getScheduledProtocolSpecs() {
-    return protocolSpecs.stream().collect(Collectors.toUnmodifiableList());
+  public List<NumberScheduledProtocolSpec> getScheduledProtocolSpecs() {
+    return protocolSpecs.stream().toList();
   }
+
+  /** Tuple that associates a {@link ProtocolSpec} with a given block number level starting point */
+  public record NumberScheduledProtocolSpec(long block, ProtocolSpec spec)
```
fab-10, line 125: rename `block` to `blockNumber`?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilderTest.java
```diff
@@ -187,4 +189,22 @@ public void modifierOnDefinedMilestoneIsAppliedButDoesNotGetAnExtraMilestoneCrea
...
   }
+
+  @Test
+  public void isOnForkBoundary() {
+    when(configOptions.getBerlinBlockNumber()).thenReturn(OptionalLong.of(1L));
+    when(configOptions.getLondonBlockNumber()).thenReturn(OptionalLong.of(2L));
+    when(configOptions.getMergeNetSplitBlockNumber()).thenReturn(OptionalLong.of(4L));
+    final HeaderBasedProtocolSchedule protocolSchedule = builder.createProtocolSchedule();
+
+    assertThat(protocolSchedule.isOnForkBoundary(header(0))).isEqualTo(true);
```
fab-10, line 200: which fork boundary is 0?

fab-10 (commented):
LGTM, just a renaming and a question

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilderTest.java
siladu: Frontier. This is how ProtocolSchedule works, for better or worse :)
https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilder.java#L157

Note, that there's no 'default' or 0 value for the timestamp-based schedule though since we need to fall back to number-based schedule instead.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MutableProtocolSchedule.java
siladu: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HeaderBasedProtocolSchedule.java
```diff
@@ -34,4 +36,8 @@ public interface HeaderBasedProtocolSchedule {
   String listMilestones();
 
   Stream<Long> streamMilestoneBlocks();
+
+  boolean anyMatch(Predicate<ScheduledProtocolSpec> predicate);
+
+  boolean isOnForkBoundary(final BlockHeader blockHeader);
```
jframe, line 42: The name is inconsistent with the other naming using fork instead of milestone. So suggest `isOnMilestoneBoundary`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ScheduledProtocolSpec.java
```diff
@@ -14,21 +14,11 @@
...
-  }
-
-  public long getBlock() {
-    return block;
-  }
+/**
+ * Tuple that associates a {@link ProtocolSpec} with a given block number or timestamp level
+ * starting point
+ */
+public interface ScheduledProtocolSpec {
```
jframe, line 21: What does this type represent now? It's no longer a tuple with a block or timestamp, it just has a spec() method.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/TimestampScheduleBuilderTest.java
```diff
@@ -149,4 +149,22 @@ public void streamMilestoneBlocksReturnTimestampsInOrder() {
 
     assertThat(schedule.streamMilestoneBlocks()).containsExactly(FIRST_TIMESTAMP_FORK, 2L, 3L, 5L);
   }
+
+  @Test
+  public void isOnForkBoundary() {
```
jframe, line 154: This doesn't seem like the right place for this test as it's not implemented on the builder. Suggest putting in a TimestampScheduleTest instead.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MutableProtocolSchedule.java
```diff
@@ -74,37 +77,51 @@ public ProtocolSpec getByBlockNumber(final long number) {
...
   }
 
-  public List<ScheduledProtocolSpec> getScheduledProtocolSpecs() {
-    return protocolSpecs.stream().collect(Collectors.toUnmodifiableList());
+  public List<NumberScheduledProtocolSpec> getScheduledProtocolSpecs() {
+    return protocolSpecs.stream().toList();
   }
+
+  /** Tuple that associates a {@link ProtocolSpec} with a given block number level starting point */
+  public record NumberScheduledProtocolSpec(long blockNumber, ProtocolSpec spec)
```
jframe, line 125: Suggest using `BlockNumberScheduledProtocolSpec` instead, number by itself doesn't mean much.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HeaderBasedProtocolSchedule.java
siladu: good spot - done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ScheduledProtocolSpec.java
siladu: Have made this a record that is used by both types of schedule

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MutableProtocolSchedule.java
siladu: See refactoring comment above

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/TimestampScheduleBuilderTest.java
siladu: I agree. Current test structure follows ProtocolScheduleBuilderTest which conflates tests for both the builder and class itself. 

Since this issue affects multiple Schedule methods on both schedule types, I would like to handle this all in one go as one PR though, can play straight after this.

consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/TransitionProtocolScheduleTest.java
```diff
@@ -224,6 +224,68 @@ public void getByBlockHeader_delegatesToPostMergeScheduleWhenTimestampScheduleDo
...
+
+    when(timestampSchedule.isOnMilestoneBoundary(any(BlockHeader.class))).thenReturn(false);
+    when(preMergeProtocolSchedule.isOnMilestoneBoundary(any(BlockHeader.class))).thenReturn(true);
+
+    assertThat(transitionProtocolSchedule.isOnMilestoneBoundary(blockHeader)).isTrue();
+    verifyNoInteractions(postMergeProtocolSchedule);
+  }
+
+  @Test
+  public void isOnForkBoundary_delegatesToPostMergeSchedule() {
```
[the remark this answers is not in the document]
siladu: Thank you. Surprised intellij didn't pick some of these up tbh!

jframe (approved):

PR #5201, opened 2023-03-10
Remove getByBlockNumber from ProtocolScheduleBuilder

gfukushima:
## Documentation

## Changelog

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilder.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilder.java
@@ -192,7 +192,12 @@ public class ProtocolScheduleBuilder extends AbstractProtocolScheduleBuilder {
         .ifPresent(
             daoBlockNumber -> {
               final ProtocolSpec originalProtocolSpec =
-                  protocolSchedule.getByBlockNumber(daoBlockNumber);
+                  protocolSchedule.getScheduledProtocolSpecs().stream()
+                      .filter(x -> x.getBlock() <= daoBlockNumber)
+                      .findFirst()
+                      .get()
+                      .getSpec();
+
               addProtocolSpec(
                   protocolSchedule,
                   daoBlockNumber,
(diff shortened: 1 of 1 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilder.java
```diff
@@ -192,7 +192,12 @@ protected void postBuildStep(final MainnetProtocolSpecFactory specFactory) {
         .ifPresent(
             daoBlockNumber -> {
               final ProtocolSpec originalProtocolSpec =
-                  protocolSchedule.getByBlockNumber(daoBlockNumber);
+                  protocolSchedule.getScheduledProtocolSpecs().stream()
+                      .filter(x -> x.getBlock() <= daoBlockNumber)
```
siladu, line 196: I think this approach is acceptable, but it does have some minor issues. It technically breaks encapsulation so someone might attempt something like this...
```
 blockAddedEvent -> {
   if (protocolSchedule.getScheduledProtocolSpecs().stream()
      .anyMatch(s ->
         s.milestone() == blockAddedEvent.getBlock().getHeader().getTimestamp()
         || s.milestone() == blockAddedEvent.getBlock().getHeader().getNumber())) {
```

which leaks the complexity of numbers vs timestamp outside of the schedule.

We could limit the visibility of getScheduledProtocolSpecs() to be in the same package (although would have to deal with CombinedProtocolSchedule).
We could also leave it only implemented in MutableProtocolSchedule, but that will cause problems when we want to unify the schedules.

gfukushima: Closing this PR in favor of #5159

gfukushima:
Closing this PR in favor of https://github.com/hyperledger/besu/issues/5159

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilder.java
jframe: I like the simplicity of this, but I don't think it's the right way to do this. As was said above we just reimplementing the getByBlockNumber functionality outside of the protocol schedule.

gfukushima: This PR has already been closed in favor of another PR, thanks though!

PR #5202, opened 2023-03-12
Verify default logging is INFO

shemnon:
Add an acceptance test to verify that default logging is at least INFO.


besu/src/main/java/org/hyperledger/besu/Besu.java
```diff
@@ -64,7 +64,6 @@ public static void main(final String... args) {
   private static Logger setupLogging() {
     try {
       // This call is to test if log4j classes are available
```
siladu, line 66: Is this log still relevant?

besu/src/main/java/org/hyperledger/besu/Besu.java
```diff
@@ -64,7 +64,6 @@ public static void main(final String... args) {
...
-      ((Log4J2LoggerFactory) Log4J2LoggerFactory.INSTANCE).newInstance("");
```
siladu, line 67: Not sure why removing this magic fixes it?

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/LoggingTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Test
+  public void testDefaultLoggingIsAtLeastInfo() throws IOException {
+    final BesuNode node = besu.runCommand("");
+    cluster.startConsoleCapture();
+    cluster.runNodeStart(node);
+
+    // Wait for log to contain an info level
+    WaitUtils.waitFor(
+        10, () -> assertThat(cluster.getConsoleContents()).doesNotContain("| INFO |"));
```
siladu, line 37: Although the fix works, this test still passes even with the broken code...The test class itself still prints INFO logs.

siladu (changes requested):

siladu:
@shemnon I played #5206 because users apparently run off main. 

I am struggling to come up with a useful test for this fix though.
Part of the problem may be that the ATs run as a thread rather than a process, however `startConsoleCapture` is not supported by `runBesuAsProcess` mode

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/LoggingTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.io.IOException;
+
+import org.junit.Test;
+
+public class LoggingTest extends AcceptanceTestBase {
+
+  @Test
+  public void testDefaultLoggingIsAtLeastInfo() throws IOException {
+    final BesuNode node = besu.createArchiveNode("logger");
```
siladu, line 30: ah nice, I was thinking the previous incantation was creating logging when it shouldn't but it was the other way around.

Any insight into why this factory method creates logging whereas the other one didn't?
These ATs are still in the realm of magic for me....something about runCommand vs runNodeStart?

besu/src/main/java/org/hyperledger/besu/Besu.java
```diff
@@ -62,7 +62,13 @@ public static void main(final String... args) {
   }
 
   private static Logger setupLogging() {
-    InternalLoggerFactory.setDefaultFactory(Log4J2LoggerFactory.INSTANCE);
+    try {
+      InternalLoggerFactory.setDefaultFactory(Log4J2LoggerFactory.INSTANCE);
+    } catch (Throwable t) {
+      System.out.printf(
+          "Could not set netty log4j logger factory: %s - %s%n",
+          t.getClass().getSimpleName(), t.getMessage());
```
siladu, line 70: If you didn't do this would it continue running besu without error?

Also wondering why we're using a netty factory here and not slf4j/log4j?

siladu (approved):
Just some questions for my own (and any other dev reading this) understanding

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/LoggingTest.java
shemnon: It's magic to me too.  It looks like run command is just there for the help test.  It requires an argument and an empty argument was breaking stuff.

The more reliable way to check for logging is to create a real running instance, this seemed as good as any other.

besu/src/main/java/org/hyperledger/besu/Besu.java
shemnon: I'm not sure why we configure log4j with netty, it's probably some netty magic.

The reason to wrap it in an exception is that I'm setting things up so log4j can be removed at the classpath level and things continue to function.  (a) in case there is another log4shell and (b) in case someone uses the out-of-band slf4j configurations or a wrapper class to point the logging at java.util.logging, logback, google logging, or elsewhere, and then wants to remove unneeded classes

PR #5206, opened 2023-03-13
Fix console logging

siladu:
Based on https://github.com/hyperledger/besu/pull/5202

Getting a fix onto main while we work on a valid acceptance test.


jframe (approved):
LGTM. Was able to manually verify logging is working again now

PR #5221, opened 2023-03-15
Remove getbyblocknumber from flexibleprivacyprecompiledcontract

### LINKED ISSUE #5160: Investigate FlexiblePrivacyPrecompiledContract usage in RunnerBuilder.createPrivateTransactionObserver()
Part of #4789

gfukushima:
## PR description
This PR aims to remove the usage of getByBlockNumber from the FlexiblePrivacyPrecompiledContract. As far as analysis shows the only requirement is to get a protocolSpec (it doesn't matter which one) currently it always fetches the protocolSpec for block number 1 which can vary according to the network config. We're changing that get the protocolSpec for block 0 which is the only block we always have the header.

AT nonmainnet on circle CI
Pipeline - 21048

## Fixed Issue(s)

Fixes #5160 

## Documentation

## Changelog


besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -796,7 +796,7 @@ public Runner build() {
             .build();
     vertx.deployVerticle(filterManager);
 
-    createPrivateTransactionObserver(filterManager, privacyParameters);
+    createPrivateTransactionObserver(filterManager, privacyParameters, context.getBlockchain());
```
siladu, line 799: nit: could just pass through the BlockHeader

siladu (approved):

macfarla (approved):
looks ok to me. block 0 vs block 1 isn't going to make any difference I don't think

besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -1344,17 +1347,19 @@ private void createLogsSubscriptionService(
 
   private void createPrivateTransactionObserver(
```
jframe, line 1348: Don't know how feasible this is, but adding a test for this would be good. But it might already be covered by a test at another level.

PR #5224, opened 2023-03-16
Add getForNextBlockHeader to protocol schedule

jframe:
## PR description
Add getForNextBlockHeader to protocol schedule

## Documentation

## Changelog

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
@@ -75,7 +75,8 @@ public class EthFeeHistory implements JsonRpcMethod {
     final Optional<List<Double>> maybeRewardPercentiles =
         request.getOptionalParameter(2, Double[].class).map(Arrays::asList);
 
-    final long chainHeadBlockNumber = blockchain.getChainHeadBlockNumber();
+    final BlockHeader chainHeadHeader = blockchain.getChainHeadHeader();
+    final long chainHeadBlockNumber = chainHeadHeader.getNumber();
     final long resolvedHighestBlockNumber =
         highestBlock
             .getNumber()
@@ -106,7 +107,7 @@ public class EthFeeHistory implements JsonRpcMethod {
             .map(blockHeader -> blockHeader.getBaseFee().orElse(Wei.ZERO))
             .orElseGet(
                 () ->
-                    Optional.of(protocolSchedule.getByBlockNumber(nextBlockNumber).getFeeMarket())
+                    Optional.of(protocolSchedule.getForNextBlockHeader(chainHeadHeader, chainHeadHeader.getTimestamp()).getFeeMarket())
                         .filter(FeeMarket::implementsBaseFee)
                         .map(BaseFeeMarket.class::cast)
                         .map(
--- a/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
+++ b/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
@@ -17,6 +17,7 @@ package org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.hyperledger.besu.ethereum.core.InMemoryKeyValueStorageProvider.createInMemoryBlockchain;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
@@ -65,7 +66,7 @@ public class EthFeeHistoryTest {
   public void params() {
     final ProtocolSpec londonSpec = mock(ProtocolSpec.class);
     when(londonSpec.getFeeMarket()).thenReturn(FeeMarket.london(5));
-    when(protocolSchedule.getByBlockNumber(anyLong())).thenReturn(londonSpec);
+    when(protocolSchedule.getForNextBlockHeader(any(), anyLong())).thenReturn(londonSpec);
     // should fail because no required params given
     assertThatThrownBy(this::feeHistoryRequest).isInstanceOf(InvalidJsonRpcParameters.class);
     // should fail because newestBlock not given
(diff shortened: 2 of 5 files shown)


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
```diff
@@ -85,10 +86,14 @@ public void params() {
   public void allFieldsPresentForLatestBlock() {
```
siladu, line 86: Reckon it's worth having tests that cover Paris and Shanghai?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/UnsignedIntParameter.java
```diff
@@ -28,6 +28,12 @@ public UnsignedIntParameter(final String value) {
     checkArgument(this.value >= 0);
   }
 
+  @JsonCreator
+  public UnsignedIntParameter(final int value) {
+    this.value = value;
+    checkArgument(this.value >= 0);
+  }
```
siladu, line 35: 👍

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -106,7 +107,11 @@ public JsonRpcResponse response(final JsonRpcRequestContext request) {
             .map(blockHeader -> blockHeader.getBaseFee().orElse(Wei.ZERO))
             .orElseGet(
                 () ->
-                    Optional.of(protocolSchedule.getByBlockNumber(nextBlockNumber).getFeeMarket())
+                    Optional.of(
+                            protocolSchedule
+                                .getForNextBlockHeader(
+                                    chainHeadHeader, chainHeadHeader.getTimestamp())
```
siladu, line 113: Think we discussed having a comment here to explain why we're using head timestamp instead of timestamp for next block?

jframe: We can only estimate the timestamp for the next block by adding the seconds per slot to the head block's timestamp i.e. add 12 when it's mainnet. This is only an estimate though as it's possible that a block could be skipped due to a validator not proposing a block.

I'm using the parent headers timestamp which is wrong but is simple. However, since the fee market doesn't change in Shanghai or Cancun it's adding a new fee market not changing the base fee one we can get away with this simple implementation for now.

The better approach we considered was to check the difference between timestamps. There are many edge cases with this though. We need to consider the possibility of the blocks that are missed. And what to do if we don't have all the block headers.

Perhaps we should just add some predefined constants for the networks though?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/PoWMinerExecutor.java
```diff
@@ -75,7 +75,10 @@ public PoWBlockMiner createMiner(
     final PoWSolver solver =
         new PoWSolver(
             nonceGenerator,
-            protocolSchedule.getByBlockNumber(parentHeader.getNumber() + 1).getPoWHasher().get(),
+            protocolSchedule
+                .getForNextBlockHeader(parentHeader, 0)
+                .getPoWHasher()
+                .get(), // timestamp schedule is never used for pow mining
```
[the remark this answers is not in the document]
jframe: Moved out of the constructor hopefully it's clearer now

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
jframe: No, don't think it's worth doing that. Nothing special about the boundaries as far EthFeeHistory is concerned it just passes down the timestamp.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
siladu: Unfortunately, there's no way of accurately predicting the next block's timestamp, which is a significant difference between block number vs timestamp for switching fork.

e.g. we have a network with 10 slots per epoch, 1 second per slot, genesis @ t = 1 (Epoch 1, Slot 1).
We have a fork configured at Epoch 2, Slot 1 @ t = 11 (forks are currently always configured at the start of an Epoch).
Let's say the proposers miss Epoch 1, Slot 9, 10 and Epoch 2, Slot 1, 2, 3
Last block of previous fork is Epoch 1, Slot 8 @ t = 8; First block of next fork is Epoch 2, Slot 4 @ t = 14.

So Block 8 @ t = 8 -> (fork @ t = 11) -> Block 9 @ t = 14

Even if we know slots per epoch = 1, then we will still predict that Block 9 should be @ t = 9 which won't hit the fork.

A dummy value of 1 minute might be pragmatic and future-proof but feels very hacky 🙈

siladu:
> Is the `ProtocolSchedule` interface going to be removed at some time?

@fab-10 Current plan is to unify MutableProtocolSchedule and DefaultTimestampSchedule. Having a ProtocolSchedule interface may still make sense though, since other schedules exist outside of mainnet.

siladu (approved):
LGTM

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
rolfyone: You'd have to fix both sides of the computation of this I think, which would look something like
 - block timestamp EIP to suggest that block timestamp is always the slot -> timestamp computation, so the EL block timestamp is locked to the slot (predictably computed as expected slot time)
 - then your computation here is trivial, but currently you've theoretically got a 12 sec margin of error... we're asking for a block 4 seconds into the slot before you're computing a block for, and the block could be up to 4 seconds late (then it'd be reorged out if a CL spec change suggestion gets up).
 
 If all EL's could converge on using the timestamp of the block you're proposing for, then this is a trivial thing, but I think ideally it'd be in spec (I haven't checked if it's already specified in that way but i'd be surprised if it is...)

siladu: @fab-10 I'm guessing the reason you need it is related to an optimisation rather than part of the specification itself? I am assuming that if we're querying the txpool then we are building a block and therefore should have a timestamp from the CL.

jframe: In this specific case for eth_feeHistory since the fee market is either pre-london or post-london we don't need to worry about an exact timestamp. But if the fee market is changed in a future fork then I think we will need to have something in a spec.

jframe: I wanted to avoid adding general a getForNextBlockHeader method that doesn't take a timestamp as I don't think there is a simple answer that isn't a hack right now and might depend on the context. The safest way is to always provide the timestamp which we do have when we are proposing a block.

@fab-10 I will leave this open for another day for you to have another look and add comments or block it if you don't think it shouldn't go in as is. I think this is safe for the eth_feeHistory case but we need to come up with a better way for the general case and that might require some spec changes as well.

fab-10: The part I am reviewing is not the block production, but when accepting transactions to the pool, it is an edge case that could only affect fee computation just before a hard fork, but after the hard fork the computation will be fine again, so I am fine with this PR since the only side effect seems to be a non optimal block production in case we are the ones that build the hard fork block.

fab-10 (approved):

PR #5231, opened 2023-03-16
remove parallelism from AT task

macfarla:
This task is now only taking 3 minutes with parallelism at 6. If it takes less than unit tests without parallelism there's no need. 

Occasionally seeing this error on 1 of the 6 runners `No tests found for given includes` 
Because for the parallelism, we're still passing all the AT classes into the splitter, it is possible that none of them match
https://app.circleci.com/pipelines/github/hyperledger/besu/21077/workflows/d91f36d9-7103-44c0-887a-46430e0331b9/jobs/129673/parallel-runs/1?filterBy=FAILED

PR #5232, opened 2023-03-17
Mainnet ATs - reduce parallelism

macfarla:
Reduce parallelism to 4

Occasionally seeing this error on 1 of the 6 runners `No tests found for given includes`
Because for the parallelism, we're still passing all the AT classes into the splitter, it is possible that none of them match eg
https://app.circleci.com/pipelines/github/hyperledger/besu/21077/workflows/d91f36d9-7103-44c0-887a-46430e0331b9/jobs/129673/parallel-runs/1?filterBy=FAILED

Alternative to #5231


siladu (approved):
LGTM

garyschulte:
going to take the liberty of pushing a github action change into this PR to see if we can a) break the logjam and b) get the AT's passing so they do not exacerbate the build problems with flaky fails

macfarla:
closing in favour of #5234

PR #5234, opened 2023-03-17
Mainnet AT reduce parallelism and self-hosted runners

garyschulte:
## PR description
same as #5232 but with additional self-hosted runner config


jflo (approved):
Approving this due to blocked build queues so actions can run on self hosted runners. Will revert this asap after build is done.

PR #5238, opened 2023-03-19
Minor improvements to logging

siladu:
* During startup print "Besu version ..." instead of just "Besu ..." to make searching for the version in the logs easier.

* eth_getLogs now prints request along with exception:
```
2023-03-20 12:42:17.918+10:00 | Test worker | DEBUG  | EthGetLogs | eth_getLogs request JsonRpcRequest{id=null, method='eth_getLogs', params=[FilterParameter{fromBlock=BlockParameter{type=NUMERIC, number=Optional[0]}, toBlock=BlockParameter{type=NUMERIC, number=Optional[50]}, fromAddress=[], toAddress=[], addresses=[], topics=[], maybeBlockHash=Optional.empty, logsQuery=LogsQuery{addresses=[], topics=[], after=Optional.empty, count=Optional.empty, isValid=true}], version='2.0', isNotification=true} failed
java.lang.IllegalArgumentException: Requested range exceeds maximum range limit
	at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.EthGetLogs.lambda$response$3(EthGetLogs.java:90) ~[main/:?]
	at java.util.Optional.orElseGet(Optional.java:364) ~[?:?]
	at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.EthGetLogs.response(EthGetLogs.java:70) ~[main/:?]
	...
```

* Better formatting of the fast sync retry interval: Instead of printing
```
ERROR | FastSyncDownloader | Encountered an unexpected error during fast sync. Restarting sync in PT5S seconds.
```
we now print:
```
ERROR | FastSyncDownloader | Encountered an unexpected error during fast sync. Restarting sync in 5 seconds.
```

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/JsonRpcRequest.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/JsonRpcRequest.java
@@ -136,4 +136,22 @@ public class JsonRpcRequest {
   public <T> Optional<T> getOptionalParameter(final int index, final Class<T> paramClass) {
     return parameterAccessor.optional(params, index, paramClass);
   }
+
+  @Override
+  public String toString() {
+    return "JsonRpcRequest{"
+        + "id="
+        + id
+        + ", method='"
+        + method
+        + '\''
+        + ", params="
+        + Arrays.toString(params)
+        + ", version='"
+        + version
+        + '\''
+        + ", isNotification="
+        + isNotification
+        + '}';
+  }
 }
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetLogs.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetLogs.java
@@ -103,7 +103,11 @@ public class EthGetLogs implements JsonRpcMethod {
                 });
 
     if (ex.get() != null) {
-      LOG.debug("eth_getLogs call failed: ", ex.get());
+      LOG.atDebug()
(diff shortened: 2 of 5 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetLogs.java
```diff
@@ -103,7 +103,11 @@ public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
                 });
 
     if (ex.get() != null) {
-      LOG.debug("eth_getLogs call failed: ", ex.get());
+      LOG.atDebug()
+          .setMessage("eth_getLogs request {} failed: {}")
+          .addArgument(requestContext.getRequest())
```
macfarla, line 108: should there be a second addArgument?

siladu: Good spot, I was printing the exception explicitly but don't need second arg after using the fluent api.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/JsonRpcRequest.java
```diff
@@ -136,4 +136,22 @@
   public <T> Optional<T> getOptionalParameter(final int index, final Class<T> paramClass) {
     return parameterAccessor.optional(params, index, paramClass);
   }
+
+  @Override
+  public String toString() {
+    return "JsonRpcRequest{"
+        + "id="
+        + id
```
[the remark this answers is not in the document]
jframe: Looks like JsonRpcRequestId is missing a toString

macfarla (approved):

PR #5239, opened 2023-03-19
Extract tests for Schedule out of the ScheduleBuilder tests

siladu:


macfarla (approved):

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleTest.java
```diff
@@ -32,6 +38,31 @@
 public class ProtocolScheduleTest {
```
jframe, line 38: nit: rename this test to `MutableProtocolScheduleTest` for consistency?

jframe (approved):
LGTM

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleTest.java
siladu: Agree

PR #5240, opened 2023-03-20
Move protocol schedule streamMilestoneBlocks into test scope

siladu:
It's now only used by the tests following removal of production code paths in https://github.com/hyperledger/besu/commit/08b822a860cd12133fcfa53d6ff0057bc03d56f8 A simple decorater would have been a nicer implementation but I didn't want to expose the protocolSpecs from within the schedule hence why I've used inheritance.

ForksIdsNetworkConfigTest has a particularly hairy construction since it relies on TransitionProtocolSchedule but I've isolated this as an inner class. I am hopeful that we can refactor ForksIdsNetworkConfigTest out of existance eventually and avoid using streamMilestoneBlocks at the same time.

Furthermore, the end goal is have a unified schedule that can handle milestone as blockNumber or timestamp, so we won't need two versions of things such as TimestampStreamingProtocolSchedule


ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/DefaultTimestampScheduleTest.java
```diff
@@ -86,17 +86,6 @@ public void getByBlockHeader_whenSpecNotFoundReturnsNull() {
     assertThat(schedule.getByBlockHeader(BLOCK_HEADER)).isNull();
   }
 
-  @Test
-  public void streamMilestoneBlocksReturnTimestampsInOrder() {
```
gfukushima, line 90: Think this test was important, has it been replace by another one? I think isOnMilestoneBoundary has a different purpose and and would be nice to have a test to always guarantee we are still delivering the forks and the right order.

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/TimestampStreamingProtocolSchedule.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class TimestampStreamingProtocolSchedule extends DefaultTimestampSchedule
+    implements MilestoneStreamingProtocolSchedule {
+
+  public TimestampStreamingProtocolSchedule(final DefaultTimestampSchedule timestampSchedule) {
+    super(timestampSchedule);
+  }
+
+  @Override
+  public Stream<Long> streamMilestoneBlocks() {
```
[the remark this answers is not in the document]
siladu: > abstract the concept of MilestoneBlock
Not sure I follow exactly.

`streamMilestoneBlocks`: milestone === fork (but milestone name used in the ProtocolSchedule classes)
`streamMilestones` would be fine too but they do still refer to blocks, whether it's number or timestamp based, so the Blocks suffix is accurate if a little redundant. 

I do wonder if the original author was using Block to mean BlockNumber? So `streamMilestones` or `streamMilestoneNumber` might be better here.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/DefaultTimestampScheduleTest.java
siladu: streamMilestoneBlocks isn't used in production any more, so do we want tests for all of our test-support code is the really the question here...

gfukushima: I agree, we should not test a method that is not being used anymore, but how do we guarantee the forks are in the right order now? Is there a separate test for that somewhere  else?

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/TimestampStreamingProtocolSchedule.java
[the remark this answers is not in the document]
gfukushima: I think I was one who was always using the milestone word but maybe fork was a better word to use even in the ProtocolSpec class.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/DefaultTimestampScheduleTest.java
siladu: Yes, the order is tested in the ScheduleBuilderTests: https://github.com/hyperledger/besu/blob/main/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/TimestampScheduleBuilderTest.java#L61
https://github.com/hyperledger/besu/blob/main/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/TimestampScheduleBuilderTest.java#L104

There is also some validation in the code itself: https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/TimestampScheduleBuilder.java#L56

The above test was covering that streamMilestoneBlocks returned them in the same order.

gfukushima (approved):
lgtm

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/TimestampStreamingProtocolSchedule.java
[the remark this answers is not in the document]
siladu: Yeh both are used pretty liberally in the code...and also https://ethereum.org/en/history/ 😆

jframe (approved):

Issue #5098, opened 2023-02-16
Add detail messages to particular cases for INVALID_PARAMS


macfarla:
"invalid params" error code is used a lot, but always with the same message, which can hide a lot of detail. 

eg for #4212 it's hard to get any clues as to the nature of the problem. Could add further values to the JsonRpcError enum with the same code -32062 but with different messages that might help the user diagnose or at least report issues
eg
missing param
wrong type for param
unable to decode param (eg RLP)

JsonRpcError
  INVALID_PARAMS(-32602, "Invalid params"),

macfarla:
Another instance  - when transaction type is incorrect, eth_sendRawTransaction can respond with invalid params but it's not obvious what the problem is https://discord.com/channels/905194001349627914/938504958909747250/1108931495797727354

chenqping:
yes, invalid params don't indication where is wrong

macfarla:
changed default to internal error not invalid params https://github.com/hyperledger/besu/pull/5506

macfarla:
suggest to follow the pattern used here for peer disconnect reasons https://github.com/hyperledger/besu/pull/6899

Issue #4212, opened 2022-08-03
Intermittent 400 Bad Request when sending a transaction


samsondav:
We are seeing intermittent errors trying to send transactions on Besu (consensus client - teku). It only occurs on some RPC nodes and not all the time.

We are using `eth_sendRawTransaction` call.

Occasionally, we see this response:

`400 Bad Request: {\n  \"jsonrpc\" : \"2.0\",\n  \"id\" : 60246,\n  \"error\" : {\n    \"code\" : -32602,\n    \"message\" : \"Invalid params\"\n  }\n}`

This is a blocker for us to run Chainlink on Besu. We need reliable transaction sending.

Could be related to this issue? https://github.com/hyperledger/besu/issues/2251

jflo:
Can you provide an example call please? A curl statement would be fine.

yorickdowne:
I'm going to play go-between here. "Literally just send any transaction" ... I assume if you just send enough tx, do include those that call contracts, you'll see the issue.

lyxer123:
In response to the above situation, this happens to my program from time to time, but after I use the metamask wallet to send data once, my original program can work again. (my program works the same as metamask)

lyxer123:
I use 4G wireless module to post the eth_sendRawTransaction：this error  {\n \"jsonrpc\" : \"2.0\",\n \"id\" : 60246,\n \"error\" : {\n \"code\" : -32602,\n \"message\" : \"Invalid params\"\n }\n}  occured many times。

when error occured，I use metamask to send a Transaction using same wallet address,then the 4G wireless module can send data to besu with the  same wallet address.

macfarla:
Hi @samsondav we are taking a look at RPC issues this sprint. Are you able to provide any more context re the intermittency of this issue - anything unusual about how you're sending the transactions in general, or the ones that fail? 

We're also working on adding more info to the "invalid params" error logging so hopefully that may assist #4326

samsondav:
Don't have any more context, it seems completely random, but if you send enough tx you will see it.

baxterjfinch:
I am also seeing this behavior intermittently on our private 4 node QBFT network. We load test ~10,000 transactions of the same contract interaction type, and we see about 4 of these `400` `Invalid params` error returned.


```
"rawTx": {
    "type": "0x0",
    "nonce": "0x5f32",
    "gasPrice": "0x3e8",
    "maxPriorityFeePerGas": null,
    "maxFeePerGas": null,
[...]

PR #5241, opened 2023-03-20
Invalid params data

macfarla:
Added `data` to error in case of INVALID_PARAMS to give users more info as to what's gone wrong

See #4212 - possible solution to #5098

Not sure if this will break some hive tests?

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/BaseJsonRpcProcessor.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/BaseJsonRpcProcessor.java
@@ -41,8 +41,10 @@ public class BaseJsonRpcProcessor implements JsonRpcProcessor {
     try {
       return method.response(request);
     } catch (final InvalidJsonRpcParameters e) {
+      JsonRpcError invalidParamsError = JsonRpcError.INVALID_PARAMS;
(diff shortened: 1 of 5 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/execution/BaseJsonRpcProcessor.java
```diff
@@ -41,8 +41,10 @@ public JsonRpcResponse process(
     try {
       return method.response(request);
     } catch (final InvalidJsonRpcParameters e) {
+      JsonRpcError invalidParamsError = JsonRpcError.INVALID_PARAMS;
+      invalidParamsError.setData(e.getMessage());
```
siladu, line 45: This doesn't touch the Engine API does it?
I think error messages are part of the spec in some cases for that: https://github.com/ethereum/execution-apis/blob/main/src/engine/common.md#errors

siladu (commented):
Do we need to do something similar for web sockets or graphQL? Maybe it'll work for those already, I just noticed only http tests were updated

macfarla:
On further inspection, calling setData() on an enum value seems like a dodgy hack so I think we need to refactor JsonRpcError to separate these two concepts. 

Closing this PR, needs further thought. Created an issue https://github.com/hyperledger/besu/issues/5437

PR #5244, opened 2023-03-20
Metrics via dagger

jflo:
Introduces Dagger to provide dependencies.  

- Migrates an existing MetricsSystem module up into the metrics gradle module.
- Makes BesuCommand accessible from Dagger components, to enable configuration reuse.
- Uses CachedMerkleTrieLoader as an example of how to get and use a MetricSystem.


besu/src/main/java/org/hyperledger/besu/Besu.java
```diff
@@ -14,13 +14,9 @@
...
 package org.hyperledger.besu;
 
-import org.hyperledger.besu.chainexport.RlpBlockExporter;
-import org.hyperledger.besu.chainimport.JsonBlockImporter;
-import org.hyperledger.besu.chainimport.RlpBlockImporter;
 import org.hyperledger.besu.cli.BesuCommand;
 import org.hyperledger.besu.cli.logging.BesuLoggingConfigurationFactory;
-import org.hyperledger.besu.controller.BesuController;
-import org.hyperledger.besu.services.BesuPluginContextImpl;
+import org.hyperledger.besu.components.DaggerBesuComponent;
```
siladu, line 19: Presume this is generated by Dagger?

besu/src/main/java/org/hyperledger/besu/components/BesuComponent.java
```diff
@@ -0,0 +1,69 @@
...
+
+/** An application context that knows how to provide dependencies based on Dagger setup. */
+@Singleton
+@Component(
+    modules = {
+      BesuCommandModule.class,
+      MetricsSystemModule.class,
+      CachedMerkleTrieLoaderModule.class
+    })
+public interface BesuComponent {
```
siladu, line 39: BesuComponent seems quite generic to me. Could/should we be more precise with the name? Or does it refer to the one and only component for the `besu` gradle module? 
Will this end up being _the_ application context, or one of many (e.g. one per module)?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1333,8 +1334,15 @@ static class TxPoolOptionGroup {
   private Collection<EnodeURL> staticNodes;
   private BesuController besuController;
   private BesuConfiguration pluginCommonConfiguration;
+
+  private BesuComponent besuComponent;
```
siladu, line 1338: What's the difference between BesuComponent and DaggerBesuComponent?

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -461,6 +476,7 @@ public BesuControllerBuilder chainPruningConfiguration(
    * @return the besu controller
    */
   public BesuController build() {
+    // checkNotNull(besuComponent, "Dagger should have provided a BesuComponent");
```
siladu, line 479: comment should be removed?

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -493,7 +509,10 @@ public BesuController build() {
             reorgLoggingThreshold,
             dataDirectory.toString());
 
-    final CachedMerkleTrieLoader cachedMerkleTrieLoader = new CachedMerkleTrieLoader(metricsSystem);
+    final CachedMerkleTrieLoader cachedMerkleTrieLoader =
+        besuComponent == null
```
siladu, line 513: In what scenario would besuComponent be null here?

besu/src/test/java/org/hyperledger/besu/cli/CommandTestAbstract.java
```diff
@@ -167,6 +168,8 @@ public abstract class CommandTestAbstract {
   @Mock
   protected Logger mockLogger;
 
+  @Mock protected BesuComponent mockContext;
```
siladu, line 171: mix of 'Context' and 'Component' being used in the var names

ethereum/referencetests/src/reference-test/external-resources
siladu, line 1: see my message about fixing this on #besu-contributors

gradle.properties
```diff
@@ -2,7 +2,7 @@ version=23.1.3-SNAPSHOT
 
 org.gradle.welcome=never
 # Set exports/opens flags required by Google Java Format and ErrorProne plugins. (JEP-396)
-org.gradle.jvmargs=-Xmx4g \
+org.gradle.jvmargs=-Xmx5g \
```
siladu, line 5: keep cranking it!! 😆

metrics/core/src/main/java/org/hyperledger/besu/metrics/MetricsConfigurationModule.java
```diff
@@ -11,27 +11,23 @@
...
-import org.hyperledger.besu.metrics.MetricsSystemFactory;
 import org.hyperledger.besu.metrics.prometheus.MetricsConfiguration;
-import org.hyperledger.besu.plugin.services.MetricsSystem;
 
 import javax.inject.Singleton;
 
 import dagger.Module;
 import dagger.Provides;
 
-@SuppressWarnings("WeakerAccess")
```
siladu, line 28: Intrigued to know why you were able to delete this

siladu (commented):
Looks pretty good to me, but I haven't fully got my head around Dagger yet so this is a first pass. Sorry for all the questions :)

Love the fact this is targetting BesuCommand...I believe that's the most coupled class?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
jflo: BesuComponent is ours, the Dagger_ prefix indicates it is the implementation that Dagger generates for us during precompile. At some point, once we've "daggerized all the things" we won't need to use these because Dagger will more fully compose the object graph.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
jflo: Good catch. This is leftover from a prior revision where we were passing in the BesuComponent after construction time.

besu/src/test/java/org/hyperledger/besu/cli/CommandTestAbstract.java
jflo: This was intentional per the description of Component above, but I am happy to change if this is misleading.  Eventually we will introduce subcomponents, which I would name something like P2PComponent, but then refer to the top-level component as a single application context.

metrics/core/src/main/java/org/hyperledger/besu/metrics/MetricsConfigurationModule.java
jflo: It moved packages, now the module is in the same package that MetricsConfiguration is in.

besu/src/main/java/org/hyperledger/besu/components/BesuComponent.java
siladu: Any reason not to call the whole class ApplicationContext? Or is that to Spring-y 😛

besu/src/test/java/org/hyperledger/besu/cli/CommandTestAbstract.java
siladu: I guess I'd favour whatever is idiomatic for Dagger. My objection here is about having two variable names for the same thing, would prefer if we stick with one or the other everywhere.

siladu: I'd be fine with any of mockContext, mockBesuContext, mockApplicationContext, mockBesuComponent, as long as we use the same in the prod code.

siladu:
@jflo I attempted a second pass after reading up on Dagger, but I think the recent merge was dodgy and has polluted the PR files so it's hard to review now.

Are you back to working on it or is it still ready for review?

besu/src/main/java/org/hyperledger/besu/components/BesuComponent.java
jflo: Yes, trying to adopt Daggers convention

siladu (commented):
Forgot to submit this comment yesterday

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/cache/CachedMerkleTrieLoaderModule.java
garyschulte, line 1: does this need to be in this package for visibility reasons?  could we put all of the modules in the besu project so we can find them?

garyschulte (approved):
Love it.  as mentioned in comment, I wish we had a way to collect all of the module definitions though.  It would be wise to get a handful of approvals before merging since this is a 'whole repo' kind of change, even if incremental

besu/src/main/java/org/hyperledger/besu/components/BesuComponent.java
```diff
@@ -0,0 +1,69 @@
...
+import dagger.Component;
+import org.slf4j.Logger;
+
+/** An application context that knows how to provide dependencies based on Dagger setup. */
+@Singleton
+@Component(
+    modules = {
+      BesuCommandModule.class,
+      MetricsSystemModule.class,
+      CachedMerkleTrieLoaderModule.class
```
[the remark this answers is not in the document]
siladu: Would love to get more maintainers eyes on this PR, particularly regarding this question, before this is merged and a precedent for how to design modules is set :)

besu/src/test/java/org/hyperledger/besu/cli/CommandTestAbstract.java
macfarla: mockBesuComponent makes sense to me

macfarla (commented):
still getting my head around dagger also. This change doesn't get us out of passing the metricsSystem around eg to transactionPool - it just makes setup easier?

gfukushima (commented):
Also still getting my head around dagger. Could even try implementing another module. Do you have any suggestions of what would be a good one to get started with to familiarise with dagger?

besu/src/main/java/org/hyperledger/besu/Besu.java
jflo: Yes, see explanation in later comment.

besu/src/main/java/org/hyperledger/besu/components/BesuComponent.java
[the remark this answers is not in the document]
jflo: Devs should DEFINITELY not assume this establishes a precedent of any kind. The right place to do this could be a judgement call. If something is worth being Daggerized within its submodule (evmtool is an example) then the dagger stuff should live right next to its implementation. In other cases, there could be a module that is only ever useful in a BesuContext (like configuration related things).

jflo:
> still getting my head around dagger also. This change doesn't get us out of passing the metricsSystem around eg to transactionPool - it just makes setup easier?

Correct. Once transactionPool is itself constructed by Dagger, then we no longer need to pass it or the BesuContext around ourselves.

jflo:
> Do you have any suggestions of what would be a good one to get started with to familiarise with dagger?

The ideal candidate would be something used rarely, but deeeeeply within the overall object graph.

PR #5245, opened 2023-03-20
use tags compatible with github runners

garyschulte:
## PR description
revert GHA tags to ones that will trigger with github runners, also tag jobs which require x86 architecture with 'x64'


macfarla (approved):

garyschulte:
seeing the build failures with the self hosted runners that are picking up jobs, I am inclined to not merge this change

PR #5247, opened 2023-03-21
Plugin API -  Blockchain Service 

Gabriel-Trintinalia:
## PR description
Introduces a new feature to the plugin API, enabling third-party plugins to access the blockchain service and retrieve information about a specific block based on its number.


macfarla (approved):
LGTM

PR #5251, opened 2023-03-21
Merkle trie radix abstraction

garyschulte:
## PR description

Abstract the MerklePatriciaTrie to support tries with a different radix


matkt (approved):
Added a commit to clean some part of the code and reverting the reference test modification . LGTM

PR #5253, opened 2023-03-22
Cleanup after GHA docker user

garyschulte:
## PR description
Docker actions that write to files during actions leave files in the repository owned by the docker user rather than the runner user.  This leaves the runner filesystem in a state that makes non-docker actions fail.


siladu (approved):
LGTM

Issue #5164, opened 2023-03-05
Consider deleting getByBlockNumber usage in CombinedProtocolScheduleFactory


jframe:
Part of #4789

gfukushima:
Delete still an option but #5162 might mean we can use the new BFT schedule

PR #5254, opened 2023-03-22
Isolate bft getByBlockNumber

### LINKED ISSUE #5162: BFT sequence number usage
Part of #4789

siladu:
Fixes #5162 and #5164 

Introduce BftProtocolSchedule with a copy of MutableProtocolSchedule.getByBlockNumber
This method will be deleted from the ProtocolSchedule interfaces and so will only exist in BftProtocolSchedule (although it may be renamed in a subsequent commit).

Rename existing *BftProtocolSchedule classes to append Builder for clarity (and inline with other ProtocolScheduleBuilder classes)

### DIFF
--- a/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilder.java
+++ b/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilder.java
@@ -36,7 +36,7 @@ import java.util.Map;
 import java.util.function.Function;
 
 /** Defines the protocol behaviours for a blockchain using a BFT consensus mechanism. */
-public abstract class BaseBftProtocolSchedule {
+public abstract class BaseBftProtocolScheduleBuilder {
 
   private static final BigInteger DEFAULT_CHAIN_ID = BigInteger.ONE;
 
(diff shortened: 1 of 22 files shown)


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BftProtocolSchedule.java
```diff
@@ -0,0 +1,46 @@
...
+import org.hyperledger.besu.ethereum.mainnet.ScheduledProtocolSpec;
+
+public class BftProtocolSchedule extends MutableProtocolSchedule {
+
+  public BftProtocolSchedule(final MutableProtocolSchedule mutableProtocolSchedule) {
+    super(mutableProtocolSchedule);
+  }
+
+  @Override
+  public ProtocolSpec getByBlockNumber(final long number) {
```
siladu, line 31: Should we rename this or does it still make sense?

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilder.java
```diff
@@ -72,15 +72,18 @@ public ProtocolSchedule createProtocolSchedule(
...
-            config,
-            DEFAULT_CHAIN_ID,
-            specAdapters,
-            privacyParameters,
-            isRevertReasonEnabled,
-            config.isQuorum(),
-            evmConfiguration)
-        .createProtocolSchedule();
+    final MutableProtocolSchedule mutableProtocolSchedule =
+        (MutableProtocolSchedule)
```
jframe, line 76: Bit nasty need to do this casting, but don't see any way around it

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BftProtocolSchedule.java
jframe: Still make sense, don't think we need to rename it

jframe (approved):

jframe (approved):
LGTM

Issue #5260, opened 2023-03-23
Unify Protocol Schedule for Block Number and Timestamp


siladu:
- [x] Migrate usages of getByTimestamp -> getByBlockHeader
- [ ] Unify MutableProtocolSchedule and DefaultTimestampSchedule. Make getByBlockHeader call private getByTimestamp orElse getByBlockNumber
- [ ] Unify StreamingProtocolSchedules and extend new 'ProtocolSchedule'
- [ ] BftProtocolSchedule extends new 'ProtocolSchedule'
- [ ] Unstitch in TransitionProtocolSchedule?

Should end up with this hierarchy:

```
                            ----> PrivacySupportingProtocolSchedule 
                           |
                           | 
         ----->    ProtocolSchedule         <------
        |          - public getByBlockHeader       |
        |          - private getByTimestamp        |
        |          - private getByBlockNumber      |
        |                                          |
        |                                          |
BftProtocolSchedule                MilestoneStreamingProtocolSchedule
 - public getByBlockNumber          - public streamMilestones
```

PR #5261, opened 2023-03-24
Replace usage of getByTimestamp in WithdrawalsValidatorProvider

siladu:
Use schedule.getByBlockHeader for newPayload, constructing a dummy header for the sole purpose of getting the correct WithdrawalsValidator for this block. See reviews comments below for context.

Use schedule.getForNextBlockHeader for forkchoiceUpdated.

Once we remove usages of schedule.getByTimestamp, we can unify the protocol schedules.

Part of #5260


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -104,8 +104,13 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
     final Optional<List<Withdrawal>> maybeWithdrawals =
         Optional.ofNullable(blockParam.getWithdrawals())
             .map(ws -> ws.stream().map(WithdrawalParameter::toWithdrawal).collect(toList()));
-    if (!getWithdrawalsValidator(timestampSchedule, blockParam.getTimestamp())
-        .validateWithdrawals(maybeWithdrawals)) {
+    final Optional<BlockHeader> maybeParentHeader =
+        protocolContext.getBlockchain().getBlockHeader(blockParam.getParentHash());
+    boolean ableToValidateWithdrawals = maybeParentHeader.isPresent();
```
siladu, line 109: This does change behaviour such that for an "invalid withdrawals"-block, we will end up returning SYNCING if the parent header is missing. Before this PR, we would return an "Invalid params" error, however that may be incorrect - I am waiting fir clarification about the spec w.r.t syncing.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -104,8 +104,13 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+    if (ableToValidateWithdrawals
```
siladu, line 110: This does have a side-effect change of behaviour such that for an "invalid withdrawals"-block, we will end up returning SYNCING if the parent header is missing. Before this PR, we would return an "Invalid params" error, however that may be incorrect - I am waiting fir clarification about the spec w.r.t syncing.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/WithdrawalsValidatorProvider.java
```diff
@@ -15,17 +15,22 @@
...
   static WithdrawalsValidator getWithdrawalsValidator(
-      final TimestampSchedule timestampSchedule, final long newPayloadTimestamp) {
+      final TimestampSchedule timestampSchedule,
+      final BlockHeader parentBlockHeader,
+      final long newPayloadTimestamp) {
 
-    return timestampSchedule
-        .getByTimestamp(newPayloadTimestamp)
+    return Optional.ofNullable(
+            timestampSchedule.getForNextBlockHeader(parentBlockHeader, newPayloadTimestamp))
```
siladu, line 33: Using getForNextBlockHeader makes sense for forkchoiceUpdated.

For newPayload, it's a bit dubious due to the parentHeader change in behaviour mentioned above. 
An alternative for newPayload would be to use getByBlockHeader and construct a dummy BlockHeader instead, which would require a second version of getWithdrawalsValidator.

We can of course use the BlockHeader that is constructed later on in newPayload, however it's a bit of a chicken and egg because it takes the withdrawals as a parameter and also required changing the order that we respond with INVALID in newPayload which I wanted to avoid if possible. Think a dummy header is preferable over this, since we can't guarantee validity of the whole header at this point.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
siladu: I got some clarification on syncing...on the one hand it makes sense to defer SYNCING until after we've performed all possible validation, bearing in mind not all validation can take place without the parent header.
On the other hand, there might be a performance trade-off: if a CL is spamming us with requests then breaking out early to respond with SYNCING may help.

Performance aside, I think the proper order is to respond with INVALID, if not SYNCING, so currently thinking I shouldn't rely on the parent header being present to validate withdrawals

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/WithdrawalsValidatorProvider.java
jframe: Agree that for FCU it makes sense to use getForNextBlockHeader.

For newPayload would be better to use the actual header we construct, but we could get away with a dummy header. If we are using a dummy header though using getForNextBlockHeader with the new timestamp and chain head would also work since we only use the timestamp from that block in the check.

ethereum/referencetests/src/reference-test/external-resources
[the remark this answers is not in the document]
siladu: Nope - good spot, that keeps trying to sneak in!

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/WithdrawalsValidatorProvider.java
siladu: As discussed, plan to use getForNextBlockHeader for FCU and getByBlockHeader(dummyHeader) for newPayload.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/WithdrawalsValidatorProvider.java
```diff
@@ -15,17 +15,42 @@
...
+            .timestamp(blockTimestamp)
+            .number(blockNumber)
+            .buildBlockHeader();
+    return Optional.ofNullable(timestampSchedule.getByBlockHeader(blockHeader))
+        .map(ProtocolSpec::getWithdrawalsValidator)
+        // TODO Withdrawals this is a quirk of the fact timestampSchedule doesn't fallback to the
+        // previous fork. This might be resolved when
+        // https://github.com/hyperledger/besu/issues/4789 is played
+        // and if we can combine protocolSchedule and timestampSchedule.
+        .orElseGet(WithdrawalsValidator.ProhibitedWithdrawals::new);
```
jframe, line 44: Can the duplicated code between this function and the other getWithdrawalsValidator function be moved into a common function

jframe, line 44: This fallback will only work falling back from Shanghai. I guess that's okay though as long as 4789 is done before Cancun.

siladu: As eluded to in the TODO comment, this fallback will disappear when we unify the schedules in the next PR

siladu: See above

jframe (approved):

PR #5266, opened 2023-03-24
reset cache after heal

matkt:
## PR description

We found that sometimes after the heal we still have old snapshot in the cache because after the clearTrieLog some transaction pool thread are still running 
The fix is to clear the cache after the snapsync and also add a snapshot of the  new head


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -338,6 +338,9 @@ public TrieLogManager getTrieLogManager() {
   @Override
   public void setArchiveStateUnSafe(final BlockHeader blockHeader) {
```
garyschulte, line 339: this method name should change to reflect its use.  IMO it isn't setting the archive unsafe for use like the name implies, but rather is resetting the archive cache and adding the new pivot as the only entry.  Perhaps `resetArchiveStateTo(BlockHeader)`

Similarly, BonsaiWorldState.setArchiveUnsafe should be something more along the lines of `resetWorldStateTo(BlockHeader)`

These methods could use javadoc describing the intent

garyschulte (changes requested):
LGTM, naming and documenting intent is probably wise considering the effort went into making bonsai more dev friendly

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
matkt: updated the code

garyschulte (approved):
🚢

PR #5269, opened 2023-03-24
Update most dependecies to latest version

fab-10:
## PR description

Vertx, Prometheus and Mockito are not update to latest version since there are some breaking changes and we need to update our code, will track their update in separate tickets.


jflo (approved):
Since this updates a number of dependencies, perhaps it is about time to re-generate verification-metadata.xml (As described in https://docs.gradle.org/current/userguide/dependency_verification.html#sec:verification-metadata-hygiene) from scratch to clean out prior versions.

fab-10:
> Since this updates a number of dependencies, perhaps it is about time to re-generate verification-metadata.xml (As described in https://docs.gradle.org/current/userguide/dependency_verification.html#sec:verification-metadata-hygiene) from scratch to clean out prior versions.

There is some value in keeping the verification for older version,, in the case we need to rollback an update, we alreadu have the right checksum w/o having to trust it again, but this should be very rare and so I agree that make sense to do a cleanup now.

PR #5274, opened 2023-03-28
continue despite test failures

jflo:
I would like to propose some blasphemy:

we should run the weekly sonarqube job with gradle --continue so it proceeds on, regardless of test failure.

this week a flaky test failed, and so we got no analysis.
the sonarqube job produces no artifacts of any use, it is meant as a weekly diagnostic, so while blasphemous, I believe this to be safe.


macfarla (approved):
LGTM

PR #5275, opened 2023-03-29
[MINOR] Add log to show where we're reading the jwt from

gfukushima:
## PR description
Add a log to print the path where we loaded the engine jwt for better ux.

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -2370,7 +2370,6 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
       if (Objects.nonNull(engineRPCOptionGroup.engineJwtKeyFile)
           && java.nio.file.Files.exists(engineRPCOptionGroup.engineJwtKeyFile)) {
         engineConfig.setAuthenticationPublicKeyFile(engineRPCOptionGroup.engineJwtKeyFile.toFile());
-        logger.info("Engine JWT loaded from: {}", engineRPCOptionGroup.engineJwtKeyFile);
       } else {
         logger.warn(
             "Engine API authentication enabled without key file. Expect ephemeral jwt.hex file in datadir");
@@ -3635,6 +3634,11 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
       builder
           .setEnginePort(engineJsonRpcConfiguration.getPort())
           .setEngineApis(engineJsonRpcConfiguration.getRpcApis());
+    }
+    if (engineJsonRpcConfiguration.isAuthenticationEnabled()
+        && engineJsonRpcConfiguration.getAuthenticationPublicKeyFile() != null) {
+      builder.setEngineJwtFile(
+          engineJsonRpcConfiguration.getAuthenticationPublicKeyFile().getAbsolutePath());
     }
 
     if (rocksDBPlugin.isHighSpecEnabled()) {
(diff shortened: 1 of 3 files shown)


besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
```diff
@@ -142,13 +143,24 @@ public ConfigurationOverviewBuilder setEngineApis(final Collection<String> engin
...
     return this;
   }
 
+  /**
+   * Sets the engine jwt file path.
+   *
+   * @param engineJwtFilePath the engine apis
+   * @return the builder
+   */
+  public ConfigurationOverviewBuilder setEngineJwtFile(final String engineJwtFilePath) {
```
siladu, line 159: nit: could use Path or File here to make this even safer, e.g. if somehow the secret token ended up getting passed in by mistake

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -3630,10 +3630,17 @@ private String generateConfigurationOverview() {
           .setRpcHttpApis(jsonRpcConfiguration.getRpcApis());
     }
 
-    if (engineJsonRpcConfiguration != null && engineJsonRpcConfiguration.isEnabled()) {
-      builder
-          .setEnginePort(engineJsonRpcConfiguration.getPort())
-          .setEngineApis(engineJsonRpcConfiguration.getRpcApis());
+    if (engineJsonRpcConfiguration != null) {
+      if (engineJsonRpcConfiguration.isEnabled()) {
```
siladu, line 3634: Why are these broken up?

I'm struggling to imagine a scenario where jwt token is needed without the engine api being enabled.

ajsutton (approved):
LGTM.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -3630,10 +3630,17 @@ private String generateConfigurationOverview() {
...
+        builder
+            .setEnginePort(engineJsonRpcConfiguration.getPort())
+            .setEngineApis(engineJsonRpcConfiguration.getRpcApis());
+      }
+      if (engineJsonRpcConfiguration.isAuthenticationEnabled()
+          && engineJsonRpcConfiguration.getAuthenticationPublicKeyFile() != null) {
```
siladu, line 3640: If someone mistakenly doesn't configure it, what will we print? I think there may be some other error so probably ok

gfukushima: Prints nothing, engineJsonRpcConfiguration.getAuthenticationPublicKeyFile() will be null

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
gfukushima: Think you're right. Refactored this. Thank you

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
siladu: ah, it generates one and logs a warning:
`2023-03-29 13:31:29.817+10:00 | main | WARN  | Besu | Engine API authentication enabled without key file. Expect ephemeral jwt.hex file in datadir`

Do you think we should also print out the location of this ephemeral one in the configuration overview?

siladu (approved):
LGTM

Issue #5249, opened 2023-03-21
QBFT validator smart contract method does not work with london fork with Fees.


D3athGr1p:
I want to use EIP-1559 fee stucture but when i try using london fork it does not work with validator smart contract. it works fine with blockheader selection method. i tried both way using ```londonBlock: 0``` and 
```
"transitions": {
      "qbft": [
        {
          "block": 25,
          "validatorselectionmode": "contract",
          "validatorcontractaddress": "0x42699A7612A82f1d9C36148af9C77354759b210b"
        }
      ]
    }
```

but i'm getting this error
![image](https://user-images.githubusercontent.com/86771928/226526894-06e906c4-647c-46fe-99a0-1b840f7b60b9.png)

as for genesis file  :
```
{
  "config": {
    "chainId": 878145,
    "londonBlock": 0,
    "berlinBlock": 0,
    "qbft": {
      "blockperiodseconds": 2,
      "epochlength": 30000,
      "requesttimeoutseconds": 4,
      "blockreward": "1500000000000000000"
    },
    "transitions": {
      "qbft": [
        {
          "block": 50,
          "validatorselectionmode": "contract",
          "validatorcontractaddress": "0xb9A219631Aed55eBC3D998f17C3840B7eC39C0cc"
        }
      ]
    }
  },
  "nonce": "0x0",
  "timestamp": "0x58ee40ba",
  "gasLimit": "0x47b760",
  "difficulty": "0x1",
  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "coinbase": "0x0000000000000000000000000000000000000000",
  "alloc": {
    "f17f52151EbEF6C7334FAD080c5704D77216b732" : {
      "balance" : "0x446c3b15f9926687d2c40534fdb56400"
    }
  },
  "extraData": "0xf87aa00000000000000000000000000000000000000000000000000000000000000000f8549447349b0e16718f7439261616c48fd1b92a9d64d7945c2ec9937230a20ab9a6008bc32952d0551c14fc94903e1eedcb60cc27a386ee2879274b2a445bd72a9457a71c05d16403165c21ce27914573ef5ffad819c080c0"
}
```

I also tried londonBlock at 55 and my `"validatorselectionmode": "contract"` at block 50 but still got same issue.

as for validator smart contract I'm using [ConsenSys](https://github.com/ConsenSys/validator-smart-contracts) and this works fine without london block

notatestuser:
[...]

PR #5276, opened 2023-03-29
Fix BFT creating invalid blocks on London when zero base fee is used

jframe:
## PR description
This fixes an issue where on IBFT and QBFT invalid blocks are created by the proposer when using London with a zeroBaseFee.

The cause of this is that the ZeroBaseFeeMarket was returning true for the implementsDataFee check which meant that the excessDataGas field was added to the block header. This is only valid after Shanghai.

## Fixed Issue(s)

fixes #5270

PR #5277, opened 2023-03-29
Make QBFT validator smart contract mode work with london fork

siladu:
This should work when gas fees are used as well as for free gas networks

Override the transactionSimulator's default TransactionValidationParams with one that allows for exceeding the account balance (which effectively zeros the baseFee). This mimics the way that eth_estimateGas and eth_call are implemented.

Fixes https://github.com/hyperledger/besu/issues/5249


siladu:
> I have considered running ./gradlew acceptanceTestNonMainnet locally if my PR affects non-mainnet modules

I would like to run this but it doesn't work locally (maybe an M1 issue?)

matkt (changes requested):
Is the reference tests modification normal ?

siladu:
> Is the reference tests modification normal ?

Good spot. No I can't seem to shake that from my PRs...or undo it!

siladu:
> > Is the reference tests modification normal ?
> 
> Good spot. No, I didn't want to commit it...I can't seem to shake that from my PRs...or undo it!

Actually, I do want to commit it because other PRs had reverted it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/ZeroBaseFeeMarket.java
```diff
@@ -38,9 +38,4 @@ public Wei computeBaseFee(
   public ValidationMode baseFeeValidationMode(final long blockNumber) {
     return ValidationMode.NONE;
   }
-
-  @Override
-  public boolean implementsDataFee() {
-    return true;
-  }
```
siladu, line 45: Had to do this to get `retrievesValidatorsFromValidatorContract_LondonFork_ZeroBaseFee` passing. It's stolen from https://github.com/hyperledger/besu/pull/5276

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
```diff
@@ -263,10 +277,19 @@ public TestContext build() {
     if (genesisFile.isPresent()) {
       try {
         final GenesisState genesisState = createGenesisBlock(genesisFile.get());
+        Block genesisBlock = genesisState.getBlock();
```
jframe, line 280: Could just set the base fee in the genesis block of the genesis file instead of overriding it here

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
```diff
@@ -398,7 +425,14 @@ private static ControllerAndState createControllerAndFinalState(
             : Collections.emptyMap();
     final QbftConfigOptions qbftConfigOptions = createGenesisConfig(useValidatorContract);
 
-    genesisConfigOptions.byzantiumBlock(0);
+    if (useLondonMilestone) {
+      genesisConfigOptions.londonBlock(0);
+    } else {
+      genesisConfigOptions.berlinBlock(0);
+    }
```
jframe, line 432: Might be simpler to just use the MainnetProtocolSchedule with the genesis file instead. I think this and the above comment about zero base fees come down to how we want to write these tests. Would we rather create custom genesis files or use a generic genesis file with config in the tests?

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
siladu: I avoided setting it in the genesis since that's not how zeroBaseFee config works (it kicks in post-genesis for the next block you create). The only impact this has is on the error you get when the transaction simulator configuration fix isn't in place.

However, I've just realised there's an edge case if you have a network with validator contract at genesis: we attempt to access the contract _before_ we create the next block and so don't "zeroBaseFee" doesn't get a chance to set the baseFee to zero. So in reality, I should get the same error as for "zeroBaseFee: false".

So I will remove this whole block of code in TestContextBuilder.

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
siladu: I'm not sure I follow exactly, but refactoring how we load the genesis is out of the scope of this PR I reckon.

siladu: I think it would be more straight forward to load a full genesis file rather than just the genesis block....but yes we'd end up with lots of genesis files.

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
siladu: removed

consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/support/TestContextBuilder.java
jframe: I think prefer less genesis files to deal with so I think this approach is better after all.

jframe (approved):
LGTM

PR #5289, opened 2023-03-31
Bootstrap Linea configuration + transaction max calldata size validation

fab-10:
## PR description

With this PR we introduce Linea network, based on Paris, and create the plumbing for its configuration options, using transaction max calldata size validation as first simple new feature.


fab-10:
Wrong place

Issue #4471, opened 2022-10-03
Provide way to set extraData in execution blocks post-merge


mikeygrier:
### Description
As a Validator, I want to be able to specify the value of extraData in the execution blocks that I propose

### Acceptance Criteria
* Update existing pre-merge command line options and configuration to work in the post-merge world OR
* Provide new post-merge options that allow this value to be set e.g. `--validator-extra-data`

### Steps to Reproduce (Bug)
1. Configure a validator
2. Use existing `--miner-extra-data` or `miner-extra-data` config to set extraData in blocks built by Besu
3. Propose a block
4. See block proposed with null extraData field

**Expected behavior:** The value specified in the miner-extra-data configuration should make its way into the block's extraData field.

**Actual behavior:** The extraData field is null in the block.

**Frequency:** 100%

### Versions (Add all that apply)
* Software version: besu/v22.7.4/linux-x86_64/oracle-java-18
* Java version: Java(TM) SE Runtime Environment (build 18.0.2+9-61)
* OS Name & Version: Arch Linux
* Kernel Version: 5.19.12-arch1-1

### Additional Information (Add any of the following or anything else that may be relevant)
* System info - 16GB Memory, Quad core i5-3570

MysticRyuujin:
Bump

PR #5291, opened 2023-04-01
For PoS, use extraData that was configured in the miner-extra-data cli option

siladu:
This was hardcoded to "0x" as part of The Merge implementation. 

Fixes https://github.com/hyperledger/besu/issues/4471

Tested locally with https://github.com/siladu/beku-timestamp


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -77,8 +77,7 @@ public class MergeCoordinator implements MergeMiningCoordinator, BadChainListene
   /** The Merge block creator factory. */
   protected final MergeBlockCreatorFactory mergeBlockCreatorFactory;
   /** The Extra data. */
-  protected final AtomicReference<Bytes> extraData =
```
siladu, line 80: Can't think of a good reason why this needs to be AtomicReference

fab-10: since there is the `setExtraData` method that could change this field after the object has been created, it should be an `AtomicReference` or `volatile`

siladu: That's fair, but I don't think the MiningCoordinators are actually used by multiple threads in reality? Looks to me like the only actual usage of setExtraData is in JsonBlockImporter which is just a convenience. Might be better to just make this immutable.

jframe: Think it would be better to leave this as it is using AtomicReference. Whilst the current usage is restricted to JsonBlockImporter it's a public method we can guarantee no one else will call it on a different thread.

As for JsonBlockImporter it's probably not used much as there isn't a JsonBlockExporter but would prefer to not break it. It would be a useful tool to keep.

fab-10: Agree with Jason, until there is a public setter, it should be thread safe.

siladu: yep, I'm not about to refacotr JsonBlockImporter in this PR, so will use AtomicRef 👍

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -77,8 +77,7 @@ public class MergeCoordinator implements MergeMiningCoordinator, BadChainListene
...
-      new AtomicReference<>(Bytes.fromHexString("0x"));
+  protected AtomicReference<Bytes> extraData = new AtomicReference<>(Bytes.fromHexString("0x"));
```
garyschulte, line 80: AtomicReference can/should be final

siladu: oops, good spot

garyschulte (approved):
🚢

PR #5288, opened 2023-03-31
Log an error when RPC responds with INTERNAL_ERROR

siladu:
I think this would impact all RPCs, except maybe engine API. I am interesting in eth_getBlockByNumber for debugging purposes.

Error looks like this:
```
Error processing method: eth_getBlockByNumber ["latest",false]
java.lang.NullPointerException: Cannot invoke \"org.hyperledger.besu.ethereum.core.BlockHeader.getHash()\" because \"headHeader\" is null
at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.EthGetBlockByNumber.latestResult(EthGetBlockByNumber.java:86)
at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.AbstractBlockParameterMethod.findResultByParamType(AbstractBlockParameterMethod.java:90)
at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.AbstractBlockParameterMethod.response(AbstractBlockParameterMethod.java:102)
at org.hyperledger.besu.ethereum.api.jsonrpc.execution.BaseJsonRpcProcessor.process(BaseJsonRpcProcessor.java:44)
at org.hyperledger.besu.ethereum.api.jsonrpc.execution.TracedJsonRpcProcessor.process(TracedJsonRpcProcessor.java:41)
at org.hyperledger.besu.ethereum.api.jsonrpc.execution.TimedJsonRpcProcessor.process(TimedJsonRpcProcessor.java:45)
at org.hyperledger.besu.ethereum.api.jsonrpc.execution.AuthenticatedJsonRpcProcessor.process(AuthenticatedJsonRpcProcessor.java:51)
at org.hyperledger.besu.ethereum.api.jsonrpc.execution.JsonRpcExecutor.execute(JsonRpcExecutor.java:92)
at org.hyperledger.besu.ethereum.api.handlers.JsonRpcExecutorHandler.executeRequest(JsonRpcExecutorHandler.java:118)
at org.hyperledger.besu.ethereum.api.handlers.JsonRpcExecutorHandler.executeJsonObjectRequest(JsonRpcExecutorHandler.java:131)
at org.hyperledger.besu.ethereum.api.handlers.JsonRpcExecutorHandler.lambda$handler$0(JsonRpcExecutorHandler.java:78)
at io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)
[...]

Issue #5281, opened 2023-03-29
Nimbus Besu connection disruption


yorickdowne:
### Description

Companion issue to https://github.com/status-im/nimbus-eth2/issues/4773, as it's unclear to me whether it's a Nimbus issue, a Besu issue, or both

Nimbus 23.3.2 with Besu 23.1.2, the Engine API connection gets disrupted. Reported on Discord; I think I can reproduce this. Fresh Nimbus/Besu sync on Goerli, with Nimbus checkpoint synced and Besu snap synced.

### Additional Information (Add any of the following or anything else that may be relevant)

Debug logs are too large to attach, placed them on dropbox: https://www.dropbox.com/s/69sejeefa9i85kt/nimbus-besu-connection-woes.debug.log.tar.gz?dl=0

Log examples from Nimbus
```
consensus_1             | WRN 2023-03-29 20:25:07.252+00:00 Connection to EL node degraded             topics="elmon" url=http://execution:8551 failedRequest=getBlockByNumber statusCode=0 err="{\"code\":-32603,\"message\":\"Internal error\"}"
consensus_1             | INF 2023-03-29 20:25:09.810+00:00 Connection to EL node restored             topics="elmon" url=http://execution:8551
```

and from Besu
```
execution_1             | java.net.SocketException: Connection reset
execution_1             |     at java.base/sun.nio.ch.SocketChannelImpl.throwConnectionReset(SocketChannelImpl.java:394)
execution_1             |     at java.base/sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:426)
execution_1             |     at io.netty.buffer.PooledByteBuf.setBytes(PooledByteBuf.java:258)
execution_1             |     at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1132)
execution_1             |     at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:357)
execution_1             |     at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:151)
execution_1             |     at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:722)
execution_1             |     at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:658)
execution_1             |     at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:584)
[...]

PR #5292, opened 2023-04-02
Use cached header for latest eth_getBlockByNumber

siladu:
While syncing, it is possible that we access the cached header's number before the block itself has had chance to be imported. In other words there is a mismatch between cached blockNumber and non-cached blockHeader that is being retrieved for the RPC.

Discovered after adding debug in https://github.com/hyperledger/besu/pull/5288

## Fixed Issue(s)
https://github.com/hyperledger/besu/issues/5281

## Testing

Tested X_SNAP with each CL on sepolia as well as nimbus on goerli.
Without the fix: prysm, lighthouse and nimbus received the error (see https://github.com/hyperledger/besu/issues/5281#issuecomment-1491753557).
With the fix: no errors


macfarla (approved):
LGTM

Issue #5004, opened 2023-01-26
In-Protocol Deposits Invesigation 


non-fungible-nelson:
### Description
As a user of Besu, I would like access to in-protocol deposits, to allow for more thorough pruning of chain-data and potential improvements to checkpoint sync. With in-protocol access to deposits, my Besu node can start a sync after the deposit contract, and can prune older chain data and still get what I need to stake.

[Context](https://hackmd.io/@n0ble/deposits-breakout)

### Acceptance Criteria
* Spec changes in Besu
* Prototype in Besu
* Acceptance of [EIP-6110](https://eips.ethereum.org/EIPS/eip-6110)

PR #5295, opened 2023-04-03
 EIP-6110: Add deposit validation and apis

naviechan:
## PR description
This PR is the continuation of #5055. It is the final PR for [EIP-6110](https://eips.ethereum.org/EIPS/eip-6110). The following are the main features introduced:

- Decode and extract deposit from transaction receipt log
- Introduce `depositContractAddress` field in genesis file
- Validate deposits in a block against the transaction receipt logs
- Update `engine_newPayloadV2` and `engine_getPayloadV2` according to the [spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/experimental/eip6110.md). Some of the non-functional components were partially implemented in the previous PR

## Fixed Issue(s)

#5004 

## Testing
- [x] Local block production pre-V6110
- [x] Public testnet sync
- [x] Public mainnet sync


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
```diff
@@ -39,7 +46,13 @@ class ProhibitedDeposits implements DepositsValidator {
     private static final Logger LOG = LoggerFactory.getLogger(ProhibitedDeposits.class);
 
     @Override
-    public boolean validateDeposits(final Optional<List<Deposit>> deposits) {
+    public boolean validateDepositParameters(final Optional<List<Deposit>> deposits) {
```
siladu, line 49: nit: wondering if this should just be `validateDepositParameter`

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidatorTest.java
```diff
@@ -59,7 +122,34 @@ public void invalidateProhibitedDepositsRoot() {
...
+
+    final TransactionReceipt receipt =
+        new TransactionReceipt(null, 0L, List.of(LOG_1, LOG_2), Optional.empty());
+
+    assertThat(new DepositsValidator.AllowedDeposits().validateDeposits(block, List.of(receipt)))
+        .isTrue();
+  }
+
+  @Test
+  public void validateAllowedDeposits2() {
```
siladu, line 138: Please can you use a more descriptive name to distinguish what each test is doing

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadV2Test.java
```diff
@@ -67,6 +67,27 @@ public void shouldReturnBlockForKnownPayloadId() {
                   .isEqualTo(mockHeader.getPrevRandao().map(Bytes32::toString).orElse(""));
             });
     verify(engineCallListener, times(1)).executionEngineCalled();
+
+    // should return deposits for a post-V6110 block
```
siladu, line 71: Think this could be in a separate test to the post-shanghai one

siladu (commented):
Did a first pass looks pretty straight forward to me. The deposit parsing and validation is simple and easy to understand. Some initial comments, mostly naming.

Next pass I will review against the EIP and execution-api spec in more detail.

acceptance-tests/tests/src/test/resources/jsonrpc/engine/test-cases/18_v6110_prepare_payload.json
```diff
@@ -0,0 +1,46 @@
+{
+  "request": {
+    "jsonrpc": "2.0",
+    "method": "engine_forkchoiceUpdatedV2",
```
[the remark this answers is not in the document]
naviechan: > Why isn't this engine_forkchoiceUpdatedV6110?

I don't think we need to have a new method for forkchoiceUpdated as the current V2 is suffice. Also it's worth pointing out there was a recent modification to the `engine-api` which removes all the V6110 methods but instead modifying `engine_newPayloadV2` and `engine_getPayloadV2`

datatypes/src/main/java/org/hyperledger/besu/datatypes/Address.java
```diff
@@ -74,6 +74,9 @@ public class Address extends DelegatingBytes {
   public static final Address KZG_POINT_EVAL = Address.precompiled(0x14);
   /** The constant ZERO. */
   public static final Address ZERO = Address.fromHexString("0x0");
+  /** The deposit address. */
+  public static final Address DEPOSIT_ADDRESS =
+      Address.fromHexString("0x00000000219ab540356cbb839cbe05303d7705fa");
```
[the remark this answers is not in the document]
naviechan: Good catch, it is not. I will make it configurable

ahamlat:
To accurately gauge the impact of this PR on Besu's performance, particularly regarding block processing, the most effective approach is to deploy it on a dedicated environment with real data. It's preferable to use an environment with many deposits, and then profile the node to observe any impact on performance. In the meantime, I've identified two pertinent methods to assess performance by conducting microbenchmarking with JMH

- DepositDecoder.decodeFromLog
- BodyValidation.depositRoots

During block processing, both methods are invoked, with DepositDecoder.decodeFromLog being called iteratively within a loop that depends on the quantity of receipts and logs per receipt present in the block

**Block processing** 

```
AbstractEngineNewPayload.syncResponse
        BodyValidation.depositsRoot
	MergeCoodinator.rememberBlock
		MergeCoodinator.validateBlock
			MainnetBlockValidator.validateAndProcessBlock
				MainnetBlockValidator.validateAndProcessBlock
					MainnetBlockBodyValidator.validateBodyLight
						AllowedDeposits.validateDeposits
							DepositDecoder.decodeFromLog (called several times)
						AllowedDeposits.validateDepositsRoot
							BodyValidation.depositsRoot	
```

**Block creation**

```
AbstractBlockCreator.createBlock
	BodyValidation.depositsRoot
	DepositDecoder.decodeFromLog (called several times)
```
The microbenchmarking of both methods (Keccak256 was added to enable a comparison point)

```
Benchmark                                     Mode  Cnt       Score      Error  Units
DepositDecoder.decodeFromLogBenchmark         avgt  100   12261.105 ±   80.786  ns/op
DepositDecoder.keccakk256Benchmark            avgt  100     306.937 ±    0.377  ns/op
BodyValidation.depositsRootBenchmark            avgt  100  238915.938 ± 4451.589  ns/op
```
If I'm not wrong, BodyValidation.depositRoots is called twice during block processing, and it takes 1/4 ms to execute, I think we can save one execution but even with both executions, the overhead is acceptable.

DepositDecoder.decodeFromLog overhead will depends on the number of receipts per block and the number of logs per receipt. If there're for example 1000 logs to process, this represents 13 ms overhead for block processing and block creation.

I started also a node on mainnet with PR to test against main on current fork (previous commit of this PR 795b7c4d), I don't see any performance regression on block processing or FCU 
image


From what I gathered, I don't see a blocker to merge in termes of performance

siladu:
I've started a regression, fresh X_CHECKPOINT sync of 3 BONSAI nodes: sepolia, goerli, mainnet.

siladu:
> I've started a regression test, fresh X_CHECKPOINT sync of 3 BONSAI nodes: sepolia, goerli, mainnet.

This has passed regression: all nodes got in sync and have stayed in sync. No errors in the logs.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -195,8 +197,19 @@ protected BlockCreationResult createBlock(
...
-          Optional.empty(); // TODO 6110: Extract deposits from transaction receipts
+      final DepositsValidator depositsValidator = newProtocolSpec.getDepositsValidator();
+      Optional<List<Deposit>> maybeDeposits = Optional.empty();
+
+      if (depositsValidator instanceof DepositsValidator.AllowedDeposits) {
+        final List<Deposit> depositsFromReceipts =
+            transactionResults.getReceipts().stream()
+                .flatMap(receipt -> receipt.getLogsList().stream())
+                .map(DepositDecoder::decodeFromLog)
+                .toList();
```
siladu, line 208: Are we missing the depositContractAddress filter here?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/DepositDecoderTest.java
```diff
@@ -47,4 +52,21 @@ void shouldDecodeDeposit() {
...
+
+    Log log = new Log(address, data, topics);
+    Deposit deposit = DepositDecoder.decodeFromLog(log);
+
+    System.out.println(deposit);
```
siladu, line 70: Should be asserting instead of printing here I think

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidatorTest.java
```diff
@@ -15,25 +15,89 @@
...
   private final BlockDataGenerator blockDataGenerator = new BlockDataGenerator();
+  private static Deposit DEPOSIT_1;
+  private static Deposit DEPOSIT_2;
+  private static Log LOG_1;
+  private static Log LOG_2;
+  private static Address DEPOSIT_CONTRACT_ADDRESS;
+
+  @BeforeAll
+  public static void setup() {
+    System.out.println("setting");
```
siladu, line 52: remove sout please

siladu (commented):
Looking very close now! 
Couple of minor things, main one is a potentially missing depositContractAddress filter.

Also can you add a changelog entry please :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/DepositContract.java
```diff
@@ -0,0 +1,63 @@
...
+      new Event(
+          "DepositEvent",
+          Arrays.asList(
+              new TypeReference<DynamicBytes>() {},
+              new TypeReference<DynamicBytes>() {},
+              new TypeReference<DynamicBytes>() {},
+              new TypeReference<DynamicBytes>() {},
+              new TypeReference<DynamicBytes>() {}));
+
+  protected DepositContract(
```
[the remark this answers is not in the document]
naviechan: Interestingly if I set it to `private`, it will complain the constructor is never used. I will leave it as `protected` unless there is a better way to make it compile

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/DepositDecoder.java
```diff
@@ -39,6 +42,22 @@ public static Deposit decode(final RLPInput rlpInput) {
...
+    byte[] rawPublicKey = (byte[]) eventValues.getNonIndexedValues().get(0).getValue();
+    byte[] rawWithdrawalCredential = (byte[]) eventValues.getNonIndexedValues().get(1).getValue();
+    byte[] rawAmount = (byte[]) eventValues.getNonIndexedValues().get(2).getValue();
+    byte[] rawSignature = (byte[]) eventValues.getNonIndexedValues().get(3).getValue();
+    byte[] rawIndex = (byte[]) eventValues.getNonIndexedValues().get(4).getValue();
+
+    return new Deposit(
+        BLSPublicKey.wrap(Bytes.wrap(rawPublicKey)),
+        Bytes32.wrap(Bytes.wrap(rawWithdrawalCredential)),
+        GWei.of(Bytes.wrap(rawAmount).reverse().toLong()),
```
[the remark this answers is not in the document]
naviechan: > For Withdrawals, the CL side converted amounts into big endian before sending data to EL via engine API so we were able to keep the EL as big endian.
> 
> Was the deposit amount stored in little endian because it originated from the CL? It's shame little endian has leaked into the EL here slightly, but I guess the data is already in the contract so can't really change it after the fact 🤔

Yea exactly. The deposit contract use little endian hence we need to reverse it: https://github.com/ethereum/consensus-specs/blob/dev/solidity_deposit_contract/deposit_contract.sol#L119

I have improved the readability of the code and added a comment

config/src/main/resources/experimental.json
```diff
@@ -35,5 +35,6 @@
       "comment": "private key and this comment are ignored.  In a real chain, the private key should NOT be stored",
       "balance": "90000000000000000000000"
     }
-  }
+  },
+  "depositContractAddress": "0x00000000219ab540356cbb839cbe05303d7705fa"
```
siladu, line 39: This should be within the `config` block at the top, where you have it here makes it part of the genesis block.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/DepositContract.java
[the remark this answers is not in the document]
siladu: package-private (i.e. removing `protected` keyword will work).
By "suppress warning" I meant the "never used" warning you're talking about. Protected is ok, package-private slightly better (in terms of minimising visibility)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
```diff
@@ -64,10 +77,41 @@ public boolean validateDepositsRoot(final Block block) {
...
-    public boolean validateDeposits(final Optional<List<Deposit>> deposits) {
-      return true; // TODO 6110: Validate deposits in the next PR
+    public boolean validateDepositParameter(final Optional<List<Deposit>> deposits) {
+      return deposits.isPresent();
+    }
+
+    @Override
+    public boolean validateDeposits(final Block block, final List<TransactionReceipt> receipts) {
+      if (block.getBody().getDeposits().isEmpty()) {
+        LOG.warn("Deposits must not be null when Deposits are activated");
```
macfarla, line 98:
```suggestion
        LOG.warn("Deposits must not be empty when Deposits are activated");
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
```diff
@@ -39,7 +46,13 @@ class ProhibitedDeposits implements DepositsValidator {
...
+    public boolean validateDepositParameter(final Optional<List<Deposit>> deposits) {
+      return deposits.isEmpty();
+    }
+
+    @Override
+    public boolean validateDeposits(final Block block, final List<TransactionReceipt> receipts) {
+      Optional<List<Deposit>> deposits = block.getBody().getDeposits();
       final boolean isValid = deposits.isEmpty();
       if (!isValid) {
         LOG.warn("Deposits must be null when Deposits are prohibited but were: {}", deposits);
```
macfarla, line 58:
```suggestion
        LOG.warn("Deposits must be empty when Deposits are prohibited but were: {}", deposits);
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
macfarla: small nitpick - null has a specific meaning in java and technically using an optional here means it is `empty` rather than `null`

acceptance-tests/tests/src/test/resources/jsonrpc/engine/genesis.json
```diff
@@ -13,6 +13,7 @@
     "berlinBlock":0,
     "londonBlock":0,
     "shanghaiTime":10,
+    "experimentaleipstime":20,
```
macfarla, line 16:
```suggestion
    "experimentalEipsTime":20,
```

macfarla: just for readability :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/DepositDecoder.java
```diff
@@ -39,6 +44,25 @@ public static Deposit decode(final RLPInput rlpInput) {
     return new Deposit(publicKey, depositWithdrawalCredential, amount, signature, index);
   }
 
+  public static Deposit decodeFromLog(final Log log) {
+    Contract.EventValuesWithLog eventValues = DepositContract.staticExtractDepositEventWithLog(log);
+    byte[] rawPublicKey = (byte[]) eventValues.getNonIndexedValues().get(0).getValue();
+    byte[] rawWithdrawalCredential = (byte[]) eventValues.getNonIndexedValues().get(1).getValue();
+    byte[] rawAmount = (byte[]) eventValues.getNonIndexedValues().get(2).getValue();
+    byte[] rawSignature = (byte[]) eventValues.getNonIndexedValues().get(3).getValue();
+    byte[] rawIndex = (byte[]) eventValues.getNonIndexedValues().get(4).getValue();
```
macfarla, line 53: these can all be final

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/DepositDecoder.java
```diff
@@ -39,6 +44,25 @@ public static Deposit decode(final RLPInput rlpInput) {
     return new Deposit(publicKey, depositWithdrawalCredential, amount, signature, index);
   }
 
+  public static Deposit decodeFromLog(final Log log) {
+    Contract.EventValuesWithLog eventValues = DepositContract.staticExtractDepositEventWithLog(log);
+    byte[] rawPublicKey = (byte[]) eventValues.getNonIndexedValues().get(0).getValue();
```
macfarla, line 49:
```suggestion
    final byte[] rawPublicKey = (byte[]) eventValues.getNonIndexedValues().get(0).getValue();
```

macfarla (approved):
looks ok to me - a few minor comments

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
siladu: For context, I think this is following what I did from Withdrawals and I specifically used "null" in order to match the spec definition:
https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-3
The spec has been rewritten since then so it's not as obvious even for withdrawals https://github.com/ethereum/execution-apis/commit/939255f467f9c5700a65426159afdfc0e0a66060

and deposits doesn't even mention "null" anymore https://github.com/ethereum/execution-apis/blob/main/src/engine/experimental/eip6110.md

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -195,8 +197,24 @@ protected BlockCreationResult createBlock(
...
+                .flatMap(receipt -> receipt.getLogsList().stream())
+                .filter(
+                    log ->
+                        ((DepositsValidator.AllowedDeposits) depositsValidator)
+                            .getDepositContractAddress()
+                            .equals(log.getLogger()))
+                .map(DepositDecoder::decodeFromLog)
+                .toList();
+        maybeDeposits = Optional.of(depositsFromReceipts);
+      }
```
siladu, line 215: is there a test that covers this logic?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/DepositsValidatorProvider.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  static DepositsValidator getDepositsValidator(
+      final ProtocolSchedule protocolSchedule,
+      final BlockHeader parentBlockHeader,
+      final long timestampForNextBlock) {
+
+    return getDepositsValidator(
+        protocolSchedule.getForNextBlockHeader(parentBlockHeader, timestampForNextBlock));
+  }
```
siladu, line 46: don't think we need this version of the method, it's not used.

For withdrawals, this version is used in AbstractEngineForkchoiceUpdated because withdrawals are sent in the payload, but that isn't the case for deposits.

siladu (commented):
Just a couple more minor things

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadResultV2.java
```diff
@@ -39,8 +41,9 @@ public EngineGetPayloadResultV2(
       final BlockHeader header,
       final List<String> transactions,
       final Optional<List<Withdrawal>> withdrawals,
+      final Optional<List<Deposit>> deposits,
```
jframe, line 44: It's not part of this PR so I couldn't add a comment directly, but it looks like this replaces the `EngineGetPayloadResultV6110` that was added in the previous PR and `EngineGetPayloadResultV6110` can now be removed.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -195,8 +197,24 @@ protected BlockCreationResult createBlock(
...
+      Optional<List<Deposit>> maybeDeposits = Optional.empty();
+
+      if (depositsValidator instanceof DepositsValidator.AllowedDeposits) {
+        final List<Deposit> depositsFromReceipts =
+            transactionResults.getReceipts().stream()
+                .flatMap(receipt -> receipt.getLogsList().stream())
+                .filter(
+                    log ->
+                        ((DepositsValidator.AllowedDeposits) depositsValidator)
+                            .getDepositContractAddress()
```
jframe, line 210: Doesn't seem quite to be exposing a getter on the DepositValidator. Would rather we pass in the depositContractAddress from the config.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
```diff
@@ -64,10 +77,41 @@ public boolean validateDepositsRoot(final Block block) {
...
+            Deposit deposit = DepositDecoder.decodeFromLog(log);
+            expectedDeposits.add(deposit);
+          }
+        }
+      }
+
+      return actualDeposits.equals(expectedDeposits);
     }
 
     @Override
```
jframe, line 117: Not part of this PR but part of the DepositsValidator code. In validateDepositsRoot the error message when the deposits root doesn't match has "Invalid block: transaction root mismatch" instead of "deposit root mismatch"

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
```diff
@@ -64,10 +77,41 @@ public boolean validateDepositsRoot(final Block block) {
...
+      for (TransactionReceipt receipt : receipts) {
+        for (Log log : receipt.getLogsList()) {
+          if (depositContractAddress.equals(log.getLogger())) {
+            Deposit deposit = DepositDecoder.decodeFromLog(log);
+            expectedDeposits.add(deposit);
+          }
+        }
+      }
+
+      return actualDeposits.equals(expectedDeposits);
```
jframe, line 114: Adding a warn log when the deposits don't match would be nice

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/EnginePayloadParameter.java
```diff
@@ -64,7 +65,8 @@ public EnginePayloadParameter(
       @JsonProperty("logsBloom") final LogsBloomFilter logsBloom,
       @JsonProperty("prevRandao") final String prevRandao,
       @JsonProperty("transactions") final List<String> transactions,
-      @JsonProperty("withdrawals") final List<WithdrawalParameter> withdrawals) {
+      @JsonProperty("withdrawals") final List<WithdrawalParameter> withdrawals,
+      @JsonProperty("deposits") final List<DepositParameter> deposits) {
```
jframe, line 69: The `DepositParameter` was created in the previous PR comparing the json structure against the Engine API spec for 6110 and but I think the field names are different.

It has `pubKey` instead of `pubkey` from spec and `withdrawal_credentials` instead of `withdrawalCredentials` from spec.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
macfarla: yeah would be good to have it consistently make sense across the board - I did notice it's null in the json test files, but it didn't make sense just looking at the java code

macfarla:
the PKI unit test failures will be fixed by a merge from main

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
naviechan: Cool, I will modify `DepositsValidator` and leave `WithdrawalsValidator` untouched

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/EngineGetPayloadResultV2.java
naviechan: Yea it fell off the radar

acceptance-tests/tests/src/test/resources/jsonrpc/engine/test-cases/13_shanghai_get_payload.json
```diff
@@ -37,6 +37,7 @@
             "amount": "0x4"
           }
         ],
+        "deposits": null,
```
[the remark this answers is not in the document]
naviechan: > This block is pre-6110 so be shouldn't be returning the deposit field

I think same with it retuning null `withdrawals` pre-shanghai, it should return null `deposits` for pre-6110. Can you please confirm @siladu ?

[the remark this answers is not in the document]
siladu: Discussed this with @jframe, haven't tested it, but I suspect withdrawals with do the same thing. If so, then it's probably a bug in the existing code so IMO can be fixed outside this PR.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DepositsValidator.java
```diff
@@ -64,10 +77,47 @@ public boolean validateDepositsRoot(final Block block) {
...
+            Deposit deposit = DepositDecoder.decodeFromLog(log);
+            expectedDeposits.add(deposit);
+          }
+        }
+      }
+
+      boolean isValid = actualDeposits.equals(expectedDeposits);
+
+      if (!isValid) {
+        LOG.warn("Deposits validation failed. Deposits from block body do not match deposits from logs");
```
siladu, line 117: Would be great to add at least the block hash here so we can tell where the problem was, unless that would already happen in the level above?

You could also add the actual vs expected deposits to make debugging easier, but probably would need to go at DEBUG or TRACE level in a second log message, it would be too noisy for WARN.

acceptance-tests/tests/src/test/resources/jsonrpc/engine/test-cases/20_v6110_get_payload.json
```diff
@@ -0,0 +1,49 @@
...
+            "amount": "0x1"
+          },
+          {
+            "index": "0x1",
+            "validatorIndex": "0x0",
+            "address": "0xfe3b557e8fb62b89f4916b721be55ceb828dbd73",
+            "amount": "0x2"
+          }
+        ],
+        "deposits" : [],
```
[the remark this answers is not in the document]
jframe: Alternatively this could be added later once we a testnet/devnet to get some test data from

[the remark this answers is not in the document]
naviechan: > Alternatively this could be added later once we a testnet/devnet to get some test data from

Yes I think this is better as we are still setting up the testnet

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
naviechan: > Doesn't seem quite right to be exposing a getter on the DepositValidator. Would rather we pass in the depositContractAddress from the config.

Yea I thought the same. However I am not sure how to do it without making a big change. Should I add the `depositContractAddress` as a param of `createBlock()`? If so, seems like I would need to update all the callers of `createBlock()`. Would be great if you could give me a pointer.

jframe: Can't think of a way that doesn't end up changing at 10 or so files. 

I was thinking of passing it into the constructor of the `AbstractBlockCreator`, which would also mean passing it into the `MergeBlockCreator` and `MergeCoordinator` and constructing the `MergeCoordinator` using the value of the deposit contract from the genesis in the `MergeBesuControllerBuilder`. Still quite a few changes. There are other BlockCreator instances the Pow, Bft but they could be instantiated with an Optional.empty for the deposit contract.

naviechan:
@jframe @siladu I have updated the way the `depositContractAddress` get passed into the `AbstractBlockCreator`, not sure if I have changed it correctly so please take a look. Particularly the logic of setting a default address `DEFAULT_DEPOSIT_CONTRACT_ADDRESS` in the case it is absent while creating block.

Since `DepositsValidator.AllowedDeposits.validateDeposits()` still needs a `depositContractAddress` to filter the logs, it still relies on `MaineProtocolSpecs` to pass in the address when initializing. 

To summarize, we need `depositContractAddress` for two things, one for filtering logs to get deposits to include in block body in `createBlock()`, and one for filtering logs to validate deposits.
The former has `depositContractAddress` passed around like this: `BesuControllerBuilder` -> `MergeCoordinator` -> `MergeBlockCreator` with `MergeBlockCreator` setting the default address if absent
The latter is like this: `MainnetProtocolSpecs` -> `DepositsValidator.AllowedDeposits()` -> `DepositsValidator.AllowedDeposits.validateDeposits()`

naviechan:
Looks like there couple tests failing because `MergeBesuControllerBuilder.createTransitionMiningCoordinator()` now reads genesis config. I will fix them when you guys are okay with my proposed change change on `createBlock()`.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -195,8 +203,24 @@ protected BlockCreationResult createBlock(
...
+      Optional<List<Deposit>> maybeDeposits = Optional.empty();
+
+      if (depositsValidator instanceof DepositsValidator.AllowedDeposits) {
+        final List<Deposit> depositsFromReceipts =
+            transactionResults.getReceipts().stream()
+                .flatMap(receipt -> receipt.getLogsList().stream())
+                .filter(
+                    log ->
+                        depositContractAddress
+                            .orElse(DEFAULT_DEPOSIT_CONTRACT_ADDRESS)
```
siladu, line 216: I don't think we should handle the default value in this class.

I think in this case it's fine to use orElseThrow since we shouldn't get into this case for non-mainnet usecases.
We'll probably need to retest with non-mainnet networks to make sure, I'm happy to do this testing.

siladu (commented):
Approach to depositContractAddress looks fine, just the small thing about the default value.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
siladu: ...in order for this error to be thrown I think the genesis file would need to have `experimentalEipsTime` configured and be missing some `depositContractAddress` configuration. I think this is acceptable for this PR.

Me and @jframe discussed a refactoring of this class to move the withdrawals and deposit code into MergeBlockCreator (which should maybe now be called 'ProofOfStakeBlockCreator') since withdrawals and deposits are specific to mainnet/PoS.
I think that refactoring would reduce the risk of misconfiguration further. 

Maybe a default is still the way to go but it should be as close to when the config is read-in rather than when it is used...maybe @jframe has an opinion on this too.

jframe: Agree that @siladu that we shouldn't have a default defined. We can assume that if this EIP is enabled then we will have a deposit contract defined in the relevant network genesis file.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
jframe: We definitely should have a unit test for this logic

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
jframe: That refactoring to move the mainnet specific features out of the AbstractBlockCreator would be nice to have but I think that is something for the future that could be done in another PR.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -195,8 +203,24 @@ protected BlockCreationResult createBlock(
 
       throwIfStopped();
 
-      final Optional<List<Deposit>> maybeDeposits =
-          Optional.empty(); // TODO 6110: Extract deposits from transaction receipts
+      final DepositsValidator depositsValidator = newProtocolSpec.getDepositsValidator();
+      Optional<List<Deposit>> maybeDeposits = Optional.empty();
+
+      if (depositsValidator instanceof DepositsValidator.AllowedDeposits) {
```
jframe, line 209: Rather than checking the type of the validator, we could enable this based on whether the `depositContractAddress` has a value

jframe: Sorry @naviechan I gave you a bad suggestion here, this shouldn't be changed. This can't be changed to enable this based on whether `depositContractAddress` as this value will be defined and added to all the network genesis files.

naviechan: No worries, I think its better for the condition to be `depositsValidator instanceof DepositsValidator.AllowedDeposits && depositContractAddress.isPresent()`?

jframe: Yes that's makes sense, good idea

siladu:
Reran some regression tests (deposits disabled) on latest changes with a local interop and sepolia sync ✅ 
Think this is ready to merge if @jframe is happy with latest changes.

jframe:
> Reran some regression tests (deposits disabled) on latest changes with a local interop and sepolia sync ✅ Think this is ready to merge if @jframe is happy with latest changes.

Looks fine to me

jframe (approved):

siladu:
Thanks @jframe. I will hold off merging until we are sure main branch is clear: there are some issues on main found during the 23.4.2 soak.

siladu (approved):

Issue #4696, opened 2022-11-17
Ethstats connection won't be retried properly if the ethstats server runs with TLS/SSL


skylenet:
### Description
When forwarding metrics via the `--ethstats` flag to an ethstats server that serves under TLS/SSL, besu falls back to plain HTTP if the TLS connection fails and will never try again. The only way to establish a connection again is by restarting besu and hoping that the first connectivity attempt works. 

### Acceptance Criteria
Besu should be able to try to reconnect to an ethstats server that has TLS/SSL configured. 

### Steps to Reproduce (Bug)
1 Have a ethstats server (I'll name it ethstats.local:443 on the following steps) running with TLS/SSL 
2. Stop the ethstats server to simulate that it's not available.
3. Run besu with `--ethstats=mynode:mysecret@ethstats.local:443`
4. See that it will log something like `2022-11-17 12:46:54.910+00:00 | vert.x-eventloop-thread-4 | ERROR | EthStatsService | Failed to reach the ethstats server Failed to create SSL connection (trying without ssl)`
5. Start the ethstats server with TLS/SSL
6. Notice that besu won't be able to connect to it because it won't retry with TLS/SSL again. It will stick to plain HTTP.

### Versions (Add all that apply)
* Software version: [`besu --version`]: hyperledger/besu:22.10.0

non-fungible-nelson:
reprioritizing - Pari from EF devops has asked for some of this. alongside #3792

Issue #3792, opened 2022-05-05
Improve DNS use from Ethstats


jflo:
Besu needs DNS/IP:Port whereas others allow for just DNS and let the reverse proxy route them to the correct port. My reverse proxy forwards the traffic to the new ethstats server, but Besu doesn't like that since it tries to send traffic to the port 3000 of the specified DNS server (doesn't touch the reverse proxy then). 

So I can switch to IP:Port of the new server, but I'd say this is a QoL thing that should be fixed. 

### Acceptance Criteria
* Besu continues reporting to an Ethstats server that has changed it's ip/port under it's DNS entry.

### Steps to Reproduce (Bug)
1. Change the IP of a reverse proxy between Besu and Ethstats

**Expected behavior:** Reporting continues

**Actual behavior:** Reporting stops, error message in logs saying:

` Failed to reach the ethstats server finishConnect(..) failed: Connection refused: ethstats.mainnetshadowfork3.ethdevops.io/164.92.169.47:3000`

usmansaleem:
networkaddress.cache.ttl

usmansaleem:
@jflo hey Justin, can you ping me either on Discord or slack, I would like to chat about this ticket. One possible solution is to add system property networkaddress.cache.ttl=60 in Besu startup script. The other is to re-verify if the IP address of reverse proxy host is changing more often or the IP address of ethstats is changing more often?

usmansaleem:
The other option is to modify the ethstats connecting code and set https://vertx.io/docs/apidocs/io/vertx/core/dns/AddressResolverOptions.html#setCacheMaxTimeToLive-int-.

usmansaleem:
@non-fungible-nelson hey Matt, can you ping me regarding above issue, I want to get some further clarification from the user regarding the above issue and whether it is already sorted out or not. Cheers, Uzi.

non-fungible-nelson:
@jflo might know this - based on the mainnet shadow fork logs, this is likely also with the EF. @usmansaleem did you connect with @parithosh on the improvements?

parithosh:
@siladu gave me an image to test out and that showed no ethstats issues, I'm honestly not sure which branch or fix he gave me to try, but perhaps he can chime in

usmansaleem:
@parithosh That was probably cut againt the main branch that contained the ethstats related enhancements. Next Besu release should also contain these fixes. I am going to close this ticket, in case you face any issues, feel free to re-open this issue or create a new one. Thanks, /uzi

PR #5301, opened 2023-04-04
Ethstat retry fix

usmansaleem:
## PR description
Fix ethstats retry logic. Attempt to connect with and without ssl. If `--ethstats` is specified without a port, use port 443 (for wss://) and port 80 (for ws://) as defaults instead of 3000. This will help in k8 or similar environment where ethstats is running behind a reverse proxy.

Introduced optional `--ethstats-cacert-file`. This option specifies the path to the root CA (Certificate Authority) certificate file that has signed ethstats server certificate (or reverse proxy certificate). This option is optional. Useful in non-production setup.

## Fixed Issue(s)

#4696 
Potentially #3792 

## Local Testing
https://github.com/usmansaleem/ethstats-proxy


besu/src/main/java/org/hyperledger/besu/cli/options/stable/EthstatsOptions.java
```diff
@@ -27,6 +28,7 @@ public class EthstatsOptions implements CLIOptions<NetstatsUrl> {
 
   private static final String ETHSTATS = "--ethstats";
   private static final String ETHSTATS_CONTACT = "--ethstats-contact";
+  private static final String ETHSTATS_PEM_TRUST = "--ethstats-pem-trust";
```
siladu, line 31: Not very clear this is a filepath

How about `ethstats-pem-file` or `ethstats-trust-pem-file`?

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/NetstatsUrl.java
```diff
@@ -48,8 +53,9 @@ static NetstatsUrl fromParams(final String url, final String contact) {
             .nodeName(netStatsUrl.group(1))
             .secret(netStatsUrl.group(2))
             .host(netStatsUrl.group(3))
-            .port(Integer.parseInt(Optional.ofNullable(netStatsUrl.group(5)).orElse("3000")))
+            .port(Integer.parseInt(Optional.ofNullable(netStatsUrl.group(5)).orElse("-1")))
```
siladu, line 56: What's the impact of switching from 3000 to -1 here?

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
```diff
@@ -213,6 +229,20 @@ public void start() {
...
 
+  private String getEthStatsHost() {
+    return String.format(
+        "%s://%s:%s",
+        webSocketConnectOptions.isSsl() ? "wss" : "ws",
+        netstatsUrl.getHost(),
+        getWsPort(netstatsUrl, webSocketConnectOptions.isSsl()));
+  }
+
+  private void flipSSL() {
```
siladu, line 240: nit:
```suggestion
  private void toggleSSL() {
```

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
```diff
@@ -147,22 +147,43 @@ public EthStatsService(
...
+    if (netstatsUrl.getPort() >= 0) {
+      return netstatsUrl.getPort();
+    }
+    return isSSL ? 443 : 80;
   }
 
   /** Start. */
   public void start() {
-
+    flipSSL();
```
siladu, line 183: Why are we flipping it?

besu/src/main/java/org/hyperledger/besu/cli/options/stable/EthstatsOptions.java
```diff
@@ -77,8 +86,23 @@ public String getEthstatsContact() {
...
+   * @return Path to CA file. null if no CA file to set.
+   */
+  public Path getEthstatsPemTrust() {
+    return ethstatsPemTrust;
+  }
+
   @Override
   public List<String> getCLIOptions() {
-    return Arrays.asList(ETHSTATS + "=" + ethstatsUrl, ETHSTATS_CONTACT + "=" + ethstatsContact);
+    final ArrayList<String> options = new ArrayList<>();
```
jframe, line 100: nit: type on the left could be just List<String>

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/NetstatsUrl.java
```diff
@@ -38,7 +40,10 @@ public interface NetstatsUrl {
 
   String getContact();
 
-  static NetstatsUrl fromParams(final String url, final String contact) {
+  @Nullable
+  Path getPemTrust();
```
jframe, line 44: Perhaps this interface is misnamed but including the pemTrust as part of this type representing an URL doesn't make sense to me.

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
jframe: This doesn't make sense to me either. I thought we would flip after a connection failure. If I understand this correctly if we have SSL specified we will try without SSL, possibly fail then do the connect again flipping to enable SSL again?

usmansaleem: The naming of the method can be changed. The `reconnect` logic calls `start` method again after n seconds. We want to switch from ssl to non-ssl, and then to ssl on each reconnect attempt.

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/NetstatsUrl.java
usmansaleem: The usage of getPort is in EthStatsService class. There we can derive the 443/80 ports if this port is set to -1. The other option might be to use Option<Integer> to represent the port instead of -1.

besu/src/main/java/org/hyperledger/besu/cli/options/stable/EthstatsOptions.java
usmansaleem: Renamed to `--ethstats-cacert`, which uses commonly used pattern `--cacert` in *nix cli tools to specify the same.

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/NetstatsUrl.java
usmansaleem: I've renamed this interface to `EthStatsConnectOptions` to better represent its usage and intent.

besu/src/main/java/org/hyperledger/besu/cli/options/stable/EthstatsOptions.java
siladu: -file prefix seemed to be the most popular form used in besu.

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
siladu: I assumed start is called on startup as well in which case the toggle doesn't make sense to me. Are you saying start is only called first time we reconnect? What if we reconnect more than once?

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/NetstatsUrl.java
siladu: I see the breaking change mentioned in the changelog now 👍

besu/src/main/java/org/hyperledger/besu/cli/options/stable/EthstatsOptions.java
jframe: Agree, we are using the -file suffix elsewhere

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
jframe: The start is called by the Runner.startExternalServices. So initially we will toggle the SSL. There is also a 10-second delay between each retry. So if we expect to use SSL we try without SSL fail wait for 10s and then connect with SSL. This doesn't seem right to me.

Why do we need to flip-flop between SSL and non-SSL?

usmansaleem: Thats the behavior of geth. They try to connect to ethstats first with ssl, then with non-ssl. We are replicating the same behavior i.e. first attempt with ssl, next attempt with non-ssl so-on so forth.

jframe (approved):

siladu (approved):
LGTM

MysticRyuujin:
Can I ask how one is suppose to use the URL now? This broke my deployment...

I used to have this:
```
--ethstats=MysticRyuujin-Besu-Teku:<redacred>@ethstats.dev:443 \
```

That no longer works, I get:
```
The plain HTTP request was sent to HTTPS port
```

If I remove `:443` from the end I get this:
```
301 Moved Permanently
```

If I prefix `wss://` to the front, I get this:
```
Invalid netstats URL syntax. Netstats URL should have the following format 'nodename:secret@host:port' or 'nodename:secret@host'
```

So...what's the appropriate way to set this?

non-fungible-nelson:
@siladu any insight with Uzi out of office?

siladu:
@MysticRyuujin does `--ethstats=MysticRyuujin-Besu-Teku:<redacted>@ethstats.dev:3000` work?

MysticRyuujin:
Hello @siladu - the EF DevOps team had an issue with their ethstats server so I just now got around to testing, that doesn't seem to work either, it's not clear to me why you're recommending port 3000 here but, no that's not working.

MysticRyuujin:
Ok, removing `443` from the end works now that the ethstats.dev server is actually working :laughing:

siladu:
> it's not clear to me why you're recommending port 3000 here

@MysticRyuujin As this PR description notes: 
> use port 443 (for wss://) and port 80 (for ws://) as defaults instead of 3000. 

besu was previously defaulting to 3000 for the ethstats server. I believe ethstats server still use this port but more recently they have been but behind a reverse proxy so you need to use standard ports instead.

Glad it's working for you now :)

PR #5305, opened 2023-04-04
Env var to debug child processes.

jflo:
export BESU_DEBUG_CHILD_PROCESS_PORT=5005 and when running ProcessBesuNodeRunner (i.e. acceptance tests) the child process will suspend and wait for a debugger to attach.


acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/ProcessBesuNodeRunner.java
```diff
@@ -421,6 +421,29 @@ public void startNode(final BesuNode node) {
...
+        String prevJavaOpts = processBuilder.environment().get("JAVA_OPTS");
+        if (prevJavaOpts == null) {
+          processBuilder.environment().put("JAVA_OPTS", debugOpts);
+        } else {
+          processBuilder.environment().put("JAVA_OPTS", prevJavaOpts + debugOpts);
+        }
+
+      } else {
+        LOG.debug(
+            "Child process may be attached to by specifying -DBESU_DEBUG_CHILD_PROCESS_PORT=<port>");
```
garyschulte, line 440: -D specifies a System property, not an envar.  if you intend for it to be a system property we should use System.getProperty rather than System.getenv (above)

pinges (commented):
Are you sure that there is always only one child process? What happens if there are more than one?

jflo:
> Are you sure that there is always only one child process? What happens if there are more than one?

Definitely not. Enabling this when running the entire suite wouldn't make much sense, we are assuming the dev has at least isolated the problem to a specific test they are trying to debug, which should be running as a single process.

fab-10:
> > Are you sure that there is always only one child process? What happens if there are more than one?
> 
> Definitely not. Enabling this when running the entire suite wouldn't make much sense, we are assuming the dev has at least isolated the problem to a specific test they are trying to debug, which should be running as a single process.

An option is to just increment the port number for each new child process

garyschulte (approved):
🚢

PR #5306, opened 2023-04-04
obfuscate jwt trace logs

garyschulte:
## PR description

Obfuscate JWT token in trace logs, leaving enough to compare with what would have been sent for debugging purposes.

## Fixed Issue(s)

addresses static analysis alerts 597 and 810


garyschulte:
closing

PR #5311, opened 2023-04-04
Handle busy exception on fast/snap WS Download process

### LINKED ISSUE #5300: Handle RocksDB.Busy exception in the worldstate download pipeline
When using the OpmitisticTransactionDB RocksDB can return Status.Busy when committing changes to the database (likely caused by multiple tx trying to write to the same key).

The worldstate download pipeline currently halts when we get a busy exception.

We need to catch and handle this type of exception to prevent the sync halts due to a temporary write conflict to the db.

gfukushima:
## PR description
This PR is an attempt to stop the worldstate download pipeline from breaking when RocksDB returns a Status.Busy Exception.
We currently don't handle that scenario and forest does seem to trigger that often.
I suspect Bonsai rarely hits this scenario but it could also leverage from the exception handler.

## Fixed Issue(s)

Fixes #5300


gfukushima:
Closing PR after getting 100% of instances eventually triggering world state issues

Issue #5296, opened 2023-04-03
Flaky JsonRpcHttpServiceTest.getBlockByNumberForLatest test


siladu:
Following https://github.com/hyperledger/besu/commit/27fd97f6b578d94fb06fcdde779596dd90bd0738, `org.hyperledger.besu.ethereum.api.jsonrpc.JsonRpcHttpServiceTest.getBlockByNumberForLatest` is flaky

siladu:
Will leave open for a couple of days to make sure test is no longer flaky

PR #5313, opened 2023-04-05
Attempt to fix flaky getBlockByNumberForLatest test

siladu:
Lock down timestamp

## Fixed Issue(s)

#5296


gfukushima (approved):
nit: I think the timestamp hard set there is probably unnecessary since you're already returning the blockWithMetadata

siladu:
> nit: I think the timestamp hard set there is probably unnecessary since you're already returning the blockWithMetadata

@gfukushima You might be right, but couldn't recreate the flakiness locally so it's a conservative shot in the dark.

PR #5328, opened 2023-04-11
Rocksdb plugin to support OptimisticTransactionDb and TransactionDb

### LINKED ISSUE #5300: Handle RocksDB.Busy exception in the worldstate download pipeline

gfukushima:
## PR description
Tis PR introduces some changes to the RocksDB plugin to allow using the DB in either pessimistic or optimistic transaction mode.
Currently pessimistic is tied to the usage of Forest data format.

The refactor was done with the intent to keep bonsai as it is currently with usage of the optimistic transaction db.

This is just a initial draft not ready for review yet. 

## Fixed Issue(s)

Partially fixes #5300 

Scenarios tested:

**Sepolia**
- [x] Fast-sync Forest (uses Pessimistic mode)
- [x] Snap-sync Bonsai (uses Optimistic mode)   

**Goerli**
- [x] Fast-sync Forest (uses Pessimistic mode)
- [x] Snap-sync Bonsai (uses Optimistic mode)

**Mainnet**
- [x] Snap-sync Forest (uses Pessimistic mode) not finished yet
- [x] Snap-sync Bonsai (uses Optimistic mode)
- [x] Checkpoint-sync Forest (uses Pessimistic mode) not finished yet 
- [x] Checkpoint-sync Bonsai (uses Optimistic mode)

**Backwards compatibility**
- [x] Roll back a Forest synced node to 23.1.2 forcing it to use Optimistic 
- [ ] Update a current synced Forest node that is using Optimistic to use the new version with pessimistic to ensure this won't require resyncs. (not done yet)


non-fungible-nelson:
Can you explain the benefits of these modes/this PR? 

I think a review from @garyschulte at some point be useful as well (not urgent prior to the next release).

garyschulte:
The main benefit is that it simplifies concurrency for forest mode.  Bonsai relies on Snapshots, which require OptimisticTransactionDb.  

For some reason Optimistic causes transaction locking timeouts for forest mode.  We could spend time on forest to discover why, but if we can avoid the issue by reverting toTransactionDb instead, that is a win. 

I think there is value in having the ability to configure both modes with the rocksdb plugin, so that neither storage format forces the other mode to use a specific transaction mode.

gfukushima:
It is pretty much impossible to sync a node currently using Forest (fast/snap/checkpoint). The worldstate download pipeline gets a lot of locking timeouts like Gary said. Switching Forest back to Transaction mode puts Forest back on track to get the the initial sync working.

gfukushima:
> I think there is value in having the ability to configure both modes with the rocksdb plugin, so that neither storage format forces the other mode to use a specific transaction mode.

@garyschulte 
I actually thought about introducing a new flag at the plugin level since this is actually related to the plugin itself and not the stoage-format we're using but eventually this would allow users to make no-functional combinations such as bonsai and pessimistic (which we probably don't want to).
We could possibly tie the TxDB mode with data-storage-format at the CLI level with some flags validations.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
```diff
@@ -177,19 +181,35 @@ public KeyValueStorage create(
...
-
-            segmentedStorage =
-                new RocksDBColumnarKeyValueStorage(
-                    rocksDBConfiguration,
-                    segmentsForVersion,
-                    ignorableSegments,
-                    metricsSystem,
-                    rocksDBMetricsFactory);
+            if (isForestStorageFormat) {
+              LOG.info("FOREST mode detected, using pessimistic DB.");
```
garyschulte, line 185: I wouldn't call it pessimistic per se 😆 .  maybe locking ?

plugins/rocksdb/src/test/java/org/hyperledger/besu/plugin/services/storage/rocksdb/unsegmented/RocksDBColumnarKeyValueStorageTest.java
```diff
@@ -299,7 +300,7 @@ private SegmentedKeyValueStorage<RocksDbSegmentIdentifier> createSegmentedStore(
       final Path path,
       final List<SegmentIdentifier> segments,
       final List<SegmentIdentifier> ignorableSegments) {
-    return new RocksDBColumnarKeyValueStorage(
+    return new OptimisticRocksDBColumnarKeyValueStorage(
```
garyschulte, line 303: Could we do similar test with TransactionDb / PessimisticRocksDBColumnarKeyValueStorage.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
```diff
@@ -177,19 +181,35 @@ public KeyValueStorage create(
...
           }
           final RocksDbSegmentIdentifier rocksSegment =
               segmentedStorage.getSegmentIdentifierByName(segment);
-          return new SegmentedKeyValueStorageAdapter<>(
-              segment, segmentedStorage, () -> segmentedStorage.takeSnapshot(rocksSegment));
+
+          if (isForestStorageFormat) {
+            return new SegmentedKeyValueStorageAdapter<>(segment, segmentedStorage);
+          } else {
+            return new SegmentedKeyValueStorageAdapter<>(
```
garyschulte, line 210: I am not a fan of our adapter pattern for rocksdb, HOWEVER it would be clarifying if forest had its own adapter that did not implement SnappableKeyValueStorage.  As it is, Pessimistic db would report as being Snappable.  Forest code never takes a snapshot, so it is mostly academic, but could lead to problems in the future.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/PessimisticRocksDBColumnarKeyValueStorage.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Not supported for Pessimistic.
+   *
+   * @param segment the segment
+   * @return the RocksDb columnar key value snapshot
+   * @throws StorageException the storage exception
+   */
+  @Override
+  public RocksDBColumnarKeyValueSnapshot takeSnapshot(final RocksDbSegmentIdentifier segment) {
```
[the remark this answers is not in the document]
garyschulte: adapter comment from above should have probably gone here.

garyschulte (commented):
LGTM, minor suggestions about naming and adapter split.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/PessimisticRocksDBColumnarKeyValueStorage.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.List;
+
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksDBException;
+import org.rocksdb.TransactionDB;
+import org.rocksdb.WriteOptions;
+
+/** Pessimistic RocksDB Columnar key value storage */
+public class PessimisticRocksDBColumnarKeyValueStorage extends RocksDBColumnarKeyValueStorage {
```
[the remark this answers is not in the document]
jframe: I was using the Pessimistic to describe that mode myself. Didn't realise it was official so happy to leave it as that.

plugins/rocksdb/src/test/java/org/hyperledger/besu/plugin/services/storage/rocksdb/unsegmented/RocksDBColumnarKeyValueStorageTest.java
gfukushima: Extended the current class to create and use TransactionDB and OtimisticTransactionDB. Thanks for the suggestion.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
gfukushima: changed to TransactionDB and moved that to the debug level.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
gfukushima: Done.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/PessimisticRocksDBColumnarKeyValueStorage.java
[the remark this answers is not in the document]
gfukushima: gone

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
macfarla: pessimistic occurs elsewhere in naming (class name)

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/TransactionDBRocksDBColumnarKeyValueStorage.java
```diff
@@ -0,0 +1,91 @@
...
+import org.rocksdb.TransactionDB;
+import org.rocksdb.WriteOptions;
+
+/** Pessimistic RocksDB Columnar key value storage */
+public class TransactionDBRocksDBColumnarKeyValueStorage extends RocksDBColumnarKeyValueStorage {
+
+  private final TransactionDB db;
+
+  /**
+   * The constructor of PessimisticRocksDBColumnarKeyValueStorage
```
macfarla, line 38: pessimistic has been renamed

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/TransactionDBRocksDBColumnarKeyValueStorage.java
```diff
@@ -0,0 +1,91 @@
...
+import org.hyperledger.besu.services.kvstore.SegmentedKeyValueStorageTransactionTransitionValidatorDecorator;
+
+import java.util.List;
+
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksDBException;
+import org.rocksdb.TransactionDB;
+import org.rocksdb.WriteOptions;
+
+/** Pessimistic RocksDB Columnar key value storage */
```
macfarla, line 32: pessimistic has been renamed

macfarla (commented):
javadoc - needs a find/replace for pessimistic

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
macfarla: javadoc

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/OptimisticRocksDBColumnarKeyValueStorage.java
```diff
@@ -0,0 +1,102 @@
...
+   * @throws StorageException the storage exception
+   */
+  @Override
+  public SegmentedKeyValueStorage.Transaction<RocksDbSegmentIdentifier> startTransaction()
+      throws StorageException {
+    throwIfClosed();
+    final WriteOptions writeOptions = new WriteOptions();
+    writeOptions.setIgnoreMissingColumnFamilies(true);
+    return new SegmentedKeyValueStorageTransactionTransitionValidatorDecorator<>(
+        new RocksDBColumnarKeyValueStorage.RocksDbTransaction(
```
jframe, line 86: RocksDbTransaction could be imported like it is in the TransactionDBRocksDBColumnarKeyValueStorage so this code is little shorter

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/OptimisticRocksDBColumnarKeyValueStorage.java
```diff
@@ -0,0 +1,102 @@
...
+            db.beginTransaction(writeOptions), writeOptions));
```
jframe, line 87: Shame this is duplicated between TransactionDBRocksDBColumnarKeyValueStorage and this class. But probably small enough to not matter that much

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -68,48 +66,51 @@
...
-      final MetricsSystem metricsSystem,
-      final RocksDBMetricsFactory rocksDBMetricsFactory)
-      throws StorageException {
-    this(configuration, segments, List.of(), metricsSystem, rocksDBMetricsFactory);
-  }
+  private final MetricsSystem metricsSystem;
+  private final RocksDBMetricsFactory rocksDBMetricsFactory;
+  private final RocksDBConfiguration configuration;
+  private Map<Bytes, String> segmentsById;
+  /** RocksDB DB options */
```
jframe, line 94: nit: there are a lot more comments here now, but maybe it's just me but I don't think they add much value in this case

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/TransactionDBRocksDBColumnarKeyValueStorage.java
gfukushima: Fixed

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/OptimisticRocksDBColumnarKeyValueStorage.java
gfukushima: Done

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/OptimisticRocksDBColumnarKeyValueStorage.java
gfukushima: necessary due to the type of instance being used in the db object. beginTransaction has different implementations for OptimisticTransactionDB and TransactionDB.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
gfukushima: Added to comply with the javadoc rules

jframe (approved):

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
```diff
@@ -149,55 +153,71 @@ public KeyValueStorage create(
...
-          throw new IllegalStateException(
-              String.format(
-                  "Developer error: A supported database version (%d) was detected but there is no associated creation logic.",
-                  databaseVersion));
+
+        final RocksDbSegmentIdentifier rocksSegment =
+            segmentedStorage.getSegmentIdentifierByName(segment);
+
+        if (isForestStorageFormat) {
+          return new NonSnappableSegmentedKeyValueStorageAdapter<>(segment, segmentedStorage);
```
garyschulte, line 207: not to be too naming nitpicky, but IMO we should define these in the 'positive', like SnappableSegmentedKeyValueStorageAdapter and Segmented...Adapter

garyschulte (approved):
LGTM, just one more nitpick about class naming

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/NonSnappableSegmentedKeyValueStorageAdapter.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.stream.Stream;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+/**
+ * The type Segmented key value storage adapter.
+ *
+ * @param <S> the type parameter
+ */
+public class NonSnappableSegmentedKeyValueStorageAdapter<S> implements KeyValueStorage {
```
garyschulte, line 35: if we have SegmentedKVStorageAdapter as the base class, then the Snappable version can inherit from it and just implement the SnappableKVStorage interface.

garyschulte (approved):
🚢

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
gfukushima: done

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/NonSnappableSegmentedKeyValueStorageAdapter.java
gfukushima: done

PR #5330, opened 2023-04-11
Update get account method in the accumulator

matkt:
## PR description

This PR fixes an issue discovered in the accumulator, which could occasionally result in an invalid version of the account. Additionally, the PR addresses a side effect associated with this fix that impacts trace_call.


garyschulte (approved):
GTG

garyschulte (dismissed):
looks like tracing refactor might depend on this behavior now?

ahamlat:
In terms of performance, we've observed that running a node with this PR results in similar performance to running another node on the same hardware.
The yellow line represents this PR
image

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -89,40 +90,43 @@ public Optional<TransactionSimulatorResult> process(
...
   }
 
-  public Optional<TransactionSimulatorResult> process(
+  public <U> Optional<U> process(
       final CallParameter callParams,
       final TransactionValidationParams transactionValidationParams,
       final OperationTracer operationTracer,
+      final BiFunction<
+              MutableWorldState, Optional<TransactionSimulatorResult>, ? extends Optional<U>>
+          mapper,
```
siladu, line 129: This is a pretty gnarly function type - can this be any clearer, maybe extracting into FunctionalInterface with a good name? A comment at least? 

I currently have no idea what different classes it can return and the variable name `mapper` provides no clues.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
```diff
@@ -105,7 +105,7 @@ void cloneFromUpdater(final BonsaiWorldStateUpdateAccumulator source) {
 
   @Override
   public Account get(final Address address) {
-    return super.getAccount(address);
+    return super.get(address);
```
siladu, line 108: What is the difference between getAccount and get and why is it significant?

siladu:
> Additionally, the PR addresses a side effect associated with this fix that impacts eth_call.

What was the side-effect?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
matkt: > > Additionally, the PR addresses a side effect associated with this fix that impacts eth_call.
> 
> What was the side-effect?

it was not ethcall it was traceCall

In this part we are trying to get the account from the DB [https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyper[…]i/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java#L48) before this PR it was from the cache. The problem in this case the worldstate is already close at this moment [https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyper[…]edger/besu/ethereum/api/jsonrpc/internal/methods/TraceCall.java](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceCall.java#L105)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
matkt: The difference between these two methods arises from the fact that getAccount reads from the Bonsai accumulator cache before attempting to read from the database. This means that there can be a cached version available. On the other hand, get first checks if there is an updated or deleted account, and if not, it reads from the database. It does not read from the accumulator cache. In this particular case, it seems to be causing an issue.

To summarize, Bonsai (not Forest) introduces the concept of an accumulator. In the case of the getAccount and other methods in the Worldstate Updater, we first read from the accumulator before reading from the database. The modification of this line was made to optimize SLOAD by reading accounts from the accumulator instead of fetching them from the database. However, it appears that we did not fully understand all the cases related to this call, and we missed this particular case. Therefore, it is better to rollback the change. Additionally, with all the other optimizations implemented, this one doesn't seem to make a significant difference anymore.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
matkt: Done hope it will b clear like that

matkt:
> > Additionally, the PR addresses a side effect associated with this fix that impacts eth_call.
> 
> What was the side-effect?

Sorry , problem it's on TraceCall . See the description of the problem. In this part we are trying to get the account from the DB [https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyper[…]i/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java#L48) before this PR it was from the cache. The problem in this case the worldstate is already close at this moment [https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyper[…]edger/besu/ethereum/api/jsonrpc/internal/methods/TraceCall.java](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/TraceCall.java#L105)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
matkt: https://github.com/hyperledger/besu/pull/5330/files#diff-5bf48e64343e7cc59a7ded7bbaaa71872d71b012172d763d5ee5baa1070fa37c

non-fungible-nelson:
Hey @matkt - ready to be merged?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright ConsenSys AG.
```
siladu, line 2: Should be `Copyright contributors to Hyperledger Besu.`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Optional;
+import java.util.function.BiFunction;
+
+/**
+ * This class is passed to the transaction simulator and allows executing code before the copy of
+ * the worldstate is closed. When processing a transaction with the simulator, a copy of the
+ * worldstate is created, and once the simulation process is complete, the copy is closed. In some
+ * cases the simulator fills a Tracer that requires updates from the worldstate. If we want to
+ * generate a trace after the execution, it would fail. This interface allows us to execute our code
+ * before the worldstate is closed, ensuring that we can generate our objects without any issues.
```
siladu, line 28: Essentially a callback function that is called upon the worldstate being closed?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
```diff
@@ -0,0 +1,52 @@
...
+   * mutableWorldState, which represents the worldstate before it is closed, and
+   * transactionSimulatorResult, which holds the result of the transaction simulation. The method
+   * returns an Optional object containing the created object. The specific type of the object can
+   * vary based on the implementation and is determined by the caller's needs. The purpose of this
+   * method is to provide a hook for executing custom code or generating objects before the
+   * worldstate is closed, allowing for any necessary processing or operations to be performed.
+   *
+   * @param mutableWorldState The worldstate before it is closed.
+   * @param transactionSimulatorResult The result of the transaction simulation.
+   * @return The object that has been created.
```
siladu, line 46: Still not clear to me what the objects that are being created are can you extend a Type to make this clear, or is that not possible for this set of objects?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * This class is passed to the transaction simulator and allows executing code before the copy of
+ * the worldstate is closed. When processing a transaction with the simulator, a copy of the
+ * worldstate is created, and once the simulation process is complete, the copy is closed. In some
+ * cases the simulator fills a Tracer that requires updates from the worldstate. If we want to
+ * generate a trace after the execution, it would fail. This interface allows us to execute our code
+ * before the worldstate is closed, ensuring that we can generate our objects without any issues.
+ * Then, the simulator will properly close the copy of the worldstate.
+ */
+public interface PreWorldStateCloseGuard<U>
```
siladu, line 31: IIUC, would `onWorldStateClose` make sense?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
siladu: Or provide an example maybe?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
```diff
@@ -0,0 +1,52 @@
...
+        MutableWorldState, Optional<TransactionSimulatorResult>, Optional<? extends U>> {
+
+  /**
+   * This method executes code before the worldstate is closed. It takes two parameters:
+   * mutableWorldState, which represents the worldstate before it is closed, and
+   * transactionSimulatorResult, which holds the result of the transaction simulation. The method
+   * returns an Optional object containing the created object. The specific type of the object can
+   * vary based on the implementation and is determined by the caller's needs. The purpose of this
+   * method is to provide a hook for executing custom code or generating objects before the
+   * worldstate is closed, allowing for any necessary processing or operations to be performed.
```
siladu, line 42: This makes it really clear, thanks 👍

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
siladu: Thanks for the explanation! 

In another PR maybe, could we rename these methods to indicate the difference in cache behaviour? Or is this one of the ones that is shared with Forest so not so straightforward?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
siladu: so IIUC, it was working before because data survives in the cache even after the worldstate is closed?

siladu (commented):
Thanks for the extra comments, it really helps clear things up 👍 
Couple more questions and minor change suggestions

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
matkt: yes you got it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
matkt: as it's used by forest it can be tricky but I think at the end we need to have the code more bonsai friendly and less forest friendly. maybe when we will  have bonsai working with archive mode we can maybe remove forest

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
matkt: yes just that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
matkt: in fact it is impossible to give a type of object because precisely the person who calls this method can return the type he wants, that means that the call is not limited by a particular type. If you want to return a BigInteger you can if you want to restart a Transaction you can. It's clearly the type you will define during the call (the magic of generic type)

```java

Optional<BigInteger> process = transactionSimulator.process(null, null, null,
            (mutableWorldState, transactionSimulatorResult) -> Optional.of(BigInteger.ONE), null);

Optional<Transaction> process2 = transactionSimulator.process(null, null, null,
            (mutableWorldState, transactionSimulatorResult) -> Optional.of(transactionSimulatorResult.orElseThrow().getTransaction()), null);

Optional<Bytes> process3 = transactionSimulator.process(null, null, null,
            (mutableWorldState, transactionSimulatorResult) -> Optional.of(transactionSimulatorResult.orElseThrow().getOutput()), null);

```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
matkt: not sure this interface is not closed when the worldstate is close but before

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
siladu: I get the concept, I guess I'm just not clear why such a generic function is needed in the context of this code. Maybe a onWorldStateClosed observer pattern would avoid the unwieldy function. I don't think that refactoring is worth it though as I think with your comments it is clear enough to understand the point of this code now :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
siladu: hmm yes I see how you got to your name, tricky one.

`worldStateWillClose()` 🤷 😆 

nit: PreWorldStateCloseGuard function name is a noun rather than a verb

siladu (approved):
Good to go IMO, just some nitpicks remain

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/PreWorldStateCloseGuard.java
matkt: I will try to think better for another pr

matkt: I will use that i think ilt will be better PreCloseStateHandler

PR #5331, opened 2023-04-11
Reduce UInt256 in EVMOperations

shemnon:
## PR description

Improve the performance of the EVM by reducing the usage of UInt256 in almost all operations and moving the main switch to a Java 17 expression switch


shemnon:
Benchmark results - https://docs.google.com/spreadsheets/d/1qbvlAveoLCiWbhbxouHgcpFeDVjws7UD0dgZghJZ6oU/edit?usp=sharing

Macbook Pro M1 MAX

Top line result - 10-15% "all in"
Target opcodes - 100-500% improvement.  Most notably in JumpDest& SDIV

evm/src/main/java/org/hyperledger/besu/evm/operation/SDivOperation.java
```diff
@@ -71,10 +70,12 @@ public static OperationResult staticOperation(final MessageFrame frame) {
...
 
-      final byte[] padding = new byte[32 - resultBytes.size()];
-      Arrays.fill(padding, result.signum() < 0 ? (byte) 0xFF : 0x00);
+      frame.pushStackItem(Bytes32.leftPad(resultBytes, result.signum() < 0 ? (byte) 0xFF : 0x00));
 
-      frame.pushStackItem(UInt256.fromBytes(Bytes.concatenate(Bytes.wrap(padding), resultBytes)));
+      //      final byte[] padding = new byte[32 - resultBytes.size()];
+      //      Arrays.fill(padding, result.signum() < 0 ? (byte) 0xFF : 0x00);
+      //
+      //      frame.pushStackItem(Bytes.concatenate(Bytes.wrap(padding), resultBytes));
```
macfarla, line 78: did you mean to leave this commented code?

macfarla (commented):
a few comments on parts I don't understand 
overall I think it's fine
performance improvement sounds fab 
think a second opinion would be good :)

macfarla:
@shemnon should this get a changelog entry?

evm/src/test/java/org/hyperledger/besu/evm/operations/SarOperationTest.java
```diff
@@ -221,9 +209,9 @@ public void shiftOperation() {
     when(frame.stackSize()).thenReturn(2);
     when(frame.getRemainingGas()).thenReturn(100L);
     when(frame.popStackItem())
-        .thenReturn(UInt256.fromBytes(Bytes32.fromHexStringLenient(shift)))
-        .thenReturn(UInt256.fromHexString(number));
+        .thenReturn(Bytes32.fromHexStringLenient(shift))
+        .thenReturn(Bytes.fromHexString(number));
```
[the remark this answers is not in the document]
shemnon: That's something worth benchmarking, but the number of times we need to take a 256 bit number and consider it as a 8 or 32 bit number are very high (memory access, shifts, etc).  So my first pass was trimmed bytes, where we can assess size much quicker.

I'll look into Bytes32 or possibly ByteBuffer integrations in a future pass. It did cross my mind that consistently sized words may improve garbage collection implications and (by extension) speed.

evm/src/main/java/org/hyperledger/besu/evm/operation/ByteOperation.java
```diff
@@ -36,20 +35,21 @@ public ByteOperation(final GasCalculator gasCalculator) {
     super(0x1A, "BYTE", 2, 1, gasCalculator, gasCalculator.getVeryLowTierGasCost());
   }
 
-  private static UInt256 getByte(final UInt256 seq, final UInt256 offset) {
-    if (!offset.fitsInt()) {
-      return UInt256.ZERO;
+  private static Bytes getByte(final Bytes seq, final Bytes offset) {
+    Bytes trimmedOffset = offset.trimLeadingZeros();
+    if (trimmedOffset.size() > 1) {
+      return Bytes.EMPTY;
```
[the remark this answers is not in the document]
shemnon: Semantically they are the same on the stack, it's quicker to get an int or byte zero out of a Bytes.EMPTY than a UInt256.ZERO (UInt256 would have to check 7 ints as zero before returning the last one, and possibly bit checking for size).

The tests that detected the Tuweni 3.2 constant values bug (where Bytes32.ZERO couldn't be trimmed to 20 bytes in length) would trip this if it was a problem.

evm/src/main/java/org/hyperledger/besu/evm/operation/SDivOperation.java
shemnon: No, somehow it passed my sonar checks.  Will remove

evm/src/main/java/org/hyperledger/besu/evm/operation/AbstractCallOperation.java
```diff
@@ -184,7 +183,7 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
       frame.expandMemory(outputDataOffset(frame), outputDataLength(frame));
       frame.incrementRemainingGas(gasAvailableForChildCall(frame) + cost);
       frame.popStackItems(getStackItemsConsumed());
-      frame.pushStackItem(UInt256.ZERO);
+      frame.pushStackItem(Bytes.EMPTY);
```
garyschulte, line 186: FAILURE_STACK_ITEM here?

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
```diff
@@ -41,10 +43,10 @@ public BlockHashOperation(final GasCalculator gasCalculator) {
   @Override
   public Operation.OperationResult executeFixedCostOperation(
       final MessageFrame frame, final EVM evm) {
-    final UInt256 blockArg = UInt256.fromBytes(frame.popStackItem());
+    final Bytes blockArg = frame.popStackItem().trimLeadingZeros();
```
garyschulte, line 46: might this lead to problems if we are not aligned on a word boundary?

evm/src/main/java/org/hyperledger/besu/evm/operation/ByteOperation.java
```diff
@@ -36,20 +35,21 @@ public ByteOperation(final GasCalculator gasCalculator) {
...
     }
+    final int index = trimmedOffset.toInt();
 
-    final int index = offset.intValue();
-    if (index >= 32) {
-      return UInt256.ZERO;
+    int size = seq.size();
+    int pos = index - 32 + size;
```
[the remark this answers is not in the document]
macfarla: ahhh I skimmed over the trimLeadingZeros. makes sense now.

evm/src/main/java/org/hyperledger/besu/evm/operation/JumpDestOperation.java
```diff
@@ -21,6 +21,9 @@
 /** The Jump dest operation. */
 public class JumpDestOperation extends AbstractFixedCostOperation {
 
+  /** constant for a successful jumpdest * */
+  public static final OperationResult jumpdestSuccess = new OperationResult(1L, null);
```
garyschulte, line 25: suggest all caps on public static final

macfarla (approved):

garyschulte (approved):
the slicing is a bit scary tbh, but happy to see evm perf improvements.  This PR would be a good candidate for fuzzing on a private testnet.  What kind of testing did you do already?

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
shemnon: It won't.  The call to `blockArg.toLong()` on line 54 handles any byte length of 8 or less. That's the only use of this variable.

evm/src/main/java/org/hyperledger/besu/evm/operation/AbstractCallOperation.java
shemnon: done

evm/src/main/java/org/hyperledger/besu/evm/operation/JumpDestOperation.java
shemnon: done

evm/src/main/java/org/hyperledger/besu/evm/operation/SubOperation.java
```diff
@@ -48,12 +51,20 @@ public Operation.OperationResult executeFixedCostOperation(
    * @return the operation result
    */
   public static OperationResult staticOperation(final MessageFrame frame) {
-    final UInt256 value0 = UInt256.fromBytes(frame.popStackItem());
-    final UInt256 value1 = UInt256.fromBytes(frame.popStackItem());
+    final BigInteger value0 = new BigInteger(1, frame.popStackItem().toArrayUnsafe());
+    final BigInteger value1 = new BigInteger(1, frame.popStackItem().toArrayUnsafe());
 
-    final UInt256 result = value0.subtract(value1);
+    final BigInteger result = value0.subtract(value1);
```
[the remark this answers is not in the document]
shemnon: I expect there are some JVM optimizations targeting that stretch of code.  It doesn't look like it's directly an @Intrinsic function but it ought to be.  Also, the zero check in UInt256 is kind of expensive.

shemnon:
Before I press Merge I was going to give it a day or so on Marius's fuzzer.  It found the other corner cases fairly quickly.

non-fungible-nelson:
How did it go? @shemnon

shemnon:
Fuzz testing was useful.

Issue #5327, opened 2023-04-11
Besu memory usage, investigate OpenJ9 JDK implementation


ahamlat:
### Description
As an Besu user, I want Besu to use as less system memory as possible

The memory usage (RSS) of Besu was compared between HotSpot and OpenJ9 JDK implementations, and the results showed that OpenJ9 consumes significantly less memory due to its GenCon GC algorithm, which is better suited for Besu.

You can see below memory usage of Besu release 23.1.2 on a 32 VM with -Xmx8g.

![image](https://user-images.githubusercontent.com/5099602/231114724-08720472-0209-4e02-aea2-979df3517496.png)

We are currently in the process of testing a new OpenJ9 docker image, and one of the nodes is running on a VM with 16 GiB of RAM and the -Xmx4g flag. This configuration has resulted in reducing IO reads for Besu, likely due to the additional OS cache that is available when using less memory (note that cache reads are not counted in IO reads).

![image](https://user-images.githubusercontent.com/5099602/231120623-48422fd8-3921-4f38-a9d7-774e927eddde.png)

System OS cache + buffer when using OpenJ9 JRE
<img width="1649" alt="image" src="https://user-images.githubusercontent.com/5099602/231122858-d6fa838b-c085-4a72-a13c-97acca391df0.png">

ahamlat:
Running OpenJ9 for more than a month have given interesting results. We can see in the screenshot below that Besu memory usage was under 5.2 GiB for the last 30 days
<img width="1643" alt="image" src="https://github.com/hyperledger/besu/assets/5099602/429836ef-8816-4a77-a51e-7b0d406e2a62">

To avoid any performance overhead with OpenJ9 on Docker for ARM64, we tested also different GC tuning on G1 GC (for OpenJDK HotSpot), and have noticed good results. The GC flags used are : _**-XX:G1ConcRefinementThreads=2 -XX:G1HeapWastePercent=15 -XX:MaxGCPauseMillis=100**_

![image](https://github.com/hyperledger/besu/assets/5099602/4971ceda-ecc4-4355-820f-61d1a6072f3f)

ahamlat:
If Besu users notice big memory usage and want to reduce it, they can either use OpenJ9 JDK implementation or OpenJDK hotSpot implementation with these GC flags  _**-XX:G1ConcRefinementThreads=2 -XX:G1HeapWastePercent=15 -XX:MaxGCPauseMillis=100**_

PR #5335, opened 2023-04-12
Update OpenJ9 Docker image

ahamlat:
## PR description
Update OpenJ9 docker image to use Ubuntu 22.04 as a base image.
I have different performance results depending on the OS. OpenJ9 x86_64 docker image has better block processing time that running natively Besu with a Hotspot JVM. The screenshots below are from AWS instances m6a.2xlarge

**1. X86_64 Linux architecture**

image

OpenJ9 JVM demonstrates significantly lower memory usage compared to HotSpot JVM, resulting in a reduction of 6 GiB in this particular scenario.

image

**5 days later ( Last 6 days)**
Block processing timings
image

Memory usage
image

**2. Aarch 64 Linux architecture**

OpenJ9 aarch64 docker image block processing is 6% slower than a hotspot docker image (on aarch64). 

image

Despite the HotSpot JVM's ability to free up memory, OpenJ9 still showcases very low memory usage on aarch64 architecture. Although HotSpot has caught up with OpenJ9 and now consumes a similar amount of memory, OpenJ9 remains a solid choice for those looking to minimize their memory footprint.

image

**3 days later (Last 4 days)**
Block processing timings
image

Memory usage
image

## Fixed Issue(s)
https://github.com/hyperledger/besu/issues/5327


garyschulte (approved):
🚢

PR #5337, opened 2023-04-12
Updating confusing --bonsai-maximum-back-layers-to-load option to --bonsai-historical-block-limit + remove deprecated snapshots config 

non-fungible-nelson:
## PR description

Changed the mentioned flags. Alias used to keep backwards compatibility. Also removed now deprecated snapshots option. 

## Fixed Issue(s)
Ambiguity


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Update most dependencies to latest version [#5269](https://github.com/hyperledger/besu/pull/5269)
 - If jemalloc is used, print its version in the configuration overview [#4738](https://github.com/hyperledger/besu/pull/4738)
 - Add metrics for accounts and storage reads (Flat database vs Merkle Patricia Trie) [#5315](https://github.com/hyperledger/besu/pull/5315)
+- Renamed --bonsai-maximum-back-layers-to-load option to --bonsai-historical-block-limit for clarity [#5337](https://github.com/hyperledger/besu/pull/5337)
```
garyschulte, line 12: should mention that we removed the -Xbonsai-use-snapshots flag also, just in case some old configs were still using it

non-fungible-nelson:
gotta love more red than green on a PR

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -48,22 +45,13 @@ public class DataStorageOptions implements CLIOptions<DataStorageConfiguration>
   private final DataStorageFormat dataStorageFormat = DataStorageFormat.FOREST;
 
   @Option(
-      names = {BONSAI_STORAGE_FORMAT_MAX_LAYERS_TO_LOAD},
+      names = {BONSAI_STORAGE_FORMAT_MAX_LAYERS_TO_LOAD, "--bonsai-maximum-back-layers-to-load"},
       paramLabel = "<LONG>",
       description =
           "Limit of back layers that can be loaded with BONSAI (default: ${DEFAULT-VALUE}).",
```
macfarla, line 51:
```suggestion
          "Limit of historical layers that can be loaded with BONSAI (default: ${DEFAULT-VALUE}).",
```

macfarla (approved):

macfarla:
there is one unit test failing
```
java.lang.AssertionError: 
Expecting actual:
  "Invalid value for option '--bonsai-maximum-back-layers-to-load': 'ten' is not a long

To display full help:
besu [COMMAND] --help
"
to contain:
  "Invalid value for option '--bonsai-historical-block-limit': 'ten' is not a long" 
	at org.hyperledger.besu.cli.BesuCommandTest.parsesInvalidBonsaiTrieBonsaiHistoricalBlockLimit(BesuCommandTest.java:1912)
```

macfarla:
> there is one unit test failing


@non-fungible-nelson I believe PicoCLI is choosing the `longest` name to throw into that error message. 

You could look at how host-whitelist is handled - it's a hidden alternative (separate option) to host-allowlist - but then you have to handle the case where both are supplied

also if you don't hide that option name it will still display in the help ->

```
      --bonsai-historical-block-limit,
        --bonsai-maximum-back-layers-to-load=<LONG>
                             Limit of historical layers that can be loaded with
                               BONSAI (default: 512).
```

Alternatively - if we don't mind users still seeing the old name, you could just remove that test - it's not critical

non-fungible-nelson:
I think seeing both is OK if you're querying for help directly. As long as we are clear in the docs and it supplies to the client correctly.

PR #5338, opened 2023-04-13
Set version as 23.1.3 and minor changelog update

siladu:


ethereum/referencetests/src/reference-test/external-resources
jframe, line 1: Did you mean to change this?

CHANGELOG.md
```diff
@@ -2,10 +2,6 @@
 
 ## 23.1.3
```
jframe, line 4: We should add a note that this is a hotfix for the nimbus issue

ethereum/referencetests/src/reference-test/external-resources
siladu: that guy again! good spot

siladu: ah, actually it's fine I think. It'll have been wrong on the release branch

jframe (approved):
LGTM

PR #5339, opened 2023-04-13
Remove RocksDb deprecated option maxBackgroundCompactions 

gfukushima:
## PR description
This PR cleans up the besu rocksDB plugin removing all the `maxBackgroundCompactions` usage. Might be useful to get a following PR to introduce `maxBackgroundJobs` if we still want to support this type of customisation in the plugin


macfarla (approved):
LGTM - just needs a changelog entry

ahamlat (approved):
LGTM, I just wanted to add this comment from RocksDB documentation :
_"NOT SUPPORTED ANYMORE: RocksDB automatically decides this based on the value of max_background_jobs. For backwards compatibility we will set `max_background_jobs = max_background_compactions + max_background_flushes` in the case where user sets at least one of `max_background_compactions` or `max_background_flushes` (we replace -1 by 1 in case one option is unset). Specifies the maximum number of concurrent background compaction jobs, submitted to the default LOW priority thread pool. If you're increasing this, also consider increasing number of threads in LOW priority thread pool. For more information, see Default: -1"_
I also checked on the instance running this PR, the value of Options.max_background_compactions is now -1
image

It was 4 before
image

Issue #2569, opened 2021-07-21
Improve error message when verifying enode syntax with xdns-enabled


jlcs-es:
### Description

When deploying multiples nodes on private network, the same `BESU_BOOTNODES` env variable is set with all the enodes.
This is ok as a node will detect its own enode.

Now, in a private networking environment such a Kubernetes, the option `BESU_XDNS_ENABLED` is set to true so we can point to the Pods hostnames instead of IPs that will change.

Internally, Besu seems to retrieve from DNS the IP and then substitute it in the enode string.
If any of the nodes listed on the `BESU_BOOTNODES` option is not yet ready, the Kubernetes DNS will not report any IP, and then the Besu command will fail with this message:

```log
Invalid enode URL syntax. Enode URL should have the following format 'enode://<node_id>@<ip>:<listening_port>[?discport=<discovery_port>]'.  Invalid ip address.
```

If the option `BESU_XDNS_UPDATE_ENABLED` is also enabled, Besu does not substitute any IP in the enode string and passes the boot up validations.

**Expected behavior:**

If `BESU_XDNS_ENABLED` is set to true, but `BESU_XDNS_UPDATE_ENABLED` isn't, the error message displayed should be more clear on what is actually happening.

Example:

```
Invalid ip address or DNS query resolved an invalid ip.
```

### Versions
* Software version: `besu/v21.7.1/linux-x86_64/adoptopenjdk-java-11` (docker latest tag at the time of writting).

matkt:
thanks for your feedback we will try to improve the message in this case

matt-nelson-csi:
@daniellehrner easy logging change?

NickSneo:
@non-fungible-nelson @macfarla Shall we close this issue?

non-fungible-nelson:
all you Nick! You handled the changes ;)

PR #5346, opened 2023-04-14
Improved error message when verifying enode syntax with xdns-enabled

NickSneo:
## PR description
Improved error message when verifying enode syntax with xdns-enabled

## Fixed Issue(s)
https://github.com/hyperledger/besu/issues/2569

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/peers/EnodeURLImpl.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/peers/EnodeURLImpl.java
@@ -80,13 +80,19 @@ public class EnodeURLImpl implements EnodeURL {
       checkStringArgumentNotEmpty(value, "Invalid empty value.");
       return fromURI(URI.create(value), enodeDnsConfiguration);
     } catch (final IllegalArgumentException e) {
-      String message =
-          String.format(
-              "Invalid enode URL syntax '%s'. Enode URL should have the following format 'enode://<node_id>@<ip>:<listening_port>[?discport=<discovery_port>]'.",
-              value);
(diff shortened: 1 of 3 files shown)


NickSneo:
the test fromString_withHostnameEnodeURLShouldWorkWhenDnsEnabledAndUpdateEnabled() in unitTests which is failing is not due to my main changes. Actually it was incorrect before since the function is with dns UpdateEnabled but it was formerly set to false and I just corrected it. Also it is passing in my local setup.

NickSneo:
@macfarla Can you plz review?

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/peers/EnodeURLImpl.java
```diff
@@ -80,13 +80,20 @@ public static EnodeURL fromString(
...
-      String message =
-          String.format(
-              "Invalid enode URL syntax '%s'. Enode URL should have the following format 'enode://<node_id>@<ip>:<listening_port>[?discport=<discovery_port>]'.",
-              value);
-      if (e.getMessage() != null) {
-        message += " " + e.getMessage();
+      String message = "";
+      if (enodeDnsConfiguration.dnsEnabled() && !enodeDnsConfiguration.updateEnabled()) {
+        message =
+            "Invalid IP address or DNS query resolved an invalid IP. --Xdns-enabled is true but --Xdns-update-enabled flag is false.";
```
macfarla, line 86:
```suggestion
            "Invalid IP address (or DNS query resolved an invalid IP). --Xdns-enabled is true but --Xdns-update-enabled flag is false.";
```

macfarla (approved):

PR #5356, opened 2023-04-18
[MINOR] junit5

macfarla:
Migrate some tests to junit5; remove some unnecessary mockings


besu/src/test/java/org/hyperledger/besu/RunnerBuilderTest.java
```diff
@@ -72,25 +72,25 @@
...
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.jupiter.MockitoExtension;
 
-@RunWith(MockitoJUnitRunner.class)
+@ExtendWith(MockitoExtension.class)
 public final class RunnerBuilderTest {
 
-  @Rule public TemporaryFolder dataDir = new TemporaryFolder();
+  @TempDir private static Path dataDir;
```
jframe, line 93: Don't think this needs to be static

besu/src/test/java/org/hyperledger/besu/RunnerTest.java
```diff
@@ -159,7 +159,7 @@ public void fastSyncFromGenesis() throws Exception {
 
   private void syncFromGenesis(final SyncMode mode, final GenesisConfigFile genesisConfig)
       throws Exception {
-    final Path dataDirAhead = temp.newFolder().toPath();
+    final Path dataDirAhead = Files.createTempDirectory(temp, "db-ahead");
```
siladu, line 162: I thought @TempDir created the dir automatically?

besu/src/test/java/org/hyperledger/besu/RunnerBuilderTest.java
```diff
@@ -393,4 +399,12 @@ public void noEngineApiNoServiceForMethods() {
     assertThat(runner.getJsonRpcPort()).isPresent();
     assertThat(runner.getEngineJsonRpcPort()).isEmpty();
   }
+
+  private void setupBlockchainAndBlock() {
```
gfukushima, line 403: can't we just call this method once in the setup()? Just trying to understand what the advantages of having it called in each test instead of in the setup

siladu (approved):

gfukushima (approved):

besu/src/test/java/org/hyperledger/besu/RunnerTest.java
macfarla: it does but for runnerbuilder we create the database dir and we also need the parent dir

jframe (approved):

Issue #5354, opened 2023-04-17
UnifiedProtocolSchedule Tidy up


siladu:
Separated some low impact changes out of [this PR](https://github.com/hyperledger/besu/pull/5310) as it was already too big: 

- [x] Fold AbstractProtocolScheduleBuilder into ProtocolScheduleBuilder
- [x] Tidy up WithdrawalsValidatorProvider
- [x] Maybe no need to createSchedule inside JsonRpcMethodsFactory, as per https://github.com/hyperledger/besu/pull/5310#discussion_r1172086249
- [x] Unify StreamingProtocolSchedules and extend UnifiedProtocolSchedule
- [x] Rename UnifiedProtocolSchedule -> DefaultProtocolSchedule
- [ ] Get rid of UnifiedProtocolSchedule.putMilestone

[New Issue] AT for ensuring shanghai EVM changes are enabled when shanghai is enabled

PR #5357, opened 2023-04-18
Remove merge block processor as block rewards are already disabled in protocol schedule

jframe:
## PR description
Remove the merge block processor as block rewards are already disabled in the protocol schedule by using skipZeroBlockRewards=true and blockreward=0.

Testing that syncing still works
- [x] Local test using Beku
- [x] Bonsai snapsync on Goerli
- [x] Bonsai snapsync on Sepolia
- [ ] Bonsai snapsync on Mainnet

## Fixed Issue(s)

Part of #5354


jframe:
Testing that syncing still works
- [ ] Bonsai snapsync on Goerli
- [ ] Bonsai snapsync on Sepolia
- [ ] Bonsai snapsync on Mainnet

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeProtocolSchedule.java
```diff
@@ -93,7 +93,6 @@ private static ProtocolSpecBuilder applyMergeSpecificModifications(
             (gasCalculator, jdCacheConfig) ->
                 MainnetEVMs.paris(
                     gasCalculator, chainId.orElse(BigInteger.ZERO), EvmConfiguration.DEFAULT))
-        .blockProcessorBuilder(MergeBlockProcessor::new)
```
gfukushima, line 96: Isn't the MergeBlockProcessor Optimized in terms of all the rewardCoinbase code path? If we remove this and start calling the rewardCoinbase method from the MainnetBlockProcessor aren't we executing more checks and code than we're supposed to?

jframe: I don't think the MergeBlockProcessor is that optimised. We exit the rewardCoinbase method early due to https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockProcessor.java#L55 with two very simple checks as opposed to the single isPostMerge check from the MergeBlockProcessor.

gfukushima: If what you mentioned in the PR description is true: skipZeroBlockRewards=false then that if is probably skipped

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeProtocolSchedule.java
```diff
@@ -104,11 +103,8 @@ private static ProtocolSpecBuilder applyMergeSpecificModifications(
 
   private static ProtocolSpecBuilder unapplyMergeModificationsFromShanghaiOnwards(
       final ProtocolSpecBuilder specBuilder) {
-    // TODO Withdrawals Get rid of MergeBlockProcessor
     // inherits merge config from MainnetProtocolSpecs.parisDefinition
-    // This would be Function.identify() but MergeBlockProcessor can't be used in
-    // MainnetProtocolSpecs due to circular dependency
-    return specBuilder.blockProcessorBuilder(MergeBlockProcessor::new);
+    return specBuilder;
```
siladu, line 107: Wondering if we should use Function::identity instead now? 
I think it's better not to leave this method around to discourage future changes.

I like that this method has a name though...maybe creating a named FunctionalInterface version of identity() would be a good compromise?

siladu (approved):
Just a nitpick really

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeProtocolSchedule.java
jframe: I left the function just for the name. A comment would probably just as well though. So will remove it add a comment instead.

jframe: done

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeProtocolSchedule.java
jframe: The skipZeroBlockRewards is set to true from spurious dragon onwards https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java#L267. So for Paris onwards where blockrewards=0 then we return immediately without trying apply rewards.

gfukushima: You're are right, spurious dragon changes that to true, I was misled by the PR description.

gfukushima (approved):

PR #5371, opened 2023-04-20
Retire ECIP-1049 network

diega:
## PR description
With the [withdraw](https://github.com/ethereumclassic/ECIPs/pull/486) of ECIP-1049 we can remove the network meant for testing.
This PR also removes Keccak pow/hasher but I'm good with keeping those.


diega:
@macfarla I continued syncing an ETC mainnet full bonsai node and I started to full sync Mordor and ETH mainnet bonsai nodes. All without issues after ~2 hours

PR #5372, opened 2023-04-20
Bonsai TrieLogFactory serializer/deserializer

garyschulte:
## PR description
This PR introduces a TrieLogFactory interface and default implementation that will allow for a pluggable implementation of bonsai trie log serialization and deserialization.  

This pr also introduces a StorageSlotKey record type that enables accumulators and serializers to utilize slotKeys and/or slotKeyHashes when writing trieLogs.

## Fixed Issue(s)

related to protocol-misc # 755


macfarla (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogLayer.java
```diff
@@ -98,10 +98,7 @@ public TrieLogLayer addAccountChange(
   public TrieLogLayer addCodeChange(
       final Address address, final Bytes oldValue, final Bytes newValue, final Hash blockHash) {
     checkState(!frozen, "Layer is Frozen");
-    code.put(
-        address,
-        new BonsaiValue<>(
-            oldValue == null ? Bytes.EMPTY : oldValue, newValue == null ? Bytes.EMPTY : newValue));
+    code.put(address, new BonsaiValue<>(oldValue, newValue, newValue == null));
```
garyschulte, line 101: @matkt I believe we were previously returning Bytes.EMPTY here because the old BonsaiLayeredWorldState would serve state directly from TrieLogs.  

Since the 5123 bonsai refactor, we should be able to treat this consistently as null.  Otherwise a constructed trielog != that same trielog serialized and then deserialized.  It was the difference between code == null vs code == 0x.

ahamlat:
Based on my observations, I haven't noticed any decrease in performance. The time it takes to process blocks is very similar, and after analyzing the CPU profiling, I haven't identified any additional strain.

Screenshot 2023-04-21 at 09 39 27

PR #5392, opened 2023-04-26
remove zero lead bytes of log and not save log bloom filder

matkt:


non-fungible-nelson:
Block headers and receipts - 0s

PR #5405, opened 2023-04-28
Decompose putMilestone into putBlockNumberMilestone and putTimestampMilestone

### LINKED ISSUE #5354: UnifiedProtocolSchedule Tidy up

siladu:
Minor refactor for readability and clarity.

Originally, I was trying to avoid timestamp or blockNumber details leaking outside of the ProtocolSchedule and its Builder.
I wanted to remove putMilestone from the public interface entirely, since for the mainnet use case, this detail can be kept inside the ProtocolScheduleBuilder. 
This would have meant building the milestones first and passing them into ProtocolSchedule's constructor, but this was not feasible given the way the ProtocolScheduleBuilder works, specifically the fact that ProtocolSpecBuilder relies on receiving a ProtocolSchedule which implies milestones must be added after the ProtocolSchedule itself has been constructed.

Since my original plan is not feasible, I think this is a better approach if we have to support a public put operation anyway.

Fixes #5354


jframe (approved):
LGTM

PR #5406, opened 2023-04-28
Ensure parisSpecificModifications are unapplied for all subsequent forks

### LINKED ISSUE #5404: Ensure mergeSpecificModifications are unapplied for all subsequent forks
The current implementation has a bug that relies on shanghaiTime being configured which might not be the case for future devnets/testnets, e.g. if they start with cancunTime: 0 and don't configure shanghaiTime at all.

https://github.com/hyperledger/besu/blob/4875ba3a64d9a438c452e5c855a6fe2f83d5f5d8/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/MergeProtocolSchedule.java#L72-L75

A better approach would be to use `config.getForkBlockTimestamps().stream().findFirst().isPresent()`

siladu:
...not just Shanghai

This is because you could have a genesis configured with e.g. cancunTime but leave shanghaiTime unconfigured.

Fixes #5404


jframe (approved):
LGTM

PR #5423, opened 2023-05-03
Remove block creation task map entries when cancelling

siladu:
Fixes potential memory leak - probably only significant for clients running lots of validators

## Testing

- [x] beku-timestamp with validators


fab-10 (approved):
LGTM.
No related to this PR, but would you consider renaming `blockCreationTask` to `blockCreationTasks`?

siladu:
> LGTM. No related to this PR, but would you consider renaming `blockCreationTask` to `blockCreationTasks`?

I think only one task exists at any one time now, so we could potentially make it a singleton map or another data structure even.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -308,21 +308,26 @@ private void cancelAnyExistingBlockCreationTasks(final PayloadIdentifier payload
...
         (pid, blockCreationTask) -> {
           blockCreationTask.cancel();
-          return blockCreationTask;
+          return null;
         });
   }
 
+  @Override
+  public void finalizeProposalById(final PayloadIdentifier payloadId) {
+    LOG.debug("Finalizing block proposal for payload id {}", payloadId);
```
jframe, line 327: Is this going to spam the logs? Most of the time I think we would cancel the block creation as we are constantly trying to rebuild a better one.

siladu: I haven't changed anything with this, just refactored a bit. This is logged every time engine_getPayload is called, it's the happy path.

jframe (approved):
LGTM

siladu:
> > LGTM. No related to this PR, but would you consider renaming `blockCreationTask` to `blockCreationTasks`?
> 
> I think only one task exists at any one time now, so we could potentially make it a singleton map or another data structure even.

@fab-10 I went your rename suggestion since this is a larger refactor, and the plural is more accurate according to current implementation https://github.com/hyperledger/besu/pull/5423/commits/427c37511374b8b145e931cb2a15d83c15bbb6d9

PR #5424, opened 2023-05-03
Acceptance Test for Shanghai EVM changes (push0)

### LINKED ISSUE #5403: Acceptance Test for shanghai EVM changes
This is [already tested at the unit test level](https://github.com/hyperledger/besu/blob/4875ba3a64d9a438c452e5c855a6fe2f83d5f5d8/consensus/merge/src/test/java/org/hyperledger/besu/consensus/merge/MergeProtocolScheduleTest.java#L58-L82) and we should first check if this type of test is already present in Hive.

If not, then an Acceptance Test for ensuring shanghai EVM changes are enabled when shanghai is enabled would be useful. The context is that MergeProtocolSchedule.applyMergeSpecificModifications has now been responsible twice for causing a bug where the shanghai (or subsequent forks) configuration is reverted back to paris. For Shanghai, this only has impact on the EVM configuration.

One way to do this would be to send a transaction containing a PUSH0.

siladu:
Fixes #5403

Splits Shanghai AT test cases out from paris.
Idea is to have an AT per "fork transition", i.e. ExecutionEngineShanghaiAcceptanceTest for paris -> shanghai
next one would be ExecutionEngineCancunAcceptanceTest for shanghai -> cancun


acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/jsonrpc/ExecutionEngineAcceptanceTest.java
```diff
@@ -25,8 +25,8 @@
 
 @RunWith(Parameterized.class)
 public class ExecutionEngineAcceptanceTest extends AbstractJsonRpcTest {
```
jframe, line 27: Suggest that this test is renamed to ExecutionEngineParisAcceptanceTest

acceptance-tests/tests/src/test/resources/jsonrpc/engine/shanghai/genesis.json
```diff
@@ -0,0 +1,65 @@
...
...1717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171717171760019055",
```
jframe, line 40: Is this to test the max init size changes?

siladu: It turned out not to be used, I think it's only `...00000100` that's being called by the push0 transaction in the test.

I copied it from https://github.com/ethereum/tests/blob/develop/GeneralStateTests/Shanghai/stEIP3855-push0/push0.json#L158-L178 as I wasn't sure which one I would need initially.

I suspect might be related to:
```
                "0" : "single_push0",
                "1" : "1024_push0",
                "2" : "1025_push0",
```

jframe (approved):

PR #5428, opened 2023-05-03
Fix the number of RocksDB log files to the last one week

ahamlat:
Signed-off-by: Ameziane H <ameziane.hamlat@consensys.net

## PR description
Fix the number of RocksDB LOG files to 5 with 100 MB for each LOG file.
These logs are not to be confused with [number].log files which represent WAL files.

EDIT : Following Simon suggestion and Fabio's idea, I changed the policy retention to keep 7 daily log files.


plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -78,6 +78,10 @@
   protected static final long ROCKSDB_BLOCKCACHE_SIZE_HIGH_SPEC = 1_073_741_824L;
   /** RocksDb memtable size when using the high spec option */
   protected static final long ROCKSDB_MEMTABLE_SIZE_HIGH_SPEC = 1_073_741_824L;
+  /** RocksDb number of log files to keep on disk */
+  private static final long NUMBER_OF_LOG_FILES_TO_KEEP = 5;
+  /** RocksDb Max size of each log file */
+  private static final long MAX_LOG_FILE_SIZE = 104_857_600L;
```
siladu, line 84: Does this mean we will always fill up 500MB worth of logs if the node has been up for long enough, or might the files get rotated for reasons other than hitting the 100MB size?

siladu (commented):
My preference would be to default to a lot less than 500MB overall log size and allow the user to configure it themselves if they want more for debugging purposes. I'm not sure what the appropriate size might be, but keeping logs for a few days to a week seems reasonable.

siladu:
Will this automatically cleanup any existing LOG.old* files that already exist?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
ahamlat: From what I noticed, there is a new log file for each restart.

ahamlat:
I didn't find a way in RocksDB API to limit the size of the LOG files to some period, like the last week. 
Form what I understand setLogFileTimeToRoll will specify the time interval for the info log file to roll (in seconds), but you still need to define the number of log files.
https://github.com/facebook/rocksdb/blob/a11f1e12ca95ae6795f0a0674c3a3dbf82c157e9/logging/auto_roll_logger.cc#L36
My estimation on a staking node is between 15 and 20 MB per day, this rate is bigger during sync period as there're more compaction and more workload in general.

siladu:
> My estimation on a staking node is between 15 and 20 MB per day, this rate is bigger during sync period as there're more compaction and more workload in general.

How about 7 files with a max of 20MB each?

ahamlat:
> How about 7 files with a max of 20MB each?

It is actually possible like you suggested in the first place, we just need to set the time to roll to one day and keep seven files. I'm going to do it, thanks to @fab-10

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -78,6 +78,10 @@
   protected static final long ROCKSDB_BLOCKCACHE_SIZE_HIGH_SPEC = 1_073_741_824L;
   /** RocksDb memtable size when using the high spec option */
   protected static final long ROCKSDB_MEMTABLE_SIZE_HIGH_SPEC = 1_073_741_824L;
+  /** RocksDb number of log files to keep on disk */
+  private static final long NUMBER_OF_LOG_FILES_TO_KEEP = 5;
```
fab-10, line 82: In the PR title you say one week, while here the number is 5, should it be 7?

fab-10 (commented):
LGTM, just a question

fab-10 (approved):

macfarla (approved):

siladu (approved):
LGTM, glad the time roll worked :)

PR #5429, opened 2023-05-03
Move Big EOF out of Cancun

shemnon:
Move the implementation of "big" eof into Future EIPs, out of cancun. Also, create space for Prague, Osaka, and Bogota forks in the EVM code.


evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
```diff
@@ -419,7 +419,7 @@ public static EVM constantinople(
   }
 
   /**
-   * Constantinople operations registry.
+   * Constantinople operations operation registry.
```
siladu, line 422: nit: "operations operation" reads weird to me

evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
```diff
@@ -816,7 +1010,7 @@ public static void registerCancunOperations(
    * @return the evm
    */
   public static EVM futureEips(final BigInteger chainId, final EvmConfiguration evmConfiguration) {
-    return shanghai(chainId, evmConfiguration);
+    return futureEips(new CancunGasCalculator(), chainId, evmConfiguration);
```
siladu, line 1013: Can't comment inline as it's not in the diff, but I think the experimentalEips equivalent method is missing `new CancunGasCalculator()`

siladu: Why do we need the new it up at this level anyway?

siladu (approved):
LGTM, couple of minor comments

evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
shemnon: Buffalo buffalo buffalo buffalo buffalo.

operation registry for the operations in Constantinople.

evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
shemnon: Every EVM needs a gas calculator.  I suppose we could use static instance since they don't have state.

shemnon: Making it configurable also allows for gas schedules to be change by downstream consumers.

PR #5430, opened 2023-05-03
Eip 6780 selfdestruct

shemnon:
* Wire in a check to see if an account is new in the TX scope
* Create a new SelfDestruct Opcode that only deletes accounts
  that register as new within tx scope.


evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructEIP6780Operation.java
```diff
@@ -0,0 +1,82 @@
...
+    if (frame.isStatic()) {
+      return new OperationResult(cost, ExceptionalHaltReason.ILLEGAL_STATE_CHANGE);
+    } else if (frame.getRemainingGas() < cost) {
+      return new OperationResult(cost, ExceptionalHaltReason.INSUFFICIENT_GAS);
+    }
+
+    final Address address = frame.getRecipientAddress();
+    final MutableAccount account = frame.getWorldUpdater().getAccount(address).getMutable();
+
+    if (account.isNewAccount()) {
```
jframe, line 63: Would prefer to pass in a flag to enable this extra check. Had to compare this and the `SelfDestructEIP6780Operation` to see that this extra if was the only difference. That and there is a bit of duplication between the two.

shemnon: done

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -27,13 +27,28 @@
 /** The Self destruct operation. */
 public class SelfDestructOperation extends AbstractOperation {
 
+  final boolean eip6780Semantics;
```
siladu, line 30:
```suggestion
  final boolean eip6780Enabled;
```

evm/src/test/java/org/hyperledger/besu/evm/operations/SelfDestructOperationTest.java
```diff
@@ -0,0 +1,185 @@
...
+  @Mock private WorldUpdater worldUpdater;
+  @Mock private WrappedEvmAccount account;
+  @Mock private MutableAccount mutableAccount;
+  @Mock private EVM evm;
+  @Mock private MutableAccount newMutableAccount;
+
+  private final SelfDestructOperation operation =
+      new SelfDestructOperation(new ConstantinopleGasCalculator());
+
+  private final SelfDestructOperation newOperation =
```
siladu, line 68:
```suggestion
  private final SelfDestructOperation eip6780Operation =
```

siladu, line 68:
```suggestion
  private final SelfDestructOperation eip6780Operation =
```

evm/src/test/java/org/hyperledger/besu/evm/operations/SelfDestructOperationTest.java
```diff
@@ -0,0 +1,185 @@
...
+          );
+
+  private MessageFrame messageFrame;
+  @Mock private WorldUpdater worldUpdater;
+  @Mock private WrappedEvmAccount account;
+  @Mock private MutableAccount mutableAccount;
+  @Mock private EVM evm;
+  @Mock private MutableAccount newMutableAccount;
+
+  private final SelfDestructOperation operation =
```
siladu, line 65:
```suggestion
  private final SelfDestructOperation frontierOperation =
```

siladu, line 65:
```suggestion
  private final SelfDestructOperation frontierOperation =
```

evm/src/main/java/org/hyperledger/besu/evm/worldstate/UpdateTrackingAccount.java
```diff
@@ -148,6 +148,11 @@ public Map<UInt256, UInt256> getUpdatedStorage() {
     return updatedStorage;
   }
 
+  @Override
+  public boolean isNewAccount() {
+    return account == null;
```
siladu, line 153: Same here, not clear why this means it's a new account.

For example, https://github.com/hyperledger/besu/pull/5430/files#diff-7f0493f7e9a429807ad4302ee3c0d24aaae4bebcb4ee73f9e2148db4b50d80bbR147

```
return new WrappedEvmAccount(track(new UpdateTrackingAccount<>(newAccount)))
```

we create a new account and pass it to UpdateTrackingAccount constructor, but this isNewAccount would return false.

siladu (commented):
Looks pretty straight forward overall 👍 

Some questions and name change suggestions

evm/src/test/java/org/hyperledger/besu/evm/operations/SelfDestructOperationTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    setUp(contract, beneficiary, balanceHex);
+    final Operation.OperationResult operationResult = operation.execute(messageFrame, evm);
+    assertThat(operationResult).isNotNull();
+    assertThat(messageFrame.getSelfDestructs()).contains(Address.fromHexString(contract));
+
+    verify(account).getBalance();
+    verify(account).isEmpty();
+    if (!contract.equals(beneficiary)) {
+      verify(mutableAccount).incrementBalance(Wei.fromHexString(balanceHex));
```
jflo, line 151: The variables `account` and  `mutableAccount` representing both the contract and a potential beneficiary makes this test hard to read. Is there a more clear way we can name these?

jflo (commented):
Functionally this looks good, but I've left comments with regard to readability.

evm/src/test/java/org/hyperledger/besu/evm/operations/SelfDestructOperationTest.java
```diff
@@ -0,0 +1,185 @@
...
+    } else {
+      assertThat(messageFrame.getSelfDestructs()).isEmpty();
+    }
+
+    verify(account).getBalance();
+    verify(account).isEmpty();
+    if (!contract.equals(beneficiary)) {
+      verify(mutableAccount).incrementBalance(Wei.fromHexString(balanceHex));
+    }
+    verify(mutableAccount).setBalance(Wei.ZERO);
```
siladu, line 181: What does it mean for the contract to be equal to the beneficiary?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiAccount.java
```diff
@@ -141,7 +147,20 @@ public static BonsaiAccount fromRLP(
...
-        context, address, Hash.hash(address), nonce, balance, storageRoot, codeHash, mutable);
+        context,
+        address,
+        Hash.hash(address),
+        nonce,
+        balance,
+        storageRoot,
+        codeHash,
+        mutable,
+        false);
```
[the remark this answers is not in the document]
shemnon: If we are reading from RLP that means it's been serialized and we are reading from disk, hence not new in the transaction.  There is no account serialziation within the scope of a transaction.

An account can have a zero nonce and not be new if someone sent funds to it speculatively, either as a burn or planning for it to match an EOA or future contract create.

evm/src/test/java/org/hyperledger/besu/evm/operations/SelfDestructOperationTest.java
shemnon: I refactored the tests so that benificiary and contract accounts are no longer shared.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiAccount.java
```diff
@@ -107,6 +111,7 @@ public BonsaiAccount(
     updatedStorage.putAll(toCopy.updatedStorage);
 
     this.mutable = mutable;
+    this.newAccount = false;
```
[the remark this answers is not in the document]
shemnon: done.

evm/src/main/java/org/hyperledger/besu/evm/worldstate/UpdateTrackingAccount.java
shemnon: Ok, in the cases where the new account is mutable I changed the code to delegate to that account for newness checking, otherwise it require absence of a parent to be the indicator for newness.

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleAccount.java
```diff
@@ -82,6 +82,11 @@ public SimpleAccount(
     this.code = code;
   }
 
+  @Override
+  public boolean isNewAccount() {
+    return parent == null;
```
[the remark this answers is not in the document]
shemnon: This is for self-contained executions.  If the account exists in the state file (basically the genesis file) then it is not new and would have the account created referencing that parent account.  If it's not in the state file then it has no parent.

I noticed that this class has leaked into the TransactionPool, for transaction validation.  (a) there is no EVM execution there and (b) the TransactionPool probably shouldn't be using it and should be creating it's own class.

evm/src/test/java/org/hyperledger/besu/evm/operations/SelfDestructOperationTest.java
shemnon: Contract address is the same as the beneficiary address for the self destruct.  This is reflecting implementation details where we don't increment the account being self-destructed if it's the same account.

shemnon: Balance is indeed ether is lost.  State tests confirm this is the expected outcome.

```
./ethereum/evmtool/build/install/evmtool/bin/evm state-test ./ethereum/referencetests/src/reference-test/external-resources/GeneralStateTests/stSystemOperationsTest/suicideAddress.json --json --json-alloc 
-------8<--------
{"pc":0,"op":48,"gas":"0xef038","gasCost":"0x2","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"ADDRESS"}
{"pc":1,"op":96,"gas":"0xef036","gasCost":"0x3","memSize":0,"stack":["0x95e7baea6a6c7c4c2dfeb977efac326af552d87"],"depth":1,"refund":0,"opName":"PUSH1"}
{"pc":3,"op":85,"gas":"0xef033","gasCost":"0x5654","memSize":0,"stack":["0x95e7baea6a6c7c4c2dfeb977efac326af552d87","0x0"],"depth":1,"refund":0,"opName":"SSTORE"}
{"pc":4,"op":48,"gas":"0xe99df","gasCost":"0x2","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"ADDRESS"}
{"pc":5,"op":255,"gas":"0xe99dd","gasCost":"0x1388","memSize":0,"stack":["0x95e7baea6a6c7c4c2dfeb977efac326af552d87"],"depth":1,"refund":0,"opName":"SELFDESTRUCT"}
{
 "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b": {
  "balance": "0xde0b6b3a75b2232",
  "nonce": "0x1"
 },
}
{"output":"","gasUsed":"0xbbeb","time":3454083,"Mgps":"13.928","test":"suicideAddress","fork":"Shanghai","d":0,"g":0,"v":0,"postHash":"0x74d6e975dfc354eca790f5a3bee573e0019bb92967c9131ddc2dc622f297e266","postLogsHash":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","pass":true}
```

`0x95e7baea6a6c7c4c2dfeb977efac326af552d87` was the executing contract, and had balance

jflo (approved):

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleAccount.java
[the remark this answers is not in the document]
siladu: cc @fab-10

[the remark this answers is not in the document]
fab-10: I am just doing some work there, and will remove that reference

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -60,7 +75,9 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
     final Address address = frame.getRecipientAddress();
     final MutableAccount account = frame.getWorldUpdater().getAccount(address).getMutable();
 
-    frame.addSelfDestruct(address);
+    if (!eip6780Semantics || account.isNewAccount()) {
```
garyschulte, line 78: If we were processing a (malformed) block that included the account creation in a prior transaction, would the account still be marked as newAccount and allow a self destruct in a subsequent transaction?  

At least in the case of bonsai,  I would expect the accumulator would have the account cached with isNewAccount == true, since I don't see anywhere where we are resetting that flag at transaction boundary.

garyschulte (commented):
Do we have pre-existing self destruct reference tests that validate the behavior of self destruct recipient address the same as the self destruct address?

shemnon:
./ethereum/referencetests/src/reference-test/external-resources/GeneralStateTests/stSystemOperationsTest/suicideAddress.json handles the self-destruct to self question.  RefTests haven't been filled for Cancun.  I'll add a evmtool test based off of this.

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
shemnon: Change of plans, I re-wrote the new account detection to not depend on changes to the Account interface.

garyschulte (approved):
👍

shemnon:
Waiting on https://github.com/ethereum/EIPs/pull/7308 to settle before check in.  We pass execution-spec-tests as the exist right now, without alteration.

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
siladu: Looks like this wasn't changed or the old name crept back in

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -60,19 +75,22 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
     final Address address = frame.getRecipientAddress();
     final MutableAccount account = frame.getWorldUpdater().getAccount(address).getMutable();
 
-    frame.addSelfDestruct(address);
+    if (!eip6780Semantics || frame.wasCreatedInTransaction(address)) {
+      frame.addSelfDestruct(address);
+    }
 
     final MutableAccount recipient =
```
siladu, line 82: Think it would help the understanding of this PR to rename this to `beneficiary` (or even`selfDestructBeneficiary`) to match the unit test and [the spec](https://ethereum.github.io/execution-specs/autoapi/ethereum/shanghai/vm/instructions/system/index.html?highlight=selfdestruct#ethereum.shanghai.vm.instructions.system.selfdestruct).

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -60,19 +75,22 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
     final Address address = frame.getRecipientAddress();
```
siladu, line 75: Perhaps a little verbose, but while reviewing, I renamed this to `addressOfAccountToMaybeDestruct`

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -60,19 +75,22 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
...
+    }
 
     final MutableAccount recipient =
         frame.getWorldUpdater().getOrCreate(recipientAddress).getMutable();
 
-    if (!account.getAddress().equals(recipient.getAddress())) {
-      recipient.incrementBalance(account.getBalance());
+    Wei contractBalance = account.getBalance();
+    if (eip6780Semantics || !account.getAddress().equals(recipient.getAddress())) {
+      recipient.incrementBalance(contractBalance);
```
siladu, line 87: With the renames, this statement becomes a lot clearer IMO...

`!addressOfAccountToMaybeDestruct.equals(selfDestructBeneficiary.getAddress())) {
      selfDestructBeneficiary.incrementBalance(balanceOfDestructedAccount);`

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -60,19 +75,22 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
...
     }
 
     // add refund in message frame
-    frame.addRefund(recipient.getAddress(), account.getBalance());
+    frame.addRefund(recipient.getAddress(), contractBalance);
 
-    account.setBalance(Wei.ZERO);
+    account.decrementBalance(contractBalance);
```
siladu, line 93: Might be worth a comment to explain the case where this balance should not be set to zero, as before.

Seems like for most cases this is equivalent to setting to zero, except the case where the contract account being "hollowed" (because it was created in a prior transaction) had an existing balance in it...IIUC, this is a rare and specific usage of CREATE2?

siladu (commented):
Another pass based on the new changes. I prefer this messageFrame approach 👍

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
shemnon: Settled on `beneficiary` and `originator`

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
shemnon: Spec has `originator`, so I went with that.

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
shemnon: Worse than that, there was a bug relating to repeated self-destructs of the same address.  I rewrote the logic into something much simpler.

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -27,54 +27,77 @@
...
+  public SelfDestructOperation(final GasCalculator gasCalculator, final boolean eip6780Semantics) {
     super(0xFF, "SELFDESTRUCT", 1, 0, gasCalculator);
+    this.eip6780Semantics = eip6780Semantics;
   }
 
   @Override
   public OperationResult execute(final MessageFrame frame, final EVM evm) {
-    final Address recipientAddress = Words.toAddress(frame.popStackItem());
-
+    // fist calculate cost.  There's a bit of yak shaving getting values to calculate the cost
```
siladu, line 56:
```suggestion
    // first calculate cost.  There's a bit of yak shaving getting values to calculate the cost
```

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -27,54 +27,77 @@
...
+    final Address originatorAddress = frame.getRecipientAddress();
+    final Wei originatorBalance = frame.getWorldUpdater().get(originatorAddress).getBalance();
 
     final long cost =
-        gasCalculator().selfDestructOperationGasCost(recipientNullable, inheritance)
-            + (accountIsWarm ? 0L : gasCalculator().getColdAccountAccessCost());
+        gasCalculator().selfDestructOperationGasCost(beneficiaryNullable, originatorBalance)
+            + (beneficiaryIsWarm ? 0L : gasCalculator().getColdAccountAccessCost());
 
+    // with the cost we can test for two early exits: static or not enough gas
```
siladu, line 70: good comment 👍

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -27,54 +27,77 @@
...
-    final MutableAccount recipient =
-        frame.getWorldUpdater().getOrCreate(recipientAddress).getMutable();
+    // Do the "sweep," all modes send all originator balance to the beneficiary account.
+    originatorAccount.decrementBalance(originatorBalance);
+    beneficiaryAccount.incrementBalance(originatorBalance);
 
-    if (!account.getAddress().equals(recipient.getAddress())) {
-      recipient.incrementBalance(account.getBalance());
+    // If we are actually destroying the contract (pre-Cancun or same-tx-create) we need to
+    // explicitly zero out the account balance (destroying ether if the originator is the
```
siladu, line 88: typo?

```suggestion
    // explicitly zero out the account balance (destroying even if the originator is the
```

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -27,54 +27,77 @@
...
 
-    final MutableAccount recipient =
-        frame.getWorldUpdater().getOrCreate(recipientAddress).getMutable();
+    // Do the "sweep," all modes send all originator balance to the beneficiary account.
+    originatorAccount.decrementBalance(originatorBalance);
+    beneficiaryAccount.incrementBalance(originatorBalance);
 
-    if (!account.getAddress().equals(recipient.getAddress())) {
-      recipient.incrementBalance(account.getBalance());
+    // If we are actually destroying the contract (pre-Cancun or same-tx-create) we need to
```
siladu, line 87: for consistency with the rest of the comment/variable names

```suggestion
    // If we are actually destroying the originator account (pre-Cancun or same-tx-create) we need to
```

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -27,54 +27,77 @@
...
+    final MutableAccount originatorAccount =
+        frame.getWorldUpdater().getAccount(originatorAddress).getMutable();
+    final MutableAccount beneficiaryAccount =
+        frame.getWorldUpdater().getOrCreate(beneficiaryAddress).getMutable();
 
-    final MutableAccount recipient =
-        frame.getWorldUpdater().getOrCreate(recipientAddress).getMutable();
+    // Do the "sweep," all modes send all originator balance to the beneficiary account.
+    originatorAccount.decrementBalance(originatorBalance);
+    beneficiaryAccount.incrementBalance(originatorBalance);
```
siladu, line 85: Like that this is closer to the execution-spec 👍

siladu (commented):
Readability changes look good. 

> Spec has originator, so I went with that.

I actually wonder if the execution-spec has the wrong name in this case. Yellowpaper defines "sender" and "original transactor", saying that original transactor (originator) can change when using CREATE. 
But I think [even the YP has a mistake in its definition of it](https://github.com/ethereum/yellowpaper/blob/beacfbd29ccc04dfd790f2282d4c4b1c0afc9b5a/Paper.tex#L794), when compared to [the ORIGIN opcode](https://ethereum.github.io/execution-specs/autoapi/ethereum/shanghai/vm/instructions/environment/index.html?highlight=origin#origin). 

I think it's the _sender_ that should change during CREATE. Indeed, [this is how Besu appears to implements it](https://github.com/hyperledger/besu/blob/47a61ddb1185e400520e0ae9fa9cfd0b577d4fc1/evm/src/main/java/org/hyperledger/besu/evm/operation/AbstractCreateOperation.java#L156-L160). 
To me, it makes sense that the transaction originator never changes and the sender can change if CREATE is involved. 

I _think_ that means for self-destruct we should be dealing with the _sender_ rather than the originator.

shemnon:
Both Geth's dev-branch and ours agree on the testing results (not all execution-spec-tests appear to be correct, and geth/besu disagree with the spec tests in the same way).

So I'm considering it up to standard finally.

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
siladu: I read it as "either" rather than "ether" :)

PR #5436, opened 2023-05-05
Send rawtx conditional

macfarla:
Related to account abstraction. 

Implementation of this endpoint https://notes.ethereum.org/@yoav/SkaX2lS9j
* basically the same as eth_sendRawTransaction, with the ability to specify a number of conditions
- [x] have put X in the RPC name for now - spec may not be finalized
- [x] limit max number of user-specified constraints
- [x]  check expected storageHash against blockchainQueries actual value (needs a new method to expose this)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractEthSendRawTransaction.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright ConsenSys AG.
```
fab-10, line 2: Besu contributors

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthSendRawTransactionConditional.java
```diff
@@ -0,0 +1,166 @@
...
+      final SendRawTransactionConditionalParameter conditions =
+          conditionalParameter(requestContext);
+      if (!withinRange(
+          blockchainQueries.get().headBlockNumber(),
+          conditions.getBlockNumberMin(),
+          conditions.getBlockNumberMax())) {
+        return getJsonRpcErrorResponse(
+            requestContext,
+            USER_SPECIFIED_CONDITIONS_NOT_MET,
+            "block number not within specified range");
```
fab-10, line 88: is it costly to add details about the block numbers here?

this apply also to the other error descriptions below

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/SendRawTransactionConditionalParameter.java
```diff
@@ -0,0 +1,190 @@
...
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.units.bigints.UInt256;
+
+public class SendRawTransactionConditionalParameter {
+
+  private final Optional<Long> fromBlock;
+  private final Optional<Long> toBlock;
+  // TODO may need a custom deserializer because the hash here could actually be a map
+  private final Optional<Map<Address, KnownAccountInfo>> knownAccounts;
+  private final Optional<Long> timestampMin, timestampMax;
```
fab-10, line 48: nit: different styles used to declare vars of the same type

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -200,6 +201,20 @@ public Optional<UInt256> storageAt(
...
 
+  /**
+   * Returns the storage root hash for the given address.
+   *
+   * @param address The address of the account that owns the storage being queried.
+   * @param blockNumber The blockNumber that is being queried.
+   * @return The storage root hash for the account being queried.
+   */
+  public Optional<Hash> storageRoot(final Address address, final long blockNumber) {
+    final Hash blockHash =
```
fab-10, line 212: why in case the block is not found we do not just return empty, instead of searching for the `Hash.EMPTY` block? What does a search with block `Hash.EMPTY` return?

fab-10 (commented):
Overall looks good, just minor things and some questions.

One thing is still not clear to me about the lifecycle of a tx sent by this method, should the conditional be evaluated also after the tx has been accepted?
For example given a block range condition, if it is satisfied when accepting the tx, but not anymore satisfied when the tx is selected to be included in a block, should the tx be included or discarded?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthSendRawTransactionConditional.java
macfarla: added

macfarla:
> Overall looks good, just minor things and some questions.
> 
> One thing is still not clear to me about the lifecycle of a tx sent by this method, should the conditional be evaluated also after the tx has been accepted? For example given a block range condition, if it is satisfied when accepting the tx, but not anymore satisfied when the tx is selected to be included in a block, should the tx be included or discarded?

changed to draft while we're waiting on clarification on this question

non-fungible-nelson:
Suggest we ask in the ERC implementers Telegram the question Fabio posed. DM me for details

non-fungible-nelson:
Adding context from Yoav:

It must be evaluated against the current state at the time of inclusion in a block.

The purpose of this RPC is to create a transaction that will only be included if the conditions are met at the time it is included.  It's a cheap alternative to the otherwise-expensive alternative of running the transaction and ensuring non-reversion (like the Flashbots relay does).  The sender does the work of mapping the minimum conditions for the transaction to succeed, so the validator/sequencer doesn't have to run the entire transaction.  But if the conditions are no longer met, the transaction might revert despite the sender's best effort.  That's where the above guarantee fills the gap by dropping the transaction.

macfarla:
think it makes more sense to do this in a plugin

non-fungible-nelson:
If we want to support account abstraction on many chains, we may want to ship this in vanilla Besu, so its picked up in any upstream changes. What do you think?

PR #5438, opened 2023-05-05
Enable blobs on Blockchain and trie log column families

ahamlat:
## PR description
Enable blobs on Blockchain and trie log column families.
Only values that are bigger than 4 KiB are written to Blobs


ahamlat:
Closing this PR as another PR #5475 with good results is already merged.

PR #5444, opened 2023-05-09
Update dependencies - commons-net

### LINKED ISSUE #5386: Address Vulnerabilities in graphql-java, commons_net, antler
### Description
As a Besu user, I want currently critical and high vulnerabilities reported in Besu addressed so that any solution based on Besu can maintain minimum security baselines. 

| Package | Vulnerability |  Severity | Current Version | Fixed Version |
| -------- | ------------- | --------- | --------------- | ------------- | 
| graphql-java | CVE-2023-28867 | High  |  19.2 | ~> 19.4 |
| commons_net |  -  | Medium | 3.8.0  | ~> 3.9.0 | 
| antlr  | - | Medium | 4.10.1 | ~> 4.11.0 |

### Acceptance Criteria
* All the three Vulnerabilities above are applied required fixes 

### Steps to Reproduce (Bug)
1. Vulnerability Scanning against common vulnerability DBs

### Logs (if a bug)
Please post relevant logs from Besu (and the consensus client, if running proof of stake) from before and after the issue. 

* Software version: 23.1.3

macfarla:
Refs #5386 
This PR forces the commons-net version to be 3.9.0 but we can't do the same for antlr. antlr version needs to match the version in tuweni. So we need a tuweni release before we can update antlr.
This PR has the version we want but it's not yet released https://github.com/apache/incubator-tuweni/pull/479


gradle/versions.gradle
```diff
@@ -147,6 +147,8 @@ dependencyManagement {
       entry 'tuweni-units'
     }
 
+    dependency 'commons-net:commons-net:3.9.0'
```
siladu, line 150: Does it have parent dependency? Would it be good to add a comment to upgrade that in the future and remove this.

siladu (approved):
Maybe some extra stuff in verification-metadata but don't think it really matters?

gradle/versions.gradle
macfarla: done

Issue #4043, opened 2022-07-02
Besu fails to create db if `--data-path` is a symlink and there is no existing db


SeaMonkey82:
```
➜  eth mkdir /mnt/eth3/besu-symlink-test
➜  eth ln -s /mnt/eth3/besu-symlink-test besu-symlink
➜  eth ~/Downloads/merge-testnets/ropsten-beacon-chain/besu/build/install/besu/bin/besu --data-path=/home/seamonkey/eth/besu-symlink
2022-07-01 20:09:51.676-04:00 | main | INFO  | Besu | Using LibEthPairings native alt bn128
2022-07-01 20:09:51.677-04:00 | main | INFO  | Besu | Using the native implementation of the signature algorithm
2022-07-01 20:09:51.680-04:00 | main | INFO  | Besu | Starting Besu version: besu/v22.7.RC1-dev-043191a4/linux-x86_64/openjdk-java-11
2022-07-01 20:09:51.962-04:00 | main | INFO  | Besu | Engine API authentication enabled without key file. Expect ephemeral jwt.hex file in datadir
2022-07-01 20:09:51.964-04:00 | main | INFO  | Besu | Static Nodes file = /home/seamonkey/eth/besu-symlink/static-nodes.json
2022-07-01 20:09:51.965-04:00 | main | INFO  | StaticNodesParser | StaticNodes file /home/seamonkey/eth/besu-symlink/static-nodes.json does not exist, no static connections will be created.
2022-07-01 20:09:51.965-04:00 | main | INFO  | Besu | Connecting to 0 static nodes.
2022-07-01 20:09:51.967-04:00 | main | INFO  | Besu | Security Module: localfile
2022-07-01 20:09:51.976-04:00 | main | INFO  | RocksDBKeyValueStorageFactory | No existing database detected at /home/seamonkey/eth/besu-symlink. Using version 1
Failed to retrieve the RocksDB database meta version: /home/seamonkey/eth/besu-symlink

To display full help:
besu [COMMAND] --help
➜  eth 
```

daniellehrner:
Thank you for reporting the issue. We will look into it.

matthew1001:
I thought I'd take a look a this and it looks like Java document the behaviour of `Files.createDirectories()` as throwing `FileAlreadyExistsException` if the directory exists but is not a directory: https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createDirectories

[...]

PR #5453, opened 2023-05-15
Only create DB data dir if it doesn't already exist

matthew1001:
## PR description
This PR skips creation of the database directory if it exists. While the call to `Files.createDirectories()` (which we use to initialise it if the DB needs creating) should be a no-op in this case, symlinks cause it to fail so the safer thing to do is confirm that the directory doesn't actually exist before making that call.

## Fixed Issue(s)
Fixed https://github.com/hyperledger/besu/issues/4043


jflo:
Looks good, but could use a unit test to cover it.

matthew1001:
> Looks good, but could use a unit test to cover it.

Good point, I'll add something today and push a new commit.

jflo (approved):
Thanks for the contribution!

PR #5465, opened 2023-05-16
handle cases where dagger besuComponent is null

garyschulte:
## PR description
fix cases where BesuCommand is constructed without a dagger context


gfukushima (approved):
lgtm

PR #5467, opened 2023-05-17
JsonRpcHttpService - Add BodyHandler limit size 

Gabriel-Trintinalia:
## PR description
Body Size Limit for BodyHandler in Vert.x Route: A limit on body size for the BodyHandler is introduced to prevent memory overuse and improve application robustness. (Set to 5MB)

`--rpc-http-max-request-content-length ` overrides the limit.


macfarla (approved):
LGTM

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpService.java
```diff
@@ -320,6 +321,7 @@ private Router buildRouter() {
         .route()
         .handler(
             BodyHandler.create()
+                .setBodyLimit(BODY_LIMIT)
```
jframe, line 324: If this limit is breached, what error is returned? And do we translate that to a JsonRpcError?

Gabriel-Trintinalia: > If this limit is breached, what error is returned? And do we translate that to a JsonRpcError?

HTTP response status 413 - `Request Entity Too Large`. It does not translate to a JsonRpcError since it is more like a client error.

macfarla (approved):

jframe (approved):
LGTM

PR #5473, opened 2023-05-19
Invalid transaction type - report to user

macfarla:
If transaction validation fails because of "Invalid transaction type", report that error back to the user (previously was "invalid params")
This will make situations like this easier to troubleshoot -
https://discord.com/channels/905194001349627914/938504958909747250/1108931495797727354


Gabriel-Trintinalia (approved):
LGTM

PR #5474, opened 2023-05-19
remove old EF bootnodes

gfukushima:
## PR description
This PR removes old EF bootnodes .Would like someone from the EF team validate this before we move forward with that.
@skylenet @parithosh Could please confirm these bootnodes don't exist anymore?


parithosh:
Yup, I can confirm that the bootnodes do not exist and the list after this update will match our current live bootnode stack

siladu (approved):
LGTM, reckon this will have a positive impact on peering @gfukushima ?

gfukushima:
Don't think it will be significant, I'd say this is more a clean up to the codebase

PR #5471, opened 2023-05-18
Introduce variables storage

fab-10:
## PR description

This is the first PR of a series to reorganize the data in db, in order to pave the way to future optimizations.
The first goal that we want to achieve is the optimization of the db for data that is static, leveraging [the BlobDB](https://github.com/facebook/rocksdb/wiki/BlobDB) feature.
For static data like blockchain or trie logs, BlobDB is a better choice since it reduces a lot the write amplification during compaction, with the result of faster initial sync and less wear out of the disk.

Now we have variable data mixed with static data in the blockchain column family, and this is not optimal, so this PR moved the _variables_ into the dedicated column family `VARIABLES` that we can further optimize for frequently changing data in future.

The PR is meant automatically migrates existing variables to the new location on startup. 

Logs during migration
```
{"@timestamp":"2023-06-06T10:14:15,156","level":"INFO","thread":"main","class":"KeyValueStoragePrefixedKeyBlockchainStorage","message":"Migrated key chainHeadHash to variables storage","throwable":""}
{"@timestamp":"2023-06-06T10:14:15,157","level":"INFO","thread":"main","class":"KeyValueStoragePrefixedKeyBlockchainStorage","message":"Migrated key finalizedBlockHash to variables storage","throwable":""}
{"@timestamp":"2023-06-06T10:14:15,158","level":"INFO","thread":"main","class":"KeyValueStoragePrefixedKeyBlockchainStorage","message":"Migrated key safeBlockHash to variables storage","throwable":""}
{"@timestamp":"2023-06-06T10:14:15,162","level":"INFO","thread":"main","class":"KeyValueStoragePrefixedKeyBlockchainStorage","message":"Migrated key forkHeads to variables storage","throwable":""}
{"@timestamp":"2023-06-06T10:14:15,163","level":"INFO","thread":"main","class":"KeyValueStoragePrefixedKeyBlockchainStorage","message":"Migrated key local-enr-seqno to variables storage","throwable":""}
```

It is possible to revert this change in case the user want to run a previous version of Besu, before downgrading the Beus, the user must run the subcommand `storage revert-variables` with the same configuration use to normally start Besu, and only after that it is possible to install a previous version of Besu.

Logs during revert
```
2023-06-06 10:15:33.201+00:00 | main | INFO  | DatabaseMetadata | Lookup database metadata file in data directory: /data/besu
[...]

PR #5475, opened 2023-05-19
Blobdb for static data

### LINKED ISSUE #4607: Implement and benchmark rocksdb BlobDB for blockchain storage
### Description
As the blockchain storage is append only and block sizes tend to be > 10kb, besu should be able to benefit from RocksDB's [Integrated BlobDB](http://rocksdb.org/blog/2021/05/26/integrated-blob-db.html).  Compaction should be cheaper as the blocks themselves would not be subject tocompaction.

### Acceptance Criteria
* implement blobdb for blockchain storage
* optionally implement blobdb for all values in excess of some fixed size
* benchmark block import via snap-sync with a variety of minimum blob sizes to determine the performance impact
* evaluate the impact of blobdb on overall iops during snap-sync
* evaluate memory consumption for both on and off-heap

fab-10:
## PR description

This PR is built on top of https://github.com/hyperledger/besu/pull/5471, so check it first.
[To only show the diff against #5471](https://github.com/fab-10/besu/compare/variables-storage...fab-10:besu:blobdb-for-static-data) 

This PR introduces optimizations for RocksDB column families that contain static data, i.e. key value entries that are never changed, but only added, like an append log storage, the best example is the blockchain storage, that benefit if [BlobDB ](https://github.com/facebook/rocksdb/wiki/BlobDB) is used for them, since the write amplification is highly reduced, with the result of faster initial sync and less wear out of SSD.

Enabling BlobDB for a column family does not make the db incompatible with a previous version of Besu, however before downgrading Besu you need to run the subcommand `storage revert-variables` to revert changes done in PR #5471 

Preliminary test shows that checkpoint sync is faster with BlobDB enabled, more tests are ongoing, and will share the results later.

## Fixed Issue(s)

fixes #4607


jflo (commented):
So far so good. Have you considered re-using the DataStoreModule from evmtool? We might be able to pull that up into Besu, and use that to provide VariablesStorage to anything that needs it, via BesuComponent.

fab-10:
> So far so good. Have you considered re-using the DataStoreModule from evmtool? We might be able to pull that up into Besu, and use that to provide VariablesStorage to anything that needs it, via BesuComponent.

Not analyzed that part, and given my little knowledge of that part, I will check with you what is necessary for that.

garyschulte:
the diff from 5471 looks good to me.  I am not sure why the failing unit and acceptance tests.

fab-10:
> the diff from 5471 looks good to me. I am not sure why the failing unit and acceptance tests.

unnecessary stubbing, fixed

ahamlat:
Adding some metrics on this PR
### Sync and block block processing timings


**Sync Time**
image

We can also see the difference in block import time, especially from some block number where the node running this PR has a much better block import time.

image

image

**Block processing (post sync)**
Block processing time is very close between both nodes.

image

We can notice also the same performance regarding Forkchoice update (FCU) 

image

### System metrics

Block import time is better because with BlobDB there is [less compaction and thus less write amplification](https://github.com/facebook/rocksdb/wiki/BlobDB).

**Without PR #5475**
image
We can notice below a huge IO activity with a lot of IO waits, 28 ms on average on writes.
image


**With PR #5475**
image
With this PR, there is less IO activity with less IO waits, 9 ms on average on writes.
image

garyschulte (approved):
🚢

PR #5479, opened 2023-05-19
Remove bonsai accumulator trielogfactory

garyschulte:
## PR description
Fixes a refactor omission.  TrieLogFactory was still statically constructed and used in accumulator.  This would lead to besu writing trie logs to its database with the default TrieLog serializer rather than one (potentially) configured via plugin.


matkt (approved):
LGTM

PR #5491, opened 2023-05-24
Ignore min-block-occupancy-ratio option when on PoS

fab-10:
## PR description

[`min-block-occupancy-ratio` option ](https://besu.hyperledger.org/en/stable/public-networks/reference/cli/options/#min-block-occupancy-ratio) make sense on PoW networks where you need to be faster that other nodes in finding a block, so you can set be satisfied when the block produced is enough filled, but on PoS you do not need to compete with other nodes, since you have an allocated slot to produce the block, so in this case make sense to always try to fill the block, until the remaining gas is less than the minimum needed for the smaller transaction.
So for PoS the `min-block-occupancy-ratio` option is ignored since we always try to fill 100% of the block.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Breaking Changes
 - Add request content length limit for the JSON-RPC API (5MB) [#5467](https://github.com/hyperledger/besu/pull/5467)
+- `min-block-occupancy-ratio` options is now ignored on PoS networks [#5491](https://github.com/hyperledger/besu/pull/5491)
```
macfarla, line 7: is this a breaking change?

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
```diff
@@ -405,14 +438,119 @@ public void transactionSelectionStopsWhenSufficientBlockOccupancyIsReached() {
...
+            miningBeneficiary,
+            Wei.ZERO,
+            MIN_OCCUPANCY_100_PERCENT);
+
+    final long minTxGasCost = getGasCalculator().getMinimumTransactionCost();
+
+    // Add 4 transactions to the Pending Transactions
+    // 1) 90% of block (selected)
+    // 2) 90% of block (skipped since too large)
+    // 3) dot not fill the block, but leaves less gas than the min for a tx (selected)
```
macfarla, line 512:
```suggestion
    // 3) do not fill the block, but leaves less gas than the min for a tx (selected)
```

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
```diff
@@ -405,14 +438,119 @@ public void transactionSelectionStopsWhenSufficientBlockOccupancyIsReached() {
...
+            transactionProcessor,
+            blockHeader,
+            Wei.ZERO,
+            miningBeneficiary,
+            Wei.ZERO,
+            MIN_OCCUPANCY_100_PERCENT);
+
+    final long minTxGasCost = getGasCalculator().getMinimumTransactionCost();
+
+    // Add 4 transactions to the Pending Transactions
```
macfarla, line 457: nit - if you numbered these 0..4 it would match the code

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
```diff
@@ -405,14 +438,119 @@ public void transactionSelectionStopsWhenSufficientBlockOccupancyIsReached() {
...
+    // 1) 90% of block (selected)
```
macfarla, line 510: nit - if you numbered these 0..4 it would match the code

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
```diff
@@ -467,15 +470,34 @@ private boolean transactionTooLargeForBlock(final Transaction transaction) {
...
-    final double gasUsed = transactionSelectionResults.getCumulativeGasUsed();
-    final double occupancyRatio = gasUsed / gasAvailable;
+    final long gasAvailable = processableBlockHeader.getGasLimit();
+    final long gasUsed = transactionSelectionResults.getCumulativeGasUsed();
+    final long gasRemaining = gasAvailable - gasUsed;
+
+    final double occupancyRatio = (double) gasUsed / (double) gasAvailable;
     LOG.trace(
-        "Min block occupancy ratio {}, gas used {}, available {}, used/available {}",
+        "Min block occupancy ratio {}, gas used {}, available {}, remaining {}, used/available {}",
```
macfarla, line 479: is it confusing to log the min block occupancy ratio if we're ignoring it?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
```diff
@@ -467,15 +470,34 @@ private boolean transactionTooLargeForBlock(final Transaction transaction) {
...
   }
+
+  private boolean blockFull() {
+    final long gasAvailable = processableBlockHeader.getGasLimit();
+    final long gasUsed = transactionSelectionResults.getCumulativeGasUsed();
+    final long gasRemaining = gasAvailable - gasUsed;
+
+    if (gasRemaining < gasCalculator.getMinimumTransactionCost()) {
+      LOG.trace(
+          "Block full, remaining gas {} is less that minimum transaction gas cost {}",
```
macfarla, line 496:
```suggestion
          "Block full, remaining gas {} is less than minimum transaction gas cost {}",
```

macfarla (approved):

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
gfukushima: I think we're logging the current occupancyRatio here, not the min

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeBlockCreator.java
```diff
@@ -71,7 +69,7 @@ public MergeBlockCreator(
         protocolContext,
         protocolSchedule,
         minTransactionGasPrice,
-        minBlockOccupancyRatio,
+        1.0,
```
gfukushima, line 72: Might be better if we use a constant here like we did in the test

gfukushima (approved):
lgtm

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
gfukushima: I think we're logging the current occupancyRatio here, not the min

CHANGELOG.md
fab-10: yes, but I think no one is really using this on a PoS network, because it makes no sense to force smaller blocks there.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
fab-10: technically we are not ignoring it, but for PoS we are setting it to be always 100%

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
fab-10: good catch, will renumber

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeBlockCreator.java
fab-10: done, and also added a javadoc to explain why

PR #5493, opened 2023-05-24
Add MCOPY Operation (EIP-5656)

shemnon:
Add the MCOPY operation targeting cancun.
Testing required adding memory to the text fixture.


jflo (approved):
Pretty straightforward, looks good.

evm/src/main/java/org/hyperledger/besu/evm/operation/MCopyOperation.java
```diff
@@ -0,0 +1,56 @@
...
+/** The Memory copy operation. */
+public class MCopyOperation extends AbstractOperation {
+
+  /**
+   * Instantiates a new Memory Copy operation.
+   *
+   * @param gasCalculator the gas calculator
+   */
+  public MCopyOperation(final GasCalculator gasCalculator) {
+    super(0x5e, "MCOPY", 3, 0, gasCalculator);
```
garyschulte, line 35: EIP spec is different obviously, but it looks like there has been no pushback on https://hackmd.io/@shemnon/CancunOpcodeAllocation ?  It could be worth a TODO just in case

shemnon: That's the reason I'm holding off on a commit.  (a) mcopy is not commited yet and (b) opcode renumbering is also pending.

shemnon:
@jflo, @garyschulte, can one of you put a "changes requested" review on this so it doesn't accidentally get merged?  After this mornings ACD it's not the slam dunk it once was.

garyschulte (changes requested):
Placeholder 'changes requested' until broader community consensus on opcodemrestructuring

evm/src/main/java/org/hyperledger/besu/evm/operation/MCopyOperation.java
shemnon: Opcode number settled on 0x5e - https://github.com/ethereum/EIPs/pull/7155

PR #5498, opened 2023-05-25
Stop evaluating txs for sender after the first skipped one

fab-10:
## PR description

This PR is built on top of  #5492 and #5491 so check them first.

This PR introduce a simple optimization to avoid to waste time validating transactions that will not be selected during the current block creation process. 
Basically if a sender have more than one prioritized candidate transactions, and for some reason any of them is not selected during the block creation process, it make no sense to keep evaluating any more transaction after the skipped one, since it will not be selected due to the nonce gap.

Enabling the selection stats implemented in #5492 is it possible to see the difference before and after this PR, the wasted validation work that we are removing is represented by the `INVALID_TRANSIENT(NONCE_TOO_HIGH)` label.

Before
```
Selection stats: Totals[Evaluated=2000, Selected=109, Skipped=1891, Dropped=0]; Detailed[INVALID_TRANSIENT(GAS_PRICE_BELOW_CURRENT_BASE_FEE)=1712, INVALID_TRANSIENT(NONCE_TOO_HIGH)=158, INVALID_TRANSIENT(TX_TOO_LARGE_FOR_REMAINING_GAS)=21, SELECTED=109]
```

After
```
Selection stats: Totals[Evaluated=1743, Selected=126, Skipped=1617, Dropped=0]; Detailed[INVALID_TRANSIENT(GAS_PRICE_BELOW_CURRENT_BASE_FEE)=1600, INVALID_TRANSIENT(TX_TOO_LARGE_FOR_REMAINING_GAS)=17, SELECTED=126]
```


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/LayeredPendingTransactions.java
```diff
@@ -321,64 +321,69 @@ public synchronized void selectTransactions(
...
+
+                          if (res.stop()) {
+                            completed.set(true);
+                          }
+
+                          if (!res.selected()) {
+                            // avoid processing other txs from this sender if this one is skipped
+                            // since the following will not be selected due to the nonce gap
+                            skipSenders.add(candidatePendingTx.getSender());
+                            LOG.trace("Skip sender {}", candidatePendingTx.getSender());
```
macfarla, line 367:
```suggestion
                            LOG.trace("Skipping tx from sender {}", candidatePendingTx.getSender());
```

macfarla: how does the sender get removed from the skipSenders list?

fab-10: He should not be removed for the remaining of this selection task, since evaluating his txs is wasted time after the first non selected. This list only last for a single block creation, so he could be picked again in a following selection task.

macfarla (approved):
LGTM

PR #5396, opened 2023-04-27
Add plugin API to select Transactions

pinges:
## PR description
This PR adds an API that can be used to select Transactions while they are added to a block

PR #5499, opened 2023-05-25
Revert "Add plugin API to select Transactions (#5396)"

siladu:
This reverts commit 65bcc557e41cca2568588a21441aa5d8658a5e79.

Due to breaking Clique (in the ATs at least)


.circleci/config.yml
```diff
@@ -401,6 +401,9 @@ workflows:
       - acceptanceTests:
           requires:
             - assemble
+      - acceptanceTestsNonMainnet:
+          requires:
+            - assemble
```
macfarla, line 406: are we putting this back in

macfarla (approved):

macfarla:
cc @pinges

.circleci/config.yml
siladu: no, was just a temporary test to prove it passes before merging, reverting now

PR #5500, opened 2023-05-25
Specify optional scheme (ws or wss) in Ethstats url option

### LINKED ISSUE #5494: Optionally specify ws:// or wss:// to Ethstats connection url
### Description
Modify [ethstats option](https://besu.hyperledger.org/en/stable/public-networks/reference/cli/options/?h=ethstats#ethstats) so that it can optionally take ws or wss to specify if it wants to use ssl or plain websocket connection. This will override the current behavior of switching between ssl and non-ssl between retry connections.

### Acceptance Criteria
* ethstats URI can take ws:// and wss:// as scheme. ws:// will not use SSL. wss:// will use SSL to establish websocket connection.
* If protocol is not specified in URI, keep the current logic i.e. try to connect to ethstats server using secure websocket, then using plain websocket (with 10 seconds interval).
* If port is not specified, ws:// based URI will default to port 80 and wss:// will default to port 443. (Note: If ethstats server is not running behind a proxy, it usually uses port 3000. In such case, the Besu ethstats URI should mention the port in the URI as we do not default to port 3000 anymore.
* Update relevant unit/AT.

usmansaleem:
## PR description
- Allow scheme (ws:// or wss://) optionally to be specified in ethstats url. If user specify the scheme, the connection logic doesn't need to switch from ssl to non-ssl on each retry logic.
- Change order of ethstats service start after main ethereum loop is up because ethstats service needs local node address.

## Fixed Issue(s)

Fixes #5494


ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/EthStatsConnectOptions.java
```diff
@@ -48,22 +46,48 @@ static EthStatsConnectOptions fromParams(
...
+      } else {
+        // append ws:// to make a valid URI while keeping scheme as null
+        uri = URI.create("ws://" + url);
+        scheme = null;
+      }
+
+      if (scheme != null) {
+        // make sure that scheme is either ws or wss
+        if (!scheme.equalsIgnoreCase("ws") && !scheme.equalsIgnoreCase("wss")) {
+          throw new IllegalArgumentException("EthStats URI only support ws:// or wss:// scheme.");
```
macfarla, line 65:
```suggestion
          throw new IllegalArgumentException("Ethstats URI only support ws:// or wss:// scheme.");
```

macfarla (approved):

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/EthStatsConnectOptions.java
```diff
@@ -48,22 +46,48 @@ static EthStatsConnectOptions fromParams(
...
-            .build();
+      // if scheme is not specified in the URI, user info (nodename) gets converted to scheme.
+      final URI uri;
+      final String scheme;
+      if (url.matches("^.*://.*$")) {
+        // construct URI
+        uri = URI.create(url);
+        scheme = uri.getScheme();
+      } else {
+        // append ws:// to make a valid URI while keeping scheme as null
```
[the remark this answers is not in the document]
usmansaleem: done

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/util/EthStatsConnectOptions.java
usmansaleem: done

PR #5502, opened 2023-05-26
run non-mainnet ATs on PRs, split into sub groups

macfarla:
Run all ATs on PRs, but split into 
* mainnet
* Clique + BFT
* Privacy + Permissioning


siladu (approved):
LGTM - let's try it out

Issue #5415, opened 2023-05-02
Blockchain Sync Improvements: Chain Data & Pipeline 


non-fungible-nelson:
## Description

Besu needs to be able to sync the chain data as quickly as possible. It also needs to be more aware of the state of the client (i.e. is the worldstate downloaded? Can I begin validator duties?) and differ or parallelize the chain download to improve the sync UX for Besu stakers and RPC providers. 

### Goals
- Faster time from start to validating by differing and/or improving chain download
- Refactor chain pipeline for post-Merge environment

### Acceptance Criteria:
* New pipeline strategy for chain-data
  * Download blockchain in the background (?) Use 8 seconds of free slot time for background download
  * Better post-Merge sync target based on the CL blocks and what we receive from the CL
  * If we have deposit logs (Teku), immediately start validating when worldstate download is complete 
   
### Considerations
* Consider implications for private networks (leave fast unchanged, update snap)
  * auto-detect lock-step sync with CL, switch to optimistic head rather than relying on fcU
* Peering for improving sync performance
  * When receiving an empty response, retry logic polls new peers #5271
  * Flag peer as useless/demoted

non-fungible-nelson:
#5373 ref 

Need to re-evaluate post-Merge sync strategy. Q3 perhaps...

non-fungible-nelson:
#5535 has more context

Issue #5271, opened 2023-03-28
Make all tasks switching tasks


pinges:
### Description
At the moment we have "switching" tasks, that go through the list of our peers when it is necessary to retry a task, and we have "retrying" tasks, that select the "best" peer every time we need to retry that task. As the best peer was selected the first time this task was called, there is a very good chance that we are selecting the same peer to retry that task. It makes more sense to select a different peer, which the "switching" tasks do.

There are tasks that we want to call with a specific peer, like getting a block or transaction when it was announced by one of our peers. We need to make sure that this can still happen.

### Acceptance Criteria
* There are no 'retrying' tasks left

fab-10:
this is partially implemented in https://github.com/hyperledger/besu/pull/5508 and https://github.com/hyperledger/besu/pull/5509

non-fungible-nelson:
Need a status on this ticket

Matilda-Clerke:
This might be a bit redundant after my refactoring work is in place. See https://www.notion.so/consensys/EthTask-and-EthPeer-system-ba1f00f85f5e4b9c94a0dd9c661f2c5c?d=d8303d044a8d40e590b48ea29885e154#567a120382c544a9b6e9431a3c951926 and https://github.com/hyperledger/besu/pull/7602

non-fungible-nelson:
Closing in favor of #7602

PR #5508, opened 2023-05-29
Use retry switching peer for world state download tasks

fab-10:
## PR description

Built on top of #5509, so please check it first. [Link to only see diff again #5509](https://github.com/fab-10/besu/compare/demote-peer-with-empty-responses-when-retrying...fab-10:besu:use-retry-switching-peer)

With this PR all the world state download tasks used by snap sync are updated to use the retry switching peer strategy, to reduce the chance that we keep sending requests to the same bad peer.

relates to https://github.com/hyperledger/besu/issues/5415 and #5271


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/RetryingGetStorageRangeFromPeerTask.java
```diff
@@ -44,8 +43,13 @@ private RetryingGetStorageRangeFromPeerTask(
...
       final Bytes32 endKeyHash,
       final BlockHeader blockHeader,
-      final MetricsSystem metricsSystem) {
-    super(ethContext, 4, data -> data.proofs().isEmpty() && data.slots().isEmpty(), metricsSystem);
+      final MetricsSystem metricsSystem,
+      final int maxRetries) {
+    super(
+        ethContext,
+        metricsSystem,
+        data -> data.proofs().isEmpty() && data.slots().isEmpty(),
```
matkt, line 51: A bad peer will be a peer with this kind of result
I’m not sure you are checking if the response is like that in order to switch to another peer

fab-10: check this other PR https://github.com/hyperledger/besu/pull/5509, that checks and records the empty response in the `AbstractRetryingPeerTask`

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/ethtaskutils/RetryingMessageTaskTest.java
```diff
@@ -155,7 +166,8 @@ public void completesWhenPeersAreTemporarilyUnavailable()
     final T requestedData = generateDataToBeRequested();
 
     // Execute task and wait for response
-    final EthTask<T> task = createTask(requestedData);
+    // +2 is required for switching peers tasks where the max retries = number of peers
```
siladu, line 169: useful comment 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -320,7 +320,9 @@ public Stream<EthPeer> streamAvailablePeers() {
   }
 
   public Stream<EthPeer> streamBestPeers() {
-    return streamAvailablePeers().sorted(getBestChainComparator().reversed());
+    return streamAvailablePeers()
+        .filter(EthPeer::isFullyValidated)
```
siladu, line 324: Why is this required?

siladu (commented):
Diff looks pretty straightforward. Couple of small comments. Don't want to approve while the parent PR is in flight

matkt:
Just a question. A peer that is useless for the worldstate part may not be for the blockchain part. Is there not a risk of losing valid nodes for a certain type of data by marking useless when it could still help us.

macfarla:
Blocked by #6609 - without that we would disconnect peers, because we are trying peers that do not serve snap data.

jframe:
@fab-10 Closing this, reopen if think we still need it

PR #5509, opened 2023-05-29
Record empty responses when retrying a peer task

fab-10:
## PR description

`AbstractRetryingPeerTask` has a way to understand if the result of a try is empty, but it only uses this information to discriminate if the result is a partial one, instead is very useful to also use the emptiness information to demote the peer and eventually disconnect it in case it sends too many useless responses, as done in this PR.
In the making of this PR, I discovered that there were opportunities to improve the code and simplify the writing of retrying tasks, so I refactored and documented the code so that any class extending `AbstractRetryingPeerTask` should not set the final task result by themself, but instead implements the `emptyResult` and `successfulResult` to report the status of the request, so that the final setting of the task result is always a duty of `AbstractRetryingPeerTask`, removing the different approaches used before.

relates to #5415 and #5271

## Tests

- [ ] Checkpoint Sync
- [ ] Snap Sync
- [ ] Fast Sync
- [ ] Full sync


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
```diff
@@ -136,7 +125,7 @@ private void refreshPeers() {
     // If we are at max connections, then refresh peers disconnecting one of the failed peers,
     // or the least useful
 
-    if (peers.peerCount() >= peers.getMaxPeers()) {
+    if (peers.peerCount() >= peers.getPeerLowerBound()) {
```
macfarla, line 128: should this be upperBound?

macfarla (commented):
nice refactor. one q on a change from getMaxPeers to getPeerLowerBound

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingPeerTask.java
```diff
@@ -93,11 +94,21 @@ protected void executeTask() {
               if (error != null) {
                 handleTaskError(error);
               } else {
-                // If we get a partial success, reset the retry counter.
-                if (!isEmptyResponse.test(peerResult)) {
-                  retryCount = 0;
+                if (successfulResult(peerResult)) {
+                  result.complete(peerResult);
```
matkt, line 98: I wonder how it worked before without this success part. Is this something that is done automatically and that we don't really need?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
fab-10: I want to force Besu to actively search for new peers, and my understanding is that this happens when the number of peers is below the lower bound

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingPeerTask.java
fab-10: Before each extending task had to set the final result by itself, using `result.complete`, and another thing that was a bit confusing to me, is that the returned value of a task is not the final result until you set `result.complete` and so I tried to centralize this part here and make the writing of extending tasks easier.

matkt: good idea

matkt (commented):
LGTM. Waiting for the test results

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingSwitchingPeerTask.java
macfarla: maybe add a comment (or edit the existing comment) because the code is now doing something slightly different

siladu:
SGTM. Do we demote the peer as soon as we get one empty response from it? If so, are there not cases where we receive an empty result but still want to retry with the same peer?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/RetryingGetStorageRangeFromPeerTask.java
```diff
@@ -72,11 +73,16 @@ protected CompletableFuture<StorageRangeMessage.SlotRangeData> executePeerTask(
...
-            peerResult -> {
-              result.complete(peerResult.getResult());
-              return peerResult.getResult();
-            });
+    return executeSubTask(task::run).thenApply(PeerTaskResult::getResult);
+  }
+
+  @Override
+  protected boolean emptyResult(final StorageRangeMessage.SlotRangeData peerResult) {
+    return peerResult.proofs().isEmpty() && peerResult.slots().isEmpty();
```
garyschulte, line 81: same here, an empty response might be a signal that we are asking for old ranges.

macfarla:
Blocked by #6609 - without that we would disconnect peers, because we are trying peers that do not serve snap data.

Matilda-Clerke:
@fab-10 I haven't included any feedback regarding peer performance or quality in the refactor yet, so it should be possible to ensure we provide feedback for empty responses too. I'll take a good look at this PR to make sure I understand the intended behaviour.

Issue #5466, opened 2023-05-17
Besu help - unite the tx-pool CLI options


macfarla:
to reproduce: `besu -h`
Expected output - all the tx-pool options together
Actual (extract from) output - two groups of tx-pool options
```
      --tx-pool-disable-locals[=<Boolean>]
                             Set to true if transactions sent via RPC should
                               have the same checks and not be prioritized over
                               remote ones (default: false)
      --tx-pool-enable-save-restore[=<Boolean>]
                             Set to true to enable saving the txpool content to
                               file on shutdown and reloading it on startup
                               (default: false)
      --tx-pool-limit-by-account-percentage=<DOUBLE>
                             Maximum portion of the transaction pool which a
                               single account may occupy with future
                               transactions (default: 0.001)
      --tx-pool-save-file=<STRING>
                             If saving the txpool content is enabled, define a
                               custom path for the save file (default: txpool.
                               dump in the data-dir)
  -V, --version              Print version information and exit.
...
Tx Pool Options
      --tx-pool-max-size=<INTEGER>
                             Maximum number of pending transactions that will
                               be kept in the transaction pool (default: 4096)
      --tx-pool-price-bump=<INTEGER>
                             Price bump percentage to replace an already
                               existing transaction  (default: 10)
      --tx-pool-retention-hours=<INTEGER>
                             Maximum retention period of pending transactions
                               in hours (default: 13)
```

gtebrean:
proposed PR merged into main

PR #5512, opened 2023-05-29
refactor txpool options

gtebrean:
## PR description
Add all tx-pool CLI options under a single group https://github.com/hyperledger/besu/issues/5466

## Fixed Issue(s)
Fixes, #5466 

Result
Screenshot 2023-05-30 at 19 40 37


gtebrean:
I moved the 4 tx-pool stable options from BesuCommand in TransactionPoolOptions in order to can have them in single group.
I tried also to add them in a separated specific TransactionPool group but as TransactionPoolOptions class contains unstable options it didn't work in loading them properly from both stable and unstable options.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1197,39 +1188,6 @@ static class PermissionsOptionGroup {
           "Sets target gas limit per block. If set, each block's gas limit will approach this setting over time if the current gas limit is different.")
   private final Long targetGasLimit = null;
 
-  // Tx Pool Option Group
-  @CommandLine.ArgGroup(validate = false, heading = "@|bold Tx Pool Options|@%n")
```
macfarla, line 1201: I'd like to keep this grouping for ux. I get your point about the unstable options - can we keep the unstable options on the TransactionPoolOptions class (they don't appear in the help anyway) and put the remaining tx-pool options within this group in BesuCommand

macfarla (changes requested):

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
gtebrean: updated

macfarla (approved):
LGTM

Issue #5456, opened 2023-05-15
eth_getCode unpredictable behaviour


cbayraktaroglu:
### Description
The response from "eth_getCode"rpc call is unpredictable. Consecutive runs of the same call may result with no code returned. (Potential Bug)

### Steps to Reproduce (Bug)
1. Initialize the client
```
client, err := ethclient.Dial("besu node rpc url")

```
2. Get the address format ready
```
targetAddress := common.HexToAddress("0x3E7d1eAB13ad0104d2750B8863b489D65364e32D")
```

3. Send the request
```
code, err := client.CodeAt(context.Background(), common.HexToAddress("0x3E7d1eAB13ad0104d2750B8863b489D65364e32D"), nil)
```

4. Check the returned bytes length
```
log.Println(len(code))
```

**Expected behavior:**
The length of the returned bytes array should be the same between the runs

**Actual behavior:** 
Non predictable behaviour, sometimes the length is 9571 and sometimes 0

**Frequency:**
1 in 10 runs returns the length of 0

### Additional Information (Add any of the following or anything else that may be relevant)
* Besu setup info - version 23.4.0

non-fungible-nelson:
Hey can you please provide your nodes configuration files? And your CL config if you're on Mainnet/public testnet?

non-fungible-nelson:
Reference - #5038, #5076, #5123 @garyschulte

cbayraktaroglu:
Hi, here are the command line arguments used:

For Besu:
```
        - '--sync-mode=X_SNAP'
        - '--data-storage-format=BONSAI'
        - '--network=mainnet'
        - '--data-path=/var/lib/besu/data'
        - '--metrics-enabled=true'
        - '--metrics-host=0.0.0.0'
        - '--metrics-port=8008'
        - '--engine-rpc-enabled=true'
        - '--engine-jwt-secret=/var/lib/besu/jwtsecret'
        - '--engine-rpc-port=8551'
        - '--engine-host-allowlist=*'
        - '--rpc-http-enabled=true'
        - '--rpc-http-host=0.0.0.0'
        - '--rpc-http-port=8545'
        - >-
          --rpc-http-api=ETH,NET,CLIQUE,DEBUG,MINER,NET,PERM,ADMIN,EEA,TXPOOL,PRIV,WEB3
        - '--rpc-http-cors-origins=*'
        - '--host-whitelist=*'
        - '--graphql-http-enabled'
        - '--rpc-ws-enabled=true'
[...]

PR #5517, opened 2023-05-30
Log worldstate availability

garyschulte:
## PR description

PR adds debug logging to worldstate access path, and a warning when we try to access closed worldstate.
Added in response to #5456, and seems suitable to add to main.


macfarla (approved):
LGTM

PR #5525, opened 2023-06-01
[MINOR] switch QBFT tests to Junit5

macfarla:
statemachine tests have plenty of unnecessary mockings so did not unpick that as part of this PR, went with lenient mockings instead.


fab-10 (approved):

jframe (approved):
LGTM

PR #5536, opened 2023-06-05
EIP - 4844

jflo:
Implements EIP-4844. 

- introduces a Hardfork class to the protocol schedule system
- new Engine APIs required for CL to work on 4844
- new DataGas type for tracking block cost for 4844 data
- new VersionedHash type to reflect that a versioned hash is not quite a pure sha256
- incorporates wrapped jc-kzg library for KZG point evaluations
- New transaction type, and domain objects for constituent parts to represent the Blobs, KZGCommitments, and Proofs used for 4844
- RLP encoders and decoders to support new transaction type
- gas pricing calculators for the new type of gas
- plugin-api version was changed


macfarla (commented):
got 1/3 through! can we remove the ssz tuweni snapshot lib?

macfarla (commented):
some more comments :)

jflo:
Build has sync'd with mainnet here:  

https://grafana.observability.consensys.net/d/_Mqt4ksnz/besu-full?orgId=19&var-system=dev-besu-clc-teku-ohio-mainnet-dev_4844_devnet6&var-sync_task_quantile=All&var-executor=All&var-vertx_pool_name=All&var-protocol_inbound=All&var-protocol_outbound=All&var-diskdevices=%5Ba-z%5D%2B%7Cnvme%5B0-9%5D%2Bn%5B0-9%5D%2B&var-overview_engine_percentiles=All&from=now-24h&to=now&refresh=1m

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/feemarket/TransactionPriceCalculator.java
```diff
@@ -42,4 +44,34 @@ static TransactionPriceCalculator eip1559() {
       return price;
     };
   }
+
+  private static BigInteger fakeExponential(
```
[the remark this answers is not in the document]
jflo: It does sound fishy, but thats how it's named in the spec. I'll add a comment linking here:  https://eips.ethereum.org/EIPS/eip-4844#cryptographic-helpers

gradle/check-licenses.gradle
```diff
@@ -192,24 +192,4 @@ task lazyDownloadLicenses() {
...
-    def xml = new XmlParser().parse("${rootProject.buildDir}/reports/license/license-dependency.xml")
-    xml.each { license ->
-      if (!acceptedLicenses.contains((license.@name).toLowerCase())) {
-        def depStrings = []
-        license.dependency.each { depStrings << it.text() }
-        bads = bads + depStrings + " => ${license.@name} \n"
-      }
-    }
-    if (bads != "") {
-      throw new GradleException("Some 3rd parties are using licenses not in our accepted licenses list:\n" +
```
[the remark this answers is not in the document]
jflo: I probably merged incorrectly at some point, will align with whats on main.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -299,6 +331,109 @@ public ValidationResult<TransactionInvalidReason> validateTransactionSignature(
...
+      return ValidationResult.invalid(
+          TransactionInvalidReason.INVALID_BLOBS,
+          "transaction blobs are empty, cannot verify without blobs");
+    }
+
+    BlobsWithCommitments blobsWithCommitments = transaction.getBlobsWithCommitments().get();
+
+    final long blobsLimit =
+        gasLimitCalculator.currentDataGasLimit()
+            / gasCalculator.dataGasUsed(blobsWithCommitments.getBlobs().size());
```
Gabriel-Trintinalia, line 347: DATA_GAS_LIMIT / DATA_GAS_PER_BLOB * blobCount;  
Is this correct? Why blobsLimit depends on the blob size? 
We would have 786432/ 2**17 * (6 blobs) =  786432 / 786432 = 1

Did I miss anything?

datatypes/src/main/java/org/hyperledger/besu/datatypes/Hash.java
```diff
@@ -61,6 +61,16 @@ public static Hash hash(final Bytes value) {
...
   }
 
+  /**
+   * Convert value to sha256 hash.
+   *
+   * @param value the bytes to hash
+   * @return the hash of those bytes
+   */
+  public static Hash sha256(final Bytes value) {
+    return new Hash(org.hyperledger.besu.crypto.Hash.sha256(value));
```
shemnon, line 71: I don't like this.  The datatype Hash object has always meant a container for a Keccak-256 hash, because that was the only hash type used internally.  We either need to add a type field showing what type the hash is, and check it when we use it (the critical part) or create a new HashSha256 type.

shemnon: The solution is probably to promote a datatypes variant of VersionedHash into the datatypes package.

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/CancunGasCalculator.java
```diff
@@ -22,13 +24,27 @@
...
+  private CancunGasCalculator(final int maxPrecompile) {
+    super(maxPrecompile);
+  }
 
   private static final long TLOAD_GAS = WARM_STORAGE_READ_COST;
   private static final long TSTORE_GAS = WARM_STORAGE_READ_COST;
 
   private static final long DATA_GAS_PER_BLOB = 1 << 17;
-  private static final long TARGET_DATA_GAS_PER_BLOCK = 1 << 18;
+  private static final long TARGET_DATA_GAS_PER_BLOCK = 393216L;
```
shemnon, line 47: Since it's not a simple 1-shift, a hex representation would make it clearer it's a round hex number.  `0x60000`

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/CancunGasCalculator.java
```diff
@@ -42,18 +58,34 @@ public long getTransientStoreOperationGasCost() {
...
-  public long computeExcessDataGas(final long parentExcessDataGas, final int newBlobs) {
-    final long consumedDataGas = dataGasCost(newBlobs);
-    final long currentExcessDataGas = parentExcessDataGas + consumedDataGas;
+  /**
+   * Retrieves the target data gas per block.
+   *
+   * @return The target data gas per block.
+   */
+  public long getTargetDataGasPerBlock() {
+    return TARGET_DATA_GAS_PER_BLOCK;
```
shemnon, line 71: To be clear, this is here because the target gas per block is baked into the protocol and doesn't float like the block gas limit?

jflo: Yep, thats correct

evm/src/main/java/org/hyperledger/besu/evm/operation/DataHashOperation.java
```diff
@@ -47,12 +47,12 @@ public DataHashOperation(final GasCalculator gasCalculator) {
 
   @Override
   public OperationResult execute(final MessageFrame frame, final EVM evm) {
-    int blobIndex = frame.popStackItem().toInt();
-    final Optional<List<Hash>> maybeHashes = frame.getVersionedHashes();
+    Bytes blobIndex = frame.popStackItem();
+    UInt256 blobIndexUInt256 = UInt256.fromBytes(blobIndex);
```
shemnon, line 51: I can optimize this later, but this will be slower than it should be.

(a) UInt256 is always slow.  Better to count leading zero bytes and convert to int if there are enough leading zero bytes.
(b) all actions other than the pop should be done after the check to see if versioned hashes are even present.

jflo: good point, this will nearly always fit in something as small as even a byte.

evm/src/test/java/org/hyperledger/besu/evm/operations/DataHashOperationTest.java
```diff
@@ -57,7 +60,7 @@ void pushesZeroOnBloblessTx() {
 
     EVM fakeEVM = mock(EVM.class);
 
-    DataHashOperation getHash = new DataHashOperation(new LondonGasCalculator());
+    DataHashOperation getHash = new DataHashOperation(new ShanghaiGasCalculator());
```
shemnon, line 63: Should this be `CancunGasCalculator`?

shemnon (dismissed):
Most of my comments are nits and non blockers, except the bit about `VersionedHash`  I think we need to make it  type in the `datatypes` module and ensure we consistently use it instead of a mix of `Hash` and `Bytes32`.

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockHeader.java
```diff
@@ -208,4 +208,12 @@ default Optional<Bytes32> getPrevRandao() {
...
   @Unstable
   Optional<? extends Quantity> getExcessDataGas();
+
+  /**
+   * The data_gas_used of this header.
+   *
+   * @return The data_gas_used of this header.
+   */
+  @Unstable
+  long getDataGasUsed();
```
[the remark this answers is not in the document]
jflo: It is an unsigned int64, and there is another PR that is addressing the long/DataGas mismatch:  

https://github.com/hyperledger/besu/pull/5654

evm/src/main/java/org/hyperledger/besu/evm/operation/DataHashOperation.java
shemnon: https://github.com/hyperledger/besu/blob/b3d3f54fff04e5aa2cccf96ab0a44ecfb580af31/evm/src/main/java/org/hyperledger/besu/evm/operation/ByteOperation.java#L39

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockHeader.java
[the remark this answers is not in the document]
shemnon: That PR has worse performance implications than staying with longs, there was about a 20% evm speed improvement when I moved to longs for gas away from the bespoke type the PR is proposing. 

Even though it's an unsigned 64 bit long there are `Long` methods that correctly handle comparisons and multiplication/division.  I would recommend settling on the primitive type.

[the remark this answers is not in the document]
jflo: understood, but that gives me readability concerns. when a reader sees a long, they assume it is signed, and would naturally interact with it as signed. Is there a way we can use this as a primitive but encode the "unsigned" semantics somehow? extend Long maybe?

datatypes/src/main/java/org/hyperledger/besu/datatypes/Hash.java
jflo: lol I had initially done that, but wasn't sure if the EVM wanted to be Cancun aware. Will promote it and update MessageFrame signatures to use it.

evm/src/test/resources/org/hyperledger/besu/evm/precompile/pointEvaluationPrecompile.json
```diff
@@ -11,11 +11,6 @@
       "FailureMode": "busted versioned hash",
       "Valid": false
     },
-    {
```
shemnon, line 154: Why did this test case get removed?

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockHeader.java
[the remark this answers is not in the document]
shemnon: Use https://checkerframework.org/manual/#signedness-checker and annotate it with `org.checkerframework.errorprone.checker.signedness.qual.Unsigned`.  We should look into running the signedness checker as part of the build, since unsigned longs are very much a part of the EVM.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Hash.java
shemnon: Having the EVM Cancun aware is fine.  In fact it is preferable.  The message frame already has new fields for the versioned hash, adding a new type doesn't add any incompatibilities not already added by that.

evm/src/test/resources/org/hyperledger/besu/evm/precompile/pointEvaluationPrecompile.json
jflo: Because it is now handled in the constructor of VersionedHash and isn't a failure case for the precompile itself. Corresponding test for the VersionedHash class itself is pending.

shemnon:
I've reviewed datatypes, evm, plugin, and build code and I am satisfied.  However I have not reviewed any of the code in besu, acceptance-tests, consensus, or ethereum directories as I expect other maintainers are better positioned to review those.  Ping me next week if I still need to review those directories.

I would tag this as "approved" however that alone would clear it for merge when I feel other areas need different eye, so I'll hold off.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
jflo: Thinking about it, i'm not sure this check belongs here at all. if all the transactions sum to exceed the datagas limit for the block, the block would be invalid. a transaction using too much datagas is not invalid on its own, but only invalid in the context of the block.  I think we should remove this check.

Gabriel-Trintinalia: Agreed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/ZeroBaseFeeMarket.java
```diff
@@ -38,4 +38,9 @@ public Wei computeBaseFee(
   public ValidationMode baseFeeValidationMode(final long blockNumber) {
     return ValidationMode.NONE;
   }
+
+  @Override
+  public boolean implementsDataFee() {
+    return true;
```
siladu, line 44: This was a bug introduced in https://github.com/hyperledger/besu/pull/4992 which was removed in https://github.com/hyperledger/besu/pull/5276 so it needs removing again I think.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/feemarket/ZeroBaseFeeMarketTest.java
```diff
@@ -141,12 +141,12 @@ public void isBeforeForkBlockShouldBeFalse() {
   }
 
   @Test
-  public void implementsDataFeedShouldReturnFalse() {
-    assertThat(zeroBaseFeeMarket.implementsDataFee()).isFalse();
+  public void implementsDataFeeShouldReturnTrue() {
+    assertThat(zeroBaseFeeMarket.implementsDataFee()).isTrue();
```
siladu, line 145: Why was this changed to true?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadV3.java
```diff
@@ -0,0 +1,96 @@
...
+
+      if (shanghai.isPresent() && builtAt >= this.shanghai.get().milestone()) {
+        return new JsonRpcSuccessResponse(
+            request.getRequest().getId(),
+            blockResultFactory.payloadTransactionCompleteV2(blockWithReceipts));
+      }
+
+      return new JsonRpcSuccessResponse(
+          request.getRequest().getId(),
+          blockResultFactory.payloadTransactionCompleteV1(blockWithReceipts.getBlock()));
```
siladu, line 89: This reads well! 👏

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/EnginePayloadParameter.java
```diff
@@ -66,7 +69,10 @@ public EnginePayloadParameter(
       @JsonProperty("prevRandao") final String prevRandao,
       @JsonProperty("transactions") final List<String> transactions,
       @JsonProperty("withdrawals") final List<WithdrawalParameter> withdrawals,
-      @JsonProperty("deposits") final List<DepositParameter> deposits) {
+      @JsonProperty("dataGasUsed") final UnsignedLongParameter dataGasUsed,
+      @JsonProperty("deposits") final List<DepositParameter> deposits,
+      @JsonProperty("excessDataGas") final String excessDataGas,
```
siladu, line 74: nit: deposits after 4844 params

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/UnsignedLongParameter.java
```diff
@@ -24,7 +24,7 @@ public class UnsignedLongParameter {
 
   @JsonCreator
   public UnsignedLongParameter(final String value) {
-    this.value = Long.decode(value);
+    this.value = value != null ? Long.decode(value) : 0;
```
siladu, line 27: Is 0 a safe default for all UnsignedLongParameters?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/ExecutionEngineJsonRpcMethods.java
```diff
@@ -74,56 +80,87 @@ protected String getApiGroup() {
...
+              new EngineGetPayloadBodiesByHashV1(
+                  consensusEngineServer, protocolContext, blockResultFactory, engineQosTimer),
+              new EngineGetPayloadBodiesByRangeV1(
+                  consensusEngineServer, protocolContext, blockResultFactory, engineQosTimer),
+              new EngineExchangeCapabilities(
+                  consensusEngineServer, protocolContext, engineQosTimer),
+              new EnginePreparePayloadDebug(
+                  consensusEngineServer, protocolContext, engineQosTimer, mergeCoordinator.get())));
+
+      if (protocolSchedule.anyMatch(p -> p.spec().getName().equalsIgnoreCase("cancun"))) {
```
siladu, line 152: Why does the introduction of V3 need to be triggered by Cancun? With Shanghai, some CLs started calling the next API version before the fork was triggered.

siladu (dismissed):
Reviewed acceptance-tests, consensus and part of engine API so far.

Requesting changes due to what I believe is a ZeroBaseFeeMarket bug

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/feemarket/ZeroBaseFeeMarketTest.java
siladu: ^ ignore - guess it was due to merging the implementsDataFee commit

acceptance-tests/tests/src/test/resources/jsonrpc/engine/eip6110/test-cases/06_eip6110_get_payload.json
```diff
@@ -39,7 +39,7 @@
         ],
         "deposits" : [],
         "blockNumber": "0x2",
-        "blockHash": "0x58ea3e01b03ac17c68ed3e3d724a021408273fac8a86f42cb30a26be8e93fbe9",
+        "blockHash": "0x4c4418c408aeadb4659d31d1c05108f26fabf713bb6f8cc487dba8424a725bf5",
```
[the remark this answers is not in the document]
jflo: @Gabriel-Trintinalia has details, basically new input fields to the hash

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/ExecutionEngineJsonRpcMethods.java
jflo: It's more for configuration consistency than timing. If you can't hit V3 of this api, maybe you don't have cancun enabled, and you're gonna need to fix that.

Will consider removing this though, and just always enabling it since the api itself is timestamp aware.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ScheduledProtocolSpec.java
```diff
@@ -25,10 +27,24 @@ public interface ScheduledProtocolSpec {
 
   boolean isOnMilestoneBoundary(ProcessableBlockHeader header);
 
-  long milestone();
+  // long milestone();
+  Hardfork fork();
 
   ProtocolSpec spec();
 
+  public record Hardfork(String name, long milestone) implements Comparable<Hardfork> {
```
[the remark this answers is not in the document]
jflo: Not sure we _need_ it since it is a readability improvement, but it encapsulates transition points and allows them to be compared, which is used by Engine APIs.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsBundleV1.java
```diff
@@ -0,0 +1,75 @@
...
+import org.hyperledger.besu.ethereum.core.Block;
+import org.hyperledger.besu.ethereum.core.BlockWithReceipts;
+import org.hyperledger.besu.ethereum.core.Transaction;
+import org.hyperledger.besu.ethereum.mainnet.ProtocolSchedule;
+
+import java.util.List;
+
+import io.vertx.core.Vertx;
+
+public class EngineGetBlobsBundleV1 extends AbstractEngineGetPayload {
```
fab-10, line 36: wasn't this merge with getPayloadV3?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
```diff
@@ -0,0 +1,130 @@
...
+          || payloadParameter.getExcessDataGas() == null) {
+        return Optional.of(new JsonRpcErrorResponse(reqId, INVALID_PARAMS));
+      }
+    } else {
+      if (payloadParameter.getDataGasUsed() != null
+          || payloadParameter.getExcessDataGas() != null) {
+        return Optional.of(new JsonRpcErrorResponse(reqId, UNSUPPORTED_FORK));
+      }
+    }
+    return Optional.empty();
```
fab-10, line 76: should we check versioned hashes param here too?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/EnginePayloadParameter.java
```diff
@@ -48,6 +48,9 @@ public class EnginePayloadParameter {
   private final List<String> transactions;
   private final List<WithdrawalParameter> withdrawals;
   private final List<DepositParameter> deposits;
+  private final List<Bytes32> versionedHashes;
+  private final Long dataGasUsed;
+  private final String excessDataGas;
```
fab-10, line 53: since they are both `QUANTITY` of 64 bits, it is better to uniform them using the same type for both

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/UnsignedLongParameter.java
fab-10: instead of supporting null, what if introducing an `optionalUnsignedLongParameters`?

fab-10 (commented):
in progress 52/155, please check my comments

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/UnsignedLongParameter.java
jflo: I'm not sure. I'm more sure that 'null' is unsafe... would you prefer 0, null, or a thrown exception of some sort here? Other options?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -276,6 +299,35 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
     }
   }
 
+  private Optional<String> validateBlobTransactions(
```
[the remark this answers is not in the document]
jflo: It's not used anywhere else, and is very CL specific; the error message is informational, not an enumeration that the CL can use.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsBundleV1.java
jflo: removed

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/EnginePayloadParameter.java
jflo: Agreed, and I have a PR that unifies those. #5654 
@shemnon has some performance concerns, I have some readability concerns, and we think we can find a compromise. see convo here https://github.com/hyperledger/besu/pull/5536#discussion_r1247180442

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ScheduledProtocolSpec.java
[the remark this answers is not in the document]
siladu: I like how it reads in the client code for sure. I haven't thought through the readability of using ScheduledProtocolSpec instead. 

My concern is that there is a potential comprehension/complexity issue with introducing yet another Type that represents something about a fork... there's already milestone, ProtocolSpec, ScheduledProtocolSpec, ForkSpec (for bft)

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/UnsignedLongParameter.java
siladu: I guess I don't understand the impact on any existing UnsignedLongParameter value. If there is no impact, i.e. no bug introduced by this, then I'm easy with any of those. Exception might be more explicit and good 'fail early' practice though.

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/CombinedProtocolScheduleFactory.java
```diff
@@ -52,7 +52,9 @@ public BftProtocolSchedule create(
           Optional.ofNullable(forkSpecs.higher(spec)).map(ForkSpec::getBlock);
       protocolSchedule.getScheduledProtocolSpecs().stream()
           .filter(protocolSpecMatchesConsensusBlockRange(spec.getBlock(), endBlock))
-          .forEach(s -> combinedProtocolSchedule.putBlockNumberMilestone(s.milestone(), s.spec()));
+          .forEach(
+              s ->
+                  combinedProtocolSchedule.putBlockNumberMilestone(s.fork().milestone(), s.spec()));
```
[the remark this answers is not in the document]
siladu: The HardFork layer of indirection:
`s.fork().milestone()` instead of `s.milestone()`

See https://github.com/hyperledger/besu/pull/5536#discussion_r1251804232

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
```diff
@@ -0,0 +1,130 @@
...
+  private Optional<JsonRpcResponse> validateBlobTransactions(
+      final EnginePayloadParameter blockParam,
+      final Object reqId,
+      final List<Transaction> blobTransactions,
+      final Optional<List<String>> maybeVersionedHashParam) {
+
+    List<Bytes32> versionedHashesParam =
+        maybeVersionedHashParam
+            .map(strings -> strings.stream().map(Bytes32::fromHexStringStrict).toList())
+            .orElseGet(ArrayList::new);
```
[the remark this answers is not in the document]
siladu: @fab-10 Not as elegant, but would `.orElseGet(() -> new ArrayList())` have the same effect?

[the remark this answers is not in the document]
fab-10: @siladu  `.orElseGet(() -> new ArrayList())` is equal to `.orElseGet(ArrayList::new)`, what I suggest is when you need to return an immutable empty array `List.of()` does not allocate memory.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3Test.java
```diff
@@ -0,0 +1,164 @@
...
+            new JsonRpcRequest(
+                "2.0",
+                RpcMethod.ENGINE_NEW_PAYLOAD_V3.getMethodName(),
+                new Object[] {payload, versionedHashes})));
+  }
+
+  @Override
+  @Ignore
+  public void shouldRespondWithSyncingDuringBackwardsSync() {}
+}
```
siladu, line 164: Is this hiding a bug?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/UnsignedLongParameter.java
fab-10: I agree with the backward compatibility, and looking at similar classes this will bit a bit odd, since none is supporting null, so we probably need to find another way to reach the same goal

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -228,7 +228,26 @@ protected BlockCreationResult createBlock(
...
+                    parentHeader.getDataGasUsed().orElse(0L)));
+
+        final int newBlobsCount =
+            transactionResults.getTransactionsByType(TransactionType.BLOB).stream()
+                .map(tx -> tx.getVersionedHashes().orElseThrow())
+                .mapToInt(List::size)
+                .sum();
+
+        newDataGasUsed = gasCalculator.dataGasUsed(newBlobsCount);
+      }
```
siladu, line 251: Think this is missing unit tests?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/GasLimitCalculator.java
```diff
@@ -16,13 +16,15 @@
 
 public interface GasLimitCalculator {
 
+  static final long DATA_GAS_LIMIT = 786432;
```
siladu, line 19: Is this a duplicate of MAX_DATA_GAS_PER_BLOCK?

Maybe out of scope of this PR, but does anyone think there's value in having all the specified constants grouped together in one place so they as easily searchable/comparable to the spec?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
```diff
@@ -57,18 +59,26 @@ public void writeTo(final RLPOutput out) {
     out.startList();
 
     header.writeTo(out);
-    body.writeTo(out);
+    out.writeList(body.getTransactions(), Transaction::writeTo);
+    out.writeList(body.getOmmers(), BlockHeader::writeTo);
+    body.getWithdrawals().ifPresent(withdrawals -> out.writeList(withdrawals, Withdrawal::writeTo));
+    body.getDeposits().ifPresent(deposits -> out.writeList(deposits, Deposit::writeTo));
```
siladu, line 65: Why have we duplicated the contents of BlockBody.writeTo() here?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
```diff
@@ -57,18 +59,26 @@ public void writeTo(final RLPOutput out) {
...
   public static Block readFrom(final RLPInput in, final BlockHeaderFunctions hashFunction) {
     in.enterList();
     final BlockHeader header = BlockHeader.readFrom(in, hashFunction);
-    final BlockBody body = BlockBody.readFrom(in, hashFunction);
+    final List<Transaction> transactions = in.readList(Transaction::readFrom);
+    final List<BlockHeader> ommers = in.readList(rlp -> BlockHeader.readFrom(rlp, hashFunction));
+    final Optional<List<Withdrawal>> withdrawals =
+        in.isEndOfCurrentList() ? Optional.empty() : Optional.of(in.readList(Withdrawal::readFrom));
+    final Optional<List<Deposit>> deposits =
+        in.isEndOfCurrentList() ? Optional.empty() : Optional.of(in.readList(Deposit::readFrom));
```
siladu, line 78: Same point about duplicating BlockBody.readFrom

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -558,6 +620,13 @@ private Bytes32 getOrComputeSenderRecoveryHash() {
    * @param out the output to write the transaction to
    */
   public void writeTo(final RLPOutput out) {
+    // there is context missing here. Are we writing a transaction (possibly type 0x03, Blobs) to be
+    // gossiped out,
+    // or to be written within a block.  Gossiped type 0x03s need to include blobs and proofs.
+    // Writing transactions
+    // to a new block being produced, or to a block being imported, should not include blobs or
+    // proofs.
+    // TODO: remove this method, require direct interaction with TransactionEncoder
```
siladu, line 629: Checking this TODO is being left in intentionally.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionBuilderTest.java
```diff
@@ -68,11 +71,27 @@ public void zeroBlobTransactionIsInvalid() {
...
+    String actualMessage =
+        assertThrows(
+                InvalidParameterException.class,
+                () ->
+                    new BlobsWithCommitments(
+                        List.of(new KZGCommitment(Bytes.of(1))), List.of(), List.of()))
+            .getMessage();
+    final String expectedMessage = "There must be an equal number of blobs, commitments and proofs";
+    assertThat(actualMessage).isEqualTo(expectedMessage);
+  }
```
siladu, line 96: Should be in BlobsWithCommitementsTest

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -617,9 +686,15 @@ private void memoizeHashAndSize() {
...
-    size = rlpOutput.encodedSize();
+    if (transactionType.supportsBlob()) {
+      if (getBlobsWithCommitments().isPresent()) {
+        size = TransactionEncoder.encodeOpaqueBytes(this).size();
+      }
+    } else {
+      final BytesValueRLPOutput rlpOutput = new BytesValueRLPOutput();
+      TransactionEncoder.encodeForWire(transactionType, bytes, rlpOutput);
+      size = rlpOutput.encodedSize();
+    }
```
siladu, line 697: Is this covered by units somewhere, struggling to spot it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -1002,7 +1082,11 @@ public int hashCode() {
   @Override
   public String toString() {
     final StringBuilder sb = new StringBuilder();
-    sb.append(isContractCreation() ? "ContractCreation" : "MessageCall").append("{");
+    sb.append(
+            transactionType.supportsBlob()
+                ? "Blob"
+                : isContractCreation() ? "ContractCreation" : "MessageCall")
+        .append("{");
```
siladu, line 1089: also wonder about units for this

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/BlobTransactionDecoder.java
```diff
@@ -0,0 +1,108 @@
...
+        .accessList(
+            input.readList(
+                accessListEntryRLPInput -> {
+                  accessListEntryRLPInput.enterList();
+                  final AccessListEntry accessListEntry =
+                      new AccessListEntry(
+                          Address.wrap(accessListEntryRLPInput.readBytes()),
+                          accessListEntryRLPInput.readList(RLPInput::readBytes32));
+                  accessListEntryRLPInput.leaveList();
+                  return accessListEntry;
```
siladu, line 80: this inner list isn't covered by units

siladu (commented):
> Reviewed acceptance-tests, consensus and part of engine API so far.

Second pass 76/152: finished engine, api, started on ethereum

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java
```diff
@@ -212,8 +216,9 @@ public void writeTo(final RLPOutput out) {
     if (withdrawalsRoot != null) {
       out.writeBytes(withdrawalsRoot);
     }
-    if (excessDataGas != null) {
-      out.writeUInt256Scalar(excessDataGas);
+    if (excessDataGas != null && dataGasUsed != null) {
+      out.writeLongScalar(dataGasUsed);
```
fab-10, line 220: unsignedLongScalar?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -1094,8 +1182,8 @@ public static class Builder {
 
     protected Optional<BigInteger> chainId = Optional.empty();
 
-    protected Optional<BigInteger> v = Optional.empty();
```
fab-10, line 1097: Q: why we do not need `v` anymore?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/BodyValidationTest.java
```diff
@@ -58,6 +59,7 @@ public void calculateWithdrawalsRoot() throws IOException {
     }
   }
 
+  @Ignore // TODO: RLP encoding has changed, so testdata needs to be updated
```
fab-10, line 62: do you plan to re-enable this test after this PR?

fab-10 (commented):
152/152, last comments

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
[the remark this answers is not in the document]
siladu: ah got you. () -> is lazy whereas method reference is eagerly evaluated, they are subtly different - that's what I thought you were getting at by mentioning allocation.
So I think with Arraylist::new, there might be an allocation even if the map maybeVersionedHashParam is present.

https://docs.oracle.com/javase/specs/jls/se13/html/jls-15.html#jls-15.13.3

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/EnginePayloadParameter.java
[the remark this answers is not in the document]
siladu: Would ignoring omitted json fields suffice for all versions? That is assuming we only ever add/remove fields rather than rename.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/BodyValidationTest.java
siladu: cc @naviechan

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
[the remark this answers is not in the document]
jflo: I share that nit, and will probably refactor this toward stronger typing during upcoming rpc-api review.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/encoding/TransactionRLPDecoderTest.java
```diff
@@ -32,7 +32,7 @@
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 
-class TransactionDecoderTest {
+class TransactionRLPDecoderTest {
```
siladu, line 35: Is this "as opposed to SSZ"? Should we update the production class names as well to match?

Not for this PR, but maybe the other Encoders/Decoders should be renamed to

siladu (commented):
155/155 - I skipped the evm stuff since Danno covered it.

It's mostly a superficial review, I haven't reviewed against the spec because I assume that is well covered by the tests and devnets.

Overall, decent test coverage and good code comments.
Most of my PR comments are nits or questions, only potential blockers would be missing units for the more complex logic.

Since I complained about the size of Fabio's PR, it's only fair I say the same about this 😆 

I'd be happy to approve once outstanding comments are considered.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -107,7 +113,9 @@ public class Transaction
   private final TransactionType transactionType;
 
   private final SignatureAlgorithm signatureAlgorithm = SignatureAlgorithmFactory.getInstance();
-  private final Optional<List<Hash>> versionedHashes;
+  private final Optional<List<VersionedHash>> versionedHashes;
+
+  private final Optional<BlobsWithCommitments> blobsWithCommitments;
```
siladu, line 118: Think these should be added to the equals method as well

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockHeader.java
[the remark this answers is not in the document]
jflo: I have reverted it to use longs, and will introduce checker framework in a later branch.  

https://github.com/hyperledger/besu/issues/5696

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
```diff
@@ -0,0 +1,130 @@
...
+  public String getName() {
+    return RpcMethod.ENGINE_NEW_PAYLOAD_V3.getMethodName();
+  }
+
+  @Override
+  protected Optional<JsonRpcResponse> validateForkSupported(
+      final Object reqId, final EnginePayloadParameter payloadParameter) {
+    var cancun = timestampSchedule.hardforkFor(s -> s.fork().name().equalsIgnoreCase("Cancun"));
+
+    if (cancun.isPresent() && payloadParameter.getTimestamp() >= cancun.get().milestone()) {
```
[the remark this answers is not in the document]
jflo: Yes, and discussions are ongoing to revise designs toward that. Devnet-8 avoids it.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
jflo: This is handled optionally in the parent class.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3Test.java
jflo: This has been moved up to abstract parent and is no longer ignored.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -97,13 +104,32 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+
+    final Optional<BlockHeader> maybeParentHeader =
+        protocolContext.getBlockchain().getBlockHeader(blockParam.getParentHash());
 
     LOG.atTrace()
         .setMessage("blockparam: {}")
         .addArgument(() -> Json.encodePrettily(blockParam))
         .log();
 
+    /*
```
fab-10, line 126: is this required? if yes could you add a comment to explain

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -172,7 +198,10 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
             blockParam.getPrevRandao(),
             0,
             maybeWithdrawals.map(BodyValidation::withdrawalsRoot).orElse(null),
-            null,
+            blockParam.getDataGasUsed() == null ? null : blockParam.getDataGasUsed(),
```
fab-10, line 201:
```suggestion
            blockParam.getDataGasUsed(),
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/GasLimitCalculator.java
jflo: Yes it is a duplicate, but could be fixable with the right refactor. That limit is a per-block concern, so likely does not belong in this class, but rather CancunTargetingGasLimitCalculator instead.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
jflo: This was necessary when SSZ was in the picture, will try to revert.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetPayloadV3.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public EngineGetPayloadV3(
+      final Vertx vertx,
+      final ProtocolContext protocolContext,
+      final MergeMiningCoordinator mergeMiningCoordinator,
+      final BlockResultFactory blockResultFactory,
+      final EngineCallListener engineCallListener,
+      final ProtocolSchedule schedule) {
+    super(vertx, protocolContext, mergeMiningCoordinator, blockResultFactory, engineCallListener);
+    this.shanghai = schedule.hardforkFor(s -> s.fork().name().equalsIgnoreCase("Shanghai"));
```
fab-10, line 51: is it possible to use enums instead of strings here?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -169,10 +171,10 @@ protected BlockCreationResult createBlock(
           createPendingBlockHeader(timestamp, maybePrevRandao, newProtocolSpec);
       final Address miningBeneficiary =
           miningBeneficiaryCalculator.getMiningBeneficiary(processableBlockHeader.getNumber());
-      final Wei dataGasPrice =
+      Wei dataGasPrice =
```
fab-10, line 174: why no more final?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -228,11 +230,11 @@ protected BlockCreationResult createBlock(
 
       throwIfStopped();
 
-      final DataGas newExcessDataGas = computeExcessDataGas(transactionResults, newProtocolSpec);
+      final GasUsage usage = computeExcessDataGas(transactionResults, newProtocolSpec);
```
fab-10, line 233: `usage` is quire generic here, perhaps rename to `dataGasUsage`?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -280,8 +285,11 @@ List<Deposit> findDepositsFromReceipts(final TransactionSelectionResults transac
         .toList();
   }
 
-  private DataGas computeExcessDataGas(
-      TransactionSelectionResults transactionResults, ProtocolSpec newProtocolSpec) {
+  record GasUsage(DataGas excessDataGas, DataGas used) {}
```
fab-10, line 288: rename to `DataGasUsage`?

fab-10 (commented):
another pass

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/UnsignedLongParameter.java
jflo: Turns out this is only ever used by test code, and never passes in null, so this is fine to revert.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
jflo: It is, I have added a skeleton of one, but completing it turned into a rabbit hole, so I will leave the test ignored for now, and complete it once this branch is merged to main.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -154,6 +160,10 @@ public BlockProcessingResult processBlock(
           transactionReceiptFactory.create(
               transaction.getType(), result, worldState, currentGasUsed);
       receipts.add(transactionReceipt);
+      if (LOG.isTraceEnabled()) {
+        LOG.trace("warning, intermittent state root debugging on, performance hit");
+        LOG.trace("intermittent state root: {}", worldState.rootHash());
+      }
```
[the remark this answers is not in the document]
jflo: This has been removed.

jflo:
this work has been ported to https://github.com/jflo/besu/tree/EIP-4844 and already merged into main via https://github.com/hyperledger/besu/pull/5724

Thanks to all who contributed comments!

PR #5542, opened 2023-06-07
Improve log of p2p messages

fab-10:
## PR description
For most p2p messages the `toString` was not implemented, and that made it harder to trace exchanged messages. This PR implements a base `toString` that prints the raw data, and if implemented in the sub class also the decoded format of the data.

Since message data could be quite large, the log system is configured to easily change the max amount of data written, according to the user needs, by default 1000 chars are written for raw data and decoded data.


gfukushima (commented):
Need to fix the CodeQL failure

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/PeerInfoTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.List;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.junit.jupiter.api.Test;
+
+class PeerInfoTest {
+
+  @Test
+  public void toStringIsSanitized() {
```
gfukushima, line 27: Nice to have this test. Thanks

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/rlpx/wire/PeerInfo.java
```diff
@@ -97,9 +99,12 @@ public Bytes getNodeId() {
   }
 
   public Address getAddress() {
-    final SECPPublicKey remotePublicKey =
-        SignatureAlgorithmFactory.getInstance().createPublicKey(nodeId);
-    return Util.publicKeyToAddress(remotePublicKey);
+    if (address == null) {
```
gfukushima, line 102: Nice small optimisation

gfukushima (approved):
lgtm

PR #5548, opened 2023-06-08
Fix graphql storageAt for bonsai

### LINKED ISSUE #5524: [GraphQL] eth_getStorageAt reporting the wrong storage
When using the graphQLcombined with Bonsai the we report the wrong storage.
This is likely due to some optimizations besu has when fetching an account (we don't load code and storage).
#5523 fixes the account code loading. But the storage will probably require a little extra effort since we don't have enough context to the right slot at that part of the code.
A new world state snapshot could be open in the AccountAdapter (which has the queried slot) provided that we pass the block number to that class. 

This has been detected by the GraphQL hive test below

```
26_eth_getStorageAt
expected value(s):
{
  "data": {
    "block": {
      "account": {
        "storage": "0xaabbccffffffffffffffffffffffffffffffffffffffffffffffffffffffffee"
      }
    }
  }
} 

got: {
  "data": {
    "block": {
      "account": {
        "storage": "0x0000000000000000000000000000000000000000000000000000000000000000"
      }
    }
  }
}
```

gfukushima:
## PR description
This PR adds a check to the AccountAdapter used by the graphQL suite to ensure that if the account is a bonsaiAccount we open a new snapshot to retrieve the storage slot requested.

## Fixed Issue(s)

Fixes #5524


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/internal/pojoadapter/BlockAdapterBase.java
```diff
@@ -142,16 +142,12 @@ public Long getNumber() {
   }
 
   public AccountAdapter getAccount(final DataFetchingEnvironment environment) {
-
     final BlockchainQueries query = getBlockchainQueries(environment);
     final long bn = header.getNumber();
+    final Address address = environment.getArgument("address");
```
shemnon, line 147: Is there a reason to take this variable outside of the closure scope?

shemnon (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/internal/pojoadapter/BlockAdapterBase.java
gfukushima: No strong reason it was to make the Function of getAndMapWorldState as clean as possible. Happy to move this code back to it's original place if you think it's not worth keep the function more readable.

PR #5581, opened 2023-06-11
Migrate util module to JUnit 5.0

7suyash7:
Migrate the util module from Junit 4.0 to Junit 5.0 
#5556


7suyash7:
If someone could re-run ci for `acceptanceTestsPrivacy` that would be great because it shows that all the tests have passed upon going to the details?

macfarla (approved):

PR #5606, opened 2023-06-15
Option to disable txpool

fab-10:


macfarla:
@fab-10 is this PR one you're going to come back to?

fab-10:
I still want to add this option, but not sure when, and probably with a different approach, so I can close this.

PR #5611, opened 2023-06-16
Ensure that blobs, proofs and  commitments have the same number of elements

### LINKED ISSUE #5488: Limit blob related fields
All arrays of blob related fields should be checked to be less than specified length.
(currently 1024 at time of writing)
https://github.com/ethereum/execution-apis/pull/404/files

Gabriel-Trintinalia:
## PR description
Ensure blobs proofs commitments have the same size

## Fixed Issue(s)
fixes #5488

### DIFF
--- a/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionBuilderTest.java
+++ b/ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionBuilderTest.java
@@ -17,16 +17,20 @@ package org.hyperledger.besu.ethereum.core;
 
 import static java.util.stream.Collectors.toUnmodifiableSet;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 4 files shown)


ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionBuilderTest.java
```diff
@@ -76,4 +81,17 @@ public void zeroBlobTransactionIsInvalid() {
...
+  @Test
+  public void blobsWithCommitmentsMustHaveSameNumberOfElements() {
+    String actualMessage =
+        assertThrows(
+                InvalidParameterException.class,
+                () ->
+                    new BlobsWithCommitments(
+                        List.of(new KZGCommitment(Bytes.of(1))), List.of(), List.of()))
+            .getMessage();
+    String expectedMessage = "There must be an equal number of blobs, commitments and proofs";
```
macfarla, line 94:
```suggestion
    final String expectedMessage = "There must be an equal number of blobs, commitments and proofs";
```

macfarla (approved):

Issue #5597, opened 2023-06-14
ConcurrentModificationException in TransactionPool.addRemoteTransaction()


matthew1001:
### Description

```
2023-06-13 19:49:44.061+00:00 | EthScheduler-Transactions-0 | ERROR | Besu | Uncaught exception in thread "EthScheduler-Transactions-0"
java.util.ConcurrentModificationException
	at java.base/java.util.TreeMap$ValueSpliterator.forEachRemaining(TreeMap.java:3226)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at org.hyperledger.besu.ethereum.eth.transactions.sorter.AbstractPendingTransactionsSorter.signalInvalidAndGetDependentTransactions(AbstractPendingTransactionsSorter.java:539)
	at org.hyperledger.besu.ethereum.eth.transactions.sorter.AbstractPendingTransactionsSorter.signalInvalidAndRemoveDependentTransactions(AbstractPendingTransactionsSorter.java:546)
	at org.hyperledger.besu.ethereum.eth.transactions.TransactionPool.addRemoteTransaction(TransactionPool.java:340)
	at org.hyperledger.besu.ethereum.eth.transactions.TransactionPool.lambda$addRemoteTransactions$9(TransactionPool.java:268)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.SortedOps$SizedRefSortingSink.end(SortedOps.java:357)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:510)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
[...]

PR #5612, opened 2023-06-16
Add lock to creation of pending transaction stream

matthew1001:
## PR description
This PR does 2 things:

1. Takes a lock on `pendingTransactions` when creating a stream of the transactions
2. Creates a copy of the current transactions when creating the stream, rather than streaming the original list

I have tested the changes with a cut down version `PendingTransactionsForSender` class. There are a couple of points to note for the PR reviewer:

1. I don't know this area of the code well enough to know if there is likely to be a noticeable performance impact of this change. However, without it (or some alternative fix) we'll continue to see `ConcurrentModificationException` be thrown.
2. `List.copyOf` will throw a `ConcurrentModificationException` if the underlying list is changed during the call. Currently `PendingTransactionsForSender.getPendingTransactions(...)` uses that call as well, but without taking a lock on the list. I think this call could fail in the same way and I'm happy to make that change in this PR but would appreciate feedback from the reviewer first.
3. The code path that hit the issue was ultimately down to `AbstractPendingTransactionsSorter.signalInvalidAndRemoveDependentTransactions` being called which wasn't expected for the test I was running. However, the reason for the transaction and its dependent transactions being considered invalid is only logged out if trace is enabled. I wonder if this happens so often that logging it out as a `debug` message instead would be too noisy?

## Fixed Issue(s)
fixes https://github.com/hyperledger/besu/issues/5597


macfarla:
@fab-10 can you review this one?

fab-10 (commented):
Having spent a lot of time on the txpool lately, I reckon that is not easy to fully grasp the concurrency management of this implementation of the txpool at the first view, so I can say that your fix should works because the other paths are covered by the main lock on the `AbstractPendingTransactionsSorter` class, and for the same reason there is an alternative implementation of this fix that should avoid the copy, that I will propose so you can try it.
In any case I do not think that the copy is a big overhead, but instead it is more consistent to managed all the synchronization in the `AbstractPendingTransactionsSorter` that act as the entry point. 
As said in the ticket evaluate switching to the new layered txpool.

fab-10:
@matthew1001 it seems that I cannot open a PR against your fork, so please check the alternative fix here https://github.com/fab-10/besu/commit/2ecc7b95527c79ccbbece596ff52879a818dc98c

matthew1001:
Thanks for the background @fab-10, that's really useful and it's great to know there's a new & improved version of the tx-pool to try. I'd be happy to try it out in our test environment. Regarding the alternative fix, I'll check why you couldn't raise a PR against my fork (probably just an unintentional setting in github) but I can see the point of taking the lock one level higher so the lower level code doesn't need to take a copy of the transactions.

matthew1001:
@fab-10 I've just pushed your suggested change to my branch so the updated PR should be using your alternative fix now

fab-10 (approved):
LGTM, had you the chance to test this implementation in your network?

matthew1001:
Not yet @fab-10 it will probably be a week or two for us to pull the fix into our own build of Besu and run enough tests to see if we hit it again

fab-10:
if you agree we can merge after the test

matthew1001:
@fab-10 we've had this fix in our test environment for the past week or so and I've not seen any recurrence of the problem. Happy for you to merge it if you're happy that's long enough to flush out any issues.

fab-10:
Thanks for the test, for it is fine to merge it now.

PR #5618, opened 2023-06-19
[4844] Check params earlier and move blob validation to a new method

Gabriel-Trintinalia:
## PR description
Check the params earlier and move blob validation to a new method

In case CL sends a payload with invalid blob params for a known header, we should return invalid. 

Expected Behaviour: return invalid

Actual Behaviour: Returns are valid because we have already seen that payload.

This PR moves the validation earlier in the flow and encapsulates the validation login into a method for readability.


macfarla (approved):
LGTM

Gabriel-Trintinalia:
closed in favour of https://github.com/hyperledger/besu/pull/5663

PR #5619, opened 2023-06-19
Reload flat db strategy on clear

matkt:
## PR description

During the snapsync we can have a nullpointer because of a clear we are doing when we are starting from scratch


macfarla (approved):
LGTM

PR #5629, opened 2023-06-21
Decouple JsonRpcError enum from the data field

### LINKED ISSUE #5437: Decouple JsonRpcError enum from the data field
JsonRpcError enum has code/message. It also has a data field which is used sometimes - currently for revert reason. But it's a bit of a hack because it gets reset as soon as that enum field gets used again (so, for example, you can't use this in unit tests to test actual vs expected values of the data field). 

~~See #5421 for an example~~

See #5241 for an example

gtebrean:
## PR description
Extracts data field from JsonRpcError by moving the ENUM values in RpcErrorType and keeping JsonRpcError only for building the expected json error format.

As the Enum class has been renamed to RpcErrorType a lot of files are appearing as changed but the only major changes are in:

JsonRpcError
RpcErrorType
JsonRpcErrorResponse
PrivCall
EthCall
JsonRpcUnauthorizedResponse

## Fixed Issue(s)

Fixes #5437


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/JsonRpcErrorResponse.java
```diff
@@ -21,15 +21,22 @@
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.google.common.base.MoreObjects;
 
-@JsonPropertyOrder({"jsonrpc", "id", "error"})
+@JsonPropertyOrder({"jsonrpc", "id", "error", "data"})
 public class JsonRpcErrorResponse implements JsonRpcResponse {
```
gtebrean, line 25: To avoid adding all `data: null` in the test resorce file we can add:
`@JsonInclude(value = JsonInclude.Include.NON_NULL)`
So response will be build without any null field and I think will make more sense than returning a data=null. I wanted first to do it like this but I thought maybe you will want to have it returned.

gtebrean:
closing this due missing signatures of some commits

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/PluginJsonRpcMethod.java
```diff
@@ -53,13 +53,13 @@ public JsonRpcResponse response(final JsonRpcRequestContext request) {
       final Object result = function.apply(() -> request.getRequest().getParams());
       return new JsonRpcSuccessResponse(request.getRequest().getId(), result);
     } catch (final PluginRpcEndpointException ex) {
-      final JsonRpcError error = PLUGIN_INTERNAL_ERROR;
-      error.setData(ex.getMessage());
+      final JsonRpcError error = new JsonRpcError(PLUGIN_INTERNAL_ERROR, ex.getMessage());
```
macfarla, line 56: much nicer

macfarla (commented):
LGTM. touches a lot of files but the change is pretty straight forward. 

would like a second opinion though since it's used so widely

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/JsonRpcErrorResponse.java
```diff
@@ -69,4 +74,12 @@ public int hashCode() {
...
     return MoreObjects.toStringHelper(this).add("id", id).add("error", error).toString();
   }
+
+  @JsonIgnore
+  public RpcErrorType getErrorType() {
+    return Arrays.stream(RpcErrorType.values())
+        .filter(e -> e.getCode() == error.getCode() && e.getMessage().equals(error.getMessage()))
+        .findFirst()
+        .get();
+  }
```
siladu, line 84: nit: Might be nicer to have an errorType property on JsonRpcError that is ignored for serialization purposes, so we can just call error.getErrorType() instead of this look-up.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/JsonRpcUnauthorizedResponse.java
```diff
@@ -56,11 +61,19 @@ public boolean equals(final Object o) {
...
     return Objects.hash(id, error);
   }
+
+  @JsonIgnore
+  public RpcErrorType getErrorType() {
+    return Arrays.stream(RpcErrorType.values())
+        .filter(e -> e.getCode() == error.getCode() && e.getMessage().equals(error.getMessage()))
+        .findFirst()
+        .get();
+  }
```
siladu, line 78: nit: same point about preferring `JsonRpcError.errorType`

siladu (approved):
Good change 👍 
Nothing major, just a couple of nitpicks.

Thanks for pointing out the important changes for this PR in the description, very helpful!

consensus/clique/src/test/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/CliqueGetSignersAtHashTest.java
```diff
@@ -130,6 +130,6 @@ public void failsOnInvalidBlockHash() {
         .thenReturn(Optional.empty());
 
     final JsonRpcErrorResponse response = (JsonRpcErrorResponse) method.response(request);
-    assertThat(response.getError().name()).isEqualTo(JsonRpcError.INTERNAL_ERROR.name());
+    assertThat(response.getErrorType().name()).isEqualTo(RpcErrorType.INTERNAL_ERROR.name());
```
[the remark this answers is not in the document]
siladu: Think there might be some more of these too.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/JsonRpcErrorResponse.java
gtebrean: Updated but I need to do one lookup anyway when I deserialise the response but is better than doing it every time when requested.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/JsonRpcUnauthorizedResponse.java
gtebrean: same as above

gtebrean:
Comments fixed, @macfarla please check again.

siladu:
Thanks @gtebrean, updates look good. I'll give @macfarla a bit more time to re-review and then I am happy to merge it for you.

macfarla (approved):
LGTM

PR #5634, opened 2023-06-22
Start with txpool disabled

### LINKED ISSUE #5636: Failure at startup when enabling layered txpool before initial sync done
### Description

Enabling the layered txpool #5290 before the initial sync is done, causes Besu to fail to start with this error

```
2023-06-23 12:48:11.172+02:00 | main | INFO  | TransactionPoolFactory | Using layered transaction pool
class org.hyperledger.besu.ethereum.mainnet.feemarket.LegacyFeeMarket cannot be cast to class org.hyperledger.besu.ethereum.mainnet.feemarket.BaseFeeMarket (org.hyperledger.besu.ethereum.mainnet.feemarket.LegacyFeeMarket and org.hyperledger.besu.ethereum.mainnet.feemarket.BaseFeeMarket are in unnamed module of loader 'app')
```

Workaround until the fix is ready, is to enabled the layered txpool only after Besu is synced.

### Steps to Reproduce (Bug)
Start Besu with `--Xlayered-tx-pool=true`

Only when syncing

* Software version: >= 23.4.1

fab-10:
## PR description

This PR review the way the txpool is enabled/disabled in a more intuitive way and also[ fix a bug ](https://github.com/hyperledger/besu/issues/5636)that happens when enabling the layered txpool when not synced.

The main point of disabling the txpool, is to avoid its overhead during the initial sync (or resync) phases, since pending transactions are always evaluated and processed against the current head, so if we are syncing there is no value in processing them, and we can wait until the sync is done to enabled the txpool.
The current implementation is a bit convoluted, since the txpool is always enabled at startup, but if the initial sync is not done, then the txpool is disabled, to be re-enabled after the sync is done, while it is straightforward to just start with the txpool disabled, and enabled it when the sync is done.

## Fixed Issue(s)

fixes #5636

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
@@ -60,6 +60,8 @@ import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
@@ -522,25 +524,28 @@ public class TransactionPool implements BlockAddedObserver {
     }
   }
 
-  public void setEnabled() {
+  public CompletableFuture<Void> setEnabled() {
     if (!isEnabled()) {
       pendingTransactions = pendingTransactionsSupplier.get();
       pendingTransactionsListenersProxy.subscribe();
       isPoolEnabled.set(true);
       subscribeConnectId =
           OptionalLong.of(ethContext.getEthPeers().subscribeConnect(this::handleConnect));
-      saveRestoreManager.loadFromDisk();
+      return saveRestoreManager.loadFromDisk();
     }
+    return CompletableFuture.completedFuture(null);
   }
 
-  public void setDisabled() {
+  public CompletableFuture<Void> setDisabled() {
     if (isEnabled()) {
       isPoolEnabled.set(false);
       subscribeConnectId.ifPresent(ethContext.getEthPeers()::unsubscribeConnect);
       pendingTransactionsListenersProxy.unsubscribe();
-      saveRestoreManager.saveToDisk(pendingTransactions);
+      final PendingTransactions pendingTransactionsToSave = pendingTransactions;
       pendingTransactions = new DisabledPendingTransactions();
+      return saveRestoreManager.saveToDisk(pendingTransactionsToSave);
     }
+    return CompletableFuture.completedFuture(null);
   }
 
   public boolean isEnabled() {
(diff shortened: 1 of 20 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -121,13 +129,13 @@ private void initLogForReplay() {
         .log();
   }
 
-  public void saveToDisk() {
+  private synchronized void saveToDisk(final PendingTransactions pendingTransactionsToSave) {
```
shemnon, line 132: Why pick two arbitrary methods to place the `synchronized` keyword on.  What fields are they looking to prevent unsynchronized access on?  Do all accesses to those fields get locked?

Consider using `com.google.errorprone.annotations.concurrent.GuardedBy` annotations.

fab-10: I just want that these 2 methods are executed serially, in case of different threads enabling/disabling the pool in rapid succession (that should never happen now).

shemnon: That's my point, we are treating synchronization lightly.  

If they shouldn't be executed concurrently then you should be able to identify the fields that would conflict.  Then we can make sure that no other public method will allow the same conflicts that executing these two methods in parallel would cause.

Consider using `import java.util.concurrent.locks.ReentrantLock` if your goal is to prevent writing and reading at the same time.  This will limit the impact of the lock (not locking on the entire TransactionPool object, which may have unforseen impacts).  It also has timeout-enabled lock acquisitions, which should be considered to ensure shutdown doesn't fully freeze in extreme circumstances, or take an extended time such as if multiple shutdown requests have been initiated.

fab-10: Ok, I see your point, will change to avoid locking the entire class, and will change the write to check and kill if an already running write, same for the read, so there is no unbounded queueing on shutdown and it can finish in few seconds as for my tests with >100k txs. At that point there should be no need to put a timeout but in case its place should be in the `Runner::stop`.

fab-10: Please check the updated implementation

shemnon: Use the timeout variant, with something large like a minute or ten minutes.  A bug in the loading code could prevent orderly shutdown of the entire service. And this is the sort of bug that will sneak in as an unrelated change to some other feature, not that it exists today.

It's better to have a shutdown missing pending transactions than a service that refuses to shutdown.  A non-timeout lock request creates that risk.

fab-10: Managing the timeout on shutdown in `Runner::stop` where all timeout on services are managed then.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -640,4 +574,136 @@ private void onAdded(final Transaction transaction) {
...
+
+    void loadFromDisk() {
+      serializeAndDedupOperation(this::executeLoadFromDisk, readInProgress);
+    }
+
+    private void serializeAndDedupOperation(
+        final Runnable operation,
+        final AtomicReference<CompletableFuture<Void>> operationInProgress) {
+      if (configuration.getEnableSaveRestore()) {
+        diskAccessLock.lock();
```
shemnon, line 599: This call to lock should use `tryLock(long time, TimeUnit unit)` to ensure it doesn't wait forever getting a lock.

shemnon (approved):

PR #5638, opened 2023-06-23
Updates for execution-spec-tests

shemnon:
## PR description

Updates needed to keep executing execution-spec-tests
* add withdrawals support
* add access lists to transactions
* accept t8n and b11r CLI args that are zero length or all whitespace
* Enumerate forks in t8n and b11r help
* remove JSON wrapping from t8n transaction RLP


ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/B11rSubCommand.java
```diff
@@ -162,6 +195,11 @@ public void run() {
           config.set("txs", t8nReader.readTree(reader));
         }
       }
+      if (!withdrawals.equals(stdinPath)) {
+        try (FileReader reader = new FileReader(withdrawals.toFile(), StandardCharsets.UTF_8)) {
+          config.set("withdrawals", t8nReader.readTree(reader));
```
jflo, line 200: I assume using `t8nReader` as a name for the ObjectMapper here is just copypasta?

jflo (approved):

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/B11rSubCommand.java
shemnon: The irony being this was to remove copypasta.  Will fix.

PR #5639, opened 2023-06-24
Update evmtool graalvm build

shemnon:
## PR description

Update EVMTool's GraalVM build
* add new class to reflection config
* move CLI output to be system-err oriented
* make logger in evmtool nop
* exclude some problematic jars from compilation
  - mixed versions of bouncy castle
  - netty initializations of log4j


ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/JsonUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.fasterxml.jackson.core.JsonParser.Feature;
+import com.fasterxml.jackson.core.util.DefaultIndenter;
+import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import org.apache.tuweni.bytes.Bytes;
+
+public class JsonUtils {
```
jflo, line 29: Seems like a good candidate to have Dagger provide an ObjectMapper configured this way.

jflo (approved):

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/JsonUtils.java
shemnon: It would be, but t8n and b11r are not dagger configured (yet).

PR #5640, opened 2023-06-26
Eth/68: treat transaction size as scalar unsigned int

fab-10:
## PR description

Align Besu to what other clients are doing, treating the sizes as an array of scalar ints and not 4 bytes, this also prevent being disconnected from Geth due to `invalid-message` error:

```
DEBUG[06-19|15:15:40.347] Ethereum peer connected                  id=a1697db12e1f2986 conn=inbound name=besu/v23.4.0/linux-x...
DEBUG[06-19|15:15:40.360] Message handling failed in `eth`         id=a1697db12e1f2986 conn=inbound err="invalid message: message msg #8 (157028 bytes): invalid message: (code 8) (size 157028) rlp: non-canonical integer (leading zero bytes) for uint32, decoding into (eth.NewPooledTransactionHashesPacket68).Sizes[0]"
DEBUG[06-19|15:15:40.360] Removing p2p peer                        peercount=10 id=a1697db12e1f2986 duration=12.956ms     req=false err="invalid message: message msg #8 (157028 bytes): invalid message: (code 8) (size 157028) rlp: non-canonical integer (leading zero bytes) for uint32, decoding into (eth.NewPooledTransactionHashesPacket68).Sizes[0]"
```

Contacting the author of the EIP for making it clearer in which data type is used for this message, see https://github.com/ethereum/EIPs/pull/7233

### DIFF
--- a/ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/AbstractRLPInput.java
+++ b/ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/AbstractRLPInput.java
@@ -318,6 +318,12 @@ abstract class AbstractRLPInput implements RLPInput {
   }
 
   @Override
+  public long readUnsignedIntScalar() {
+    checkScalar("unsigned int scalar", 4);
+    return readLongScalar();
+  }
+
+  @Override
   public BigInteger readBigIntegerScalar() {
     checkScalar("arbitrary precision scalar");
     final BigInteger res = getUnsignedBigInteger(currentPayloadOffset, currentPayloadSize);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Bug Fixes
 - Use the node's configuration to determine if DNS enode URLs are allowed in calls to `admin_addPeer` and `admin_removePeer` [#5584](https://github.com/hyperledger/besu/pull/5584)
+- Align the implementation of Eth/68 `NewPooledTransactionHashes` to other clients, using unsigned int for encoding size. [#5640](https://github.com/hyperledger/besu/pull/5640)
 
 ### Download Links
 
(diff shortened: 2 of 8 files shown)


Gabriel-Trintinalia (approved):
LGTM

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/AbstractRLPInput.java
```diff
@@ -317,6 +317,12 @@ public int readIntScalar() {
     return res;
   }
 
+  @Override
+  public long readUnsignedIntScalar() {
+    checkScalar("unsigned int scalar", 4);
```
pinges, line 322: So you are enforcing that there are no leading zeros now! That means that we are disconnecting older Besu nodes that do not have this PR :-)
I guess it is the right thing to do, it is just funny that we will disconnect older besu nodes ...
What we could do is being lenient with the decoding (allowing the leading zeros), but do the right thing with the encoding, at least for a short time ...

pinges (approved):
I've made a suggestion about being lenient (at least for a short while) so we do not disconnect older Besu nodes. 
Otherwise LGTM

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/AbstractRLPInput.java
fab-10: Good point, anyway we are not disconnecting other nodes in case of an error, but just skipping this message, so should be safe

fab-10: indeed you are right, and we disconnect other Besus, so I am implementing a backward compatibility decoding

fab-10: made `ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/encoding/TransactionAnnouncementDecoder.java` lenient when reading size

PR #5641, opened 2023-06-27
Remove BLS12 from cancun

shemnon:
## PR description

BLS12 signatures were speculatively added in cancun in late 2022.  This is no longer accurate.  Move the implementation to the "Future_EIPs" fork so that it will remain testable and not accidentally get included accidentally.  Also add missing fluent APIs.


jflo (approved):
nice cleanup, much improved.

PR #5643, opened 2023-06-27
[4844] CancunGasCalculator ->  Fix Excess Data Gas Calculation

Gabriel-Trintinalia:
## PR description
Fixing the formula according to `calc_excess_data_gas` in https://eips.ethereum.org/EIPS/eip-4844


jflo (changes requested):
I'm surprised none of the unit tests were impacted, which lead me to discover we have no coverage for this.  Can you please add unit tests to cover it?

Gabriel-Trintinalia:
> I'm surprised none of the unit tests were impacted, which lead me to discover we have no coverage for this. Can you please add unit tests to cover it?

done

pinges (approved):
LGTM

Issue #5377, opened 2023-04-21
Investigate and fix the Hive rpc-compat issues


gfukushima:
https://hive.ethpandaops.io/#/test/generic/1759805723-7162953a09ee00461c5ff2121146b945

gfukushima:
Update, after fixing the issue with block import (https://github.com/hyperledger/besu/commit/7fff05a4cc8f463a468ed5b9a557ba316b69c7f3)
RPC-compat is running and current status (Hive built: 02/05/2023, 20:10:11 commit: 972a1801) is:
23/05/2023, 12:56:25	rpc-compat	besu	✕ Fail (16 / 45)

there was a PR to enable ADMIN API which will bring that number down:
https://github.com/ethereum/hive/commit/2e43bacfa72697ec2e31c6d716bdab003a298233

macfarla:
https://github.com/hyperledger/besu/pull/8419

macfarla:
Current failures https://docs.google.com/spreadsheets/d/1UkIJ9Sgj_Bd4AfjSUOmIIq-t2rfzHUbBDSQR_3glEuw/edit?gid=0#gid=0

macfarla:
I’ve made a couple of attempts at addressing some hive test failures this week. But I’ve had to abandon these attempts because I can’t change the behavior for individual RPCs without breaking hive tests in other areas.

In Besu we have an abstract class `AbstractBlockParameterOrBlockHashMethod` that is used by several RPCs - so any RPC that gets a block by block number or hash will have the same error handling behavior. Including:
debug_accountAt, debug_setHead, eth_call, eth_getBlockReceipts, eth_getProof, eth_simulateV1, eth_getBalance, eth_getCode, eth_getStorageAt, eth_getTransactionCount

The spec, and the hive tests, are inconsistent with error handling expectations.

Specific issues I have struck this week:

return null if block not found
so there’s a hive test for getBlockReceipts
but fixing this hive test broke hive tests for eth_simulate that use a future block. Is it expected that we would want different behavior here or should eth_simulate also return null if the requested block is not found?
draft besu PR that fixes 3 hive tests and breaks 2 https://github.com/hyperledger/besu/pull/9303 

also this indicates there’s a gap in rpc-compat hive tests for error handling for the other RPCs in the above list

error code 3 for revert
this PR was merged recently https://github.com/ethereum/execution-apis/pull/600
think this should match execution_reverted code here https://github.com/ethereum/execution-apis/pull/650 
[...]

PR #5647, opened 2023-06-28
Fix getAccountProof for Bonsai

gfukushima:
## PR description
This PR fix the eth_getProof RPC method when using Bonsai by adding the worldStateProof to BonsaiWorldStateProvider.
By passing a snapshot of the ws that we want the proof for the BonsaiWorldStateProvider should be able to make normal usage of the worldStateProof to generate the proofs.

## Fixed Issue(s)

Part of #5377


matkt (changes requested):
seems to be good

do we can add test for bonsai in this class WorldStateProofProviderTest.  the current tests are only using forest

you can use this class as example SnapWorldDownloadStateTest we are using Parameterized to call the same tests with bonsai and forest

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/proof/WorldStateProofProviderTest.java
```diff
@@ -39,21 +44,43 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.junit.runners.Parameterized;
 
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(Parameterized.class)
```
siladu, line 49: nit: would be preferable to use JUnit 5 syntax if possible

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/proof/WorldStateProofProviderTest.java
```diff
@@ -39,21 +44,43 @@
...
   @Before
   public void setup() {
+    if (dataStorageFormat.equals(DataStorageFormat.BONSAI)) {
+      worldStateStorage =
+          new BonsaiWorldStateKeyValueStorage(
+              new InMemoryKeyValueStorageProvider(), new NoOpMetricsSystem());
+    } else {
+      worldStateStorage = new WorldStateKeyValueStorage(new InMemoryKeyValueStorage());
+    }
+
```
siladu, line 83:
```suggestion
    worldStateStorage =
        switch (dataStorageFormat) {
          case BONSAI -> new BonsaiWorldStateKeyValueStorage(
              new InMemoryKeyValueStorageProvider(), new NoOpMetricsSystem());
          case FOREST -> new WorldStateKeyValueStorage(new InMemoryKeyValueStorage());
        };
```

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/proof/WorldStateProofProviderTest.java
```diff
@@ -88,8 +115,13 @@ public void getProofWhenWorldStateAvailable() {
...
     worldStateTrie.put(addressHash, RLP.encode(accountValue::writeTo));
-    worldStateTrie.commit(updater::putAccountStateTrieNode);
-
+    worldStateTrie.commit(
+        (location, hash, value) -> {
+          updater.putAccountStateTrieNode(location, hash, value);
+          if (location.isEmpty()) {
+            updater.saveWorldState(Bytes.EMPTY, worldStateTrie.getRootHash(), value);
+          }
+        });
```
siladu, line 124: Why is this necessary?

matkt:
can we verify this GetProof method (compare with other client ?) from what i see getProof doesn't look good. I worked on this part a long time ago but I have a doubt now

Normally GetProof must return the siblings at each level in order to validate the proof. That doesn't seem like it regarding the code. I implemented a getProof method in another component so we can fix that easily but just wanted to be sure i'm not missing something

IMO it should be L1+L2+X for X

*         o
*        / \
*       o   o (L1)
*       / \
*  (L2)o  x

looking at the code it is L1+L2+X

* (L1)  o
*        / \
* (L2)o   o 
*       / \
*    o  x

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/proof/WorldStateProofProviderTest.java
siladu: Maybe worth a comment explaining this?

gfukushima:
I've tested and we do provide the right proofs but there's seems to be an issue for proofs there are not the latest state in bonsai. Will move this PR to draft until I get back to it.

gfukushima:
Closing this until I get back to it.

PR #5653, opened 2023-06-28
[WIP] Store RLP encoding to avoid encoding operations

ahamlat:
## PR description
The goal of this PR is to store RLP encoding bytes insides different objects  (transaction, transactionReceipt, BlockBody and BlockHeader) to avoid encoding operations when we need the get the RLP encoding.

This PR has also an improvement on Sync process to avoid transaction root validation and receipt root validation twice.

Before this PR
image

After this PR (We can notice that receipts root validation appears only once)
image


ahamlat:
Close to split into two PRs

PR #5656, opened 2023-06-28
Add hooks to AbstractCreateOperation for library users

shemnon:
## PR description

Add hooks for a successful contract create, a failed contract create, and an invalid contact create. Users of the library will be able to customize create responses without having to replace core logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+
 # Changelog
 
 ## 23.4.5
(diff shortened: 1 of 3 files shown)


siladu (commented):
LGTM but would like a second opinion so I will leave for someone closer to this code to approve.

Maybe warrants a changelog item?

shemnon:
Changelog added.  @diega @jflo can you review?

diega (approved):
I'm ok to merge this. I personally would have preferred an interface for a listener, but that's just stylistic and I'm not sure if that would impose some sort of performance drawback

shemnon:
A Listener would have imposed another field to do an indirect dispatch against and another object to preserve identity.

Here's an example of what it will do once new code hits hedera: https://github.com/hashgraph/hedera-services/blob/96ac267a596d611b650ecec98bc29d190e3c85bb/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/operations/CustomCreate2Operation.java#L68C1-L74

The Operations are all loaded in at EVM create time, and live for the duration of the EVM.  Composition over inheritance would make it more interesting.  If more of these hook requests come in for different operations I'll look into a EVM wide listener that can be registered.

One advantage of this setup is that if if there are no custom overrides (like in the Ethereum Mainnet setup) the hotspot compiler will elide the entire call away in time.  (first noticing it's monomrphic it will attempt to inline it, then when it's inlining an empty method it will disappear).

PR #5659, opened 2023-06-30
Only update peer with the first and last block

pinges:
## PR description
We only need to update the peers chain state with the highest block header, which is the first or the last block in the list. Currently we are trying to update with every block header, which can be 200 or so.


Gabriel-Trintinalia (approved):
lgtm

PR #5662, opened 2023-06-30
Extend OperationTracer with new methods

daniellehrner:
## PR description

This PR extends the `OperationTracer` with 2 new methods:

* traceStartTransaction, which complements the existing traceEndTransaction
* isExtendedTracing, which returns `false` by default, but can be overrriden by other tracer implementations

Extended tracing is needed in cases where the current tracing implementations does not provide sufficient insight into a transaction. This is for example required for zkEVM implementations as those often require a more detailed insight into a transaction for the proof generation. This PR introduces extended tracing in case of a successful contract creation. More extending tracing will be added in future PRs.

Further the plugin API for tracing is extended as well. It adds 2 new methods:

* traceStartBlock
* traceEndBlock

Both take a block header and a block body as a parameter. It was required to use those 2 instead of a `Block` object, because the latter is causing circular dependency issues that would require a very big refactoring to resolve them.


evm/src/main/java/org/hyperledger/besu/evm/processor/ContractCreationProcessor.java
```diff
@@ -175,6 +175,9 @@ public void codeSuccess(final MessageFrame frame, final OperationTracer operatio
             contractCode.size(),
             frame.getRemainingGas());
         frame.setState(MessageFrame.State.COMPLETED_SUCCESS);
+        if (operationTracer.isExtendedTracing()) {
```
shemnon, line 178: Why only in extended?  Why not log this account creation all the time?

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/tracer/ExtendedOperationTracer.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.plugin.services.tracer;
+
+import org.hyperledger.besu.evm.tracing.OperationTracer;
+import org.hyperledger.besu.plugin.data.BlockBody;
+import org.hyperledger.besu.plugin.data.BlockHeader;
+
+public interface ExtendedOperationTracer extends OperationTracer {
```
shemnon, line 21: Since all the interfaces are block related I would recommend getting block in the name.  Perhaps `BlockAwareOperationTracer.`

daniellehrner: Yes, that is a good name. I've changed it.

evm/src/main/java/org/hyperledger/besu/evm/processor/ContractCreationProcessor.java
daniellehrner: It was not logged before, only in the failure cases. So it would change the existing behavior for anybody using the current tracer implementation. I think it makes sense to always log it though. So we might just change it to always log it.

evm/src/main/java/org/hyperledger/besu/evm/tracing/OperationTracer.java
```diff
@@ -63,6 +63,13 @@ default void tracePrecompileCall(
   default void traceAccountCreationResult(
       final MessageFrame frame, final Optional<ExceptionalHaltReason> haltReason) {}
 
+  /**
+   * Trace the end of a transaction.
```
macfarla, line 67:
```suggestion
   * Trace the start of a transaction.
```

besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
```diff
@@ -114,6 +117,9 @@ private void trace(final Block block, final OperationTracer tracer) {
                                 maybeParentHeader
                                     .flatMap(BlockHeader::getExcessDataGas)
                                     .orElse(DataGas.ZERO));
+
+                    tracer.traceStartTransaction(transaction.getGasLimit());
```
delehef, line 121: Could we get the to/from addresses here?

shemnon: Would it be easier to pass in the Transaction object as the argument and let the tracer implementations pick out gas limit and to/from?

delehef: Not only easier, but also probably smarter :)

daniellehrner: I did the change and now the parameter is a `Transaction` object. I needed to move the around the interface of the transaction, which causes a lot of changes in different files. But only the imports are changing, nothing else.

evm/src/main/java/org/hyperledger/besu/evm/processor/ContractCreationProcessor.java
daniellehrner: All the current tests are passing, so there is no problem in general. Do you think it makes sense to log it all the time? I am not sure why we did not do in the first place. But that might just have been an oversight.

macfarla (approved):
LGTM

shemnon (approved):
I'm hesitant to move Transaction into datatypes, but I cannot think of a better alternative that preserves the ability to be concise on method parameters.  I am also seeing some advantage of bring the notion of a transaction in a stronger way to the EVM, so there are pluses to be sure.

So with hesitation, LGTM.

PR #5664, opened 2023-07-02
[4844] Read ExcessDataGas and ExcessDataGas from genesis if Cancun is present

Gabriel-Trintinalia:
## PR description
Fix for genesis mismatch


spencer-tb:
We recently added the hive pyspec simulator to https://hivecancun.ethdevops.io/, which runs the tests from https://github.com/ethereum/execution-spec-tests using the EngineAPI 

In order to start running the tests the genesis state needs to include the `excessDataGas` and `dataGasUsed`, otherwise we get a genesis mismatch -> leading to a `SYNCING` status when adding the first block with `NewPayloadV3` via the EngineAPI, as opposed to a `VALID` or expected `INVALID` status.

Gabriel-Trintinalia:
Closed in favour of #5677

Issue #5528, opened 2023-06-02
Minimun gas price ignored


DomingoMr:
### Description
I want to set the minimum gas price to 0 so I can deploy smart contracts without problems.

### Acceptance Criteria
Parameter `--min-gas-price` not ignored

### Steps to Reproduce (Bug)
1. Use the command npx `quorum-dev-quickstart` to run a local besu network (`besu version` 22.10.3).
2. Deploy a smart contract to this local network.

**Expected behavior:**
To deploy the smart contract.

**Actual behavior:**
I get this error:

`ProviderError: Gas price below configured minimum gas price
    at HttpProvider.request (/home/domingo/token-erc20/node_modules/hardhat/src/internal/core/providers/http.ts:78:19)
    at LocalAccountsProvider.request (/home/domingo/token-erc20/node_modules/hardhat/src/internal/core/providers/accounts.ts:182:36)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at EthersProviderWrapper.send (/home/domingo/token-erc20/node_modules/@nomiclabs/hardhat-ethers/src/internal/ethers-provider-wrapper.ts:13:20)`

If I look for Besus logs:

`{"timestamp":"2023-06-02T13:06:21,722","container":"39d421c9fa71","level":"INFO","thread":"main","class":"Besu","message":"Starting Besu","throwable":""}
{"timestamp":"2023-06-02T13:06:24,125","container":"39d421c9fa71","level":"WARN","thread":"main","class":"Besu","message":"--min-gas-price has been ignored because --miner-enabled was not defined on the command line.","throwable":""}
{"timestamp":"2023-06-02T13:06:27,088","container":"39d421c9fa71","level":"INFO","thread":"main","class":"Besu","message":"
`

As you can see, `--min-gas-price` is ignored. I tried to put in the config file this instruction `miner-enabled=true` but the network does not work.

**Frequency:**
Always I try to deploy a smart contract

### Versions (Add all that apply)
* Software version: [`22.10.3`]
* Java version: [`17`]
* OS Name & Version: [`WSL2`]

non-fungible-nelson:
@matthew1001 this is similar to the config cleanup you were working on? Or did I just recall you hitting the same hurdle 😅

matthew1001:
I certainly had to use a slightly odd combination of settings (the discord discussion is here -> https://discord.com/channels/905194001349627914/938504958909747250/1110826605447753759)

[...]

PR #5669, opened 2023-07-04
Add genesis configuration isPoa() convenience function, and use it in validation of `--miner-enabled` option

matthew1001:
## PR description
This PR indirectly relates to issue https://github.com/hyperledger/besu/issues/5528.

The intention of the PR is to change the validation logic of the `--miner-enabled` option, particularly with regard to `--min-gas-price` option which is a common requirement for private POA chains.

Specifically, in the case where a POA consensus algorithm has been configured (`qbft`, `ibft2`, `ibft`, or `clique`) validation of options that would otherwise require `--miner-enabled` to be set, is skipped.

## Fixed Issue(s)
No specific issue is open for this.


config/src/main/java/org/hyperledger/besu/config/GenesisConfigOptions.java
```diff
@@ -65,6 +65,13 @@ public interface GenesisConfigOptions {
    */
   boolean isClique();
 
+  /**
+   * Is poa boolean.
```
siladu, line 69:
```suggestion
   * Is a Proof of Authority network.
```

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2133,7 +2134,10 @@ private void issueOptionWarnings() {
             "--remote-connections-max-percentage"));
 
     // Check that block producer options work
-    if (!isMergeEnabled()) {
+    if (!isMergeEnabled()
+        && !(Optional.ofNullable(genesisConfigOptions)
+            .orElseGet(() -> new StubGenesisConfigOptions())
+            .isPoa())) {
```
siladu, line 2140: Wonder if we can get away with `!isMergeEnabled && isEthHash` instead?

Note, currently public networks genesis' are either clique or ethash, but future public testnets (proof of stake at genesis) might have the consensus unspecified. These networks shouldn't care about miner-enabled either I think.

siladu: Had a play and I think that `!isMergeEnabled && isEthHash` is what we want for both checks, i.e. the `miner-stratum-enabled` one underneath this (I think it's a bug it doesn't have `!isMergeEnabled`)

PoS at genesis networks that omit the consensus engine will default to "unknown" in the besu code, so it works for them too.

Maybe eventually it can simply be `isEthHash` but might have to always support a "merged" mainnet config (where it was originally ethash).

siladu: ...so yes, the isPoa is probably redundant for this, though potentially useful for other things at some point?

matthew1001: Thanks @siladu yes for this case I think `isEthHash()` makes sense. Maybe it's worth leaving `isPoa()` as a general utility that could help make future code changes/logic a little clearer. I'll submit a new commit that changes the current logic but leaves `isPoa()` in. Let me know what you think.

matthew1001: Realised that `getActualGenesisConfigOptions()` provides the combined output of a genesis file, network config, + overrides, so no refactoring required.

matthew1001: @siladu I think I'm happy with the current shape of the PR:

- switched to using `isEthash()`
- moved `--miner-stratum` option warning check inside same block as `--miner-enabled`
- left `isPoa()` in as a helper function
- added some new unit tests for QBFT & IBFT2 option checking

Ready for you to take another look at when you have a mo.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2133,7 +2133,7 @@ private void issueOptionWarnings() {
             "--remote-connections-max-percentage"));
 
     // Check that block producer options work
-    if (!isMergeEnabled()) {
+    if (!isMergeEnabled() && getActualGenesisConfigOptions().isEthHash()) {
```
siladu, line 2136: Nice! Didn't know about this one.

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -3684,7 +3698,7 @@ public void stratumMiningIsEnabledWhenSpecified() throws Exception {
   @Test
   public void stratumMiningOptionsRequiresServiceToBeEnabled() {
 
-    parseCommand("--miner-stratum-enabled");
+    parseCommand("--network", "dev", "--miner-stratum-enabled");
```
siladu, line 3701: Why was this needed?

siladu (approved):
Looks good!

siladu:
@matthew1001 Do you have merge permission or need a maintainer to merge it?

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
matthew1001: I needed to modify the test in a way that would ensure it was being run with a) `merge` enabled and b) `ethash` enabled, so that the warning message it was testing for would be produced. The default network is `mainnet.json` which doesn't have `ethash` enabled but `dev.json` does so I ended up going with that.

matthew1001:
> @matthew1001 Do you have merge permission or need a maintainer to merge it?

No not yet @siladu, still working towards maintainer status so you or one of the other maintainers will need to merge it please.

PR #5670, opened 2023-07-04
Test updates for cancun execution-spec-tests

shemnon:
## PR description

Update t8n tool to support new fields (data gas related) for Cancun in the execution-spec-tests,
also, refactor so the t8n server benefits from the changes.

### DIFF
--- a/ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/t8n/cancun-blobs-per-tx.json
+++ b/ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/t8n/cancun-blobs-per-tx.json
@@ -6,66 +6,85 @@
     "--input.env=stdin",
     "--output.result=stdout",
     "--output.alloc=stdout",
-    "--output.body=stdout",
-    "--state.fork=Shanghai",
+    "--output.body=txs.rlp",
+    "--state.fork=Cancun",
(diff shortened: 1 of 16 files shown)


ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/t8n/cancun-blobs-per-tx.json
```diff
@@ -0,0 +1,105 @@
+{
+  "cli": [
+    "t8n",
+    "--input.alloc=stdin",
+    "--input.txs=stdin",
+    "--input.env=stdin",
+    "--output.result=stdout",
+    "--output.alloc=stdout",
+    "--output.body=txs.rlp",
+    "--state.fork=Cancun",
```
jflo, line 10: If we're on Cancun, why do we expect the transaction to fail for being an unsupported type?

jflo (approved):

ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/t8n/cancun-blobs-per-tx.json
shemnon: I have a much more involved fork addressing 4844 integrations.

PR #5673, opened 2023-07-05
Introduce transaction validator interface

fab-10:


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
```diff
@@ -472,10 +465,6 @@ private boolean isTransientValidationError(final TransactionInvalidReason invali
         || invalidReason.equals(TransactionInvalidReason.NONCE_TOO_HIGH);
   }
 
-  private boolean isIncorrectNonce(final ValidationResult<TransactionInvalidReason> result) {
```
siladu, line 475: Why is this being removed? Is the functionality happening elsewhere?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/PermissionTransactionFilter.java
```diff
@@ -11,11 +12,12 @@
  * specific language governing permissions and limitations under the License.
  *
  * SPDX-License-Identifier: Apache-2.0
+ *
  */
 package org.hyperledger.besu.ethereum.core;
 
 @FunctionalInterface
-public interface TransactionFilter {
+public interface PermissionTransactionFilter {
```
siladu, line 20: 👍

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockTransactionSelector.java
fab-10: yes it has been generalized, now every invalid reason is recorded in the `TransactionSelectionResults.invalidTransactions` map, not only for incorrect nonce, and the caller have full access to all invalid transactions

siladu (approved):
LGTM

PR #5675, opened 2023-07-05
Upgrade BouncyCastle libraries

shemnon:
## PR description

Upgrade bouncy castle to v1.75.  This involved a change in maven
coordinates for other modules.


ethereum/api/build.gradle
```diff
@@ -125,7 +125,7 @@ dependencies {
 
 artifacts { testSupportArtifacts testSupportJar }
 
-task generateTestBlockchain() {
+tasks.register('generateTestBlockchain') {
```
fab-10, line 128: not a gradle expert, is this just a code style change or has it also other impacts?

fab-10 (approved):

ethereum/api/build.gradle
shemnon: IntelliJ keeps tagging it as a lint issue.

PR #5682, opened 2023-07-06
Introduce transaction validator interface (phase 2)

fab-10:
## PR description

The focus of this PR is to complete the work started in #5673 and remove the odd `setPermissionTransactionFilter` from the interface.
The `setPermissionTransactionFilter` method is only relevant for private network when [permissioning](https://besu.hyperledger.org/stable/private-networks/concepts/permissioning) is enabled, so it make sense to have a dedicated validator for that, so `PermissionTransactionValidator` has been introduced to handle that use case, while it delegates all the other validations to `MainnetTransactionValidator`.
To obtain that, as common, another layer of indirection was required, since it is not possible to create the `PermissionTransactionValidator` when building the protocol schedule, due to a cyclic dependency, so the need for `TransactionValidatorFactory` that is available through the protocol schedule, and for efficiency create and memoize the actual transaction validator on first usage when all the required prerequisite, to build any kind of transaction validator, are satisfied.


macfarla (approved):
LGTM

PR #5686, opened 2023-07-07
Bonsai based reference test worldstate

garyschulte:
## PR description
@shemnon pointed out that the reference tests support in besu does not use bonsai worldstate, nor does evmtool.  This PR creates a bonsai-based worldstate and accumulator for ReferenceTestsWorldState.

### Change Summary:
* creates ReferenceTestWorldState interface, BonsaiReferenceTestWorldState and DefaultReferenceTestWorldState impl's
* creates preimageProxy functionality for reference test worldstate
* implements account storage streaming for bonsai
* adds overload of streamFromKey that includes endKey
* implements iterator based streamFromKey for BonsaiSnapshotWorldstate
* removes archive from BonsaiWorldState
* since bonsai checks calculated state root against blockheader stateroot:
  * adds currentStateRoot to ReferenceTestEnv, adds stateRoot to evmtool t8n test cases 
  * rewrites constructed block headers from GeneralStateState reference tests with the expected stateroot from post

### Discovered/Fixed Defects
* fixes premature stream end with streamStorageFlatDatabase for unsorted storage
* fixes account storage clearing issue 

## Fixed Issue(s)

#5688


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -39,21 +44,29 @@
...
+import org.apache.tuweni.rlp.RLP;
+import org.apache.tuweni.units.bigints.UInt256;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("unused")
 public class BonsaiWorldStateKeyValueStorage implements WorldStateStorage, AutoCloseable {
-
+  Bytes32 BYTES32_MAX_VALUE =
+      Bytes32.fromHexString("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
```
matkt, line 69: this one can be final and static

also maybe we can use `UInt256.MAX_VALUE` or `Bytes32.repeat((byte) 0xff) `

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -223,7 +218,7 @@ private void updateTheAccounts(
       final BonsaiAccount updatedAccount = bonsaiValue.getUpdated();
       try {
         if (updatedAccount == null) {
-          final Hash addressHash = Hash.hash(accountKey);
+          final Hash addressHash = preImageProxy.hashAndSavePreImage(accountKey);
```
matkt, line 221: it's clearly too much modification just for testing. I will be happy if we have a way to avoid that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
```diff
@@ -127,14 +132,18 @@ public EvmAccount createAccount(final Address address, final long nonce, final W
       bonsaiValue = new BonsaiValue<>(null, null);
       accountsToUpdate.put(address, bonsaiValue);
     } else if (bonsaiValue.getUpdated() != null) {
-      throw new IllegalStateException("Cannot create an account when one already exists");
+      if (bonsaiValue.getUpdated().isEmpty()) {
```
matkt, line 135: maybe 

```java
else if (bonsaiValue.getUpdated() != null && bonsaiValue.getUpdated().isEmpty()) {

    return new WrappedEvmAccount(track(new UpdateTrackingAccount<>(bonsaiValue.getUpdated())));

} else {

    throw new IllegalStateException("Cannot create an account when one already exists");

}

```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
```diff
@@ -290,6 +299,19 @@ public void commit() {
               final BonsaiAccount updatedAccount;
               final BonsaiValue<BonsaiAccount> updatedAccountValue =
                   accountsToUpdate.get(updatedAddress);
+
+              final Map<StorageSlotKey, BonsaiValue<UInt256>> pendingStorageUpdates =
```
matkt, line 303: we need to fullsync some nodes to be sure we are not adding a consensus issue 
(goerli, sepolia, and maybe etc at the minimum)

ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/state-test/blockhash.json
```diff
@@ -61,7 +61,7 @@
       "post": {
         "Shanghai": [
           {
-            "hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+            "hash": "0xd14c10ed22a1cfb642e374be985ac581c39f3969bd59249e0405aca3beb47a47",
```
matkt, line 64: is it normal to change that ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -79,15 +92,25 @@ public class BonsaiWorldStateKeyValueStorage implements WorldStateStorage, AutoC
 
   protected final Subscribers<BonsaiStorageSubscriber> subscribers = Subscribers.create();
 
+  final BonsaiPreImageProxy preImageProxy;
```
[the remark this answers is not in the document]
garyschulte: I think we could create a subclass of BonsaiWorldStateKeyValueStorage that takes the preImageProxy, since we are only using it to recover hashes for reference test worldstate.  Similarly we could move streamAccounts into the subclass.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/FlatDbReaderStrategy.java
```diff
@@ -158,7 +158,7 @@ public Map<Bytes32, Bytes> streamStorageFlatDatabase(
             .streamFromKey(
                 ACCOUNT_STORAGE_STORAGE,
                 Bytes.concatenate(accountHash, startKeyHash).toArrayUnsafe())
-            .takeWhile(pair -> Bytes.wrap(pair.getKey()).slice(0, Hash.SIZE).equals(accountHash))
+            .filter(pair -> Bytes.wrap(pair.getKey()).slice(0, Hash.SIZE).equals(accountHash))
```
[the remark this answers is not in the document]
garyschulte: I think the right answer is to add the filter above the takeWhile rather than replace with it.  That won't impact performance and it will ensure that unsorted flatDb storage doesn't prematurely end the stream (like in the case of layered)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldState.java
garyschulte: I will look at deferring to a hash() method we can call in BonsaiWorldState that we can override in the reference test worldstate with a preimageProxy.  Before I do I want to collect some cache hit metrics and perf numbers on mainnet with an in-memory preImage proxy to see if using the preimage proxy would be beneficial vs hashing every time.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/FlatDbReaderStrategy.java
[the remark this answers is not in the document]
garyschulte: I ended up creating streamFromKey(start, end), that way each storage impl can do what is best for its underlying data.  Also I noticed the snapshot streamFromKey was not efficient and switched to an iterate.seek() impl

ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/state-test/blockhash.json
garyschulte: yes, bonsai cares about state root, whereas forest does not.  This value was just a placeholder afaik.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldState.java
garyschulte: In BonsaiWorldState and the accumulator I added a hashAndSavePreImage function that by default does not save pre-images and just override this in the referencetest worldstate.  The only tradeoff here is that I had to make the accumulator mutable via a protected method.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
garyschulte: I snap synced a goerli node a few days ago and it has been going well.  I will start a full sync on goerli and another on sepolia and let them run over the weekend

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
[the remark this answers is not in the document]
matkt: seems to be good like that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java
matkt: could you try to run a mainnet snapsync with the heal flat db flag ? I see that this PR is modifying a lot the stream of the flat db datas and it will be nice if can be sure that it's ok

matkt:
Seems to be good. if we can just check a snapsync with the heal flat db flag it will be nice

garyschulte:
> Seems to be good. if we can just check a snapsync with the heal flat db flag it will be nice

I am snapsyncing mainnet over the weekend, but I think it is experiencing the 99% error :/

maybe I can rebase off of your heal PR

matkt:
> > Seems to be good. if we can just check a snapsync with the heal flat db flag it will be nice
> 
> I have been snapsyncing mainnet over the weekend, but I think it is experiencing the 99% error :/ The good news is that sepolia completed a full sync without issue. Goerli is at about 3 million blocks behind, but appears it is going to reach head without error. Just the mainnet snap sync with flatdb heal is outstanding.
> 
> maybe I can rebase off of your heal PR

I think it's a good idea. After the sync I think I will approve. Everything seems to be fine now

garyschulte:
> I think it's a good idea. After the sync I think I will approve. Everything seems to be fine now

The snap sync of mainnet was good.  full sync of goerli and sepolia likewise. 👍

PR #5693, opened 2023-07-11
Do not leak references to PendingTransactions

fab-10:
## PR description

This PR complete the work done in #5634, basically since the latter is creating and modifying the `pendingTransactions` object at runtime, references to that field outside the txpool, must not be allowed anymore, otherwise object that cached a reference to `pendingTransactions` could see an uninitialized or stale object.

With this PR it is not possible anymore to get references directly to the `pendingTransactions` object and all the interaction previously done with it, now go through `TransactionPool` that manages that lifecycle of the `pendingTransactions` object.

Do not be scared by the number of changes in this PR, since the it is mostly about using `TransactionPool` instead of `PendingTransactions` and unit tests adjustments.

## Fixed Issue(s)

fix an issue introduced by #5634 that causes block proposals with 0 transactions.


ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/LondonFeeMarketBlockTransactionSelectorTest.java
```diff
@@ -86,21 +107,22 @@ public void eip1559TransactionCurrentGasPriceLessThanMinimumIsSkippedAndKeptInTh
             Wei.ZERO,
             MIN_OCCUPANCY_80_PERCENT);
 
-    // tx is willing to pay max 6 wei for gas, but current network condition (baseFee == 1)
+    // tx is willing to pay max 7 wei for gas, but current network condition (baseFee == 1)
```
macfarla, line 110: this one is still 6, no?

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/LondonFeeMarketBlockTransactionSelectorTest.java
```diff
@@ -143,44 +165,29 @@ public void eip1559LocalTransactionCurrentGasPriceLessThanMinimumIsSelected() {
     // tx is willing to pay max 6 wei for gas, but current network condition (baseFee == 1)
```
macfarla, line 165:
```suggestion
    // tx is willing to pay max 7 wei for gas, but current network condition (baseFee == 1)
```

macfarla (approved):
couple of nits, otherwise LGTM

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/LondonFeeMarketBlockTransactionSelectorTest.java
fab-10: right

PR #5699, opened 2023-07-13
Update worldstate state change behavior in case of FCU 

matkt:
## PR description

We have identified a potential issue with the FCU mechanism. Based on the logs, it seems that Besu considers each new head as a reorganization (reorg) because of this line https://github.com/hyperledger/besu/blob/main/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java#L610 so we are not updating the worldstate if the new head is not a direct descendant of the current head, which is rare in general. As a result, we enter the "rewindToBlock" method https://github.com/hyperledger/besu/blob/main/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java#L633, which performs a chain reorg (in this case, a move forward of the chain) without changing the world state. However, with Prysm, we only encounter this case and keep advancing the chain without advancing the world state. Consequently, with each "engineNewPayload" call, we need to apply a larger number of trie logs to process the new block, which leads to a node crash.

Upon examining the code, it is indeed not normal to have a scenario where we change the blockchain without changing the world state.


ahamlat:
I conducted a test on the mainnet using this PR and version 23.4.4. During the experiment, I halted Prysm on both instances for a duration of 2 hours. Upon observation, the node operating with this PR successfully recovered, whereas the node running version 23.4.4 encountered a continuous decline in performance due to the lack of world state updates. Consequently, the node eventually crashed

image

I also tested different Consensus clients with different scenarios 

- Initial Sync (performs backwards sync as last stage)
- Scenario 1: CL restart
- Scenario 2: Besu restart
- Scenario 3: CL short downtime: 60 mins
- Scenario 4: CL long downtime: more than 20 hours
- Scenario 5: Besu short downtime: 60 mins
- Scenario 6: Besu longer downtime: more than 20 hours

Scenario/CL | Nimbus | Teku | Prysm | Lighthouse | Lodestar
-- | -- | -- | -- | -- | --
Initial Sync | ✅ | ✅ | ✅ | ✅ | ✅
Scenario 1 | ✅ | ✅ | ✅ | ✅ | ✅
Scenario 2 | To do | ✅ | ✅ | ✅ | ✅
Scenario 3 | ✅ Backward sync | ✅ | ✅ | ✅ | ✅
Scenario 4 | ✅ Backward sync | ✅ | ✅ | ✅ | ✅
Scenario 5 | ✅ Backward sync | ✅ | ✅ | ✅ | ✅
Scenario 6 | Ongoing test | ✅ Backward sync | ✅ | ✅ Backward sync | ✅

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -458,13 +458,16 @@ private boolean canRetryBlockCreation(final Throwable throwable) {
   @Override
   public Optional<BlockHeader> getOrSyncHeadByHash(final Hash headHash, final Hash finalizedHash) {
     final var chain = protocolContext.getBlockchain();
+    final var chainHead = chain.getChainHeadBlockNumber();
     final var maybeHeadHeader = chain.getBlockHeader(headHash);
 
-    if (maybeHeadHeader.isPresent()) {
-      LOG.atDebug()
+    if (maybeHeadHeader.isPresent()
+        && Math.abs(maybeHeadHeader.get().getNumber() - chainHead) < 500) {
```
garyschulte, line 465: this is only necessary for bonsai, perhaps we can defer to the WorldStateArchive to provide this limit?  eventually an archive-friendly version of bonsai would not need this also

garyschulte: if we use backward sync mechanism as an asynchronous worldstate move, we should probably make backward sync aware of the fact that we already have those blocks.  That is an optimization though and we can do that in a subsequent PR.

garyschulte (commented):
This looks good to me, but we should call out explicitly that we are using backward sync as an asynchronous way of moving head.  We would probably benefit from having an explicit behavior for this whether it is part of backward sync service or not.  As it is, when we encounter this condition, I believe we are going to unnecessarily download blocks we already have in blockchain storage.

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
gfukushima, line 465: Can we replace the number here for a named constant to describe where it comes from?

matkt: Good Catch. Indeed I used this number for the tests but we must definitely use the Max Layer that we have for bonsai

matkt: https://github.com/hyperledger/besu/blob/ec6a6b1b73751575e3657eff2b837d7083071f37/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/AbstractTrieLogManager.java#L42

gfukushima: isn't that configurable though?

matkt: Yes there is a flag to change this value

garyschulte: if we defer to a WorldStateArchive method that provides this limit, we can skip the behavior for forest since it should not have a worldstate move issue.

matkt: I removed this part. I will just keep the worldtsate head update modification https://github.com/hyperledger/besu/pull/5699/files#diff-6da2d65dd7a0564527e8ce3ad7c4482f599fb2af30e37316c47854e5aed2e977R610 . I'm doing new test now with this last version

matkt: I removed this modification and reduced the scope of this PR. normally it will be enough to fix the problem (testing now)

matkt:
I conducted the same test as @ahamlat  on the mainnet using this PR (after my last modification) and version 23.7.1

- Initial Sync (performs backwards sync as last stage)
- Scenario 1: CL restart
- Scenario 2: Besu restart
- Scenario 3: CL short downtime: 60 mins
- Scenario 4: CL long downtime: more than 24 hours
- Scenario 5: Besu short downtime: 60 mins
- Scenario 6: Besu longer downtime: more than 24 hours

Scenario/CL | Nimbus | Teku | Prysm | Lighthouse | Lodestar
-- | -- | -- | -- | -- | --
Initial Sync | ✅ | ✅ | ✅ | ✅ | ✅
Scenario 1 |   |   |  |  | 
Scenario 2 |   |   |  |  | 
Scenario 3  |   |   |  |  | 
Scenario 4  |   |   |  |  | 
Scenario 5  |   |   |  |  | 
Scenario 6  |   |   |  |  |

garyschulte (approved):
This makes much more sense than the prior behavior.  I don't think we should have been doing a chain reorg without moving the worldstate.

🚢

gfukushima (approved):

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -607,33 +607,30 @@ private boolean setNewHead(final MutableBlockchain blockchain, final BlockHeader
...
-        .setMessage("New head {} is a chain reorg, rewind chain head to it")
+        .setMessage("Failed to move the worldstate forward to hash {}, not moving chain head")
         .addArgument(newHead::toLogString)
         .log();
-    return blockchain.rewindToBlock(newHead.getHash());
+    return false;
   }
 
-  private boolean forwardWorldStateTo(final BlockHeader newHead) {
+  private boolean moveWorldStateTo(final BlockHeader newHead) {
```
gfukushima, line 633: This rename makes me feel that we could use this method to move the ws "backwards", is that correct?

matkt: yes we can do rollback and rollfoward with this code

Issue #5576, opened 2023-06-10
Besu randomly crashes and stops syncing


yuri-moens:
**Expected behavior:** It works

**Actual behavior:** It doesn't work

**Frequency:** About 1-2 hours after restarting Besu

### Logs (if a bug)
```
2023-06-10 01:18:12.790+00:00 | vert.x-worker-thread-0 | INFO  | AbstractEngineNewPayload | Imported #17,446,465 / 127 tx / 16 ws / base fee 18.07 gwei / 8,465,311 (28.2%) gas / (0xe35c859e2798ebcc951e3743215dc69a0d60c04047c1c86bb646bdbbde47335b) in 0.600s. Peers: 50
2023-06-10 01:18:24.763+00:00 | vert.x-worker-thread-0 | INFO  | DefaultSynchronizer |
####################################################################################################
#                                                                                                  #
# Besu has identified a problem with its worldstate database.                                      #
# Your node will fetch the correct data from peers to repair the problem.                          #
# Starting the sync pipeline...                                                                    #
#                                                                                                  #
####################################################################################################
2023-06-10 01:18:24.765+00:00 | vert.x-worker-thread-0 | INFO  | TransactionPoolFactory | Disabling transaction handling during re-sync
2023-06-10 01:18:24.769+00:00 | vert.x-worker-thread-0 | WARN  | BonsaiWorldStateProvider | Invalid slot found for account 0xdfde78d2baec499fe18f2be74b6c287eed9511d7 at location 0x0c0c
2023-06-10 01:18:25.011+00:00 | vert.x-worker-thread-0 | INFO  | DefaultSynchronizer | Starting synchronizer.
2023-06-10 01:18:25.012+00:00 | vert.x-worker-thread-0 | INFO  | FastSyncDownloader | Starting sync
2023-06-10 01:18:25.058+00:00 | nioEventLoopGroup-3-5 | INFO  | SnapWorldStateDownloader | Downloading world state from peers for pivot block 17445216 (0x3643700ec320aa10c77861639c27734fd3f82364c028f527686720dac70b0635). State root 0x3eeffb63b373cab4a1f201e96f6d3a23b9ca27973860b68f1c3857055f669267 pending requests 0
2023-06-10 01:18:26.006+00:00 | nioEventLoopGroup-3-5 | INFO  | SnapWorldDownloadState | Running world state heal process from peers with pivot block 17445216
[...]

PR #5700, opened 2023-07-13
Promote segmented storage

### LINKED ISSUE #5362: Clean shutdown
### Description
Shutdowns with SEGFAULT and OOM can cause corrupted databases and rocksDB BUSY/LOCK. 

Need to gather information on what areas of improvement are available to us. 

To-Do:
- [ ] Look into rocksDB shut-down with atomic transactions for flatDB and trie **OR**
- [ ] Determine if we can commit some/all column families at the same time (see [here](https://github.com/hyperledger/besu/issues/5362#issuecomment-1594812509))

garyschulte:
## PR description
At a high level, this PR promotes SegmentedKeyValueStorage into the plugin-api for storage.  

The primary benefit of this besides addressing tech debt is that we can directly use segmented storage and atomically commit across multiple segments.  This is mainly beneficial for bonsai worldstate, which comprises four separate key value store segments.  

This is motivated by "unclean" worldstate storage that can result from besu app or the host machine crashing for a variety of reasons.  RocksDB handles unclean shutdowns very well, but if the besu application was in the midst of a worldstate commit, each of the individual worldstate storages might be in different states of commit/rollback at the time of the crash.  By having composed storage, we can rely on the guarantees of rocksdb to atomically commit or rollback changes in the event of a crash or halt, like what resulted in #5576 

## Fixed Issue(s)

fixes #5362 #5576


ethereum/core/src/test/java/org/hyperledger/besu/ethereum/worldstate/MarkSweepPrunerTest.java
```diff
@@ -147,7 +147,8 @@ public void sweepBefore_shouldSweepStateRootFirst() {
     stateRoots.forEach(
         stateRoot -> {
           final InOrder thisRootsOrdering = inOrder(hashValueStore, worldStateStorage);
-          thisRootsOrdering.verify(hashValueStore).remove(stateRoot);
+          // disable this validation since the underlying storage no longer directly uses .remove()
+          // thisRootsOrdering.verify(hashValueStore).remove(stateRoot);
```
garyschulte, line 151: looking for feedback here - do we care to jump through hoops to keep the pruning tests passing when pruning is a deprecated feature?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright ConsenSys AG.
```
daniellehrner, line 2: The copyright should be changed to Hyperledger Contributors

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchiveTest.java
```diff
@@ -15,272 +15,279 @@
...
-import org.junit.Before;
-import org.junit.Ignore;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.Mock;
-import org.mockito.Mockito;
-import org.mockito.junit.MockitoJUnitRunner;
-
-@RunWith(MockitoJUnitRunner.class)
+// import static org.assertj.core.api.Assertions.assertThat;
```
daniellehrner, line 18: Everything is this file is commented out. So it should be safe to delete it. Or are those tests only temporarily commented out?

daniellehrner (changes requested):

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateArchiveTest.java
garyschulte: Thanks, I missed this one since it was collapsed in the default view.  refactored and re-added.

daniellehrner (approved):

PR #5701, opened 2023-07-13
Changes to allow evmtool t8n-server to work with execution-spec-tests

shemnon:
An omnibus of minor changes needed for t8n-server to work with the EFs new execution-spec-tests framework

* Reduce logging output
* Fix json library mismatch between t8n and t8n-server
* Add hook to enumerate supported forks
* temporarily map Shanghai+6780 to Cancun
* add to main distro under 'evmtool' name


shemnon:
Docs change would include this section: https://besu.hyperledger.org/stable/public-networks/reference/evm-tool
* point out the tool is now standard to the distribution
* the name is now `evmtool` to not conflict with Geth's `evm` command.
   * We do aim to be CLI compatible with Geth's `evm` command so bugs pointing out incompatibilities are welcome.
* For the sub-commands (`block-builder`/`b11r`, `code-validate`, `state-test`, `transition`/`t8n`, and `t8n-server) I don't think extensive documentation is needed.  Point to https://ethereum-tests.readthedocs.io/en/develop/t8ntool-ref.html and https://ethereum.github.io/execution-spec-tests/ .  These sub commands exist solely to serve those testing code bases and are not meant for typical user interactions.

danceratopz:
I can confirm this is working nicely with ethereum/execution-spec-tests/pull/206 apart from some tests failing to fill. There seems to be 2 sources of fails, one is due to the incorrect chain id being specified and the other is to do with tx playback protection. I think both of these can (and perhaps even should) be remedied on our side, will let you know early next week once I've taken a closer look.

danceratopz:
So, on our side, I've fixed whether playback protection is enabled/disabled in some of the test cases. I pushed these to ethereum/execution-spec-tests/pull/206.

However, as we don't set the `protected` field in transactions any more, it would be helpful, if we could check for playback protection using `v`. I made a suggestion in shemnon/besu/pull/14.

ethereum/execution-spec-tests/pull/206/ (https://github.com/ethereum/execution-spec-tests/pull/206/commits/739173c7ed7e1f9aa27166b8b24f22c0d57bed49) in combination with shemnon/besu/pull/14 fills all test cases up and including Shanghai successfully with:

```
fill --evm-bin=/path/to/besu/ethereum/evmtool/build/install/evmtool/bin/evm
```

siladu (commented):
Thanks for the descriptive PR! 
The changes themselves look good.

I would only question _why_ we are adding the evm bin to the standard distribution?
It won't be needed for the vast majority of users AFAICT and I'm concerned we may inconvenience more people than we help with this (cue the Discord questions about which binary to use and why things don't match the install guides etc)

shemnon:
Size wise it's a script and a small jar.

But the reason is to improve user experience.  Running execution-spec-tests should't require installing raw source and executing it.  It's like stripping javac from the java distro, Oracle stopped doing that 7 years ago when they sunset the JRE.

Note that Geth ships a whole lot more with their homebrew formula and linux packages.  Notably adding `clef`, `devp2p`, `abigen`, `bootnode`, `evm`, `rlpdump` and `puppeth`.  People have no problem identifying that 'geth' runs geth, similarly 'besu' runs besu.

shemnon:
> However, as we don't set the protected field in transactions any more, it would be helpful, if we could check for playback protection using v. I made a suggestion in https://github.com/shemnon/besu/pull/14.

Merged.

diega (approved):
👍

siladu:
> Note that Geth ships a whole lot more with their homebrew formula and linux packages.

Homebrew, yes, but if I download the linux package from https://geth.ethereum.org/downloads then it's just a single geth binary.

> People have no problem identifying that 'geth' runs geth, similarly 'besu' runs besu.

I do agree that it should be easy enough for users to figure it out and looking at the common setups guides the instructions will still work, so I have no objection based on impaired usability for less savvy users:

- https://besu.hyperledger.org/stable/public-networks/get-started/install/binary-distribution#linux--unix
- https://someresat.medium.com/guide-to-staking-on-ethereum-ubuntu-prysm-581fb1969460
- https://www.coincashew.com/coins/overview-eth/guide-or-how-to-setup-a-validator-on-eth2-mainnet/part-i-installation/step-3-installing-execution-client/besu#3.-setup-and-configure-systemd

siladu (approved):

PR #5708, opened 2023-07-16
replace ArrayList with HashSet for rpcMethodExists check

thinkAfCod:
## PR description
When executing the RpcMethod.rpcMethodExists(String rpcMethod) method, using HashSet instead of ArrayList provides a very very small performance improvement of a few nanoseconds. 😂😂

## Fixed Issue(s)
nothing


macfarla:
@thinkAfCod you need to add signoff to your commits to make the DCO happy.

siladu (approved):
LGTM, good spot

fab-10 (commented):
will be nice to have a micro benchmark to understand the improvement

macfarla:
> will be nice to have a micro benchmark to understand the improvement

@thinkAfCod do you have numbers from your testing that you could add to evidence to support the improvement?

shemnon (approved):

shemnon:
Design wise a O(ln) lookup is better than an O(n), so even without benchmarks I'm OK with this.  Whether we will see improvement in typical use cases is a valid question, but some cases with broken clients they may come to the conclusion their RPCs will fail quicker so they can increase the TPS of calls they should know fail.

PR #5711, opened 2023-07-20
Return all not selected transactions, not only invalid ones

fab-10:
## PR description

Apart from some little code cleanup, the main goal of this PR is to add, to the result of a block transaction selection, all the evaluated but the not selected transactions, with the reason they were not selected, this will potentially give more data during block creation, that could try different strategies to include some of the invalid transient transactions.


macfarla (approved):
LGTM - do you think it needs a changelog entry?

fab-10:
> LGTM - do you think it needs a changelog entry?

No changelog since there is not yet anything relevant to the users in this PR, following PRs will have.

PR #5726, opened 2023-07-27
Correctly cache the TransactionValidator instance on creation

fab-10:
## PR description

in #5682 `Suppliers.memoize` was misused and so a new instance was created at every `get` invocation, instead of correctly return the cached instance created at creation time


macfarla (approved):
LGTM

PR #5727, opened 2023-07-28
Make smart contract permissioning features work with london fork

siladu:
Override the transactionSimulator's default TransactionValidationParams with one that allows for exceeding the account balance (which effectively zeros the baseFee). This mimics the way that eth_estimateGas and eth_call are implemented. Similar change to #5277

Update ATs to use londonBlock (existing genesis allocs necessitate zeroBaseFee as well)


macfarla (approved):
LGTM

PR #5732, opened 2023-07-30
Add missing methods to transaction interface

### LINKED ISSUE #5731: Add missing methods to Transaction interface
Currently the `Transaction` interface does not declare all of the public methods of the Transaction implementation in `org.hyperledger.besu.ethereum.core`. Because the `Transaction` interface is for example used in the plugin API, plugins cannot access those missing methods. The interface needs to be updated.

daniellehrner:
## PR description

Missing methods which exist in the transaction class, but not the interface were added. This required the creation of a `AccessListEntry` interface in `org.hyperledger.besu.datatypes` and it's use in all the places that have used the class directly before.

## Fixed Issue(s)
fixes #5731


datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +186,83 @@ default Optional<? extends Quantity> getMaxFeePerDataGas() {
...
+
+  /**
+   * Return the maximum fee per gas the sender is willing to pay for this transaction.
+   *
+   * @return max fee per gas in wei
+   */
+  Wei getMaxGasPrice();
+
+  /**
+   * Return the maximum fee per gas the sender is willing to pay for this transaction.
```
macfarla, line 234:
```suggestion
   * Return the effective priority fee per gas for this transaction.
```

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +186,83 @@ default Optional<? extends Quantity> getMaxFeePerDataGas() {
...
+   *
+   * @param maybeBaseFee base fee in case of EIP-1559 transaction
+   * @return max fee per gas in wei
```
macfarla, line 237:
```suggestion
   * @return priority fee per gas in wei
```

datatypes/src/main/java/org/hyperledger/besu/datatypes/AccessListEntry.java
```diff
@@ -0,0 +1,22 @@
+package org.hyperledger.besu.datatypes;
+
+import java.util.List;
+
+import org.apache.tuweni.bytes.Bytes32;
+
+/** An access list entry as defined in EIP-2930 */
+public interface AccessListEntry {
```
shemnon, line 8: Can this be a record instead of an interface?

Even if it's not converted to a record, I'd rather see the whole class pulled in from the evm module to the datatypes module instead of an interface with a single implementation.

shemnon (changes requested):

datatypes/src/main/java/org/hyperledger/besu/datatypes/AccessListEntry.java
daniellehrner: As suggested I've removed the interface and created a record instead.

datatypes/src/main/java/org/hyperledger/besu/datatypes/AccessListEntry.java
```diff
@@ -12,35 +12,18 @@
...
-   * @param address the address
-   * @param storageKeys the storage keys
-   */
-  public AccessListEntry(final Address address, final List<Bytes32> storageKeys) {
-    this.address = address;
-    this.storageKeys = storageKeys;
-  }
-
+/** An access list entry as defined in EIP-2930 */
+public record AccessListEntry(Address address, List<Bytes32> storageKeys) {
```
garyschulte, line 26: 👍

garyschulte (approved):
🚢

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
    */
   TransactionType getType();
+
+  /**
+   * Boolean which indicates the transaction has associated cost data, whether gas price or 1559 fee
+   * market parameters.
+   *
+   * @return whether cost params are present
+   */
+  boolean hasCostParams();
```
shemnon, line 197: This is a Besu implementation detail.  I don't think it needs to be pulled up.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+
+  /**
+   * Returns the number of blobs this transaction has, or 0 if not a blob transaction type
+   *
+   * @return return the count
+   */
+  int getBlobCount();
```
shemnon, line 204: This can be derived from `getVersionedHashes` and `getBlobsWithCommitments`

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
shemnon, line 197: This is a Besu implementation detail. I don't think it needs to be pulled up.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+
+  /**
+   * Returns the signature used to sign the transaction.
+   *
+   * @return the signature used to sign the transaction
+   */
+  SECPSignature getSignature();
```
shemnon, line 211: This can be derived from the `R`, `S`, and `V` fields.  Also, it presumes ECDSA forever.  If non-ECDSA keys are added we would want a more generic Signature type.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+
+  /**
+   * Returns the public key extracted from the signature.
+   *
+   * @return the public key
+   */
+  Optional<String> getPublicKey();
```
shemnon, line 218: Public key and not address?

The impl already does an extraction from the signature when called.  This can be derived from `R`, `S`, and `V`

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+
+  /**
+   * Returns whether the transaction is a contract creation
+   *
+   * @return {@code true} if this is a contract-creation transaction; otherwise {@code false}
+   */
+  boolean isContractCreation();
```
shemnon, line 225: This can be derived from `getTo` being empty.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+
+  /**
+   * Return the maximum fee per gas the sender is willing to pay for this transaction.
+   *
+   * @return max fee per gas in wei
+   */
+  Wei getMaxGasPrice();
```
shemnon, line 232: This can be derived from fields already in the transaction interface.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+
+  /**
+   * Return the effective priority fee per gas for this transaction.
+   *
+   * @param maybeBaseFee base fee in case of EIP-1559 transaction
+   * @return priority fee per gas in wei
+   */
+  Wei getEffectivePriorityFeePerGas(final Optional<Wei> maybeBaseFee);
```
shemnon, line 240: This can be derived from fields already in the transaction interface.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+   * @return optional access list
+   */
+  Optional<List<AccessListEntry>> getAccessList();
+
+  /**
+   * Writes the transaction to RLP
+   *
+   * @param out the output to write the transaction to
+   */
+  void writeTo(final RLPOutput out);
```
shemnon, line 275: One of my goals is to remove the RLP dependency from the datatypes module.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+
+  /**
+   * Calculates the up-front cost for the gas and data gas the transaction can use.
+   *
+   * @param gasPrice the gas price to use
+   * @param blobGasPrice the data gas price to use
+   * @param totalBlobGas the gas the transaction can use
+   * @return the up-front cost for the gas the transaction can use.
+   */
+  Wei getUpfrontGasCost(final Wei gasPrice, final Wei blobGasPrice, final long totalBlobGas);
```
shemnon, line 285: This can be derived from fields already in the transaction interface.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+   * Calculates the up-front cost for the transaction.
+   *
+   * <p>The up-front cost is paid by the sender account before the transaction is executed. The
+   * sender must have the amount in its account balance to execute and some of this amount may be
+   * refunded after the transaction has executed.
+   *
+   * @param totalBlobGas the gas the transaction can use
+   * @return the up-front gas cost for the transaction
+   */
+  Wei getUpfrontCost(final long totalBlobGas);
```
shemnon, line 297: This can be derived from fields already in the transaction interface.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -183,4 +187,126 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
...
+   * calculated as:
+   *
+   * <p>min((baseFeePerGas + maxPriorityFeePerGas), maxFeePerGas)
+   *
+   * <p>Otherwise, return gasPrice for legacy transactions.
+   *
+   * @param baseFeePerGas optional baseFee from the block header, if we are post-london
+   * @return the effective gas price.
+   */
+  Wei getEffectiveGasPrice(final Optional<Wei> baseFeePerGas);
```
shemnon, line 311: This can be derived from fields already in the transaction interface.

shemnon (commented):
APIs are forever, and I am hesitant to add fields that can be derived from other fields in an effort to keep the API surface as small as possible.  That was the point of the separate `datatypes` module.  Unless there is a compelling "This use case requires X" I would rather keep those derived data methods out of the interface.

4844 and access list fields are good, as there are no other ways to get at that data.

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: Removed

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
daniellehrner: I've replaced it with a new generic method `Bytes encoded();`: [see here](https://github.com/hyperledger/besu/blob/12b88e1550988a9eb934460830fc469f0c87adb8/datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java#L221)

daniellehrner:
@shemnon I've adressed the changes, could you please review again?

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
shemnon: After diving into what it would take to remove it I'm thinking it should come back.

Because the CREATE operation depends on the RLP definitions to get the contract addresses the EVM module will still require the RLP, it's not just coming into EVM as a tag-along of datatypes.  So removing RLP from datatypes is not really needed, as some of the RLP dependencies once it is determined they should stay are best handled where they are at design-wise.

For performance we will want to use a passed in RLPOutput object as well.

shemnon (approved):
Optional: - move `bytes encoded()` to `void writeTo(final RLPOutput out);`

PR #5734, opened 2023-07-31
Address import performance issues

shemnon:
## PR description

Address a performance regression observed in import testing -
* ensure we are on a single tuweni version
* factor out an unneeded concatenate


macfarla (approved):
LGTM

ahamlat:
The results of microbenchmarking for both methods indicate that the new implementation, which avoids concatenation, shows a marginal performance improvement : 
```
Benchmark                                                         Mode  Cnt    Score   Error  Units
BenchmarkEncodeOpaqueBytes.benchmarkEncodeOpaqueBytesConcatenate  avgt   50  709.356 ± 5.901  ns/op
BenchmarkEncodeOpaqueBytes.benchmarkEncodeOpaqueBytesEncode       avgt   50  703.564 ± 1.937  ns/op
```
The PR also fixes the performance regression in 23.7.0

image

ahamlat (approved):

PR #5742, opened 2023-08-01
added worldview service to plugin api

### LINKED ISSUE #5740: Allow plugins to access any account
A Besu plugin can right now access the history of the blockchain, but not it's state. With a new `WorldViewService` we can access the world state and retrieve any account from it.

daniellehrner:
## PR description

Add a new `WorldViewService` which allows to get any account of the world state.

## Fixed Issue(s)
fixes #5740

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -169,6 +169,7 @@ import org.hyperledger.besu.plugin.services.SecurityModuleService;
 import org.hyperledger.besu.plugin.services.StorageService;
 import org.hyperledger.besu.plugin.services.TraceService;
 import org.hyperledger.besu.plugin.services.TransactionSelectionService;
+import org.hyperledger.besu.plugin.services.WorldViewService;
 import org.hyperledger.besu.plugin.services.exception.StorageException;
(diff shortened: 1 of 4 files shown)


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1760,6 +1762,11 @@ private void startPlugins() {
                 besuController.getProtocolContext().getWorldStateArchive()),
             besuController.getProtocolSchedule()));
 
+    besuPluginContext.addService(
+        WorldViewService.class,
+        new WorldViewServiceImpl(
+            besuController.getProtocolContext().getWorldStateArchive().getMutable()));
```
garyschulte, line 1768: It is quite risky to expose the 'root' worldstate from bonsai.  Could we instead expose BlockchainQueries or a similar functional interface like getAndMapWorldState() that has some safety to prevent accidental mutation?  

https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java#L904-L915

I believe the TracerService plugin already has blockchainQueries, we can keep with that pattern.

garyschulte (changes requested):
exposing that particular view from bonsai is 😨

PR #5748, opened 2023-08-03
Force the use of jdk8+ version of Bouncy Castle library

fab-10:
## PR description
This avoid that any dependency can include the jdk5 version in the distribution.

Implement the solution described here https://docs.gradle.org/7.6/userguide/dependency_capability_conflict.html

Had to inline the `xor` function since it is now deprecated in newer versions of Bouncy Castle.


fab-10:
@ahamlat did some new test with the last version of Bouncy Castle on a Standard_D4s_v5 instance, and I found that the differences between the distribution are very small, as a side node running on newer jdk gives a performance boost.
Could you validate the results?

```
Bouncy Castle 1.76

jdk15t018 (java19)

Benchmark                    (SIZE)  Mode  Cnt    Score   Error  Units
KeccakBench.benchmarkKeccak      32  avgt   50  540.001 ± 4.328  ns/op

jdk18on (java19)

Benchmark                    (SIZE)  Mode  Cnt    Score   Error  Units
KeccakBench.benchmarkKeccak      32  avgt   50  544.367 ± 3.604  ns/op

jdk15t018 (java17)

Benchmark                    (SIZE)  Mode  Cnt    Score   Error  Units
KeccakBench.benchmarkKeccak      32  avgt   50  608.889 ± 7.984  ns/op

jdk18on (java17)

Benchmark                    (SIZE)  Mode  Cnt    Score    Error  Units
KeccakBench.benchmarkKeccak      32  avgt   50  620.537 ± 14.868  ns/op
```

garyschulte (approved):
This is JDK 1.8+ right? not jdk 18+

Looks good to me, but it would be wise to get :eyes: from @ahamlat

fab-10:
> This is JDK 1.8+ right? not jdk 18+

Correct, let me amend the title

ahamlat (approved):
I was able to reproduce similar numbers regarding JDK 17 and JDK 19. JDK 21 has even better performance for both versions and I didn't notice a difference in performance.

These are some results on Mac M1 Aarch64
```
Java 21 - jdk15to18 

Benchmark                                         (SIZE)  Mode  Cnt    Score   Error  Units
HashBenchmarking.benchmarkKeccakVersionjdk15to18      32  avgt   50  290.137 ± 2.660  ns/op


Java 21 - jdk18on
Benchmark                                       (SIZE)  Mode  Cnt    Score   Error  Units
HashBenchmarking.benchmarkKeccakVersionjdk18on      32  avgt   50  290.327 ± 2.950  ns/op

PR #5754, opened 2023-08-04
Benchmarking in evmtool

shemnon:
## PR description

Add a new benchmarking sub-command to `evmtool`


besu/src/main/java/org/hyperledger/besu/cli/options/unstable/NativeLibraryOptions.java
```diff
@@ -53,7 +53,7 @@ public class NativeLibraryOptions {
           "Per default a native library is used for modexp. "
               + "If the Java implementation should be used instead, this option must be set to false",
       arity = "1")
-  private final Boolean nativeModExp = Boolean.TRUE;
+  private final Boolean nativeModExp = Boolean.FALSE;
```
daniellehrner, line 56: Why was this changed to `FALSE`? Shouldn't we prefer the native one per default?

besu/src/main/java/org/hyperledger/besu/cli/options/unstable/NativeLibraryOptions.java
```diff
@@ -53,7 +53,7 @@ public class NativeLibraryOptions {
           "Per default a native library is used for modexp. "
```
daniellehrner, line 53: If the change in line 56 is correct this description needs to be changed to default `false`

daniellehrner (changes requested):
Clarification needed regarding native mod exp command line.

daniellehrner (approved):

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/benchmarks/BenchmarkExecutor.java
```diff
@@ -0,0 +1,156 @@
...
+    this.iterations = iterations;
+  }
+
+  /** Run benchmarks with warmup and iterations set to MATH style benchmarks. */
+  protected BenchmarkExecutor() {
+    this(MATH_WARMUP, MATH_ITERATIONS);
+  }
+
+  /**
+   * Run the benchmark with the speicific args. Execution will be done warmup + iterations times
```
macfarla, line 92:
```suggestion
   * Run the benchmark with the specific args. Execution will be done warmup + iterations times
```

PR #5755, opened 2023-08-07
EVMTool contract create processing

shemnon:
## PR description

The "code" mode of the EVMTool did not support the side effects of a contract create. Expose the TransactionProcessor method that handles that switching.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
@@ -510,21 +510,17 @@ public class MainnetTransactionProcessor {
     }
   }
 
-  protected void process(final MessageFrame frame, final OperationTracer operationTracer) {
+  public void process(final MessageFrame frame, final OperationTracer operationTracer) {
     final AbstractMessageProcessor executor = getMessageProcessor(frame.getType());
 
     executor.process(frame, operationTracer);
   }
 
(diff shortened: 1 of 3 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -510,21 +510,17 @@ public TransactionProcessingResult processTransaction(
...
-      case MESSAGE_CALL:
-        return messageCallProcessor;
-      case CONTRACT_CREATION:
-        return contractCreationProcessor;
-      default:
-        throw new IllegalStateException("Request for unsupported message processor type " + type);
-    }
+    return switch (type) {
+      case MESSAGE_CALL -> messageCallProcessor;
+      case CONTRACT_CREATION -> contractCreationProcessor;
```
daniellehrner, line 522: Don't we still need the `default` case?

daniellehrner (changes requested):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
shemnon: Actually no.  Because this is an enum switch the compiler requires that all enums have a case, and if so Intellij marks the default as dead code (which it is).

Since type is never null this covers the same cases with no dead-code.

daniellehrner (approved):

PR #5756, opened 2023-08-07
Trigger `OperationTracer` on contexts enter & exit

### LINKED ISSUE #5728: Trace context changes in the EVM
Geth offers a more fine grained tracing capability than Besu. Currently in Besu it is not possible to trace the EVM on a context change (via call, create or selfdestruct). In Geth those methods are defined as:

```
// CaptureEnter is called when EVM enters a new scope (via call, create or selfdestruct).
func (t *opcounter) CaptureEnter(op vm.OpCode, from common.Address, to common.Address, input []byte, gas uint64, value *big.Int) {}

// CaptureExit is called when EVM exits a scope, even if the scope didn't
// execute any code.
func (t *opcounter) CaptureExit(output []byte, gasUsed uint64, err error) {}
```
For consistency I would name those methods `traceContextEnter` and `traceContextExit`.

delehef:
## PR description

Extends the `OperationTracer` interface to be triggered on contexts enter & exit.

@daniellehrner @shemnon 

## Fixed Issue(s)

Fixes #5728


evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -200,20 +207,12 @@ public void process(final MessageFrame frame, final OperationTracer operationTra
...
-      revert(frame);
-    }
-
-    if (frame.getState() == MessageFrame.State.COMPLETED_SUCCESS) {
-      completedSuccess(frame);
-    }
-
-    if (frame.getState() == MessageFrame.State.COMPLETED_FAILED) {
-      completedFailed(frame);
+    switch (frame.getState()) {
```
daniellehrner, line 210: AFAIK the `switch` statement is not equivalent to the chain of `if`s. More than one if condition could be triggered, if the state changes during the execution of `exceptionalHalt()`, `revert()`, `completedSuccess()` or `completedFailed()`. I think we need to revert this change.

evm/src/main/java/org/hyperledger/besu/evm/tracing/OperationTracer.java
```diff
@@ -69,7 +69,7 @@ default void traceAccountCreationResult(
    *
    * @param transaction the transaction which will be processed
    */
-  default void traceStartTransaction(final Transaction transaction) {}
+  default void traceStartTransaction(@SuppressWarnings("unused") final Transaction transaction) {}
```
daniellehrner, line 72: I think the `@SuppressWarnings` annotations are not strictly needed here. The CI pipeline should be okay with it not being used, because before it passed for `traceStartTransaction` as well.

daniellehrner (changes requested):

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
delehef: Fair point.

evm/src/main/java/org/hyperledger/besu/evm/tracing/OperationTracer.java
delehef: To be frank I didn't understand why it complained about unused parameters in `traceStartTransaction` but not in `traceEndTransaction`...

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -192,6 +208,9 @@ public void process(final MessageFrame frame, final OperationTracer operationTra
       codeExecute(frame, operationTracer);
 
       if (frame.getState() == MessageFrame.State.CODE_SUSPENDED) {
+        if (operationTracer != null) {
```
shemnon, line 211: Suspended does not mean the context is exiting.  Really what it means is that we are starting a new sub-context and the particulars of how this state machine is written we need to restart from the beginning.

This should be removed.

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -146,7 +152,10 @@ protected void revert(final MessageFrame frame) {
    *
    * @param frame The message frame
    */
-  private void completedSuccess(final MessageFrame frame) {
+  private void completedSuccess(final MessageFrame frame, final OperationTracer operationTracer) {
+    if (operationTracer != null) {
```
shemnon, line 156: I'd rather see these in the `process` method and these signatures not altered (like was done with the `CODE_SUSPENDED` case).  The `contextEnter` call also occurs there.

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -158,7 +167,10 @@ private void completedSuccess(final MessageFrame frame) {
    *
    * @param frame The message frame
    */
-  private void completedFailed(final MessageFrame frame) {
+  private void completedFailed(final MessageFrame frame, final OperationTracer operationTracer) {
+    if (operationTracer != null) {
```
shemnon, line 171: I'd rather see these in the `process` method and these signatures not altered (like was done with the `CODE_SUSPENDED` case).  The `contextEnter` call also occurs there.

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -124,7 +124,10 @@ private void clearAccumulatedStateBesidesGasAndOutput(final MessageFrame frame)
    *
    * @param frame The message frame
    */
-  private void exceptionalHalt(final MessageFrame frame) {
+  private void exceptionalHalt(final MessageFrame frame, final OperationTracer operationTracer) {
+    if (operationTracer != null) {
+      operationTracer.traceContextExit(frame);
+    }
```
shemnon, line 130: Note that in line 134 the state is set to `COMPLETED_FAILED` and this will result in context edit called twice

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -136,7 +139,10 @@ private void exceptionalHalt(final MessageFrame frame) {
    *
    * @param frame The message frame
    */
-  protected void revert(final MessageFrame frame) {
+  protected void revert(final MessageFrame frame, final OperationTracer operationTracer) {
+    if (operationTracer != null) {
+      operationTracer.traceContextExit(frame);
+    }
```
shemnon, line 145: Note that in line 147 the state is set to `COMPLETED_FAILED` so the context exit will be called twice.

shemnon (changes requested):
This change needs tests.  Either directly such as how `ExtendedOperationTracerTest` tested the code or indirectly like how `traceEndTransaction` is integrated into the testing of the `T8nExecutor`.

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
delehef: In which scenario can the `contextEnter` happen here?

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -208,31 +198,45 @@ public void process(final MessageFrame frame, final OperationTracer operationTra
...
-        if (operationTracer != null) {
-          operationTracer.traceContextExit(frame);
-        }
         return;
       }
 
       if (frame.getState() == MessageFrame.State.CODE_SUCCESS) {
+        if (operationTracer != null) {
+          operationTracer.traceContextExit(frame);
+        }
```
shemnon, line 207: This is unneeded.  We should only do this on `COMPLETED_SUCCESS` and `COMPLETED_FAILURE`

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -208,31 +198,45 @@ public void process(final MessageFrame frame, final OperationTracer operationTra
...
         codeSuccess(frame, operationTracer);
       }
     }
 
     if (frame.getState() == MessageFrame.State.EXCEPTIONAL_HALT) {
-      exceptionalHalt(frame, operationTracer);
+      if (operationTracer != null && frame.getMessageStackDepth() > 0) {
+        operationTracer.traceContextExit(frame);
+        exitTraced = true;
+      }
```
shemnon, line 216: This is unneeded.  We should only do this on `COMPLETED_SUCCESS` and `COMPLETED_FAILURE`

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -208,31 +198,45 @@ public void process(final MessageFrame frame, final OperationTracer operationTra
...
+      exceptionalHalt(frame);
     }
 
     if (frame.getState() == MessageFrame.State.REVERT) {
-      revert(frame, operationTracer);
+      if (operationTracer != null && !exitTraced && frame.getMessageStackDepth() > 0) {
+        operationTracer.traceContextEnter(frame);
+        exitTraced = true;
+      }
```
shemnon, line 224: This is unneeded.  We should only do this on `COMPLETED_SUCCESS` and `COMPLETED_FAILURE`

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -208,31 +198,45 @@ public void process(final MessageFrame frame, final OperationTracer operationTra
...
+      revert(frame);
     }
 
     if (frame.getState() == MessageFrame.State.COMPLETED_SUCCESS) {
-      completedSuccess(frame, operationTracer);
+      if (operationTracer != null && !exitTraced && frame.getMessageStackDepth() > 0) {
+        operationTracer.traceContextEnter(frame);
+        exitTraced = true;
```
shemnon, line 231: We don't need to track `exitTraced`

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -196,7 +184,9 @@ private void codeExecute(final MessageFrame frame, final OperationTracer operati
    * @param operationTracer the operation tracer
    */
   public void process(final MessageFrame frame, final OperationTracer operationTracer) {
-    if (operationTracer != null) {
+    boolean exitTraced = false;
```
shemnon, line 187: We don't need to track `exitTraced`

daniellehrner (approved):

shemnon:
I was hoping for a test more like https://github.com/hyperledger/besu/blob/532797fc1725d3264ca51d083cb2b8a89123f034/evm/src/test/java/org/hyperledger/besu/evm/StandardJsonTracerTest.java#L33, where there are  no mocks.  A custom tracer that outputs the tracing results into a data structure or text would then be verified.

With mocks-only testing I just don't feel it proves that it's not behaving in surprising ways.

daniellehrner:
@shemnon I added another test without mocks, based on EIP-3155. Could you please review again?

shemnon (approved):

PR #5760, opened 2023-08-14
Cancun genesis tests

jflo:
- fixes genesis block calculations when started post cancun
- refactors engine api tests


shemnon (approved):
Are these reference or spec tests we are adjusting the error messages for?  And I assume this is all cancun related?

PR #5763, opened 2023-08-15
EvmTool fixes

shemnon:
## PR description

A mixed collection of EVMTool fixes
* Ensure ECDSA algo is pre-configured
* used clamped math in an old gas calculator
* use correct gas calculator in fluent APIs


evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
```diff
@@ -546,7 +547,7 @@ public static EVM berlin(final EvmConfiguration evmConfiguration) {
    * @return the evm
    */
   public static EVM berlin(final BigInteger chainId, final EvmConfiguration evmConfiguration) {
-    return berlin(new IstanbulGasCalculator(), chainId, evmConfiguration);
+    return berlin(new BerlinGasCalculator(), chainId, evmConfiguration);
```
macfarla, line 550: this makes sense, just wondering if changing this is going to break anything? it's been this way for a long time!

macfarla (commented):
LGTM just one Q on the MainnetEVMs change

evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
shemnon: It's broken and wrong.  Should have caught it earlier.  It doesn't align with what the real definition of berlin is.  Also, only impacts users of the relatively new fluent evm APIs.  One researcher found a differential fuzzing bug based off of it, and this was the bug.

macfarla: ok makes sense then

macfarla (approved):

PR #5766, opened 2023-08-17
Issue 5719 - remove Pretty JSON and make it user configurable

### LINKED ISSUE #5712: Change default to remove Pretty JSON and make it user configurable
Pretty JSON is costly in the RPC stack and in file sizes. This issue tracks removing Pretty JSON as the default for Besu and creates a new flag:

`--pretty-json-enabled` with a default of `false`

This will be considered a breaking change and should be included in the RC series.

gtebrean:
## PR description
Added --pretty-json-enabled as CLI option as it was asked in the issue mentioned bellow.

Here I attached an example with the output of a RPC request when Besu was ran with --pretty-json-enabled (first case) and without the argument in second case:
Screenshot 2023-08-17 at 16 44 00

## Fixed Issue(s)
Fixes #5712


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -779,6 +780,12 @@ static class JsonRPCHttpOptionGroup {
         paramLabel = MANDATORY_LONG_FORMAT_HELP,
         description = "Specifies the maximum request content length. (default: ${DEFAULT-VALUE})")
     private final Long rpcHttpMaxRequestContentLength = DEFAULT_MAX_REQUEST_CONTENT_LENGTH;
+
+    @Option(
+        names = {"--pretty-json-enabled"},
+        description =
+            "Force pretty format output to be enabled/disabled (default: ${DEFAULT-VALUE})")
```
macfarla, line 787:
```suggestion
            "Enable JSON pretty print format (default: ${DEFAULT-VALUE})")
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Additions and Improvements
 - Added `benchmark` subcommand to `evmtool` [#5754](https://github.com/hyperledger/besu/issues/5754)
+- Added `--pretty-json-enabled` argument option to CLI in order to remove Pretty JSON as the default for improving memory usage [#5766](https://github.com/hyperledger/besu/pull/5766)
```
macfarla, line 11:
```suggestion
- JSON output is now compact by default. This can be overridden by the new `--pretty-json-enabled` CLI option. [#5766](https://github.com/hyperledger/besu/pull/5766)
```

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -779,6 +780,12 @@ static class JsonRPCHttpOptionGroup {
         paramLabel = MANDATORY_LONG_FORMAT_HELP,
         description = "Specifies the maximum request content length. (default: ${DEFAULT-VALUE})")
     private final Long rpcHttpMaxRequestContentLength = DEFAULT_MAX_REQUEST_CONTENT_LENGTH;
+
+    @Option(
+        names = {"--pretty-json-enabled"},
```
macfarla, line 785:
```suggestion
        names = {"--json-pretty-print-enabled"},
```

macfarla: json should definitely come first. and I think "pretty print" rather than just "pretty"

macfarla (commented):
does this CLI option cover both HTTP and websockets?

gtebrean:
> does this CLI option cover both HTTP and websockets?

Yes because is setting the format in `JsonRpcConfiguration` which is used by both. But anyway the websockets it returns the response without pretty print, unless `jq`,as example, is not specified as option.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Additions and Improvements
 - Added `benchmark` subcommand to `evmtool` [#5754](https://github.com/hyperledger/besu/issues/5754)
+- JSON output is now compact by default. This can be overridden by the new `--pretty-json-enabled` CLI option. [#5766](https://github.com/hyperledger/besu/pull/5766)
```
macfarla, line 11:
```suggestion
- JSON output is now compact by default. This can be overridden by the new `--json-pretty-print-enabled` CLI option. [#5766](https://github.com/hyperledger/besu/pull/5766)
```

macfarla (approved):

macfarla:
Docs change required - add a note to docs that all HTTP and Websockets responses will be compact JSON by default, and the `--json-pretty-print-enabled` can be used to pretty-print the output if desired.

PR #5768, opened 2023-08-17
Forkchoice v3

### LINKED ISSUE #5735: engine_forkchoiceUpdateV3
We now have a v3. it is a simple extension of 2 to add some error handling.

https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#engine_forkchoiceupdatedv3

jflo:
## Fixed Issue(s)
fixes #5735


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineGetPayload.java
```diff
@@ -61,6 +76,11 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext request) {
     if (blockWithReceipts.isPresent()) {
       final var proposal = blockWithReceipts.get();
       LOG.atDebug().setMessage("assembledBlock {}").addArgument(() -> proposal).log();
+      ValidationResult<RpcErrorType> forkValidationResult =
+          validateForkSupported(proposal.getHeader().getTimestamp());
+      if (!forkValidationResult.isValid()) {
+        return new JsonRpcSuccessResponse(request.getRequest().getId(), forkValidationResult);
```
macfarla, line 82: should this be an error response?

jflo: You'd think so, but no. Those are reserved for JsonRPCprotocol type errors.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
```diff
@@ -106,4 +123,22 @@ public final JsonRpcResponse response(final JsonRpcRequestContext request) {
   }
 
   public abstract JsonRpcResponse syncResponse(final JsonRpcRequestContext request);
+
+  protected ValidationResult<RpcErrorType> validateForkSupported(final long blockTimestamp) {
```
pinges, line 127: It is a bit funny, that the implementation for the cancun versions of the RPCs is in here, whereas the "default" implementation that is just returning valid is in all the other RPCs.
Might be easier to understand if we were to put this one in to the V3 RPCs and have the "valid" one here.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
```diff
@@ -106,4 +123,22 @@ public final JsonRpcResponse response(final JsonRpcRequestContext request) {
...
+    if (protocolSchedule.isPresent()) {
+      var cancun =
+          protocolSchedule.get().hardforkFor(s -> s.fork().name().equalsIgnoreCase("Cancun"));
```
pinges, line 130: This could be done once in the  constructor.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
```diff
@@ -54,24 +44,12 @@ public String getName() {
...
-      return ValidationResult.invalid(
-          RpcErrorType.INVALID_PARAMS, "Missing versioned hashes field");
-    } else if (maybeParentBeaconBlockRoot.isEmpty()) {
-      return ValidationResult.invalid(
-          RpcErrorType.INVALID_PARAMS, "Missing parent beacon block root field");
-    }
-    if (payloadParameter.getTimestamp() < cancunTimestamp) {
-      return ValidationResult.invalid(RpcErrorType.UNSUPPORTED_FORK, "Fork not supported");
+    } else {
+      return ValidationResult.valid();
```
pinges, line 52: we have lost some of the checks here. We need to check that the versioned hash params and the parent beacon block root are available when we are calling the v3 RPC.

jflo: They are checked in the parent abstract class, via the validateParams method.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
jflo: it is, but prior versions have no notion of validating on a supported fork, and all future versions will. So the default case moving forward should check fork boundaries, and prior versions can just assume validity till they are removed.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
```diff
@@ -45,18 +48,36 @@ public enum EngineStatus {
   private static final Logger LOG = LoggerFactory.getLogger(ExecutionEngineJsonRpcMethod.class);
   protected final Optional<MergeContext> mergeContextOptional;
   protected final Supplier<MergeContext> mergeContext;
+  protected final Optional<ProtocolSchedule> protocolSchedule;
   protected final ProtocolContext protocolContext;
   protected final EngineCallListener engineCallListener;
+  protected final Optional<ScheduledProtocolSpec.Hardfork> cancun;
```
garyschulte, line 54: this is a V3 concern right?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
garyschulte: syncResponse is calling the much more robust validateBlobs, but it doesn't look like validateBlobs enforces that the parameters are present.  Does the spec for v3 require `expectedBlobVersionedHashes` to be present even if the list is empty?

garyschulte (commented):
nice refactor.  just a couple questions

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
macfarla: done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
macfarla: done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/ExecutionEngineJsonRpcMethod.java
jflo: Yes, but used across 3 different apis, and will likely be used by more hardforks in the future.

Gabriel-Trintinalia (approved):
LGTM

PR #5772, opened 2023-08-18
Layered txpool by default and txpool options hoverhaul

fab-10:
## PR description

This PR started with the goal to switch to the new layered by default, adding the possibility to select the old implementation if needed, introducing a new option `--tx-pool` that can take the values `layered` (default) and `legacy`, all the previously unstable options for the layered pool `--tx-pool-layer-max-capacity`, `--tx-pool-max-prioritized`, `--tx-pool-max-future-by-sender` are not stable.
Options specific to each implementation have been grouped, also in the help section, and validation has been added to fail in case options for a specific implementation are misused with the other implementation.

While implementing this, I found that the txpool options were a bit messy, stable and unstable options mixed, many options cluttering `BesuCommand` and the presence of boilerplate (and error prone) code, so I decided to use this PR to also review, clean and improve the code:
- to move all the txpool options away from `BesuCommand` and split them correctly between stable and unstable, 
- made `getCLIOptions` to dynamically generate its output, via reflection, avoiding to manually add, and then maintain, any option, that is error prone. Most of the changes are related to this feature, that inspect by reflection the options in the class. Since these classes are only used at startup or during unit tests the performance of the reflection should not be an issue
- made the custom converters to return directly the high level type instead of the raw type, that seems the right way to use them.
- fixed some issues that happens when using the layered txpool by default in tests.

If the work done for txpool options, is found to be effective, then I can apply the same to other groups of options to clean and uniform them.


besu/src/main/java/org/hyperledger/besu/cli/converter/exception/DurationConversionException.java
```diff
@@ -0,0 +1,39 @@
...
+   * @param value the value
+   */
+  public DurationConversionException(final String value) {
+    super(format("'%s' is not a long", value));
+  }
+
+  /**
+   * Instantiates a new Duration conversion exception for negative value.
+   *
+   * @param millis the value
```
fab-10, line 34:
```suggestion
   * @param value the millis
```

macfarla (approved):
LGTM

PR #5774, opened 2023-08-18
Add EIPTests to ReferenceTests

shemnon:
## PR description

Add the EIPTests directory to the set of reference tests.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -291,6 +299,18 @@ public ValidationResult<TransactionInvalidReason> validateTransactionsBlobs(
...
+          TransactionInvalidReason.INVALID_BLOBS,
+          "transaction blob transactions must specify one or more versioned hashes");
+    }
+
+    return ValidationResult.valid();
+  }
+
+  public ValidationResult<TransactionInvalidReason> validateTransactionsBlobs(
+      final Transaction transaction) {
+
```
shemnon, line 313: @jflo can you make sure this section won't cause problems?  I needed to validate a type 3 / blob transaction has a versioned hash but it didn't have any of the blob data to go with it.  Seems like a typical block replay use case.

shemnon, line 313: @jflo can you make sure this section won't cause problems? I needed to validate a type 3 / blob transaction has a versioned hash but it didn't have any of the blob data to go with it. Seems like a typical block replay use case.

jflo: This is correct, and won't cause problems. You only have blob data with network transactions which arrived via gossip. When replaying them out of a block we only have the versioned hashes.

jflo (approved):

PR #5776, opened 2023-08-21
fix snapsync issue with forest

matkt:
## PR description

Fixing an issue during snapsync with forest related to the heal step

the step checking whether the node existed or not during the healing process could lead to a null pointer exception due to the null nodeHash, which was only utilized within the context of forest.


matkt:
Tests done

Scenario/StorageFormat | BONSAI | FOREST
-- | -- | --
Initial Snapsync | ✅ | ✅

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/WorldStateKeyValueStorage.java
```diff
@@ -78,6 +78,11 @@ private Optional<Bytes> getTrieNode(final Bytes32 nodeHash) {
     }
   }
 
+  @Override
+  public Optional<Bytes> getUnSafeTrieNode(final Bytes key) {
+    return keyValueStorage.get(key.toArrayUnsafe()).map(Bytes::wrap);
+  }
```
fab-10, line 84: a style suggestion, rename method to `getTrieNodeUnsafe` to uniform with other _unsafe_ methods like `toArrayUnsafe`

fab-10 (approved):
LGTM this fix the sync on the Sepolia instance, just a style suggestion.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/WorldStateKeyValueStorage.java
matkt: done

PR #5786, opened 2023-08-23
fix excess blob gas calculation when creating blobs

pinges:
## PR description
This fixes an issue where we were calculating the excess blob gas in a wrong way when generating a block proposal.


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -317,9 +318,7 @@ private GasUsage computeExcessBlobGas(
       // casting parent excess blob gas to long since for the moment it should be well below that
       // limit
       BlobGas excessBlobGas =
-          BlobGas.of(
-              gasCalculator.computeExcessBlobGas(
-                  parentHeader.getExcessBlobGas().map(BlobGas::toLong).orElse(0L), newBlobsCount));
+          ExcessBlobGasCalculator.calculateExcessBlobGasForParent(newProtocolSpec, parentHeader);
```
macfarla, line 321: is there a test to cover this or are we relying on hive tests here?

Gabriel-Trintinalia (approved):
LGTM, I am working on adding some tests for this.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
pinges: Gabriel is in the process of adding some tests for that! I'll get this merged and let Gabriel add the tests :-)

PR #5789, opened 2023-08-23
Make referencetests build script Windows friendly

fab-10:
## PR description

#5774 broken builds on Windows due to the different separation char, this PR just replace the '/' with the system separator char as provided by the jvm


Gabriel-Trintinalia (approved):
LGTM

PR #5792, opened 2023-08-24
EngineNewPayload - Validate parameters before fork validation

Gabriel-Trintinalia:
Parameters validation should occur before fork validation

> Specification
> This method follows the same specification as [engine_newPayloadV2](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_newpayloadv2) with the addition of the following:
> 
> 1. Client software MUST check that provided set of parameters and their fields strictly matches the expected one and return -32602: Invalid params error if this check fails. Any field having null value MUST be considered as not provided.
> 
> 2. Client software MUST return -38005: Unsupported fork error if the timestamp of the payload does not fall within the time frame of the Cancun fork.
> ...


pinges (approved):
LGTM

PR #5793, opened 2023-08-24
Remove versionedHashes from EngineNewPayload

Gabriel-Trintinalia:
Remove versionedHashes from EngineNewPayload. 

It does not belong to the payload parameter:
https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#executionpayloadv3


pinges (approved):
LGTM

PR #5795, opened 2023-08-24
add log for unexpected pipeline error

matkt:
## PR description

This PR allows displaying unexpected errors during the sync. Indeed, until now, it was necessary to go into debug mode to see them and therefore we could have stuck pipelines without knowing what the error is.


gfukushima (approved):
lgtm, just a question: have we manually checked that this is not going to be a noisy log during sync?

matkt:
> lgtm, just a question: have we manually checked that this is not going to be a noisy log during sync?

Yes normally it's fine . I tried sync on mainnet, goerli and sepolia and I also tried with the forest node to be sure that unexpected issues are displayed

PR #5803, opened 2023-08-25
EngineNewPayload -  Fix Invalid Block Hash return

Gabriel-Trintinalia:
EngineNewPayload should not return the  latestValidHash when blockHash validation fails:

Current:
{status: INVALID_BLOCK_HASH, latestValidHash: `latestValidHash`, validationError: errorMessage | null} if the blockHash validation has failed

Expected:
{status: INVALID_BLOCK_HASH, latestValidHash: `null`, validationError: errorMessage | null} if the blockHash validation has failed


gfukushima (approved):
lgtm

macfarla (approved):
should we have a unit test for this?

PR #5804, opened 2023-08-25
don't check block timestamp drift for PoS

### LINKED ISSUE #5800: hive test failing - block too far in future
Besu validates (see TimestampValidationRuleTest) that blocks are no more than 15s into the future, as per https://eips.ethereum.org/EIPS/eip-1482

geth has this validation rule for ethash https://github.com/ethereum/go-ethereum/blob/4e474c74dc2ac1d26b339c32064d0bac98775e77/consensus/ethash/consensus.go#L45

but no longer in beacon consensus https://github.com/ethereum/go-ethereum/blob/master/consensus/beacon/consensus.go#L226

Think we can remove this validation rule from mainnet PoS consensus

macfarla:
Geth has removed this check (block too far in the future) since the merge. Fixes #5800 - and 9 related hive tests that start with `cancun/eip4788_beacon_root/beacon_root_contract/`


Gabriel-Trintinalia (approved):
LGTM

PR #5819, opened 2023-08-29
fix ForkId if there are no Forks and the starting timestamp is not 0

pinges:
## PR description
Make sure that we are passing the fork id check if there are no forks and the genesis timestamp is not 0


Gabriel-Trintinalia (approved):
lgtm

PR #5823, opened 2023-08-29
[4844] EngineNewPayloadV3 - Validate non null required paremeters

Gabriel-Trintinalia:
## Description
This PR enhances the validation of null for the `EngineNewPayloadV3` parameters. Previously, these parameters were extracted using the `getOptionalParameter` method, which returns an `Optional.empty `if the parameters are null. This approach was insufficient as the engine can accept empty lists of versionedHashes, but cannot accept a null parameter.

The main changes introduced by this PR include:

- Modification of the parameter extraction method to allow for more robust null validation.
- Adjustment of the validation logic to correctly handle empty lists of versionedHashes and reject null parameters.
- Will now return "Invalid Parameters" if any required parameters are null.
- Renames `EnginePayloadParameter` to `EngineExecutionPayloadParameter` to match spec naming
- validateParams becomes `getAndCheckEngineNewPayloadRequestParams`

Fixes the following Hive tests:

- NewPayloadV3 Versioned Hashes, Nil Hashes
- NewPayloadV3 Versioned Hashes, Nil Hashes (Syncing)


jflo (changes requested):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3.java
```diff
@@ -52,20 +51,12 @@ public String getName() {
 
   @Override
   protected ValidationResult<RpcErrorType> validateParameters(
-      final EnginePayloadParameter payloadParameter,
-      final Optional<List<String>> maybeVersionedHashParam,
-      final Optional<String> maybeBeaconBlockRootParam) {
-    if (payloadParameter.getBlobGasUsed() == null || payloadParameter.getExcessBlobGas() == null) {
+      final EngineNewPayloadRequestParameter params) {
+    if (params.getExecutionPayload().getBlobGasUsed() == null
+        || params.getExecutionPayload().getExcessBlobGas() == null) {
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: `maybeVersionedHashParam` == null would never happen to be true in the previous code as maybeVersionedHashParam was being extracted using `getOptionalList` which returns Optional.Empty in case of `null`.

maybeVersionedHashParam is now treated as a required parameter for V3 and will throw if is missing.  `BeaconBlockRoot` will also throw if `null`. 

There are two tests in this PR that test this validation.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV1.java
```diff
@@ -60,6 +64,29 @@ protected EngineStatus getInvalidBlockHashStatus() {
     return INVALID_BLOCK_HASH;
   }
 
+  @Override
+  public EngineNewPayloadRequestParameter getAndCheckEngineNewPayloadRequestParams(
+      final JsonRpcRequestContext requestContext) {
+    final EngineExecutionPayloadParameter enginePayload =
+        requestContext.getRequiredParameter(0, EngineExecutionPayloadParameter.class);
+    if (enginePayload.getWithdrawals() != null) {
+      throw new InvalidJsonRpcParameters("non-null Withdrawals pre-shanghai");
```
jflo, line 73: This feels like some API version leakage. We don't want to have to continuously update these ancient version apis as we evolve, and add more and more parameter checks and errors.

It might be simpler to create a versioned request parameter, and add strict JSON translation, so it throws an error on unexpected properties. That way the api never needs to be modified later, but will always error out if the client tries to call it with irrelevant fields.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3Test.java
```diff
@@ -55,7 +52,7 @@
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith(MockitoExtension.class)
-public class EngineNewPayloadV3Test extends EngineNewPayloadV2Test {
+public class EngineNewPayloadV3Test extends AbstractEngineNewPayloadTest {
```
jflo, line 55: See my suggestion above. If we stop that leakage, we should be able to still re-use the testing of prior versions.

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV3Test.java
```diff
@@ -108,20 +105,8 @@ public void shouldInvalidVersionedHash_whenShortVersionedHash() {
...
-    final EnginePayloadParameter payload =
-        mockEnginePayload(mockHeader, Collections.emptyList(), null, null);
-
-    ValidationResult<RpcErrorType> res =
-        method.validateParameters(
-            payload,
-            Optional.of(List.of()),
-            Optional.of("0x0000000000000000000000000000000000000000000000000000000000000000"));
-    assertThat(res.isValid()).isTrue();
+    final JsonRpcRequestContext mock = getRequestContextMock(null, 1L, "1", new String[0], "0x0");
```
jflo, line 108: nit: would rename this mockRequestContext() so it implies it is creating something, not getting an already existing thing.

jflo (commented):
See comments, would like to regain the test coverage lost by not extending the api test for the previous version.

macfarla:
@Gabriel-Trintinalia @jflo is this PR redundant now? Ok to close?

PR #5826, opened 2023-08-30
enforce that BlobTransactions have at least one blob

pinges:
## PR description
make sure that a blob transaction has at least one blob.


datatypes/src/main/java/org/hyperledger/besu/datatypes/BlobsWithCommitments.java
```diff
@@ -38,11 +38,15 @@ public BlobsWithCommitments(
       final List<Blob> blobs,
       final List<KZGProof> kzgProofs,
       final List<VersionedHash> versionedHashes) {
+    if (blobs.size() < 1) {
```
pinges, line 41:
```suggestion
    if (blobs.size() == 0) {
```

Gabriel-Trintinalia (approved):
lgtm

PR #5831, opened 2023-08-31
add versioned hashes and number of blobs to toString()

pinges:
## PR description
Adding missing information to the toString() method of Transaction


macfarla (approved):
LGTM

Issue #5835, opened 2023-09-01
Provide a world view to the start/end transaction hooks in tracing API


delehef:
### Description

Currently, the start/end transaction hooks in the tracing API only provide information about the transaction itself.

It would be beneficial for more advanced tracers to provide an access to a non-mutable state of the world, _e.g._ to retrieve information about the accounts implicated in the transaction.

PR #5836, opened 2023-09-01
Add world context to transaction tracing API

delehef:
## PR description

Add world context to the start/end transaction hooks in the tracing API.

## Fixed Issue(s)

fixes: #5835


ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java
```diff
@@ -270,6 +270,7 @@ static T8nResult runTest(
       final TransactionProcessingResult result;
       try {
         tracer = tracerManager.getManagedTracer(i, transaction.getHash());
+        tracer.traceStartTransaction(worldStateUpdater, transaction);
```
delehef, line 273: I am not sure why the tracer was not called here previously?

matkt:
could you explain how you will use this updater. I don't think it's a good idea to pass the worldUpdater to the trace object. I think you want to use this worldUpdater later in order to retrieve some account or other data. 

in bonsai you have a huge risk that this objet will be a closed snapshot and it will not work . I you need a specific data from this updater I think it's better to take it directly and pass the value to the trace object

matkt:
> could you explain how you will use this updater. I don't think it's a good idea to pass the worldUpdater to the trace object. I think you want to use this worldUpdater later in order to retrieve some account or other data.
> 
> in bonsai you have a huge risk that this objet will be a closed snapshot and it will not work . I you need a specific data from this updater I think it's better to take it directly and pass the value to the trace object

ok we did a check and it's fine regarding that .

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java
matkt: it seems you added the traceEndTransaction @diega . do you know if it's normal to not have the startTransaction ?

matkt (commented):
LGTM for me it's ok but I waiting for @diega regarding T8nExecutor class and maybe @shemnon

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java
diega: Yes, it looks like it was needed. I never used the `--json|--trace` flag from the `t8n` tool, so I cannot say if/how it was working without the tracer start. Probably @shemnon can confirm it

macfarla:
@diega @shemnon any objections to this approach?

matkt (approved):

diega:
What's the practical use of this change? is there some other PR giving more context on why is this new interface needed?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java
shemnon: It feels like this should be called somewhere in [MainnetTransactionProcessor.processTransaction] (https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java#L260) rather than as part of the lead-up.

shemnon: same with endTransaction.

shemnon (changes requested):
I want to see some tests for this.  They can be unit tests, or the full JSON tests using the T8n spec tracer, but I would like to have some output verified so that if the call to  trace methods get moved around incorrectly or removed something breaks as part of the build.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java
delehef: @shemnon given that `gasUsed` can't be computed from within `processTransaction`, that implies to either:

 - dirty `processTransaction` signature to pass `gasUsed` only for the sake of tracing;
 - just drop `gasUsed` from the `traceEndTransaction` API.

What do you think?

delehef:
> What's the practical use of this change?

This change is part of a larger set of changes that are helping in the development of the Linea project.

shemnon (approved):

PR #5838, opened 2023-09-01
Fix snapsync heal

matkt:
## PR description

This PR improves the snapsync, particularly the healing stage.

- It enhances the stability of the trie healing by avoiding the bug that causes the sync to get stuck at 99%.
-  It improves the tracking of invalid accounts to improve the healing of the flatDB.
- It fix some issues related to the healing of the flat db.

We are also adding more comment and some cleaning of the code

## TESTS

- 2 mainnet nodes (snapsync + snapsync with flat db heal)
- 2 goerli nodes (snapsync + snapsync with flat db heal)
- 6 sepolia nodes (snapsync + snapsync with flat db heal)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
```diff
@@ -274,15 +297,15 @@
...
+  public synchronized void addAccountToHealingList(final Bytes account) {
+    if (!accountsHealingList.contains(account)) {
+      snapContext.addAccountToHealingList(account);
+      accountsHealingList.add(account);
     }
   }
 
-  public HashSet<Bytes> getAccountsToBeRepaired() {
-    return accountsToBeRepaired;
+  public HashSet<Bytes> getAccountsHealingList() {
```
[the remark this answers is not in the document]
matkt: ignore this one because it's not in the scope of this PR. I'm just renaming the method in this PR . Don't want to modify the logic

macfarla (commented):
PR looks good @matkt - but I'm curious as to which change exactly fixed the 99% bug?

matkt:
> PR looks good @matkt - but I'm curious as to which change exactly fixed the 99% bug?

mainly related to https://github.com/hyperledger/besu/pull/5838/files#diff-e341896fe8d23fac861ef07ae0a8edd6d3412c4ffd356af6b53698507d62f1b9R116

This was related to accounts that were marked for healing at a later stage. The problem was that the storage for these accounts became empty in the meantime. As a result, we were trying to heal an empty storage and were asking the peer for the storage root node. Since the storage was empty, the peers returned an empty response, and we kept trying indefinitely because, from our perspective, we were waiting for a response.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
```diff
@@ -385,25 +408,25 @@ public void setPivotBlockSelector(final DynamicPivotBlockSelector pivotBlockSele
...
-        if (!pivotBlockSelector.isBlockchainBehind()) {
-          snapSyncState.setWaitingBlockchain(false);
-        }
-        if (!snapSyncState.isWaitingBlockchain()) {
-          blockObserverId.ifPresent(blockchain::removeObserver);
-          blockObserverId = OptionalLong.empty();
-          reloadTrieHeal();
-        }
+      final AtomicBoolean foundNewPivotBlock = new AtomicBoolean(false);
+      pivotBlockSelector.check(
```
garyschulte, line 414: is it not a problem that `check` executes async code when it determines that the pivot block is too far behind ?  For example could we incorrectly return no new pivot if we have no peers.  

The async code in DynamicPivotBlockSelector makes this hard to reason about

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/AccountFlatDatabaseHealingRangeRequest.java
```diff
@@ -95,12 +95,12 @@ public Stream<SnapDataRequest> getChildRequests(
       downloadState.getMetricsManager().notifyRangeProgress(HEAL_FLAT, endKeyHash, endKeyHash);
     }
 
-    Stream.of(existingAccounts.entrySet(), removedAccounts.entrySet())
+    Stream.of(existingAccounts.entrySet(), flatDbAccounts.entrySet())
         .flatMap(Collection::stream)
```
garyschulte, line 99: non blocking, but Streams.concat() would be more readable than Stream.of(..).flatMap(Collection::stream).

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/AccountTrieNodeHealingRequest.java
```diff
@@ -112,6 +112,9 @@ public Stream<SnapDataRequest> getRootStorageRequests(final WorldStateStorage wo
                   account.size() - getLocation().size()))
           .map(RLP::input)
           .map(StateTrieAccountValue::readFrom)
+          .filter(
+              stateTrieAccountValue ->
+                  !stateTrieAccountValue.getStorageRoot().equals(MerkleTrie.EMPTY_TRIE_NODE_HASH))
```
macfarla, line 117: maybe add a comment here since this is the crucial change
don't request accounts with empty storage since this indicates they can't be healed

garyschulte (commented):
Generally LGTM.  The comments are very helpful.  It would be helpful to have a joint review though, to clear up some questions.  

Also there seems to be a problem with the upstream version metadata, not related to the PR.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
matkt: Indeed, the code for the Pivot Block Selector is a bit complex to understand and could benefit from more detailed descriptions.

This class fetches pivot blocks in advance to avoid having to search at the time of the call. I don't have the exact numbers, but for example:

I call the check method and if the distance to the head is >60 and <126, I will cache a pivot block. I won't use it right now. On the next call, when the distance to the head is close to 126, I will take the one cached. This means that every time we call this method and we are in this range(>60 and <128), we are looking for a pivot block to put in a cache ,  This is to have fast access to the next pivot block

In conclusion, there will be no problem as it will be synchronous (every time we will take the pivot from cache).

matkt: it would probably be beneficial to simplify this pivot block selection code as it's more technical debt. Previously, obtaining a pivot block required querying 5 peers and reaching a consensus, which could be time-consuming. Now, it's simply a matter of taking the last safe block. Therefore, this logic may no longer be necessary.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/AccountFlatDatabaseHealingRangeRequest.java
matkt: do you prefer something like that ? 

```java
Stream.concat(existingAccounts.entrySet().stream(), flatDbAccounts.entrySet().stream())
        .forEach(
        ```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/AccountTrieNodeHealingRequest.java
matkt: added

garyschulte (approved):
🚢

PR #5844, opened 2023-09-05
Range tracing with worldstate

matkt:
## PR description

Implement a method to trace a range of blocks and have access to the worldstate before and after the tracing


besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
```diff
@@ -88,61 +91,108 @@ public void traceBlock(final Hash hash, final BlockAwareOperationTracer tracer)
...
+            .mapToObj(
+                number ->
+                    blockchain
+                        .getBlockByNumber(number)
+                        .orElseThrow(() -> new RuntimeException("Block not found " + number)))
+            .toList();
     Tracer.processTracing(
         blockchainQueries,
-        block.getHash(),
+        blocks.get(0).getHash(),
```
garyschulte, line 123: This looks clean and succinct.  But I think we at least need a unit test that builds block 1, 2, 3 where the transactions of block 2 and 3 would fail to execute against the parent worldstate of block 1, but succeed when executed against their correct parent worldstates.  

When I first looked at the implementation it looked like we would be executing all of the blocks against the (unmutated) parent worldstate of block[0]

garyschulte (commented):
Looks good, but I think we need at least one test to prove the worldstate mutation is working as anticipated.

besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
```diff
@@ -88,61 +91,108 @@ public void traceBlock(final Hash hash, final BlockAwareOperationTracer tracer)
...
+   * @param tracer an instance of OperationTracer
+   */
+  @Override
+  public void trace(
+      final long fromBlockNumber,
+      final long toBlockNumber,
+      final Consumer<WorldState> beforeTracing,
+      final Consumer<WorldState> afterTracing,
+      final BlockAwareOperationTracer tracer) {
+    checkArgument(tracer != null);
```
garyschulte, line 110: it might be wise to implement some rational limit on the number or size of the blocks we are tracing.  If only to prevent an unintentional DoS of besu with super long block ranges

besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
matkt: good catch I added tests and found a bug. It's working correctly now

besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
matkt: I think this method is private and only used internally with plugin API. This one is too much in the deep layers. I think if we want to add a limitation it need to be done and managed before calling this one (RPC part etc). 

Also close to impossible to define a good limit. it will depends on the block size and the network configuration .

besu/src/test/java/org/hyperledger/besu/services/TraceServiceImplTest.java
garyschulte, line 1: 👍

garyschulte (approved):
🚢

Issue #5856, opened 2023-09-06
NullPointerException when fast syncing QBFT chain


matthew1001:
### Description
```
2023-09-06 09:09:07.295+01:00 | BftProcessorExecutor-QBFT-0 | ERROR | EventMultiplexer | State machine threw exception while processing event \{BlockTimerExpiry{Round Identifier=ConsensusRoundIdentifier{Sequence=378, Round=0}}\}
java.lang.IllegalStateException: Block creation failed unexpectedly. Will restart on next block added to chain.
        at org.hyperledger.besu.ethereum.blockcreation.AbstractBlockCreator.createBlock(AbstractBlockCreator.java:292)
        at org.hyperledger.besu.ethereum.blockcreation.AbstractBlockCreator.createBlock(AbstractBlockCreator.java:156)
        at org.hyperledger.besu.ethereum.blockcreation.AbstractBlockCreator.createBlock(AbstractBlockCreator.java:142)
        at org.hyperledger.besu.consensus.qbft.statemachine.QbftRound.createAndSendProposalMessage(QbftRound.java:130)
        at org.hyperledger.besu.consensus.qbft.statemachine.QbftBlockHeightManager.handleBlockTimerExpiry(QbftBlockHeightManager.java:136)
        at org.hyperledger.besu.consensus.common.bft.statemachine.BaseBftController.handleBlockTimerExpiry(BaseBftController.java:167)
        at org.hyperledger.besu.consensus.common.bft.EventMultiplexer.handleBftEvent(EventMultiplexer.java:65)
        at java.base/java.util.Optional.ifPresent(Optional.java:178)
        at org.hyperledger.besu.consensus.common.bft.BftProcessor.run(BftProcessor.java:65)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.NullPointerException: Cannot invoke "org.hyperledger.besu.ethereum.eth.transactions.PendingTransactions.selectTransactions(org.hyperledger.besu.ethereum.eth.transactions.PendingTransactions$TransactionSelector)" because "this.pendingTransactions" is null
        at org.hyperledger.besu.ethereum.eth.transactions.TransactionPool.selectTransactions(TransactionPool.java:526)
[...]

PR #5861, opened 2023-09-08
Don't start BFT mining coordinators until initial sync has completed

matthew1001:
## PR description
This PR is a suggested approach to resolving the NPE described in issue https://github.com/hyperledger/besu/issues/5856 whereby a QBFT block expiry timer causes a null pointer if the QBFT node is still syncing with the rest of the chain with `--sync-mode=FAST` or `--sync-mode=X_SNAP`

I have made the fix specific to BFT mining coordinators so that the behaviour of other consensus mining algorithms is unaffected by the change. I tested a lower-level fix to just for `null` in `TransactionPool.selectTransactions()` and that does indeed stop the NPE and allow the node to complete syncing successfully. However I have left the check out of this PR because it feels more appropriate to avoid any reason for trying to select transactions for a BFT miner that hasn't sync'd yet.

I've noted that while this prevents the NPE for `--sync-mode=X_SNAP` and the sync appears to complete successfully, the node does not participate in BFT block validation or proposals. I think that's a separate issue, and I assume it hasn't worked before since the NPE would have been hit. I suspect a new issue should be opened for that to be looked at separately but I'll wait for the conclusion of this PR.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/5856


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/BftMiningCoordinator.java
```diff
@@ -122,7 +124,13 @@ public void awaitStop() throws InterruptedException {
 
   @Override
   public boolean enable() {
-    return true;
+    // Return true if we're already running or idle, or successfully switch to idle
+    if (state.get() == State.RUNNING
+        || state.get() == State.IDLE
+        || state.compareAndSet(State.PAUSED, State.IDLE)) {
```
macfarla, line 130: what's the difference between paused and idle?

macfarla (approved):
LGTM. one small question

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/BftMiningCoordinator.java
matthew1001: The existing logic had `IDLE`, `RUNNING`, and `STOPPED`, where `IDLE` meant "not started yet". But I wanted to add a "not enabled yet" state where enabling has to happen before starting. So `PAUSED` is intended to mean "Waiting to be enabled, at which point we are IDLE until started".

matthew1001: There's an argument that the function calls (`start()`, `stop()`, `enable()`) should move the coordinator into the same state - STARTED, STOPPED, ENABLED etc. but I didn't want to refactor that too much

PR #5863, opened 2023-09-08
Merge MutableAccount and EVMAccount

shemnon:
## PR description

Merge MutableAccount and EVMAccount functionalities by removing EVMAccount, all calls to getMutable, and the WrappedEVMAccount that was wrapping non-EVMAccounts in a mutable fashion.  Instead, use a MutableAccount in all cases an EVMAccount would have been used.  This also tends to reduce a level of layering in many places.


macfarla (approved):
LGTM

PR #5865, opened 2023-09-11
Feature/multi version flat db

### LINKED ISSUE #5846: Implement Bonsai archive storage format
### Description
As a besu user, I want to have fast, lightweight access to historical data at block-level granularity.  
Kikori is envisioned as:
* for flat db's,  an evolution of bonsai flat db, with block-level historical state data encoded in key suffixes
* for state trie, implement a checkpointed state trie strategy to enable faster-than-single-log-rolling access to historical state tries 

### Acceptance Criteria
* implement kikori flat db reader & writer strategy
* implement checkpointed state tries solution
* implement kikori worldstate and worldstate provider infrastructure to support kikori worldstate (supplying historical flat db based worldstates and optionally/lazily provided versioned state tries)
* story for bonsai -> kikori migration

*

garyschulte:
## PR description
Draft version of multi-version flat db
* creates bonsai context concept
* creates FlatDbArchiveStrategy
* kikori worldstate provider integrated into BonsaiWorldStateProvider (not suitable for eth_getProof)

There are a handful of hacks that are marked with TODOs that are there for testing expedience.

skipping CI while in draft form

## Fixed Issue(s)

fixes #5846

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
@@ -23,6 +23,7 @@ import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.ethereum.bonsai.cache.CachedMerkleTrieLoader;
 import org.hyperledger.besu.ethereum.bonsai.cache.CachedWorldStorageManager;
 import org.hyperledger.besu.ethereum.bonsai.storage.BonsaiWorldStateKeyValueStorage;
+import org.hyperledger.besu.ethereum.bonsai.storage.flat.ArchiveFlatDbStrategy;
 import org.hyperledger.besu.ethereum.bonsai.trielog.TrieLogManager;
 import org.hyperledger.besu.ethereum.bonsai.worldview.BonsaiWorldState;
 import org.hyperledger.besu.ethereum.bonsai.worldview.BonsaiWorldStateUpdateAccumulator;
@@ -159,6 +160,19 @@ public class BonsaiWorldStateProvider implements WorldStateArchive {
     if (shouldPersistState) {
       return getMutable(blockHeader.getStateRoot(), blockHeader.getHash());
     } else {
+      // TODO this needs to be better integrated && ensure block is canonical
+      // HACK for kikori PoC, if we have the trielog for this block, we can assume we have it in
+      // flatDB
+      // although, in practice we can only serve canonical chain worldstates and need to fall back
+      // to state rolling if the requested block is a fork.
+      if (this.worldStateStorage.getFlatDbStrategy() instanceof ArchiveFlatDbStrategy
+          && trieLogManager.getTrieLogLayer(blockHeader.getBlockHash()).isPresent()) {
+
+        var contextSafeCopy = worldStateStorage.getContextSafeCopy();
+        contextSafeCopy.getFlatDbStrategy().updateBlockContext(blockHeader);
+        return Optional.of(new BonsaiWorldState(this, contextSafeCopy));
+      }
+
       final BlockHeader chainHeadBlockHeader = blockchain.getChainHeadHeader();
       if (chainHeadBlockHeader.getNumber() - blockHeader.getNumber()
           >= trieLogManager.getMaxLayersToLoad()) {
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -2111,7 +2111,7 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
 
     CommandLineUtils.failIfOptionDoesntMeetRequirement(
         commandLine,
-        "--Xsnapsync-synchronizer-flat option can only be used when -Xsnapsync-synchronizer-flat-db-healing-enabled is true",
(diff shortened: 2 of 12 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -159,6 +160,19 @@ public Optional<MutableWorldState> getMutable(
...
+      // HACK for kikori PoC, if we have the trielog for this block, we can assume we have it in
+      // flatDB
+      // although, in practice we can only serve canonical chain worldstates and need to fall back
+      // to state rolling if the requested block is a fork.
+      if (this.worldStateStorage.getFlatDbStrategy() instanceof ArchiveFlatDbStrategy
+          && trieLogManager.getTrieLogLayer(blockHeader.getBlockHash()).isPresent()) {
+
+        var contextSafeCopy = worldStateStorage.getContextSafeCopy();
+        contextSafeCopy.getFlatDbStrategy().updateBlockContext(blockHeader);
+        return Optional.of(new BonsaiWorldState(this, contextSafeCopy));
```
matkt, line 173: not sure why we are doing that . seems not clear to clone only the flatdb . It seems to be better to follow the logic of bonsai to clone all the storage if we need a clone and pass this clone to BonsaiWorldstate . like that no need to have a specific clone for flat db . 


just passing a clone to the worldstate and the internal flat db will check directly the snapshot automatically

I'm also not fan to clone everytime. we had problem before because of that . I think we can use the same code as before . using the clone in the cache and if not present create a new one (without rollback)

```java
return trieLogManager.getHeadWorldState(blockchain::getBlockHeader)
```
When we will have the checkpointed trie we will have to add again the rollback. maybe an optional rolling depending if we need the state or not

matkt: just saw that the current implemention is a hack for testing 👍

garyschulte: There isn't really any cloning exactly, contextSafeClone exists only to directly reuse the flatdb storage, but with a different bonsai context for a different block number.  For example we don't want RPC queries to modify the bonsai context, because that will mess up the block suffixing when we persist.  

We need to guard the bonsai context for the primary worldstate, but otherwise as long as we are not persisting, there is no need to snapshot or clone the database itself.

So clone is probably not the best name for this, better naming suggestions welcome :)

garyschulte: Perhaps we should have a different BonsaiWorldState subclass that better reflects that it is non-persisting and limited to flatdb only for reads

matkt: I think Layered storage is already doing this. It is not a clone and prevents modification. I would like to limit the number of classes to keep the same logic as much as possible. We also have the notion of Frozen BonsaiWorldState . The idea imo is to have a single worldstate and give a storage to this one. And at the storage level it will be different . Maybe I'm wrong but I think we have to think about this because we did the refactoring to reduce the number of worldstate class etc and I think that we must avoid redoing what we removed 


So I would say that we should have just a Layered Storage that wraps it and give it to the worldstate. And we freeze the worldstate as we are doing with bonsai before rpc call. Then find out if we should have a special worldstate for flat DB. I think we have to wait to see the implementation of the checkpointed trie. We could access in one worldstate to the flat db directly and the checkpointed trie in lazy mode.

matkt: Freeze's call is already wrapping the storage with layered storage https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldState.java#L592

Here 

https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/BonsaiWorldStateProvider.java#L177

matkt: Finally, I think that it is in the creation of the worldstate storage that we will give the context of the block number. 

BonsaiWorldstate doesn't need to know this.

garyschulte:
Context from conversation with @mattnelson, this implementation is a single column family implementation.  Read and write performance degrades as multiple versions 'pile up' in the worldstate column family.  A partitioned version of this implementation that uses separate column families for 'hot' (read/write) and a 'history' (read-only) should enable prevent degradation of execution performance, and should provide adequate historical RPC performance for use cases like block explorers.

@matkt, @jframe  - this would be like a manual checkpoint implementation, where the check point is dynamic, and we use trie logs to push old state into the history partition.

matkt:
> Context from conversation with @mattnelson, this implementation is a single column family implementation. Read and write performance degrades as multiple versions 'pile up' in the flat worldstate column families. A partitioned version of this implementation that uses separate column families for 'hot' (read/write) and a 'history' (read-only) should prevent degradation of execution performance, and should provide adequate historical RPC performance for use cases like block explorers.
> 
> @matkt, @jframe - this would be like a manual checkpoint implementation, where the check point is dynamic, and we use trie logs to push old state into the history partition.

seems to be a good idea

PR #5879, opened 2023-09-14
Tune G1GC to reduce Besu memory footprint

fab-10:
## PR description

These flags have been suggested by @ahamlat , and we tested them for months now, they have the effect of reducing the off heap memory used by Besu, without affecting the overall performance.


shemnon:
These changes looks reasonable, and should be fine in the default binary.  

However I expect some enterprises and non-mainnet versions may want to use different VMs, different GC, and different tuning.  Can we get a new `besu-untuned` launcher that has none of these tuning parameters enabled?  That way a non-mainnet environment can put their own tuning in via environmental parameters and not have them conflict with default settings.

fab-10:
@shemnon added the untuned script, first time I touched the create start scripts feature, please check if it is ok

shemnon (approved):

PR #5889, opened 2023-09-18
Code storage by hash

### LINKED ISSUE #5388: Change bytecode storage to be by hash as opposed to by address
Today, code storage is stored by account address, but we need to store by code hash. Duplicated smart contracts can point to the same hashed bytecode. We also must serve data by hash in the context of the snapsync server. 

* Do a comparison with a full sync node to compare the improvement in storage reduction
* Measure the performance impact of the change on syncing and block import

- [x] Change  code storage to be stored by code hash instead of account address
- [x] Backwards capability for code storage by account
- [x] Compare stored code values from existing column family to new column to ensure consistency
- [x] Performance test syncing to ensure read/write performance caused by changes to tracking code hashes by address have minimal impact
- [x] Experimental flag to enable code storage by code hash
- [x] Compare storage size taken by new column families with old storage mode
- [ ] Track code hashes by address so that self-destruct can removed when all code storage is unreferenced

jframe:
## PR description
Storing code by hash instead of by address. This is needed so that Besu can serve snapsync as we need to be able to retrieve the code by hash.

After syncing with this change it's not possible to downgrade to an older version of Besu. If we think this is needed I could add a subcommand to do the conversion.

**Testing**
- Checkpoint sync on Sepolia, Goerli and Mainnet
- Confirmed the same code bytes were stored for both column families by running a comparison tool on the stored code bytes https://github.com/hyperledger/besu/pull/6019/files#diff-9c4d660b9f0b067bbaf230c40ff79f2a53fef78e88c0c5e368b43e3891d0f68bR227
- Compatibility testing to ensure that existing databases continue to work
- Compared storage of code size on mainnet. Before: 9 GiB After: 2GiB. Column family to store code counts i.e. `CODE_HASH_COUNT` is using 38 MiB.
- Compared FCU and new payload times on mainnet with and without code changes showing no significant difference in times
Screenshot 2023-10-16 at 10 07 15 am

## Fixed Issue(s)

fixed #5388


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/DefaultCodeStorageStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Override
+  public void removeFlatCode(
+      final SegmentedKeyValueStorageTransaction transaction,
+      final Hash accountHash,
+      final Hash codeHash) {
+    final long codeHashCount = getCodeHashCount(transaction, codeHash);
+    final long updatedCodeHashCount =
+        codeHashCount > 0 ? codeHashCount - 1 : 0; // ensure count min value is 0
+    updateCodeHashCount(transaction, codeHash, updatedCodeHashCount);
```
matkt, line 59: do we want to also remove the counter when codeHashCount is zero ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/LegacyCodeStorageStrategy.java
siladu, line 1: Is this covered by some existing unit tests?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateTest.java
```diff
@@ -0,0 +1,140 @@
...
+      new BonsaiWorldState(
+          new BonsaiWorldStateLayerStorage(
+              new BonsaiWorldStateKeyValueStorage(
+                  new InMemoryKeyValueStorageProvider(), new NoOpMetricsSystem())),
+          cachedMerkleTrieLoader,
+          trieLogManager);
+
+  @ParameterizedTest
+  @MethodSource("priorAndUpdatedEmptyAndNullBytes")
+  void doesNothingWhenMarkedAsDeletedButAlreadyDeleted(final Bytes prior, final Bytes updated) {
```
siladu, line 68: nit: test names don't really indicate it's related to updating the code storage.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateTest.java
siladu, line 1: 👍

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
```diff
@@ -37,6 +39,7 @@ public class RocksDBTransaction implements SegmentedKeyValueStorageTransaction {
   private final Transaction innerTx;
   private final WriteOptions options;
   private final Function<SegmentIdentifier, ColumnFamilyHandle> columnFamilyMapper;
+  private final ReadOptions readOptions = new ReadOptions().setVerifyChecksums(false);
```
siladu, line 42: Why is this needed?

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
```diff
@@ -270,6 +270,24 @@ public void put(
...
+    public Optional<byte[]> get(final SegmentIdentifier segmentIdentifier, final byte[] key) {
+      final Optional<byte[]> updatedValue =
+          updatedValues
+              .computeIfAbsent(segmentIdentifier, __ -> new HashMap<>())
+              .getOrDefault(Bytes.wrap(key), Optional.empty());
+      final Set<Bytes> removedKeys =
+          this.removedKeys.computeIfAbsent(segmentIdentifier, __ -> new HashSet<>());
+
+      // consider the values in order of precedence of removed values, updated values, existing
+      // values
```
siladu, line 283: can the comment explain _why_ this precedence order is needed please? 

It's not clear to me why we need to query removedKeys at all, e.g. what if the same key was added a second time?

siladu (commented):
What regression testing has been done for FOREST mode? 

Will the new column families only be created for BONSAI mode - i.e. there should be no impact when downgrading a FOREST database?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/DefaultCodeStorageStrategy.java
jframe: good idea will add that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -248,15 +251,29 @@ private void updateCode(
               worldStateUpdater.getCodeToUpdate().entrySet()) {
             final Bytes updatedCode = codeUpdate.getValue().getUpdated();
             final Hash accountHash = codeUpdate.getKey().addressHash();
-            if (updatedCode == null || updatedCode.isEmpty()) {
-              bonsaiUpdater.removeCode(accountHash);
+            final Bytes priorCode = codeUpdate.getValue().getPrior();
+
+            // code hasn't changed then do nothing
+            if (Objects.equals(priorCode, updatedCode)
```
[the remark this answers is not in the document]
jframe: Didn't notice any performance impact on the engine API FCU or new payload times

jframe:
> Looks pretty good so far. Mostly just some questions and naming...
> 
> What regression testing has been done for FOREST mode?
> 
> Will the new column families only be created for BONSAI mode - i.e. there should be no impact when downgrading a FOREST database?

The new columns should only exist for BONSAI mode. There is no impact on FOREST as the existing column `CODE_STORAGE` also only applies to BONSAI.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/LegacyCodeStorageStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.plugin.services.storage.SegmentedKeyValueStorage;
+import org.hyperledger.besu.plugin.services.storage.SegmentedKeyValueStorageTransaction;
+
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes;
+
+public class LegacyCodeStorageStrategy implements CodeStorageStrategy {
```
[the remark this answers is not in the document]
jframe: That's a good idea, these names aren't great. Will probably steal your suggested names.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/LegacyCodeStorageStrategy.java
jframe: It's covered by the code storage tests in `BonsaiWorldStateKeyValueStorage`.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateTest.java
jframe: Good point missed will fix those names up

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
jframe: The get method I added to the transaction requires readOptions. So added it with the same configuration that are being used for `RocksDBColumnarKeyValueStorage`

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
jframe: I was basing this on the order in which we apply the changes in the commit method of `SegmentedKeyValueStorageTransaction` which applies the remove last. So a remove will always override an update being applied last.

The more I think about this I think this get method and the commit method are incorrect and we should be applying the operations in order that they made to the `SegmentedKeyValueStorageTransaction`.

non-fungible-nelson:
Reminder to revisit count

jframe:
@matkt @garyschulte Ready for another review when you get a chance. I've changed the PR to reuse the existing CODE_STORAGE column family instead of introducing two new column families and have added a `--Xsnapsync-server-enabled` to toggle the functionality.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -517,7 +541,10 @@ public Hash frontierRootHash() {
     return calculateRootHash(
         Optional.of(
             new BonsaiWorldStateKeyValueStorage.Updater(
-                noOpSegmentedTx, noOpTx, worldStateStorage.getFlatDbStrategy())),
+                noOpSegmentedTx,
+                noOpTx,
+                worldStateStorage.getFlatDbStrategy(),
+                worldStateStorage.getWorldStateStorage())),
```
garyschulte, line 547: We have CodeStorageStrategy, why do we need to expose composed worldstate storage ? in FlatDbStrategy we are not even using this param and instead use codeStorageStrategy.  

If we get rid of this unnecessary usage, we can remove the accessor from Bonsai WK KV storage class.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/FlatDbStrategy.java
```diff
@@ -172,15 +177,16 @@
   public void putFlatCode(
       final SegmentedKeyValueStorageTransaction transaction,
       final Hash accountHash,
-      final Bytes32 codeHash,
-      final Bytes code) {
-    transaction.put(CODE_STORAGE, accountHash.toArrayUnsafe(), code.toArrayUnsafe());
+      final Hash codeHash,
+      final Bytes code,
+      final SegmentedKeyValueStorage keyValueStorage) {
```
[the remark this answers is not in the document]
garyschulte: as mentioned elsewhere, this seems vestigial from prior implementations of code by hash

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/RollingImport.java
```diff
@@ -61,11 +60,7 @@ public static void main(final String[] arg) throws IOException {
     final SegmentedInMemoryKeyValueStorage worldStateStorage =
         (SegmentedInMemoryKeyValueStorage)
             provider.getStorageBySegmentIdentifiers(
-                List.of(
-                    ACCOUNT_INFO_STATE,
-                    CODE_STORAGE,
-                    ACCOUNT_STORAGE_STORAGE,
-                    TRIE_BRANCH_STORAGE));
+                List.of(ACCOUNT_INFO_STATE, ACCOUNT_STORAGE_STORAGE, TRIE_BRANCH_STORAGE));
```
garyschulte, line 63: I was going to call out that not including code storage in world state storage would make this test fail for any log rolling that included code.  But I don't know whether this class is even useful anymore.  It looks like it was part of an early prototype PR but isn't used anywhere anymore.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class BonsaiWorldStateTest {
```
garyschulte, line 46: nice to have explicit rather than just implicit bonsai worldstate tests 👍

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/storage/keyvalue/VariablesKeyValueStorageTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.hyperledger.besu.services.kvstore.InMemoryKeyValueStorage;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class VariablesKeyValueStorageTest {
```
garyschulte, line 26: 👍

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
```diff
@@ -57,6 +60,21 @@ public RocksDBTransaction(
...
+  public Optional<byte[]> get(final SegmentIdentifier segmentId, final byte[] key)
+      throws StorageException {
+    try (final OperationTimer.TimingContext ignored = metrics.getWriteLatency().startTimer()) {
+      return Optional.ofNullable(
+          innerTx.get(columnFamilyMapper.apply(segmentId), readOptions, key));
+    } catch (final RocksDBException e) {
+      if (e.getMessage().contains(NO_SPACE_LEFT_ON_DEVICE)) {
+        logger.error(e.getMessage());
+        System.exit(0);
+      }
```
garyschulte, line 73: I don't think a read can cause NO_SPACE_LEFT_ON_DEVICE 🤔

garyschulte (changes requested):
It looks like we are going to be deleting code by hash in this impl rather than leaving it as we discussed on discord?  

A few minor requested changes, generally LGTM

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
jframe: Yes, good point. Don't need that.

jframe:
> It looks like we are going to be deleting code by hash in this impl rather than leaving it as we discussed on discord?
> 
> A few minor requested changes, generally LGTM



> It looks like we are going to be deleting code by hash in this impl rather than leaving it as we discussed on discord?
> 
> A few minor requested changes, generally LGTM

Where I thought we got to was that by default we would delete code as that would be the best option for most users. But have a config option to not delete as the only case we wouldn't want to delete would be if we needed it for a future optimised tracing call. Didn't add that to this PR as it was getting quite messy with adding all the config options I thought I could do that in another PR once I had the mess sorted out. But at this point maybe yet another param isn't such an issue.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiSnapshotWorldStateKeyValueStorage.java
```diff
@@ -50,7 +50,8 @@ public BonsaiSnapshotWorldStateKeyValueStorage(
         parentWorldStateStorage.flatDbStrategy,
         segmentedWorldStateStorage,
         trieLogStorage,
-        metricsSystem);
+        metricsSystem,
+        parentWorldStateStorage.useCodeHashStorageMode);
```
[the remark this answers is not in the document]
jframe: Agree it is getting quite unruly with all those config options. Not sure of the right answer for cleaning it up right now. But definitely in another PR.

ethereum/core/src/test-support/java/org/hyperledger/besu/ethereum/core/InMemoryPrivacyStorageProvider.java
```diff
@@ -38,14 +37,6 @@ public static WorldStateArchive createInMemoryWorldStateArchive() {
...
   }
 
-  public static MutableWorldState createInMemoryWorldState() {
-    final InMemoryPrivacyStorageProvider provider = new InMemoryPrivacyStorageProvider();
-    return new DefaultMutableWorldState(
-        provider.createWorldStateStorage(),
-        provider.createWorldStatePreimageStorage(),
-        EvmConfiguration.DEFAULT);
-  }
-
```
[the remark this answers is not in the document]
jframe: Yeah it was dead code. Not related to the other changes. Just doing a bit of cleanup

jframe:
> > It looks like we are going to be deleting code by hash in this impl rather than leaving it as we discussed on discord?
> > A few minor requested changes, generally LGTM
> 
> > It looks like we are going to be deleting code by hash in this impl rather than leaving it as we discussed on discord?
> > A few minor requested changes, generally LGTM
> 
> Where I thought we got to was that by default we would delete code as that would be the best option for most users. But have a config option to not delete as the only case we wouldn't want to delete would be if we needed it for a future optimised tracing call. Didn't add that to this PR as it was getting quite messy with adding all the config options I thought I could do that in another PR once I had the mess sorted out. But at this point maybe yet another param isn't such an issue.

Changed agreed on from DM,
- On deleting code. Add that extra flag to disable deleting code. Will be enabled by default as is best for the majority of users
- Removing the --Xsnap-sync-server-enabled flag and adding a feature flag for this feature instead so it's not confusing with snap server needing to be enabled

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -1080,6 +1082,24 @@ WorldStateArchive createWorldStateArchive(
...
         trieLogPruner.initialize();
+
+        final boolean snapSyncServerEnabled =
+            syncConfig.getSnapSyncConfiguration().isServerEnabled();
+        variablesStorage
+            .isCodeStoredUsingCodeHash()
+            .ifPresentOrElse(
+                codeStoredUsingCodeHash -> {
+                  if (snapSyncServerEnabled && !codeStoredUsingCodeHash) {
+                    throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
jframe: Changed the feature to `--Xbonsai-code-using-code-hash-enabled` so it's confusing with snap sync server needs to be enabled.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/RollingImport.java
jframe: Wasn't sure what this was for. Think it's an old testing tool? It's in the test package so wasn't worried about not including the code storage in world state storage.

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -90,6 +90,14 @@ static class Unstable {
         description =
             "Enables code storage using code hash instead of by account hash. (default: ${DEFAULT-VALUE})")
     private boolean bonsaiCodeUsingCodeHashEnabled = DEFAULT_BONSAI_CODE_USING_CODE_HASH_ENABLED;
+
+    @CommandLine.Option(
+        hidden = true,
+        names = {"--Xbonsai-code--using-code-hash-deletion-enabled"},
```
garyschulte, line 97: nit: double dash

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -76,6 +76,7 @@ public class BonsaiWorldStateKeyValueStorage implements WorldStateStorage, AutoC
 
   protected final ObservableMetricsSystem metricsSystem;
   protected final boolean useCodeHashStorageMode;
+  protected final boolean deleteCodeInCodeHashStorageMode;
```
garyschulte, line 79: ultra descriptive variable name :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueStorageProviderBuilder.java
```diff
@@ -61,6 +68,7 @@ public KeyValueStorageProvider build() {
     return new KeyValueStorageProvider(
         segments -> storageFactory.create(segments, commonConfiguration, metricsSystem),
         worldStatePreImageStorage,
-        (ObservableMetricsSystem) metricsSystem);
+        (ObservableMetricsSystem) metricsSystem,
+        deleteCodeInCodeHashStorageMode);
```
garyschulte, line 72: We are leaking bonsai configuration concerns into the storage interfaces.  Can we not derive this in BonsaiWorldStateKeyValueStorage the same way we derive flatdb mode?  

We don't have (or shouldn't have) the synchronizer to make this db configuration for us the way flat db does.  I do wish we could push this into bonsai specific config rather than making it part of the storage provider interface.  

DataStorageFormat enum *could* have a defaulted member var for this config, and we could override it with a fluent function in the DataStorageFormat.  That somewhat breaks the concept of the enum being immutable.  

perhaps we just go with this and fast-follow with a PR that encapsulates storage config options.

garyschulte (approved):
Not a fan of leaking bonsai config into storage interfaces, but in the interest of expedience and to limit the PR scope, 🚢  and we can fast-follow with a storage option config encapsulation pr

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueStorageProviderBuilder.java
```diff
@@ -48,6 +49,12 @@ public KeyValueStorageProviderBuilder withMetricsSystem(final MetricsSystem metr
     return this;
   }
 
+  public KeyValueStorageProviderBuilder withDeleteCodeInCodeHashStorageMode(
```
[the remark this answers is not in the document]
jframe: Missed adding that in have wired in now

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
jframe: done

jframe:
> Not a fan of leaking bonsai config into storage interfaces, but in the interest of expedience and to limit the PR scope, 🚢 and we can fast-follow with a storage option config encapsulation pr

Don't like it either. But I think the alternative is to do some refactoring around how the `BonsaiWorldStateKeyValueStorage` gets initialised. Would like to clean that up. But prefer to do that in another PR so this PR doesn't get any bigger.

garyschulte (approved):
👍

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
```diff
@@ -47,6 +47,8 @@ interface Unstable {
     long DEFAULT_BONSAI_TRIE_LOG_RETENTION_THRESHOLD = 512L;
     long MINIMUM_BONSAI_TRIE_LOG_RETENTION_THRESHOLD = DEFAULT_BONSAI_TRIE_LOG_RETENTION_THRESHOLD;
     int DEFAULT_BONSAI_TRIE_LOG_PRUNING_LIMIT = 30_000;
+    boolean DEFAULT_BONSAI_CODE_USING_CODE_HASH_ENABLED = false;
+    boolean DEFAULT_BONSAI_CODE_USING_CODE_HASH_DELETION_ENABLED = true;
```
gfukushima, line 51: shouldn't the default be false here too?

jframe: I want the default to be true as generally, we should delete code from storage as part of a self-destruct. The special case is for tracing which most people won't use where could in the future use the code along with the trie logs to do a trace without needing to roll the world state.

gfukushima: My main concern is, that this is a change in the current behavior that is not documented anywhere because this PR isn't adding any changelog since this is introduced under an experimental flag, is my understanding correct?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/storage/flat/CodeHashCodeStorageStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+      final Bytes code) {
+    transaction.put(CODE_STORAGE, prefixKey(CODE_PREFIX, codeHash), code.toArrayUnsafe());
+  }
+
+  @Override
+  public void removeFlatCode(
+      final SegmentedKeyValueStorageTransaction transaction,
+      final Hash accountHash,
+      final Hash codeHash) {
+    // TODO JF add reference counting so that code can be removed when there are no more usages
```
garyschulte, line 51: 👍

jframe: good idea, I'll update the comment with the issue

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/storage/flat/FlatDbStrategy.java
```diff
@@ -53,9 +53,14 @@ public abstract class FlatDbStrategy {
...
   protected final Counter getStorageValueFlatDatabaseCounter;
+  private final CodeStorageStrategy codeStorageStrategy;
 
-  public FlatDbStrategy(final MetricsSystem metricsSystem) {
+  public FlatDbStrategy(final MetricsSystem metricsSystem, final boolean useCodeStoredByHash) {
     this.metricsSystem = metricsSystem;
+    this.codeStorageStrategy =
+        useCodeStoredByHash
+            ? new CodeHashCodeStorageStrategy()
+            : new AccountHashCodeStorageStrategy();
```
garyschulte, line 64: can we move this up into FlatDbStrategyProvider, and pass a CodeStorageStrategy rather than a boolean ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/storage/flat/FlatDbStrategyProvider.java
```diff
@@ -50,10 +52,15 @@ public void loadFlatDbStrategy(final SegmentedKeyValueStorage composedWorldState
     // if  flatDbMode is not loaded or has changed, reload flatDbStrategy
     if (this.flatDbMode == null || !this.flatDbMode.equals(newFlatDbMode)) {
       this.flatDbMode = newFlatDbMode;
+      final boolean isCodeStoredByHashEnabled =
+          dataStorageConfiguration.getUnstable().getBonsaiCodeStoredByCodeHashEnabled();
+      LOG.info(
+          "Bonsai flat db mode with code stored using code hash enabled = {}",
+          isCodeStoredByHashEnabled);
```
garyschulte, line 59: the risk of driving this from command line only is that we can fall out of consensus if the user changes this config parameter without resyncing.  It isn't fun to plumb in the check  but we should try to derive the codehash vs accounthash as it exists in the database and only allow the cli parameter to affect an as-yet-unsynced node.

we already have the notion of using `loadFlatDbStrategy(..)` in BonsaiWorldStateKeyValueStorage, and it will override the CLI supplied flat db strategy based on what it finds.  We can work this check into that method and ensure that we use the correct code storage strategy based on what is in the db, and perhaps log a big ugly warning message if the CLI param does not match the db storage

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/trie/bonsai/worldview/BonsaiWorldStateTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class BonsaiWorldStateTest {
```
garyschulte, line 46: IMO we should add a test that asserts the correct FlatDBStrategy (with the correct CodeStorageStrategy) is constructed when the db storage and the cli param disagree.

garyschulte (changes requested):
Looks good, the only significant concern is the ability for code storage CLI param to affect besu's worldstate consistency when it is initially synced with a different code storage method

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/storage/flat/FlatDbStrategy.java
jframe: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/storage/flat/FlatDbStrategyProvider.java
jframe: Added detection code in the `FlatDbStrategyProvider` which will warn if there is an inconsistency and use the strategy in the database.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/trie/bonsai/worldview/BonsaiWorldStateTest.java
jframe: Added some tests for that in the `FlatDbStrategyProvider` since that's where it's created

garyschulte (approved):
🚢   great test coverage

non-fungible-nelson:
💯 🚀

PR #5890, opened 2023-09-18
Update holesky with fixed extraData, genesis time, shanghaiTime

siladu:
Incorporates https://github.com/eth-clients/holesky/pull/73
and also @fab-10's changes from https://github.com/eth-clients/holesky/pull/72


gfukushima (commented):
LGTM, double-checked against the expected block hash and it matches:
0xb5f7f912443c940f21fd611f12828d75b534364ed9e95ca4e307729a4661bde4

I'm going to wait until https://github.com/eth-clients/holesky/pull/73 gets merged to approve this.

garyschulte (commented):
It looks like the missing brace would have parsed the remaining accounts as storage fields?  I presume this was possible because of a json parsing policy that ignores missing fields, otherwise we should have thrown when "balance":xx wasn't part of the json schema for storage fields.

can we add stricter json parsing for storage fields and/or test coverage?

siladu:
@fab-10 I have some doubt about the removal of cancunTime (which is currently breaking the tests). 

Eth-clients genesis didn't have it but geth did: https://github.com/ethereum/go-ethereum/blob/90d5bd85bcf2919ac2735a47fde675213348a0a6/params/config.go#L84

Nethermind didn't: https://github.com/NethermindEth/nethermind/blob/8ce10d324c05a44dc1fb9e07e6b8618350addac9/src/Nethermind/Nethermind.Specs/HoleskySpecProvider.cs#L15

I think it's probably fine either way...just not sure if "fork next" having a difference value affects peering

besu/src/test/java/org/hyperledger/besu/ForkIdsNetworkConfigTest.java
```diff
@@ -84,10 +84,9 @@ public static Collection<Object[]> parameters() {
           new Object[] {
             NetworkName.HOLESKY,
             List.of(
-                new ForkId(Bytes.ofUnsignedInt(0xa200f558L), 1694790240L),
-                new ForkId(Bytes.ofUnsignedInt(0x840a3b53L), 2000000000L),
-                new ForkId(Bytes.ofUnsignedInt(0x30771f90), 0L),
-                new ForkId(Bytes.ofUnsignedInt(0x30771f90), 0L))
+                new ForkId(Bytes.ofUnsignedInt(0xc61a6098L), 1696000704L),
```
siladu, line 87: Couldn't find a reference for this, would be good to cross reference with other clients once they've made the changes

fab-10:
> @fab-10 I have some doubt about the removal of cancunTime (which is currently breaking the tests).
> 
> Eth-clients genesis didn't have it but geth did: https://github.com/ethereum/go-ethereum/blob/90d5bd85bcf2919ac2735a47fde675213348a0a6/params/config.go#L84
> 
> Nethermind didn't: https://github.com/NethermindEth/nethermind/blob/8ce10d324c05a44dc1fb9e07e6b8618350addac9/src/Nethermind/Nethermind.Specs/HoleskySpecProvider.cs#L15
> 
> I think it's probably fine either way...just not 100% sure if "fork next" having a different value affects peering

Ok, now I remember why I added it in the first place, so should I readd it or have you found a way to fix the tests?

siladu:
> Ok, now I remember why I added it in the first place, so should I readd it or have you found a way to fix the tests?

@fab-10 I fixed the tests and confirmed with Stefan that a future cancunTime shouldn't affect peering so think we're good.
Would still like to confirm with other EL clients if the forkId is the expected value before we re-launch

jframe:
> It looks like the missing brace would have parsed the remaining accounts as storage fields? I presume this was possible because of a json parsing policy that ignores missing fields, otherwise we should have thrown when "balance":xx wasn't part of the json schema for storage fields.
> 
> can we add stricter json parsing for storage fields and/or test coverage?

Definitely a good idea to do that. Happy for that to be in another PR to not hold the release for the Holesky relaunch. Even better would be to be compatible with the same genesis config that other clients use.

macfarla:
Created this issue for tightening up the parsing https://github.com/hyperledger/besu/issues/5900

config/src/main/resources/holesky.json
```diff
@@ -827,196 +826,196 @@
...
+    },
+    "0xC9CA2bA9A27De1Db589d8c33Ab8EDFa2111b31fb": {
+      "balance": "0xd3c21bcecceda1000000"
+    },
+    "0x4BC656B34De23896fa6069C9862F355b740401aF": {
+      "balance": "0x084595161401484a000000"
     }
   },
   "coinbase": "0x0000000000000000000000000000000000000000",
   "baseFeePerGas": "0x3B9ACA00",
```
jframe, line 1013: Why does the Besu config need `baseFeePerGas` but the default genesis.json doesn't?

siladu: Just double checked and we technically don't because it is the default initial value: https://github.com/hyperledger/besu/blob/main/config/src/main/java/org/hyperledger/besu/config/GenesisConfigFile.java#L43

However, if we change it in this PR then we will no longer exactly match https://github.com/eth-clients/holesky/blob/main/custom_config_data/besu.json if anyone cares about that.

I actually prefer it being in, more explicit, since it is actually part of the genesis block.

garyschulte (approved):
🚢  LGTM.  AFAIK, next fork shouldn't present a peering problem until we 'get there'

PR #5892, opened 2023-09-19
Add updated storage to evmtool json trace

shemnon:
## PR description

Add the EIP-3155 "storage" option to the standard tracer, with the caveat only updated storage is logged.


macfarla (approved):
LGTM

Issue #5898, opened 2023-09-19
ConcurrentModificationException in Bonsai Tries durring reference tests


shemnon:
### Description

While I was updating the reference tests I got a transient `ConcurrentModificationException` in one of the test cases.

```
ava.lang.RuntimeException: Error parsing test case file /Users/dannoferrin/git/github.com/shemnon/besu/ethereum/referencetests/build/resources/referenceTest/GeneralStateTests/stBadOpcode/opcE0DiffPlaces.json to class class org.hyperledger.besu.ethereum.referencetests.GeneralStateTestCaseSpec
	at org.hyperledger.besu.testutil.JsonTestParameters.parseFile(JsonTestParameters.java:299)
	at org.hyperledger.besu.testutil.JsonTestParameters.generate(JsonTestParameters.java:247)
	at org.hyperledger.besu.testutil.JsonTestParameters.generate(JsonTestParameters.java:235)
	at org.hyperledger.besu.ethereum.vm.GeneralStateReferenceTestTools.generateTestParametersForConfig(GeneralStateReferenceTestTools.java:124)
	at org.hyperledger.besu.ethereum.vm.generalstate.GeneralStateReferenceTest_215.getTestParametersForConfig(GeneralStateReferenceTest_215.java:23)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.params.provider.MethodArgumentsProvider.lambda$provideArguments$1(MethodArgumentsProvider.java:46)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
[...]

PR #5899, opened 2023-09-19
Update reference tests to 12.4

shemnon:
## PR description

Update reference tests to 12.4
* Some in-memory storage changed to ConcurrentMap
* exclude cancun from all EIP tests, EIP-4788 still in flux
* Add new fields to ReferenceTestEnv, and re-order the reflected constructor for clarity and ease of development.

## Fixed Issue(s)

includes speculative changes for #5898 concurrency issue.  `services/kvstore/...` can be dropped out if there are other fixes, but transient failures result in blockchain and general state tests when removed


matkt (approved):
I don't know why I decided not to put a concurrenthashmap. anyway it's good for me

macfarla (approved):

PR #5887, opened 2023-09-16
SnapServer using FlatDB Strategy

garyschulte:
## PR description
PR to implement SnapServer using FlatDB Strategy.   

marked Draft until we have code storage by code hash rolled out.

## Fixed Issue(s)

#3165

Issue #5864, opened 2023-09-11
Implement multiversion flat db reader & writer strategy


garyschulte:
### Description
Implement a multi-version flat db and context, such that we can access any historical state from the bonsai flat db.

### Acceptance Criteria
* mergable, working multi-version flat db format
* metrics, diff of flat db and multi-version flat
  * size difference
  * write performance diff
  * read performance diff
  * block execution diff

Issue #3165, opened 2021-12-13
Implement snapsync as a server


matkt:
The snapsync as a server must be fixed and certainly optimized. Check if our implementation is good enough regarding DOS protection. And check the Bonsai database issue 
To see if we want to implement snapshot or if we want to find a way to use bonsai for that

non-fungible-nelson:
blocked on #5319

siladu:
@garyschulte Create a story for conversion piece please

PR #5901, opened 2023-09-19
Add FlatDbStrategy

garyschulte:
## PR description
This is a prerequisite PR for #5865 and #5887.  

It:
* expands the concept of FlatDbReaderStrategy to FlatDbStrategy, including write operations.
* adds getNearestTo implementation to SegmentedKeyValueStorage and its implementations
* adds getNearestTo test coverage and expands KeyValueStorage tests to SegmentedKeyValueStorage implementations

This PR will allow for parallel progress on SnapServer server and Bonsai Archive, and mitigate merge conflicts that would otherwise arise

## Fixed Issue(s)

related to #3165 and #5864


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/FlatDbStrategy.java
```diff
@@ -117,6 +118,65 @@ public Optional<Bytes> getCode(
...
+
+  /*
+   * Removes the storage value for the given account hash and storage slot key, using the world state root hash supplier, storage root supplier, and node loader.
+   */
+  public void removeFlatAccountStorageValueByStorageSlotHash(
+      final SegmentedKeyValueStorageTransaction transaction,
+      final Hash accountHash,
+      final Hash slotHash) {
+    transaction.remove(
+        ACCOUNT_STORAGE_STORAGE, Bytes.concatenate(accountHash, slotHash).toArrayUnsafe());
```
macfarla, line 158: why is it called ACCOUNT_STORAGE_STORAGE ? there's no ACCOUNT_STORAGE, right?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/ArchiveFlatDbReaderStrategyTest.java
```diff
@@ -0,0 +1,18 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+package org.hyperledger.besu.ethereum.bonsai.storage.flat;
+
+public class ArchiveFlatDbReaderStrategyTest {}
```
matkt, line 18: is this needed in this PR ?

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -92,6 +92,26 @@ public Optional<byte[]> get(final SegmentIdentifier segmentId, final byte[] key)
     }
   }
 
+  @Override
+  public Optional<NearestKeyValue> getNearestTo(
+      final SegmentIdentifier segmentIdentifier, final Bytes key) throws StorageException {
+    var ourNearest = super.getNearestTo(segmentIdentifier, key);
+    var parentNearest = parent.getNearestTo(segmentIdentifier, key);
```
matkt, line 99: I'm not a fan of var. it not easy to understand the type we have

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
```diff
@@ -111,6 +112,30 @@ public Optional<byte[]> get(final SegmentIdentifier segmentIdentifier, final byt
     }
   }
 
+  @Override
+  public Optional<NearestKeyValue> getNearestTo(
```
matkt, line 116: I think we need to find another way to do that. it will be huge work for snapsync as a server

matkt (changes requested):

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/ArchiveFlatDbReaderStrategyTest.java
garyschulte: Nope, will remove 👍

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
garyschulte: will be explicit.  👍

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
garyschulte: Although we could use getNearest and return an iterator for snap sync server, I am thinking streamFlatAccounts and streamFlatStorage should work better and more easily for snapsync server.  

getNearest is really just for point-wise access to keys when in archive mode.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/FlatDbStrategy.java
garyschulte: yeah, not my favorite name, but it is what the segment enum calls it.  It is the underlying storage for account storage, but makes for a weird name.  I am all for renaming the flat segments to ACCOUNT_INFO_FLAT, ACCOUNT_STORAGE_FLAT, and CODE_STORAGE_FLAT or something else that is more meaningful.  I'd rather do that in another PR though since it would touch a lot of places

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -92,6 +92,26 @@ public Optional<byte[]> get(final SegmentIdentifier segmentId, final byte[] key)
...
+    if (ourNearest.isPresent() && parentNearest.isPresent()) {
+      // Both are present, return the one closer to the key
+      int ourDistance = ourNearest.get().key().commonPrefixLength(key);
+      int parentDistance = parentNearest.get().key().commonPrefixLength(key);
+      return (ourDistance <= parentDistance) ? ourNearest : parentNearest;
+    } else if (ourNearest.isPresent()) {
+      // Only ourNearest is present
+      return ourNearest;
+    } else {
+      // Either both are not present, or only parentNearest is present
```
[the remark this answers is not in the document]
garyschulte: yes, if neither are present there was nothing at or before the requested key and we return empty. Callers should expect to handle the optional empty return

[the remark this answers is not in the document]
garyschulte: > does this end up being Optional.empty if neither are present? What happens then?
This question made me add another case that checks we correctly return Optional.empty().  Though the rocksdb nearestTo would behave that way, the In-memory comparators are not.  adding a test and added a filter to the streaming implementation.  👍

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
matkt: ah yes got it not sure why I was thinking this one will be used for snapsync :)

matkt (approved):
LGTM

PR #5916, opened 2023-09-21
renamed PayloadTuple and made a separate class

macfarla:
adds flexibility for adding more info to the Payload object eg OperationTracer to be used by plugins


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PayloadWrapper.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.consensus.merge;
+
+import org.hyperledger.besu.consensus.merge.blockcreation.PayloadIdentifier;
+import org.hyperledger.besu.ethereum.core.BlockWithReceipts;
+
+/** Wrapper for payload plus extra info. */
+public class PayloadWrapper {
```
fab-10, line 21: it seems it could be a record

fab-10 (approved):

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/PayloadWrapper.java
macfarla: done

Gabriel-Trintinalia (approved):
lgtm

PR #5917, opened 2023-09-21
fix geth rlpx ping command

pinges:
## PR description
The geth cmd 
./build/bin/devp2p rlpx ping enodeAddress
expects the client with address enodeAddress to respond with a HelloMessage right after successfully receiving the handshake message from the initiator.
Existing behaviour in Besu is to wait for the HelloMessage from the initiator before sending the HelloMessage back.


ethereum/referencetests/src/reference-test/external-resources
daniellehrner, line 1: It seems the reference tests have been updated by accident.

daniellehrner (changes requested):
Looks good to me. The only thing that need to be reverted is the update of the reference tests.

macfarla (approved):

PR #5920, opened 2023-09-21
Improve performance when promoting transaction from next layers

fab-10:
## PR description

During the implementation of #5921, a performance issue surfaced in the way the promotion of transactions, from lower layers, was implemented.
The issue was that, to keep the code simple, the promotion was done for one tx at once, but with the improvement on the prioritized layer done #5921, this approach is no more practical since result in a quadratic complexity (number of confirmed txs per number of senders in the ready layer), so the solution is to do the promotion only once after all the confirmed txs have been processed, so the time is linear.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/SparseTransactions.java
```diff
@@ -126,36 +125,86 @@ protected void internalReplaced(final PendingTransaction replacedTx) {
...
-      } else {
-        final long firstNonce = senderTxs.firstKey();
-        final int newGap = (int) (firstNonce - (promotedTx.getNonce() + 1));
-        if (newGap != 0) {
-          updateGap(sender, 0, newGap);
+  public List<PendingTransaction> promote(
+      final Predicate<PendingTransaction> promotionFilter,
+      final long freeSpace,
+      final int freeSlots) {
+    long accSpace = 0;
```
macfarla, line 143: usedSpace?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/EndLayer.java
```diff
@@ -115,8 +115,9 @@ public OptionalLong getCurrentNonceFor(final Address sender) {
   }
 
   @Override
-  public PendingTransaction promote(final Predicate<PendingTransaction> promotionFilter) {
-    return null;
+  public List<PendingTransaction> promote(
+      final Predicate<PendingTransaction> promotionFilter, final long l, final int freeSlots) {
```
macfarla, line 119:
```suggestion
      final Predicate<PendingTransaction> promotionFilter, final long freeSpace, final int freeSlots) {
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractPrioritizedTransactions.java
```diff
@@ -104,8 +105,9 @@ protected void internalRemove(
   }
 
   @Override
-  public PendingTransaction promote(final Predicate<PendingTransaction> promotionFilter) {
-    return null;
+  public List<PendingTransaction> promote(
+      final Predicate<PendingTransaction> promotionFilter, final long l, final int freeSlots) {
```
macfarla, line 109:
```suggestion
      final Predicate<PendingTransaction> promotionFilter, final long freeSpace, final int freeSlots) {
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/ReadyTransactions.java
```diff
@@ -139,30 +141,51 @@ public Stream<PendingTransaction> stream() {
...
+    final List<PendingTransaction> promotedTxs = new ArrayList<>();
+
+    // first find all txs that can be promoted
+    search:
+    for (final var senderFirstTx : orderByMaxFee.descendingSet()) {
+      final var senderTxs = txsBySender.get(senderFirstTx.getSender());
+      for (final var candidateTx : senderTxs.values()) {
+        if (promotionFilter.test(candidateTx)) {
+          accSpace += candidateTx.memorySize();
+          if (promotedTxs.size() < freeSlots && accSpace <= freeSpace) {
```
macfarla, line 158: maybe a comment to explain why < for slots and <= for space

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/SparseTransactions.java
macfarla: allocatedSpace?

fab-10: here `acc` stays for accumulated space, to keep how much space has been accumulated by the selected txs till now, do you think it is better to rename it?

macfarla: accumulated also works

macfarla (approved):
makes sense!

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/SparseTransactions.java
fab-10: renaming to `accumulatedSpace`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/ReadyTransactions.java
fab-10: can I ask what you get from the code about the difference? to understand what needs clarification

macfarla: freeSlots and freeSpace have similar names - how much is free - so why does the if statement check < for one and <= for the other?

why not <= for freeSlots?

fab-10: slots always increment by 1, while for space it could be any amount, so you can read it like that

`if (promotedTxs.size() + 1 <= freeSlots && accumulatedSpace <= freeSpace)`

but the `+1` is redundant knowing that `promotedTxs.size()` can only increment by 1 at time, and we can shortcut with `<`

macfarla: right so it's because the accumulatedSpace is already incremented at this point but for the slots it happens after. makes sense. it's ok for me :)

PR #5921, opened 2023-09-21
Always enforce promotion filter for transactions in the prioritized layer

fab-10:
## PR description

This PR is built on top of #5920, so please check it first ([relative changes](https://github.com/fab-10/besu/compare/fab-10:besu:txpool-promotion-performance...prioritized-promotion-filter2))

This PR is preparatory for more work on prioritizing local txs, that consist in more small PRs that have been split to make it easier to review.
Before the PR, the prioritized layer also kept txs that could not be selected to be included in the next block, because as far as there was still space txs were added and kept there, regardless they were or not satisfying the promotion filter anymore.
This PR improves the content of the prioritized layer, keeping there only txs that always satisfy the promotion filter, and thus are candidate for inclusion in the next block, this avoid to waste some work during txs selection, and pave the way for an easier prioritization of local txs.
Currently the promotion filter says that a tx is willing to pay at list the base fee, on base fee market network, otherwise is always true.

Simulation of block production shows no regression, actually metrics about number of txs in the block, block value and gas used are, on average, better with this PR vs 23,7,2 version:
image

image

image


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractPrioritizedTransactions.java
```diff
@@ -77,6 +82,12 @@ protected void internalReplaced(final PendingTransaction replacedTx) {
   }
 
   private boolean hasPriority(final PendingTransaction pendingTransaction) {
+    // if it does not pass the promotion filter, then has not priority
+    if (!promotionFilter(pendingTransaction)) {
+      return false;
+    }
+
+    // if there is space add it, otherwise check if it has move value than the last one
```
macfarla, line 90:
```suggestion
    // if there is space add it, otherwise check if it has more value than the last one
```

macfarla (approved):

PR #5930, opened 2023-09-22
Add parameters to EVM library fluent API

shemnon:
## PR description

Add the ability to configure more parameters in the fluent API. Specifically contract address, coinbase, difficulty, mixHash/prevRandao, baseFee, block number, timestamp, gas limit, previous block hashes, and versioned hashes.


evm/src/main/java/org/hyperledger/besu/evm/MainnetEVMs.java
```diff
@@ -413,11 +424,16 @@ public static EVM constantinople(final EvmConfiguration evmConfiguration) {
    */
   public static EVM constantinople(
       final GasCalculator gasCalculator, final EvmConfiguration evmConfiguration) {
+    var version = EvmSpecVersion.CONSTANTINOPLE;
+    return constantiNOPEl(gasCalculator, evmConfiguration, version);
+  }
+
+  private static EVM constantiNOPEl(
```
garyschulte, line 431: Even old New York was once New Amsterdam

non-fungible-nelson: giantCase

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleBlockValues.java
```diff
@@ -15,7 +15,105 @@
...
+  long timestamp = 1;
+  long gasLimit = Long.MAX_VALUE;
+
+  @Override
+  public Bytes getDifficultyBytes() {
+    return difficultyBytes;
+  }
+
+  /**
+   * Sets teh difficulty of the block
```
garyschulte, line 42: not sure if typo or just for color.

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleBlockValues.java
```diff
@@ -15,7 +15,105 @@
...
 import org.hyperledger.besu.evm.frame.BlockValues;
 
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+
 /** A concrete BlockValues object that takes all the defaults */
-public class SimpleBlockValues implements BlockValues {}
+public class SimpleBlockValues implements BlockValues {
```
garyschulte, line 27: non-blocking ofc,  but this class is not particularly fluent

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
```diff
@@ -107,11 +120,16 @@ public void revert() {
 
   @Override
   public void commit() {
-    parent.accounts.putAll(accounts);
+    accounts.forEach(
+        (address, account) -> {
+          if (!account.commit()) {
+            accounts.put(address, account);
+          }
+        });
```
garyschulte, line 128: this is peculiar - if we cannot commit the account, we re-add it back to accounts?  should this be `parent.accounts.put(address,account)` ?

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
```diff
@@ -60,8 +60,9 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
     final boolean beneficiaryIsWarm =
         frame.warmUpAddress(beneficiaryAddress) || gasCalculator().isPrecompile(beneficiaryAddress);
 
-    final Address originatorAddress = frame.getRecipientAddress();
-    final Wei originatorBalance = frame.getWorldUpdater().get(originatorAddress).getBalance();
+    final Address originatorAddresss = frame.getRecipientAddress();
+    final MutableAccount originatorAccount = frame.getWorldUpdater().getAccount(originatorAddresss);
```
garyschulte, line 64: spelling nitpick (sss).  also originatorAddress is unused apart from getAccount.

garyschulte (approved):
LGTM, minor nits and question regarding commit in SimpleWorld but otherwise GTG

evm/src/main/java/org/hyperledger/besu/evm/operation/SelfDestructOperation.java
shemnon: Had it fixed in another stash. ;)

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
shemnon: Also yes, the parent.
SimpleWorld lacks deep testing and is meant for "toy" environments.  I would like to run general world state reference tests against the EVM lib alone, but there would be some ref tests duplication.  And we need a fake data store substrate, which is what this fills.  Real testing will be when ref tests pass.

garyschulte (approved):
🚢

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleBlockValues.java
shemnon: The EVMExecutor is fluent, and sets it via fluent .  This is then passed into the EVM.  Nested fluency seems pointless, and the block values hold the data rather than the executor

https://github.com/hyperledger/besu/pull/5930/files#diff-c691ed5f63ebd8d1a3908a282b65313fdea66411c51a77b2e811b5ba9c09e27fR873

PR #5933, opened 2023-09-25
Cache last blocks data (block headers, block bodies, transactions' receipts and total difficulty)

ahamlat:
## PR description
Cache last blocks data (block headers, block bodies, transactions' receipts and total difficulty).
The default number of blocks to keep in the cache is 512.


ahamlat:
#6009

PR #5940, opened 2023-09-26
[4844] [Hive] Fix fcuV3 parameter return 

Gabriel-Trintinalia:
## PR description
Fix some hive tests

- Return JsonRpcErrorResponse instead of JsonRpcSuccessResponse when fork not supported or parameters invalid.
- Validate parameter before fork validation
- Validate getParentBeaconBlockRoot on EngineForkchoiceUpdatedV3

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdatedV3.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdatedV3.java
@@ -18,6 +18,7 @@ import org.hyperledger.besu.consensus.merge.blockcreation.MergeMiningCoordinator
 import org.hyperledger.besu.ethereum.ProtocolContext;
 import org.hyperledger.besu.ethereum.api.jsonrpc.RpcMethod;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.EngineForkchoiceUpdatedParameter;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.parameters.EnginePayloadAttributesParameter;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.RpcErrorType;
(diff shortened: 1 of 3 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdatedV3.java
```diff
@@ -48,14 +49,21 @@ public String getName() {
...
+      final EngineForkchoiceUpdatedParameter fcuParameter,
+      final Optional<EnginePayloadAttributesParameter> maybePayloadAttributes) {
     if (fcuParameter.getHeadBlockHash() == null) {
       return ValidationResult.invalid(RpcErrorType.INVALID_PARAMS, "Missing head block hash");
     } else if (fcuParameter.getSafeBlockHash() == null) {
       return ValidationResult.invalid(RpcErrorType.INVALID_PARAMS, "Missing safe block hash");
     } else if (fcuParameter.getFinalizedBlockHash() == null) {
       return ValidationResult.invalid(RpcErrorType.INVALID_PARAMS, "Missing finalized block hash");
     }
+    if (maybePayloadAttributes.isPresent()) {
```
macfarla, line 61: this is the parameter checking you mean that will potentially conflict with the larger refactor?

macfarla (approved):
LGTM

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineForkchoiceUpdatedV3.java
macfarla: I think fine to fix it like this in the short-term to clear the noise for the hive tests

Gabriel-Trintinalia: We would not necessarily have to validate this if the plans are to use strict JSON validation

PR #5941, opened 2023-09-26
updated beacon root and modulus to match DRAFT eip

macfarla:
Still draft but this time it _says_ final https://github.com/ethereum/EIPs/pull/7672#issuecomment-1734562996

Also renamed the modulus value so it matches the name used in the EIP


macfarla:
last updated here https://github.com/hyperledger/besu/pull/5903

jflo (approved):

jflo:
I'm doing it again!

PR #5948, opened 2023-09-26
Transaction pool unit tests refactoring to remove duplicated code

fab-10:
## PR description

Refactor transaction pool tests to remove all the duplicated code that was introduced when implementing the layered txpool.
Now all tests are defined in `AbstractTransactionPoolTest` and extending classes are used to setup the right env for fee market and txpool implementation.
This make the tests more future proof, since in this way a new test is run against all the implementations and fee markets, while before you had to remember to code the test twice.


macfarla (commented):
LGTM

macfarla (approved):

PR #5955, opened 2023-09-27
rlpx - Send empty list instead of Empty Bytes for the Ping and Pong message

Gabriel-Trintinalia:
## PR description
Send empty list instead of Empty Bytes for Ping and Pong message
https://github.com/ethereum/devp2p/blob/master/rlpx.md#ping-0x02


macfarla (approved):
LGTM

PR #5961, opened 2023-09-28
Apply fcu even on invalid payload

jflo:
Fixes more Hive tests when a payload that is invalid is passed along with an FCU


macfarla (approved):
LGTM. just spotless

fab-10:
Could you add an example of the invalid payload?

PR #5963, opened 2023-09-28
allow BLOB txs during shanghai

jflo:
covers first block after the fork case, merge after #5961


macfarla:
reference test fail https://app.circleci.com/pipelines/github/hyperledger/besu/24242/workflows/f8d4d4b8-9426-401e-bc5e-c7b6a34459ec/jobs/155813/tests#failed-test-0

fab-10 (changes requested):
I do not think this is the right way, since adding Blob tx here means that they are considered valid even before Cancun, that is against the protocol specs, since we can accept as valid blocks containing them when still on Shanghai.
We need to understand better the scope of the Hive test and in case change it

jflo:
> I do not think this is the right way, since adding Blob tx here means that they are considered valid even before Cancun, that is against the protocol specs, since we can accept as valid blocks containing them when still on Shanghai. We need to understand better the scope of the Hive test and in case change it.
> 
> I remember that this was discussed during a ACDE call, and we agreed that tests should not expect blob tx to be present in the pool for the fork block.

ok, that makes sense. Will close this PR, and need more time to open a more extensive one that allows the set of allowedTransaction types to change as the protocol schedule passes.

PR #5966, opened 2023-09-28
Use PendingTransaction in BlockTransactionSelector

fab-10:
## PR description

This is just a refactoring, no functionality change, to have the `PendingTransaction` object during the selection of transactions, since in the `PendingTransaction` there are additional metadata about the candidate transaction that can be used by the selection process, and more will be added in a next PR to better support priority senders.


Gabriel-Trintinalia (approved):
LGTM, I am wondering if it makes sense to have it named `PooledTransaction` instead of `PendingTransaction`, since PendingTransaction are transactions that are in the transaction pool and not confirmed yet. That would also match the methods `GetPooledTransactions` and `NewPooledTransactionHashes` as well as the EncodingContext in the transaction encoder

fab-10:
@Gabriel-Trintinalia it seems that _pending_ and _pooled_ are both used as synonyms, [here for example](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transaction-exchange), but I have no objections if you want to evaluate a renaming

PR #5973, opened 2023-10-02
[4844] EngineNewPayload - Add Blob Count validation

Gabriel-Trintinalia:
## PR description
Add Blob Count validation to EngineNewPayload - Fix invalid_blob_count hive tests


macfarla (approved):
LGTM

PR #5974, opened 2023-10-02
Target to use about 25MB for the new layered txpool by default

fab-10:
## PR description

Reduce the default memory usage of the txpool, to limit the added amount of memory used by Besu.
There 2 memory limited layers, ready and sparse, so to target 25MB we set the limit for a layer to 12.5MB


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolConfiguration.java
```diff
@@ -64,7 +64,7 @@ enum Implementation {
   boolean DEFAULT_ENABLE_SAVE_RESTORE = false;
 
   File DEFAULT_SAVE_FILE = new File(DEFAULT_SAVE_FILE_NAME);
-  long DEFAULT_PENDING_TRANSACTIONS_LAYER_MAX_CAPACITY_BYTES = 50_000_000L;
+  long DEFAULT_PENDING_TRANSACTIONS_LAYER_MAX_CAPACITY_BYTES = 12_500_000L;
```
siladu, line 67: Should this be 25MB or is there a non-obvious doubling going on somewhere?

fab-10: Updated the description:

> There 2 memory limited layers, ready and sparse, so to target 25MB we set the limit for a layer to 12.5MB

siladu (approved):
Reckon this warrants a changelog entry?

fab-10:
waiting for more metrics before merging

fab-10:
> Reckon this warrants a changelog entry?

Existing changelog entry about the layered txpool, updated

Issue #5911, opened 2023-09-20
SnapSync pipeline halting on RocksDB Busy exception during flat db heal


garyschulte:
### Description
During snap-sync, especially on a low IO instance, a deadlock can occur that will halt the snap sync pipeline.  See stack trace in logs below.

Besu should implement a retry policy that will catch the exception and retry the commit before throwing, rather than allowing the exception to abend the pipeline.

### Acceptance Criteria
* Snap sync pipeline is more tolerant of IO saturated machines

### Steps to Reproduce (Bug)
1. start snap sync of mainnet with flat db healing, on an IO restricted instance

**Expected behavior:** [What you expect to happen]
snap sync to finish with full flat db heal

**Actual behavior:** [What actually happens]
When IO contention is too high, a RocksDB Busy exception will be thrown and abend the sync pipeline

**Frequency:** [What percentage of the time does it occur?]

### Logs (if a bug)

```
{“@timestamp”:“2023-09-19T13:05:45,780",“level”:“INFO”,“thread”:“EthScheduler-Services-26 (batchPersistTrieNodeData)“,”class”:“SnapsyncMetricsManager”,“message”:“Healed 74500 world state trie nodes, Peer count: 25”,“throwable”:“”}
{“@timestamp”:“2023-09-19T13:05:55,384”,“level”:“INFO”,“thread”:“EthScheduler-Services-26 (batchPersistTrieNodeData)“,”class”:“Pipeline”,“message”:“Unexpected exception in pipeline. Aborting.“,”throwable”:” org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Busy
	at org.hyperledger.besu.plugin.services.storage.rocksdb.RocksDBTransaction.commit(RocksDBTransaction.java:95)
	at org.hyperledger.besu.services.kvstore.SegmentedKeyValueStorageTransactionValidatorDecorator.commit(SegmentedKeyValueStorageTransactionValidatorDecorator.java:65)
	at org.hyperledger.besu.ethereum.bonsai.storage.BonsaiWorldStateKeyValueStorage$Updater.commit(BonsaiWorldStateKeyValueStorage.java:472)
	at org.hyperledger.besu.ethereum.eth.sync.snapsync.PersistDataStep.persist(PersistDataStep.java:80)
[...]

Issue #5807, opened 2023-08-25
Forest mode failing to sync with snapsync


matkt:
### Description

Sometimes we can have a rocksdb timeout issue during the snapsync with forest

[...]

Issue #5549, opened 2023-06-08
Fresh snap sync did not start worldstate heal


yorickdowne:
### Description

A fresh sync of mainnet Ethereum on Besu using snap sync got through FastImportBlocksStep, but did not start worldstate heal. It was "stuck" here.

### Steps to Reproduce (Bug)

- Sync mainnet
- Likely a highly intermittent issue

### Logs (if a bug)

```
eth-main-execution-1  | 2023-06-06 23:45:55.873+00:00 | EthScheduler-Services-73 (importBlock) | INFO  | FastImportBlocksStep | Block import progress: 17407253 of 17423064 (99%)
eth-main-execution-1  | 2023-06-06 23:46:27.843+00:00 | EthScheduler-Services-73 (importBlock) | INFO  | FastImportBlocksStep | Block import progress: 17408453 of 17423064 (99%)
eth-main-execution-1  | 2023-06-06 23:48:34.674+00:00 | EthScheduler-Services-78 (importBlock) | INFO  | FastImportBlocksStep | Block import progress: 17415357 of 17423064 (99%)
eth-main-execution-1  | 2023-06-07 01:40:37.386+00:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
eth-main-execution-1  | 2023-06-07 07:40:34.310+00:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
eth-main-execution-1  | 2023-06-07 13:30:44.083+00:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
eth-main-execution-1  | 2023-06-07 19:20:40.366+00:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
eth-main-execution-1  | 2023-06-08 01:10:29.142+00:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
eth-main-execution-1  | 2023-06-08 07:00:31.256+00:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
eth-main-execution-1  | 2023-06-08 12:10:36.405+00:00 | Timer-0 | INFO  | DNSResolver | Resolved 2409 nodes
```

Node restarted at this point

```
eth-main-execution-1  | 2023-06-08 13:59:56.533+00:00 | main | INFO  | DefaultP2PNetwork | Node address 0xcbe24acf964d074e77a3e604410094a053702dfe
[...]

PR #5976, opened 2023-10-02
Retry on busy and locktimeout failures for rocksdb transactions

garyschulte:
## PR description
Simple default retry policy for rocksdb busy and locktimeout exceptions.

**This may or may not remedy the busy exceptions encountered in the listed issues, but it will at a minimum provide better signal regarding whether these are transient issues due to compaction, or more systemic issues

## Fixed Issue(s)

related to #5911 
related to #5549 
related to #5807

### DIFF
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
@@ -22,6 +22,8 @@ import org.hyperledger.besu.plugin.services.storage.SegmentIdentifier;
 import org.hyperledger.besu.plugin.services.storage.SegmentedKeyValueStorageTransaction;
 
 import java.util.EnumSet;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 
 import org.rocksdb.ColumnFamilyHandle;
@@ -141,6 +143,10 @@ public class RocksDBTransaction implements SegmentedKeyValueStorageTransaction {
               retryLimit,
               ex.getStatus().getCodeString());
           try {
+            if (ex.getStatus().getSubCode().equals(Status.SubCode.Deadlock)) {
+              // deadlock detection returns immediately, backoff and wait if deadlock is detected
+              retryBackoff();
+            }
             retryAction.retry();
           } catch (RocksDBException ex2) {
             maybeRetryRocksDBAction(ex2, attemptNumber + 1, retryLimit, retryAction);
@@ -148,6 +154,31 @@ public class RocksDBTransaction implements SegmentedKeyValueStorageTransaction {
         }
       } else {
         throw new StorageException(ex);
+      }
+    }
+
+    long BASE_TIMEOUT = 1000; // Base timeout in milliseconds
+    long MAX_TIMEOUT = 10000; // Max timeout in milliseconds
+    long DECAY_TIME = 5000; // Time in milliseconds after which the timeout decays
+    AtomicLong timeout = new AtomicLong(BASE_TIMEOUT);
+    AtomicLong lastCallTime = new AtomicLong(System.currentTimeMillis());
+
+    static void retryBackoff() {
+      try {
+        long currentTime = System.currentTimeMillis();
+        long delay = timeout.get();
+        // If the function hasn't been called for DECAY_TIME milliseconds, decay the timeout back to
+        // the base value
+        if (currentTime - lastCallTime.get() > DECAY_TIME) {
+          delay = BASE_TIMEOUT;
+          timeout.set(BASE_TIMEOUT);
+        }
+        TimeUnit.MILLISECONDS.sleep(delay);
+        // Increase the timeout for the next call, up to the maximum
+        timeout.updateAndGet(t -> Math.min(t * 2, MAX_TIMEOUT));
+        lastCallTime.set(currentTime);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt(); // Preserve interrupt status
       }
     }
   }
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
@@ -90,7 +90,11 @@ public class RocksDBTransaction implements SegmentedKeyValueStorageTransaction {
     commit(0, 3);
   }
 
-  /** commit transaction with an explicit number of retries. */
+  /**
+   * commit transaction with an explicit number of retries.
+   *
+   * @param retryLimit limit for the number of retry attempts
+   */
   public void commitWithRetries(final int retryLimit) throws StorageException {
     commit(0, retryLimit);
   }
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
@@ -90,12 +90,7 @@ public class RocksDBTransaction implements SegmentedKeyValueStorageTransaction {
     commit(0, 3);
   }
 
-  /**
-   * commit transaction with an explicit number of retries.
-   *
-   * @param retryLimit limit for the number of retry attempts
-   * @throws StorageException
-   */
+  /** commit transaction with an explicit number of retries. */
   public void commitWithRetries(final int retryLimit) throws StorageException {
     commit(0, retryLimit);
   }
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
@@ -107,13 +107,13 @@ public class RocksDBTransaction implements SegmentedKeyValueStorageTransaction {
         logger.error(e.getMessage());
         System.exit(0);
       } else if (e.getMessage().equals(ERR_BUSY) && attemptNumber < retryLimit) {
-        logger.info(
+        logger.warn(
             "RocksDB Busy exception caught on attempt {} of {}, retrying",
             attemptNumber,
             retryLimit);
         commit(attemptNumber + 1, retryLimit);
       } else if (e.getMessage().equals(ERR_LOCK_TIMED_OUT) && attemptNumber < retryLimit) {
-        logger.info(
+        logger.warn(
             "RocksDB Lock Timeout exception caught on attempt {} of {}, retrying",
             attemptNumber,
             retryLimit);
--- a/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
+++ b/plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
@@ -106,13 +106,13 @@ public class RocksDBTransaction implements SegmentedKeyValueStorageTransaction {
       if (e.getMessage().contains(ERR_NO_SPACE_LEFT_ON_DEVICE)) {
         logger.error(e.getMessage());
         System.exit(0);
-      } else if (e.getMessage().equals(ERR_BUSY) && attemptNumber < retryLimit) {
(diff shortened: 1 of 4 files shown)


plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBTransaction.java
```diff
@@ -85,14 +87,35 @@ public void remove(final SegmentIdentifier segmentId, final byte[] key) {
...
-      if (e.getMessage().contains(NO_SPACE_LEFT_ON_DEVICE)) {
+      if (e.getMessage().contains(ERR_NO_SPACE_LEFT_ON_DEVICE)) {
         logger.error(e.getMessage());
         System.exit(0);
+      } else if (e.getMessage().equals(ERR_BUSY) && attemptNumber < retryLimit) {
+        logger.trace(
+            "RocksDB Busy exception caught on attempt {} of {}, retrying",
+            attemptNumber,
+            retryLimit);
+        commit(attemptNumber + 1, retryLimit);
```
matkt, line 109: do you think a direct retry without any delay will fix the problem ? maybe setting the level to info like that we can debug and see if this fix can unblock this kind of situation ?

garyschulte: I haven't yet been able to reproduce a busy exception, but I bumped the logging to warn.  It seems reasonable to leave it at that level since it likely will provide good signal and shouldn't spam the logs due to the transaction commit timeout.

jframe:
I remember that @gfukushima tried a similar approach in https://github.com/hyperledger/besu/pull/5287/files but from what recall the retry didn't fix the busy error.

garyschulte:
> I remember that @gfukushima tried a similar approach in https://github.com/hyperledger/besu/pull/5287/files but from what recall the retry didn't fix the busy error.

Thanks I didn't know he already tried this.  

Was he able correlate this to rocksdb activity?  The rocks code indicates that 'Status::Busy' should be transient and the remedy is to retry.  I have some instances trying to reproduce this now, hoping to get some activity metrics and and rocksdb logs to see what might be causing the contention.

jframe:
> > I remember that @gfukushima tried a similar approach in https://github.com/hyperledger/besu/pull/5287/files but from what recall the retry didn't fix the busy error.
> 
> Thanks I didn't know he already tried this.
> 
> Was he able correlate this to rocksdb activity? The rocks code indicates that 'Status::Busy' should be transient and the remedy is to retry. I have some instances trying to reproduce this now, hoping to get some activity metrics and and rocksdb logs to see what might be causing the contention.

Not sure about correlating it to rocksdb activity. Too long ago. He is on holiday right now. Have asked around, hopefully someone remembers more than I do.

Agree that doing a retry definitely seems like the right thing to do.

siladu:
> hopefully someone remembers more than I do.
Maybe @matkt ?

garyschulte:
I have been able to reproduce this somewhat readily by snap syncing a forest node with slow storage.  

I think Forest is particularly vulnerable to these transient locking/busy issues, because it is using [TransactionDB mode](https://github.com/facebook/rocksdb/wiki/Transactions#transactiondb), which locks keys during transaction write rather than during commit.

So far, I have seen it exclusively on `.put()` rather than `.commit()`, so I tweaked the implementation to retry all RocksDBTransaction actions that encounter a non-fatal exception and am re-testing

garyschulte:
I have been able to reproduce this somewhat readily by snap syncing a forest node with slow storage.

I think Forest is particularly vulnerable to these transient locking/busy issues, because it is using [TransactionDB mode](https://github.com/facebook/rocksdb/wiki/Transactions#transactiondb), which locks keys during transaction write rather than during commit.

So far, I have seen it exclusively on .put() rather than .commit(), so I tweaked the implementation to retry all RocksDBTransaction actions that encounter a non-fatal exception and am re-testing

garyschulte:
Interestingly, this *did* fix one of the TimedOut(LockTimeout) exceptions on one thread on scheduler-8.  But 60 seconds later the same thing happened to both scheduler-8 and scheduler-13.  It *looks* like they were both trying to obtain locks and ended up in a deadlock kind of situation:

```
{"@timestamp":"2023-10-05T23:50:50,129","level":"WARN","thread":"EthScheduler-Services-8 (batchPersistStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 0 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:45,556","level":"WARN","thread":"EthScheduler-Services-8 (batchPersistStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 0 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:45,627","level":"WARN","thread":"EthScheduler-Services-13 (batchPersistLargeStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 0 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:46,557","level":"WARN","thread":"EthScheduler-Services-8 (batchPersistStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 1 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:46,630","level":"WARN","thread":"EthScheduler-Services-13 (batchPersistLargeStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 1 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:47,558","level":"WARN","thread":"EthScheduler-Services-8 (batchPersistStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 2 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:47,630","level":"WARN","thread":"EthScheduler-Services-13 (batchPersistLargeStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 2 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:48,558","level":"WARN","thread":"EthScheduler-Services-8 (batchPersistStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 3 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
{"@timestamp":"2023-10-05T23:51:48,631","level":"WARN","thread":"EthScheduler-Services-13 (batchPersistLargeStorageData)","class":"RocksDBTransaction","message":"RocksDB Transient exception caught on attempt 3 of 3, status: TimedOut(LockTimeout), retrying.","throwable":""}
```

It is interesting that the retry worked for the first lock timeout, but 8 and 13 together seemed to be deadlocked in the same time frame, for the same timeout duration.  Perhaps what we might need to do is introduce a static timeout supplier that forces an incremental backoff wait.  That way at least one of the threads might be able to get the lock and progress while the other one waits.

I will try that on this node since it seems to readily reproduce it within an hour or so of starting a mainnet sync.

matkt:
If the bug arrives largely during the sync. I can try to make a logic of retry but at the pipeline level in addition to this pr. Instead of failing pipeline on a lock I restart the step later. Your PR is trying to manage the commit in general but I think we can also try to have better management of rocksdb error on the pipeline side too

garyschulte:
> If the bug arrives largely during the sync. I can try to make a logic of retry but at the pipeline level in addition to this pr. Instead of failing pipeline on a lock I restart the step later. Your PR is trying to manage the commit in general but I think we can also try to have better management of rocksdb error on the pipeline side too

There seems to be a variety of places this can occur.  I think it is better to address it systemically rather than having individual processes duplicate the retry logic.  If we find that some code paths are particularly prone to this or need special care, we can always catch the storage exception that is eventually thrown and apply process-specific logic there.

matkt:
> > If the bug arrives largely during the sync. I can try to make a logic of retry but at the pipeline level in addition to this pr. Instead of failing pipeline on a lock I restart the step later. Your PR is trying to manage the commit in general but I think we can also try to have better management of rocksdb error on the pipeline side too
> 
> There seems to be a variety of places this can occur. I encountered LockTimeout most often during sync in state trie commit visitors, when they are just doing a transaction `put()`. I haven't been able to reproduce the Busy exception during a bonsai sync, but Optimistic transaction db seems to have a different behavior/pattern since it resolves locks at transaction `.commit()` time.
> 
> I think it is better to address it systemically rather than having individual processes duplicate the retry logic. If we find that some code paths are particularly prone to this or need special care, we can always catch the storage exception that is eventually thrown and apply process-specific logic there.

Ok i will try to also start some sync next week hoping to have it with bonsai

garyschulte:
@matkt - perhaps I spoke too soon about not trying to handle this in the sync process.  With an incremental backoff retry, I am still seeing LockTimeouts.  Sometimes a single thread will get locktimeouts and will resolve, but inevitably two separate threads are trying to write to overlapping key ranges and a deadlock occurs.  It doesn't matter how long we wait since each process/transaction is trying to get a lock on the keys of the other.  This doesn't happen with OptimisticTransactionDB because conflicts are resolved at commit time rather than write.

We will have to do some instrumenting/logging to test this theory, but I think we have 2 separate worldstate download threads that are trying to write to some of the same trie keys during the commit phase.

However... I suspect the better option is going to be to ditch TransactionDB for forest and go back to OptimisticTransactionDB with transaction retries.  IIRC, we reverted forest to TransactionDB to alleviate "Busy" exceptions.  The current retry strategy in this PR might make TransactionDB unnecessary for forest, and we can avoid the concurrency/deadlock issue.

matkt:
> @matkt - perhaps I spoke too soon about not trying to handle this in the sync process. With an incremental backoff retry, I am still seeing LockTimeouts. Sometimes a single thread will get locktimeouts and will resolve, but inevitably two separate threads are trying to write to overlapping key ranges and a deadlock occurs. It doesn't matter how long we wait since each process/transaction is trying to get a lock on the keys of the other. This doesn't happen with OptimisticTransactionDB because conflicts are resolved at commit time rather than write.
> 
> We will have to do some instrumenting/logging to test this theory, but I think we have 2 separate worldstate download threads that are trying to write to some of the same trie keys during the commit phase.
> 
> However... I suspect the better option is going to be to ditch TransactionDB for forest and go back to OptimisticTransactionDB with transaction retries. IIRC, we reverted forest to TransactionDB to alleviate "Busy" exceptions. The current retry strategy in this PR might make TransactionDB unnecessary for forest, and we can avoid the concurrency/deadlock issue.

For the sync part. The fix can be very simple, that's why I propose this. It is juste to catch the error during a persist and to marke the task as failed. The pipeline therefore will process this task again later with the pipeline

matkt:
> > @matkt - perhaps I spoke too soon about not trying to handle this in the sync process. With an incremental backoff retry, I am still seeing LockTimeouts. Sometimes a single thread will get locktimeouts and will resolve, but inevitably two separate threads are trying to write to overlapping key ranges and a deadlock occurs. It doesn't matter how long we wait since each process/transaction is trying to get a lock on the keys of the other. This doesn't happen with OptimisticTransactionDB because conflicts are resolved at commit time rather than write.
> > We will have to do some instrumenting/logging to test this theory, but I think we have 2 separate worldstate download threads that are trying to write to some of the same trie keys during the commit phase.
> > However... I suspect the better option is going to be to ditch TransactionDB for forest and go back to OptimisticTransactionDB with transaction retries. IIRC, we reverted forest to TransactionDB to alleviate "Busy" exceptions. The current retry strategy in this PR might make TransactionDB unnecessary for forest, and we can avoid the concurrency/deadlock issue.
> 
> For the sync part. The fix can be very simple, that's why I propose this. It is juste to catch the error during a persist and to marke the task as failed. The pipeline therefore will process this task again later with the pipeline

I think timeout is the Optimistic version of Busy

garyschulte:
> For the sync part. The fix can be very simple, that's why I propose this. It is juste to catch the error during a persist and to marke the task as failed. The pipeline therefore will process this task again later with the pipeline .Apart from that I think that the probability of having this error outside the sync should be limited. Because we do not have too many other cases where we have several threads that persist in the Worldstate at tthe same

Seems reasonable.  It isn't obvious that allowing both threads to write is going to result in a correct trie either, so allowing one or both to fail and retry is probably the better strategy.

Looking back over all of the "Busy" issues, it seems all of them have been related to worldstate sync.  If we handle the sync failure gracefully in the pipeline, there really isn't any need for TransactionDB anymore afaict.  We know that Optimistic is more performant, so if TransactionDB isn't 'fixing' anything, we should remove it as well.

garyschulte:
FYI I am getting persistent and repeated timeouts with this PR when snapsyncing forest.  The retry might be grouping together requests that are getting overlapping range responses.  I will let it run for a few more hours and see if it makes any headway or if it is deadlocked.

garyschulte:
closing this in favor of #6004

PR #5980, opened 2023-10-03
Only use the builder to create transactions

fab-10:
## PR description

Remove all the overloaded `Transaction`' constructors, and only allow to create a transaction using the `Builder`


jframe (approved):
LGTM - good to get rid of those extra constructors

Issue #5977, opened 2023-10-02
Tuned default values for the layered tx pool that respect memory constraints 


non-fungible-nelson:
https://wiki.hyperledger.org/display/BESU/Memory+usage+investigations+on+23.7.3-RC

@fab-10 @ahamlat - please use this new issue to track the RC3 related memory tuning.

PR #5985, opened 2023-10-04
Transaction detachedCopy to optimize txpool memory usage

fab-10:
## PR description

When a tx is decoded from RLP, some of its fields are reference to slices of the encoded RLP byte array, the latter could be shared by many transactions at the same time, for example when decoding the block body, every tx object keep a reference to the large RLP byte array, that could be hundred of KB long, the same happens when we receive a bunch on txs via p2p in the same message.
This approach works fine if the tx objects are short lived, since it means allocating and copying less data around, but does not fit well for pending txs, since they could live much longer in the txpool and a single tx could keep a reference to a large byte array, while actually only it only needs a small portion of it.

This PR solve this memory inefficiency, by creating a _detachedCopy_ of the tx when this is added to the txpool. A _detachedCopy_ is a copy that does not share any byte array with others. This also changes previous assumptions on the estimation of the amount of memory used by a pending tx, so the calculation have been updated and extended to cover also the Blob tx estimation.

relates #5977


jflo (approved):
love it, pls merge asap and I will close my competing pr.

PR #5986, opened 2023-10-04
Validation ordering

jflo:
- corrects a number of validation errors by adjusting priority of evaluations
- covers and corrects a subtle issue where besu would propose bad blocks in the event of a re-org prior to a proposal which restored blob transactions to mempool.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -469,6 +469,10 @@ && strictReplayProtectionShouldBeEnforcedLocally(chainHeadBlockHeader)
       return ValidationResultAndAccount.invalid(
           TransactionInvalidReason.INVALID_TRANSACTION_FORMAT,
           "EIP-1559 transaction are not allowed yet");
+    } else if (transaction.getType().equals(TransactionType.BLOB)
```
macfarla, line 472: this is the change that fixed the invalid block proposals, right?

macfarla (approved):
looks pretty straight forward. couple of comments

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
jflo: exactly! I considered updating a transaction validator to handle it, but decided against. it's not an invalid tx, it's just in the wrong context.

Gabriel-Trintinalia: can we use transaction.getType().supportsBlob() here?

PR #5997, opened 2023-10-06
Use existing Bytes48 for KZGCommitment and KZGProof

fab-10:
## PR description

Since we have a native `Bytes48` datatype in Tuweni, we can use it for `KZGCommitment` and `KZGProof` to also have better checks and sematic


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -339,12 +339,11 @@ public TransactionProcessingResult processTransaction(
           gasCalculator.accessListGasCost(accessListEntries.size(), accessListStorageCount);
       final long gasAvailable = transaction.getGasLimit() - intrinsicGas - accessListGas;
       LOG.trace(
-          "Gas available for execution {} = {} - {} - {} - {} (limit - intrinsic - accessList - data)",
+          "Gas available for execution {} = {} - {} - {} (limit - intrinsic - accessList)",
           gasAvailable,
           transaction.getGasLimit(),
           intrinsicGas,
-          accessListGas,
-          blobGas);
```
garyschulte, line 347: did you intend for this to be part of the bytes48 pr ?

garyschulte (approved):
LGTM, just a question about intent regarding the couple non-Bytes48 related logging items in the PR

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
fab-10: it is a little log fix, that I included to avoid creating a separate PR just for it

PR #6004, opened 2023-10-09
add retry logic on sync pipeline for rocksdb issue

matkt:
## PR description

This Pull Request allows retrying recoverable RocksDB errors such as 'busy' or 'lock timeout' during the initial synchronization. This helps to unblock the pipeline if possible and prevents pipelines from getting stuck at 99%.

## Fixed Issue(s)

#5911


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/StorageExceptionManager.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.hyperledger.besu.ethereum.eth.sync;
+
+import org.hyperledger.besu.plugin.services.exception.StorageException;
+
+public final class StorageExceptionManager {
+
+  private static final String rocksdbClassName = "RocksDBException";
+  private static final String ERR_BUSY = "Busy";
+  private static final String ERR_LOCK_TIMED_OUT = "TimedOut(LockTimeout)";
```
garyschulte, line 23:
```suggestion
  EnumSet<Status.Code> RETRYABLE_STATUS_CODES =
      EnumSet.of(Status.Code.TimedOut, Status.Code.TryAgain, Status.Code.Busy);
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/StorageExceptionManager.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public static boolean canRetryOnError(final StorageException e) {
+    return e.getMessage().contains(rocksdbClassName)
+        && (e.getMessage().contains(ERR_BUSY) || e.getMessage().contains(ERR_LOCK_TIMED_OUT));
+  }
```
garyschulte, line 28: less brittle to use enum values

```suggestion
      return Optional.of(e.getCause())
          .filter(z -> z instanceof RocksDBException)
          .map(RocksDBException.class::cast)
          .map(RocksDBException::getStatus)
          .map(Status::getCode)
          .map(RETRYABLE_STATUS_CODES::contains)
          .orElse(false);
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/StorageExceptionManager.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.eth.sync;
+
+import org.hyperledger.besu.plugin.services.exception.StorageException;
+
```
garyschulte, line 18:
```suggestion
import org.rocksdb.RocksDBException;
import org.rocksdb.Status;
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/PersistDataStep.java
```diff
@@ -33,24 +42,35 @@ public List<Task<NodeDataRequest>> persist(
...
+                }
+              });
+      updater.commit();
+    } catch (StorageException storageException) {
+      if (canRetryOnError(storageException)) {
+        // We reset the task by setting it to null. This way, it is considered as failed by the
+        // pipeline, and it will attempt to execute it again later.
+        LOG.info("retry on rocksdb issue " + storageException.getMessage());
+        tasks.forEach(nodeDataRequestTask -> nodeDataRequestTask.getData().setData(null));
+      } else {
```
garyschulte, line 70: can we not fail the future for an individual task in the foreach rather than failing the whole list?  or is that a lot more plumbing to do in the pipeline to handle a single failed task future?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldStateDownloadProcess.java
```diff
@@ -231,6 +231,22 @@ public SnapWorldStateDownloadProcess build() {
...
+      2. Storage Data Pipeline (fetchStorageDataPipeline): Running parallel to the account data pipeline, this pipeline downloads the storage of smart contracts.
+          If all slots cannot be downloaded at once, tasks are created in the fetchLargeStorageDataPipeline to download the storage by range, allowing parallelization of large account downloads.
+
+      3. Code Data Pipeline (fetchCodePipeline): This pipeline, running concurrently with the account and storage data pipelines, is responsible for downloading the code of the smart contracts.
+
+      4. Large Storage Data Pipeline (fetchLargeStorageDataPipeline): This pipeline is used when the storage data for a smart contract is too large to be downloaded at once.
+          It enables the storage data to be downloaded in ranges, similar to the account data.
+
+      5. Healing Phase: Initiated after all other pipelines have completed their tasks, this phase ensures the integrity and completeness of the downloaded data.
+      */
```
garyschulte, line 249: 🙏

garyschulte:
FYI I am getting persistent and repeated timeouts with this PR when snapsyncing forest. The retry might be grouping together requests that are getting overlapping range responses. I will let it run for a few more hours and see if it makes any headway or if it is deadlocked.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/PersistDataStep.java
matkt: not possible because all modification are in the same updater and we cannot commit only a part of a transaction

matkt:
> FYI I am getting persistent and repeated timeouts with this PR when snapsyncing forest. The retry might be grouping together requests that are getting overlapping range responses. I will let it run for a few more hours and see if it makes any headway or if it is deadlocked.
> 
> edit: the deadlock did resolve, but it took a couple hours. Once resolved, sync progressed normally until it hit another problematic range. The test instance I used today spent ~6 hours failing and retrying every second, after 12 hours it is roughly 37% complete. So, 👍 it does progress, but it seems like there should be a better / faster strategy we can use.

yes clearly I think maybe we need to take some times to avoid writing same datas in differents thread but could be a mitigation . I also think verkle trie sync will simplify the snapsync stragegy as we have only one trie so maybe waiting 🤔

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/StorageExceptionManager.java
matkt: cannot work because the exception is like that
“Unexpected exception in pipeline. Aborting.“,”throwable”:” org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Busy"

the cause will be a Storage exception and the message "org.rocksdb.RocksDBException: Busy"

matkt:
I restarted a sync to test

matkt:
> FYI I am getting persistent and repeated timeouts with this PR when snapsyncing forest. The retry might be grouping together requests that are getting overlapping range responses. I will let it run for a few more hours and see if it makes any headway or if it is deadlocked.
> 
> edit: the deadlock did resolve, but it took a couple hours. Once resolved, sync progressed normally until it hit another problematic range. The test instance I used today spent ~6 hours failing and retrying every second, after 12 hours it is roughly 37% complete. So, 👍 it does progress, but it seems like there should be a better / faster strategy we can use.

is you sync finished correctly ?

garyschulte:
> is your sync finished correctly ?

It did finish worldstate sync, the heal was still running when the process was shut down but it looked like it was going to finish.  :+1:

matkt:
@garyschulte do we want to keep log level to info ? or maybe adding a logic to log only one time in order to avoid spam but give us a way to debug and understand that retry occured

garyschulte:
> @garyschulte do we want to keep log level to info ? or maybe adding a logic to log only one time in order to avoid spam but give us a way to debug and understand that retry occured

Seems like a good idea to reduce or rate limit the logging.  on the low iops forest node, the retry logs were once per second for several hours.

matkt:
> > @garyschulte do we want to keep log level to info ? or maybe adding a logic to log only one time in order to avoid spam but give us a way to debug and understand that retry occured
> 
> Seems like a good idea to reduce or rate limit the logging. on the low iops forest node, the retry logs were once per second for several hours.

I modified the code to not diplay all issues, I will start a simple snapsync just to be sure and I think we will be able to merge it

matkt:
Sync worked, Imo we can merge if you are ok

garyschulte (approved):
🚢

PR #6010, opened 2023-10-10
[Plugin API] - TransactionSelector - Send TransactionSelectionResult to the plugin when not transaction is not selected

Gabriel-Trintinalia:
## PR description
Adds the `TransactionSelectionResult` to `TransactionSelector.onTransactionNotSelected` so that plugins have more context.

Addresses the comment in this PR https://github.com/hyperledger/besu/pull/6005#r1351753669


fab-10 (approved):
LGTM, and the cleanup makes it more readable

PR #6012, opened 2023-10-10
Standardize pubkey and public key naming in deposit

naviechan:
## PR description
Currently there are various naming for public key (`publicKey`, `pubkey`, `pubKey`) in Deposit/DepositReceipt. This PR is to standardize them to `pubkey` such that they are consistent with the naming convention in [execution-api](https://github.com/ethereum/execution-apis/blob/584905270d8ad665718058060267061ecfd79ca5/src/engine/experimental/eip6110.md?plain=1#L31)


siladu (approved):
Probably only matters that the JsonProperty matches the spec, but I like the consistency.

I wish the spec was pubKey or publicKey though, I think pubkey is inconsistent with the rest of the spec, e.g.
```
- `pubkey`: `DATA`, 48 Bytes
- `withdrawalCredentials`: `DATA`, 32 Bytes
```

Issue #5979, opened 2023-10-03
besu t8n require post state to be provided still


winsvega:
still getting this issue 

```
.retesteth/besu/setup.sh WARNING: Besu daemon failed to start, will use besu evm t8n instead
/home/ubuntu/.retesteth/besu/start.sh --state.fork Berlin --state.reward 0 --state.chainid 1 --input.alloc /dev/shm/4a92005a-d5cb-4287-b4fe-2d286b62b0dc/alloc.json --input.txs /dev/shm/4a92005a-d5cb-4287-b4fe-2d286b62b0dc/txs.rlp --input.env /dev/shm/4a92005a-d5cb-4287-b4fe-2d286b62b0dc/env.json --output.basedir /dev/shm/4a92005a-d5cb-4287-b4fe-2d286b62b0dc --output.result out.json --output.alloc outAlloc.json --output.errorlog /dev/shm/4a92005a-d5cb-4287-b4fe-2d286b62b0dc/error.json

Res:

RAlloc:

Tool log: 
java.lang.RuntimeException: World State Root does not match expected value, header 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 calculated 0xcd39e0cdd18f8f811911222ae6779341663d0293e1a3d9501da7ac2f4da9b277
	at org.hyperledger.besu.ethereum.bonsai.worldview.BonsaiWorldState.verifyWorldStateRoot(BonsaiWorldState.java:435)
	at org.hyperledger.besu.ethereum.referencetests.BonsaiReferenceTestWorldState.verifyWorldStateRoot(BonsaiReferenceTestWorldState.java:91)
	at org.hyperledger.besu.ethereum.bonsai.worldview.BonsaiWorldState.persist(BonsaiWorldState.java:393)
	at org.hyperledger.besu.evmtool.T8nExecutor.runTest(T8nExecutor.java:380)
	at org.hyperledger.besu.evmtool.T8nSubCommand.run(T8nSubCommand.java:285)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2026)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
[...]

PR #6013, opened 2023-10-10
add method to disable root verification for T8n and not Reference tests

matkt:
## PR description

add method to disable root verification for T8n and not Reference tests

## Fixed Issue(s)

#5979


ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/BonsaiReferenceTestWorldState.java
```diff
@@ -87,7 +89,7 @@ public ReferenceTestWorldState copy() {
    */
   @Override
   protected void verifyWorldStateRoot(final Hash calculatedStateRoot, final BlockHeader header) {
-    if (!header.getStateRoot().equals(Hash.ZERO)) {
+    if (!disableRootHashVerification) {
```
garyschulte, line 92: ~~If we are going to do it this way, we should probably also set disableRootHashVerification in the retesteth subcommand.  I don't think it is being used currently, but we shouldn't leave it broken~~

Nevermind - it looks like retesteth subcommand is still using forest

garyschulte (approved):
🚢

PR #6014, opened 2023-10-10
implements tx sizing for blob txs

jflo:
sizing was deferring to 1559 sizing for blob transactions.


fab-10:
This should be already covered in https://github.com/hyperledger/besu/pull/5985

PR #6022, opened 2023-10-11
Add test to check that TX selection doesn't prioritize TXs from the same sender

matthew1001:
## PR description
Note that this PR is dependent on these PRs:
 - https://github.com/hyperledger/besu/pull/6106
 - https://github.com/hyperledger/besu/pull/6146

(My testing showed that https://github.com/hyperledger/besu/pull/6146 was required and had been missed off https://github.com/hyperledger/besu/pull/6106)

This PR now simply adds unit tests which demonstrate that the PRs ^^ mean we no longer require a specific CLI option `--Xtx-pool-disable-sender-grouping`

Re-ordering the TX selection under those PRs meant that grouping of transactions per sender was no longer an issue in the legacy TX pool.

~~The intent of this PR is to give the user the option to control whether the transaction pool selection logic prefers to fill the block with all available transactions from each sender it has transactions for, or to treat the transactions equally, regardless of which sender they came from. Discussions with enterprise users have shown that the existing/default behaviour is not what they expect, and further it can cause an active sender to inadvertently monopolize the TX selection and effectively block other senders from having their transactions mined.~~

~~The PR adds an experimental flag `--Xtx-pool-disable-sender-grouping`. When this flag is set, the transaction selection pays no regard to the account the transaction was from.~~

~~Since this is an experimental feature, I have not added the `doc-change-required` label as I don't believe we document experimental features?~~


matthew1001:
@fab-10 I've been looking at a suitable approach to avoid transactions from a single sender monopolizing the transaction selection process and preventing other senders' transactions from being included in a block. The current draft is a working implementation, but I'm not very happy with it.

It feels like the logic in `AbstractPendingTransactionsSorter.selectTransactions()` ( https://github.com/hyperledger/besu/blob/03a8335bf94e613fff8e4564b432069588e8e6a3/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsSorter.java#L244C20-L244C20) is fighting against the logic in `GasPricePendingTransactionsSorter` (https://github.com/hyperledger/besu/blob/03a8335bf94e613fff8e4564b432069588e8e6a3/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/GasPricePendingTransactionsSorter.java#L38):

```
private final NavigableSet<PendingTransaction> prioritizedTransactions =
      new TreeSet<>(
          comparing(PendingTransaction::isReceivedFromLocalSource)
              .thenComparing(PendingTransaction::getGasPrice)
              .thenComparing(PendingTransaction::getAddedAt)
              .thenComparing(PendingTransaction::getSequence)
              .reversed());
```

`prioritizedTransactions` returns the transactions in reverse order (i.e. the most recently added transactions, quite likely the ones with the highest nonces, are the first to be considered for selection). `AbstractPendingTransactionsSorter.selectTransactions()` therefore has to "defer" those with higher nonces, and when it reaches a valid nonce it grabs all of the deferred ones for that sender. Hence the current situation where if the first valid nonce TX is for `sender1`, and `sender1` has 100s of transactions deferred, those are chosen in preference to any of `sender2`'s.

My first go at making this configurable (via `--tx-pool-disable-sender-grouping`) is the following:

 1. Keep the same ordering as today, where most recent TXs are considered first
 2. Keep the same deferral mechanism
 3. When a valid TX nonce is found, evaluate it but leave the deferred ones in their list
 4. At the end, if any deferred transactions for any senders remain, consider them one account at a time (i.e. round robin across accounts giving each one a chance to have its next deferred transaction considered for selection)
 5. Continue step 4 until no deferred transactions for any sender accounts remain

The implementation in the PR needs tidying up (I've got a bunch of old-school for-loops to replace with streams etc) but I wanted to get feedback before continuing. The extra unit tests I've added demonstrate that it works. But I can't help thinking that if the order of `priorizedTransactions` was the oldest transactions to be submitted are considered first then the logic would be simpler and would work for both the original logic and the new logic. I think it would be something like:

```
private final NavigableSet<PendingTransaction> prioritizedTransactions =
      new TreeSet<>(
          comparing(PendingTransaction::getSequence, Comparator.reverseOrder())
              .thenComparing(PendingTransaction::getAddedAt, Comparator.reverseOrder())
              .thenComparing(PendingTransaction::getGasPrice)
              .thenComparing(PendingTransaction::isReceivedFromLocalSource));
```

(going by memory - the above may not be correct, but just to give you the idea). However, I think that's a fairly fundamental change to the prioritized transactions ordering so I'm guessing it would have a bunch of side affects.

fab-10:
@matthew1001 sorry but I could not review this in details until next week. 
In the mean time I want to share an approach I was thinking about yesterday when we talked about the replacement of txs in a gas free market. 
Now we have the possibility to select the txpool implementation, via the `tx-pool` options, and thus could be possible to introduce an implementation tailored for a gas free market, that does not order by price, try to be FIFO (or fair) and accept any replacement.
Pretty simple, the only complication is tracking the nonce gaps, but for it we could just reuse the Ready and Sparse layers of the layered txpool, making possible to customize their sorted and the selector.

fab-10:
FYI: @7suyash7

matthew1001:
Would you be opposed to an interim fix along the lines of the change in this PR? I think I agree that a custom pool implementation for these use-cases would be preferable, but a tactical non-breaking change to provide an option for the current pool implementation would be very useful in the mean time.

Out of interest I raised a draft PR (https://github.com/hyperledger/besu/pull/6106) to see if it regressed any existing behaviour and get feedback on whether it's a valid change to make as a pre-req to this PR. It causes transactions to be sorted by _**ascending**_ sequence number and added-time, not _**descending**_ order. I can't think of a downside to that, and it would allow this PR to be much simplified.

besu/src/main/java/org/hyperledger/besu/cli/options/stable/TransactionPoolOptions.java
```diff
@@ -42,6 +42,7 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
   private static final String TX_POOL_ENABLE_SAVE_RESTORE = "--tx-pool-enable-save-restore";
   private static final String TX_POOL_SAVE_FILE = "--tx-pool-save-file";
   private static final String TX_POOL_PRICE_BUMP = "--tx-pool-price-bump";
+  private static final String TX_POOL_DISABLE_SENDER_GROUPING = "--tx-pool-disable-sender-grouping";
```
fab-10, line 45: I think it is better to make this an experimental options for the moment, just prefix it with an `X`

besu/src/main/java/org/hyperledger/besu/cli/options/stable/TransactionPoolOptions.java
```diff
@@ -88,6 +89,15 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
       arity = "1")
   private Percentage priceBump = TransactionPoolConfiguration.DEFAULT_PRICE_BUMP;
 
+  @CommandLine.Option(
+      names = {TX_POOL_DISABLE_SENDER_GROUPING},
+      paramLabel = "<Boolean>",
+      description =
+          "Disable sender grouping of transactions during selection. (default: ${DEFAULT-VALUE})",
+      arity = "0..1")
+  private Boolean disableSenderTXGrouping =
```
fab-10, line 98: nit: s/TX/Tx

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsSorter.java
```diff
@@ -252,14 +252,17 @@ public void selectTransactions(final TransactionSelector selector) {
...
+
       while (prioritizedTransactions.hasNext()) {
         final PendingTransaction highestPriorityPendingTransaction = prioritizedTransactions.next();
         final AccountTransactionOrder accountTransactionOrder =
             accountTransactions.computeIfAbsent(
                 highestPriorityPendingTransaction.getSender(), this::createSenderTransactionOrder);
 
         for (final PendingTransaction transactionToProcess :
-            accountTransactionOrder.transactionsToProcess(highestPriorityPendingTransaction)) {
+            accountTransactionOrder.transactionsToProcess(
```
fab-10, line 263: You said this could change after #6106, so will review this later

fab-10 (commented):
I am fine, finding an interim way to make this txpool fairer, will continue the review after #6106 is sorted out

besu/src/main/java/org/hyperledger/besu/cli/options/stable/TransactionPoolOptions.java
matthew1001: I guess I should move it to https://github.com/hyperledger/besu/blob/ee7a66fc785a8a504d1e501e355492e89f6e73a3/besu/src/main/java/org/hyperledger/besu/cli/options/unstable/TransactionPoolOptions.java as well?

fab-10: correct

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 24.1.0
```
fab-10, line 3: The next version is 23.10.2, unless you specifically want to add it to the next quarterly

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsTestBase.java
```diff
@@ -320,6 +346,104 @@ public void shouldNotNotifyDroppedListenerWhenTransactionAddedToBlock() {
...
+          return SELECTED;
+        });
+
+    assertThat(parsedTransactions.size()).isEqualTo(6);
+
+    assertThat(parsedTransactions.get(0)).isEqualTo(transaction1Sdr2);
+    assertThat(parsedTransactions.get(1)).isEqualTo(transaction2Sdr2);
+    assertThat(parsedTransactions.get(2)).isEqualTo(transaction3Sdr2);
+    assertThat(parsedTransactions.get(3))
+        .isEqualTo(transaction2); // Transaction 2 is actually the lowest nonce for this sender
```
fab-10, line 393: since the name of the tx is misleading with its nonce, you could consider to rename

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsTestBase.java
```diff
@@ -320,6 +346,104 @@ public void shouldNotNotifyDroppedListenerWhenTransactionAddedToBlock() {
...
+          return SELECTED;
+        });
+
+    assertThat(parsedTransactions.size()).isEqualTo(6);
+
+    // Check that by setting --tx-pool-disable-sender-grouping=true then sdr 1 hasn't monopolized
+    // the pool selection just because its transaction (transaction 3) is the first one to be parsed
+    // by the selector
+    assertThat(parsedTransactions.get(0)).isEqualTo(transaction1Sdr2);
+    assertThat(parsedTransactions.get(1)).isEqualTo(transaction2);
```
fab-10, line 444: Worth to add assertion for the remaining txs to be sure the fairness is kept after the first ones

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsTestBase.java
```diff
@@ -489,6 +613,10 @@ public void shouldReplaceTransactionWithSameSenderAndNonce_multipleReplacements(
             transactions.addTransaction(
                 createRemotePendingTransaction(independentTx), Optional.empty()))
         .isEqualTo(ADDED);
+    assertThat(
```
fab-10, line 616: this seems a duplication, is it wanted?

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -228,6 +231,16 @@ static class Unstable {
...
         TransactionPoolConfiguration.Unstable.ETH65_TRX_ANNOUNCED_BUFFERING_PERIOD;
+
+    @CommandLine.Option(
+        names = {TX_POOL_DISABLE_SENDER_GROUPING},
+        paramLabel = "<Boolean>",
+        hidden = true,
+        description =
+            "Disable sender grouping of transactions during selection. (default: ${DEFAULT-VALUE})",
+        arity = "0..1")
+    private Boolean disableSenderTXGrouping =
```
fab-10, line 242:
```suggestion
    private Boolean disableSenderTxGrouping =
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolConfiguration.java
```diff
@@ -47,6 +47,11 @@ default Duration getEth65TrxAnnouncedBufferingPeriod() {
     default int getTxMessageKeepAliveSeconds() {
       return DEFAULT_TX_MSG_KEEP_ALIVE;
     }
+
+    @Value.Default
+    default Boolean getDisableSenderTXGrouping() {
```
fab-10, line 52:
```suggestion
    default Boolean getDisableSenderTxGrouping() {
```

fab-10 (commented):
suggested renaming to uniform the style

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsTestBase.java
matthew1001: Fixed

matthew1001:
@fab-10 after updating the TX selection ordering, transactions are no longer grouped by sender (at least not in a way that is unfair to specific senders). I've repurposed this PR to add unit tests which demonstrate that to be the case.

matthew1001:
@fab-10 I've marked this as auto-merge if/when you're happy with the new test.

fab-10 (approved):

PR #6000, opened 2023-10-09
[WIP] Trie log pruning 

siladu:
Added some subcommands for debugging...

(based on https://github.com/ahamlat/RocksdDB-Column-Families-Size)
```
Usage: besu operator x-rocksdb [-hV] [COMMAND]
Print RocksDB information
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
Commands:
  usage  Prints disk usage
```

```
Usage: besu operator x-trie-log [-hV] [--block[=<LONG>]] [--from[=<LONG>]] [--to
                                [=<LONG>]] [COMMAND]
Manipulate trie logs
      --block[=<LONG>]   The block
      --from, --from-block-number[=<LONG>]
                         Start of the block number range
  -h, --help             Show this help message and exit.
      --to, --to-block-number[=<LONG>]
                         End of the block number range
  -V, --version          Print version information and exit.
Commands:
  count   This command counts all the trie logs
  list    This command lists all the trie logs
  delete  Deletes the trie logs stored under the specified hash or block number
            range.
  prune   This command prunes all trie log layers below the specified block
            number, including orphaned trie logs.
```

Steps to install on a box:
```
cd /opt/besu
sudo wget -O 6000.tar.gz https://output.circle-artifacts.com/output/job/4d3d3d9c-25d5-41f1-ba04-d888dc1e4413/artifacts/0/distributions/besu-23.10.1-SNAPSHOT.tar.gz
sudo tar zxf 6000.tar.gz --transform s/besu-23.10.1-SNAPSHOT/6000/
time sudo /opt/besu/6000/bin/besu --config-file=/etc/besu/config.toml operator x-rocksdb usage
```

```
# offline only command:
time sudo /opt/besu/6000/bin/besu --config-file=/etc/besu/config.toml operator x-trie-log list
```

Issue #5390, opened 2023-04-25
Removal of old trie logs specified by historical block limit


non-fungible-nelson:
There is no need to keep all the trie logs for a near-head node as the state for old blocks is rarely needed. This will place a configurable limit on how many trie logs are kept reducing storage needed for a node.

* Add configurable limit for number trie logs to be kept. This should be the same as the Bonsai read limit config 
--bonsai-historical-block-limit parameter
* Prune old trie logs that are not needed
* Trie logs are stored by block hash so there will need to be a way of determining with block hashes to remove the trie logs for
* The logic used for the block pruning which keeps a list of block hashes for each block number could possibly be reused for pruning the trie log
* This should be enabled independently of the chain data pruning feature and can be used together
* Measure storage saved by having the trie log pruning enabled
* Test that reorgs have their trie log pruned

Acceptance Criteria
* DB size of long running node should be the same before and after resync

Notes:
* The `ChainDataPruner` could be a helpful place to start and reuse some of the same logic

Tasks:
- [x] First iteration of `--Xbonsai-trie-log-pruning-enabled`: https://github.com/hyperledger/besu/pull/6026 
- [x] Consider using safeBlock instead of 512 layers to retain
  - would be a nice strategy to prune only blocks older than RETAINED_LAYERS that have a subsequent finalized blockhash
- [x] Consider belts and braces protection during rollback/forward in case pruned trielog may be in use
  - not sure how useful this is now we only prune finalized blocks
- [x] DB size of long running node should be the same before and after resync - 
  - [x]  without flat db https://gist.github.com/siladu/0a6fe37fe6de350cef233136e237a6e4
  - [x] with flat db https://gist.github.com/siladu/a200ca16b5189a827e14eaa74be21ad7
  - [x] with flat db, using debug_worldSyncState https://gist.github.com/siladu/0a6fe37fe6de350cef233136e237a6e4
- [x] Measure storage saved by having the trie log pruning enabled - underway on `*simon-6026*` and `*simon-flatdb*` nodes
     ```
       crontab -e 
       
       0 16 * * * echo "\n\n=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=\n$(date)\n" >> /home/simon.dudley/rocksdb-usage.txt 2>&1
       1 16 * * * time /opt/besu/6000/bin/besu --config-file=/etc/besu/config.toml operator x-rocksdb usage >> /home/simon.dudley/rocksdb-usage.txt 2>&1
     ```
[...]

PR #6026, opened 2023-10-12
Add trie log pruning triggered after trie log persist

siladu:
First piece of https://github.com/hyperledger/besu/issues/5390

TrieLogPruner loads ~all~ a limited number of trie logs on startup to preload the pruner ~cache~ queue, based on loadingLimit with a default value of 30,000 blocks (configured with `--Xbonsai-trie-log-prune-limit`)

Each time a trie log is persisted it is added to the pruner ~cache~ queue and then the pruner is run against the ~cache~ queue, which will prune trie logs associated with block numbers **below** the `--Xbonsai-trie-log-retention-threshold` (defaults to 0 === disabled). 
The retention-threshold is approximately the size of the trie log pruning ~cache~ queue (it is measured in blocks but the multimap ~cache~ queue may contain forks, so they can differ).

### Trie log backlog management 
There are two reasons the database might contain old trie logs that don't exist in the ~cache~ queue:
1. Pruning was enabled on an existing node that already has trie logs in the database (backlog could be quite large in this case).
1. Besu was restarted, clearing the pruner ~cache~ queue but leaving the database with the trie logs associated with the blocks _above_ the retention-threshold.

The backlog is gradually pruned: a single set of trie logs are loaded each time besu is started up, which ensures pruneable trie logs aren't forgotten when besu gets restarted, whilst avoiding any complicated logic for progressively managing the backlog. 

A loadingLimit > retention-threshold is desirable since if besu is restarted, we want to at least preload the amount of trie logs that were forgotten by the ~cache~ queue.

Once backlog is cleared, each prune run should just be a single trie log.

Depending on the size of the backlog, it is possible that it will never be cleared. This is particularly true for nodes that have been running before this feature was enabled. To mitigate this, we can offer one of two options:
1. A one-off subcommand to trim the trie logs down to the retention-threshold (future PR, see the WIP https://github.com/hyperledger/besu/pull/6000)
1. Resync besu

### TODO:
- [x] Test with local node using smaller defaults for retain/prune window
- [x] Add feature toggle
- [x] Feasibility of streaming all trie logs on startup? Edit: Based on [test results below](https://github.com/hyperledger/besu/pull/6026#issuecomment-1763991623) it's probably not feasible to stream all in one go for nodes with a large backlog of trielogs
- [x] Test on real nodes using current defaults: retain latest 512 blocks; backlog prune window of 1000 blocks;
- [x] Test mainnet to determine reasonable pruningLimit, which will impact besu startup time.
	- Suggest 30,000 loading/pruningLimit - this impacts startup by ~4 seconds.
- [x] Deal with "orphaned" trie logs (related to building: ones that aren't referenced in the `blockchain`; neither canonical nor a fork)
  - [x] Prune any orphaned trie logs found during the cache preload
  - [x] ~Prevent block building from saving orphaned trie logs~ see below
- [x] Test engine API perf impact 
	- [x] [First iteration](https://github.com/hyperledger/besu/pull/6026#issuecomment-1789761107)
	- [x] [Second iteration](https://github.com/hyperledger/besu/pull/6026#issuecomment-1798027390)
- [x] Test pruner cache/queue size
  - [x] https://github.com/hyperledger/besu/pull/6026#issuecomment-1813837274
- [x] Reorg test
  - [x] https://github.com/hyperledger/besu/pull/6026#issuecomment-1811938360
- [x] Test on existing validator canary once code is finalized (which should have orphaned trielogs)
  - [x] https://github.com/hyperledger/besu/pull/6026#issuecomment-1815546252

### TODO Next PR(s):
- [x] Consider using safeBlock instead of 512 layers to retain
  - would be a nice strategy to prune only blocks older than RETAINED_LAYERS that have a subsequent finalized blockhash
- [ ] Consider belts and braces protection during rollback/forward in case pruned trielog may be in use
  - not sure how useful this is now we only prune finalized blocks
- [ ] Prevent block building from saving orphaned trie logs 
- [ ] Consider an option to only prune forks
- [ ] Subcommand to one-off prune the backlog, see https://github.com/hyperledger/besu/pull/6000
- [ ] Move pruning to async thread
- [ ] Prune in batches?


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -203,6 +204,10 @@ public Optional<byte[]> getTrieLog(final Hash blockHash) {
     return trieLogStorage.get(blockHash.toArrayUnsafe());
   }
 
+  public Stream<byte[]> streamTrieLogs() {
+    return trieLogStorage.streamKeys();
```
siladu, line 208: Used to preload the pruner. Need to test the feasibility of this on real nodes, might need to limit it.

matkt: yes seems to be dangerous to do that without limit

garyschulte: agreed. If we do not close the RocksIterator, that is going to lead to problems.  I would suggest passing in a Consumer<TrieLog> or something like that so we can compartmentalize the iterator open/close

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.stream.Stream;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.TreeMultimap;
+import org.apache.tuweni.bytes.Bytes32;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TrieLogPruner {
```
garyschulte, line 39: I would rather see the pruning implemented in a subclass of CachedWorldStorageManager.  This feels like a management function that could benefit from knowing the current state of cache.  we could ensure that DEFAULT_MAX_BLOCKS_TO_PRUNE doesn't conflict with AbstractTrieLogManager.RETAINED_LAYERS, etc.

Having a separate pruner class breaks encapsulation imo.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/AbstractTrieLogManager.java
```diff
@@ -88,6 +90,9 @@ public synchronized void saveTrieLog(
       } finally {
         if (success) {
           stateUpdater.commit();
+          trieLogPruner.rememberTrieLogKeyForPruning(
+              forBlockHeader.getNumber(), forBlockHeader.getBlockHash().toArrayUnsafe());
```
garyschulte, line 94: since block number isn't unique for trielogs, we will get collisions during reorgs

siladu: yep, the data structure hold forks for pruning too (it's a multimap)

siladu:
Test results for this first iteration on some `~3 week old` nodes with `~150K trie logs` using 
`numBlocksToRetain = 512;`
`pruningWindowSize = 1000;`

Most of the work is in streaming all the trie log entries into the `knownTrieLogKeysByDescendingBlockNumber` cache at startup, which currently blocks from besu starting up e.g. 

```
{"@timestamp":"2023-10-16T01:04:07,335","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loading trie logs from database...","throwable":""}
{"@timestamp":"2023-10-16T01:04:47,761","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loaded 134958 trie logs from database","throwable":""}
```

| Network | Number of trie log | Loading Time (s) |
|---------|--------------------|------------------|
| Sepolia | 165213 | 32 |
| Goerli | 141391 | 30 |
| Mainnet | 134956 | 40 |
| Mainnet | 515 |  24 |

The pruning windows themselves were performant since RocksDB marks for deletion, the range of time taken:
`pruningWindowSize = 1000: 10 - 27 ms`
`pruningWindowSize = 1: 2 - 4ms`  (normal operation once backlog is cleared)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: I think the main problem with this is that at the `CachedWorldStorageManager` level, the `cachedWorldStatesByHash` cache only maintains canonical blocks, not forks...in other words `CacheWorldStorageManager.addCachedLayer` (and therefore scrubCachedLayers) is only called for canonical blocks (when the head is moved after an FCU AFAICT).

So we'd miss out on pruning forks this way.

Am still investigating potential benefits of the pruning being aware of this cache

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
siladu: Now I'm simply using `streamKeys().limit(..)`, hopefully that is safe!

[`8658aec` (#6026)](https://github.com/hyperledger/besu/pull/6026/commits/8658aec93fbe1d54d0166247062808f53dbc8629)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: As discussed in person, these concepts (and therefore caches) are separated out now. There may still be a task to align these in code rather than relying on configuration.

siladu:
>  Test mainnet to determine reasonable pruningLimit, which will impact besu startup time.

Testing various number of trie log limits for preloading pruning cache.
Tested on mainnet nodes ~5 weeks old with ~240,000 trie logs in the database.

| Instance | Loading Limit | Loading Time avg (s) |
|---------|--------------------|------------------|
| dev-elc-besu-nimbus-mainnet-simon-memory-23.7.2-1 | 500 | 0.646 |
| dev-elc-besu-nimbus-mainnet-simon-memory-23.7.2-1 | 5000 | 1.4 |
| dev-elc-besu-nimbus-mainnet-simon-memory-23.7.2-1 | 50000 | 8 |
| dev-elc-besu-nimbus-mainnet-simon-memory-23.7.2-1 | 25000 | 3.6 |
| dev-elc-besu-nimbus-mainnet-simon-memory-23.7.2-2 | 30000 | 4.4 |
| dev-elc-besu-nimbus-mainnet-simon-memory-23.7.3-RC-1 | 30000 | 4.6 |

siladu:
> Test engine API perf impact

This is using the first iteration of this code that has been running for a while...

besu-trielog-pruning-perf

siladu:
Ready for code review, but leaving in draft until testing is complete.

matkt:
My question is do we want to have another protection to not prune block before finalization . For example even if your value is 10, the pruner will keep more than 10 because we don't have yet a finalization, my question is how to manage a long period without finalization . So maybe we need to take care of the threeshold as long as this trielog are after a block finalization

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,186 @@
...
+    knownTrieLogKeysByDescendingBlockNumber.put(blockNumber, trieLogKey);
+  }
+
+  void pruneFromCache() {
+
+    final long retainAboveThisBlock = blockchain.getChainHeadBlockNumber() - numBlocksToRetain;
+    final long retainAboveThisBlockOrFinalized =
+        blockchain
+            .getFinalized()
+            .flatMap(blockchain::getBlockHeader)
```
siladu, line 108: Maybe could save a db call if we combine into one transaction

siladu:
> My question is do we want to have another protection to not prune block before finalization...

@matkt This was on the TODO list for next PR anyway, but I decided to include it in this one (pending testing):
https://github.com/hyperledger/besu/pull/6026/commits/9ac5efad0cb817eedb2a7b09eb5f8bfdfd7d3874

Note, this adds 2 db calls to find the finalized block number (every time we save a trie log), maybe that is something we can optimize to 1 db call. I will test performance impact of the extra 2 db calls anyway.

matkt:
> > My question is do we want to have another protection to not prune block before finalization...
> 
> @matkt This was on the TODO list for next PR anyway, but I decided to include it in this one (pending testing): [9ac5efa](https://github.com/hyperledger/besu/commit/9ac5efad0cb817eedb2a7b09eb5f8bfdfd7d3874)
> 
> Note, this adds 2 db calls to find the finalized block number (every time we save a trie log), maybe that is something we can optimize to 1 db call. I will test performance impact of the extra 2 db calls anyway.
> 
> Another potential issue this could introduce in a long period of finalization is unbounded cache growth, though it's quite a small cache 🤔 I suspect besu would run into other problems first before hitting this one.

ok just wanted to be sure we are not missing that, honestly I think it will be better to start by a pruning after finalization and add the threshold in a second step like that we have everytime something safe . But as you already did the implementation I think it’s ok

Also In my opinion having pruning-enabled will be better not fan to have =0 to disable the feature

> Note, this adds 2 db calls to find the finalized block number (every time we save a trie log), maybe that is something we can optimize to 1 db call. I will test performance impact of the extra 2 db calls anyway.
We are using that during snapsync in order to have last safeblock I think we can maybe use the same logic and like that we will not need to read database https://github.com/hyperledger/besu/blob/6a2d41fccc00eac110a8cec27bc16cc2df47da0a/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java#L860

besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
```diff
@@ -266,13 +290,22 @@ public String build() {
...
+    lines.add("Using " + txPoolImplementation + " transaction pool implementation");
+
     if (isHighSpec) {
       lines.add("Experimental high spec configuration enabled");
     }
 
-    lines.add("Using " + txPoolImplementation + " transaction pool implementation");
     lines.add("Using " + worldStateUpdateMode + " worldstate update mode");
 
+    if (trieLogRetentionThreshold > 0) {
```
matkt, line 301: do we can use a second flag to disable the feature instead of 0 ?

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -52,6 +55,25 @@ public class DataStorageOptions implements CLIOptions<DataStorageConfiguration>
...
+    @CommandLine.Option(
+        hidden = true,
+        names = {"--Xbonsai-trie-log-retention-threshold"},
+        description =
+            "The number of blocks for which to retain trie logs. 0 is unlimited. (default: ${DEFAULT-VALUE})")
+    private long bonsaiTrieLogRententionThreshold = DEFAULT_BONSAI_TRIE_LOG_RETENTION_THRESHOLD;
+
+    @CommandLine.Option(
+        hidden = true,
+        names = {"--Xbonsai-trie-log-prune-limit"},
```
matkt, line 71: I understood that we don't want to use pruning ? And for me the name and the description doesn't match a lot maybe `retention_trie_log_cache_limit`?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,187 @@
...
+          final Optional<BlockHeader> header = blockchain.getBlockHeader(hash);
+          if (header.isPresent()) {
+            knownTrieLogKeysByDescendingBlockNumber.put(header.get().getNumber(), blockHashAsBytes);
+            count.getAndIncrement();
+          } else {
+            // prune orphaned blocks (sometimes created during block production)
+            rootWorldStateStorage.pruneTrieLog(blockHashAsBytes);
+          }
+        });
+    LOG.atInfo().log("Loaded {} trie logs from database", count);
```
matkt, line 90: do we need info for that ? debug is not enough ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,187 @@
...
+    this.pruningLimit = pruningLimit;
+    this.loadingLimit = pruningLimit; // same as pruningLimit for now
+  }
+
+  public void initialize() {
+    preloadCache();
+  }
+
+  private void preloadCache() {
+    LOG.atInfo()
```
matkt, line 72: do we need info for that ? debug is not enough ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
matkt: why not using the same logic as snapsync, https://github.com/hyperledger/besu/blob/6a2d41fccc00eac110a8cec27bc16cc2df47da0a/besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java#L860
I think we are not calling the database with this method, maybe only the first time

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,187 @@
...
+            .getFinalized()
+            .flatMap(blockchain::getBlockHeader)
+            .map(ProcessableBlockHeader::getNumber)
+            .map(finalizedBlock -> Math.min(finalizedBlock, retainAboveThisBlock))
+            .orElse(retainAboveThisBlock);
+
+    LOG.atTrace()
+        .setMessage(
+            "min((chainHeadNumber: {} - numBlocksToRetain: {}), finalized: {})) = retainAboveThisBlockOrFinalized: {}")
+        .addArgument(blockchain.getChainHeadBlockNumber())
```
matkt, line 117: maybe we can call this method only one time https://github.com/hyperledger/besu/pull/6026/files#diff-60f4623b862cc8abfc075f66848a976b23470f2d6f1ce1124773835116594110R105

matkt (changes requested):

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
matkt: not sure about my proposition

besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
siladu: sure thing, any suggestions for a name (ideally that avoids "pruning")?

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
siladu: This is not the same as the retention threshold so would be concerned about confusion there. Perhaps we need a better overall feature name to group them.
 
This option is being used both as a cache loading limit as well as a prune limit, with the loading limit being the more significant factor in performance. I like "cache-limit", although not strictly true technically the cache can grow higher than this limit, especially with the finalization changes.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: Maybe debug is OK. It's only printed once at startup though. Thought it might be useful to discuss with users who may ask about disk reduction once this is released.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: thanks, that might be a nice way to avoid the db...will look into it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: true, I figured it would only get called during trace logging, but still.

besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
siladu: Gone with `--Xbonsai-trie-log-pruning-enabled` against for now https://github.com/hyperledger/besu/pull/6026/commits/bde0bd27ca9dc347c6f212f471207503e6e630e1

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
siladu: So we have:
--Xbonsai-trie-log-pruning-enabled
and
--Xbonsai-trie-log-pruning-limit

where "bonsai-trie-log-pruning" is like the feature name.

This pruning limit is being used as the "cache preload limit" (`loadingLimit`) as well as the "pruning window limit": https://github.com/siladu/besu/blob/3a779ea53ed2c9a2b438eb91c2d85231ebd59b1d/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java#L133

I think `bonsai-trie-log-cache-preload-limit` could equally work and still be accurate.

`bonsai-trie-log-cache-limit` is perhaps a little misleading.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: This is because it introduces a small delay (currently tested as 4-8 seconds with default limit of 30,000) 
so I wanted to indicate to the user what the delay was for...similar to the "compacting database..." message

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: chainHead is cached so it's not a db call, but I've made sure we're only executing it when trace is enabled...
https://github.com/hyperledger/besu/pull/6026/commits/c74b104fc0779f8e8375b0f3e63e7e3e85d710b4

besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
matkt: if we really want to remove pruning maybe this one ? `Xbonsai-trie-log-eviction-enabled`just a proposition not really pushing for that

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
matkt: is this cache is used also for block processing or validation bonsai-trie-log-cache-preload-limit ?

matkt: just trying to understand if it's only used for the pruning or also for trielog access in general

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
matkt: ah ok maybe fine if it's only displayed once

siladu:
Second iteration tested on the following deployments: 

| Instance | Description | Commit Ref |
|---------|--------------------|------------------|
| dev-elc-besu-teku-mainnet-simon-6026-control-* | Last merged in change from main | https://github.com/hyperledger/besu/commit/236779d3ceb216aa4155258f228032a75c05d6d2 |
| dev-elc-besu-teku-mainnet-simon-6026-prune-* | Without finalized block check: no extra db calls due to pruner cache (except on startup) | https://github.com/hyperledger/besu/pull/6026/commits/bd53bdf2af91d3013d6ed96277b1a7d839d506c4 |
| dev-elc-besu-teku-mainnet-simon-6026-prune-finalized-* | Same as `-prune` but with extra finalized block check: [two extra db calls](https://github.com/siladu/besu/blob/b0c72ca94ebd4a3ed7142034651314d8e703ec7e/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java#L107-L109) | https://github.com/hyperledger/besu/pull/6026/commits/6d22bf7c56a977901cc9c7f60307562568444a20

besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
siladu: I will discuss name with Matt, but I think for now it doesn't matter while it's an experimental option. 
Despite what I'm said before, I'm currently thinking `pruning` might be OK after all because it accurately describes what it does and eventually I don't expect most users will need to know/worry about this feature.

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
siladu: > is this cache is used also for block processing or validation bonsai-trie-log-cache-preload-limit ?
This cache is only accessed by the trieLogPruner and trieLogManager, which adds to it upon saving a trie log.

> just trying to understand if it's only used for the pruning or also for trielog access in general
yes, only pruning

matkt (approved):
LGTM as it's experimental feature, we will have to do other PRs to do flag renaming and call db optimization

besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
```diff
@@ -266,13 +301,22 @@ public String build() {
...
 
-    lines.add("Using " + txPoolImplementation + " transaction pool implementation");
     lines.add("Using " + worldStateUpdateMode + " worldstate update mode");
 
+    if (isTrieLogPruningEnabled) {
+      lines.add("Trie log pruning enabled:");
+      lines.add("  - retention threshold: " + trieLogRetentionThreshold + " blocks");
+      if (trieLogPruningLimit != null) {
+        lines.add("  - prune limit: " + trieLogPruningLimit + " blocks");
+      }
```
jframe, line 317: nit: seems a little verbose to be using three lines in the config logging. can to be shortened to one line?

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -42,16 +48,41 @@ public class DataStorageOptions implements CLIOptions<DataStorageConfiguration>
...
+        hidden = true,
+        names = {"--Xbonsai-trie-log-pruning-enabled"},
+        description = "Enable trie log pruning. (default: ${DEFAULT-VALUE})")
+    private boolean bonsaiTrieLogPruningEnabled = DEFAULT_BONSAI_TRIE_LOG_PRUNING_ENABLED;
+
+    @CommandLine.Option(
+        hidden = true,
+        names = {"--Xbonsai-trie-log-retention-threshold"},
+        description =
+            "The number of blocks for which to retain trie logs. (default: ${DEFAULT-VALUE})")
```
jframe, line 76: Can the minimum value be specified in the description as well

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,188 @@
...
+  private void preloadCache() {
+    LOG.atInfo()
+        .setMessage("Loading first {} trie logs from database...")
+        .addArgument(loadingLimit)
+        .log();
+    final Stream<byte[]> trieLogKeys = rootWorldStateStorage.streamTrieLogKeys(loadingLimit);
+    final AtomicLong count = new AtomicLong();
+    trieLogKeys.forEach(
+        blockHashAsBytes -> {
+          Hash blockHash = Hash.wrap(Bytes32.wrap(blockHashAsBytes));
```
jframe, line 80: nit: final

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,188 @@
...
+          final Optional<BlockHeader> header = blockchain.getBlockHeader(blockHash);
+          if (header.isPresent()) {
+            trieLogBlocksAndForksByDescendingBlockNumber.put(header.get().getNumber(), blockHash);
```
jframe, line 83: Is it possible that not enough trielog are not loaded such that we don't fill the gap between head and minimum retention threshold? I think it could only happen if there are a very large number of trie logs as the pruning limit in most cases would load all the tries.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,188 @@
...
+            .dropWhile((e) -> e.getKey() > retainAboveThisBlockOrFinalized)
+            .limit(pruningLimit);
+
+    final List<Long> blockNumbersToRemove = new ArrayList<>();
+
+    final AtomicInteger count = new AtomicInteger();
+    pruneWindowEntries.forEach(
+        (e) -> {
+          for (Hash blockHash : e.getValue()) {
+            rootWorldStateStorage.pruneTrieLog(blockHash);
```
jframe, line 141: Is this wrapped in a transaction?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,188 @@
...
+            count.getAndIncrement();
+          }
+          blockNumbersToRemove.add(e.getKey());
+        });
+
+    blockNumbersToRemove.forEach(trieLogBlocksAndForksByDescendingBlockNumber::removeAll);
```
jframe, line 147: If there is an error with pruning a trielog could end up with an inconsistency in the trieLogBlocksAndForksByDescendingBlockNumber map since only some trielogs might be removed.

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
siladu: Limitation of PicoCLI unfortunately, these are pre-defined constants

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: As discussed, there will always be future opportunities to prune the gap each time besu is started (and the cache is preloaded). This works because the pruningLimit >> retainedTrieLogs. It might be worth enforcing this property in code to make sure the trie logs can't end up growing again.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: No, will consider it for performance reasons, maybe in future PR.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
siladu: Good catch, I think if StorageException occurs, we will bubble up too far. Catching this error inside the foreach should allow me to keep `blockNumbersToRemove` consistent. Any trie logs that missed pruning due to error will remain in the cache and will be picked up in the next pruner execution (next block).

siladu: Handled this here https://github.com/hyperledger/besu/pull/6026/commits/4681a858386dc9165b9850b427a1a26bfbbbc500

besu/src/main/java/org/hyperledger/besu/cli/ConfigurationOverviewBuilder.java
siladu: Done this here: https://github.com/hyperledger/besu/pull/6026/commits/6b275220acfe18531b1fcb91e2a5351986679946
Looks like this:
```
# Trie log pruning enabled: retention: 512; prune limit: 30000                                     #
```

siladu:
I setup a reorg test based on [this hive test](https://github.com/ethereum/hive/blob/900b75d75fe681b72be1a5f9025cc3e89c80b714/simulators/ethereum/engine/suites/engine/invalid_ancestor.go#L71-L74), with some modifications to send finalized blocks trailing the head by 10 blocks. Note it uses a mock CL.

In the abbreviated logs below, we:
 - build a chain and start finalizing it
 - simultaneously build a sidechain starting from 6
 - reorg to a sidechain at Step 16
 - reorg back to the original chain at Step 26.

The test was setup with an artificially low `--Xbonsai-trie-log-retention-threshold=1` to ensure that we were relying on the finalizedBlock check to prevent premature pruning.

```
Step | FCU | Block | FinalizedBlock
## 0. FCU 0
## 1. FCU 1
## 2. FCU 2
## 3. FCU 3
## 4. FCU 4
## 5. FCU 5 (common ancestor)
## 6. FCU 6
## 7. FCU 7
## 8. FCU 8
## 9. FCU 9
## 10. FCU 10
## 11. FCU 11 FINALIZED 1
## 12. FCU 12 FINALIZED 2
## 13. FCU 13 FINALIZED 3
## 14. FCU 14 FINALIZED 4
## 15. FCU 15 FINALIZED 5
## 16. FCU 6' FINALIZED 5
## 17. FCU 7' FINALIZED 5
## 18. FCU 8' FINALIZED 5
## 19. FCU 9' FINALIZED 5
## 20. FCU 10' FINALIZED 5
## 21. FCU 11' FINALIZED 5
## 22. FCU 12' FINALIZED 5
## 23. FCU 13' FINALIZED 5
## 24. FCU 14' FINALIZED 5
## 25. [INVALID] FCU 15' [INVALID] FINALIZED 5
## Resend the latest correct fcU
## 26. FCU 15 FINALIZED 5
## 27. FCU 16 FINALIZED 6
## 28. FCU 17 FINALIZED 7
```

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPrunerTest.java
```diff
@@ -43,8 +43,10 @@ public class TrieLogPrunerTest {
   private BonsaiWorldStateKeyValueStorage worldState;
   private Blockchain blockchain;
 
+  @SuppressWarnings("BannedMethod")
   @BeforeEach
   public void setup() {
+    Configurator.setLevel(LogManager.getLogger(TrieLogPruner.class).getName(), Level.TRACE);
```
matkt, line 51: why this one is needed for the test ?

matkt (approved):
LGTM seems to be fine with finalized block modification

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPrunerTest.java
siladu: It's not _needed_ - purely convenience for comprehending the tests. I thought it was useful enough to leave in but happy to remove.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogPruner.java
```diff
@@ -0,0 +1,195 @@
...
+    preloadQueue();
+  }
+
+  private void preloadQueue() {
+    LOG.atInfo()
+        .setMessage("Loading first {} trie logs from database...")
+        .addArgument(loadingLimit)
+        .log();
+    try {
+      final Stream<byte[]> trieLogKeys = rootWorldStateStorage.streamTrieLogKeys(loadingLimit);
```
jframe, line 74: Think this should be done in a try-with-resources so that we close the stream

siladu: done https://github.com/hyperledger/besu/pull/6026/commits/13f0040867102e9c6b0a627b3d61eb792769d0f0

siladu:
Test pruner cache/queue size, with default value of 512 retention, size of the cache/queue is **~140KB**
queue-size-512-trielog-keys

`140 / 12` = **~275 bytes per trielog**

In a non-finality event, this **cache is unbounded**, however since the size is small I don't think this will be a problem. More likely other parts of besu would break first.

For context, the [recent non-finality event](https://medium.com/offchainlabs/post-mortem-report-ethereum-mainnet-finality-05-11-2023-95e271dfd8b2) lasted a max of 9 epochs. Max of 32 blocks per epoch = `9 * 32 = 288 blocks` which within the default bound of 140KB.

To reach Megabyte magnitude, the non-finality event would have to be 3,636 blocks long (1 GB / 275 bytes)
To reach Gigabyte magnitude, the non-finality event would have to be 3,636,363 blocks long (1 GB / 275 bytes)

jframe (approved):
LGTM

ahamlat:
Checking the performance and the implementation, I think this should be done asynchronously to not have an overhead on the block processing time. Also, I think handling this case on either OnblockAdded event or onTrieLogAdded event, and process the event asynchronously.
Nonetheless, I didn't observe any performance degradation. Given that this is hidden behind an experimental flag, I approve the PR from a performance overhead perspective. The async optimization can be done in another PR.

siladu:
### Test on existing validator canary once code is finalized (which should have orphaned trielogs)

**TL;DR it works 🎉** 

Tested on `prd-elc-besu-lighthouse-sepolia-bonsai-snap`

Trie log count before enabling pruning:
```
trieLog count: 742311
 - canonical count: 681039
 - fork count: 217
 - orphaned count: 61055
 ```

As expected, there are orphaned trielogs due to block creation. The number is high because this sepolia node is 20 / 1972 validators = 1% of the network so does an unusual amount of block creation compared to mainnet nodes.

#### Log snippets

```
# Besu version 23.10.3-dev-81115bd9                                                                #
...
# Trie log pruning enabled: retention: 512; prune limit: 30000                                     #
```

Example of the initial load picking up a batch of trielogs. Note, we don't control the order of the trielog loading due to the way they are stored - it appears to be quite random. The upshot is that not all of them are eligible for pruning immediately because they may fall inside the retention window...
```
2023-11-16T09:19:55,439 INFO TrieLogPruner "Loading first 30000 trie logs from database..."
2023-11-16T09:20:04,694 INFO TrieLogPruner "Loaded 27528 trie logs from database"
2023-11-16T09:20:04,928 DEBUG TrieLogPruner "pruned 27470 trie logs from 27469 blocks"
```

After that (and every restart), there's a period of no pruning while the queue builds up its 512 block retention window. This period lasts ~2 hours. Note pruning triggers on both block processing thread and block building thread...
```
{"@timestamp":"2023-11-16T09:20:21,302","level":"TRACE","thread":"vert.x-worker-thread-0","class":"TrieLogPruner","message":"pruned 0 trie logs for blocks {}","throwable":""}
...
{"@timestamp":"2023-11-16T09:32:56,171","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"pruned 0 trie logs for blocks {}","throwable":""}
...
{"@timestamp":"2023-11-16T11:12:12,820","level":"TRACE","thread":"vert.x-worker-thread-0","class":"TrieLogPruner","message":"pruned 0 trie logs for blocks {}","throwable":""}
```

During this time we prune the odd trielog, which I believe occurs when an queue entry that was preloaded during startup becomes eligible.

Once the queue size reaches 512 retention window, we start pruning ~1 block every block (one in, one out).
At this point you will notice the trielogs are pruned in order since it's a sliding window by block number...
```
{"@timestamp":"2023-11-16T11:12:24,376","level":"TRACE","thread":"vert.x-worker-thread-0","class":"TrieLogPruner","message":"pruned 1 trie logs for blocks {4704297=[0x1e335684e0c7d41a1a3bc13b321e2f27b73672c6a37515e1c735bc8ac213c393]}","throwable":""}
{"@timestamp":"2023-11-16T11:12:36,385","level":"TRACE","thread":"vert.x-worker-thread-0","class":"TrieLogPruner","message":"pruned 1 trie logs for blocks {4704298=[0x2875008c2ee12ab366d7969e6682a9928260c359e93333e0513ca87015df3a56]}","throwable":""}
{"@timestamp":"2023-11-16T11:13:00,985","level":"TRACE","thread":"vert.x-worker-thread-0","class":"TrieLogPruner","message":"pruned 1 trie logs for blocks {4704299=[0x913742463b968f41cc4e101844493c34e87cd444def91ca386ed6c09b218cb16]}","throwable":""}
```

When we propose a block, we actually save multiple trie logs, but only the latest one makes it into the chain (hence the orphaned trielogs). During pruning (512 blocks later), these will show up as a list of trie logs stored against the same block number (similar to forks).
```
{"@timestamp":"2023-11-16T09:32:56,018","level":"INFO","thread":"vert.x-worker-thread-0","class":"MergeCoordinator","message":"Start building proposals for block 4704354 identified by 0x00688f3cd98f37a0","throwable":""}
...
// FCU thread creating the initial empty block...
{"@timestamp":"2023-11-16T09:32:56,017","level":"TRACE","thread":"vert.x-worker-thread-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0xd2b16e03e6d312a9989da949ac06a7cf6319e575f63ef1fbedba2b0ce9ce275b","throwable":""}
// Block creation loop...
{"@timestamp":"2023-11-16T09:32:56,171","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0x6724630e7d47e454de20b95a913a4997a3ce8bf138c04e413045bab896435aa9","throwable":""}
{"@timestamp":"2023-11-16T09:32:56,677","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0x656bae99dd673404531b9fe95b09bf0a281df9c803d0c11b5b054838046ecd3f","throwable":""}
{"@timestamp":"2023-11-16T09:32:57,144","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0x4379fe26a8e77cfaa4668065fceca9cefe2fbb64ee0213eedc3993196177743d","throwable":""}
{"@timestamp":"2023-11-16T09:32:57,728","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0xc87cdd96177387b20327573e682a23147215c99597599a9280848e9c54152622","throwable":""}
{"@timestamp":"2023-11-16T09:32:58,262","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0x3e3a13b87a3ef3b342581fab6d3794a563aae79f5c11feb73b2415355da341d8","throwable":""}
{"@timestamp":"2023-11-16T09:32:58,744","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0x39b3f1ca18ca4a6c5a15dd3b8076e03724393412953c4d378cbb4caf61b213c8","throwable":""}
{"@timestamp":"2023-11-16T09:32:59,251","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0x9cf0aea39d630ffcf79458a89348ccb584991051ba80e4672df83fca40fa8bcc","throwable":""}
{"@timestamp":"2023-11-16T09:32:59,756","level":"TRACE","thread":"EthScheduler-BlockCreation-0","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 4704354; blockHash 0x0ec8491df4952c450503933d43e8c8163881a335f00816ed37b8bd0f42bb3085","throwable":""}
...
{"@timestamp":"2023-11-16T11:24:24,577","level":"TRACE","thread":"vert.x-worker-thread-0","class":"TrieLogPruner","message":"pruned 9 trie logs for blocks {4704354=[0x0ec8491df4952c450503933d43e8c8163881a335f00816ed37b8bd0f42bb3085, 0x39b3f1ca18ca4a6c5a15dd3b8076e03724393412953c4d378cbb4caf61b213c8, 0x3e3a13b87a3ef3b342581fab6d3794a563aae79f5c11feb73b2415355da341d8, 0x4379fe26a8e77cfaa4668065fceca9cefe2fbb64ee0213eedc3993196177743d, 0x656bae99dd673404531b9fe95b09bf0a281df9c803d0c11b5b054838046ecd3f, 0x6724630e7d47e454de20b95a913a4997a3ce8bf138c04e413045bab896435aa9, 0x9cf0aea39d630ffcf79458a89348ccb584991051ba80e4672df83fca40fa8bcc, 0xc87cdd96177387b20327573e682a23147215c99597599a9280848e9c54152622, 0xd2b16e03e6d312a9989da949ac06a7cf6319e575f63ef1fbedba2b0ce9ce275b]}","throwable":""}
```

PR #6027, opened 2023-10-12
Mining options refactor

fab-10:
## PR description

Before adding new mining options for upcoming features, I took some time to review and refactor mining options.
Mining parameters are a bit odd when compared to other options, since they do not follow the same pattern of using immutables, as for txpool options, stable options are defined in `BesuCommand`, and since that file is already too big, adding more options there should be avoided. 

For the refactoring I have reapplied most of the solutions used for the refactoring of txpool options (see #5772 if interested), with some novelty, since for mining configuration a new feature is required, that is having the possibility to update the value of some options at runtime.
One notable difference with #5772, is that _stable_ and _unstable_ options are no more split on two classes, since it make more sense to keep all the related options in a single place, also because some validations could make use of a mix of stable and unstable options, and for that it is better to have all of them in a single place.

To implement the _update at runtime_ feature, in a thread safe way, unfortunately it is not possible to use the immutables library directly, so instead of using annotations, some code has to be added to the `MiningParameters` class to manage the updatables options. If the proposed solution prove to be effective and looks good, after it stabilize, we can also think of extending the immutables library to support this, or introduce our own annotations, or find another library that is best suited for managing configuration.
The _updatable_ parameters work like that: 
- on startup their initial value is set, as usual from config file or CLI, 
- these parameters have also a _set_, along with the normal _get_ one, 
- the values are kept in _volatile_ fields, to make it multi thread friendly
- Note that initial values are kept in case it is needed, even if the parameters  then updated at runtime, 

The proposed implementation also has the goal to centralize all the parameters in the `MiningParameters` class, and only passing this class around instead of passing single parameters, like `coinbase`, `extraData`, etc... so it should help readability and make easier to manage these parameter.

More refactoring may be proposed in following PRs, to avoid making this one too big, for example `MiningParameters` could be renamed to `MiningConfiguration` for consistency with the naming of similar classes, then could make sense to pass mining parameters directly to RPC methods that update things like `coinbase`, `targetGasLimit` etc... instead of passing through other objects.


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
```diff
@@ -166,13 +166,12 @@ public MergeCoordinator(
     this.protocolSchedule = protocolSchedule;
     this.blockBuilderExecutor = blockBuilderExecutor;
     this.mergeContext = protocolContext.getConsensusContext(MergeContext.class);
-    this.miningParameters = miningParams;
     this.backwardSyncContext = backwardSyncContext;
-    this.targetGasLimit =
-        miningParameters
-            .getTargetGasLimit()
-            // TODO: revisit default target gas limit
```
macfarla, line 174: this TODO got removed - is it still relevant?

macfarla (approved):
looks ok to me

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1489,6 +1429,7 @@ public void run() {
 
       // Need to create vertx after cmdline has been parsed, such that metricsSystem is configurable
       vertx = createVertx(createVertxOptions(metricsSystem.get()));
+      miningParameters = getMiningParameters();
```
Gabriel-Trintinalia, line 1432: Is there a reason why this needs to happen here? Inside of the Run method? Can we move it to the configure method?

fab-10:
I am running this PR on some nodes, and will wait to see there are no regressions before merging to main.

besu/src/main/java/org/hyperledger/besu/controller/IbftBesuControllerBuilder.java
```diff
@@ -145,6 +145,7 @@ protected MiningCoordinator createMiningCoordinator(
 
     final Address localAddress = Util.publicKeyToAddress(nodeKey.getPublicKey());
     final BftProtocolSchedule bftProtocolSchedule = (BftProtocolSchedule) protocolSchedule;
+    miningParameters.setCoinbase(localAddress);
```
siladu, line 148: Why do we need to do this here now?

siladu (commented):
Nice PR.
+1 for making BesuCommand smaller!

PR is a bit too big to do a thorough review without spending a lot more time on it so can't explicitly approve it. Seems a fairly safe change so won't block either.  Spot checked a few interesting parts and have some comments.

fab-10:
> PR is a bit too big to do a thorough review without spending a lot more time on it so can't explicitly approve it. Seems a fairly safe change so won't block either. Spot checked a few interesting parts and have some comments.

Thanks, will review your comment, and since there is no rush to merge this, so feel free to think about it and review in steps

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java
fab-10: let's ask to @garyschulte, that wrote the comment, what he think it is still TODO

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/MiningParameters.java
```diff
@@ -20,385 +20,264 @@
...
-      this.stratumNetworkInterface = stratumNetworkInterface;
-      return this;
-    }
+    Optional<Iterable<Long>> nonceGenerator();
+  }
 
-    public Builder stratumPort(final int stratumPort) {
-      this.stratumPort = stratumPort;
-      return this;
+  static class UpdatableRuntimeValues {
```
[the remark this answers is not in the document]
fab-10: renamed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/MiningParameters.java
```diff
@@ -20,385 +20,264 @@
...
-    private long posBlockCreationMaxTime = DEFAULT_POS_BLOCK_CREATION_MAX_TIME;
-
-    private long posBlockCreationRepetitionMinDuration =
-        DEFAULT_POS_BLOCK_CREATION_REPETITION_MIN_DURATION;
-
-    public Builder() {
-      // zero arg
-    }
+  @Value.Immutable
+  public interface UpdatableInitValues {
```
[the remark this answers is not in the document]
fab-10: renamed

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
fab-10: removed since call is not needed anymore

besu/src/main/java/org/hyperledger/besu/cli/options/MiningOptions.java
```diff
@@ -0,0 +1,303 @@
...
+  private Bytes extraData = DEFAULT_EXTRA_DATA;
+
+  @Option(
+      names = {"--min-block-occupancy-ratio"},
+      description = "Minimum occupancy ratio for a mined block (default: ${DEFAULT-VALUE})",
+      arity = "1")
+  private Double minBlockOccupancyRatio = DEFAULT_MIN_BLOCK_OCCUPANCY_RATIO;
+
+  @Option(
+      names = {"--min-gas-price"},
```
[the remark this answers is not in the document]
fab-10: This option is used when selecting txs for block creation, for any kind of network, so it make sense to group here, and I agree that the `miner` naming is now misleading, since we have different ways of building and proposing blocks, so a renaming could help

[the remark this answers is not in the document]
macfarla: Agree changing Miner Options to something like "Block Builder Options" makes sense

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -307,6 +306,9 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   final org.hyperledger.besu.cli.options.stable.TransactionPoolOptions
       stableTransactionPoolOptions = TransactionPoolOptions.create();
 
+  @CommandLine.ArgGroup(validate = false, heading = "@|bold Miner Options|@%n")
```
macfarla, line 309:
```suggestion
  @CommandLine.ArgGroup(validate = false, heading = "@|bold Block Builder Options|@%n")
```

macfarla (approved):
couple of comments. I think "Mining Options" could be renamed but that could also be a separate PR and it's not a dealbreaker

PR #6011, opened 2023-10-10
Optimize Eth_feeHistory RPC method

ahamlat:
This PR enhances the Eth_feeHistory method through the implementation of a new cache for storing computed rewards. 
Furthermore, it introduces enhancements in the process of calculating rewards.

The tests below are done on a Standard_D4s_v5 Azure VM (4 vCPU, 16 GiB RAM)

### Before this PR :
Besu wasn't able to handle 5 TPS for eth_feeHistory (with 1024 blocks) for near head calls

flood eth_feeHistory localhost:8545 -d 300 -r 5 --metrics p50 p90 p99
image

**CPU profiling**
image

### After this PR
Besu is able to handle 50 TPS with 99th percentile in 139 ms. Combining this PR with PR https://github.com/hyperledger/besu/pull/6009, Besu was able to handle 500 TPS on near head calls with 1024 blocks.

flood eth_feeHistory localhost:8545 -d 300 -r 50 --metrics p50 p90 p99

image

**CPU profiling**
image

PR #6035, opened 2023-10-16
Minor optimization on Eth_feeHistory

ahamlat:
This PR has a minor optimization on Eth_feeHistory, on top of [this other optimization on Eth_feeHistory ](https://github.com/hyperledger/besu/pull/6011).

Instead of sorting rewardPercentiles for each block header, this PR sorts only once and the sorted list is used as a parameter for each call to compute rewards.


macfarla (approved):
LGTM

PR #6046, opened 2023-10-17
Unsigned timestamps and blob gas used

jflo:
Merge after pr #6037 , was built on top of that.

the pyspec simulator in Hive likes to check fields for max capacities, which showed us a lot of places where using signed longs is a problem. This pr fixes many of them.

broader adoption of CheckerFramework is suggested at a later time, so that we can see where else we are affected by this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1093,6 +1093,22 @@ distributions {
   }
 }
 
+configurations.all {
+  resolutionStrategy {
+    dependencySubstitution {
+      def os = System.getProperty("os.name").toLowerCase()
+      if (os.contains("windows")) {
+        substitute module('org.eclipse.platform:org.eclipse.swt.${osgi.platform}') with module("org.eclipse.platform:org.eclipse.swt.win32.win32.x86_64:3.124.0")
+      }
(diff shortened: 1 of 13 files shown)


Gabriel-Trintinalia (approved):
LGTM to fix the hive tests. Nice work with the UnsignedLongParameter. It would be beneficial if we could standardise the handling of blobGasUsed and excessBlobGas across the AbstractEngineNewPayload and other related areas. Currently, despite their similar types, one is BlobGas and the other is UnsignedLong.

Something to mention is Besu's capacity to manage the maximum value. Even though we accept this value, Besu might not be capable of handling it. It's extremely unlikely to have a blobGas of this size, but we need to consider the potential implications and error handling if such a value were to be encountered.  

For instance, in the `BlobGasValidationRule`, we currently use: `long parentBlobGasUsed = parent.getBlobGasUsed().orElse(0L);` What could happen here?

This, however, is a problem that already exists and as you said in the PR's description, could be addressed with a broader use of the CheckerFramework or/and making use of the BlobGas Uint64 methods to perform those calculations

PR #6052, opened 2023-10-18
not pruning block during the initial sync

matkt:
## PR description

I propose a modification to the block pruning feature so that we can lower the minimum limit to 512 instead of 7200. Why 512? Simply to allow a minimum number of blocks close to the head. This is also the number we chose for the world state on Bonsai and seems to be a good compromise between pruning and accessibility of recent history.

7200 was chosen to avoid pruning during the initial sync, so I made a modification to prune during the initial sync by querying the SyncState.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
@@ -53,8 +53,7 @@ public class ChainDataPruner implements BlockAddedObserver {
   @Override
   public void onBlockAdded(final BlockAddedEvent event) {
     if (!isInitialSyncPhaseDoneSupplier.get()) {
-      LOG.debug(
-          "Unable to run block pruning because the initial synchronization is still in progress.");
+      // not run block pruning because the initial synchronization is still in progress.
       return;
     }
     final long blockNumber = event.getBlock().getHeader().getNumber();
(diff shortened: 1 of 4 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -32,22 +33,30 @@ public class ChainDataPruner implements BlockAddedObserver {
...
     this.blocksToRetain = blocksToRetain;
     this.pruningFrequency = pruningFrequency;
+    this.isInitialSyncPhaseDoneSupplier = isInitialSyncPhaseDoneSupplier;
     this.pruningExecutor = pruningExecutor;
   }
 
   @Override
   public void onBlockAdded(final BlockAddedEvent event) {
+    if (!isInitialSyncPhaseDoneSupplier.get()) {
+      LOG.debug(
```
jframe, line 56: This could be quite noisy on debug level during the initial sync I think. Maybe we don't need a log here.

jframe (approved):
LGTM - although doing a sync with option enabled would be best way to confirm it works as intended

macfarla (approved):
LGTM. echo @jframe 's comment about doing a sync to verify - has this been done?

garyschulte:
Is this PR still relevant/active?

macfarla:
from @matkt No we can close

PR #6054, opened 2023-10-18
Update reference tests to Cancun

shemnon:
Update reference tests to cancun tests.


shemnon:
Taking off of draft as this is now referencing a committed change.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ParentBeaconBlockRootHelper.java
```diff
@@ -49,8 +48,7 @@ static void storeParentBeaconBlockRoot(
     final UInt256 timestampIndex = UInt256.valueOf(timestampReduced);
     final UInt256 rootIndex = UInt256.valueOf(timestampExtended);
 
-    account.setStorageValue(
-        timestampIndex, UInt256.fromBytes(Bytes.of(Longs.toByteArray(timestamp))));
+    account.setStorageValue(timestampIndex, UInt256.fromBytes(Bytes.ofUnsignedLong(timestamp)));
```
macfarla, line 51: is there a conflict here with the change made by @jflo here https://github.com/hyperledger/besu/pull/6046/files#diff-61db834b59eae5ce5c438462505de1add8fa244deda830742060d15f668a9806

macfarla (commented):
one comment

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ParentBeaconBlockRootHelper.java
shemnon: This is the same change but with better APIs.  The #6046 change was to store values as unsigned longs.  The old way created multiple Bytes objects, which causes bad performance (there was a prior optimization addressing deeply wrapped bytes).  The use of [`Bytes.ofUnsignedLong`](https://github.com/tmio/tuweni/blob/55d33bb5a3930e5b5ba919c0e40e0551a1e71612/bytes/src/main/java/org/apache/tuweni/bytes/Bytes.java#L423-L458) skips the multiple object creations.

jflo: ha! awesome, i hated how I did that.

jflo (approved):
LGTM. TIL a needed API.

PR #6066, opened 2023-10-20
Update to production KZG ceremony output

jflo:
- update to new library


macfarla (commented):
looks ok but no idea how to verify some of the kzg files. Is there a way to verify? or is it a just trust me situation

jflo:
There is a reference test that covers them using commitments from Hive in ExecutionEngineCancunBlockBuildingAcceptanceTest.java

I'm working on updating the test vectors for the kzg precompile now.

evm/src/test/resources/kzg-trusted-setups/mainnet.txt
shemnon, line 1: Why is this file in the test directory, shouldn't it be in main?

jflo: as it stands today, the standalon evm only tries to init the KZGPrecompile contract by a network name from test code. Other usages it is passing in a file explicitly, so there was no need to bundle it in the binary.  

Happy to move it though.

shemnon: From the perspective of KZG as a cryptogrpahic primitive like ECDSA the mainnet data will be the most widespread "curve" in use.  I think the mainnet data should be included with the EVM as a default curve.

Are the testnets and shadownets not using the mainnet values?  Will tooling to use the proofs switch to other curves too?

shemnon: A standalone jar that only has the curve data may be an option if you don't want to wire it directly into the EVM jar.

jflo: I think it is quite unlikely that we use anything but the mainnet setup now that it is available, and having to use config flags to override it is reasonable.

I've moved it.

macfarla (approved):

Issue #6058, opened 2023-10-19
[FEATURE REQUEST] Set  DEFAULT_LOWEST_INVALID_KNOWN_NONCE_CACHE configurable, like others transaction pool options


tzdesing:
### Description
We are In a private free gas chain , using QBFT protocol, to implement a national CBDC, when a trasanction get stucked in TX_POOL, that transactions are included in cache. We are unable to send a new transaction with higher gas, so the only way is set our TX_POOL_MAX_SIZE to 0 , and back to default value, then we are able to sent new transactons, but , the invalid transaction was included in cache of validator and we are unable to reset then. because of network rules. So always we sent new transactions BESU will check this cache, and invalidate again because we have a higher nonce.

### Acceptance Criteria
IF we are able to set LOWEST_INVALID_KNOWN_NONCE_CACHE = 0 on bootstrap of validators nodes, we can transact normally and transactions will not being invalidated every time, because we haven't a cache, in our case this cache isn't necessary.

matthew1001:
I raised a discussion on discord about exactly this issue a couple of days ago: https://discord.com/channels/905194001349627914/905205502940696607/1164623046389272699

I was considering a PR that would prevent invalid TXs entering the invalid nonce cache if the reason was `NONCE_TOO_LOW` (which is the reason I've seen it happening with). The `NONCE_TOO_LOW` error seems to be the least sensible reason to set the invalid nonce cache to that value - because by definition, you couldn't submit a new TX at that nonce anyway.

I'll potentially raise that PR, but I think the new layered pool implementation doesn't use an invalid nonce cache. So you could set `--tx-pool=layered` and see if that prevents the invalid nonce cache from blocking up the pool.

matthew1001:
I've raised https://github.com/hyperledger/besu/pull/6067 which I think could help in a large number of the cases.

matthew1001:
Note that I think this issue which I raised yesterday is closely related to the scenario described:

https://github.com/hyperledger/besu/issues/6043

Currently even if you set `--tx-pool-price-bump=0` you cannot replace existing transactions. I think for QBFT, permissioned deployments it is very difficult to manage the pool through set its max size, or its retention time, when problematic transactions make their way into the pool.

alexcostars:
@matthew1001 as you mentioned, there is a new layered pool implementation. Has this new implementation already been released?

matthew1001:
[...]

PR #6067, opened 2023-10-20
Don't put NONCE_TOO_LOW transactions into the invalid nonce cache

matthew1001:
## PR description
This is intended to improve the issue described in https://github.com/hyperledger/besu/issues/6058. I don't think it fixes every possible case, but from testing it seems to be old transactions from a node, which are now all `NONCE_TOO_LOW`, that can cause this problem.

## Fixed Issue(s)
See https://github.com/hyperledger/besu/issues/6058 - although it may not be completely fixed after this PR.


fab-10 (changes requested):
`NONCE_TOO_LOW` means the transaction has already been confirmed, not sure in which scenario this happens, anyway I am fine excluding it from the invalid nonce cache, and I will suggest to have a list of invalid reason that are excluded from this cache, at the moment it will only contain `NONCE_TOO_LOW` but this way it will be easier to exclude also other invalid reason.

matthew1001:
Thanks for the feedback @fab-10 The latest commits hopefully address your comments.

fab-10 (approved):

PR #6072, opened 2023-10-23
Decouple TrieLogManager and CachedWorldStorageManager

siladu:
Two concepts were mixed together via inheritance:
1. trie log management
2. caching world states

This PR splits these concepts out ahead of adding in trie log pruning.

CachedWorldStorageManager is added as a field alongside TrieLogManager in BonsaiWorldStateProvider.

Following previous refactors, there's now no need to have an AbstractTrieLogManager or TrieLogManager interface


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/trielog/TrieLogManager.java
```diff
@@ -16,42 +16,188 @@
...
+import org.slf4j.LoggerFactory;
 
-  void saveTrieLog(
+public class TrieLogManager {
+  private static final Logger LOG = LoggerFactory.getLogger(TrieLogManager.class);
+  public static final long LOG_RANGE_LIMIT = 1000; // restrict trielog range queries to 1k logs
+  protected final Blockchain blockchain;
+  protected final BonsaiWorldStateKeyValueStorage rootWorldStateStorage;
+
+  protected final long maxLayersToLoad;
```
siladu, line 46: Can't decide if this is better to live here or in CachedWorldStateManager

matkt (approved):
Checked several times and seems to be good for me, simple PR without big refactoring

PR #6073, opened 2023-10-23
feat: add a way to read memory without altering the word capacity

delehef:
## PR description

Objects using the `MessageFrame` may wish to immutably read the memory of the frame without altering the word capacity of the memory, i.e. a “shadow” read as it behaves transparently w.r.t. the EVM.


evm/src/main/java/org/hyperledger/besu/evm/frame/Memory.java
```diff
@@ -192,6 +194,19 @@ int getActiveWords() {
...
+  /**
+   * Returns a copy of bytes from memory.
+   *
+   * @param location The location in memory to start with.
+   * @param numBytes The number of bytes to get.
+   * @param shadow If set, perform a “shadow read”, i.e. do not increase the active words.
+   * @return A fresh copy of the bytes from memory starting at {@code location} and extending {@code
+   *     numBytes}.
+   */
+  public Bytes getBytes(final long location, final long numBytes, final boolean shadow) {
```
shemnon, line 209: Rather than layering on a "shadow" flag that results two distinct forks of a function, I would rather see a `getBytesWithoutGrowth` that will zero pad without growing the memory size.

shemnon (changes requested):
Pleas make this two unrelated functions rather than a boolean switched function.

shemnon (changes requested):
Please ensure that context enter only fires once per context ever.  Test cases indicate otherwise.

evm/src/main/java/org/hyperledger/besu/evm/frame/Memory.java
delehef: Fair point, addressed in the latest patch.

macfarla (commented):
Maybe something went wrong with a merge? 91 files modified now?

evm/src/main/java/org/hyperledger/besu/evm/frame/Memory.java
```diff
@@ -179,7 +180,8 @@ int getActiveBytes() {
    *
    * @return The current number of active words stored in memory.
    */
-  int getActiveWords() {
+  @VisibleForTesting
+  public int getActiveWords() {
```
macfarla, line 184: don't think this needs to be public any more

macfarla (approved):
Looks ok to me but would be good to get @shemnon to have another pass

evm/src/main/java/org/hyperledger/besu/evm/frame/Memory.java
delehef: It actually needs; other tests break otherwise.

evm/src/main/java/org/hyperledger/besu/evm/frame/Memory.java
```diff
@@ -202,11 +204,40 @@ public Bytes getBytes(final long location, final long numBytes) {
...
+    }
+
+    final int start = asByteIndex(location);
+
+    // Arrays.copyOfRange would throw if start > memBytes.length, so just return the expected
+    // number of zeros without expanding the memory.
+    // Otherwise, just follow the happy path.
+    if (start > memBytes.length) {
+      return Bytes.wrap(new byte[(int) numBytes]);
+    } else {
```
shemnon, line 236: This doesn't handle a "straddle" case, where part of the bytes are in unexpanded memory and the rest needs to be zero expanded.  I would expect that a call with `n` as `numBytes` should allow a `get(n-1)` to work as well as `getSize()` on the return result retuning `n`

See https://github.com/hyperledger/besu/blob/98a06e882c68c5a24c55f7e67f14662b320cdb5f/evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java#L211-L222

evm/src/test/java/org/hyperledger/besu/evm/frame/MessageFrameTest.java
```diff
@@ -0,0 +1,78 @@
...
+    final MessageFrame messageFrame = messageFrameBuilder.build();
+
+    final Bytes value = Bytes.concatenate(WORD1, WORD2);
+    messageFrame.writeMemory(0, value.size(), value);
+    int initialActiveWords = messageFrame.memoryWordSize();
+
+    assertThat(messageFrame.shadowReadMemory(64, Bytes32.SIZE)).isEqualTo((Bytes32.ZERO));
+    assertThat(messageFrame.memoryWordSize()).isEqualTo(initialActiveWords);
+
+    assertThat(messageFrame.shadowReadMemory(32, Bytes32.SIZE)).isEqualTo((WORD2));
```
shemnon, line 75: Include a test for a "straddle" read.  Such as 0-255.  Verify size and a zero at some point after the end of the normal memory

shemnon (changes requested):
Pleas handle "straddle" reads that cover expanded and unexpanded memory.

evm/src/main/java/org/hyperledger/besu/evm/frame/Memory.java
delehef: According to [`Arrays.copyOfRange` doc](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#copyOfRange-byte:A-int-int-) handles this case:

> may be greater than original.length, in which case (byte)0 is placed in all elements of the copy whose index is greater than or equal to original.length - from.

evm/src/test/java/org/hyperledger/besu/evm/frame/MessageFrameTest.java
delehef: Done.

shemnon (approved):

PR #6074, opened 2023-10-23
Trigger contextEnter/Exit for all frames, including root

delehef:
## PR description

Filtering out the root context from the `context{Enter,Exit}` tracing hook leads to an asymmetry of treatment when returning from a create/call context, where the root context is blind to the event, but non-root contexts can read it.

This PR disables this restriction to address this blindspot.


evm/src/test/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessorTest.java
```diff
@@ -161,10 +141,12 @@ void shouldTraceContextEnterExitForEip3155Test() {
             POST_EXECUTION, // STATICCALL
             CONTEXT_ENTER, // STATICCALL
             CONTEXT_EXIT, // STATICCALL
+            CONTEXT_ENTER, // Re-entry in root context
```
shemnon, line 144: This shouldn't happen.  The number of enters should match the number of exits.

shemnon: Perhaps move the trace inside the `NOT_STARTED` block?

delehef: This is a re-entry in the caller context #0 after coming back from the callee context #1; I don't see what this is an issue?

delehef: Or maybe we should have different hooks, _e.g._ `traceContextEnter` and `traceContextReEnter`, one when in the `NOT_STARTED` state, and the other one elsewhere?

delehef: In this case, any context would have one enter, one exit, and as many re-entry as child frames.

shemnon (changes requested):
Please ensure that context enter only fires once per context ever. Test cases indicate otherwise.

evm/src/test/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessorTest.java
shemnon: enter and re-enter have vastly different semantics.  enter should be idempotent.  re-enter can re-occur.

delehef:
New patch differentiates between context entry and re-entry; tests have been adapted.

evm/src/test/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessorTest.java
```diff
@@ -41,59 +41,35 @@
...
+    when(messageFrame.getWorldUpdater()).thenReturn(new ToyWorld());
     when(messageFrame.getState()).thenReturn(MessageFrame.State.COMPLETED_SUCCESS);
     when(messageFrame.getMessageFrameStack()).thenReturn(messageFrameStack);
-    when(messageFrame.getWorldUpdater()).thenReturn(worldUpdater);
 
     getAbstractMessageProcessor().process(messageFrame, operationTracer);
 
-    verify(operationTracer, times(1)).traceContextEnter(messageFrame);
-    verify(operationTracer, times(1)).traceContextExit(messageFrame);
+    verify(operationTracer).traceContextExit(messageFrame);
```
macfarla, line 63:
```suggestion
    verify(operationTracer, times(1)).traceContextEnter(messageFrame);
    verify(operationTracer, times(1)).traceContextExit(messageFrame);
```

evm/src/test/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessorTest.java
```diff
@@ -41,59 +41,35 @@
...
-  void shouldTraceContextIfStackSizeIsGreaterZeroAndFailure(final int stackSize) {
-    when(messageFrame.getMessageStackSize()).thenReturn(stackSize);
+  @Test
+  void shouldTraceExitEvenIfContextFailed() {
     when(messageFrame.getState()).thenReturn(MessageFrame.State.COMPLETED_FAILED);
     when(messageFrame.getMessageFrameStack()).thenReturn(messageFrameStack);
 
     getAbstractMessageProcessor().process(messageFrame, operationTracer);
 
-    verify(operationTracer, times(1)).traceContextEnter(messageFrame);
```
macfarla, line 96: shouldn't traceContextEnter still be called?

delehef: No, because the mocker makes the frame states directly jump to `COMPLETED_FAILED`.

Testing `traceContextEnter` here would required to mock enough of the `MessageFrame` to run the whole `process` on it.

macfarla: I think add a comment to the test to explain why contextEnter is not expected

evm/src/test/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessorTest.java
```diff
@@ -41,59 +41,35 @@
...
+  @Test
+  void shouldTraceExitForContext() {
+    when(messageFrame.getWorldUpdater()).thenReturn(new ToyWorld());
     when(messageFrame.getState()).thenReturn(MessageFrame.State.COMPLETED_SUCCESS);
     when(messageFrame.getMessageFrameStack()).thenReturn(messageFrameStack);
-    when(messageFrame.getWorldUpdater()).thenReturn(worldUpdater);
 
     getAbstractMessageProcessor().process(messageFrame, operationTracer);
 
-    verify(operationTracer, times(1)).traceContextEnter(messageFrame);
```
macfarla, line 83: same here - if there's a legit reason traceContextEnter is not expected, add a comment because it's not obvious

macfarla (approved):
Looks ok to me but would be good to get @shemnon to have another pass

shemnon (approved):

PR #6075, opened 2023-10-24
Continuous Trie Log Pruning

siladu:
- Feature toggled by `--Xtrie-log-pruning-enabled`
- Add trie log pruning after a successful `TrieLogManager.saveTrieLog`
- Each time a trie log is persisted, the current trie log is cached and the pruner is run against the oldest entries in the cache.
- This makes no attempt to manage the backlog of old trie logs, it will only prune what has been added to the cache, i.e. trie logs that have been added since the feature was enabled.
- Pruner limit exists in case of exceptional circumstances, but we should only ever be pruning all the forks for a single block number during each prune execution.

This is a second take of https://github.com/hyperledger/besu/pull/6026
Currently built on https://github.com/hyperledger/besu/pull/6072 (diff: https://github.com/siladu/besu/compare/refactor-trie-log-manager...trie-log-pruning-take2)

One downside of this PR compared to #6026 is that restarting besu may leave a gap in the pruned trie logs because we don't preload the cache. This will have to be considered as part of managing the backlog. It may mean that we can't simply rely on a "one-off" resync or prune subcommand. However, these un-pruneable trie logs should be negligible compared to the saving gained for long running nodes.


siladu:
Abandoning this approach in favour of an adapted form of the original https://github.com/hyperledger/besu/pull/6026

PR #6082, opened 2023-10-25
Add option to clique to skip creating empty blocks

jframe:
## PR description
Add an option to clique to not create empty blocks. To skip creating empty blocks the config `createemptyblocks` can be false in the clique configuration of the genesis file. By default, Clique will create empty blocks. This is the current behaviour in Clique.

This is useful for large private networks that are using Clique. As empty blocks increase the amount of storage needed.

note: That all validators must have the same setting for `createemptyblocks`.

Changes
* Clique miner has been changed to not create empty blocks when `createemptyblocks` is disabled
* Header validation for Clique checks that there are no empty blocks when `createemptyblocks` is disabled

To use this feature add `createemptyblocks` to the genesis config e.g.

```
{
  "config": {
    "londonBlock": 0,
    "clique": {
      "blockperiodseconds": 10,
      "epochlength": 30000,
      "createemptyblocks": false
    }
  },
...
}
````

### CODE REVIEW

File: besu/src/main/java/org/hyperledger/besu/controller/CliqueBesuControllerBuilder.java [hunk 1]
```diff
@@ -53,6 +53,7 @@ public class CliqueBesuControllerBuilder extends BesuControllerBuilder {
   private Address localAddress;
   private EpochManager epochManager;
   private long secondsBetweenBlocks;
+  private boolean createEmptyBlocks;
```
siladu, line 56: should this either be final or default to true?

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/blockcreation/CliqueBlockMiner.java
```diff
@@ -20,16 +20,23 @@
...
 
 import java.util.function.Function;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /** The Clique block miner. */
 public class CliqueBlockMiner extends BlockMiner<CliqueBlockCreator> {
+  private static final Logger LOG = LoggerFactory.getLogger(CliqueBlockMiner.class);
+  private static final int WAIT_BETWEEN_EMPTY_BUILD_ATTEMPTS = 1_000;
```
siladu, line 36: nit: specifying the unit would be useful - assume it's millis?

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/blockcreation/CliqueBlockMiner.java
```diff
@@ -63,4 +73,18 @@ protected boolean mineBlock() throws InterruptedException {
...
+
+  @Override
+  protected boolean shouldImportBlock(final Block block) throws InterruptedException {
+    if (createEmptyBlocks) {
+      return true;
+    }
+
+    final boolean isEmpty = block.getBody().getTransactions().isEmpty();
+    if (isEmpty) {
+      LOG.debug("Skipping empty block");
```
siladu, line 85: worth logging block details with `toLogString`?

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/headervalidationrules/CliqueNoEmptyBlockValidationRule.java
```diff
@@ -0,0 +1,47 @@
...
+   * @param parent the block header corresponding to the parent of the header being validated.
+   * @return true if the transactionsRoot in the header is not the empty trie hash.
+   */
+  @Override
+  public boolean validate(final BlockHeader header, final BlockHeader parent) {
+    final boolean hasTransactions = !header.getTransactionsRoot().equals(Hash.EMPTY_TRIE_HASH);
+    if (!hasTransactions) {
+      LOG.info(
+          "Invalid block header: {} has no transactions but create empty blocks is not enabled",
+          header.getNumber());
```
siladu, line 43: `toLogString` might be more helpful here?

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/BlockMinerTest.java
```diff
@@ -132,6 +134,54 @@ public void failureToImportDoesNotTriggerObservers() throws InterruptedException
     verify(observer, times(1)).blockMined(blockToCreate);
   }
 
+  @Test
+  public void preImportValidationFailureDoesNotImportBlock() throws InterruptedException {
```
siladu, line 138: What is "pre import"?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Accept `input` and `data` field for the payload of transaction-related RPC methods [#6094](https://github.com/hyperledger/besu/pull/6094)
 - Add APIs to set and get the min gas price a transaction must pay for being selected during block creation [#6097](https://github.com/hyperledger/besu/pull/6097)
 - TraceService: return results for transactions in block [#6086](https://github.com/hyperledger/besu/pull/6086)
+- Clique option to not create empty blocks [#6082]
```
siladu, line 15: Worth calling out the genesis option name explicitly here: `createemptyblocks`?

siladu (approved):
LGTM, few questions and nits about logging/naming.

Do we want to add this to the docs as well?

jframe:
> LGTM, few questions and nits about logging/naming.
> 
> Do we want to add this to the docs as well?

Yes I think should add it to the docs, will add the label

siladu (approved):
Updates look good, thanks 👍

gfukushima (approved):

fab-10:
Found that this option prevent syncing a network that had some empty blocks at the beginning, that scenario could happen if a network was created without this option (or with it set to true) and then `createemptyblocks` is set to false.
It would be nice if this option only applies on block creation, not during the sync.

jframe:
> Found that this option prevent syncing a network that had some empty blocks at the beginning, that scenario could happen if a network was created without this option (or with it set to true) and then `createemptyblocks` is set to false. It would be nice if this option only applies on block creation, not during the sync.

Doing the header validation to ensure that even when we are not the proposer of the block we don't accept empty blocks when the `createemptyblocks` is false. If we just do validation on the block creation then we can't guarantee there are no empty blocks in this case.

non-fungible-nelson:
@jframe what does an IBFT2 and QBFT adaptation of this PR look like?

PR #6088, opened 2023-10-25
Fix k8s nat manager logic and add `--Xnat-kube-service-namespace` flag

### LINKED ISSUE #5002: Besu always needs cluster level permissions for Kubernetes NAT manager on GKE
### Description

- Currently the Kubernetes NAT manager checks all namespaces to find the besu service here: https://github.com/hyperledger/besu/blob/9eb32836b710bd997887c5c1f77436710be9eff1/nat/src/main/java/org/hyperledger/besu/nat/kubernetes/KubernetesNatManager.java#L87
- For this to work, we need to give cluster level permissions to the `besu` deployment for it to be able to scan all the namespaces
- In some environments, granting cluster level permissions is not authorised (for example, in a lot of banks)
- Even if the besu service is in the same namespace as the besu deployment the NAT manager errors out when we do not grant cluster level permissions since it ALWAYS lists services in all namespaces.
- Falls back to `NONE`, the enode does not have the correct public IP address anymore, instead it is `0.0.0.0`

What we need is the NAT manager to work without the need of cluster level permissions on GKE.

Perhaps a fallback to just checking in the local namespace (i.e the namespace of the besu deployment) before throwing the error?

### Acceptance Criteria
* NAT manager works without cluster level permissions on GKE

### Steps to Reproduce (Bug)
1. Deploy a besu deployment, service without cluster level permissions
2. The NAT manager errors out, throws the following error:

https://github.com/hyperledger/besu/blob/9eb32836b710bd997887c5c1f77436710be9eff1/nat/src/main/java/org/hyperledger/besu/nat/kubernetes/KubernetesNatManager.java#L97

and falls back to `NONE`.
3. The enode url of the node has `0.0.0.0` as the public IP

If searching for the service across all namespaces fails, it should look for the service in the local namespace before falling back to `NONE`.

If searching for the service across all namespaces fails, it errors out, and falls back to `NONE`.

100%

* docker image we use: https://hub.docker.com/layers/hyperledger/besu/22.10.4/images/sha256-036331dda5c27e2d51524614cd628d6f55fc14583b48516e1e019aef0de92a21?context=explore
* Software version: [`besu --version`]
* Java version: [`java -version`]
* OS Name & Version: [`cat /etc/*release`]
* Kernel Version: [`uname -a`]
* Virtual Machine software & version: [`vmware -v`]
* Docker Version: [`docker version`]
* Cloud VM, type, size: [Amazon Web Services I3-large]

### Smart contract information (If you're reporting an issue arising from deploying or calling a smart contract, please supply related information)
N/A

N/A

alex123012:
## PR description
Now the Kubernetes nat manager requires cluster-wide permissions (list service resources in all namespaces). Also, it has a bug: when multiple besu networks are configured in the same cluster with the same service names in different namespaces, the nat manager may set an inappropriate IP address for the besu node due to enumerating services from all namespaces and checking only their names.

I suggest adding a new flag: `--Xnat-kube-service-namespace` to specify a concrete besu node service namespace and not use a list of services, but a get request for a specific service.

## Fixed Issue(s)

fixes #5002

### DIFF
--- a/nat/src/main/java/org/hyperledger/besu/nat/kubernetes/KubernetesNatManager.java
+++ b/nat/src/main/java/org/hyperledger/besu/nat/kubernetes/KubernetesNatManager.java
@@ -49,21 +49,26 @@ import org.slf4j.LoggerFactory;
 public class KubernetesNatManager extends AbstractNatManager {
   private static final Logger LOG = LoggerFactory.getLogger(KubernetesNatManager.class);
 
-  /** The constant DEFAULT_BESU_SERVICE_NAME_FILTER. */
-  public static final String DEFAULT_BESU_SERVICE_NAME_FILTER = "besu";
+  /** The constant DEFAULT_BESU_SERVICE_NAME. */
+  public static final String DEFAULT_BESU_SERVICE_NAME = "besu";
+
+  public static final String DEFAULT_BESU_SERVICE_NAMESPACE = "besu";
 
   private String internalAdvertisedHost;
-  private final String besuServiceNameFilter;
+  private final String besuServiceName;
+  private final String besuServiceNamespace;
   private final List<NatPortMapping> forwardedPorts = new ArrayList<>();
 
   /**
    * Instantiates a new Kubernetes nat manager.
    *
-   * @param besuServiceNameFilter the besu service name filter
+   * @param besuServiceName the besu service name
+   * @param besuServiceNamespace the besu service namespace
    */
-  public KubernetesNatManager(final String besuServiceNameFilter) {
+  public KubernetesNatManager(final String besuServiceName, final String besuServiceNamespace) {
     super(NatMethod.KUBERNETES);
-    this.besuServiceNameFilter = besuServiceNameFilter;
+    this.besuServiceName = besuServiceName;
+    this.besuServiceNamespace = besuServiceNamespace;
   }
 
   @Override
@@ -83,15 +88,7 @@ public class KubernetesNatManager extends AbstractNatManager {
       final CoreV1Api api = new CoreV1Api();
       // invokes the CoreV1Api client
       final V1Service service =
-          api
-              .listServiceForAllNamespaces(
-                  null, null, null, null, null, null, null, null, null, null)
-              .getItems()
-              .stream()
-              .filter(
-                  v1Service -> v1Service.getMetadata().getName().contains(besuServiceNameFilter))
-              .findFirst()
-              .orElseThrow(() -> new NatInitializationException("Service not found"));
+          api.readNamespacedService(besuServiceName, besuServiceNamespace, null);
       updateUsingBesuService(service);
     } catch (Exception e) {
       throw new NatInitializationException(e.getMessage(), e);
--- a/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
@@ -169,6 +169,7 @@ public class RunnerBuilder {
   private int p2pListenPort;
   private NatMethod natMethod = NatMethod.AUTO;
   private String natManagerServiceName;
+  private String natManagerServiceNamespace;
   private boolean natMethodFallbackEnabled;
   private EthNetworkConfig ethNetworkConfig;
   private EthstatsOptions ethstatsOptions;
@@ -338,6 +339,17 @@ public class RunnerBuilder {
    */
   public RunnerBuilder natManagerServiceName(final String natManagerServiceName) {
     this.natManagerServiceName = natManagerServiceName;
+    return this;
+  }
+
+  /**
+   * Add Nat manager service namespace.
+   *
+   * @param natManagerServiceNamespace the nat manager service namespace
+   * @return the runner builder
+   */
+  public RunnerBuilder natManagerServiceNamespace(final String natManagerServiceNamespace) {
+    this.natManagerServiceNamespace = natManagerServiceNamespace;
     return this;
   }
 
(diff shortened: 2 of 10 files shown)


macfarla (commented):
minor suggestions on help messages

macfarla:
@alex123012 there's a couple of unit tests failing. you should be able to reproduce locally

> Task :besu:test

BesuCommandTest > natManagerServiceNameCannotBeUsedWithNatDockerMethod FAILED
    java.lang.AssertionError at BesuCommandTest.java:1971

BesuCommandTest > natManagerServiceNameCannotBeUsedWithNatNoneMethod FAILED
    java.lang.AssertionError at BesuCommandTest.java:1981

macfarla:
code looks ok - @alex123012 how would I verify that it's working? new to kubernetes

alex123012:
So, you could try to use [kind](https://kind.sigs.k8s.io/)/[minikube](https://minikube.sigs.k8s.io/docs/start/) with https://github.com/Consensys/quorum-kubernetes repo with changed image. I've tested it with [ibft2](https://github.com/Consensys/quorum-kubernetes/blob/master/playground/kubectl/quorum-besu/ibft2).

Something like a plan:
- install minikube and [kubectl](https://kubernetes.io/docs/tasks/tools/)
- build the besu image and load it to the minikube cluster (for example, with [cache](https://minikube.sigs.k8s.io/docs/handbook/pushing/#2-push-images-using-cache-command) or other option in docs).
- change images of besu in k8s manifests to image you've build (for example, [here](https://github.com/Consensys/quorum-kubernetes/blob/master/playground/kubectl/quorum-besu/ibft2/statefulsets/validator4-statefulset.yaml#L71), but you'll need to change it in all files like in the example)
- go through quorum-kubernetes docs (minikube cluster is created in it)

If you encounter any problems, you can write to me in [telegram](https://t.me/alex_makh) or email (makhonin.a.ru@gmail.com)

alex123012:
@macfarla Wrote a few notes^

alex123012:
@macfarla Hi!
Any updates here?:)

nat/src/main/java/org/hyperledger/besu/nat/kubernetes/KubernetesNatManager.java
```diff
@@ -83,15 +88,7 @@ protected void doStart() throws NatInitializationException {
...
-          api
-              .listServiceForAllNamespaces(
-                  null, null, null, null, null, null, null, null, null, null)
-              .getItems()
-              .stream()
-              .filter(
-                  v1Service -> v1Service.getMetadata().getName().contains(besuServiceNameFilter))
-              .findFirst()
-              .orElseThrow(() -> new NatInitializationException("Service not found"));
+          api.readNamespacedService(besuServiceName, besuServiceNamespace, null);
```
garyschulte, line 91: I think service namespace should be an Optional<String> parameter to the manager.  Not having to explicitly specify the namespace is a useful feature IMO, and can prevent breaking existing clustered implementations.  

The presence or absence of the namespace can gate the behavior of finding first vs explicitly specify the target namespace

garyschulte (commented):
👍 , but we should make namespace optional in order to be backward compatible with prior behavior.

nat/src/main/java/org/hyperledger/besu/nat/kubernetes/KubernetesNatManager.java
garyschulte: @alex123012 any word on this?

macfarla:
@cdivitotawela any input on this PR?

cdivitotawela:
> @cdivitotawela any input on this PR?

Let me check whether I can run this on minikube and verify,

lancemarks:
Testing successful with Kubernetes in AWS. There was one issue I ran into, I had to name the port 'discovery' or it would throw an error. I would have liked to been able to name it 'discovery-tcp' or 'p2p-tcp'.

matthew1001:
Just wondering after talking to some others about K8S behaviour. Does the namespace need manually specifying via a config option, or can the K8S namespace of the Besu pod just be queried and used in discovering the service endpoint? A K8S service can't serve a pod in a different namespace anyway, so it's not clear if there's a reason to add the new `--Xnat-kube-service-namespace` instead of auto-discovering the pod's namespace.

onelapahead:
Hello! Building off of @lancemarks @matthew1001 's feedback I've opened a PR against our fork's patch for the time being: https://github.com/kaleido-io/besu/pull/11

I'm testing it on KinD and in AWS EKS these next few days, and will get unit tests done in the next day. But wondering what @alex123012 @cdivitotawela @macfarla think of that as an enhancement to @alex123012's work ?

To summarize:
* We should assume Besu only needs to access Services in the same Namespace as itself based on reasons stated in the PR, it simplifies the UX overall and avoids bugs like what Alex was fixing.
* We can avoid introducing another experimental CLI arg based on the above assumption.
* Users can achieve least privileges in RBAC thanks to @alex123012 since we are limiting the ServiceAccount's access to the Namespace Besu runs within.

If @alex123012 that seems alright to you, you could cherry pick my commits into this PR ?

macfarla:
@alex123012 to simplify besu codebase we are planning to deprecate k8s NAT method, so we are reluctant to make changes to it at this stage.

Suggestion from our devops team - 
I wound up just keeping it simple and using an external service to return the IP https://github.com/Consensys/ethereum-helm-charts/blob/main/charts/ethereum-helm-charts/charts/elc/templates/besu/statefulset.yml#L66 We bind directly to the host node's IP and steps to get things setup on the infra side are here https://besu.hyperledger.org/public-networks/tutorials/kubernetes

This still keeps the RBAC permissions intact and doesn't require admin rights which I prefer. Control goes back to the owner and they decide how/what is allowed

macfarla:
@alex123012  are you still engaged on this PR? it's marked as stale and will be closed in 2 weeks if no activity

macfarla:
Hi @alex123012  

We have made the decision to deprecate k8s with nat manager. So we don't want to add to this section of the code base. We've got solutions in place that bypass nat manager for k8s and thats working for private, and public nodes that we run. Our preference is to keep it simple and clean for the long term and not more band aid patches for another service mechanism or cloud provider. 

We recognize that you have gone through a lot of trial and error to create this PR but that's kinda on us for leaving this as is and not making the call earlier - and is also a symptom of the issue that we besu developers aren't experts in this part of the code base, it's not our core expertise.

If you want to keep this functionality and extend, you are welcome to fork and maintain - or consider converting the functionality to a plugin long-term?

I'm closing this PR, hope you understand the maintainers point of view here.

PR #6089, opened 2023-10-25
Introduce a timeout for the evaluation of a single tx during block creation

fab-10:
## PR description

This PR is built on top of #6044, so please check it first. [Diff between PRs ](https://github.com/fab-10/besu/compare/time-limited-block-creation...time-limited-tx-processing-block-creation)

This PR is a continuation of #6044, and introduce the possibility to configure a max amount of time the evaluation of a single tx could take during block creation, this to avoid that a single slow processing tx consumes all the time dedicated to the block creation.
If a tx evaluation hit the timeout, the tx is removed from the txpool, to avoid that it will tried again on a future block creation.

This timeout is configurable using the new experimental option:
- `--Xtxs-selection-per-tx-max-time` in milliseconds

that by default is set to the same value of `--Xtxs-selection-max-time`, so by default a single tx is allowed to use all the time reserved for block creation, but could be set to a lower value.

In case of a tx evaluation timeout a warning is logged:
```
Interrupting transaction evaluation since it is taking more than the max configured time of 10ms. Tx with hash: 0xab34d5061d5daee4a5f58f3f9ff76935963aff3ed8990474d1d7e140d6d41787 will be removed from the pool
```


fab-10:
Closing since implementing this require making tx processing thread safe, implemented a lighter version of this in #6163

PR #6090, opened 2023-10-26
[MINOR] Remove a BonsaiWorldStateProvider constructor

siladu:
Was only used for supporting test code and can instead reuse static factory from InMemoryKeyValueStorageProvider


Gabriel-Trintinalia (approved):
LGTM

PR #6094, opened 2023-10-26
RPC Parameters to accept input and data field

### LINKED ISSUE #6055: Accept `input` as well as `data` in eth_call
Context on #6041

We currently only accept `data` as the name of the payload field when during an eth_call (and presumably other transaction-related RPCs) https://github.com/hyperledger/besu/blob/a90ea056f19915a4eca787fc066c080e1789c14c/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/JsonCallParameter.java#L52-L53

The spec requires `input` as the name instead:
https://github.com/ethereum/execution-apis/blob/057892b256c181232ca22499d1801b9637192698/src/schemas/transaction.yaml#L33

We will need to continue accepting `data` to avoid forcing a lot of tooling to update.

We should add support for `input` xor `data`

gfukushima:
## PR description
This PR changes the RPC to accept `input` as well as `data`.

## Fixed Issue(s)

Fixes #6055 
Also might get rid of some issues that we had in the Hive RPC suite.


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/JsonCallParameter.java
```diff
@@ -40,6 +40,31 @@ public class JsonCallParameter extends CallParameter {
 
   private final Optional<Boolean> strict;
 
+  public JsonCallParameter(
```
shemnon, line 43: Is this used anywhere outside of tests?  Can we just rip the band-aid off and migrate to the new signaure?  IntelliJ has a tool for this Refactor>Change Signature... (command/alt F6)

shemnon (approved):
Changing signature is desired, but not required.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/JsonCallParameter.java
gfukushima: Only tests, changing class to only use a single constructor. thanks for the feedback.

gfukushima:
Docs will need to update this page https://besu.hyperledger.org/public-networks/reference/api/objects#transaction-call-object

PR #6097, opened 2023-10-26
Add API to set and get the minGasPrice at runtime

fab-10:
## PR description

Adding `miner_getMinGasPrice` and `miner_setMinGasPrice` to manipulate the [min-gas-price](https://besu.hyperledger.org/public-networks/reference/cli/options#min-gas-price) option at runtime.


Gabriel-Trintinalia:
@fab-10 see https://github.com/hyperledger/besu/pull/6099

fab-10:
> @fab-10 see #6099

Done

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/PriceTransactionSelector.java
```diff
@@ -71,30 +71,27 @@ public TransactionSelectionResult evaluateTransactionPostProcessing(
    */
   private boolean transactionCurrentPriceBelowMin(final PendingTransaction pendingTransaction) {
     final Transaction transaction = pendingTransaction.getTransaction();
-    // Here we only care about EIP1159 since for Frontier and local transactions the checks
-    // that we do when accepting them in the pool are enough
-    if (transaction.getType().supports1559FeeMarket() && !pendingTransaction.hasPriority()) {
```
gfukushima, line 76: Just double checking, did you mean to drop the first condition of this if?

gfukushima (approved):
lgtm

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/PriceTransactionSelector.java
fab-10: Yes, because now the `minGasPrice` could change at runtime, so the check done in the txpool is no more enough, and we need to redo it here for every tx

PR #6098, opened 2023-10-26
New option tx-pool-min-gas-price to set a lower bound when accepting txs to the pool

fab-10:
## PR description

This PR, is related to #6097, and decouples the min gas price (existing `min-gas-price` option) used when selecting txs during block production, from the min gas price (new `tx-pool-min-gas-price` option) that is required when txs are added to the txpool, since as consequence of #6097, the former min gas price (`min-gas-price`) can be changed at runtime and using it for evaluation incoming txs in the txpool, could result in strange behavior, where we refuse txs just because their price is just less than the `min-gas-price`, but since the latter can be later lowered, the same txs could have been included in a block just after a while, for this we want to accept them in the pool.

Thus this new option (`tx-pool-min-gas-price`) is a static value to set only if you are sure that you never want to process txs with a gas price lower than it, so to exclude what you consider _spam_ txs from polluting your txpool. 
For that reason `tx-pool-min-gas-price` can only be less than equal to the start value of `min-gas-price`.


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2825,6 +2825,14 @@ private TransactionPoolConfiguration buildTransactionPoolConfiguration() {
       txPoolConfBuilder.priceBump(Percentage.ZERO);
     }
 
+    if (getMiningParameters().getMinTransactionGasPrice().lessThan(txPoolConf.getMinGasPrice())) {
+      logger.warn(
+          "Forcing tx-pool-min-gas-price="
```
macfarla, line 2830: is this preferable to flagging it as a misconfiguration?

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -122,6 +123,15 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
       arity = "1..*")
   private Set<Address> prioritySenders = TransactionPoolConfiguration.DEFAULT_PRIORITY_SENDERS;
 
+  @CommandLine.Option(
+      names = {TX_POOL_MIN_GAS_PRICE},
+      paramLabel = "<Wei>",
+      description =
+          "Minimum price (in Wei) offered by a transaction to be added to the txpool (not to be confused with min-gas-price,"
+              + " that is applied on block creation) (default: ${DEFAULT-VALUE})",
```
macfarla, line 131: Transactions with gas price lower than this minimum will not be accepted into the txpool

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
fab-10: changed to adjust the value when only `min-gas-price` is set, and return and error when both are set, but `tx-pool-min-gas-price` > `min-gas-price`

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
fab-10: updated the description

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
fab-10: the updated code wants to be backward compatible, and to simplify the upgrade path, when users just want to set `min-gas-price=0`, without having to worry about the new option.

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -340,4 +352,14 @@ public List<String> getCLIOptions() {
   public boolean isPriceBumpSet(final CommandLine commandLine) {
     return CommandLineUtils.isOptionSet(commandLine, TransactionPoolOptions.TX_POOL_PRICE_BUMP);
   }
+
+  /**
+   * Is min gas price option set?
+   *
+   * @param commandLine the command line
+   * @return true is tx-pool-min-gas-price is set
```
macfarla, line 360:
```suggestion
   * @return true if tx-pool-min-gas-price is set
```

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2825,6 +2825,24 @@ private TransactionPoolConfiguration buildTransactionPoolConfiguration() {
       txPoolConfBuilder.priceBump(Percentage.ZERO);
     }
 
+    if (getMiningParameters().getMinTransactionGasPrice().lessThan(txPoolConf.getMinGasPrice())) {
+      if (transactionPoolOptions.isMinGasPriceSet(commandLine)) {
+        throw new ParameterException(
+            commandLine,
+            "tx-pool-min-gas-price could not be greater than the value of min-gas-price");
```
macfarla, line 2832:
```suggestion
            "tx-pool-min-gas-price cannot be greater than the value of min-gas-price");
```

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2825,6 +2825,24 @@ private TransactionPoolConfiguration buildTransactionPoolConfiguration() {
...
+
+      } else {
+        // for backward compatibility, if tx-pool-min-gas-price is not set, we adjust its value
+        // to be the same as min-gas-price, so the behavior is as before this change, and we notify
+        // the user of the change
+        logger.warn(
+            "Forcing tx-pool-min-gas-price="
+                + getMiningParameters().getMinTransactionGasPrice().toDecimalString()
+                + ", since it could not be greater than the value of min-gas-price");
```
macfarla, line 2841:
```suggestion
                + ", since it cannot be greater than the value of min-gas-price");
```

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -5248,6 +5248,31 @@ public void txpoolPriceBumpKeepItsValueIfSetEvenWhenMinGasPriceZero() {
...
+
+    final Wei txPoolMinGasPrice = transactionPoolConfigCaptor.getValue().getMinGasPrice();
+    assertThat(txPoolMinGasPrice).isEqualTo(Wei.ZERO);
+
+    assertThat(commandOutput.toString(UTF_8)).isEmpty();
+    assertThat(commandErrorOutput.toString(UTF_8)).isEmpty();
+    verify(mockLogger, atLeast(1))
+        .warn(
+            contains(
+                "Forcing tx-pool-min-gas-price=0, since it could not be greater than the value of min-gas-price"));
```
macfarla, line 5265:
```suggestion
                "Forcing tx-pool-min-gas-price=0, since it cannot be greater than the value of min-gas-price"));
```

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -5248,6 +5248,31 @@ public void txpoolPriceBumpKeepItsValueIfSetEvenWhenMinGasPriceZero() {
...
+  }
+
+  @Test
+  public void txpoolTxPoolMinGasPriceMustNotBeGreaterThanMinGasPriceZero() {
+    parseCommand("--min-gas-price", "100", "--tx-pool-min-gas-price", "101");
+    assertThat(commandOutput.toString(UTF_8)).isEmpty();
+    assertThat(commandErrorOutput.toString(UTF_8))
+        .contains("tx-pool-min-gas-price could not be greater than the value of min-gas-price");
```
macfarla, line 5273:
```suggestion
        .contains("tx-pool-min-gas-price cannot be greater than the value of min-gas-price");
```

macfarla (approved):
suggestion re wording, otherwise LGTM

PR #6080, opened 2023-10-24
Implement miner_setMinPriorityFee and miner_getMinPriorityFee

Gabriel-Trintinalia:
## PR description
Built on top of https://github.com/fab-10/besu/tree/min-priority-fee, which adds `--min-priority-fee` CLI option

Adds two new JsonRPC methods: 

- miner_setMinPriorityFee - Sets the value of `miningParameters.minPriorityFeePerGas`
- miner_getMinPriorityFee - Returns the current value of `miningParameters.minPriorityFeePerGas`

This PR does not add the transactionSelector based on `minPriorityFeePerGas`. It will be added in a different PR

PR #6099, opened 2023-10-27
setMinPriorityFee  - Return Invalid Param when invalid and use hexadecimal instead of Long

Gabriel-Trintinalia:
## PR description

Implements the suggestions of the PR  https://github.com/hyperledger/besu/pull/6080:

- Returns an Json Invalid Param when param is invalid
- Use Hexadecimal values as parameters

### Valid request:
```
Set Request:
{
   "jsonrpc":"2.0",
   "method":"miner_setMinPriorityFee",
   "params":[
      "0x0a"
   ],
   "id":1
}
Response:
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": true
}

Get Request:
{
   "jsonrpc":"2.0",
   "method":"miner_getMinPriorityFee",
   "params":[],
   "id":1
}
Response:
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0xa"
}

```

### Invalid Request - missing parameter
```
Request:
{
   "jsonrpc":"2.0",
   "method":"miner_setMinPriorityFee",
   "params":[],
   "id":1
}

Response:

{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32602,
        "message": "Invalid params",
        "data": "Missing required json rpc parameter at index 0"
    }
}
```
### Invalid Request - Invalid Parameter
```
Request
{
   "jsonrpc":"2.0",
   "method":"miner_setMinPriorityFee",
   "params":[
      "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
   ],
   "id":1
}

Response:

{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32602,
        "message": "Invalid params",
        "data": "Hex value is too large: expected at most 32 bytes but got 33"
    }
}
```

Log example:
```
2023-10-27 11:14:59.719+11:00 | vert.x-worker-thread-0 | DEBUG | JsonRpcExecutor | JSON-RPC request -> miner_setMinPriorityFee ["0x000000000000000000000000000000000000000000000000000000000000000a"]
2023-10-27 11:14:59.720+11:00 | vert.x-worker-thread-0 | DEBUG | MinerSetMinPriorityFee | min priority fee per gas changed to 10 wei
```


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/miner/MinerGetMinPriorityFee.java
```diff
@@ -36,6 +36,7 @@ public String getName() {
   @Override
   public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
     return new JsonRpcSuccessResponse(
-        requestContext.getRequest().getId(), miningParameters.getMinPriorityFeePerGas().getValue());
+        requestContext.getRequest().getId(),
+        miningParameters.getMinPriorityFeePerGas().toShortHexString());
```
fab-10, line 40: I see that other APIs (like `eth_getGasPrice`) use `Quantity:create` to format the values returned, so you may consider to use that for consistency

Gabriel-Trintinalia: done! Thanks!

fab-10 (approved):

PR #6101, opened 2023-10-27
get all account read into the cache

matkt:
## PR description

There was a slight problem on the bonsai side because all account reads did not go through a single method. One of the two add the account to the cache but the other did not. This had two consequences:

Less good performance because certain accounts had to be read several times and also all account reads were not marked in the trielog. This will fix both problems.

The difference now is that we are not tracking the accounts that we read via this method, and this allows us not to falsely consider them as touched and therefore not to delete them when they should not be.

I validated my implementation thanks to this reference tests

Name: codeCopyZero-Berlin
Name: codeCopyZero-Cancun
Name: codeCopyZero-Istanbul
Name: codeCopyZero-London
Name: codeCopyZero-Merge
Name: codeCopyZero-Shanghai

I will also start a full Goerli, full ETH sync (not complete, I aim to reach the problematic block 2,675,000) and run a snapsync node during several days.


matkt:
all sync passed and we also passed the problematic block on mainnet fullsync

matkt:
@matkt do you want to check before merging or it's fine for you ?

garyschulte (approved):
LGTM.  It would be helpful to get confirmation from @diega that this is not an issue for ETC also, since he was the one who raised the initial issue that caused us to revert this behavior.

diega:
I just passed the target block and everything works smooth now

diega (approved):
LGTM

matkt:
thanks for your feedback @diega

PR #6106, opened 2023-10-30
Reverse sort order

matthew1001:


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/GasPricePendingTransactionsSorter.java
```diff
@@ -39,8 +40,8 @@ public class GasPricePendingTransactionsSorter extends AbstractPendingTransactio
       new TreeSet<>(
           comparing(PendingTransaction::hasPriority)
               .thenComparing(PendingTransaction::getGasPrice)
-              .thenComparing(PendingTransaction::getAddedAt)
-              .thenComparing(PendingTransaction::getSequence)
+              .thenComparing(PendingTransaction::getAddedAt, Comparator.reverseOrder())
```
fab-10, line 43: the comparison of `addedAt` is redundant and can be removed, since `sequence` is incremental

fab-10 (approved):
LGTM

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/GasPricePendingTransactionsSorter.java
matthew1001: Yes I wondered whether there was any difference in practice between the `addedAt` and `sequence` checks. I've removed `addedAt` under 9b0971ff72fb364c174b5c91ba57649adb5efa66

Issue #6104, opened 2023-10-30
besu t8n fails pyspec generated test


winsvega:
Get this strange failure on a pyspec test by besu

```
Response test_importRawBlock: 0xe8f86f237690ec4a5a2fe437b0780abd3ef86265e3a6fce1c3047a8af6befb5d

Request: test_setChainParams 
{
    "params" : {
        "fork" : "Cancun",
        "terminalTotalDifficulty" : "0x00",
        "constantinopleForkBlock" : "0x00",
        "byzantiumForkBlock" : "0x00",
        "homesteadForkBlock" : "0x00",
        "chainID" : "0x01"
    },
    "accounts" : {
        "0x00000000000000000000000000000000000000aa" : {
            "balance" : "0x00",
            "code" : "0x61001c600081600b8239f334600014601857346001146010575b005b47600055600e565b60aaff",
            "nonce" : "0x00",
            "storage" : {
            }
        },
        "0x00000000000000000000000000000000000000cc" : {
            "balance" : "0x3b9aca00",
            "code" : "0x60aa3b6000600060aa3c60aa3b60006003f06000600060006000600073864bbda5c698ac34b47a9ea3bd4228802cc5ce3b620186a0f162ca1101556000600060006000600173864bbda5c698ac34b47a9ea3bd4228802cc5ce3b620186a0f1",
            "nonce" : "0x01",
            "storage" : {
            }
        },
        "0x000000000000000000000000000000000000ba11" : {
            "balance" : "0x00",
            "code" : "0x73864bbda5c698ac34b47a9ea3bd4228802cc5ce3b31620ba1aa55",
            "nonce" : "0x00",
            "storage" : {
            }
        },
        "0x000000000000000000000000000000000000ba12" : {
            "balance" : "0x00",
            "code" : "0x73864bbda5c698ac34b47a9ea3bd4228802cc5ce3b31620ba1aa55",
            "nonce" : "0x00",
            "storage" : {
            }
        },
        "0x000f3df6d732807ef1319fb7b8bb8522d0beac02" : {
            "balance" : "0x00",
            "code" : "0x3373fffffffffffffffffffffffffffffffffffffffe14604d57602036146024575f5ffd5b5f35801560495762001fff810690815414603c575f5ffd5b62001fff01545f5260205ff35b5f5ffd5b62001fff42064281555f359062001fff015500",
            "nonce" : "0x01",
            "storage" : {
            }
        },
[...]

PR #6107, opened 2023-10-30
Correct reference test blobgas calculation

shemnon:
## PR description

Fix tpyo that resulted in an NPE in t8n blob gas calculations.


shemnon:
fixes #6104

macfarla (approved):

Issue #5322, opened 2023-04-10
implement pipeline on debug_ tracing endpoints 


macfarla:
similar to what was done for trace_ tracing endpoints https://github.com/hyperledger/besu/pull/5131

macfarla:
DebugTraceBlock
DebugTraceTransaction
DebugTraceBlockByHash
DebugTraceBlockByNumber

PR #6116, opened 2023-11-03
Pipeline for debug RPC endpoints

macfarla:
Increase efficiency by using pipeline for debug block endpoints

Refs #5322


macfarla:
I got stuck on creating an acceptance test for this, couldn't see a way to have a plugin node with the extra "rollup" RPC API enabled. We might need to add another method to Besu Node factory to create a plugin node with extra CLI args

PR #6117, opened 2023-11-03
clean up the ProcessableBlockHeader

pinges:
## PR description

This removes a few fields that don't make sense and that are not used in the existing core/ProcessableBlockHeader class. 
This introduces a interface ProcessableBlockHeader in the plugin module.

### DIFF
--- a/plugin-api/src/main/java/org/hyperledger/besu/plugin/data/ProcessableBlockHeader.java
+++ b/plugin-api/src/main/java/org/hyperledger/besu/plugin/data/ProcessableBlockHeader.java
@@ -80,13 +80,6 @@ public interface ProcessableBlockHeader {
   long getTimestamp();
 
   /**
-   * Returns the mixHash before merge, and the prevRandao value after.
-   *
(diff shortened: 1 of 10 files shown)


plugin-api/src/main/java/org/hyperledger/besu/plugin/data/ProcessableBlockHeader.java
```diff
@@ -0,0 +1,108 @@
...
+import org.hyperledger.besu.plugin.Unstable;
+
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes32;
+
+/**
+ * The minimum set of data for a BlockHeader, as defined in the <a href=
+ * "https://ethereum.github.io/yellowpaper/paper.pdf">Ethereum Yellow Paper</a>.
+ */
```
Gabriel-Trintinalia, line 29: nit: we may need a better description for this interface. It is not exactly the minimum set of data but a even smaller part of the header used during block construction.

Gabriel-Trintinalia (approved):
LGTM

PR #6119, opened 2023-11-03
Update RPCs for yParity

shemnon:
## PR description

* Update the GraphQL and JSON-RPC endpoints to provide `yParity` instead of `v` for non-legacy transactions.
* Update the JSON-RPC tests to use the Hive data. 
* Add JSON-RPC tests for Shanghai and Cancun blocks/fields.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -491,15 +490,26 @@ public BigInteger getV() {
 
     final BigInteger recId = BigInteger.valueOf(signature.getRecId());
 
+    if (transactionType != null && transactionType != TransactionType.FRONTIER) {
+      // EIP-2718 typed transaction, use yParity:
+      return null;
```
shemnon, line 495: JSON-RPC specs have the V field for type 0 (required) and type 1 and 2 (Optional and deprecated) but not for type 3 (no field).  If we wanted to echo the yParity in type 1 and 2 transactions this would be the best place to do it.

As the author my bias is to not include it, but I am persuadable for types 1 and 2.

jflo (approved):
Concur with the inclination to leave as is for legacy types of transaction.

PR #6138, opened 2023-11-07
Backward sync invalid block causes Out of Memory error

### LINKED ISSUE #6118: Out of memory due to invalid block
### Description
Besu fails to import a block due to an invalid block error during BWS and then fails with an OutOfMemoryError.

From the heap dump it looks like caused by [`emitBadChainEvent`](https://github.com/hyperledger/besu/blob/main/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContext.java#L372) in  `BackwardSyncContext` accumulating a very large number of blocks and headers. It doesn't look like we ever exit the loop.

top retained objects from the heap dump
Screenshot 2023-11-03 at 4 47 59 pm

I am able to recreate this in my test environment consistently by restarting the node. It is occurring on 3/4 of the fullsync goerli nodes.

Tested using build from https://github.com/hyperledger/besu/commit/a90e4487fc67f6d9315cf9499fda534490156f50

jframe:
## PR description
Besu fails to import a block due to an invalid block error during BWS and then fails with an OutOfMemoryError.

The fix is to limit the included blocks and block headers in the `onBadChain` event. I have set the limit to be the same as what is set for the maximum number of entries cached in the `BadBlockManager`. The `BadBlockManager` is currently the only listener for this event and the limit of 100 also seems reasonable for the `onBadChain` event.

### Heap dump analysis
image (1)

image (2)

## Fixed Issue(s)

fixes #6118


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Upgrade netty to address CVE-2023-44487, CVE-2023-34462 [#6100](https://github.com/hyperledger/besu/pull/6100)
 - Upgrade grpc to address CVE-2023-32731, CVE-2023-33953, CVE-2023-44487, CVE-2023-4785 [#6100](https://github.com/hyperledger/besu/pull/6100)
 - Fix blob gas calculation in reference tests [#6107](https://github.com/hyperledger/besu/pull/6107)
+- Limit memory used in handling invalid blocks [#6139](https://github.com/hyperledger/besu/pull/6138)
```
macfarla, line 26:
```suggestion
- Limit memory used in handling invalid blocks [#6138](https://github.com/hyperledger/besu/pull/6138)
```

macfarla (approved):
LGTM just a typo in changelog entry

CHANGELOG.md
jframe: done

jflo (approved):

siladu (approved):
Nice improvement!

PR #6143, opened 2023-11-08
Synchronize access to block header

### LINKED ISSUE #6140: Chain head block is sometimes not available on the node that actually mined the block
### Description
I've been experiencing occasional `CHAIN_HEAD_NOT_AVAILABLE` errors in a single-validator QBFT chain. Adding some extra sysout logs to a local build I can see that when this happens, the failing check is logged out before the existing "Produced [#block-number]" log entry for that requested block:

```
MRW: Blockchain head returned nothing for the block hash 0x10b029608b21f63b34884b36e29030332200677ed4ad0890b46342c494e29753?
2023-11-07 11:13:29.056+00:00 | BftProcessorExecutor-QBFT-0 | INFO  | QbftBesuControllerBuilder | Produced #5,182 / 21 tx / 11 pending / 40,001,640 (40.0%) gas / (0x10b029608b21f63b34884b36e29030332200677ed4ad0890b46342c494e29753)
2023-11-07 11:13:29.094+00:00 | EthScheduler-Transactions-0 | WARN  | TransactionPool | rejecting transaction 0x140a7a8af4bd571fdbc251f11e92a84f6796728eff20bff704c1896865a8a87e due to chain head not available yet
```

### Steps to Reproduce (Bug)
1. Create a 4 node QBFT chain.
2. Only one of the nodes is a validator. On a single laptop this appears to be easy to recreate in 5-10 minutes, probably due to CPU load.
3. Enabled transaction pool trace on the validator node
4. Submit ~ 50TPS to a non-validator node
5. After a short period of time you should hit `EthScheduler-Transactions-0 | TRACE  | TransactionPool | rejecting transaction`

(Note that this error is a `WARN` error as of https://github.com/hyperledger/besu/issues/6126)

**Expected behavior:**
No `CHAIN_HEAD_NOT_AVAILABLE` errors are seen in this sort of load testing

matthew1001:
## PR description
As outlined in the comment in https://github.com/hyperledger/besu/issues/6140 it appears to be the case that access to the block header isn't thread safe when the block being requested is a newly added chain head.

## Fixed Issue(s)
Fixes #6140


garyschulte:
Should we need to synchronize blockheader access?  This isn't something that typically needs concurrency safety, except of course near-head.  

Is the transaction spam using block number or blockhash?  I can understand chain head not available if the tx spammer is getting ahead of itself by predicting what the blocknumber should be based on block time.  But if it is blockhash, where is the tx spammer getting that blockhash?  

Without context, this seems like an issue that should be solved by ensuring we commit and persist the blockchain transaction rather than paying a synchronization overhead cost when trying to access block headers.

matthew1001:
Yeah it's exactly the near-head case that the problem occurs. Basically the chain header is updated (inside a synchronized function) and then storage is updated and committed (in the same synchronized function). 

But the getter for the chain head can be called at any time, and because it's not synchronized it can retrieve the new value before the storage commit has happened.

So possibly an alternative to synchronizing on `getBlockHeader()` would be to synchronize on `getChainHeadHash()` (and other related `getChainHead*()` methods probably) if getting the chain head is less frequent compared to `getBlockHeader()`. What do you think?

matthew1001:
Going to run some tests with a change to synchronizing `getChainHead*`, not `getBlockHeader`, and if it fixes the issue I'll update the PR

matthew1001:
@garyschulte I'm running my tests again using your suggestion of overloading `getBlockHeader` by adding `getBlockHeaderSafe` which is synchronized, and the caller can choose to use in the rare event that the block header for the latest block isn't available. I'll update here once I'm happy that the issue isn't showing any more. Let me know what you think of the changes.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -418,13 +418,19 @@ private TransactionValidator getTransactionValidator() {
   private ValidationResultAndAccount validateTransaction(
       final Transaction transaction, final boolean isLocal, final boolean hasPriority) {
 
-    final BlockHeader chainHeadBlockHeader = getChainHeadBlockHeader().orElse(null);
+    // Optimistically get the chain head. getChainHeadBlockHeader() doesn't take any locks,
+    // which might mean that the latest block is still being committed to storage. If this
+    // call fails try the synchronized alternative, and if that fails give up.
+    BlockHeader chainHeadBlockHeader = getChainHeadBlockHeader().orElse(null);
```
garyschulte, line 424: how about we leave this as it is and have this logic directly in getChainBlockHeader()?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -554,6 +560,11 @@ private Optional<BlockHeader> getChainHeadBlockHeader() {
     return blockchain.getBlockHeader(blockchain.getChainHeadHash());
   }
 
+  private Optional<BlockHeader> getChainHeadBlockHeaderSafe() {
+    final MutableBlockchain blockchain = protocolContext.getBlockchain();
+    return blockchain.getBlockHeaderSafe(blockchain.getChainHeadHash());
+  }
+
```
garyschulte, line 567: how about instead of adding here we add the safe call to the existing getChainHeadBlockHeader(), e.g.:

```
  private Optional<BlockHeader> getChainHeadBlockHeader() {
    final MutableBlockchain blockchain = protocolContext.getBlockchain();
    return blockchain.getBlockHeader(blockchain.getChainHeadHash())
        .or(() -> blockchain.getBlockHeaderSafe(blockchain.getChainHeadHash()));
  }
```

garyschulte (changes requested):
Seems good to me, but I think we can DRY it up a bit

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
matthew1001: Yeah that's a good thought. I'll tidy it up and push a new commit.

matthew1001: New commits pushed @garyschulte. I'll mark as auto-merge once you're happy with them.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
matthew1001: See latest commits

fab-10:
I like the approach of the `*Safe` version of the method, let's monitor this to see if the issue is gone, otherwise as suggested before we can implement a retry policy, in case of a transient error, like chain head not found.

matthew1001:
The current PR definitely fixes the issue - no failures after >10 hours running a single validator at 50TPS. Question is whether we stick with the refactor I did under 1641b23c137127c779d96885ec32e54f8a48a0e8 or revert to having separate `getChainHeadBlockHeader()` and `getChainHeadBlockHeaderSafe()`. I don't have a strong opinion either way. With the latest commit, `getChainHeadBlockHeader()` will make 2 attempts to get a block header if the block being requested genuinely doesn't exist. But for any block `<` chain head (and most calls for the `chain head` block except in very occasional circumstances) this shouldn't affect performance. So I'm happy sticking with the latest refactor, but likewise happy to revert.

@garyschulte any thoughts? It would be nice to get this into `23.10.2` as it causes unpleasant behaviour if you hit it.

garyschulte (approved):
:+1:  LGTM, especially if this solves the case you are encountering.

PR #6146, opened 2023-11-09
Apply the same reverse sort order as https://github.com/hyperledger/b…

matthew1001:
## PR description
PR https://github.com/hyperledger/besu/pull/6106 reversed the transaction sort order, but only for the gas-price sorter.

This PR applies the same logic to the base-fee sorter.

FYI @fab-10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 23.10.2
 
 ### Breaking Changes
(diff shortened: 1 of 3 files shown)


fab-10 (approved):

matthew1001:
Just looking into the only remaining unit test failure, `shouldPrioritizeGasPriceThenSequenceInPool` - want to make sure that I've not altered the eviction order

matthew1001:
I've decided to close this PR for the time being. The priority order for selection (which this PR was intended to alter) also defines the priority order for eviction (which it wasn't). There may be an argument for evicting by highest-nonce first which is ultimately what this PR would do, but I think that needs more discussion as it's a pretty fundamental behaviour today.

matthew1001:
I've updated the change log to describe this (and https://github.com/hyperledger/besu/pull/6106) as breaking changes in `23.10.2`.

PR #6147, opened 2023-11-09
Cached blob txs

### LINKED ISSUE #6025: Blob Transactions Not Re-added to Transaction Pool During Reorg
### Description
Blob Transactions Not Re-added to Transaction Pool During Reorg. This happens because we deserialise the transaction from the block during the reorg. At this stage, the RLP encoded version of blobs does not contain the blobs with commitments data, causing the transaction to fail the transaction pool validation.

### Steps to Reproduce (Bug)
1. Include a blob transaction in a block.
2. Trigger a reorg.
3. Notice that the blob transaction is not returned to the transaction pool.

This issue is reproduced by the `shouldNotReAddBlobTxsWhenReorgHappens` test.

**Expected behaviour:** 
During a reorg, blob transactions should be re-added to the transaction pool to ensure no transactions are lost, providing a consistent user experience. This should occur even though the deserialisation process does not include the blobs with commitment data.

**Actual behaviour:** 
When a reorg occurs, blob transactions are not added back to the transaction pool.

jflo:
fixes #6025 

- adds LRU cache to retain blobs for 3 epochs in case of re-orgs so they can be restored to tx pool for future inclusion.
- worst case memory usage would be around 77MB on heap
- adds means for Transaction.Builder to copy an immutable Transaction
- upgrades errorprone, checkerframework, and caffine
- miscellaneous sonar fixes.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -240,20 +240,29 @@ private ValidationResult<TransactionInvalidReason> addTransaction(
       metrics.incrementRejected(isLocal, hasPriority, TRANSACTION_ALREADY_KNOWN, "txpool");
       return ValidationResult.invalid(TRANSACTION_ALREADY_KNOWN);
     }
+    Transaction toAdd = transaction;
+    // if adding a blob tx, and it is missing its blob, is a re-org and we should restore the blob
+    // from cache.
+    if (transaction.getType().supportsBlob() && transaction.getBlobsWithCommitments().isEmpty()) {
+      final Optional<Transaction> maybeCachedBlob = pendingTransactions.restoreBlob(transaction);
+      if (maybeCachedBlob.isPresent()) {
+        toAdd = maybeCachedBlob.get();
```
fab-10, line 249: it is more efficient to move this into `reAddTransactions` method, where reorged txs are managed, so avoiding to do this check for every incoming tx.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractTransactionsLayer.java
```diff
@@ -361,7 +365,12 @@ protected PendingTransaction processRemove(
       final Transaction transaction,
       final RemovalReason removalReason) {
     final PendingTransaction removedTx = pendingTransactions.remove(transaction.getHash());
+
     if (removedTx != null) {
+      if (removedTx.getTransaction().getBlobsWithCommitments().isPresent()
+          && CONFIRMED.equals(removalReason)) {
+        this.blobCache.cacheBlobs(removedTx.getTransaction());
+      }
```
fab-10, line 373: Since this only applies when the tx is confirmed, then a best place for it, is in the `confirmed` method, so you do not even need to check the removal reason

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractTransactionsLayer.java
```diff
@@ -87,6 +90,7 @@ public AbstractTransactionsLayer(
     metrics.initSpaceUsed(this::getLayerSpaceUsed, name());
     metrics.initTransactionCount(pendingTransactions::size, name());
     metrics.initUniqueSenderCount(txsBySender::size, name());
+    this.blobCache = new BlobCache();
```
fab-10, line 93: It is better to create the blob cache, outside, and then pass it as argument to the layers, because creating it here will result in one cache per layers that extend this class, so now there will be 3 different caches.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/sorter/AbstractPendingTransactionsSorter.java
```diff
@@ -97,6 +98,8 @@ public abstract class AbstractPendingTransactionsSorter implements PendingTransa
   protected final TransactionPoolReplacementHandler transactionReplacementHandler;
   protected final Supplier<BlockHeader> chainHeadHeaderSupplier;
 
+  private final BlobCache blobCache;
```
fab-10, line 101: just a note to say that adding the blob cache to the `legacy` txpool is a nice to have, but we do not recommend to use it for EIP4844

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/BlobCache.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class BlobCache {
+  private final Cache<VersionedHash, BlobsWithCommitments.BlobQuad> cache;
+  private static final Logger LOG = LoggerFactory.getLogger(BlobCache.class);
+
+  public BlobCache() {
+    this.cache =
+        Caffeine.newBuilder()
+            .maximumSize(6 * 32 * 3l)
+            .build(); // 6 blobs max per 32 slots per 3 epochs
```
fab-10, line 38: configured this way the cache will always fill at its maximum before starting to evict older blobs, while you said you want 3 epochs, so instead of `maximumSize` it should be configured with `expireAfterWrite`

fab-10 (changes requested):
Overall looks good, there are some changes need to make it works with layered txpool and respect the 3 epochs expiration time, then some other optimization.

datatypes/src/main/java/org/hyperledger/besu/datatypes/BlobsWithCommitments.java
```diff
@@ -15,15 +15,35 @@
...
+      return Objects.equals(blob, blobQuad.blob)
+          && Objects.equals(kzgCommitment, blobQuad.kzgCommitment)
+          && Objects.equals(kzgProof, blobQuad.kzgProof)
+          && Objects.equals(versionedHash, blobQuad.versionedHash);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(blob, kzgCommitment, kzgProof, versionedHash);
+    }
```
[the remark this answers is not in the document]
jflo: I expected that also, but tests showed there were just comparing memory pointers.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/BlobCache.java
```diff
@@ -0,0 +1,91 @@
...
+              .toList()
+              .containsAll(maybeHashes.get())) {
+            txBuilder.blobsWithCommitments(bwc);
+            return Optional.of(txBuilder.build());
+          } else {
+            LOG.debug("did not find all versioned hashes to restore from cache");
+            return Optional.empty();
+          }
+        } else {
+          LOG.debug("can't restore blobs for transaction with empty list of versioned hashes");
```
fab-10, line 77: This should never happen, since the validation ensure these are present, if we reach this code, it means that there is a bug somewhere, so should we warn here?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/BlobCache.java
```diff
@@ -0,0 +1,91 @@
...
+          return Optional.empty();
+        }
+      } else {
+        LOG.debug("can't restore blobs for transaction without list of versioned hashes");
```
fab-10, line 81: same a above

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -240,20 +240,21 @@ private ValidationResult<TransactionInvalidReason> addTransaction(
       metrics.incrementRejected(isLocal, hasPriority, TRANSACTION_ALREADY_KNOWN, "txpool");
       return ValidationResult.invalid(TRANSACTION_ALREADY_KNOWN);
     }
+    Transaction toAdd = transaction;
```
fab-10, line 243: These changes can be reverted now

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -380,8 +381,14 @@ public void onBlockAdded(final BlockAddedEvent event) {
...
     if (!reAddTransactions.isEmpty()) {
-      var txsByOrigin =
+      // if adding a blob tx, and it is missing its blob, is a re-org and we should restore the blob
+      // from cache.
+      List<Transaction> restoredTransactions =
           reAddTransactions.stream()
+              .map(t -> pendingTransactions.restoreBlob(t).orElse(t))
+              .toList();
+      var txsByOrigin =
+          restoredTransactions.stream()
```
fab-10, line 391: Opt: concatenate the streams

```suggestion
      List<Transaction> txsByOrigin =
          reAddTransactions.stream()
              .map(t -> pendingTransactions.restoreBlob(t).orElse(t))
```

fab-10 (commented):
Looks good now, few more suggestions

fab-10 (approved):

PR #4425, opened 2022-09-22
Transaction pool improvements to avoid filling the pool with not executable transactions

fab-10:
## PR description

This PR is built on top of #4413 which was used to investigate the transaction pool, with the result that currently we keep many transactions that are not possible to include in a block, because a transaction with a lower nonce is missing or invalid.
To avoid that issues first we now remember (in a cache) that that sender has an invalid transaction, and we do not accept transactions with higher nonce from that sender, of course if a replacement for the invalid transaction is sent it is accepted.
The eviction when the pool is full has changed in order to pick the sender of the oldest transaction in the pool, and the evict the transaction with the higher nonce for that sender, so no gaps in the nonce list for that sender are created.

## Fixed Issue(s)

fixes #4401 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if
    [updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #6148, opened 2023-11-09
Remove LowestInvalidNonceCache

### LINKED ISSUE #6058: [FEATURE REQUEST] Set  DEFAULT_LOWEST_INVALID_KNOWN_NONCE_CACHE configurable, like others transaction pool options

fab-10:
## PR description

LowestInvalidNonceCache was introduce in #4425, as a workaround for avoiding that non executable txs fill the txpool, while we were implementing an better solution to the problem, that become the new layered txpool that is now the default option in Besu, and is tailored for public network.
At this point we can make the simple legacy txpool more private network friendly, and the LowestInvalidNonceCache is usually creating issues these networks, so it makes sense to remove this workaround, since there should be less risk of spammy txs there, and in case users can always switch to layered.

## Fixed Issue(s)

fixes #6058


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Add `yParity` to GraphQL and JSON-RPC for relevant querise. [6119](https://github.com/hyperledger/besu/pull/6119)
 - Force tx replacement price bump to zero when zero base fee market is configured or `--min-gas-price` is set to 0. This allows for easier tx replacement in networks where there is not gas price. [#6079](https://github.com/hyperledger/besu/pull/6079)
 - Introduce the possibility to limit the time spent selecting pending transactions during block creation, using the new experimental option `Xblock-txs-selection-max-time` on PoS and PoW networks (by default set to 5000ms) or `Xpoa-block-txs-selection-max-time` on PoA networks (by default 75% of the min block time) [#6044](https://github.com/hyperledger/besu/pull/6044)
+- Remove LowestInvalidNonceCache from `legacy` transaction pool to make it more private networks friendly [#6148](https://github.com/hyperledger/besu/pull/6148)
```
macfarla, line 22: this isn't a breaking change, right? no CLI options or other user config

macfarla (approved):
LGTM

CHANGELOG.md
fab-10: correct, not a breaking change.

PR #6150, opened 2023-11-09
Refactor some transaction calculations

shemnon:
Refactor some transaction calculations, including moving upfront 
overflow checks from the constructor to the validator and optimizing 
some RLP calculations.  Also fix all lint errors.


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetTransactionByHashTest.java
```diff
@@ -192,7 +193,7 @@ private Set<PendingTransaction> getTransactionPool() {
     final BlockDataGenerator gen = new BlockDataGenerator();
     Transaction pendingTransaction = gen.transaction();
     System.out.println(pendingTransaction.getHash());
-    return gen.transactionsWithAllTypes(4).stream()
+    return gen.transactions(4, TransactionType.EIP1559).stream()
```
jflo, line 196: ran into this problem myself. i like that this makes the whole test more deterministic, but won't it reduce use case coverage now that pool won't contain non-1559 txs?

shemnon: The test only checks the first item in the pool. 
https://github.com/hyperledger/besu/pull/6150/files/94618ea3f7f7e80c8eebbcb9d8c7a7963754808f#diff-1f922da5074b3ed31f2fb253ea930a81c53cd503c27cb5be77d4d5e603a857b1L159
That is the only test that uses this method.

shemnon: fixed in #6151

fab-10 (approved):

PR #6130, opened 2023-11-06
Add `rpc-gas-cap` to CLI options

gfukushima:
## PR description
This PR introduces the CLi option`rpc-gas-cap` which allows users to limit the amount of gas used by the  RPC methods `eth_call` and `eth_getEstimateGas`. 

## Fixed Issue(s)

Fixes #6042

PR #6156, opened 2023-11-10
Add `rpc-gas-cap` via transaction simulator

### LINKED ISSUE #6042: Feature Request: eth_call gas limit (rpc.gascap)
### Description
As an RPC provider, I want a way to limit the maximum amount of gas that an `eth_call` can consume.

Besu is the only Ethereum client that does not have a configurable RPC gas limit/cap.

### Acceptance Criteria
* Implement a flag similar to Geth, Nethermind, Reth, Erigon, etc. `rpc.gascap` with a configurable gas limit.

If someone sends an eth_call that consumes a bajillion gas, it should be killed with out of gas or other similar error.

I'm not sure what the current behavior is.

gfukushima:
## PR description
This PR adds `rpc-gas-cap` directly into the transaction simulator, which seems a better approach than the one used in #6130.

## Fixed Issue(s)

Fixes #6042

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
@@ -192,6 +192,7 @@ public class RunnerBuilder {
   private JsonRpcIpcConfiguration jsonRpcIpcConfiguration;
   private boolean legacyForkIdEnabled;
   private Optional<Long> rpcMaxLogsRange;
+  private Optional<Long> rpcGasCap;
   private Optional<EnodeDnsConfiguration> enodeDnsConfiguration;
 
   /**
@@ -585,6 +586,16 @@ public class RunnerBuilder {
     this.rpcMaxLogsRange = rpcMaxLogsRange > 0 ? Optional.of(rpcMaxLogsRange) : Optional.empty();
     return this;
   }
+  /**
+   * Add Rpc gasLimit cap .
+   *
+   * @param rpcGasCap the rpc max logs range
+   * @return the runner builder
+   */
+  public RunnerBuilder rpcGasCap(final Long rpcGasCap) {
+    this.rpcGasCap = rpcGasCap > 0 ? Optional.of(rpcGasCap) : Optional.empty();
+    return this;
+  }
 
   /**
    * Add enode DNS configuration
@@ -661,7 +672,7 @@ public class RunnerBuilder {
 
     final TransactionSimulator transactionSimulator =
         new TransactionSimulator(
-            context.getBlockchain(), context.getWorldStateArchive(), protocolSchedule);
+            context.getBlockchain(), context.getWorldStateArchive(), protocolSchedule, rpcGasCap);
 
     final Bytes localNodeId = nodeKey.getPublicKey().getEncodedBytes();
     final Optional<NodePermissioningController> nodePermissioningController =
@@ -910,6 +921,7 @@ public class RunnerBuilder {
       graphQlContextMap.putIfAbsent(GraphQLContextType.SYNCHRONIZER, synchronizer);
       graphQlContextMap.putIfAbsent(
           GraphQLContextType.CHAIN_ID, protocolSchedule.getChainId().map(UInt256::valueOf));
+      graphQlContextMap.putIfAbsent(GraphQLContextType.GAS_CAP, rpcGasCap);
       final GraphQL graphQL;
       try {
         graphQL = GraphQLProvider.buildGraphQL(fetchers);
@@ -1240,7 +1252,8 @@ public class RunnerBuilder {
                 besuController.getProtocolManager().ethContext().getEthPeers(),
                 consensusEngineServer,
                 rpcMaxLogsRange,
-                enodeDnsConfiguration);
+                enodeDnsConfiguration,
+                rpcGasCap);
     methods.putAll(besuController.getAdditionalJsonRpcMethods(jsonRpcApis));
 
     final var pluginMethods =
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -1234,6 +1234,12 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   private final Long rpcMaxLogsRange = 5000L;
 
   @CommandLine.Option(
+      names = {"--rpc-gas-cap"},
+      description =
+          "Specifies the gasLimit cap for transaction simulation RPC methods. Must be >=0. 0 specifies no limit  (default: ${DEFAULT-VALUE})")
+  private final Long rpcGasCap = 0L;
+
+  @CommandLine.Option(
       names = {"--cache-last-blocks"},
       description = "Specifies the number of last blocks to cache  (default: ${DEFAULT-VALUE})")
   private final Integer numberOfblocksToCache = 0;
@@ -2892,6 +2898,7 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
             .storageProvider(keyValueStorageProvider(keyValueStorageName))
             .rpcEndpointService(rpcEndpointServiceImpl)
             .rpcMaxLogsRange(rpcMaxLogsRange)
+            .rpcGasCap(rpcGasCap)
             .enodeDnsConfiguration(getEnodeDnsConfiguration())
             .build();
 
--- a/besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
+++ b/besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
@@ -1585,6 +1585,20 @@ public class BesuCommandTest extends CommandTestAbstract {
   }
 
   @Test
+  public void rpcGasCapOptionMustBeUsed() {
+    final long rpcGasCap = 150L;
+    parseCommand("--rpc-gas-cap", Long.toString(rpcGasCap));
+
+    verify(mockRunnerBuilder).rpcGasCap(longArgumentCaptor.capture());
+    verify(mockRunnerBuilder).build();
+
+    assertThat(longArgumentCaptor.getValue()).isEqualTo(rpcGasCap);
+
+    assertThat(commandOutput.toString(UTF_8)).isEmpty();
+    assertThat(commandErrorOutput.toString(UTF_8)).isEmpty();
+  }
+
+  @Test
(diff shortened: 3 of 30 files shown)


besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -585,6 +586,16 @@ public RunnerBuilder rpcMaxLogsRange(final Long rpcMaxLogsRange) {
     this.rpcMaxLogsRange = rpcMaxLogsRange > 0 ? Optional.of(rpcMaxLogsRange) : Optional.empty();
     return this;
   }
+  /**
+   * Add Rpc gasLimit cap .
+   *
+   * @param rpcGasCap the rpc max logs range
```
siladu, line 592: typo

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2926,6 +2932,7 @@ private Runner synchronize(
             .storageProvider(keyValueStorageProvider(keyValueStorageName))
             .rpcEndpointService(rpcEndpointServiceImpl)
             .rpcMaxLogsRange(rpcMaxLogsRange)
+            .rpcGasCap(rpcGasCap)
```
siladu, line 2935: Wonder if this kind of thing should live in a configuration object. Not sure which one. Feels like rpcMaxLogsRange should live in the same one...probably an issue for another PR.

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -1587,6 +1587,20 @@ public void rpcMaxLogsRangeOptionMustBeUsed() {
     assertThat(commandErrorOutput.toString(UTF_8)).isEmpty();
   }
 
+  @Test
+  public void rpcGasCapOptionMustBeUsed() {
```
siladu, line 1591: Bit confused what this test name means...it's optional right? In what sense must it be used?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/EthJsonRpcMethods.java
```diff
@@ -96,7 +97,8 @@ public EthJsonRpcMethods(
       final TransactionPool transactionPool,
       final MiningCoordinator miningCoordinator,
       final Set<Capability> supportedCapabilities,
-      final Optional<Long> maxLogRange) {
+      final Optional<Long> maxLogRange,
+      final Optional<Long> gasCap) {
```
siladu, line 101: Was wondering if it might make sense to use Long.MAX_VALUE instead of a optional field...but also like that this same pattern as maxLogRange.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -207,10 +213,17 @@ public Optional<TransactionSimulatorResult> processWithWorldUpdater(
...
-    final long gasLimit =
+    long gasLimit =
         callParams.getGasLimit() >= 0
             ? callParams.getGasLimit()
             : blockHeaderToProcess.getGasLimit();
+    if (rpcGasCap.isPresent()) {
+      final long gasCap = rpcGasCap.get();
+      if (gasCap < gasLimit) {
+        gasLimit = gasCap;
+        LOG.info("Capping gasLimit to " + gasCap);
```
siladu, line 224: What happens when the cap is reached?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -523,6 +527,78 @@ public void shouldReturnSuccessfulResultWhenEip1559TransactionProcessingIsSucces
...
+
+    final Optional<TransactionSimulatorResult> result =
+        cappedTransactionSimulator.process(
+            callParameter,
+            TransactionValidationParams.transactionSimulator(),
+            OperationTracer.NO_TRACING,
+            1L);
+
+    assertThat(result.get().isSuccessful()).isTrue();
+    verifyTransactionWasProcessed(expectedTransaction);
```
siladu, line 563: How is this asserting that the gas was capped?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -523,6 +527,78 @@ public void shouldReturnSuccessfulResultWhenEip1559TransactionProcessingIsSucces
...
+
+    mockBlockchainForBlockHeader(blockHeader);
+    mockWorldStateForAccount(blockHeader, callParameter.getFrom(), 1L);
+
+    final Transaction expectedTransaction =
+        Transaction.builder()
+            .type(TransactionType.EIP1559)
+            .chainId(BigInteger.ONE)
+            .nonce(1L)
+            .gasLimit(GASCAP.get())
```
siladu, line 544: Is this on or over the capped boundary?

siladu (commented):
Looks straight forward enough, just want to understand what TransactionSimulatorTest is testing before approving.

nit: From a readability POV, I like the idea of a GasCappedTransactionSimulator class.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -523,6 +527,78 @@ public void shouldReturnSuccessfulResultWhenEip1559TransactionProcessingIsSucces
     verifyTransactionWasProcessed(expectedTransaction);
   }
 
+  @Test
+  public void shouldCapGasLimitWhenRpcGasCapIsDefined() {
+    final CallParameter callParameter = eip1559TransactionCallParameter(Wei.ZERO, Wei.ZERO, 5000L);
```
siladu, line 532: s/5000L/`GASCAP + 1`?

besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
gfukushima: good catch

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
gfukushima: This was added following the pattern that a lot of other CLI options follow. Agree that the name could be better but I take that it means that the option must be used when specified and throw no error.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
gfukushima: we should return a -32003 INTRINSIC_GAS_EXCEEDS_LIMIT

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
gfukushima: adopted your suggestion

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
gfukushima: this is asserting that the expected transaction processed was the one with the gasLimit overwritten with GASCAP

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -518,18 +518,19 @@ public void shouldReturnSuccessfulResultWhenEip1559TransactionProcessingIsSucces
...
             .signature(FAKE_SIGNATURE)
             .build();
-    mockProcessorStatusForTransaction(expectedTransaction, Status.SUCCESSFUL);
+    mockProcessorStatusForTransaction(expectedTransaction, Status.FAILED);
 
     final Optional<TransactionSimulatorResult> result =
         transactionSimulator.process(callParameter, 1L);
 
-    assertThat(result.get().isSuccessful()).isTrue();
+    assertThat(result.get().isSuccessful()).isFalse();
```
siladu, line 526: Why has this seemingly unrelated test been changed to a failure case? The name still suggests it should be successful too.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -559,13 +560,14 @@ public void shouldCapGasLimitWhenRpcGasCapIsDefined() {
...
     verifyTransactionWasProcessed(expectedTransaction);
   }
 
   @Test
-  public void shouldKeepOriginalGasLimitWhenCapIsHigherThanOriginalValue() {
-    final CallParameter callParameter = eip1559TransactionCallParameter(Wei.ZERO, Wei.ZERO, 200L);
+  public void shouldApplyGasCapWhenOriginalGasLimitIsLowerThanGasCap() {
```
siladu, line 568: Why do we apply the gas cap when gas limit is already lower?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
gfukushima: The actual result doesn't matter for this test in fact they are all mocked base on the param you pass in the mockProcessorStatusForTransaction method. I've refactored the tests a little bit to make it less dependent on this mock.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
gfukushima: whoops, thanks for catching this.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
gfukushima: I realize now I touched the wrong test, fixed it to be as it was originally.

siladu (approved):
LGTM

PR #6163, opened 2023-11-13
Transactions that takes too long to evaluate, during block creation, are dropped from the txpool

fab-10:
## PR description

In PR #6044, a timeout has been introduced to avoid that the transactions selection, during block creation, could take an indefinite amount of time, this PR iterates on it checking if the tx that was evaluating when the timeout occurred, took longer that the max amount of time allocated for the block creation, and if this was the case, the tx is removed from the pool, to avoid that it will be retried in future block creations.


gfukushima (approved):
lgtm

PR #6168, opened 2023-11-15
Fix EthFeeHistory reward calculation

Gabriel-Trintinalia:
### Description
fixes: #6170 
eth_feeHistory returns wrong rewards. The following request on Sepolia returns: 

```
{
   "jsonrpc":"2.0",
   "method":"eth_feeHistory",
   "params":[
      "0x5",
      "0x3FA351",
      [
         0,
         20,
         30,
         100
      ]
   ],
   "id":1
}
```

Expected Result (Green) / Actual Result (red): 
```diff
{
  "baseFeePerGas": [
    "0x76c73ce5",
    "0x7e1f5d40",
    "0x7f8aed3c",
    "0x77be8999",
    "0x68c6b866",
    "0x6716f547"
  ],
  "gasUsedRatio": [
    0.7473261666666666,
    0.5450408333333333,
    0.2554263333333333,
    0,
    0.43561246666666664
  ],
  "oldestBlock": "0x3fa34d",
  "reward": [
    [
-     "0x27c9",
-     "0x27c9",
-     "0x27c9",
-     "0x27c9"
+     "0x27c9",
+     "0x77359400",
+     "0x77359400",
+     "0x1660f131b"
    ],
    [
-     "0x27c9",
-     "0x27c9",
-     "0x27c9",
-     "0x27c9"
+     "0x27c9",
+     "0x59682f00",
+     "0x59682f00",
+     "0x15eb6f2c0"
    ],
    [
-     "0x0",
-     "0x0",
-     "0x0",
-     "0x27c9"
+     "0x0",
+     "0x27c9",
+     "0x5f5e100",
+     "0x15d4b62c4"
    ],
    [
      "0x0",
      "0x0",
      "0x0",
      "0x0"
    ],
    [
-     "0x0",
-     "0x0",
-     "0x0",
-     "0x27c9"
+     "0x0",
+     "0x27c9",
+     "0xe00ac48",
+     "0x43f510f9a"
    ]
  ]
}
```


Gabriel-Trintinalia:
### PR Test explanation

Spec: A given percentile sample of effective priority fees per gas from a single block in ascending order, weighted by gas used. Zeroes are returned if the block is empty.

The current implementation does not pass the test.

The `gasUsedAndFee` list represents a series of transactions within a block. Each transaction is represented as a pair of values: the amount of gas used by the transaction and the fee of the transaction. As we move through the list, the cumulative gas used by the transactions represents different percentiles of the total gas used in the block.

The `rewardPercentiles` list represents the specific percentiles of gas usage that we're interested in.

The `expectedRewards` list represents the fees corresponding to the percentiles in the `rewardPercentiles` list. The fee for a given percentile is the transaction fee that brings the cumulative gas used to that percentile.

For example, the first transaction in the gasUsedAndFee list uses 100 gas and has a fee of 1 Wei. This transaction represents 5% of the total gas used in the block, so the 5th percentile corresponds to a fee of 1 Wei. The second transaction uses 150 gas and has a fee of 2 Wei. This transaction brings the cumulative gas used to 12.5% of the total, so the 10th percentile corresponds to a fee of 2 Wei. And so on.

The 30th and 31st percentiles correspond to a fee of 5 Wei, because the fifth transaction in the `gasUsedAndFee` list uses 200 gas and has a fee of 5 Wei, bringing the cumulative gas used to 37.5% of the total.

So, the `expectedRewards` list makes sense because it matches the fees of the transactions at each percentile in the rewardPercentiles list.

gfukushima (approved):
lgtm

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -226,23 +226,26 @@ record TransactionInfo(Transaction transaction, Long gasUsed, Wei effectivePrior
...
-
-      while (rewardPercentileIndex < rewardPercentiles.size()
-          && 100.0 * gasUsed / block.getHeader().getGasUsed()
-              >= rewardPercentiles.get(rewardPercentileIndex)) {
-        rewards.add(transactionAndGasUsed.effectivePriorityFeePerGas);
-        rewardPercentileIndex++;
+    // Start with the gas used by the first transaction
+    double totalGasUsed = transactionsAndGasUsedAscendingEffectiveGasFee.get(0).gasUsed();
+    var transactionIndex = 0;
+    for (var rewardPercentile : rewardPercentiles) {
```
gfukushima, line 233: Would prefer this to declared as the corresponding primitive if possible

macfarla (approved):

PR #6187, opened 2023-11-20
consider peer reputation score when deciding to disconnect

macfarla:
Noticed in peering logs we are disconnecting quite useful peers with timeouts and "useless responses". This exacerbates peering issues because we are quite early to disconnect, whereas if we just wait a bit longer the peer can still be useful. More of an issue in low peer count networks. 
* consider peer reputation score when deciding whether to disconnect peer for repeated timeouts/useless responses
* increase threshold for timeout counts
* decrease score points deducted for timeout/useless response

```
➜  besu grep  "Disconnect - Outbound" besu03.log
2023-11-17 17:47:27.730+10:00 | nioEventLoopGroup-3-5 | DEBUG | EthProtocolManager | Disconnect - Outbound - 0x0b TIMEOUT - 0x45d5... - 5 peers left
2023-11-17 17:49:10.232+10:00 | nioEventLoopGroup-3-1 | DEBUG | EthProtocolManager | Disconnect - Outbound - 0x0b TIMEOUT - 0xcbba... - 4 peers left
2023-11-17 18:02:33.476+10:00 | nioEventLoopGroup-3-8 | DEBUG | EthProtocolManager | Disconnect - Outbound - 0x03 USELESS_PEER - 0x6a7... - 3 peers left
```

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputation.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputation.java
@@ -37,7 +37,7 @@ public class PeerReputation implements Comparable<PeerReputation> {
   static final int DEFAULT_MAX_SCORE = 200;
   // how much above the initial score you need to be to not get disconnected for timeouts/useless
   // responses
-  static final int HAS_BEEN_USEFUL_SCORE_INCREASE = 10;
+  private final int hasBeenUsefulThreshold;
   static final int DEFAULT_INITIAL_SCORE = 100;
   private static final Logger LOG = LoggerFactory.getLogger(PeerReputation.class);
   private static final int TIMEOUT_THRESHOLD = 5;
@@ -49,8 +49,6 @@ public class PeerReputation implements Comparable<PeerReputation> {
 
   private static final int SMALL_ADJUSTMENT = 1;
   private static final int LARGE_ADJUSTMENT = 5;
-
-  private final int initialScore;
   private int score;
 
   private final int maxScore;
@@ -63,7 +61,7 @@ public class PeerReputation implements Comparable<PeerReputation> {
     checkArgument(
         initialScore <= maxScore, "Initial score must be less than or equal to max score");
     this.maxScore = maxScore;
-    this.initialScore = initialScore;
+    this.hasBeenUsefulThreshold = Math.min(maxScore, initialScore + 10);
     this.score = initialScore;
   }
 
@@ -93,7 +91,7 @@ public class PeerReputation implements Comparable<PeerReputation> {
   }
 
   private boolean peerHasNotBeenUseful() {
-    return score - initialScore < HAS_BEEN_USEFUL_SCORE_INCREASE;
+    return score < hasBeenUsefulThreshold;
   }
 
   public void resetTimeoutCount(final int requestCode) {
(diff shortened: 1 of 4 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputation.java
```diff
@@ -34,19 +34,23 @@
 public class PeerReputation implements Comparable<PeerReputation> {
   static final long USELESS_RESPONSE_WINDOW_IN_MILLIS =
       TimeUnit.MILLISECONDS.convert(1, TimeUnit.MINUTES);
-  static final int DEFAULT_MAX_SCORE = 150;
+  static final int DEFAULT_MAX_SCORE = 200;
+  // how much above the initial score you need to be to not get disconnected for timeouts/useless
+  // responses
+  static final int HAS_BEEN_USEFUL_SCORE_INCREASE = 10;
```
siladu, line 40: nit/suggestion: I feel like it's less work for the reader to use an absolute value rather than an offset.

e.g. 
`static final int HAS_BEEN_USEFUL_THRESHOLD = initialScore + 10`

then...
```
  private boolean peerHasNotBeenUseful() {
    return score < HAS_BEEN_USEFUL_SCORE_THRESHOLD;
  }
```

Might need to do some jiggery pokery depending on how dynamic initial score is to make that work I guess

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputation.java
```diff
@@ -59,22 +63,37 @@ public PeerReputation(final int initialScore, final int maxScore) {
...
+        LOG.debug(
+            "Disconnection triggered by {} repeated timeouts for requestCode {}, peer score {}",
+            newTimeoutCount,
+            requestCode,
+            score);
+        return Optional.of(DisconnectReason.TIMEOUT);
+      }
+
+      LOG.trace(
+          "Not triggering disconnect for {} repeated timeouts for requestCode {} because peer has high score {}",
```
siladu, line 85: Should we adjust the score in this case? I might be mis-reading but if a previously useful peer becomes useless, would we never disconnect?

macfarla: the score still does get decremented on line 73

eg

```
2023-11-20 15:13:42.657+10:00 | nioEventLoopGroup-3-4 | TRACE | PeerReputation | Not triggering disconnect for exceeding useless response threshold because peer has high score 167
2023-11-20 15:13:42.716+10:00 | nioEventLoopGroup-3-2 | TRACE | PeerReputation | Not triggering disconnect for exceeding useless response threshold because peer has high score 116
2023-11-20 15:13:43.492+10:00 | nioEventLoopGroup-3-2 | DEBUG | PeerReputation | Disconnection triggered by exceeding useless response threshold, score 109
2023-11-20 14:28:19.456+10:00 | nioEventLoopGroup-3-5 | TRACE | PeerReputation | Not triggering disconnect for exceeding useless response threshold because peer has high score 117
2023-11-20 14:28:19.582+10:00 | nioEventLoopGroup-3-5 | TRACE | PeerReputation | Not triggering disconnect for exceeding useless response threshold because peer has high score 111
2023-11-20 14:28:22.063+10:00 | EthScheduler-Timer-0 | DEBUG | PeerReputation | Disconnection triggered by 13 repeated timeouts for requestCode 15, peer score 107
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/PeerReputation.java
macfarla: done

siladu (approved):
LGTM

fab-10:
It think that PR should go along with these two PR that switch peer on retry, otherwise it is possible that we get stuck with a useless peer
https://github.com/hyperledger/besu/pull/5508 https://github.com/hyperledger/besu/pull/5509

PR #6189, opened 2023-11-20
Update OpenJDK latest Docker image to use Java 21

fab-10:
## PR description

Switched base image to `ubuntu:rolling`, that follow any stable Ubuntu release, currently 23.10) since OpenJDK Java 21 is not available on latest LTS Ubuntu.

### DIFF
--- a/docker/openjdk-latest/Dockerfile
+++ b/docker/openjdk-latest/Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:rolling
 ARG VERSION="dev"
 
 RUN apt-get update && \
- apt-get install --no-install-recommends -q --assume-yes openjdk-21-jre-headless libjemalloc-dev adduser && \
+ apt-get install --no-install-recommends -q --assume-yes openjdk-21-jre-headless=21* libjemalloc-dev=5.* adduser=3* && \
  apt-get clean  && \
  rm -rf /var/lib/apt/lists/*  && \
  adduser --disabled-password --gecos "" --home /opt/besu besu && \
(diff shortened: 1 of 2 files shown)


docker/openjdk-latest/Dockerfile
```diff
@@ -1,9 +1,9 @@
 
-FROM ubuntu:22.04
+FROM ubuntu:rolling
 ARG VERSION="dev"
 
 RUN apt-get update && \
- apt-get install --no-install-recommends -q --assume-yes openjdk-19-jre-headless=19* libjemalloc-dev=5.* && \
+ apt-get install --no-install-recommends -q --assume-yes openjdk-21-jre-headless libjemalloc-dev adduser && \
```
jflo, line 6: just curious, what triggered the need for adduser ?

fab-10: it is no more present in the base image

docker/openjdk-latest/Dockerfile
```diff
@@ -1,9 +1,9 @@
 
-FROM ubuntu:22.04
+FROM ubuntu:rolling
```
shemnon, line 2: This will make reproducible builds difficult.

fab-10: what do you suggest to use for the base image?

fab-10: Fine for me, will set this to `23.10`, was using `rolling` to automatically switch to a supported version, but that could actually breaks the build, like it happened now with the missing `adduser`

jflo: I think i can accept that tradeoff for only this build profile.

fab-10: @shemnon @jflo are you fine with the current implementation?

shemnon (approved):

jflo (approved):

PR #6190, opened 2023-11-20
Allow a transaction selection plugin to specify custom selection results

fab-10:
## PR description

With this PR a transaction selection plugin can extend TransactionSelectionResult to implement its custom return values.
Plus a couple of log fixes.


ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
```diff
@@ -1189,4 +1192,48 @@ protected MiningParameters createMiningParameters(
                 .build())
         .build();
   }
+
+  private static class PluginTrasactionSelectionResult extends TransactionSelectionResult {
```
Gabriel-Trintinalia, line 1196: typo here -> `PluginTraNsactionSelectionResult`

Gabriel-Trintinalia (approved):
LGTM

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
fab-10: good catch, fixed

PR #6202, opened 2023-11-22
New cli options to limit rewards return by eth_feeHistory 

Gabriel-Trintinalia:

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -1218,6 +1218,26 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   private final Long apiGasPriceMax = 500_000_000_000L;
 
   @CommandLine.Option(
+      names = {"--api-priority-fee-boundary-enabled"},
+      hidden = true,
+      description = "Set to enable rewards boundary in eth_feeHistory (default: ${DEFAULT-VALUE})")
+  private final boolean apiPriorityFeeLimitingEnabled = false;
+
+  @CommandLine.Option(
+      names = {"--api-priority-fee-lower-bound-coefficient"},
+      hidden = true,
+      description =
+          "Coefficient for setting the lower limit of minimum priority fee in eth_feeHistory (default: ${DEFAULT-VALUE})")
+  private final Long apiPriorityFeeLowerBoundCoefficient = 100L;
+
+  @CommandLine.Option(
+      names = {"--api-priority-fee-upper-bound-coefficient"},
+      hidden = true,
+      description =
+          "Coefficient for setting the upper limit of minimum priority fee in eth_feeHistory (default: ${DEFAULT-VALUE})")
+  private final Long apiPriorityFeeUpperBoundCoefficient = 100L;
+
+  @CommandLine.Option(
       names = {"--static-nodes-file"},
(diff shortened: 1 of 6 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -203,7 +213,16 @@ public List<Wei> computeRewards(final List<Double> rewardPercentiles, final Bloc
...
     final List<TransactionInfo> transactionsInfo =
         generateTransactionsInfo(transactions, transactionsGasUsed, baseFee);
-    return calculateRewards(rewardPercentiles, block, transactionsInfo);
+
+    var realRewards = calculateRewards(rewardPercentiles, block, transactionsInfo);
+
+    // If the priority fee boundary is set, return the bounded rewards. Otherwise, return the real
+    // rewards.
+    if (apiConfiguration.isPriorityFeeLimitingEnabled()) {
+      return calculateBoundedRewards(realRewards);
```
macfarla, line 222:
```suggestion
      return boundRewards(realRewards);
```

macfarla: I think it's clearer if you remove 'calculate' from this method name

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -235,6 +254,46 @@ private List<Wei> calculateRewards(
     return rewards;
   }
 
+  /**
+   * This method returns a list of bounded rewards.
+   *
+   * @param rewards The list of rewards to be bounded.
+   * @return The list of bounded rewards.
+   */
+  private List<Wei> calculateBoundedRewards(final List<Wei> rewards) {
```
macfarla, line 263: will the default (no bounds) be used most of the time?

Gabriel-Trintinalia: yes

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2506,6 +2508,17 @@ && rpcWsAuthenticationCredentialsFile() == null
   }
 
   private ApiConfiguration apiConfiguration() {
+    if (!apiPriorityFeeLimitingEnabled
```
fab-10, line 2511: There is a standard way to do this check, that also works when someone passed the default value, using `CommandLineUtils::checkOptionDependencies` see how it is already used for examples. Then having a check that the lower bound is <= the upper bound will be nice too.

fab-10 (changes requested):
ok, just use the common approach to validate options dependencies

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
Gabriel-Trintinalia: done

fab-10 (approved):

PR #6205, opened 2023-11-23
fix collision issue

matkt:
## PR description

The idea of this PR is to revert some modification that was made to pass tests  https://github.com/hyperledger/besu/pull/5686.

Therefore, I decided to revert some modification on this PR by fixing the problem differently. Indeed, the idea of the collision test that does not pass is to selfdestruct a contract and recreate it afterwards. Instead of modifying the logic of the accumulator and listing all the storage leaves to be deleted during the selfdestruct, which seems to lead to edge cases, I prefer to let BonsaiWorldstate handle it as before and just load the storage with EMPTY_TRIE_HASH if we detect that it has been cleared before pushing the new slots after recreation. Indeed, the clearing of the storage that is done at the beginning of calculateRootHash removes the storage but does not change the storage root of the contract. Therefore, we falsely believe that the storage is not empty.


garyschulte (commented):
This will be faster and simpler, but at the cost of leaving dangling storage slots on self destruct right?   

This seems like a viable tradeoff especially in light of EIP-6780.  I just want to clarify

matkt:
In the case of just a selfdestruct it should be managed by that https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java#L275                                                                                                                                                                    
If the contract is recreated without storage we will only delete the flat and we will have the garbage in the tree https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/bonsai/worldview/BonsaiWorldStateUpdateAccumulator.java#L396 it is the logic that I understands

matkt:
Sync passed on goerli @garyschulte if you can review it please

garyschulte (approved):
Seems good 👍

PR #6206, opened 2023-11-23
Fix log index in transaction receipt

### LINKED ISSUE #6204: Incorrect logIndex in eth_getReceipt responses
### Description
This is a variant on https://github.com/hyperledger/besu/issues/4114 that was fixed a while back but using `eth_getReceipt` instead of `eth_getLogs`.

On Mainnet when you query besu with:
```
curl -XPOST --data '{"id":1,"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed"]}'
```

which is getting the receipt for https://etherscan.io/tx/0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed

It gives:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
    "blockNumber": "0x11c4b9a",
    "contractAddress": null,
    "cumulativeGasUsed": "0x430d0",
    "from": "0x64ad3f5944ef64df67726956d5829d90ae84c67d",
    "gasUsed": "0x2783b",
    "effectiveGasPrice": "0xb7ea063ae",
    "logs": [
      {
        "address": "0xc3960227e41c3f54e9b399ce216149dea5315c34",
        "topics": [
          "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
          "0x0000000000000000000000007a5bef8c4b1d511bc076a416b097eea798247301",
          "0x000000000000000000000000a18d5eceb5fe22ee3cc228fb9c32b65adcedfa13"
        ],
        "data": "0x000000000000000000000000000000000000000000000000000611c0e8c75821",
        "blockNumber": "0x11c4b9a",
        "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
        "transactionIndex": "0x1",
        "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
        "logIndex": "0x0",
        "removed": false
      },
      {
        "address": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
        "topics": [
          "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
          "0x000000000000000000000000a18d5eceb5fe22ee3cc228fb9c32b65adcedfa13",
          "0x000000000000000000000000166191ccb056a39585e4fc2aba1da3d76fb4f4d2"
        ],
        "data": "0x000000000000000000000000000000000000000000000000095cb3372ab4bd74",
        "blockNumber": "0x11c4b9a",
        "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
        "transactionIndex": "0x1",
        "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
        "logIndex": "0x1",
        "removed": false
      },
      {
        "address": "0xa18d5eceb5fe22ee3cc228fb9c32b65adcedfa13",
        "topics": [
          "0x1c411e9a96e071241c2f21f7726b17ae89e3cab4c78be50e062b03a9fffbbad1"
        ],
        "data": "0x000000000000000000000000000000000000000000000004268dcdfa2cce1e2600000000000000000000000000000000000000000000000002b4d8327e1c8c4f",
        "blockNumber": "0x11c4b9a",
        "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
        "transactionIndex": "0x1",
        "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
        "logIndex": "0x2",
        "removed": false
      },
      {
        "address": "0xa18d5eceb5fe22ee3cc228fb9c32b65adcedfa13",
        "topics": [
          "0xd78ad95fa46c994b6551d0da85fc275fe613ce37657fb8d5e3d130840159d822",
          "0x000000000000000000000000166191ccb056a39585e4fc2aba1da3d76fb4f4d2",
          "0x000000000000000000000000166191ccb056a39585e4fc2aba1da3d76fb4f4d2"
        ],
        "data": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000611c0e8c75821000000000000000000000000000000000000000000000000095cb3372ab4bd740000000000000000000000000000000000000000000000000000000000000000",
        "blockNumber": "0x11c4b9a",
        "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
        "transactionIndex": "0x1",
        "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
        "logIndex": "0x3",
        "removed": false
      },
      {
        "address": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
        "topics": [
          "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
          "0x000000000000000000000000166191ccb056a39585e4fc2aba1da3d76fb4f4d2",
          "0x0000000000000000000000007a5bef8c4b1d511bc076a416b097eea798247301"
        ],
        "data": "0x000000000000000000000000000000000000000000000000094048aebddc2f63",
        "blockNumber": "0x11c4b9a",
        "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
        "transactionIndex": "0x1",
"blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
        "logIndex": "0x4",
        "removed": false
      },
      {
        "address": "0x7a5bef8c4b1d511bc076a416b097eea798247301",
        "topics": [
          "0xc42079f94a6350d7e6235f29174924f928cc2ac818eb64fed8004e115fbcca67",
          "0x000000000000000000000000166191ccb056a39585e4fc2aba1da3d76fb4f4d2",
          "0x000000000000000000000000a18d5eceb5fe22ee3cc228fb9c32b65adcedfa13"
        ],
        "data": "0x000000000000000000000000000000000000000000000000094048aebddc2f63fffffffffffffffffffffffffffffffffffffffffffffffffff9ee3f1738a7df00000000000000000000000000000000000000000d0122ada13d6847cf88842f00000000000000000000000000000000000000000000000093b3abc7a6eb0507ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff172f",
        "blockNumber": "0x11c4b9a",
        "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
        "transactionIndex": "0x1",
        "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
        "logIndex": "0x5",
        "removed": false
      }
    ],
    "logsBloom": "0x00200002000000000000000080008000000000000000000000000000000000000000000000000000000000000000000002000000080020000000000000000000000000008000000800000008000000200000002000000000008000000000000000000000000000000040000020000000004002000000000000000010000800000080000000000000000000000000000000080000000001080000004100000000000000000000000000000000000000000000000000000000000000000000000000000002002000000000000000040000000000000000001000000000000000000000200000000000000000000000000000000410000000000000000001000000",
    "status": "0x1",
    "to": "0x166191ccb056a39585e4fc2aba1da3d76fb4f4d2",
    "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
    "transactionIndex": "0x1",
    "type": "0x2"
  }
}
```

The key thing being that the first log has `"logIndex": "0x0",` whereas this is the second transaction in the block and `logIndex` should be relative to the block, not the transaction.  Geth reports the first log entry as index `0x3`:

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
    "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
    "blockNumber": "0x11c4b9a",
    "logs": [
      {
        "transactionHash": "0x44b738650c4686cbda63061f4b5304ebc0efc56eb23b53f4a27424df454e76ed",
        "address": "0xc3960227e41c3f54e9b399ce216149dea5315c34",
        "blockHash": "0x22bc2e6d0767ce30d93dec83f122416e31e28af41a4ecf649819e6987c04dff8",
        "blockNumber": "0x11c4b9a",
        "data": "0x000000000000000000000000000000000000000000000000000611c0e8c75821",
        "logIndex": "0x3",
        "removed": false,
....
```

Wetitpig:
## PR description
Assign log index to transaction receipt according to position in block instead of in transaction.

## Fixed Issue(s)
fixes #6204

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
@@ -624,17 +624,16 @@ public class BlockchainQueries {
     // getTransactionLocation should not return if the TX or block doesn't exist, so throwing
     // on a missing optional is appropriate.
     final TransactionLocation location = maybeLocation.get();
-    final Block block = blockchain.getBlockByHash(location.getBlockHash()).orElseThrow();
-    final List<Transaction> transactions = block.getBody().getTransactions();
-    final Transaction transaction = transactions.get(location.getTransactionIndex());
-
     final Hash blockhash = location.getBlockHash();
     final int transactionIndex = location.getTransactionIndex();
+
+    final Block block = blockchain.getBlockByHash(blockhash).orElseThrow();
+    final Transaction transaction = block.getBody().getTransactions().get(transactionIndex);
+
     final BlockHeader header = block.getHeader();
     final List<TransactionReceipt> transactionReceipts =
         blockchain.getTxReceipts(blockhash).orElseThrow();
-    final TransactionReceipt transactionReceipt =
-        transactionReceipts.get(location.getTransactionIndex());
+    final TransactionReceipt transactionReceipt = transactionReceipts.get(transactionIndex);
 
     long gasUsed = transactionReceipt.getCumulativeGasUsed();
     int logIndexOffset = 0;
@@ -658,7 +657,7 @@ public class BlockchainQueries {
             transactionReceipt,
             transaction,
             transactionHash,
-            location.getTransactionIndex(),
+            transactionIndex,
             gasUsed,
             header.getBaseFee(),
             blockhash,
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
@@ -629,6 +629,7 @@ public class BlockchainQueries {
     final Transaction transaction = transactions.get(location.getTransactionIndex());
 
     final Hash blockhash = location.getBlockHash();
+    final int transactionIndex = location.getTransactionIndex();
     final BlockHeader header = block.getHeader();
     final List<TransactionReceipt> transactionReceipts =
         blockchain.getTxReceipts(blockhash).orElseThrow();
@@ -636,10 +637,14 @@ public class BlockchainQueries {
         transactionReceipts.get(location.getTransactionIndex());
 
     long gasUsed = transactionReceipt.getCumulativeGasUsed();
-    if (location.getTransactionIndex() > 0) {
-      gasUsed =
-          gasUsed
-              - transactionReceipts.get(location.getTransactionIndex() - 1).getCumulativeGasUsed();
+    int logIndexOffset = 0;
+    if (transactionIndex > 0) {
+      gasUsed = gasUsed - transactionReceipts.get(transactionIndex - 1).getCumulativeGasUsed();
+      logIndexOffset =
+          IntStream.range(0, transactionIndex)
+              .parallel()
+              .map(i -> transactionReceipts.get(i).getLogs().size())
+              .sum();
     }
 
     Optional<Long> maybeBlobGasUsed =
@@ -647,8 +652,6 @@ public class BlockchainQueries {
 
     Optional<Wei> maybeBlobGasPrice =
         getBlobGasPrice(transaction, header, protocolSchedule.getByBlockHeader(header));
-
-    final int logIndexOffset = logIndexOffset(transactionHash, transactionReceipts, transactions);
 
     return Optional.of(
         TransactionReceiptWithMetadata.create(
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
@@ -641,7 +641,6 @@ public class BlockchainQueries {
       gasUsed -= transactionReceipts.get(transactionIndex - 1).getCumulativeGasUsed();
       logIndexOffset =
           IntStream.range(0, transactionIndex)
-              .parallel()
               .map(i -> transactionReceipts.get(i).getLogs().size())
               .sum();
     }
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
@@ -625,8 +625,8 @@ public class BlockchainQueries {
     // on a missing optional is appropriate.
     final TransactionLocation location = maybeLocation.get();
     final Block block = blockchain.getBlockByHash(location.getBlockHash()).orElseThrow();
-    final Transaction transaction =
-        block.getBody().getTransactions().get(location.getTransactionIndex());
+    final List<Transaction> transactions = block.getBody().getTransactions();
+    final Transaction transaction = transactions.get(location.getTransactionIndex());
 
     final Hash blockhash = location.getBlockHash();
     final BlockHeader header = block.getHeader();
@@ -648,6 +648,8 @@ public class BlockchainQueries {
     Optional<Wei> maybeBlobGasPrice =
         getBlobGasPrice(transaction, header, protocolSchedule.getByBlockHeader(header));
 
+    final int logIndexOffset = logIndexOffset(transactionHash, transactionReceipts, transactions);
+
     return Optional.of(
         TransactionReceiptWithMetadata.create(
             transactionReceipt,
@@ -659,7 +661,8 @@ public class BlockchainQueries {
             blockhash,
             header.getNumber(),
             maybeBlobGasUsed,
-            maybeBlobGasPrice));
+            maybeBlobGasPrice,
+            logIndexOffset));
   }
 
   /**
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
@@ -183,8 +183,7 @@ public class BlockchainQueries {
    */
   public Optional<UInt256> storageAt(
       final Address address, final UInt256 storageIndex, final long blockNumber) {
-    final Hash blockHash =
-        getBlockHeaderByNumber(blockNumber).map(BlockHeader::getHash).orElse(Hash.EMPTY);
+    final Hash blockHash = getBlockHashByNumber(blockNumber).orElse(Hash.EMPTY);
 
     return storageAt(address, storageIndex, blockHash);
   }
@@ -211,8 +210,7 @@ public class BlockchainQueries {
    * @return The balance of the account in Wei.
    */
   public Optional<Wei> accountBalance(final Address address, final long blockNumber) {
-    final Hash blockHash =
-        getBlockHeaderByNumber(blockNumber).map(BlockHeader::getHash).orElse(Hash.EMPTY);
+    final Hash blockHash = getBlockHashByNumber(blockNumber).orElse(Hash.EMPTY);
 
     return accountBalance(address, blockHash);
   }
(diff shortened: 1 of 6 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -625,8 +625,8 @@ public Optional<TransactionReceiptWithMetadata> transactionReceiptByTransactionH
     // on a missing optional is appropriate.
     final TransactionLocation location = maybeLocation.get();
     final Block block = blockchain.getBlockByHash(location.getBlockHash()).orElseThrow();
-    final Transaction transaction =
-        block.getBody().getTransactions().get(location.getTransactionIndex());
+    final List<Transaction> transactions = block.getBody().getTransactions();
+    final Transaction transaction = transactions.get(location.getTransactionIndex());
```
shemnon, line 629: As a "fix it now" patch this seems like a reasonable solution.  However there will be performance implications with large blocks.  I think a longer term solution would be to add the log index into the `TransactionLocation` and save that when calculating the location.  The data can be written so the value may be missing, and when it is we use this logic.  Creating a seamless transition.  We also could add a DB migration to re-build the transaction locations in the background.

Wetitpig: How to initiate re-building transaction locations?

shemnon: @fab-10

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -624,22 +624,26 @@ public Optional<TransactionReceiptWithMetadata> transactionReceiptByTransactionH
...
-          gasUsed
-              - transactionReceipts.get(location.getTransactionIndex() - 1).getCumulativeGasUsed();
+    int logIndexOffset = 0;
+    if (transactionIndex > 0) {
+      gasUsed -= transactionReceipts.get(transactionIndex - 1).getCumulativeGasUsed();
+      logIndexOffset =
+          IntStream.range(0, transactionIndex)
+              .parallel()
+              .map(i -> transactionReceipts.get(i).getLogs().size())
+              .sum();
```
fab-10, line 646: since usually we only have few hundreds txs in a block, have you any number to confirm that doing the sum in parallel is faster than doing it serially?

fab-10 (commented):
LGTM, just a check on the use of parallel stream or not for the sum

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
fab-10: @Wetitpig `migrateVariables` was much easier to implement, since it only had to move few keys around, and did it synchronously, while for this re-build an asynchronous tasks is needed, with all the consequences of making it not hurting the overall Besu performance. 

My proposal for this is to:
1.  first support saving the location for newly imported blocks
2.  then for already imported blocks, that miss this data, calculate it and save on demand

Have not yet looking into the code, to understand how much it, but would like to hear what you think.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
Wetitpig: Not really any numbers there. I am just thinking that it does not hurt to map and sum the numbers in parallel.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
Wetitpig: Given that most transaction receipts may not be queried for multiple times I am not sure if it is worth to save the `logIndexOffset` for each transaction.
Probably I am going to try removing `parallel()` and see how much the performance impact really is.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
fab-10: just thinking that a quick JMH test could be enough to spot the difference in performance, between serial/parallel

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
fab-10: if calculating that data at runtime is good enough for the moment, we can postpone the work to persist that data until more performance is needed.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
Wetitpig: I have no idea how to do a JMH benchmark, but I tried the query with `time curl` and deducting the `user` and `sys` time
```GraphQL
query {
	blocks (from: 18676591 to: 18676640) {
		transactions { gasUsed }
	}
}
```
Every time GraphQL transverses into a transaction it would first obtain the receipt for the amount of gas used, thus running the code for obtaining log indices in the receipt.

So with 8235 transactions in these 50 blocks the results are as follows:
|`.parallel()`|Time averaged over 10 queries|
|-----------------|--------:|
| with | 8.5605 s |
| without | 8.6308 s |

So it may not be worth the hassle to parallelise

fab-10: Ok, indeed for small lists there should not be a big difference.
Just for reference, an example of a JMH test in Besu is [RLPBench](https://github.com/hyperledger/besu/blob/8f5a4443cc2e511d0df6eab5dca2723d104a46d7/ethereum/rlp/src/jmh/java/org/hyperledger/besu/ethereum/rlp/RLPBench.java#L31)

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -263,13 +260,7 @@ public Optional<Integer> getTransactionCount(final long blockNumber) {
...
       return Optional.empty();
     }
-    return Optional.of(
-        blockchain
-            .getBlockHashByNumber(blockNumber)
-            .flatMap(this::blockByHashWithTxHashes)
-            .map(BlockWithMetadata::getTransactions)
-            .map(List::size)
-            .orElse(-1));
+    return blockchain.getBlockHashByNumber(blockNumber).map(this::getTransactionCount);
```
fab-10, line 263: looks better now!

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -1079,7 +1070,7 @@ private int logIndexOffset(
         break;
       }
 
-      logIndexOffset += receipts.get(i).getLogs().size();
+      logIndexOffset += receipts.get(i).getLogsList().size();
```
fab-10, line 1073: nice opt

fab-10 (approved):

PR #6209, opened 2023-11-28
worldstate refactor

matkt:
## PR description

for verkle, I’m starting a refactor. In simple terms, many classes can distinguish whether they’re in ‘bonsai’ or ‘forest’ mode, so a generic interface doesn’t offer much value, except for some minor and synchronization parts.
So, I’ve developed a ’WorldstatestorageCoordinator class. This will have a storage strategy, which could be either ‘bonsai’ or ‘forest’. This coordinator is needed when we’re unsure about the storage mode we use, and it helps for the routing process. when we’re sure about the storage mode, we directly use the appropriate strategy.
This approach reduces the number of generic interfaces, eliminating locations that are not applicable to modes like ‘forest’ or other not clean things like that.
Will be better when we will have to add some custom method or field for verkle

### Test done on holesky
|         | Checkpoint with trielog and blockchain pruning | Snapsync with codeByhash |
|---------|--------|------------|
| Bonsai  |     OK   |    OK | 

### Test done on sepolia
|         | Full   | Checkpoint | Full with Full flat DB |
|---------|--------|------------|------------|
| Bonsai  |     OK   |     OK      |?? | 
| Forest  |       ??  |     OK        | not needed |

### Test done on mainnet
|         | Checkpoint with Full flat DB |
|---------|--------|
| Bonsai  |  OK   | 

### Test done on QBFT network 
Deploy contract, read storage , selfdestruct, block creation  and block import


ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPOutput.java
```diff
@@ -92,7 +92,7 @@ default void writeUInt256Scalar(final UInt256Value<?> v) {
   /**
    * Writes a RLP "null", that is an empty value.
    *
-   * <p>This is a shortcut for {@code writeBytes(Bytes.EMPTY)}.
+   * <p>This is a shortcut for writeUInt256Scalar{@code writeBytes(Bytes.EMPTY)}.
```
gfukushima, line 95: was this meant to here?

gfukushima (approved):
lgtm with a few minor questions

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/TrieNodeHealingRequest.java
```diff
@@ -172,19 +181,21 @@
   }
 
   public abstract Optional<Bytes> getExistingData(
-      final SnapWorldDownloadState downloadState, final WorldStateStorage worldStateStorage);
+      final SnapWorldDownloadState downloadState,
```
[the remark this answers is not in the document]
matkt: will remove it thanks

besu/src/main/java/org/hyperledger/besu/cli/subcommands/operator/BackupState.java
```diff
@@ -81,15 +81,19 @@ public void run() {
 
     final BesuController besuController = createBesuController();
     final MutableBlockchain blockchain = besuController.getProtocolContext().getBlockchain();
-    final WorldStateStorage worldStateStorage =
+    final ForestWorldStateKeyValueStorage worldStateKeyValueStorage =
```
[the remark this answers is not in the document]
matkt: modified the name

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -536,17 +540,17 @@ public Stream<StreamableAccount> streamAccounts(final Bytes32 startKeyHash, fina
...
         .orElse(null);
   }
 
   protected Optional<Bytes> getAccountStateTrieNode(final Bytes location, final Bytes32 nodeHash) {
-    return worldStateStorage.getAccountStateTrieNode(location, nodeHash);
+    return worldStateKeyValueStorage.getAccountStateTrieNode(location, nodeHash);
   }
 
-  private void writeTrieNode(
+  protected void writeTrieNode(
```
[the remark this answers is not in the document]
matkt: nop , it's now fixed

ethereum/rlp/src/main/java/org/hyperledger/besu/ethereum/rlp/RLPOutput.java
matkt: I think it's a mistake, i removed that

matkt: I think it's a mistake , I removed that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/forest/pruner/Pruner.java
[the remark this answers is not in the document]
matkt: yes I revert this modification

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/StateBackupService.java
```diff
@@ -84,12 +84,12 @@ public StateBackupService(
       final Blockchain blockchain,
       final Path backupDir,
       final EthScheduler scheduler,
-      final WorldStateStorage worldStateStorage) {
```
garyschulte, line 87: 🔪 💀

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/forest/pruner/MarkSweepPruner.java
```diff
@@ -58,7 +58,7 @@ public class MarkSweepPruner {
   private static final int MAX_MARKING_THREAD_POOL_SIZE = 2;
 
   private final int operationsPerTransaction;
-  private final WorldStateStorage worldStateStorage;
+  private final ForestWorldStateKeyValueStorage worldStateKeyValueStorage;
```
garyschulte, line 61: ❤️

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/forest/worldview/ForestMutableWorldState.java
```diff
@@ -63,18 +64,22 @@ public class ForestMutableWorldState implements MutableWorldState {
...
   }
 
   public ForestMutableWorldState(
       final Bytes32 rootHash,
-      final WorldStateStorage worldStateStorage,
+      final WorldStateKeyValueStorage worldStateKeyValueStorage,
       final WorldStatePreimageStorage preimageStorage,
       final EvmConfiguration evmConfiguration) {
-    this.worldStateStorage = worldStateStorage;
+    this.worldStateKeyValueStorage = (ForestWorldStateKeyValueStorage) worldStateKeyValueStorage;
```
garyschulte, line 82: we can probably get away from this explicit cast by changing the constructor signature.   It looks like only evmtool BlockchainModule is calling this from a context that doesn't explicitly have storage as a  ForestWorldStateKeyValueStorage.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateStorageCoordinator.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+
+public class WorldStateStorageCoordinator {
+  private final WorldStateKeyValueStorage worldStateKeyValueStorage;
+
+  public WorldStateStorageCoordinator(final WorldStateKeyValueStorage worldStateKeyValueStorage) {
```
garyschulte, line 31: 👍  to the coordinator concept.  I like that this is the central place to add/remove storage formats and their interfaces.

the only bits I do not care for are where we double-down on forest and bonsai storage modes in the signatures, the apply and consume ForStrategy methods basically.  I think we could do some lambda magic to limit the bonsai/forest stuff in these methods, like: 

```
public <RESPONSE, T extends WorldStateKeyValueStorage> RESPONSE applyForStrategy(
      final Function<T, RESPONSE> strategy) {
 ```

File: ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateStorageCoordinator.java [hunk 9]
```diff
@@ -0,0 +1,154 @@
...
+
+  public void applyOnMatchingStrategy(
+      final DataStorageFormat dataStorageFormat,
+      final Consumer<WorldStateKeyValueStorage> onStrategy) {
+    if (getDataStorageFormat().equals(dataStorageFormat)) {
+      onStrategy.accept(worldStateKeyValueStorage());
+    }
+  }
+
+  public <RESPONSE> RESPONSE applyForStrategy(
```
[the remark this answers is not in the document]
garyschulte: see comment about how to remove this via generics and/or WorldStateKeyValueStorage sub-interfaces like PathBasedWorldStorage, HashBasedWorldStorage, etc.

garyschulte (commented):
Good refactor with WorldStateStorageCoordinator.  I think we should take the opportunity to get rid of some stuff rather than perpetuate it (like the state backup stuff), but that can be in another PR (before or after this one).  

Only gripe is in the coordinator where we unnecessarily perpetuate each storage format and its worldstate as a part of the method signature

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateStorageCoordinator.java
matkt: Why I am adding Bonsai and Forest in the signature is because I no longer want a common interface between Bonsai and Forest. The coordinator will play the role of redirection and do the cast to simplify other calls. So, as I don't have a common interface, I can no longer create a single strategy interface that I pass to the coordinator. I must therefore pass both and the coordinator calls the right one. I can do that but I will have to add the cast logic in other calls and in order to simplify this call I will add complexity to the other calls 

If I recreate an interface, it's doing what we had before and I specifically wanted to avoid that. Knowing that Forest will be eliminated soon, this should also remove the coordinator and this mechanism. But maybe I didn't understand your idea.

jframe:
Don't see anything to worry about in terms of impact on private networks

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateStorageCoordinator.java
garyschulte: This is what I was suggesting:

https://github.com/garyschulte/besu/commit/aab618747dfa61abbe9360de6703991e98366cb1

I didn't remove all the applyWhen, consumeFor, etc but I think we can simplify the coordinator a lot this way

matkt: could you confirm on this PR that it's ok for you to merge with the current implementation ?

garyschulte: Confirmed 👍

garyschulte (approved):
🚢

gfukushima (commented):
left a few nits and non-blocker comments, lgtm in general and it is nice to avoid the nulls we use to pass to match the general storage format interface. This approach seems to be friendly and extensible to add verkle

gfukushima:
Took a stab fixing the only unit test broken in this PR, please have a look if it looks good to you.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -84,28 +84,30 @@ public class BonsaiWorldState
...
       final CachedWorldStorageManager cachedWorldStorageManager,
       final TrieLogManager trieLogManager,
       final EvmConfiguration evmConfiguration) {
-    this.worldStateStorage = worldStateStorage;
+    this.worldStateKeyValueStorage = worldStateKeyValueStorage;
     this.worldStateRootHash =
         Hash.wrap(
-            Bytes32.wrap(worldStateStorage.getWorldStateRootHash().orElse(Hash.EMPTY_TRIE_HASH)));
+            Bytes32.wrap(
+                worldStateKeyValueStorage.getWorldStateRootHash().orElse(getEmptyTrieHash())));
```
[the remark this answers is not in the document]
matkt: it's for verkle , because we will override the value . with verkle empty trie hash is different

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/AccountTrieNodeHealingRequest.java
```diff
@@ -56,22 +56,29 @@ public AccountTrieNodeHealingRequest(
...
+        },
+        onForest -> {
+          onForest.putAccountStateTrieNode(getNodeHash(), data);
+        });
     return 1;
   }
 
   @Override
   public Optional<Bytes> getExistingData(
-      final SnapWorldDownloadState downloadState, final WorldStateStorage worldStateStorage) {
```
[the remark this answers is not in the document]
matkt: I think it was just useless parameter that we forgot to remove. I checked and it was not used

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/forest/pruner/Pruner.java
```diff
@@ -51,7 +51,7 @@ public class Pruner {
   private final ExecutorService executorService;
 
   @VisibleForTesting
-  Pruner(
+  public Pruner(
```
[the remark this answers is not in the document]
matkt: will fix it thanks

PR #6215, opened 2023-11-29
Unpin ca-certificates-java version

shemnon:
## PR description

CI builds are currently broken because an ubuntu package update breaks with ca-certificates-java version 20190909. If we allow it to use updates to the version (such as 20190909ubuntu1.2) the build un-breaks.


garyschulte (approved):
LGTM

PR #6218, opened 2023-11-30
Allow `maxActiveConnections` for GraphQL and terminate requests based on actual connection status

Wetitpig:
## PR description

1. The option of `--graphql-http-max-active-connections` has been added to limit the numer of GraphQL requests served each time.
2. The GraphQL server has been relying on an `isAliveHandler` to terminate requests based on the timeout. However, this handler was only applied to `matchingLogs`, resulting in unterminated GraphQL executions in general even if the connection has timed out. Rather than relying on a future task, it is more appropriate to terminate requests based on the actual status of the connection. The `TimeoutHandler` should have handled the HTTP timeout while connections may also be terminated due to exceeding `maxActiveConnections`.

Docs change would be added in another PR in a later stage when code changes stabilise.

## Fixed Issue(s)


Wetitpig:
There is also a refactoring of the GraphQL tests to a base class with all server start and end methods.

Wetitpig:
Quite a number of methods common to all HTTP services (CORS checking, Hostlist checking etc.) were relocated to reuse code.
A parent class `HttpConfiguration` has been created. Are there suggestions to a better name?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/HttpConfiguration.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla, line 2:
```suggestion
 * Copyright contributors to Hyperledger Besu.
```

gradle/versions.gradle
```diff
@@ -108,7 +108,7 @@ dependencyManagement {
 
     dependency 'io.reactivex.rxjava2:rxjava:2.2.21'
 
-    dependencySet(group: 'io.vertx', version: '4.3.5') {
+    dependencySet(group: 'io.vertx', version: '4.5.0') {
```
macfarla, line 111: think this upgrade would be better as a separate PR

besu/src/main/java/org/hyperledger/besu/cli/custom/CorsAllowedOriginsProperty.java
```diff
@@ -85,14 +82,6 @@ public boolean addAll(final Collection<? extends String> collection) {
       if (domains.size() > 1) {
         throw new IllegalArgumentException("Values '*' or 'all' can't be used with other domains");
       }
-    } else {
```
macfarla, line 88: why remove this?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/HttpConfiguration.java
macfarla: new files should be copyright to besu. no need to update copyright on existing files

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/HttpConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+package org.hyperledger.besu.ethereum.api;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class HttpConfiguration {
```
macfarla, line 24: this refactor could be a separate PR

macfarla (commented):
thanks for your contrib @Wetitpig - added some comments. I think overall some good refactoring but it would be much easier to review if you could break it up into separate PRs. Might also help narrow down what's causing the acceptance test failures.

besu/src/main/java/org/hyperledger/besu/cli/custom/CorsAllowedOriginsProperty.java
Wetitpig: Because the origin would not be treated as a regex anymore. Instead it would be compared as a string literal as in `addOrigin` in vertx 4.5.0.

macfarla (changes requested):
@Wetitpig there's some good refactoring in here, great to see. If you could separate this out into some smaller PRs that are more focused, that would be much easier for us to review.

shemnon (commented):
I concur with @macfarla, I see three smaller PRs fighting for space: upgrading Vertx, refactoring the HTTP config, and then terminating GraphQL based on connection status.  The latter two has a viable argument to be in the same, but for stability and refactoring VertX should be solo, in case it impacts any other services using the library.   Makes it easier to cherry pick out and differentially test it.

macfarla (commented):
@Wetitpig apologies for the delay in reviewing. overall looks good, have added some comments for your review

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
```diff
@@ -841,7 +841,7 @@ public List<LogWithMetadata> matchingLogs(
                           transactions.get(i).getHash(),
                           i,
                           removed);
-                  logIndexOffset.addAndGet(receipts.get(i).getLogs().size());
+                  logIndexOffset.addAndGet(receipts.get(i).getLogsList().size());
```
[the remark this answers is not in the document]
Wetitpig: This change should have been part of #6206 but was missed out by that PR.

I will try to move it to another PR later.

macfarla:
@Wetitpig are you still engaged on this PR? it's marked as stale and will be closed in 2 weeks if no activity

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueries.java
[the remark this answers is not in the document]
Wetitpig: Commit deleted

macfarla:
ok I just checked out your branch to review in more detail and it doesn't compile. Are you running gradlew compile locally? If so, do you have local uncommitted changes? Let me know when you've fixed the compile errors and I can take another look. for now I'm changing to draft since it isn't ready for review

macfarla (commented):
@Wetitpig checking in on this - are you planning to continue work on this PR?

Wetitpig:
Yes... I am going to resume on this PR but probably earliest in mid-March. Thank you.

PR #6230, opened 2023-12-04
Deprecation warning if Forest pruning is enabled

fab-10:
## PR description

Forest pruning, option `pruning-enabled`, it is not maintained and this feature is superseded by Bonsai, no usage is reported also because to save disk space Bonsai is the solution and when using Forest usually the user want to keep all the history.
The proposal is to deprecated it now and remove it in few months, target 24.4 release


Gabriel-Trintinalia (approved):
LGTM

PR #6241, opened 2023-12-05
Fix the annoying "Errors occurred while build effective model" during…

fab-10:
… builds

## PR description

Found a fix for the annoying build error 
```
Errors occurred while build effective model from [...]\.gradle\caches\modules-2\files-2.1\org.eclipse.platform\org.eclipse.swt\3.124.200\2013830105b55e3a9490b38034aef3c6f6c5862e\org.eclipse.swt-3.124.200.pom:
    'dependencies.dependency.artifactId' for org.eclipse.platform:org.eclipse.swt.${osgi.platform}:jar with value 'org.eclipse.swt.${osgi.platform}' does not match a valid id pattern. in org.eclipse.platform:org.eclipse.swt:3.124.200
```

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -15,6 +15,7 @@
 
 dependencyManagement {
   dependencies {
+    applyMavenExclusions = false
 
     dependencySet(group: 'org.antlr', version: '4.11.1') {
       entry 'antlr4'


macfarla (approved):
YAY

PR #6243, opened 2023-12-06
[RPC] Use apiConfiguration to limit gasPrice in eth_getGasPrice

Gabriel-Trintinalia:
## PR description
Use apiConfiguration to limit gasPrice in eth_getGasPrice


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
```diff
@@ -279,19 +283,9 @@ private List<Wei> boundRewards(final List<Wei> rewards) {
...
-    // If the reward is greater than the upper bound, return the upper bound.
-    if (reward.compareTo(upperBound) > 0) {
-      return upperBound;
-    }
-
-    // If the reward is within the bounds, return the reward as is.
-    return reward;
+    return reward.compareTo(lowerBound) <= 0
+        ? lowerBound
+        : reward.compareTo(upperBound) > 0 ? upperBound : reward;
```
macfarla, line 288:
```suggestion
        : (reward.compareTo(upperBound) > 0) ? upperBound : reward;
```

macfarla: nit: this same logic in EthGasPrice has an extra set of () - idk which is easier to read but they should match

macfarla: another nit is that line 286 uses <= in the comparison but 288 uses strictly > but >= would have the same effect and would be more symmetrical

macfarla (approved):
couple of nits, otherwise LGTM

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java
Gabriel-Trintinalia: done

PR #6244, opened 2023-12-06
Run ATs sequentially

gfukushima:
## PR description
This PR makes the ATs to run sequentially so we reduce the parallelism and get the build working again. This might result in a longer time for ATs to complete but right now Unit tests still take longer.


macfarla (approved):
LGTM - will be interesting to see if this has any effect on concurrency limits

PR #6267, opened 2023-12-08
ETC mainnet 'Spiral' activation block

diega:
## PR description
This PR sets ETC Mainnet activation block as defined in [ECIP-1109](https://ecips.ethereumclassic.org/ECIPs/ecip-1109) to `19_250_000`. This change was already merged in core-geth in https://github.com/etclabscore/core-geth/pull/595
I also set the DNS discovery tree to what's in [core-geth#bootnodes_classic.go](https://github.com/etclabscore/core-geth/blob/v1.12.17/params/bootnodes_classic.go#L28)

### DIFF
--- a/config/src/main/resources/classic.json
+++ b/config/src/main/resources/classic.json
@@ -13,6 +13,7 @@
     "thanosBlock": 11700000,
     "magnetoBlock": 13189133,
     "mystiqueBlock": 14525000,
+    "spiralBlock": 19250000,
     "ethash": {
 
     },
--- a/config/src/main/resources/classic.json
+++ b/config/src/main/resources/classic.json
@@ -17,6 +17,7 @@
 
     },
     "discovery" : {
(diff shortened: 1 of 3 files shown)


gfukushima (approved):
lgtm

config/src/main/resources/classic.json
```diff
@@ -13,10 +13,12 @@
     "thanosBlock": 11700000,
     "magnetoBlock": 13189133,
     "mystiqueBlock": 14525000,
+    "spiralBlock": 19250000,
     "ethash": {
 
     },
     "discovery" : {
+      "dns": "enrtree://AJE62Q4DUX4QMMXEHCSSCSC65TDHZYSMONSD64P3WULVLSF6MRQ3K@all.classic.blockd.info",
```
gfukushima, line 21: Do you think this deserve an entry in the changelog too?

diega: I think we're ok without it. Hopefully there should be no difference for the users, I'm not against adding it though

Issue #6211, opened 2023-11-29
Add a new synonym for the `legacy` transaction pool


matthew1001:
### Description
As an enterprise user planning to set up a permissioned chain, I want to be able to choose a TX pool that's more suitable to my requirements, and isn't called `legacy` which implies it might be deprecated.

### Acceptance Criteria
A new transaction pool name for what is currently called the `legacy` pool is decided, and either `legacy` or the new name can be set as a value for the `--tx-pool` configuration parameter.

For maintainers and community members to vote on the most suitable term for the pool, I've added a comment for each option that was discussed in the 1[1/28/2023 EMEA/AMER contributor call](https://wiki.hyperledger.org/display/BESU/2023-11-28+Contributor+Call), with a short explanation. People can vote with a thumbs-up on the one they are most happy with and after a week or two I'll raise a PR with the new name.

**Note:** I've included a couple of "anti-suggestions" so people can see some of the options that were generally ruled out in the contributor call discussion. Thumbs-up on this will be taken to mean people agree that they are not suitable.

I've started the ball rolling with a vote for `sequenced`.

matthew1001:
**Suggestion 1:** `sequenced`

**Reasoning:** The `layered` pool primarily describes its implementation, as opposed to any particular use-case, so `sequenced` would be following that precedent. Since the legacy pool has had changes recently to make it more FIFO-like in its behaviour (see for example https://github.com/hyperledger/besu/pull/6106 / https://github.com/hyperledger/besu/pull/6022) and thus a little "fairer" from the point of view of individual senders, `sequenced` describes a little about its implementation and about how TX selection is done.

matthew1001:
**Suggestion 2:** `flat`

**Reasoning:** Similar to comment 1, `flat` describes a little about its implementation without being specific about any particular use-case.

matthew1001:
**Anti-Suggestion:** `enterprise`

**Reasoning for avoiding:** It implies that the `layered` pool isn't suitable for enterprise use-cases, which is not the case.

matthew1001:
**Anti-Suggestion 2:** `trusted`

**Reason for avoiding:** It implies too much about permissioning-like features.

fab-10:
`sequenced` (or `FIFO`) works well in the case it is used in a _gas free_ network, less if there is a gas market, in which case the main order is dictated by the gas price. 
[...]

PR #6274, opened 2023-12-11
Sequenced pool synonym for legacy pool

matthew1001:
## PR description
Introduces the new `sequenced` synonym for the `legacy` transaction pool, as discussed in https://github.com/hyperledger/besu/issues/6211.

The PR adds enough for users to start switching to the `--tx-pool=sequenced` option ahead of removal of the `legacy` pool in the future.

I've added one or two additional unit tests to check that `legacy` options still work with the `sequenced` pool, but haven't done anything more comprehensive than that because the future PR to ~~deprecate~~ remove the `legacy` pool will refactor all of the existing tests to use `sequenced` anyway.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/6211


matthew1001:
Re. docs, I think the https://besu.hyperledger.org/public-networks/concepts/transactions/pool topic probably ought to have a new sub-heading for the sequenced pool. We didn't have one before for the legacy pool but with recent changes to the legacy pool we should try to describe the use-cases the sequenced pool is aimed at and how it differs from the layered pool.

I'd be happy to provide some words to help with that docs topic.

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -295,14 +296,14 @@ public void validate(
...
+        "Could not use legacy or sequenced transaction pool options with layered implementation",
         !txPoolImplementation.equals(LAYERED),
-        CommandLineUtils.getCLIOptionNames(Legacy.class));
+        CommandLineUtils.getCLIOptionNames(Sequenced.class));
 
     CommandLineUtils.failIfOptionDoesntMeetRequirement(
         commandLine,
-        "Could not use layered transaction pool options with legacy implementation",
-        !txPoolImplementation.equals(LEGACY),
+        "Could not use layered transaction pool options with legacy or sequenced implementation",
```
macfarla, line 305:
```suggestion
        "Could not use layered transaction pool options with legacy (sequenced) implementation",
```

macfarla (approved):
LGTM . non-blocking suggestion re warning message text - just to make it clearer that they are not separate implementations

Also prob need a changelog entry

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
matthew1001: Yeah it's difficult balancing the fact that we don't want to make the sequenced pool sound like it's the "legacy" option, even though in reality they are the same implementation. At some point the legacy option will be removed entirely so we'll update these messages to just talk about sequenced in the future. Personally I'd prefer to leave it as "legacy or sequenced" for now, but happy to discuss further and we can always raise another PR to re-word if we need to?

matthew1001:
> LGTM . non-blocking suggestion re warning message text - just to make it clearer that they are not separate implementations
> 
> Also prob need a changelog entry

Thanks for the comments @macfarla. I've added a reply about the warning message wording. Re. change log entry, I've added a one-liner to it already. Did you want anything more substantial, like Fabio did for the layered pool in `23.10.0`?

Issue #6250, opened 2023-12-06
CircleCi AT's failing consistently


gfukushima:
Currently failing at the AT(mainnet) step:
ECP256R1AcceptanceTest > transactionShouldBeSuccessful FAILED

with an: Infrastructure Fail -> CONCURRENCY LIMIT.

PR #6278, opened 2023-12-11
bulk update of ATs to junit 5

macfarla:
test to see if updating all at once avoids the gradle error encountered in #6262


macfarla:
bulk update was not helpful. Doing this piecemeal. see #6250 #6362 (but there's more)

Issue #6201, opened 2023-11-22
Terminating connection duplicated


macfarla:
Terminating connection - there's (not always but a lot of the time) - 2 lines for each disconnect, with different reasons

eg

```
2023-11-22 14:27:22.907+10:00 | nioEventLoopGroup-3-7 | DEBUG | AbstractPeerConnection | Terminating connection [Connection with hashCode 933831408 inboundInitiated? false initAt 1700627242579], reason 0x01 TCP_SUBSYSTEM_ERROR
2023-11-22 14:27:22.907+10:00 | nioEventLoopGroup-3-7 | DEBUG | AbstractPeerConnection | Terminating connection [Connection with hashCode 933831408 inboundInitiated? false initAt 1700627242579], reason 0x10 SUBPROTOCOL_TRIGGERED
2023-11-22 14:27:22.952+10:00 | nioEventLoopGroup-3-2 | DEBUG | AbstractPeerConnection | Terminating connection [Connection with hashCode 241677894 inboundInitiated? false initAt 1700627242620], reason 0x01 TCP_SUBSYSTEM_ERROR
2023-11-22 14:27:22.952+10:00 | nioEventLoopGroup-3-2 | DEBUG | AbstractPeerConnection | Terminating connection [Connection with hashCode 241677894 inboundInitiated? false initAt 1700627242620], reason 0x UNKNOWN
2023-11-22 14:27:22.957+10:00 | nioEventLoopGroup-3-5 | DEBUG | AbstractPeerConnection | Terminating connection [Connection with hashCode 2120634186 inboundInitiated? false initAt 1700627242619], reason 0x01 TCP_SUBSYSTEM_ERROR
2023-11-22 14:27:22.957+10:00 | nioEventLoopGroup-3-5 | DEBUG | AbstractPeerConnection | Terminating connection [Connection with hashCode 2120634186 inboundInitiated? false initAt 1700627242619], reason 0x03 USELESS_PEER
```

an example where there's only one line

```
2023-11-22 14:27:29.089+10:00 | nioEventLoopGroup-3-7 | DEBUG | AbstractPeerConnection | New PeerConnection ([Connection with hashCode 1488492767 inboundInitiated? false initAt 1700627249089]) established with peer 0xf2fb7635a99a64f3f29948792123f75
[...]

PR #6284, opened 2023-12-12
[#6201] Duplicate "Terminating connection" message on peer connections

dlutzardo:
## PR description
There are two invocations to the method "terminateConnection" (ApiHandler and NettyPeerConnection). In the case of NettyPeerConnection it is called inside a listener.
The correct thing is to control that the terminateConnection method is executed once even if it is called several times.

## Fixed Issue(s)

#6201


pinges (approved):
LGTM

PR #6287, opened 2023-12-12
Increase scope of reference tests

shemnon:
## PR description

Add new forks to transaction tests, add bad rlp to rlp ref tests.

### DIFF
--- a/ethereum/referencetests/src/reference-test/java/org/hyperledger/besu/ethereum/core/TransactionTest.java
+++ b/ethereum/referencetests/src/reference-test/java/org/hyperledger/besu/ethereum/core/TransactionTest.java
@@ -26,6 +26,7 @@ import org.hyperledger.besu.ethereum.rlp.RLP;
 import org.hyperledger.besu.ethereum.transaction.TransactionInvalidReason;
 import org.hyperledger.besu.evm.gascalculator.BerlinGasCalculator;
 import org.hyperledger.besu.evm.gascalculator.ByzantiumGasCalculator;
+import org.hyperledger.besu.evm.gascalculator.CancunGasCalculator;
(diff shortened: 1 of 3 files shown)


ethereum/referencetests/src/reference-test/java/org/hyperledger/besu/ethereum/core/TransactionTest.java
```diff
@@ -125,8 +130,30 @@ public void london(final String name, final TransactionTestCaseSpec spec) {
...
+  @ParameterizedTest(name = "Name: {0}")
+  @MethodSource("getTestParametersForConfig")
+  public void shanghai(final String name, final TransactionTestCaseSpec spec) {
+    milestone(spec, name, "Shanghai", new ShanghaiGasCalculator(), Optional.of(Wei.of(0)));
+  }
+
+  @ParameterizedTest(name = "Name: {0}")
+  @MethodSource("getTestParametersForConfig")
+  public void cancun(final String name, final TransactionTestCaseSpec spec) {
+    milestone(spec, name, "Cancun", new CancunGasCalculator(), Optional.of(Wei.of(0)));
```
diega, line 148: I couldn't find Cancun TransactionTests in [ethereum/tests](https://github.com/ethereum/tests)#develop, are you adding this now so we don't forget them once they get added as it happened with Merge and Shangai?

diega (approved):
LGTM, great finding on this missing cases

ethereum/referencetests/src/reference-test/java/org/hyperledger/besu/ethereum/core/TransactionTest.java
shemnon: Adding so we don't forget.  If blob TXes are ever added to reference tests we will need cancun.

PR #6288, opened 2023-12-13
[MINOR] Include Enode URL in error message if parsing exception

macfarla:
eg in bootnodes config

Before

2023-12-13 12:08:50.949+10:00 | main | ERROR | Besu | Failed to start Besu Invalid IP address (or DNS query resolved an invalid IP). --Xdns-enabled is true but --Xdns-update-enabled flag is false.

After

2023-12-13 12:11:14.581+10:00 | main | ERROR | Besu | Failed to start Besu Invalid IP address 'enode://bob@bob:8000' (or DNS query resolved an invalid IP). --Xdns-enabled is true but --Xdns-update-enabled flag is false.


Gabriel-Trintinalia (approved):
LGTM

PR #6300, opened 2023-12-14
create trie package for bonsai and forest

matkt:


ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/BonsaiReferenceTestWorldState.java
macfarla, line 1: I realise this was not modified in this PR but the codeQL warning sounds like something we should look at -  

Non-synchronized override of synchronized method
Warning
Method 'addCachedLayer' overrides a synchronized method in  but is not synchronized.

macfarla (approved):
refactor looks good. There's an codeQL warning that is not related to this PR but maybe create a ticket for it 

Non-synchronized override of synchronized method
Warning
Method 'addCachedLayer' overrides a synchronized method in  but is not synchronized.

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/BonsaiReferenceTestWorldState.java
matkt: clearly maybe we need to create ticket for it. it's for reference tests so should be easy to fix

matkt: @garyschulte do you know why the synchronized is remove for this method ?

PR #6302, opened 2023-12-14
Disable txpool when not in sync

### LINKED ISSUE #6001: Disable TransactionPool observer on new blocks during Full sync
### Description
During Full sync (archive node), the transaction pool observer event is triggered for each new imported block to re-organize the transaction pool (remove block's transactions, sort one of the layers per base fee, ...etc).
Besu should disable transaction pool event during Full sync (and other syncs)

image

### Steps to Reproduce (Performance issue)
Start Besu with Full sync and Forest.

**Expected behavior:** 
Besu does not trigger the transaction pool observer event for each new block

Besu triggers the transaction pool observer event for each new block

100%

fab-10:
## PR description

With this PR the transaction pool factory is now aware if Besu is in sync or not, using the events triggered by the SyncState, and will disable the txpool as soon as the node is out of sync, so avoiding validating txs while not necessary.
This PR also implements a better way to process the onBlockAdded event in the txpool, avoiding the possibility of a tight loop that could result in high cpu usage.

## Fixed Issue(s)

fixes #6001


ahamlat:
@fab-10 I tested on a mainnet forest archive sync'ing node, and it worked. I don't see anymore the validation of remote transactions. I see the deactivation in the logs :
```
{"@timestamp":"2024-01-02T08:52:02,252","level":"INFO","thread":"nioEventLoopGroup-3-6","class":"TransactionPoolFactory","message":"Node out of sync, disabling transaction handling","throwable":""}
```
I also confirmed by checking the metrics :
image
image

However, I still see that we are announcing new transactions hashes :

image

fab-10:
@ahamlat about

>  I still see that we are announcing new transactions hashes

what you see are the _incoming_ messages that we still receive, but are just ignored and not processed, you should check the _outbound_ panel to verify that we are not sending them.

ahamlat:
Right, I was looking to the inbound messages. When checking outbound messages on the transaction pool, we can see that we have no traffic, which validates this PR.

image

macfarla (approved):

PR #6188, opened 2023-11-20
Add experimental x-trie-log subcommand for one-off backlog prune

siladu:
## PR description

Part of https://github.com/hyperledger/besu/issues/5390

## TODO

- [x] Test on a throwaway mainnet node: https://github.com/hyperledger/besu/pull/6188#issuecomment-1820156907
  - took **8m 48s** to prune 293458 trie logs (18 GB) on ~5 week old node
- [ ] Test on one of our oldest nodes to understand duration - my guess is it will be in the order of minutes, hopefully less than 1 hour.
- [ ] If it's really slow, could optimise by loading all trie logs once and reuse for count as well as prune (or just prune instead of count).

## To install on a node

```
cd /opt/besu
sudo wget -O 6188.tar.gz https://output.circle-artifacts.com/output/job/a7a06df9-cf33-4115-89d1-413c86facd71/artifacts/0/distributions/besu-23.10.3-SNAPSHOT.tar.gz
sudo tar zxf 6188.tar.gz --transform s/besu-23.10.3-SNAPSHOT/6188/
time sudo /opt/besu/6188/bin/besu --config-file=/etc/besu/config.toml storage x-trie-log count
```

## Usage and examples

```
$BESU storage x-trie-log -h
Usage: besu storage x-trie-log [-hV] [COMMAND]
Manipulate trie logs
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
Commands:
  count  This command counts all the trie logs
  prune  This command prunes all trie log layers below the retention threshold,
           including orphaned trie logs.
```

```
$BESU --genesis-file=genesis.json --data-path=besu --data-storage-format=BONSAI storage x-trie-log count
...
Counting trie logs...
trieLog count: 512
 - canonical count: 512
 - fork count: 0
 - orphaned count: 0
```

```
$BESU --genesis-file=genesis.json --data-path=besu --data-storage-format=BONSAI storage x-trie-log prune
```

```
$BESU --genesis-file=genesis.json --data-path=besu --data-storage-format=BONSAI --Xbonsai-trie-log-retention-threshold=512 --Xbonsai-trie-log-pruning-limit=513 storage x-trie-log prune
...
Counting trie logs before prune...
trieLog count: 615
 - canonical count: 512
 - fork count: 0
 - orphaned count: 103

Total to prune = 615 (total) - 512 (retention threshold) =
=> 103

Estimated number of batches = max(103 (total to prune) / 513 (batch size), 1) = 1

Pruning batch 1
[...]

PR #6303, opened 2023-12-15
Add --X-trie-log subcommand

gfukushima:
## PR description
This is a more direct approach to the previous attempt in #6188.
Tests showed that the previous approach was taking too long (3 hours) to prune a 3.5 months old (keeping the lastest 512 only).
The approach this PR takes is faster since it resets the column family and only retain the necessary trie logs (saved in files in batches of 20_000 trie logs).
Drawback of this approach is that we do store the trie logs in disk (in files) before dropping the column. Realistic speaking there's little to no reason to actually keep many more trie logs than the default(512).
In any case, I've made some adjustments to the code to allow this to go as far as a user might like.
 
Ran tests on nodes retaining 512 / 20_000 / 250_000 / 1M trie logs, all instances had approx 1.8M trie logs in the db when node was stopped for prunning.

Below some of the logs condensed
 
Real use case scenario - 512 trie logs retained (1 second)
```
2023-12-22 05:40:22.409+00:00 | main | INFO  | TrieLogHelper | Saving trie logs to retain in file (batch 1)...
2023-12-22 05:40:22.416+00:00 | main | INFO  | TrieLogHelper | Obtaining trielogs from db, this may take a few minutes...
2023-12-22 05:40:22.692+00:00 | main | INFO  | TrieLogHelper | Clear trie logs...
2023-12-22 05:40:23.131+00:00 | main | INFO  | TrieLogHelper | Restoring trie logs retained from batch 1...
2023-12-22 05:40:23.446+00:00 | main | INFO  | TrieLogHelper | Deleting files...
2023-12-22 05:40:23.452+00:00 | main | INFO  | TrieLogHelper | Prune ran successfully. Enjoy some disk space back! 🚀
```

a more extreme case where for some reason someone has been running an archieve and decides to keep only the latest 1M trie logs - 1M trie logs retained (approx 2 hours)
```
2024-01-03 01:48:18.042+00:00 | main | INFO  | BonsaiWorldStateKeyValueStorage | Bonsai flat db mode found PARTIAL
2024-01-03 01:48:18.806+00:00 | main | INFO  | TrieLogHelper | Saving trie logs to retain in file (batch 1)...
2024-01-03 01:48:18.809+00:00 | main | INFO  | TrieLogHelper | Obtaining trielogs from db, this may take a few minutes...
2024-01-03 01:48:34.970+00:00 | main | INFO  | TrieLogHelper | Saving trie logs to retain in file (batch 2)...

2024-01-03 03:42:11.016+00:00 | main | INFO  | TrieLogHelper | Deleting files...
2024-01-03 03:42:11.029+00:00 | main | INFO  | TrieLogHelper | Prune ran successfully. Enjoy some disk space back! 🚀
```

## Fixed Issue(s)

Part of #5390


besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/RocksDbUsageHelper.java
```diff
@@ -34,17 +34,19 @@ static void printUsageForColumnFamily(
       final RocksDB rocksdb, final ColumnFamilyHandle cfHandle, final PrintWriter out)
       throws RocksDBException, NumberFormatException {
     final String size = rocksdb.getProperty(cfHandle, "rocksdb.estimate-live-data-size");
+    final String numberOfKeys = rocksdb.getProperty(cfHandle, "rocksdb.estimate-num-keys");
     boolean emptyColumnFamily = false;
-    if (!size.isEmpty() && !size.isBlank()) {
+    if (!size.isEmpty() && !size.isBlank() && !numberOfKeys.isEmpty() && !numberOfKeys.isBlank()) {
```
jframe, line 39: think we can just use isBlank() as that will return true if it empty or only contains whitespace

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,227 @@
...
+          bf.flush();
+      } catch (IOException e) {
+          LOG.error(e.getMessage());
+          throw e;
+      }
+  }
+
+  private static IdentityHashMap<byte[], byte[]> readTrieLogsFromFile() throws IOException {
+
+    File file = new File(trieLogFile);
```
jframe, line 197: Wold be better to use File.createTempFile so it's created in the temp directory and we won't have any write permissions issues

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,227 @@
...
+      } catch (IOException e) {
+          LOG.error(e.getMessage());
+          throw e;
+      }
+
+    return trieLogs;
+  }
+
+  private static void deleteTrieLogFile() {
+    File file = new File(trieLogFile);
```
jframe, line 216: When we create the file we could use the File.deleteOnExit so we don't need to worry about manually deleting this file

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,227 @@
...
+      // save trielogs in a flatfile as a fail-safe
+      out.println("Obtaining trielogs to retain...");
+      hashesToRetain.forEach(
+          hash -> {
+            rootWorldStateStorage
+                .getTrieLog(hash)
+                .ifPresent(trieLog -> trieLogsToRetain.put(hash.toArrayUnsafe(), trieLog));
+          });
+      out.println("Saving trielogs to retain in file...");
+      saveTrieLogsInFile(trieLogsToRetain);
```
jframe, line 98: Not sure how the fail-safe is being used. Expected to see that if we got an error clearing and inserting entries we ready from the file or allow specifying an existing backup file to restore from.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,227 @@
...
+    } else {
+      // try to read the triLogs from the flatfile
+      trieLogsToRetain = readTrieLogsFromFile();
```
jframe, line 101: Don't understand how reading from the file helps in this case. We didn't find all the hashes but also haven't created the file at this point either unless we are relying running the command again to read the previous state

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,227 @@
...
+    rootWorldStateStorage.clearTrieLog();
+
+    // get an update and insert the trielogs we retained
+    var updater = rootWorldStateStorage.updater();
+    out.println("restore trielogs retained into db...");
+    trieLogsToRetain.forEach(
+        (key, value) -> {
+          updater.getTrieLogStorageTransaction().put(key, value);
+        });
+    updater.getTrieLogStorageTransaction().commit();
```
jframe, line 114: If this fails then do have a way of restoring the trielog from the backup?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: We need the file to exist after we finish the run if for some reason we fail to reload the trielogs into the db

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: Added a comment and instructions that will be logged for the user. Basically the idea here is, we retain the logs in a file in case something goes wrong.
In that case the user can re-run the subcommand and we will read the trielogs from the file since the column family might have been dropped already.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/RocksDbUsageHelper.java
```diff
@@ -34,17 +34,17 @@ static void printUsageForColumnFamily(
       final RocksDB rocksdb, final ColumnFamilyHandle cfHandle, final PrintWriter out)
       throws RocksDBException, NumberFormatException {
     final String size = rocksdb.getProperty(cfHandle, "rocksdb.estimate-live-data-size");
+    final String numberOfKeys = rocksdb.getProperty(cfHandle, "rocksdb.estimate-num-keys");
```
garyschulte, line 37: changes here seem fine but unrelated.  Every time I look at this class name though I think it is a CLI Usage helper, like the output that gets printed when one asks for help or gets command usage wrong.  completely unrelated, but if we are going to make changes in this pr ... :)

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
garyschulte: I think this is a failsafe in for a subsequent execution where something went terribly wrong during the truncation and restore-from-file process.

what would be nice, but not idiomatic, is if we had a few "are you sure?" or "do you want to?" prompts that could be bypassed by `-y` similar to ubuntu apt commands

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,241 @@
...
+    if (trieLogsToRetain.size() == layersToRetain) {
+      out.println("Clear trielogs...");
+      rootWorldStateStorage.clearTrieLog();
+      out.println("Restoring trielogs retained into db...");
+      recreateTrieLogs(rootWorldStateStorage, trieLogsToRetain);
+    }
+    if (rootWorldStateStorage.streamTrieLogKeys(layersToRetain).count() == layersToRetain) {
+      out.println("Prune ran successfully. Deleting file...");
+      deleteTrieLogFile();
+      out.println("Enjoy some GBs of storage back!...");
```
garyschulte, line 117: needs rocket emoji

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,241 @@
...
+    } catch (IOException e) {
+      LOG.error(e.getMessage());
+      throw e;
+    }
+  }
+
+  private static IdentityHashMap<byte[], byte[]> readTrieLogsFromFile() throws IOException {
+
+    File file = new File(trieLogFile);
+    IdentityHashMap<byte[], byte[]> trieLogs = new IdentityHashMap<>();
```
garyschulte, line 210: trielogs can vary in size, and the retention count is user configurable.  We could easily blow up the heap with bad settings.  We might want to add some sanity check on the file size here, and perhaps also in trielog streaming/save to file.

garyschulte (commented):
Looks good.  I think it wouldn't be too difficult to stream trielogs to and from the file directly to prevent OOM and heap issues.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: Agree, users could eventually set the `logs-to-retain` to a high number and this could blow up. Is there real scenario where we should handle a high number having in mind that rolling back too many layer is not something feasible in bonsai (afaik)?

garyschulte: Trielogs can be the basis of a specific worldstate for debug_ and trace_ endpoints.  But really if that were the case I imagine the node would be an archive node.  I think streaming to and from disk would likely just be more bulletproof.  

Not blocking feedback, just seems a robust mechanism

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,369 @@
...
+
+  private static void processTrieLogBatches(
+      final BonsaiWorldStateKeyValueStorage rootWorldStateStorage,
+      final MutableBlockchain blockchain,
+      final long chainHeight,
+      final long lastBlockNumberToRetainTrieLogsFor,
+      final long numberOfBatches,
+      final String batchFileNameBase) {
+
+    for (long batchNumber = 1; batchNumber <= numberOfBatches; batchNumber++) {
```
matkt, line 98: could we explain why we start at 1 ?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,369 @@
...
+    final List<Hash> trieLogKeys = new ArrayList<>();
+
+    getTrieLogKeysForBlocks(blockchain, firstBlockOfBatch, lastBlockOfBatch, trieLogKeys);
+
+    LOG.info("Saving trie logs to retain in file (batch {})...", batchNumber);
+
+    try {
+      saveTrieLogsInFile(trieLogKeys, rootWorldStateStorage, batchNumber, batchFileNameBase);
+    } catch (IOException e) {
+      LOG.error("Error saving trie logs to file: {}", e.getMessage());
```
matkt, line 138: it seems that if we have an exception we continue the process, so is it possible we missed some trielog in the restore step in case of failure here

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,369 @@
...
+  private static void restoreTrieLogBatches(
+      final BonsaiWorldStateKeyValueStorage rootWorldStateStorage,
+      final long batchNumber,
+      final String batchFileNameBase) {
+
+    try {
+      LOG.info("Restoring trie logs retained from batch {}...", batchNumber);
+      recreateTrieLogs(rootWorldStateStorage, batchNumber, batchFileNameBase);
+    } catch (IOException | ClassNotFoundException e) {
+      LOG.error("Error recreating trie logs from batch {}: {}", batchNumber, e.getMessage());
```
matkt, line 151: What should we do at this point? Do we have a way to restart from an existing restore file?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: just for the sake of naming the file, didn't want a file named "trieLogsToRetain-0"

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: I'm going to change this to throw a new Runtime exception so we abort the process here, meaning if you re-run the command it's gonna detect the files exist and skip the save to file part but it will still reset the db and try to restore from file.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: going to change this to throw a runtime exception, thanks for catching this

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: added

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: I've changed this so we process it in batches now so it should be a safer mechanism.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,359 @@
...
+
+    final long chainHeight = blockchain.getChainHeadBlockNumber();
+
+    final long lastBlockNumberToRetainTrieLogsFor = chainHeight - layersToRetain + 1;
+
+    if (!validPruneRequirements(blockchain, chainHeight, lastBlockNumberToRetainTrieLogsFor)) {
+      return;
+    }
+
+    final long numberOfBatches = calculateNumberofBatches(layersToRetain);
```
jframe, line 71: Why does this need to be done in batches?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,359 @@
...
+  private static void saveTrieLogBatches(
+      final String batchFileNameBase,
+      final BonsaiWorldStateKeyValueStorage rootWorldStateStorage,
+      final long batchNumber,
+      final List<Hash> trieLogKeys) {
+
+    LOG.info("Saving trie logs to retain in file (batch {})...", batchNumber);
+
+    try {
+      saveTrieLogsInFile(trieLogKeys, rootWorldStateStorage, batchNumber, batchFileNameBase);
```
jframe, line 127: Could this overwrite an existing batch file if the command runs after a failure?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: I'm using limited batches to provide a safety allocation of resources. Mostly a safety limit to prevent besu to get killed or exhaust resources if a user decide to retains a high number of trie logs.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,359 @@
...
+  private static final long BATCH_SIZE = 20_000;
+  private static final int ROCKSDB_MAX_INSERTS_PER_TRANSACTION = 1000;
+  private static final Logger LOG = LoggerFactory.getLogger(TrieLogHelper.class);
+
+  static void prune(
+      final DataStorageConfiguration config,
+      final BonsaiWorldStateKeyValueStorage rootWorldStateStorage,
+      final MutableBlockchain blockchain,
+      final Path dataDirectoryPath) {
+    final String batchFileNameBase = dataDirectoryPath.resolve(DATABASE_PATH) + "/" + TRIE_LOG_FILE;
```
jframe, line 58: Be better to use resolve than OS dependent separator

i.e.
`dataDirectoryPath.resolve(DATABASE_PATH).resolve(TRIE_LOG_FILE)`

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: It can't overwrite files, if there's an existing file with the same name we simply skip that batch. Think it's better to do it this way than overwriting files. specially because you've could have dropped the db already and overwriting the first batch with an empty file.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -0,0 +1,359 @@
...
+    TrieLogHelper.validatePruneConfiguration(config);
```
jframe, line 59: nit: This is in the same class no benefit in using the fully qualified path to the method. Doesn't look we are doing that anywhere else.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/trielog/TrieLogPruner.java
```diff
@@ -176,8 +180,9 @@ private NoOpTrieLogPruner(
     }
 
     @Override
-    public void initialize() {
+    public int initialize() {
       // no-op
+      return -1;
```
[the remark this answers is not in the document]
gfukushima: Think this was inherit from the old approach, I don't really mind -1 or 0.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
jframe: Missed the check on line 272 for an existing file

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: applied suggestion, thanks

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: suggestion applied

jframe (approved):

PR #6305, opened 2023-12-15
mark deleted slot during clear storage step

matkt:


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -361,14 +362,24 @@ private void clearStorage(
                     (location, key) -> getStorageTrieNode(addressHash, location, key),
                     oldAccount.getStorageRoot());
             try {
+
+              final StorageConsumingMap<StorageSlotKey, BonsaiValue<UInt256>> storageToDelete =
+                  worldStateUpdater.getStorageToUpdate().get(address);
               Map<Bytes32, Bytes> entriesToDelete = storageTrie.entriesFrom(Bytes32.ZERO, 256);
```
garyschulte, line 368: while we are here we could push this into FlatDBStrategy, get better performance than trie walking

garyschulte (commented):
This looks correct to me.  We should re-run a full sync linea test net and compare the SMT state root to be certain of the impact of the change since this will likely alter the content of trielogs.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/worldview/BonsaiWorldState.java
matkt: I think at the end we still need to climb the tree in order to clean the trie https://github.com/hyperledger/besu/pull/6305/files#diff-8de415200fe0cfb23305bb4f59cfc5fe3fd3c1d5cc02db6431001cc0447ba4c0L372

So I don't think it will help for performance

garyschulte (approved):
This was tested and released with 23.10.3.  Ready to merge 👍

PR #6306, opened 2023-12-18
Generate genesis root hash with the used data storage format

matkt:
## PR description

Before, we always generated the root hash of the genesis with forest, even if we activated bonsai. I am modifying the logic to use the correct storage format that we activated. This will facilitate a potential removal of forest and also be able to generate the genesis root hash using verkle when it is integrate


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/trielog/NoOpTrieLogManager.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class NoOpTrieLogManager extends TrieLogManager {
+
+  public NoOpTrieLogManager() {
+    super(null, null, 0, null, TrieLogPruner.noOpTrieLogPruner());
+  }
+
+  @SuppressWarnings({"UnsynchronizedOverridesSynchronized", "squid:S3551"})
+  @Override
+  public void saveTrieLog(
```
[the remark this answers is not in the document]
garyschulte: I think this can be synchronized without any problem.  It looks like Simon didn't add synchronized here when he changed the interface.  This is only used in the referencetest worldstate, so I don't think it should cause an issue to make it synchronized

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/GenesisState.java
```diff
@@ -133,17 +175,42 @@ private static void writeAccountsTo(
...
+  private static Hash calculateGenesisStateHash(
+      final DataStorageFormat dataStorageFormat, final List<GenesisAccount> genesisAccounts) {
+    final MutableWorldState worldState = loadWorldState(dataStorageFormat);
     writeAccountsTo(worldState, genesisAccounts, null);
     return worldState.rootHash();
   }
 
+  private static MutableWorldState loadWorldState(final DataStorageFormat dataStorageFormat) {
+    switch (dataStorageFormat) {
+      case BONSAI -> {
```
garyschulte, line 187: We leak a lot of storage specific logic into GenesisState here.  Can we push this into DataStorageFormat itself, and have it take a KeyValueStorageProvider param?  I think that would encapsulate the storage-format-specific creation in the format itself.

garyschulte (commented):
LGTM, just one bit about leaking data storage format details

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/GenesisState.java
matkt: I updated my code

garyschulte (approved):
🚢   Only feedback is that we could default to BONSAI for the fromConfig overload

Issue #6266, opened 2023-12-08
Accidental downgrade followed by re-upgrade leaves node broken


matthew1001:
### Description
If Besu is accidentally downgraded to a lower version, the node's data can be corrupted in such as way that when the node is restarted at the correct version the node isn't usable. E.g.

- Start node at `v23.10.2`
- Stop node
- Start node with `v23.4.1`
- Node doesn't work
- Stop node
- Start node with `v23.10.2` again
- Node doesn't work any more

### Acceptance Criteria
When starting Besu at a version lower than the highest version that has been used with that node's data, Besu should fail on startup with a clear error message, rather than attempt to start and correct the data.

non-fungible-nelson:
Hi Matt - let us know if you have any ideas here. Specifically we changed the DB format in [23.4.4](https://github.com/hyperledger/besu/releases/tag/23.4.4) and provided downgrade commands in the release notes. A change to recognize "older versions" would really only be compatible from the version its released with onward.

matthew1001:
Yeah I think my idea for a resolution to this was to detect a downgrade and prevent startup without an extra CLI arg being set. I thought I'd have a go at a draft PR along those lines and see what others think.

fab-10:
could you share the logs of every step? 
the automated db upgrade, should not left the db in an inconsistent state, and using the sub command as described here should make the db working again.

non-fungible-nelson:
@matthew1001 pinging here

matthew1001:
Sorry, didn't see @fab-10 's comment. I did post an updated in the PR a few days ago though (https://github.com/hyperledger/besu/pull/6307#issuecomment-1875125648) which would be good to get peoples' views on.

I don't have logs retained for the scenario where we hit it, but it was a fairly simple QBFT setup as outlined in the description. The chain had been mining blocks at 23.10.2 perfectly fine, then after restarting with version 23.4.1 installed the chain wouldn't mine blocks at all.

I suspect the DB revert command would have worked, but I think it would be useful to have the option to prevent a lower version of Besu from starting given that the likelihood of needing to upgrade/downgrade isn't very common and having to set an option to allow it isn't particularly onerous.

PR #6307, opened 2023-12-18
Restrict downgrade

matthew1001:
## PR description
This PR aims to prevent accidental downgrade of Besu, which can potentially cause the DB to be irrevocably corrupted.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/6266


non-fungible-nelson:
LGTM functionality wise --> what you are proposing is a change to the existing default behavior (allowing downgrades). Should we consider the inverse (a flag to prevent downgrades or just switching to `--prevent-downgrades`)? It's really a coin-toss. I think users on Mainnet would be more likely to downgrade/upgrade more often and may benefit from the default remaining the same. I am open to either, however. This does not have to be a blocker.

matthew1001:
Thanks for the comments @non-fungible-nelson I'll give that some thought and reply properly tomorrow.

In the meantime I wanted to highlight that I've just updated the description and commits somewhat, as I originally had the checks happening in the RocksDB plugin but have decided that it's better that Besu can perform this check regardless of the storage provider being used.

matthew1001:
> Should we consider the inverse (a flag to prevent downgrades or just switching to --prevent-downgrades)? It's really a coin-toss. I think users on Mainnet would be more likely to downgrade/upgrade more often and may benefit from the default remaining the same. 

Yeah I agree that it changes behaviour, but the main cost to a user who is expecting to be able to do this is to set `--allow-downgrade` and leave it turned on. The downside to someone who accidentally downgrades is potentially much worse, so I think I'd argue that it's a good reason for a behaviour/breaking change.

matthew1001:
Moving out of draft as I think this is ready for review

garyschulte (commented):
I would prefer to see this as a function of the data storage format.  Version number isn't a great proxy for breaking changes in storage format.  It could well be that a forest node could up/downgrade versions with no impact, but a bonsai node might have problems.  Same can be said for the coming verkle format.  

It is a harder nut to crack to tie the safety to storage format features, but IMO it is better to extend DatabaseMetadata with a capability matrix rather than making it version based.

Also, some commands like the storage subcommand `revert-variables` do not take into account besu version, even though their intent is to migrate to/from different data storage formats.  This kinda bricks their intent.  @fab-10

fab-10:
if I understood it right, by default when a user tries to downgrade, Besu will not start, what is written to the log in that case?

I think a protection is nice to have, and I am more inclined toward the idea proposed by @garyschulte, that applying this protection on every downgrade, since this kind of incompatible changes are not frequent.

Regarding my change in 23.4.4, with hindsight, not changing the db version was a mistake, because with it the existing  mechanism would have prevented to downgrade without manually running the revert sub command. 

To summarize, I do not think we need this general _restrict downgrade_ feature, since the existing protection is enough and we must ensure that we increment correctly the db version, when we make breaking changes in the db, and when handy we can provide migration or revert sub commands. 

To minimize the impact of the mistake, of not bumping the db version in 23.4.4, I can update the version now, this will at least prevent downgrading from >=24.1.x to <=23.4.3, unless the user manually runs the revert sub command.

matthew1001:
Hi @fab-10 @garyschulte thanks for the comments, and apologies for the slow reply. Just back from the Christmas break, hope you both had a good one.

I actually started down the route you described in your comment @garyschulte, where I looked at adding to the existing `DatabaseMetadata` structure and recording the Besu version in `DATABASA_METADATA.json`, since the aim of the PR is to add extra protection to DB data corruption. However, as the PR progressed it felt like it was a little too implementation specific and what I wanted was a more DB-agnostic approach that would work even if someone swapped in their own storage plugin as a replacement for RocksDB. So where my thinking ended up was that the new option should be at a level higher than the DB (hence moving the metadata into a specific `VERSION_METADATA.json` file) and be more of a back-stop protection that enterprises could use to spot unintentional downgrades, even if the particular DB plugin in question supports a downgrade between those particular versions.

So in my mind the PR is more of a Besu-runtime check than a DB check, which as you both point out has its own protections in place and in many cases supports a downgrade correctly. So one way to think of the PR in light of the change in mindset is to think of it as a CI/CD pipeline protection check where an enterprise might be rolling out their own builds of Besu as new versions are released, but want to ensure that what they actually end up running in production is what they expect, and that their pipeline hasn't inadvertently pulled the wrong docker tag or started a version against the wrong PVC.

It might seem over-the-top to have a feature baked into Besu to give that protection, but where production data is concerned and accidental data corruption can have quite severe consequences it's the kind of feature that enterprises really appreciate.

With all that in mind, it would be good to get both your feedback on the PR based on the above. However, I thought I'd suggest some alternatives/tweaks we could consider:

- Make the default that downgrades are allowed (along the lines @non-fungible-nelson suggested), so an enterprise user would have to set that to `false` to get the protection
  - In my mind this limits the value of the feature, because it becomes easier not to notice that your pipeline has stopped setting it. However, perhaps some of the default profile work being discussed in https://wiki.hyperledger.org/display/BESU/Change+to+Besu+Defaults+-+Q1+2024 would mitigate that by making it the default in an enterprise profile and non-default in other profiles.
  - My thinking around making it the default was that it seemed unlikely someone running a public node would regularly be starting/stopping Besu at lower versions, and that hopefully very few public-node users would be affected by it defaulting to on. 
- Change the CLI arg to take a version number (the version the user thinks they are starting)
  - This would eliminate the need to persist anything to disk as the CLI arg could be compared against the runtime version number on each startup
  - I think this suffers from the same issue above, that if you don't set it you don't notice you're not setting it until something bad happens
  - It also requires more effort to use the feature, needing to updating the CLI arg on each upgrade rather than relying on Besu to update the version number on each successful startup at a higher version.

@fab-10 As an aside from the above, I think updating the DB version now as you've suggested would be really good to do.

fab-10:
I am still evaluating pros and cons of this feature, in my experience I haven't seen many apps offering it, so it will be nice to know about other apps that have this feature will be nice to see how them works in that regards.
Of course when talking about enterprise usage, there should be other good practices to prevent and recover from data corruption/loss, outside the app itself, like backups, proper test env, QA control and change management. Human error could always happen, but they should have a plan for it.
So I would keep this as an optional feature, at least for a while, with the possibility to enable it for some profiles, if we see interest around it.

I will propose a PR to bump the version of the DB following the introduction of the variables storage.

matthew1001:
I've manually re-tested with the new `--version-compatibility-protection`:
- [ ] Without it set, the `VERSION_METADATA.json` file is populated if it doesn't exist, but no version check is made
- [ ] With it set, if the version hasn't changed from what is in `VERSION_METADATA.json` Besu starts as expected
- [ ] With it set, if the version is older than the version in `VERSION_METADATA.json` Besu fails to start with
 - `Besu version 24.1.2 is lower than version 24.1.3 that last started. Remove --version-compatibility-protection option to allow Besu to start at  the lower version (warning - this may have unrecoverable effects on the database).`

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
 - Log blob count when importing a block via Engine API [#6466](https://github.com/hyperledger/besu/pull/6466)
 - Introduce `--Xbonsai-limit-trie-logs-enabled` experimental feature which by default will only retain the latest 512 trie logs, saving about 3GB per week in database growth [#5390](https://github.com/hyperledger/besu/issues/5390)
 - Introduce `besu storage x-trie-log prune` experimental offline subcommand which will prune all redundant trie logs except the latest 512 [#6303](https://github.com/hyperledger/besu/pull/6303)
+- Option to perform version incompatibility checks when starting Besu. In this first release of the feature, if `--version-compatibility-protection` is set to true it checks that the version of Besu being started is the same or higher than the previous version. [6307](https://github.com/hyperledger/besu/pull/6307)
```
fab-10, line 18: this need to be moved, since 24.1.2 is out now

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/VersionMetadata.java
```diff
@@ -0,0 +1,147 @@
...
+  public VersionMetadata(@JsonProperty("besuVersion") final String besuVersion) {
+    this.besuVersion = besuVersion;
+  }
+
+  public String getBesuVersion() {
+    return besuVersion;
+  }
+
+  public static VersionMetadata lookUpFrom(final Path dataDir) throws IOException {
+    LOG.info("Lookup version metadata file in data directory: {}", dataDir.toString());
```
fab-10, line 61:
```suggestion
    LOG.info("Lookup version metadata file in data directory: {}", dataDir);
```

fab-10 (approved):
LGTM, just a couple of edits

CHANGELOG.md
matthew1001: Good point, fixed

ethereum/core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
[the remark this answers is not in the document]
matthew1001: I'm trying to recall what `mock` needed it for. I think it was for mocking final methods & classes, and we use it elsewhere in the tests, but I've just re-run the `VersionMetadataTest` class without it in place and they pass so for now I've removed it. Possibly I needed it before some of the refactoring that came along in later commits. If the CI/CD tasks fail I might re-instate it but for now I've deleted it.

PR #6329, opened 2023-12-22
Copy also computed fields, when doing a Transaction detached copy

fab-10:

## PR description

When doing a detached copy of a transaction, copy also the computed fields to avoid to recompute them


ahamlat:
@fab-10 I still see the execution related to getting the sender from the signature when processing a transaction

fab-10:
@ahamlat this applies only to incoming txs, not txs from block validation, but in any case is good to have to avoid double computation of some fields.

macfarla (approved):

PR #6340, opened 2024-01-03
Fix trielog shipping issue during self destruct

matkt:
## PR description

Fix trielog shipping issue during self destruct


garyschulte (approved):
🚢

PR #6343, opened 2024-01-03
Make RPC reason settable, pass execution failure reason in RPC error message

### LINKED ISSUE #4914: Add Explanation on Why Transaction is Rejected
Users are told that their transactions are rejected but don't know why they were rejected. For better UX, we should add some description to explain why it was rejected

0.5 day spike to see where in the code the tx rejection logic lives and investigate why we tell users tx is rejected. If it's not clear on the code and we need to investigate further, then defer this issue to later.

matthew1001:
## PR description
This PR adds a setter for the `reason` field in `JsonRpcError`.

If a caller sets the `reason`, it is appended to the message for the error that is taken from the `RpcErrorType` enums.

It also adds a new `RpcErrorType` of `EXECUTION_HALTED` to be used during transaction simulations (e.g. during `eth_estimateGas()` calls) so a failed gas estimation doesn't just return `INTERNAL_ERROR` when there's a problem with the execution.

The specific example I hit this in was using a Shanghai smart contract with the `JUMP0` opcode against a QBFT Besu node (QBFT doesn't currently support Shanghai).

## Fixed Issue(s)
Fixes #4914 
(See also https://github.com/hyperledger/besu/pull/5706 which improves the error messages sent back to the user for reverted transactions)

## TODO before taking out of draft
 - [x] Unit test fixes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 24.1.0-SNAPSHOT
 
 ### Breaking Changes
+- New `EXECUTION_HALTED` error returned if there is an error executing or simulating a transaction, with the reason for execution being halted. Replaces the generic `INTERNAL_ERROR` return code in certain cases which some applications may be checking for [#6343](https://github.com/hyperledger/besu/pull/6343)
 
 ### Deprecations
 - Forest pruning (`pruning-enabled` options) is deprecated and will be removed soon. To save disk space consider switching to Bonsai data storage format [#6230](https://github.com/hyperledger/besu/pull/6230)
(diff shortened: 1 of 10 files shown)


jflo (commented):
So far so good is this a fix for #4914 ?

matthew1001:
> So far so good is this a fix for #4914 ?

I think it helps to resolve part of what #4914 is asking for, along with https://github.com/hyperledger/besu/pull/5706 from August 2023 which ABI-decodes revert reasons before returning them to the user.

It can always be re-opened, but probably needs to specify which calls, which error messages etc. aren't getting back to the user.

jflo (approved):

macfarla (approved):
LGTM

Issue #6344, opened 2024-01-03
Under heavy load `eth_estimateGas` returns `INTERNAL_ERROR`


matthew1001:
### Description
Running a moderately high workload using caliper, after ~ 2000 transactions `eth_estimateGas` returns `INTERNAL_ERROR` to the caliper WS client

### Command that was run to start Besu (single validator):
`besu --data-path=data --genesis-file=../genesis.json --min-gas-price=0  --sync-mode=FULL --rpc-http-enabled --rpc-ws-enabled --rpc-http-apis=DEBUG,TXPOOL,WEB3,ETH,TRACE,QBFT,ADMIN --tx-pool-limit-by-account-percentage=0.20 --rpc-ws-max-active-connections=300 --tx-pool=sequenced -l TRACE`

### Logs
I've pinned it down to this code block:

```
final BlockHeader blockHeader = blockHeader();
if (blockHeader == null) {
    return errorResponse(requestContext, RpcErrorType.INTERNAL_ERROR);
}
```

https://github.com/hyperledger/besu/blob/e0cd89f9b5ca27763ba4b7b2589d486df81a9626/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthEstimateGas.java#L53

It was difficult to pin down the offending line using `TRACE` but adding a `new Exception().printStackTrace()` to the `JsonRpcErrorResponse(...)` constructor gave the stack trace:

```
java.lang.Exception
        at org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.JsonRpcErrorResponse.<init>(JsonRpcErrorResponse.java:47)
        at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.AbstractEstimateGas.errorResponse(AbstractEstimateGas.java:135)
        at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.AbstractEstimateGas.errorResponse(AbstractEstimateGas.java:130)
        at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.EthEstimateGas.response(EthEstimateGas.java:53)
        at org.hyperledger.besu.ethereum.api.jsonrpc.execution.BaseJsonRpcProcessor.process(BaseJsonRpcProcessor.java:44)
        at org.hyperledger.besu.ethereum.api.jsonrpc.execution.JsonRpcExecutor.execute(JsonRpcExecutor.java:92)
[...]

PR #6345, opened 2024-01-03
Use synchronized call to access the chain head block in `eth_estimateGas`

matthew1001:
## PR description
Optimistically get the chain head block without taking a lock. If that fails, retry under the lock.

This is an identical fix to https://github.com/hyperledger/besu/pull/6143.

I did consider making a more generic helper/util function to wrap the logic for any place in the code that needs it, but looking through the code I can't see any other cases where the a failed attempt to get the chain head block results in an internal error or exception. Most others treat it as optional and just retry on the next go round.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/6344


macfarla (approved):

macfarla:
@matthew1001 I think this is worth a changelog entry

Issue #5446, opened 2023-05-11
Shanghai support for QBFT/IBFT


jframe:
QBFT chains are not able to produce blocks when configured to use Shanghai. Even though Shanghai is mostly about supporting Withdrawals it does add new EVM opcodes e.g. PUSH0 that could be of use in private networks.

To reproduce configure a genesis file with:
```
{
  "config": {
    "chainid": 1337,
    "shanghaiTime": 0,
    "zeroBaseFee":true,
    "qbft": {
      "epochlength": 30000,
      "blockperiodseconds": 5,
      "requesttimeoutseconds": 10
    }
```

This is the error we get:
```
2023-05-11 10:49:44.474+10:00 | BftProcessorExecutor-QBFT-0 | WARN  | WithdrawalsValidator$AllowedWithdrawals | withdrawals must not be null when Withdrawals are activated
2023-05-11 10:49:44.474+10:00 | BftProcessorExecutor-QBFT-0 | INFO  | MainnetBlockValidator | Invalid block 1 (0x2dc8d5f94285d22afab11929b5cb872e1b158825459ef01cb7230520a36f9dfc)
2023-05-11 10:49:44.475+10:00 | BftProcessorExecutor-QBFT-0 | INFO  | ProposalPayloadValidator | Invalid Proposal Payload: block did not pass validation. Reason Optional[failed to validate output of imported block]
```

jframe:
Shanghai works with QBFT in 23.1.0 but not with 23.4.0

jframe:
Suggestion from @siladu:
Maybe just need to wire in ProhibitedWithdrawals in the "qbftSpecificModifications"

jframe:
Confirmed that this wasn't working with 231.0, the shanghaiTime was being ignored for this config and we are using defaulting to using frontier. So this isn't a regression.

NickSneo:
Did you check if it works with CLIQUE or ethash? I am getting the same error when trying to configure shanghaiTime with CLIQUE and ethash - https://github.com/hyperledger/besu/issues/5468
If it works for you then let me know the genesis file you used

jframe:
Didn't test with Clique or ethhash yet. I suspect the same issue also applies to them, as the changes for shanghaiTime were made to work with POS. Will have a look at #5468 and respond there if it's the same issue or not.

NickSneo:
Is someone working on this bug? Or maybe we can pick this to look into it?

jframe:
[...]

PR #6353, opened 2024-01-04
QBFT + shanghai support

matthew1001:
## PR description
Implement shanghai support for QBFT and IBFT chains.

Main areas of change:

- Previously the code made the assumption that BFT block creators and validators could be created ahead of time, on creation of a new `QbftRound`/`IbftRound`, since the next block number was known ahead of time. With number-based milestones such as `london` this is fine, but with timestamp-based milestones the type of EVM to use for a new round needs to be based on the time that the block is being created. As such, in many places a `protocolSchedule` is now passed down the stack instead of a `blockValidator` or a `blockImporter` to allow these to be created using the `protocolSchedule` when the timestamp is known.
- The `BftProtocolSchedule` was only designed to allow a protocol spec to be requested by block number, so a new method `getByBlockNumberAndTimestamp` that takes both the block number and timestamp has replaced `getByBlockNumber`, and determines the fork spec based on whichever milestone is most recent.

I don't think any docs changes are required since we don't currently make any mention of `shanghai` in the docs today.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/5446


consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftRound.java
```diff
@@ -115,7 +118,28 @@ public ConsensusRoundIdentifier getRoundIdentifier() {
    * @param headerTimeStampSeconds the header time stamp seconds
    */
   public void createAndSendProposalMessage(final long headerTimeStampSeconds) {
-    final Block block = blockCreator.createBlock(headerTimeStampSeconds).getBlock();
+    final Block block;
+
+    final Optional<ScheduledProtocolSpec.Hardfork> shanghaiFork =
```
jframe, line 123: Think a better approach might be to disable the withdrawals functionality for QBFT/IBFT.

You could do that by overriding the protocol schedule in BaseBftProtocolScheduleBuilder in the applyBftChanges method and configuring the protocol schedule to not use a withdrawalsProcessor and use the ProhibitedWithdrawals for the withdrawals validator i.e. same as before Shanghai.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftRoundFactory.java
```diff
@@ -110,7 +110,7 @@ public IbftRound createNewRoundWithState(
         roundState,
         blockCreator,
         protocolContext,
-        protocolSchedule.getByBlockNumber(roundIdentifier.getSequenceNumber()).getBlockImporter(),
+        protocolSchedule.getByBlockHeader(parentHeader).getBlockImporter(),
```
jframe, line 113: This needs to get the block importer based on the block we are building represented by the sequence number, not the parent block.

Instead of passing in the block importer could just get the block importer from the protocol schedule in the IbftRound importBlockToChain method.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/validation/MessageValidatorFactory.java
```diff
@@ -81,7 +81,7 @@ private SignedDataValidator createSignedDataValidator(
   public MessageValidator createMessageValidator(
       final ConsensusRoundIdentifier roundIdentifier, final BlockHeader parentHeader) {
     final BlockValidator blockValidator =
-        protocolSchedule.getByBlockNumber(roundIdentifier.getSequenceNumber()).getBlockValidator();
+        protocolSchedule.getByBlockHeader(parentHeader).getBlockValidator();
```
jframe, line 84: Similarly to the IbftRoundFactory this needs to get the block validator based on the block we are building represented by the sequence number, not the parent block.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftRound.java
matthew1001: Yeah that's a good suggestion. It does seem cleaner to avoid validating a feature that QBFT doesn't support, rather than provide the withdrawal field so that validation passes. I'll possibly start a different draft PR that uses the alternative approach you've suggested (along with your other comments below) but I'll see how I go.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftRoundFactory.java
jframe: Yeah, your right that the sequence number is the same as the chain height/block number. If a timestamp is used to schedule the Shanghai fork then there is no one-to-one mapping from timestamp to sequence/block number as there could be multiple rounds involved to create that block.

Why was suggesting to resolve the block importer from the protocol schedule in the IbftRound importBlockToChain is that happens once the block is finalised and we have a timestamp so at that point the ProtocolSchedule.getByBlockHeader could be used.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftRound.java
```diff
@@ -312,6 +315,11 @@ private void importBlockToChain() {
           blockToImport.getHash());
     }
     LOG.trace("Importing block with extraData={}", extraData);
+    final BlockImporter blockImporter =
+        ((BftProtocolSchedule) protocolSchedule)
+            .getByBlockNumberAndTimestamp(
```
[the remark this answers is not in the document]
matthew1001: Fixed

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/validation/MessageValidator.java
```diff
@@ -93,6 +95,13 @@ public boolean validateProposal(final Proposal msg) {
   }
 
   private boolean validateBlock(final Block block) {
+
+    final BlockValidator blockValidator =
+        ((BftProtocolSchedule) protocolSchedule)
+            .getByBlockNumberAndTimestamp(
```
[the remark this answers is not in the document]
matthew1001: Fixed

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftRound.java
```diff
@@ -342,6 +346,11 @@ private void importBlockToChain() {
           blockToImport.getHash());
     }
     LOG.trace("Importing proposed block with extraData={}", extraData);
+    final BlockImporter blockImporter =
+        ((BftProtocolSchedule) protocolSchedule)
```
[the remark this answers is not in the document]
matthew1001: Fixed

matthew1001:
Thanks for the latest comments @jframe. I think the latest commits resolve them - let me know what you think.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/validation/MessageValidator.java
```diff
@@ -47,19 +48,19 @@ public class MessageValidator {
    *
    * @param signedDataValidator the signed data validator
    * @param proposalConsistencyValidator the proposal consistency validator
-   * @param blockValidator the block validator
+   * @param protocolSchedule the proposal schedule
```
jframe, line 51: javadoc for this should be "the protocol schedule"

consensus/ibft/src/test/java/org/hyperledger/besu/consensus/ibft/validation/MessageValidatorTest.java
```diff
@@ -101,6 +106,12 @@ public void setup() {
             mockBftCtx,
             Optional.empty());
 
+    lenient()
+        .when(protocolSchedule.getByBlockNumberOrTimestamp(anyLong(), anyLong()))
```
jframe, line 110: Thought that the MessageValidator was using ProtocolSchedule.getByBlockHeader now?

consensus/qbft/src/integration-test/resources/genesis_validator_contract_shanghai.json
```diff
@@ -0,0 +1,46 @@
...
+			"balance": "0",
+			"code": "608060405234801561001057600080fd5b5060043610610048576000357c010000000000000000000000000000000000000000000000000000000090048063b7ab4db51461004d575b600080fd5b61005561006b565b604051610062919061017e565b60405180910390f35b606060008054806020026020016040519081016040528092919081815260200182805480156100ef57602002820191906000526020600020905b8160009054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190600101908083116100a5575b5050505050905090565b60006101058383610111565b60208301905092915050565b61011a816101d9565b82525050565b600061012b826101b0565b61013581856101c8565b9350610140836101a0565b8060005b8381101561017157815161015888826100f9565b9750610163836101bb565b925050600181019050610144565b5085935050505092915050565b600060208201905081810360008301526101988184610120565b905092915050565b6000819050602082019050919050565b600081519050919050565b6000602082019050919050565b600082825260208201905092915050565b60006101e4826101eb565b9050919050565b600073ffffffffffffffffffffffffffffffffffffffff8216905091905056fea26469706673582212206d880cf012c1677c691bf6f2f0a0e4eadf57866ffe5cd2d9833d3cfdf27b15f664736f6c63430008060033",
```
jframe, line 25: Would give more confidence if we used a Shanghai compiled contract in the Shanghai tests then we would know it's using the Shanghai EVM opcodes

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
```diff
@@ -144,6 +144,10 @@ public boolean validate(final SignedData<ProposalPayload> signedPayload) {
   }
 
   private boolean validateBlock(final Block block) {
+    if (blockValidator == null) {
+      throw new RuntimeException("Block validation not possible, no block validator.");
```
jframe, line 148: nit: this should never happen so perhaps throwing an IllegalStateException instead would be clearer

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalValidator.java
```diff
@@ -93,6 +96,16 @@ public ProposalValidator(
    */
   public boolean validate(final Proposal msg) {
 
+    checkState(
+        protocolSchedule instanceof BftProtocolSchedule,
+        "Wrong class type for protocol schedule, requires BftProtocolSchedule");
+
+    final BlockValidator blockValidator =
```
jframe, line 103: I think in this case you can also use the ProtocolSchedule.getByBlockHeader as we have the block header on the received proposal message block i.e. ```protocolSchedule.getByBlockHeader(                msg.getBlock().getHeader())```

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/RoundChangeMessageValidator.java
```diff
@@ -94,6 +96,13 @@ public boolean validate(final RoundChange msg) {
   }
 
   private boolean validateBlock(final Block block) {
+
+    final BlockValidator blockValidator =
```
jframe, line 100: Can also use ProtocolSchedule.getByBlockHeader here as we have the block header

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/validation/MessageValidator.java
matthew1001: 🤦 - fixed

consensus/qbft/src/integration-test/resources/genesis_validator_contract_shanghai.json
matthew1001: Agreed. I've recompiled with `shanghai`.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalPayloadValidator.java
matthew1001: Decided to replace with `checkState(...)` which results in an `IllegalStateException` if the predicate fails.

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/RoundChangeMessageValidator.java
matthew1001: Done 👍

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalValidator.java
matthew1001: Done 👍

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/validation/ProposalValidator.java
```diff
@@ -93,6 +96,13 @@ public ProposalValidator(
    */
   public boolean validate(final Proposal msg) {
 
+    checkState(
+        protocolSchedule instanceof BftProtocolSchedule,
+        "Wrong class type for protocol schedule, requires BftProtocolSchedule");
```
jframe, line 101: nit: don't need this check anymore

jframe (approved):

PR #6354, opened 2024-01-05
Improve the high spec flag

ahamlat:
## PR description
Improve the high spec flag (`Xplugin-rocksdb-high-spec-enabled`) by limiting it to only few column families.

We can see below the difference in memory usage with this new implementation
image


macfarla (approved):
LGTM

PR #6362, opened 2024-01-07
migrate clique tests fully to junit5

### LINKED ISSUE #6261: migrate clique and bft ATs to junit 5
Add dependency for junit.jupiter (junit5)
Migrate per https://blogs.oracle.com/javamagazine/post/migrating-from-junit-4-to-junit-5-important-differences-and-benefits
Remove dependency on junit 4
Here's an example https://github.com/hyperledger/besu/pull/5525/files

macfarla:
fixes #6261


gfukushima (approved):
lgtm, not sure if I like the fact we're naming the AcceptanceTestBase after the version of junit being used. I guess this is temporary until we finish the migration.

PR #6363, opened 2024-01-08
Import export trie log

gfukushima:
## PR description
This PR extends the x-trie-log subcommand to allow exporting and importing of a trie log.

This PR is built on top of https://github.com/hyperledger/besu/pull/6303


besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogSubCommand.java
```diff
@@ -123,6 +131,78 @@ public void run() {
...
+    @SuppressWarnings("unused")
+    @ParentCommand
+    private TrieLogSubCommand parentCommand;
+
+    @SuppressWarnings("unused")
+    @CommandLine.Spec
+    private CommandLine.Model.CommandSpec spec; // Picocli injects reference to command spec
+
+    @CommandLine.Option(
+        names = "--trie-log-hash",
```
jframe, line 189: Can the exported file contain more than one trielog? Would of expected the import to specify the filename and just import all trielogs in that file

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogSubCommand.java
```diff
@@ -123,6 +131,78 @@ public void run() {
...
+        paramLabel = DefaultCommandValues.MANDATORY_LONG_FORMAT_HELP,
+        description = "The hash of the block you want to export the trie log.",
+        arity = "1..1")
+    private String trieLogHash;
+
```
jframe, line 156: Similar to other export commands, it would be good to include the filename as a command line option

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -357,5 +351,30 @@ static void printCount(final PrintWriter out, final TrieLogCount count) {
...
+      final BonsaiWorldStateKeyValueStorage rootWorldStateStorage,
+      final Path dataDirectoryPath,
+      final Hash trieLogHash) {
+    final String trieLogFile =
+        dataDirectoryPath.resolve(DATABASE_PATH).resolve(trieLogHash.toString()).toString();
+
+    var trieLog = readTrieLogsFromFile(trieLogFile);
+
+    var updater = rootWorldStateStorage.updater();
+    trieLog.forEach((key, value) -> updater.getTrieLogStorageTransaction().put(key, value));
```
jframe, line 364: Is the foreach necessary, can there be more than one trielog for an import?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -357,5 +351,30 @@ static void printCount(final PrintWriter out, final TrieLogCount count) {
...
   }
 
+  static void importTrieLog(
+      final BonsaiWorldStateKeyValueStorage rootWorldStateStorage,
+      final Path dataDirectoryPath,
+      final Hash trieLogHash) {
+    final String trieLogFile =
+        dataDirectoryPath.resolve(DATABASE_PATH).resolve(trieLogHash.toString()).toString();
+
+    var trieLog = readTrieLogsFromFile(trieLogFile);
```
jframe, line 361: nit: not sure how important this is so feel free to ignore it doesn't make sense. But having the format in RLP to match the storage in the database trielog column family might make comparison easier without the extra encoding of the IdentityMap. Depends on the purpose of this tool.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogSubCommand.java
```diff
@@ -123,6 +131,78 @@ public void run() {
...
+    @SuppressWarnings("unused")
+    @CommandLine.Spec
+    private CommandLine.Model.CommandSpec spec; // Picocli injects reference to command spec
+
+    @CommandLine.Option(
+        names = "--trie-log-hash",
+        paramLabel = DefaultCommandValues.MANDATORY_LONG_FORMAT_HELP,
+        description = "The hash of the block you want to import the trie log.",
+        arity = "1..1")
+    private String trieLogHash;
```
jframe, line 193: rename to blockHash? the variable name suggests this is the hash of the trielog

jframe, line 193: Could be useful to be able to specify a block number instead of block hash. But only a suggestion not sure which one we will use more in practice

gfukushima: I think block hash is prob better is this case, as you could eventually request by number and not have the right block hash if your node forked at that specific block requested.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -97,16 +98,15 @@ private static void processTrieLogBatches(
...
       final long firstBlockOfBatch = chainHeight - ((batchNumber - 1) * BATCH_SIZE);
-
       final long lastBlockOfBatch =
           Math.max(chainHeight - (batchNumber * BATCH_SIZE), lastBlockNumberToRetainTrieLogsFor);
-
       final List<Hash> trieLogKeys =
           getTrieLogKeysForBlocks(blockchain, firstBlockOfBatch, lastBlockOfBatch);
 
-      saveTrieLogBatches(batchFileNameBase, rootWorldStateStorage, batchNumber, trieLogKeys);
+      LOG.info("Saving trie logs to retain in file (batch {})...", batchNumber);
```
siladu, line 108: is it useful to say batch x of y here?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogSubCommand.java
siladu: Maybe consider using a list of blocks? Perhaps next time we need to update more than one 🤷

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
siladu: I'd lean towards a more readable format rather than binary as it's hard to work with. Agree RLP is a more useful middle ground than custom binary (if that's what this is).
e.g. I wonder if this could be used to import test data...compare with RlpBlockImporter.java and JsonBlockImporter

besu/src/test/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelperTest.java
```diff
@@ -94,10 +93,11 @@ public static void setup() throws IOException {
         .getTrieLogStorageTransaction()
         .put(blockHeader5.getHash().toArrayUnsafe(), Bytes.fromHexString("0x05").toArrayUnsafe());
     updater.getTrieLogStorageTransaction().commit();
+
+    createDirectory();
```
siladu, line 97: Could use `@TempDir`?

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -42,6 +42,11 @@ public class DataStorageOptions implements CLIOptions<DataStorageConfiguration>
   private static final String BONSAI_STORAGE_FORMAT_MAX_LAYERS_TO_LOAD =
       "--bonsai-historical-block-limit";
 
+  private static final String BONSAI_TRIE_LOG_PRUNING_ENABLED =
+      "--Xbonsai-trie-log-pruning-enabled";
+
+  private static final String BONSAI_LIMIT_TRIE_LOGS_ENABLED = "--Xbonsai-limit-trie-logs-enabled";
```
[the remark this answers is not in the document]
gfukushima: moved to unstable

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: This was moved to outside the method so we don't log batches for this single trie log export.

besu/src/test/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelperTest.java
gfukushima: It is using, but I forgot to remove this form the setup. thanks

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: I'd rather have us to extend the current feature to allow the export to RLP reusing this feature if necessary, it isn't quite simple to do that right now and it would impact the sub-command prune logic as well.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogSubCommand.java
```diff
@@ -147,21 +148,38 @@ static class ExportTrieLog implements Runnable {
 
     @CommandLine.Option(
         names = "--trie-log-block-hash",
-        description = "The hash of the block you want to export the trie log of.",
+        description = "The hash of the block you want to import the trie log of",
```
siladu: Think name and description need updating now it's a list.

gfukushima: Updated the description. Think the name should still be the same following the same criteria of `rpc-http-api` for example

siladu: Makes sense, thanks!

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -97,16 +97,15 @@ private static void processTrieLogBatches(
       final String batchFileNameBase) {
 
     for (long batchNumber = 1; batchNumber <= numberOfBatches; batchNumber++) {
-
+      final String batchFileName = batchFileNameBase + "-" + batchNumber;
```
garyschulte, line 100: this might make more sense to have first/last block numbers included in the filename.  Otherwise it won't be clear what is actually in the files after an export

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -285,11 +279,10 @@ private static void saveTrieLogsInFile(
   }
 
   @SuppressWarnings("unchecked")
-  private static IdentityHashMap<byte[], byte[]> readTrieLogsFromFile(
-      final String batchFileNameBase, final long batchNumber) {
+  static IdentityHashMap<byte[], byte[]> readTrieLogsFromFile(final String batchFileName) {
```
garyschulte, line 282: Non blocking feedback - 

I am all for code reuse, but if we are going to allow for arbitrary import and export, the import files should be more readable and "createable".  


The ObjectOutputStream seems fine for backup/recovery of a pruning process, but when part of a general import/export process this file format is too inscrutable IMO.

At least for import/export we should serialize/deserialize these as json maps.  Key as the hash string, and the trielog itself as hex (or as a rich json object if we wanted to be super transparent).  In addition to being a bit more introspectable, it would allow us to create and import our own handcrafted trielogs when debugging

garyschulte (commented):
great feature :+1:  feedback mostly about import/export introspection

garyschulte:
another cadillac feature would be the ability to specify a URL for the import files.   This could have been used to recover nodes from the Jan 6 halting event.  We could have provided import command instructions and a URL to delivery the correct trielog.

besu/src/test/java/org/hyperledger/besu/cli/options/stable/DataStorageOptionsTest.java
```diff
@@ -34,17 +34,17 @@ public void bonsaiTrieLogPruningLimitOption() {
...
-        "--Xbonsai-trie-log-pruning-limit",
+        "--Xbonsai-limit-trie-logs-enabled",
+        "--Xbonsai-trie-logs-pruning-limit",
         "1");
   }
 
   @Test
   public void bonsaiTrieLogPruningLimitShouldBePositive() {
     internalTestFailure(
         "--Xbonsai-trie-log-pruning-limit=0 must be greater than 0",
```
siladu, line 45:
```suggestion
        "--Xbonsai-trie-logs-pruning-limit=0 must be greater than 0",
```

siladu: There's a few other Strings where the rename needs to happen too.

siladu (approved):
Looks good. Rename tidy-up could happen in a subsequent PR

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
jframe: The batch filenames aren't used as part of import/export subcommands the filename is taken from command line args instead. This is only used for the prune subcommand

besu/src/test/java/org/hyperledger/besu/cli/options/stable/DataStorageOptionsTest.java
jframe: will do this in a followup PR

siladu: Should be covered by https://github.com/hyperledger/besu/pull/6445

PR #6364, opened 2024-01-08
Remove junit platform directive from mainnet ats

macfarla:


macfarla:
can't do this until mainnet ATs are fully converted to junit 5

Issue #6327, opened 2023-12-21
Alias --fast-sync-min-peers to --sync-min-peers


non-fungible-nelson:
### Description
The flag `--fast-sync-min-peers` actually impacts SNAP and CHECKPOINT sync on non-POS networks. This ticket should remedy that by aliasing it with a new `--sync-min-peers`. We also need to update the documentation to reflect this fact.  <img width="1780" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1IvQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3a0ab4a9739d97e199e6fb838993e9b3c477886a/image.png" alt="image.png" /> 
Something like "This flag only impacts non-Proof of Stake networks."

One other thought for `--sync-min-peers`. FULL sync could be changed to tolerate it and just give a warning, instead of preventing node startup. Other options seem to be treated that way and that slightly more permissive approach is can be helpful in certain circumstances.

PR #6372, opened 2024-01-09
Added alias --sync-min-peers for --fast-sync-min-peers

macfarla:
Add an alias for fast-sync-min-peers

refs #6327


macfarla:
help now shows
```
      --sync-min-peers, --fast-sync-min-peers=<INTEGER>
                             Minimum number of peers required before starting
                               sync. Has effect only on non-PoS networks.
                               (default: 5)
```

jframe (approved):
LGTM

PR #6375, opened 2024-01-09
Add a sender cache to Transaction

ahamlat:
## PR description
This PR will create a new cache within the Transaction class to house all computed senders for their respective transaction hashes. 
When receiving the newPayload call, it calculates asynchronously and in parallel all the transaction hashes and senders.
This PR is expected to improve block processing time, given that either the transaction sender is determined when Besu fills its transaction pool or with the asynchronous tasks. 

**The cache memory footprint**

The cache is defined with this pair hash -> address, for 100_000 entries, it consumes currently less than 20 MiB.
image

We can find below the CPU profiling with and without this PR.

**Without this PR**
image

**With this PR**
image

We can notice also an improvement in terms of block processing time 

image


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -75,6 +77,9 @@ public class Transaction
 
   public static final BigInteger TWO = BigInteger.valueOf(2);
 
+  public static final Cache<Hash, Address> senderCache =
+      CacheBuilder.newBuilder().recordStats().maximumSize(100_000L).build();
```
fab-10, line 81: how much more memory will take a full cache? and the field should be private

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -409,16 +414,23 @@ public Optional<BigInteger> getChainId() {
    * @return the transaction sender
    */
   @Override
-  public Address getSender() {
+  public synchronized Address getSender() {
```
fab-10, line 417: probably synchronized could enforce only when `sender==null`

ahamlat: As I add another optimization to precalculate transations senders and hashes in parallel an asynchronously, I deleted synchronized to avoid too much overhead

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
ahamlat: I made it public to expose it outside the class and the package in case we need it, but you're right, it's better to make it private and update it when need more visibility. I'll share the cache metrics in this PR.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -178,6 +180,17 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
               .map(Bytes::fromHexString)
               .map(in -> TransactionDecoder.decodeOpaqueBytes(in, EncodingContext.BLOCK_BODY))
               .collect(Collectors.toList());
+      transactions.forEach(
```
fab-10, line 183: is this the same as the code below? or is it more performant to explicitly call the CompletableFuture?
```
      transactions.parallelStream().forEach(
          transaction ->
                    Address sender = transaction.getSender();
                    LOG.atTrace()
                        .setMessage("The sender for transaction {} is calculated : {}")
                        .addArgument(transaction.getHash())
                        .addArgument(sender)
                        .log();
                  );
```

fab-10 (approved):
LGTM, just a question

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
ahamlat: Good question, the difference is that parallelStream() will use the main thread as part of the threads that will calculate the transaction senders and hashes, thus the calculation is parallel but not asynchronous. Using CompletableFutures the way it is done in the other code (without join) will make these calculations asynchronous with the thread that is executing the newPayload call.

fab-10: I see, thanks for the explanation.
So a possibly _more asynchronous_ version could be ?
```
CompletableFuture.runAsync(
      transactions.parallelStream().forEach(
          transaction ->
                    Address sender = transaction.getSender();
                    LOG.atTrace()
                        .setMessage("The sender for transaction {} is calculated : {}")
                        .addArgument(transaction.getHash())
                        .addArgument(sender)
                        .log();
                  ));
```

ahamlat: Yes, I guess there is not too much difference as iterating over 200 transactions is extremely fast. One other difference is that we can imagine using a completely different thread pool with the first solution if the forkjoin pool is used a lot, by adding predefined executor to the runAsync method.
CompletableFuture.runAsync takes a runnable, so I added () ->  to your suggestion. I thing it is worth testing your suggestion but also the current implementation (in this PR) with an executor.

```
CompletableFuture.runAsync(() -> 
    transactions.parallelStream().forEach(transaction -> {
        Address sender = transaction.getSender();
        LOG.atTrace()
            .setMessage("The sender for transaction {} is calculated : {}")
            .addArgument(transaction.getHash())
            .addArgument(sender)
            .log();
    })
);
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -178,6 +180,20 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+                  .scheduleFutureTask(
+                      () -> {
+                        Address sender = transaction.getSender();
+                        LOG.atTrace()
+                            .setMessage("The sender for transaction {} is calculated : {}")
+                            .addArgument(transaction.getHash())
+                            .addArgument(sender)
+                            .log();
+                      },
+                      Duration.ofSeconds(1)));
```
fab-10, line 196: This way the execution will start 1 seconds later, are you sure this is what you want?
Could be better to use `scheduleTxWorkerTask` or `scheduleServiceTask`?

ahamlat: Nice catch, no I didn't want to delay the execution, I changed to scheduleTxWorkerTask that looks more convenient in this case.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -184,16 +184,15 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
           transaction ->
               mergeCoordinator
                   .getEthScheduler()
-                  .scheduleFutureTask(
+                  .scheduleTxWorkerTask(
```
fab-10, line 187: last suggestions, sync txWorker by default has a limit of 2 threads, so it could not leverage all the cores, for that you can change the number of thread to reflect the number of cores, or use the services executor that uses a cached thread executor with not predefined limit on threads

ahamlat: This method is not using the sync txWorker but the txWorker, which is a thread pool of a minimum 1 and a maximum number of threads 1000.

fab-10 (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -411,14 +416,21 @@ public Optional<BigInteger> getChainId() {
   @Override
   public Address getSender() {
```
sebastien-perpignane, line 417: I think you could improve readability of this method this way:

```java
  @Override
  public Address getSender() {
    if (sender == null) {
      Optional<Address> cachedSender = Optional.ofNullable(senderCache.getIfPresent(getHash()));
      sender = cachedSender.orElseGet(this::computeSender);
    }
    return sender;
  }

  private Address computeSender() {
    final SECPPublicKey publicKey =
            signatureAlgorithm
                    .recoverPublicKeyFromSignature(getOrComputeSenderRecoveryHash(), signature)
                    .orElseThrow(
                            () ->
                                    new IllegalStateException(
                                            "Cannot recover public key from signature for " + this));
    Address calculatedSender = Address.extract(Hash.hash(publicKey.getEncodedBytes()));
    senderCache.put(this.hash, calculatedSender);
    return calculatedSender;
    return Address.extract(Hash.hash(publicKey.getEncodedBytes()));
  }
```

ahamlat: Thanks @sebastien-perpignane, I will update the code with your suggestion

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -178,6 +179,19 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+      transactions.forEach(
+          transaction ->
+              mergeCoordinator
+                  .getEthScheduler()
+                  .scheduleTxWorkerTask(
+                      () -> {
+                        Address sender = transaction.getSender();
+                        LOG.atTrace()
+                            .setMessage("The sender for transaction {} is calculated : {}")
+                            .addArgument(transaction.getHash())
```
fab-10, line 191:
```suggestion
                            .addArgument(transaction::getHash)
```

fab-10 (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
ahamlat: I prefer the initial one to be honest, I think we should keep it simple, transaction::getHash will be transformed into transaction.getHash so why not just use transaction.getHash.

fab-10: there is a semantic difference when using `LOG.atTrace` with lambda or direct method call:

- `.addArgument(transaction.getHash())` means that the methos is always executed even if the log level is > TRACE
- .addArgument(transaction::getHash) means that the method is only executed if the level is <= TRACE

and since `transaction.getHash()` is potentially expensive method, it is worth to avoid calling it when the log is not done.

PR #6378, opened 2024-01-09
Upgrade `com.fasterxml.jackson` dependencies

fab-10:
## PR description

Jackson 2.15 by default limits the length of input string to a value that is lower than the string length used in some of our test, so now we need to explicitly tune this value. Setting max int since this is only related to tests.


macfarla (approved):

jframe (approved):
LGTM

PR #6379, opened 2024-01-09
Ignore bin files when running the spdx license check

mbaxter:
## PR description
Ignore generated files in `bin/` directories when running the SPDX license check.  Visual Studio Code generates build files in various bin directories, which causes this check to fail.


macfarla (approved):
LGTM

macfarla:
welcome back to besu @mbaxter 💪

PR #6381, opened 2024-01-10
Introduce TransactionEvaluationContext to pass data between selectors and plugin

fab-10:
## PR description

Some tx selector plugins needs more context about the current tx being evaluated, for example the current minGasPrice, or the effective price of the tx and possibly more data in the future, for this we introduce the TransactionEvaluationContext to hold that data and pass it across all the selectors and the plugin. Another advantage of this context is to avoid to compute multiple time the same data, in case it is needed in more that one selector.


macfarla (approved):
LGTM

PR #6390, opened 2024-01-11
Use mining beneficiary from protocol spec in TraceServiceImpl

### LINKED ISSUE #6389: Use correct mining beneficiary in TraceServiceImpl
The TraceServiceImpl has the coinbase from the block header hard coded as mining beneficiary. This is not correct for Clique networks as the mining beneficiary is recovered from the signed message in extra data. The mining beneficiary needs to be correctly calculated for all consensus options.

daniellehrner:
## PR description

instead of using coinbase from block header as mining beneficiary, we use the protocol spec to retrieve it

## Fixed Issue(s)
fixes #6389


jframe (approved):
LGTM

Gabriel-Trintinalia (approved):

PR #6392, opened 2024-01-12
reduce machine size for clique/BFT ATs

macfarla:
same change as was made in #6384 for permissioning ATs


jframe (approved):

fab-10 (approved):
LGTM since still faster than unitTests and referenceTests

PR #6394, opened 2024-01-12
Trie log prune using TrieLogEvent

gfukushima:
## PR description
This PR aims to make the pruning logic async as this feature does not need to be in the synchronous.

## Fixed Issue(s)

Part of #5390 

## Testing

Tested locally on a small chain.

Tested with a fresh holesky sync...

TrieLogPruning happening on an EthSchedule-Services thread...
```
{"@timestamp":"2024-01-19T03:04:01,050","level":"TRACE","thread":"EthScheduler-Services-4","class":"TrieLogPruner","message":"adding trie log to queue for later pruning blockNumber 765656; blockHash 0xfd2e15c6ed1d00ee050210f427869d9ff6773555e5b738888fd9855b3251ad31","throwable":""}
{"@timestamp":"2024-01-19T03:04:01,050","level":"TRACE","thread":"EthScheduler-Services-4","class":"TrieLogPruner","message":"min((chainHeadNumber: 765655 - numBlocksToRetain: 512) = 765143, finalized: 765587)) = retainAboveThisBlockOrFinalized: 765143","throwable":""}
{"@timestamp":"2024-01-19T03:04:01,050","level":"TRACE","thread":"EthScheduler-Services-4","class":"TrieLogPruner","message":"pruned 1 trie logs for blocks {765143=[0x607bc6df9eabdb70a153dc23983a384c62cfd05a46a60584716cc4114c6c0ba3]}","throwable":""}
{"@timestamp":"2024-01-19T03:04:01,050","level":"DEBUG","thread":"EthScheduler-Services-4","class":"TrieLogPruner","message":"pruned 1 trie logs from 1 blocks","throwable":""}
{"@timestamp":"2024-01-19T03:04:01,052","level":"INFO","thread":"vert.x-worker-thread-0","class":"AbstractEngineNewPayload","message":"Imported #765,656 / 18 tx / 16 ws / base fee 1.04 gwei / 4,272,115 (14.2%) gas / (0xfd2e15c6ed1d00ee050210f427869d9ff6773555e5b738888fd9855b3251ad31) in 0.093s. Peers: 25","throwable":""}
```

Post sync, besu (on holesky) is only maintaining a single EthScheduler-Services thread so thread contention is low (it's a cachedThreadPool).
Screenshot 2024-01-19 at 1 06 04 pm

Tasks per second during and after sync...

Screenshot 2024-01-19 at 1 07 30 pm


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/trielog/TrieLogPruner.java
```diff
@@ -166,34 +167,17 @@ int pruneFromQueue() {
...
-    }
-
-    @Override
-    int pruneFromQueue() {
-      // no-op
-      return -1;
+  @Override
+  public void onTrieLogAdded(final TrieLogEvent event) {
+    if (TrieLogEvent.Type.ADDED.equals(event.getType())) {
+      final TrieLogAddedEvent addedEvent = (TrieLogAddedEvent) event;
```
jframe, line 173: This cast doesn't look necessary

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/trielogs/TrieLogEvent.java
```diff
@@ -14,12 +14,17 @@
...
 /** A TrieLog event. */
 public interface TrieLogEvent {
   /** The type of the event. */
   enum Type {
     /** TrieLog added event type */
-    ADDED
+    ADDED,
+    /** dummy value (this comment required by javadoc) * */
+    @VisibleForTesting
+    UNKNOWN
```
jframe, line 27: Would prefer to not expose this on the public API. Is there another way of testing that it only works for the ADDED type?

siladu: The options I considered were:
1. This
2. Subclass TrieLogEvent for the test, but is essentially duplicating this part of the code so not really a good test.
3. No test for this case.

siladu: To be honest, I'm not sure why there is both a subclass representing the event type as well this enum...maybe don't need the enum at all?

jframe: Maybe no test is ok. Right now it's not possible to create a TrieLogEvent with anything other than an ADDED event

siladu: removed the test

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/trielog/TrieLogPruner.java
```diff
@@ -50,11 +53,13 @@ public class TrieLogPruner {
   public TrieLogPruner(
       final BonsaiWorldStateKeyValueStorage rootWorldStateStorage,
       final Blockchain blockchain,
+      final Consumer<Runnable> executeAsync,
```
siladu, line 56: Might have been more obvious to pass EthScheduler in here but that would mean making the `core` module depend on the `eth` module which is circular.

I do like the way this is decoupled from the EthScheduler implementation anyway, even if the type is a little unobvious.

jframe: The `Executor` would be clearer than using a `Consumer<Runnable>` type

jframe (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/trielog/TrieLogPruner.java
siladu: As discussed, while it happens to be the same type, the implementation I'm using does not actually represent an Executor implementation.

PR #6401, opened 2024-01-14
log a different warning if pruning is enabled with BONSAI

### LINKED ISSUE #6400: log "ignored" warning if pruning is enabled with BONSAI
(reported by a user in discord)

currently you get 

Forest pruning is deprecated and will be removed soon. To save disk space consider switching to Bonsai data storage format. 

Except my node has, and always had the following option configured --data-storage-format=BONSAI

which is confusing. 

Better would be the above if FOREST

and if BONSAI, something like "pruning is ignored with BONSAI storage"

macfarla:
log "ignored" if enabled with BONSAI vs "deprecated" if enabled with FOREST

fixes #6400


gfukushima (approved):
lgtm

PR #6406, opened 2024-01-15
cliqueBft AT task flakiness - extend no_output_timeout

macfarla:
task is flaky eg 20 min with no response https://app.circleci.com/pipelines/github/hyperledger/besu/26348/workflows/3c01f02c-3127-4b9c-965f-a869096ef3a5/jobs/173374 and https://app.circleci.com/pipelines/github/hyperledger/besu/26350/workflows/1c5f785f-f62b-4fa7-97cf-097dde63325c/jobs/173389

increase `no_output_timeout` to 30min

~~reverts #6392 - seems to have caused flakiness~~


jframe:
Could also just increase the `no_output_timeout`. I increased it to 25m in my PR as kept failing with the 20m it has right now. Might be better than increasing the machine executor size.

jframe (approved):
LGTM

PR #6407, opened 2024-01-15
Allow Xpoa-block-txs-selection-max-time to be greater than 100%

fab-10:
## PR description

In #6044 `Xpoa-block-txs-selection-max-time` options has been introduced to limit the block creation time to a fraction of the genesis block time, but there are cases where it is useful to let the block creation time last more, so with this PR is it possible to specify value greater than 100%


pinges (commented):
Could you explain why it would make sense to do that please. 
Otherwise, I think that the changes do achieve what you want to do.

fab-10:
@pinges because there could be existing networks that are used to have txs that take more than the genesis block time to process, and/or do not care too much about a fixed block time, but more about a target block time, allowing outliers.

fab-10:
Close in favor of #6423

PR #6412, opened 2024-01-16
[CI] publish tasks don't depend on test tasks

macfarla:
occasional flakiness eg https://app.circleci.com/pipelines/gh/hyperledger/besu/26365/workflows/c1646864-be7d-4a67-95dc-ab22876dbdde should not stop publish tasks

all required test tasks have already passed on PR runs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -443,10 +443,6 @@ workflows:
                 - /^release-.*/
           requires:
             - assemble
-            - integrationTests
-            - unitTests
-            - acceptanceTests
-            - referenceTests
             - buildDocker
       - publishDocker:
           filters:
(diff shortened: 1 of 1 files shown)


siladu:
> all required test tasks have already passed on PR runs

This feels that a potentially dangerous assumption at a glance. Not sure if besu does this, but some of our other repos only run tests against main branch.

siladu:
> We run all tests on PRs and then again on main.

This begs the question, do we need to run the tests on main at all if we're happy with the PR passing and are not going to prevent progress when main fails?

...but not running tests on main would make me feel uncomfortable...I am struggling to convince myself there will never be rare events that cause genuine failures on main but not on the PR.

jflo:
This convo is relevant to #6427 . In my proposed version, we do not re-run all the tests on main after it has been merged. This would be trivial to implement but it seems like a (rather large) waste of compute and clock time since the pr tests are all run against the merge result anyway. In a GHA centric world, the publish task would only run on a push to main, after checking that the required test statuses were present for the pushed head ref.

macfarla:
that makes sense @jflo - even better to not rerun the tests when it's exact duplication to run again on main. I'll close this since your PR is close

PR #6415, opened 2024-01-16
use reference tests to verify trielog generation

matkt:
## PR description

This PR use reference tests to validate trielog generation. Like that we will validate our trielog generation for all of the tests we have in the references tests suite for past and future EIPs

I was able to validate that this PR can detect the issues we had before on mainnet by running this test "GeneralStateTests/stSStoreTest/InitCollision.json" with and without [the last ](https://github.com/hyperledger/besu/pull/6359) fix


ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/BonsaiReferenceTestWorldState.java
```diff
@@ -89,6 +92,108 @@ public ReferenceTestWorldState copy() {
...
+                .formatted(parentStateRoot, generatedRootHash));
+      }
+    }
+  }
+
+  private void generateTrieLogFromState(
+      final BlockHeader blockHeader,
+      final BonsaiReferenceTestUpdateAccumulator originalUpdater,
+      final boolean isFrozen) {
+    // generate trielog
```
garyschulte, line 173: perhaps add to this comment that generating a trielog is a side effect of persisting the worldstate

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/BonsaiReferenceTestWorldState.java
```diff
@@ -101,6 +206,12 @@ protected void verifyWorldStateRoot(final Hash calculatedStateRoot, final BlockH
     // The test harness validates the root hash, no need to validate in-line for reference test
   }
 
+  @Override
+  public void persist(final BlockHeader blockHeader) {
+    validateStateRolling(blockHeader);
```
garyschulte, line 211: IDK how I feel about having this behavior _always_, especially in light of the evm performance metrics that are being derived from evmtool runs.  I don't know whether evmtool tests typically call persist, but if they do it would certainly be additional overhead to do state rolling.  @shemnon ?

garyschulte (commented):
Solid reference testing addition 👍 

Depending on feedback from Danno, we may want to make this a configurable behavior that only affects our CI reference test runs, and default this behavior to "off" so we do not impact evmtool runs.

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/BonsaiReferenceTestWorldState.java
matkt: I thought it was just used for reference tests given the name. I will make it configurable and activate it only for the reference tests

matkt: I will add a specific method for that and not use persist. It will avoid problems like that

garyschulte: another option would be to subclass the reference test worldstate for evmtool and override the state rolling.  We could fix the class name expectation that way too

shemnon: EVMTool actually uses in-memory Forest, perhaps it should be tweaked? - https://github.com/hyperledger/besu/blob/ccab8571e5648adc00a7cec4d9a9d28155508bef/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BlockchainModule.java#L67

Guido's fuzzing uses in-memory Bonsai

matkt: Yes maybe we need to update the evmtool to use bonsai or forest

matkt: Hi @garyschulte @shemnon I updated the code to have another method for this kind of check. I think it's better to keep persist method persisting and having something else for the trielog check . Let me know what do you think about the last version

matkt: for context https://github.com/hyperledger/besu/pull/6415/files#diff-b9e0afe79b33e1367ce04f1720fcccf35976bf2fbe93af84eb67bab40d8a0e9aR195

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContextTest.java
```diff
@@ -48,7 +48,7 @@
 import org.hyperledger.besu.ethereum.mainnet.MainnetProtocolSchedule;
 import org.hyperledger.besu.ethereum.mainnet.ProtocolSchedule;
 import org.hyperledger.besu.ethereum.mainnet.ProtocolSpec;
-import org.hyperledger.besu.ethereum.referencetests.DefaultReferenceTestWorldState;
+import org.hyperledger.besu.ethereum.referencetests.ForestReferenceTestWorldState;
```
garyschulte, line 51: Minor comment - IMO we should be removing Forest as the default everywhere we encounter it.

garyschulte (approved):
SHIP IT! 🚢

ethereum/referencetests/src/main/java/org/hyperledger/besu/ethereum/referencetests/BonsaiReferenceTestWorldState.java
garyschulte: > EVMTool actually uses in-memory Forest, perhaps it should be tweaked? -
> 
> https://github.com/hyperledger/besu/blob/ccab8571e5648adc00a7cec4d9a9d28155508bef/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BlockchainModule.java#L66-L68

I must be confused regarting evmtool vs T8nExecutor ...
https://github.com/hyperledger/besu/blob/main/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java#L241-L242

PR #6423, opened 2024-01-17
Promote block txs selection max time options to stable

fab-10:
## PR description

Promote block txs selection max time options to stable.
Both options were introduced in #6044, and are working fine since them, with the promotion the hard caps have been removed since that could be an issue with existing networks that are used to have longer block creation times.


pinges (approved):
LGTM

PR #6431, opened 2024-01-19
EOF Spec Updates

shemnon:
Update a few EOF features based on updated spec
* Add a prague reference test target
* Run evmtool from prestate when no code specified
* RETF and dangling immediate arg fixes


macfarla (approved):
LGTM

PR #6436, opened 2024-01-19
Implement "pending" for qbft_getValidatorsByBlockNumber

matthew1001:
## PR description
This PR adds support for "pending" as a string in qbft_getValidatorsByBlockNumber.

It returns the validators for the block that is in the process of being mined.

In certain stalled-chain scenarios where a new validator has been voted in but not enough validators are online to mine the next block, neither `qbft_getPendingVotes` nor `qbft_getValidatorsByBlockNumber("latest")` are useful since the vote is no longer pending, and the "latest" block is not the one that is failing to be mined. Similarly, `qbft_getValidatorsByBlockNumber(<next-block-number>)` cannot be used because the next block doesn't exist yet.

"pending" already exists as a block string, but isn't implemented by most JSON/RPC methods. This is only the 2nd method that implements it I believe, but it seems like a valid interpretation of "pending" in this case. All other JSON/RPC methods will continue to revert to "latest" if "pending" is specified.

## Fixed Issue(s)
No issue raised for this.

## Doc changes required

- I think https://besu.hyperledger.org/public-networks/how-to/use-besu-api/json-rpc#block-parameter needs updating to say "Use only with [eth_getTransactionCount](https://besu.hyperledger.org/public-networks/reference/api#eth_gettransactioncount) and [qbft_getValidatorsByBlockNumber](https://besu.hyperledger.org/private-networks/reference/api#qbft_getvalidatorsbyblocknumber)"


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/jsonrpc/methods/QbftGetValidatorsByBlockNumber.java
```diff
@@ -53,6 +53,21 @@ protected BlockParameter blockParameter(final JsonRpcRequestContext request) {
     return request.getRequiredParameter(0, BlockParameter.class);
   }
 
+  @Override
+  protected Object pendingResult(final JsonRpcRequestContext request) {
+    final long blockNumber = blockchainQueriesSupplier.get().headBlockNumber();
```
jframe, line 58: nit: could do this since the block number is only being used to get the head block header
```suggestion
    final BlockHeader blockHeader = getBlockchainQueries().headBlockHeader();
```

jframe (approved):
This use of pending makes sense to me. Just a minor code suggestion

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/jsonrpc/methods/QbftGetValidatorsByBlockNumber.java
matthew1001: There's also the log entry, but that can be replaced with `blockHeader.getNumber()` so seems fine to me.

PR #6437, opened 2024-01-19
Log changes in BFT validators

matthew1001:
## PR description
When the set of IBFT or QBFT validators changes between blocks, an `INFO` message is written listing the previous block number and its validators, as well as the pending block's number and validators.

This serves as a useful indicator that BFT voting has completed for a change in validators.

This log entry will not be written for public chain nodes so it should not introduce any unnecessary noise for public chain users. It isn't expected that validator changes in permissioned, BFT chains are likely to be so frequent that this log will be too noisy for those scenarios either.

I specifically didn't make it a `DEBUG` message since the validator list is key to how the BFT chain works, any changes to it are likely to be useful to an administrator/operator looking at just `INFO` logs, and as mentioned above these aren't expected to be very frequent.

Example output for a change from 1 validator node to 2 validator nodes:

```
2024-01-19 09:49:37.099+00:00 | main | INFO  | MessageValidatorFactory | Validator list change. Previous chain height 590245: [0x44740bDE71Ff5e8cd08c58a721A38Aaf62779478]. Current chain height 590246: [0x11427D85C111024762b0f6df9bfC575f4303F588, 0x44740bDE71Ff5e8cd08c58a721A38Aaf62779478].

```

## Fixed Issue(s)
No issue raised for this.


consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/validation/MessageValidatorFactory.java
```diff
@@ -83,6 +95,18 @@ public MessageValidator createMessageValidator(
     final BlockValidator blockValidator =
         protocolSchedule.getByBlockNumber(roundIdentifier.getSequenceNumber()).getBlockValidator();
     final Collection<Address> validators = getValidatorsAfterBlock(parentHeader);
+    final Collection<Address> previousValidators = getValidatorsForBlock(parentHeader);
+
+    // Log any change in validators
+    if (!(validators.containsAll(previousValidators))
```
jframe, line 101: Think this could end up logging multiple times for a height if had we multiple round changes. A better place is probably when we import the block in the IbftRound importBlockToChain method then we know the block is finalised

matthew1001: Yeah I was wondering if this could cause more than 1 log entry, so I'll look again at where the check could be made to ensure only 1 log entry per block. The specific case we were trying to diagnose was the case where the validators had changed, but immediately afterwards too many of the validators went offline to mine a new block - so putting the log entry on `imortBlockToChain` wouldn't help in that scenario. I'll have a look at where the log entry could go that is before the block is produced, but doesn't give duplicate log entries.

matthew1001: @jframe I've refactored the PR so that the validator change is checked and/or logged in `handleBlockTimerExpiry`, which seems a more sensible place to put a per-block check, and only writes a log entry for round `0`. That prevents multiple log entries if it takes several rounds to agree on a block.

I exposed the existing `MessageValiatorFactory.getValidators<For|After>Block()` methods as statics to avoid re-implementing them in `QbftBlockHeightManager/IbftBlockHeightManager`. Might be a cleaner way of re-using them - let me know what you think.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftBlockHeightManager.java
```diff
@@ -133,6 +137,30 @@ public void handleBlockTimerExpiry(final ConsensusRoundIdentifier roundIdentifie
...
+   *
+   * @param ibftRound The current round
+   */
+  private void logValidatorChanges(final IbftRound ibftRound) {
+    if (ibftRound.getRoundIdentifier().getRoundNumber() == 0) {
+      final Collection<Address> previousValidators =
+          MessageValidatorFactory.getValidatorsForBlock(ibftRound.protocolContext, parentHeader);
+      final Collection<Address> validatorsForHeight =
+          MessageValidatorFactory.getValidatorsAfterBlock(ibftRound.protocolContext, parentHeader);
+      if (!(validatorsForHeight.containsAll(previousValidators))
```
jframe, line 152: Are the two containsAll calls used so that the validators can be compared ignoring order?

jframe (approved):

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftBlockHeightManager.java
matthew1001: Not to ignore order, but to check for both new-validator and removed-validator cases.

PR #6439, opened 2024-01-19
Only accept an address from a peer if it is a valid IP address

matthew1001:
## PR description
When using the `From` address specified by a peer, ensure it is a valid IP address. Otherwise revert to using the UDP source host.

## Fixed Issue(s)
See comment https://github.com/hyperledger/besu/pull/6225#issuecomment-1900243754

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
@@ -293,7 +293,9 @@ public abstract class PeerDiscoveryAgent {
             .getPacketData(PingPacketData.class)
             .flatMap(PingPacketData::getFrom)
             .map(Endpoint::getHost)
-            .filter(abc -> !abc.equals("127.0.0.1"))
+            .filter(
+                fromAddr ->
+                    (!fromAddr.equals("127.0.0.1") && InetAddresses.isInetAddress(fromAddr)))
             .stream()
             .peek(
                 h ->


matkt (approved):
LGTM

matthew1001:
I'll submit another PR to honour DNS names DNS peering is enabled for the node, but it would be good to just get this fix in first.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -293,7 +293,9 @@ protected void handleIncomingPacket(final Endpoint sourceEndpoint, final Packet
             .getPacketData(PingPacketData.class)
             .flatMap(PingPacketData::getFrom)
             .map(Endpoint::getHost)
-            .filter(abc -> !abc.equals("127.0.0.1"))
+            .filter(
+                fromAddr ->
+                    (!fromAddr.equals("127.0.0.1") && InetAddresses.isInetAddress(fromAddr)))
```
garyschulte, line 298: it looks like we need to add `255.255.255.255` to this filter.  on mainnet we are regularly seeing logs like: 
```
{"@timestamp":"2024-01-24T06:18:16,129","level":"WARN","thread":"vert.x-eventloop-thread-3","class":"VertxPeerDiscoveryAgent","message":"Sending to peer DiscoveryPeer{status=bonding, enode=enode://5d01e72875f3485fdcce5255720696087ea7d704c6632282fbe39329db6417a6f5ef940cf05519677be7f9c31b6b553fe9192e9e9feed37db3fb97bcc7fd099e@255.255.255.255:30313, firstDiscovered=1706077069751, lastContacted=0, lastSeen=1706077069751} failed, native error code -13, packet: 0x3123aed37feabe25bd4cc2625a1d1e2b12a5a1bbe6f3ef594c1ea76abbfbae670ab30477655186a5a3bf4487f2fa183807dad26037e950e89cc07ba20eaf229437e69c2e56d44fe934082147098ca4a0bd99964d67d2046789d080f69d5be2c60101df05cb8489757c7882765f82765fcb84ffffffff8276698276698465b0abe402, stacktrace: io.netty.channel.unix.Errors$NativeIoException: sendToAddress(..) failed: Permission denied","throwable":""}
```

garyschulte: A more robust solution would be to filter by reachable addresses

PR #6441, opened 2024-01-22
Ensure that Bonsai tests with Key Value storage are using Bonsai data configuration

jframe:
## PR description
Ensure that Bonsai tests with Key Value storage are using Bonsai data configuration. As these key-value storage are explicit with Bonsai classes this doesn't cause an issue using the default forest config but for readability, it is better to use the Bonsai data storage configuration.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
@@ -31,6 +31,12 @@ public interface DataStorageConfiguration {
           .unstable(Unstable.DEFAULT)
           .build();
 
+  DataStorageConfiguration DEFAULT_BONSAI_CONFIG =
+      ImmutableDataStorageConfiguration.builder()
+          .dataStorageFormat(DataStorageFormat.BONSAI)
+          .bonsaiMaxLayersToLoad(DEFAULT_BONSAI_MAX_LAYERS_TO_LOAD)
+          .build();
+
   DataStorageFormat getDataStorageFormat();
 
   Long getBonsaiMaxLayersToLoad();
--- a/besu/src/test/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelperTest.java
+++ b/besu/src/test/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelperTest.java
@@ -76,7 +76,9 @@ class TrieLogHelperTest {
 
     inMemoryWorldState =
         new BonsaiWorldStateKeyValueStorage(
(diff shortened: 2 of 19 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/GenesisState.java
```diff
@@ -98,27 +101,28 @@ public static GenesisState fromJson(
...
-    return fromConfig(DataStorageFormat.FOREST, config, protocolSchedule);
+    return fromConfig(DataStorageConfiguration.DEFAULT_CONFIG, config, protocolSchedule);
   }
 
   /**
    * Construct a {@link GenesisState} from a JSON object.
    *
-   * @param dataStorageFormat A {@link DataStorageFormat} describing the storage format to use
+   * @param dataStorageConfiguration A {@link DataStorageFormat} describing the storage format to
+   *     use
```
matkt, line 111: "use" word seems to be strange here

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
```diff
@@ -31,6 +31,12 @@ public interface DataStorageConfiguration {
           .unstable(Unstable.DEFAULT)
```
matkt, line 31: why forest will have max layer to load ?

matkt (changes requested):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
jframe: Found out why it's needed when the build broke after removing it. It's needed because bonsaiMaxLayersToLoad is a required attribute for the Immutable value. We could make it optional and add some validation to ensure it's specified when using Bonsai mode. A better approach is probably to move all the Bonsai configuration into an Optional<BonsaiDataStorageConfiguration> as part of DataStorageConfiguration.

Not sure it's worth the effort to make a change. What do you think?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/GenesisState.java
jframe: Updated the javadoc

macfarla (approved):
I think this will fix some testing issues I'm running into when attempting to change the default from FOREST to BONSAI

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
matkt: I think it's fine as it was like that before

matkt (approved):

Issue #6324, opened 2023-12-21
Creation of minimalist-staker profile


non-fungible-nelson:
### Description
As a staker, I may want an option that maximizes my hardware value, but doesn't allow me to serve full sets of data to my peers. I cannot serve old blocks when requested, but may have a smaller disk for home staking. 

### Acceptance Criteria
The profile implements the following defaults
- Sync mode: CHECKPOINT
- Data-storage-format: BONSAI
    - bonsai-historical-block-limit: 128
- Max-peers: 25 
- -Xp2p-peer-lower-bound=10

- Chain-pruning: Enabled
    - Deposit logs / tree dependency (TEKU ONLY --> Consider later for BeKu )

non-fungible-nelson:
Consider naming change

Issue #6409, opened 2024-01-15
document the user profiles


macfarla:
create at least draft documentation for the new profiles, how to use them and how to override defaults etc.

Gabriel-Trintinalia:
# User Documentation

## Introduction to --profile CLI Option in Besu

Besu introduces a new feature, the `--profile` CLI option. This feature allows users to load existing TOML configuration files from resources, thereby enabling the override of default options. 

### How to Use the --profile CLI Option

To use the `--profile` CLI option, you need to specify it when running Besu, along with the name of the profile you want to use. For example:

```bash
besu --config-file=config.toml --profile=STAKER
```

In this case, `config.toml` is the user-provided configuration file, and `STAKER` is the pre-configured profile containing custom settings.

### Configuration Resolution

The configuration provider resolves settings in a cascading manner. It prioritizes settings in the following order:

1. Environment variables
2. Configuration file (specified by the `--config-file` option)
3. Profile file (specified by the `--profile` option)
4. Default values (used if no other configuration source is available)

For instance, if a setting is not found in the environment variables, Besu will look for it in the `config.toml` file. If the setting is not found in the `config.toml` file, Besu will then look for it in the `STAKER.toml` file. If the setting is not found in the `STAKER.toml` file, Besu will use the default value for that setting.

### Available Profiles

The `--profile` flag accepts the following profiles:

- `minimalist-staker`
- `staker`
- `privat` and `enterprise` (both alias to the same profile)

Any configuration options explicitly set in the config file or command line will overwrite any of the defaults set via the profile.

## Minimalist Staker Profile

For stakers who want to maximize their hardware value but don't want to serve full sets of data to their peers, Besu offers a `minimalist_staker` profile. This profile implements the following defaults:

- Sync mode: `CHECKPOINT`
- Data-storage-format: `BONSAI`
- bonsai-historical-block-limit: `128`
- Max-peers: `25`

To use the `minimalist_staker` profile, run Besu with the `--profile` option set to `minimalist_staker`:

```bash
besu --profile=minimalist_staker
```

Please note that any configuration options explicitly set in the config file or command line will overwrite the defaults set via the `minimalist_staker` profile.

## Enterprise/Private Profile

For private network operators who want to minimize unexpected confusion by setting sensible defaults, Besu offers an `enterprise` or `private` profile. This profile is designed to handle specific use-cases for private network operators.

[...]

Issue #6323, opened 2023-12-21
Creation of Configuration Profiles


non-fungible-nelson:
## Description
As a user, I want a `--profile` flag that I can set to easily overwrite some defaults. This way, as a new user I can only set one flag and have a good experience out of the box and get familiar with operating Besu more quickly.

### Acceptance Criteria
- New configuration option `--profile=string` 
- This flag, when set, overrides the defaults in the client with new ones from the parsed profiles.
- The profiles flag accepts the following profiles (expanded on in other tickets):
    - selfish-staker #6324
    - staker #6325
    - private network && enterprise (both should alias to the same profile) #6326
 - Any configuration options explicitly set in the config file or command line should overwrite any of the defaults set via the profile (i.e. private network defaults to `sync-mode=FAST`, if a user sets `sync-mode=FULL`, then the sync mode should be full

PR #6449, opened 2024-01-23
Add minimalist staker profile

Gabriel-Trintinalia:
## PR description

Add minimalist staker profile

Use
`besu --profile=minimalist_staker`

more: https://github.com/hyperledger/besu/issues/6409

## Fixed Issue(s)
fixes https://github.com/hyperledger/besu/issues/6324
see: #6323


Gabriel-Trintinalia:
@non-fungible-nelson maybe we should come up with a better name for this profile

config/src/main/resources/profiles/selfish-staker.toml
```diff
@@ -0,0 +1,6 @@
+sync-mode="X_CHECKPOINT"
+data-storage-format="BONSAI"
+bonsai-historical-block-limit=128
+max-peers=25
+Xp2p-peer-lower-bound=10
```
pinges, line 5: I would not use the lower bound. I checked yesterday, and it only seems to be used in one place. I would like to get rid of that config option ...

macfarla (approved):

PR #6451, opened 2024-01-23
manage empty range for storage

matkt:

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
@@ -130,7 +130,24 @@ public class RequestDataStep {
             (response, error) -> {
               if (response != null) {
                 downloadState.removeOutstandingTask(getStorageRangeTask);
-                for (int i = 0; i < response.slots().size(); i++) {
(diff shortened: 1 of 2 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
```diff
@@ -129,7 +129,24 @@ public CompletableFuture<List<Task<SnapDataRequest>>> requestStorage(
...
+                 */
+                final boolean isEmptyRange =
+                    (response.slots().isEmpty() || response.slots().get(0).isEmpty())
+                        && !response.proofs().isEmpty();
+                if (isEmptyRange) { // empty range detected
+                  slots.add(new TreeMap<>());
+                } else {
+                  slots.addAll(response.slots());
+                }
+                for (int i = 0; i < slots.size(); i++) {
```
garyschulte, line 149: if the range is empty, won't slots.size() be 0 and we will never enter this loop, and never add a response to the request?

matkt: It cannot be empty thanks to that https://github.com/hyperledger/besu/blob/8587a01be6f9d1b6a2aa7fbdca2bd2ba0a45fab9/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java#L145

PR #6452, opened 2024-01-23
Optimized Keccak Hashing for Account Storage Slots

matkt:
## PR description

The purpose of this PR is to avoid unnecessary recomputations of the Keccak hash for account storage slots during the commit phase. In most cases, we have already calculated the hashes during the transaction processing phase and can reuse these values. The idea is to add this information to the accumulator, storing each key-to-hash mapping as it's generated during the transaction. At the end of the block, during the commit phase, we will first check the cache before attempting to recalculate the hash.

The cache size should not pose a problem as it will inevitably contain fewer elements than the other maps in the accumulator. This PR also has an added benefit in cases where we need to access the same slot key for two different accounts during the transaction processing phase. In such cases, the hash will only be calculated once.


datatypes/src/main/java/org/hyperledger/besu/datatypes/StorageSlotKey.java
```diff
@@ -31,7 +32,7 @@
  */
 public class StorageSlotKey implements Comparable<StorageSlotKey> {
 
-  private final Hash slotHash;
+  private final Supplier<Hash> slotHash;
```
garyschulte, line 35: Suggest MemoizingSupplier here since the slotKey is final

matkt: yeah forgot thanks

ahamlat:
Regarding performance, there is a clear improvement on block processing with this PR as we can see in the screenshot below

image

Examining the CPU profiling also reveals the improvement in the Commit phase of block processing.

**Before this PR**

image

**After this PR**

image

ahamlat (approved):

garyschulte:
changed to Mergeable after @ahamlat confirmed testing regressions were unrelated to this PR

PR #6463, opened 2024-01-25
Skip pruning if trie log count is less than retention limit

siladu:
Rework error messaging in this case.

This can occur if pruning is attempted before the node has reached the configured retention limit (512 by default)

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
@@ -76,15 +76,11 @@ public class TrieLogHelper {
 
     final long lastBlockNumberToRetainTrieLogsFor = chainHeight - layersToRetain + 1;
 
-    final long countBeforePrune =
-        rootWorldStateStorage
-            .streamTrieLogKeys(DEFAULT_BONSAI_TRIE_LOG_PRUNING_WINDOW_SIZE)
-            .count();
     if (!validatePruneRequirements(
         blockchain,
         chainHeight,
         lastBlockNumberToRetainTrieLogsFor,
-        countBeforePrune,
+        rootWorldStateStorage,
         layersToRetain)) {
       return;
     }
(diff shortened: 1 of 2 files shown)


besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
```diff
@@ -88,11 +95,23 @@ void prune(
...
+      } else {
+        throw new IllegalStateException(
+            "There was an error deleting the trie log backup files. Please ensure besu is working before deleting them manually.");
+      }
     } else {
-      LOG.error("Prune failed. Re-run the subcommand to load the trie logs from file.");
+      throw new IllegalStateException(
+          String.format(
+              "Remaining trie logs (%d) did not match %s (%d). Trie logs backup files have not been deleted, it is safe to rerun the subcommand.",
+              countAfterPrune, BONSAI_STORAGE_FORMAT_MAX_LAYERS_TO_LOAD, layersToRetain));
```
gfukushima, line 114: Should we remove BONSAI_STORAGE_FORMAT_MAX_LAYERS_TO_LOAD from this exception message?

besu/src/test/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelperTest.java
```diff
@@ -58,7 +61,7 @@ class TrieLogHelperTest {
 
   private static final StorageProvider storageProvider = new InMemoryKeyValueStorageProvider();
   private static BonsaiWorldStateKeyValueStorage inMemoryWorldState;
-  private TrieLogHelper trieLogHelper;
+  private TrieLogHelper nonValidatingTrieLogHelper;
```
gfukushima, line 64: what you do you mean with nonValidatingTrieLogHelper?

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
siladu: why?

Should look like this 
> Remaining trie logs (0) did not match --bonsai-historical-block-limit (3). Trie logs backup files have not been deleted, it is safe to rerun the subcommand

besu/src/test/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelperTest.java
siladu: The validation enforces 512 minimum so I'm skipping it for the sake of the unit tests not having to create lots of data.

besu/src/main/java/org/hyperledger/besu/cli/subcommands/storage/TrieLogHelper.java
gfukushima: Nevermind misread that and thought we were using the default value of that option.

gfukushima (approved):

PR #6472, opened 2024-01-26
Improve flat trace generation performance

ahamlat:
## PR description
When testing one of the blocks that has a lot of subcalls (838 internal transactions), we noticed that most of the time was spent on hasRevertInSubCall method. 

image

This PR reduced the time to trace a 29 mgas block from more than 5 seconds to 1 second.


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java
```diff
@@ -530,11 +530,19 @@ private static FlatTrace.Context handleCallDataLoad(
...
-    return transactionTrace.getTraceFrames().stream()
-        .filter(traceFrame -> !traceFrame.equals(callFrame))
-        .takeWhile(traceFrame -> !traceFrame.getOpcode().equals("RETURN"))
-        .filter(traceFrame -> traceFrame.getOpcode().equals("REVERT"))
-        .anyMatch(traceFrame -> traceFrame.getDepth() == callFrame.getDepth());
+    for (int i = 0; i < transactionTrace.getTraceFrames().size(); i++) {
+      if (i + 1 < transactionTrace.getTraceFrames().size()) {
+        final TraceFrame next = transactionTrace.getTraceFrames().get(i + 1);
+        if (next.getDepth() == callFrame.getDepth()) {
+          if (next.getOpcode().equals("REVERT")) {
```
shemnon, line 537: We should add opcodeNumber to the TraceFrame and check on that instead of checking for string equality.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java
```diff
@@ -530,11 +530,19 @@ private static FlatTrace.Context handleCallDataLoad(
...
-    return transactionTrace.getTraceFrames().stream()
-        .filter(traceFrame -> !traceFrame.equals(callFrame))
-        .takeWhile(traceFrame -> !traceFrame.getOpcode().equals("RETURN"))
-        .filter(traceFrame -> traceFrame.getOpcode().equals("REVERT"))
-        .anyMatch(traceFrame -> traceFrame.getDepth() == callFrame.getDepth());
+    for (int i = 0; i < transactionTrace.getTraceFrames().size(); i++) {
+      if (i + 1 < transactionTrace.getTraceFrames().size()) {
+        final TraceFrame next = transactionTrace.getTraceFrames().get(i + 1);
+        if (next.getDepth() == callFrame.getDepth()) {
+          if (next.getOpcodeNumber() == 0xFD) { // REVERT opCode
```
shemnon, line 537: Consider adding an OPCODE constant to RevertOperation and ReturnOperation so we can reduce magic constants in the code.

shemnon (dismissed):
OPCODE request is optional.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java
ahamlat: Thanks for the suggestion, I was thinking actually of an Enum OpCode to store all the opcodes and use only the Enum in the EVM and outside, like Opcode.REVERT.getOpcodeNumber() for 0xFD
public enum Opcode {
    STOP(0x00, "STOP"),
    ADD(0x01, "ADD"),
    MUL(0x02, "MUL"),
    ...
}
In terms of performance, I guess JIT will replace by the opCodeNumber or the opcode name at runtime

ahamlat: @shemnon I've added the OpCode enum within the EVM and would like to use it across all EVM opcode operations. An implementation example has been provided for ADDMOD. Does this approach work for you ?

shemnon: This is not what I meant. Such an OpCode enum breaks exensibility downstream users use.  Please revert.

shemnon: What I intended was something like this, where a constant is added to the Operation classes, not an entirely different enum.

https://github.com/hyperledger/besu/blob/c89e4db8c345f0330774ca27ccfccff35bdbe20c/evm/src/main/java/org/hyperledger/besu/evm/operation/BlobHashOperation.java#L37

ahamlat: I actually understood your suggestion but wanted to suggest an implementation that would work for similar use cases in the future, but it looks it wasn't a good idea. I'm going to revert the last changes.

ahamlat:
@shemnon Could you review the last changes ?

shemnon (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java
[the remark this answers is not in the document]
ahamlat: This commit is outdated. Making AbstractOperation.getOpcode() static means that we need to make opcode static in AbstractOperation which not the case, opcode depends on the implementation.
What suggested by Danno is very similar in some way to what you suggest, add a static constants directly in the implementation, that is stored in the opcode field.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/flat/FlatTraceGenerator.java
```diff
@@ -530,11 +532,19 @@ private static FlatTrace.Context handleCallDataLoad(
 
   private static boolean hasRevertInSubCall(
       final TransactionTrace transactionTrace, final TraceFrame callFrame) {
-    return transactionTrace.getTraceFrames().stream()
-        .filter(traceFrame -> !traceFrame.equals(callFrame))
-        .takeWhile(traceFrame -> !traceFrame.getOpcode().equals("RETURN"))
-        .filter(traceFrame -> traceFrame.getOpcode().equals("REVERT"))
-        .anyMatch(traceFrame -> traceFrame.getDepth() == callFrame.getDepth());
+    for (int i = 0; i < transactionTrace.getTraceFrames().size(); i++) {
+      if (i + 1 < transactionTrace.getTraceFrames().size()) {
```
kinsi55, line 536: Kind of a random question (Side note: I dont really have Java exp) but what is the point of this If condition? Wouldn't starting the loop at 1 and running it until `size() - 1` have the same result? (Or if iterating in reverse has no side-effects, even starting at size() - 1 and running while its > 0)

PR #6484, opened 2024-01-29
Add segmented transaction get method

jframe:
## PR description
Add segmented transaction get method. This is needed to fully implement the reference counting for code storage changes to move to hashing by code hash.

### DIFF
--- a/services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
+++ b/services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
@@ -272,9 +272,17 @@ public class SegmentedInMemoryKeyValueStorage
 
     @Override
     public Optional<byte[]> get(final SegmentIdentifier segmentIdentifier, final byte[] key) {
-      return updatedValues
-          .computeIfAbsent(segmentIdentifier, __ -> new HashMap<>())
-          .getOrDefault(Bytes.wrap(key), Optional.empty());
+      final Optional<byte[]> updatedValue =
+          updatedValues
+              .computeIfAbsent(segmentIdentifier, __ -> new HashMap<>())
+              .getOrDefault(Bytes.wrap(key), Optional.empty());
+      final Set<Bytes> removedKeys =
+          this.removedKeys.computeIfAbsent(segmentIdentifier, __ -> new HashSet<>());
+      return removedKeys.contains(Bytes.wrap(key))
+          ? Optional.empty()
+          : updatedValue.isPresent()
+              ? updatedValue
+              : SegmentedInMemoryKeyValueStorage.this.get(segmentIdentifier, key);
     }
 
     @Override
--- a/services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
+++ b/services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
@@ -271,6 +271,13 @@ public class SegmentedInMemoryKeyValueStorage
     }
 
     @Override
+    public Optional<byte[]> get(final SegmentIdentifier segmentIdentifier, final byte[] key) {
+      return updatedValues
+          .computeIfAbsent(segmentIdentifier, __ -> new HashMap<>())
+          .getOrDefault(Bytes.wrap(key), Optional.empty());
+    }
+
+    @Override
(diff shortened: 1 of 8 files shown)


services/kvstore/src/test/java/org/hyperledger/besu/services/kvstore/AbstractSegmentedKeyValueStorageTest.java
```diff
@@ -79,4 +79,61 @@ public void assertSegmentedIsNearestTo() throws Exception {
...
+    final byte[] value2 = Bytes.fromHexString("0003").toArrayUnsafe();
+
+    tx.put(SEGMENT_IDENTIFIER, key, value1);
+    tx.remove(SEGMENT_IDENTIFIER, key);
+    tx.put(SEGMENT_IDENTIFIER, key, value2);
+    assertThat(tx.get(SEGMENT_IDENTIFIER, key).map(Bytes::wrap)).contains(Bytes.wrap(value2));
+  }
+
+  @Test
+  void txGet_returnsUpdatedValueAfterPutPut() {
```
siladu, line 128: Love the name 😁

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedKeyValueStorageTransactionValidatorDecorator.java
```diff
@@ -43,6 +44,14 @@ public SegmentedKeyValueStorageTransactionValidatorDecorator(
     this.transaction = toDecorate;
   }
 
+  @Override
+  public Optional<byte[]> get(final SegmentIdentifier segmentId, final byte[] key)
+      throws StorageException {
+    checkState(active, "Cannot invoke get() on a completed transaction.");
```
siladu, line 50: Why can't we invoke get on a completed transaction?

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
```diff
@@ -270,6 +270,13 @@ public void put(
       removedKeys.computeIfAbsent(segmentIdentifier, __ -> new HashSet<>()).remove(Bytes.wrap(key));
     }
 
+    @Override
+    public Optional<byte[]> get(final SegmentIdentifier segmentIdentifier, final byte[] key) {
+      return updatedValues
+          .computeIfAbsent(segmentIdentifier, __ -> new HashMap<>())
+          .getOrDefault(Bytes.wrap(key), Optional.empty());
```
siladu, line 277: Just checking my understanding...if we are computing empty HashMap if absent, then can we ever actually return Optional.empty() here?

siladu: Do we use this InMemory class for writing data to the database (i.e. persisting this state) or only for reading data and/or unit tests?

siladu (commented):
LGTM. Like the tests. I will leave for a storage expert to approve

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedKeyValueStorageTransactionValidatorDecorator.java
jframe: For a get maybe it is fine? I was just following the pattern that the other methods did which was not operations on a completed transaction. Hopefully, a storage expert will know the answer.

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
jframe: > Just checking my understanding...if we are computing empty HashMap if absent, then can we ever actually return Optional.empty() here?
We can because the empty hashmap won't have the key. This just ensures that we have a map for the requested segment identifier. The put and remove methods also use this pattern.

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedKeyValueStorageTransactionValidatorDecorator.java
gfukushima: I think it makes sense to throw in this case cause we want to avoid returning values we are uncertain if it's going to be committed or not. Think chances are minimal but there is a scenario where this could happen. @garyschulte what you reckon?

jframe:
No longer needed with current approach of not deleting self-destruct from code column

PR #6491, opened 2024-01-30
Remove `--engine-jwt-enabled` deprecated option

Gabriel-Trintinalia:
## PR description
Removes `--engine-jwt-enabled`. Replaced by `--engine-jwt-disabled`

It has been deprecated since 2022.


pinges (approved):
LGTM

Issue #6157, opened 2023-11-10
Bonsai flat db healing - Subcommand & Full sync 


garyschulte:
### Description
Enable flat db healing by default for bonsai when using snap-sync.  Ensure full sync databases are correctly marked as full flat.

### Acceptance Criteria
- [ ] ensure initial flat db sync strategy (full/partial) is persisted to the database
- [ ] do not allow switching between full/partial via CLI flag on restart
- [ ] ensure that full sync'd nodes are marked as full flat or partial 
- [ ] add subcommand to 'heal' a partial db into a full flat db

non-fungible-nelson:
@garyschulte can you elaborate on the first acceptance criterium ?

PR #6499, opened 2024-01-31
Flat db heal - deprecate CLI option and set to always true

macfarla:
* set flat db healing enabled to true always
* remove duplicate setup of default values from Sync Options class

refs #6157


pinges (approved):
LGTM and nice cleanup as well!

macfarla:
will wait for feedback from @garyschulte

garyschulte:
This LGTM, but there are a couple criteria of #6157 it doesn't meet:
* full sync'd nodes are not getting defaulted to fully flattened DB, see FlatDbStrategyProvider.deriveFlatDbStrategy
* we need to offer a subcommand to "heal" a node's flat db storage so that the user doesn't have to resync the whole worldstate

I think it is reasonable to handle those two in separate PRs.  They are more tricky cases

garyschulte (approved):
LGTM, comments noted

macfarla:
agree @garyschulte those changes are related but separate, let's do separate PRs

PR #6500, opened 2024-01-31
[MINOR] remove duplicate setting of default values for CLI options

macfarla:
where the default is number/boolean/string, just leave it to the initialized value for the actual field.

same as was done for Sync Options in #6499


Gabriel-Trintinalia (approved):
LGTM

PR #6509, opened 2024-01-31
give tests permission to write checks/status and put reports on pr

jflo:
Should not be merged. Demonstrating acceptance test failures publishing to the PR, if repo permissions allow it.


garyschulte (approved):
approval test

PR #6514, opened 2024-02-01
Create CODEOWNERS

ryjones:
## PR description

Creating a valid CODEOWNERS file

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# Besu Maintainers
+*       @hyperledger/besu-maintainers


shemnon (changes requested):
Why do we need to do this?  The last time we did this it flooded all our github inboxes with literally every PR.

ryjones:
@shemnon 
> Why do we need to do this? The last time we did this it flooded all our github inboxes with literally every PR.

I went through a bunch of branch rules; none of them had "require code owner reviews". The reason to do this is for some PR automation. @jflo ?

shemnon:
Has github changed automation that will then add all the codeowners as a designated reviewer to every single PR? That's what happend last time we tried to turn it on.

If it's not absolutely needed for github actions I don't want to have to dismiss every besu PR that gets posted from my github inbox.

jflo:
We don't need CODEOWNERS for anything related to a GHA based CI/CD pipeline.  Happy to research adopting it in the future, but it is not needed.

PR #6516, opened 2024-02-02
Close Sockets after testing for availablity

pinges:
## PR description
Close the Sockets after checking for their availability.
Also makes the code more readable by checking for unavailability not availability.


macfarla (approved):

Gabriel-Trintinalia (approved):
lgtm

PR #6518, opened 2024-02-02
Change contract creation halt reason when account already exists

gfukushima:
## PR description
Change the execution halt reason from INSUFFICIENT_GAS to ILLEGAL_STATE_CHANGE to reflect the real reason behind this case which is a possible contract creation collision with a already existing a account at that address.

Note that some of the trace ATs had to be adjusted after this change, possibly because the frame where we fail execution was getting included into the trace result as we don't ignore frames that halted due to INSUFFICIENT_GAS https://github.com/hyperledger/besu/blob/2a30dfba1a03ad1c2b9d42857615984c439f4970/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/vm/VmTraceGenerator.java#L110


ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/trace/specs/trace-call/trace_call_14_1_vmTrace.json
```diff
@@ -94,19 +94,6 @@
           },
           "pc" : 31,
           "sub" : null
-        }, {
```
shemnon, line 97: This should not be happening.  We need to fing the code that is dropping this trace line when the error kind changes and fix it.

shemnon (changes requested):
We need to figure out why the VM traces need to be altered to get it to pass tests and fix it.  The error text changing int he call trace is fine, but losing the CALL instruction in the VM trace breaks a lot of downstream users.

shemnon:
Probably in `org.hyperledger.besu.ethereum.api.jsonrpc.internal.results.tracing.vm.VmTraceGenerator`

ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/trace/specs/trace-call/trace_call_14_1_vmTrace.json
gfukushima: Hi @shemnon, thanks for the feedback. As I mentioned in the PR description, I knew the line was not included in the trace due to the fact that any frame that halts with a reason different from INVALID_JUMP_DESTINATION or INSUFFICIENT_GAS is ignored. I've made the necessary adjustments now.

shemnon (approved):
LGTM

Issue #5243, opened 2023-03-20
Configurable Allow/Reject List in Besu Transaction Pool


non-fungible-nelson:
## Description ## 

As a user, I would like to have an allowlist/reject list for certain types of transactions. Besu should have an aware reject-list where when transactions from certain (configurable) addresses enter my mem pool, I can eject them without much fuss. Or specify a certain allowlist where _only_ transactions from this list are accepted. This can be useful for certain private networks and public networks use pertaining to regulations or ethical concerns. 

## Acceptance Criteria ## 

Reject List
1. User defines a list of addresses that they want to reject from executing. 
2. Besu's mem-pool receives a transaction from an address on this list. 
3. Besu's mem-pool ejects the transaction and logs in the console what it has done. 
4. User does not execute reject list address tx.

Allow List
1. User defines a list of addresses from where they are interested in executing transactions. 
2. Besu's mem-pool receives a transaction from an address on this list. 
3. Besu's mem-pool executes the transaction. 
4. User executes accept list address tx.

2a. Besu's mem-pool receives a transaction from an address _not_ on the list.
3a. Besu's mem-pool ejects the transaction outright. 
4a. User does not execute transactions not on the allow-list.

non-fungible-nelson:
@fab-10 does any of the recent tx pool work make this simpler to implement?

fab-10:
not really, and I think it is better to delegate this feature to a plugin, the plugin interface could allow for generic added validation/checks of the tx on top of the standard validation /checks done by default-

macfarla:
@Brindrajsinh-Chauhan are you still working on implementing this as a plugin? We're also actively collecting feedback with the aim to improve docs and devx of the plugin API!

jflo:
Any interim work/notes/designs might be useful for EIP-7547 so users could write their own censorship detection heuristics.

macfarla:
@Brindrajsinh-Chauhan are you still working on implementing this as a plugin? We're also actively collecting feedback with the aim to improve docs and devx of the plugin API!

krsh24:
@matthew1001 raised a valid point in the linked issue ticket #6521, account permissioning which can be controlled via smart contract provides similar functionality that this spec requires. Do we need a plugin to achieve a similar functionality? Unless I got this wrong. I currently use GoQuorum and I use a smart contract-based whitelist to control the accounts whitelist. Both Besu and GoQuorum have this similar feature.

```
@CommandLine.Option(
[...]

PR #6521, opened 2024-02-04
tx-pool allow/reject list

Brindrajsinh-Chauhan:
## PR description
This PR adds the following flags to configure allow/reject list of sender addresses 

- `--tx-pool-allow-list=<comma-seprated-list-of-addresses>` - Local/Remote transactions submitted from these addresses are only allowed to be added in the pool, while transactions from other sender addresses are reject with reason `Sender account not authorized to send transactions`
- `--tx-pool-reject-list=<comma-seprated-list-of-addresses>` - Local/Remote transactions submitted from these addresses are rejected before being added to the pool with reason `Sender account not authorized to send transactions`, while transactions from other sender addresses are allowed.
- Both these flags are not allowed together, only one can be set at a time. 

**Sample Besu Log**
Setting the following config for test addresses and sending test transaction from these accounts 
`tx-pool-reject-list=["0x8C7d8409bDBA3322A65825bbe3f75ac71b9de3cf,0xe547BAa69dB4fCFd6225f428De424888dB566a41"]` for test 
```
2024-02-04 21:47:20.559+00:00 | vert.x-worker-thread-1 | TRACE | AbstractJsonRpcExecutor | {"jsonrpc":"2.0","id":83,"method":"eth_sendRawTransaction","params":["0xf8868080830f4240944fca308301a110a0118448c644654efeecf9a91180a460fe47b10000000000000000000000000000000000000000000000000000000000003039820ff1a0777743d184749836701b2bdc63c35f89523ae804291696cf6d5778e24aa1240aa07ff93a28da5138dd50277ec7b3249b03736eaa8cc99c035503131eccb6e5b04c"]}
2024-02-04 21:47:20.559+00:00 | vert.x-worker-thread-1 | DEBUG | JsonRpcExecutor | JSON-RPC request -> eth_sendRawTransaction ["0xf8868080830f4240944fca308301a110a0118448c644654efeecf9a91180a460fe47b10000000000000000000000000000000000000000000000000000000000003039820ff1a0777743d184749836701b2bdc63c35f89523ae804291696cf6d5778e24aa1240aa07ff93a28da5138dd50277ec7b3249b03736eaa8cc99c035503131eccb6e5b04c"]
2024-02-04 21:47:20.560+00:00 | vert.x-worker-thread-1 | TRACE | EthSendRawTransaction | Received local transaction MessageCall{type=FRONTIER, nonce=0, gasPrice=0 wei, gasLimit=1000000, to=0x4fca308301a110a0118448c644654efeecf9a911, value=0x0000000000000000000000000000000000000000000000000000000000000000, sig=Signature{r=54035951847700603082610381881786735130294417072039064688183051346590583301130, s=57884078091882121988970232755018579959487602981021234593703990990705896304716, recId=0}, chainId=2023, payload=0x60fe47b10000000000000000000000000000000000000000000000000000000000003039}
2024-02-04 21:47:20.560+00:00 | vert.x-worker-thread-1 | TRACE | TransactionPool | Discarded transaction from unauthorized sender 0x8c7d8409bdba3322a65825bbe3f75ac71b9de3cf
2024-02-04 21:47:20.560+00:00 | vert.x-worker-thread-1 | TRACE | AbstractJsonRpcExecutor | {"jsonrpc":"2.0","id":83,"error":{"code":-32007,"message":"Sender account not authorized to send transactions"}}
```

## Fixed Issue(s)

Fixes Issue https://github.com/hyperledger/besu/issues/5243

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
@@ -56,6 +56,8 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
       "--strict-tx-replay-protection-enabled";
   private static final String TX_POOL_PRIORITY_SENDERS = "--tx-pool-priority-senders";
   private static final String TX_POOL_MIN_GAS_PRICE = "--tx-pool-min-gas-price";
+  private static final String TX_POOL_ALLOW_LIST = "--tx-pool-allow-list";
+  private static final String TX_POOL_REJECT_LIST = "--tx-pool-reject-list";
 
   @CommandLine.Option(
       names = {TX_POOL_IMPLEMENTATION},
@@ -114,6 +116,24 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
       fallbackValue = "true",
       arity = "0..1")
   private Boolean strictTxReplayProtectionEnabled = false;
+
+  @CommandLine.Option(
+      names = {TX_POOL_ALLOW_LIST},
+      split = ",",
+      paramLabel = "Comma separated list of addresses to allow transactions",
+      description =
+          "List of sender addresses to accept transactions. If not specified, transactions from all sender addresses are allowed (default: ${DEFAULT-VALUE})",
+      arity = "1..*")
+  private Set<Address> sendersAllowList = TransactionPoolConfiguration.DEFAULT_SENDER_ALLOW_LIST;
+
+  @CommandLine.Option(
+      names = {TX_POOL_REJECT_LIST},
+      split = ",",
+      paramLabel = "Comma separated list of addresses to reject transactions",
+      description =
+          "List of sender addresses to reject transactions. If not specified, transactions from no sender addresses are is rejected (default: ${DEFAULT-VALUE})",
+      arity = "1..*")
+  private Set<Address> senderRejectList = TransactionPoolConfiguration.DEFAULT_SENDER_REJECT_LIST;
 
   @CommandLine.Option(
       names = {TX_POOL_PRIORITY_SENDERS},
@@ -268,6 +288,8 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
     options.saveFile = config.getSaveFile();
     options.strictTxReplayProtectionEnabled = config.getStrictTransactionReplayProtectionEnabled();
     options.prioritySenders = config.getPrioritySenders();
+    options.sendersAllowList = config.getSendersAllowList();
+    options.senderRejectList = config.getSendersRejectList();
     options.minGasPrice = config.getMinGasPrice();
     options.layeredOptions.txPoolLayerMaxCapacity =
         config.getPendingTransactionsLayerMaxCapacityBytes();
@@ -311,6 +333,12 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
         "Price bump option is not compatible with zero base fee market",
         !genesisConfigOptions.isZeroBaseFee(),
         List.of(TX_POOL_PRICE_BUMP));
+
+    CommandLineUtils.failIfOptionDoesntMeetRequirement(
+        commandLine, 
+        "Can only set either Reject list or allow list of senders",
+        !CommandLineUtils.isOptionSet(commandLine, TX_POOL_ALLOW_LIST),
+        List.of(TX_POOL_REJECT_LIST));
   }
 
   @Override
@@ -324,6 +352,8 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
         .saveFile(saveFile)
         .strictTransactionReplayProtectionEnabled(strictTxReplayProtectionEnabled)
         .prioritySenders(prioritySenders)
+        .sendersAllowList(sendersAllowList)
+        .sendersRejectList(senderRejectList)
         .minGasPrice(minGasPrice)
         .pendingTransactionsLayerMaxCapacityBytes(layeredOptions.txPoolLayerMaxCapacity)
         .maxPrioritizedTransactions(layeredOptions.txPoolMaxPrioritized)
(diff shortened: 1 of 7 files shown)


matthew1001:
Generally looks good @Brindrajsinh-Chauhan, I've pulled the branch and tested it out locally and it works exactly as you'd expect. I have a few initial comments which other maintainers (particularly @fab-10) might have thoughts on as well:

Higher level discussion points:

- There is a big overlap between the spec in the [original issue](https://github.com/hyperledger/besu/issues/5243) and [account permissioning](https://besu.hyperledger.org/23.10.2/private-networks/concepts/permissioning#account-permissioning) which currently supports reject lists but not allow lists.

Comments on the PR itself:

- It feels like it might be useful to support the allowed/rejected address list going in a separate file e.g. `--tx-pool-reject-file=reject.toml`. This could be in addition to your inline comma-separated list `--tx-pool-reject-list` but as address lists get bigger (particularly in the case of `--tx-pool-allow-list` a standalone file might be a useful option
  - This is partly what made me think of account permissioning, because once you start adding files to store allowed/rejected addresses the feature starts to look a lot like local account permissioning.
- Technically a transaction from a rejected address can still be executed on the node, even though it can't be accepted into the pool, because `eth_estimateGas` doesn't incorporate the allow/reject checks. It's not necessarily a high priority to prevent any execution of TXs from rejected addresses, because aside from the CPU time it's unlikely to do any harm, but it's probably worth some discussion. Similarly I guess `eth_call` would benefit from the check if `eth_estimateGas` does.

@non-fungible-nelson as you raised the original issue (I think based on some discussion on discord?) you might also have comments on some of the above.

Brindrajsinh-Chauhan:
I was also thinking initially to keep it as a file instead of a ever growing list as that would be easy to maintain. Will look into that and see how I can incorporate that. And also give a thought on the eth_estimateGas point you raised.

non-fungible-nelson:
Hi there - the language we use elsewhere in the code is "deny list," so might favor using that over "reject."

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -115,6 +117,24 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
       arity = "0..1")
   private Boolean strictTxReplayProtectionEnabled = false;
 
+  @CommandLine.Option(
+      names = {TX_POOL_ALLOW_LIST},
+      split = ",",
+      paramLabel = "Comma separated list of addresses to allow transactions",
+      description =
+          "List of sender addresses to accept transactions. If not specified, transactions from all sender addresses are allowed (default: ${DEFAULT-VALUE})",
```
macfarla, line 125:
```suggestion
          "List of sender addresses from which to accept transactions. If not specified, transactions from all sender addresses are allowed (default: ${DEFAULT-VALUE})",
```

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -115,6 +117,24 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
...
+      arity = "1..*")
+  private Set<Address> sendersAllowList = TransactionPoolConfiguration.DEFAULT_SENDER_ALLOW_LIST;
+
+  @CommandLine.Option(
+      names = {TX_POOL_REJECT_LIST},
+      split = ",",
+      paramLabel = "Comma separated list of addresses to reject transactions",
+      description =
+          "List of sender addresses to reject transactions. If not specified, transactions from no sender addresses are is rejected (default: ${DEFAULT-VALUE})",
```
macfarla, line 134:
```suggestion
          "List of sender addresses from which to deny transactions. If not specified, no transaction senders are explicitly denied (default: ${DEFAULT-VALUE})",
```

besu/src/test/resources/everything_config.toml
```diff
@@ -192,6 +192,8 @@ tx-pool-retention-hours=999
 tx-pool-max-size=1234
 tx-pool-limit-by-account-percentage=0.017
 tx-pool-min-gas-price=1000
+tx-pool-allow-list=["0xABC0000000000000000000000000000000001234"]
+tx-pool-reject-list=["0xDEF0000000000000000000000000000000001234"]
```
macfarla, line 196: won't this cause an error if both are specified in the toml file?

macfarla (commented):
added some suggestions on wording for CLI options.

My other thought is - it's a bit similar to the TransactionSelection plugin here https://github.com/Consensys/besu-sequencer-plugins/blob/main/PLUGINS.md#transaction-validation---lineatransactionvalidatorplugin - where we implemented the deny list but not (yet) the allow list. Do we want this functionality in the tx pool directly, or should we expand the plugin? cc @fab-10

matthew1001:
@macfarla where does this fit with account permissioning? I'm wondering how the docs will explain when to use account permissioning vs tx pool allow lists.

besu/src/test/resources/everything_config.toml
Brindrajsinh-Chauhan: I did not seem to cause an issue, but can double check

Brindrajsinh-Chauhan:
> https://github.com/Consensys/besu-sequencer-plugins/blob/main/PLUGINS.md#transaction-validation---lineatransactionvalidatorplugin

Thank you for sharing this, I now see the point of adding allow list in the plugin here and can modify the PR for allow list plugin if that seems to be the right path instead of standalone cli flags. would like to hear more thoughts. cc @fab-10

fab-10:
Hi @Brindrajsinh-Chauhan, my suggestion is to go with the plugin path, taking as a starting point [the basic implementation](https://github.com/Consensys/besu-sequencer-plugins/blob/main/PLUGINS.md#transaction-validation---lineatransactionvalidatorplugin) that @macfarla shared, and expanding it to support more features, for example I immagine that it will be useful to have dynamic lists that could be updated at runtime without restating, or having different sources for the lists, like  files or dbs, and possibly additional RPC APIs to admin the stuff, not saying that all of these features will be needed, but having a plugin that could be extended to support different use cases is more flexible.
If you decide to follow the plugin path, then I suggest that you create a brand new repo for it, so it can be independently managed.

Brindrajsinh-Chauhan:
Thank you for the comment. Will work on that.

macfarla:
> Thank you for the comment. Will work on that.

Hi @Brindrajsinh-Chauhan I'm going to close this PR since you're planning to work on a plugin for it. do shout out in Discord if you have questions! https://discord.gg/hyperledger

PR #6524, opened 2024-02-05
Add ETC Spiral EVM configuration

diega:
## PR description
During the last ETC network upgrade, Besu nodes got stalled because, inadvertently, [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) was _included_ instead of _omitted_.

For the future, we'll also test omissions to the protocol and not only additions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 ### Bug fixes
 - Fix the way an advertised host configured with `--p2p-host` is treated when communicating with the originator of a PING packet [#6225](https://github.com/hyperledger/besu/pull/6225)
 - Fix `poa-block-txs-selection-max-time` option that was inadvertently reset to its default after being configured [#6444](https://github.com/hyperledger/besu/pull/6444)
+- Fix ETC Spiral upgrade breach of consensus [#6524](https://github.com/hyperledger/besu/pull/6524)
 
 ### Download Links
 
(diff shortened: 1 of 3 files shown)


shemnon (approved):
Is there a need for full fluent API support for classic forks?

diega:
@shemnon I think _ProtocolSpecs_ fluency is pretty good, is that fluent API the one that you mean?

shemnon:
APIs from this directory - https://github.com/hyperledger/besu/blob/d7e5b261141da6cbb4a8131ed2feb1603e3e68f8/evm/src/main/java/org/hyperledger/besu/evm/fluent

Guido has a fuzzer he uses there.  But I'm not sure it would have caught the DIFFICULTY/RANDAO swap, even against core-geth.

diega:
That sounds really interesting, do you know where I can find some documentation on how to run the [fuzzer](https://github.com/guidovranken/cryptofuzz) with Besu's EVM?

macfarla (commented):
this could use a changelog entry

fab-10 (commented):
CI Publish job is failing due to a missing javadoc

```
> Task :evm:javadoc
/home/circleci/project/evm/src/main/java/org/hyperledger/besu/evm/ClassicEVMs.java:29: warning: no comment
public class ClassicEVMs {
       ^
```

PR #6528, opened 2024-02-06
error if snap or checkpoint sync specified with privacy

macfarla:
we enforce that you can't specify FAST sync if privacy is enabled - do the same check for SNAP and CHECKPOINT


pinges (approved):
LGTM

PR #6541, opened 2024-02-08
disable privacy tests since they are currently flaky

macfarla:
These tests are flaky right now for 2 reasons, and work is in progress to fix. 

See 

Run tessera as a process (currently using docker) for ATs https://github.com/hyperledger/besu/pull/5968

Also occasionally errors since the txpool is disabled during initial sync - it's been somewhat [mitigated](https://github.com/hyperledger/besu/pull/6479) but hasn't totally fixed it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -500,6 +500,3 @@ workflows:
     jobs:
       - assemble
       - dockerScan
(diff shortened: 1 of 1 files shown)


siladu (approved):
LGTM

Issue #6391, opened 2024-01-12
Suspicious block invalidation on hive `max-fee-per-blob-gas` test for Besu


marioevz:
### Description
We are adding validation error string check to all invalid block tests in https://github.com/ethereum/execution-spec-tests/ and therefore in hive too, so we can be sure that all clients are invalidating the block because of the correct reason, and not due to something else and we get a incorrect test-pass.

Almost all tests are passing for besu, with all exceptions making sense, and errors match what the test is expecting to verify, except for two tests that verify very similar things:
The blocks include a tx with a max-fee-per-blob-gas lower than the blob gas price for the block.

The rest of the clients invalidate the block with the proper exception, e.g. geth:
```
could not apply tx 0 [...]: max fee per blob gas less than block blob gas fee: address 0xa..B blobGasFeeCap: 1, blobBaseFee: 2
```

But Besu is invalidating due to, at least what seems like, a state root mismatch:
```
World State Root does not match expected value, header 0x242a63fa2afe3294022ba3f52bb2ffca13875b56352f17be8dcdfa72db330e4f calculated 0x98b25fddd3c355f73d63fe8a5f2f20ebfebcd6d4c06d4d98239b74fa13b7df02
```
This is weird because:
1) It's the only block on all tests that Besu invalidates with this error.
2) If the failure only happened on the state root check, it means that besu applied the tx when executing the block, which should not even attempted to.

### Steps to Reproduce (Bug)
Using branch https://github.com/marioevz/hive/tree/pyspec-check-tx-insertion-error, run:
```bash=
./hive --client-file
./configs/cancun.yaml --client besu --sim ethereum/pyspec --sim.limit "/cancun/eip4844_blobs/blo
b_txs/invalid_tx_max_fee_per_blob_gas/tests/cancun/eip4844_blobs/test_blob_txs.py::test_invalid_tx_max_fee_per_blob_gas"
```

**Expected behavior:** Besu should not try to apply the invalid transaction and invalidate the block earlier

**Actual behavior:** Besu only invalidates the block after the state root mismatch

**Frequency:** Every time

### Logs (if a bug)
```
java.lang.RuntimeException: World State Root does not match expected value, header 0x242a63fa2afe3294022ba3f52bb2ffca13875b56352f17be8dcdfa72db330e4f calculated 0x98b25fddd3c355f73d63fe8a5f2f20ebfebcd6d4c06d4d98239b74fa13b7df02
[...]

PR #6546, opened 2024-02-08
Blob gas fee checking

jflo:
fixes: #6391


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -168,6 +170,18 @@ private ValidationResult<TransactionInvalidReason> validateCostAndFee(
                 "total blob gas %d exceeds max blob gas per block %d",
                 txTotalBlobGas, gasLimitCalculator.currentBlobGasLimit()));
       }
+      if (maybeBlobFee.isEmpty()) {
+        throw new IllegalArgumentException(
+            "blob fee must be provided from blocks containing blobs");
+      } else if (maybeBlobFee.get().compareTo(transaction.getMaxFeePerBlobGas().get()) > 0) {
+        return ValidationResult.invalid(
+            TransactionInvalidReason.GAS_PRICE_BELOW_CURRENT_BASE_FEE,
```
fab-10, line 178: I think we need a new specific invalid reason for this

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/PermissionTransactionValidator.java
```diff
@@ -43,8 +43,9 @@ public PermissionTransactionValidator(
   public ValidationResult<TransactionInvalidReason> validate(
       final Transaction transaction,
       final Optional<Wei> baseFee,
+      final Optional<Wei> blobBaseFee,
       final TransactionValidationParams transactionValidationParams) {
-    return delegate.validate(transaction, baseFee, transactionValidationParams);
+    return delegate.validate(transaction, baseFee, Optional.empty(), transactionValidationParams);
```
fab-10, line 48:
```suggestion
    return delegate.validate(transaction, baseFee, blobBaseFee, transactionValidationParams);
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/TransactionValidator.java
```diff
@@ -35,6 +35,7 @@ public interface TransactionValidator {
   ValidationResult<TransactionInvalidReason> validate(
       Transaction transaction,
       Optional<Wei> baseFee,
+      Optional<Wei> blobFee,
```
fab-10, line 38:
```suggestion
      Optional<Wei> blobBaseFee,
```

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidatorTest.java
```diff
@@ -300,13 +306,60 @@ public void shouldRejectTransactionWithMaxPriorityFeeGreaterThanMaxFee() {
...
-        validator.validate(transaction, Optional.of(Wei.ONE), transactionValidationParams);
+        validator.validate(
+            transaction, Optional.of(Wei.ONE), Optional.empty(), transactionValidationParams);
     assertThat(validationResult)
         .isEqualTo(ValidationResult.invalid(MAX_PRIORITY_FEE_PER_GAS_EXCEEDS_MAX_FEE_PER_GAS));
     assertThat(validationResult.getErrorMessage())
         .isEqualTo("max priority fee per gas cannot be greater than max fee per gas");
   }
 
+  @Test
```
fab-10, line 317: Is there already a test that check the opposite, when a tx is valid?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidatorTest.java
```diff
@@ -300,13 +306,60 @@ public void shouldRejectTransactionWithMaxPriorityFeeGreaterThanMaxFee() {
...
+  public void shouldRejectTransactionWithMaxBlobPriorityFeeGreaterThanMaxBlobFee() {
```
fab-10, line 318: not sure, but it seems that should be the name

```suggestion
  public void shouldRejectTransactionWithMaxBlobPriorityFeeSmallerThanBlobBaseFee() {
```

fab-10 (changes requested):
overall looks good, some suggestions

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidatorTest.java
jflo: yes, `shouldAcceptTransactionWithAtLeastOneBlob` covers this with a tx that overpays.

fab-10 (approved):
LGTM, will be nice to have a unit test that check the tx is valid even if underpriced when it comes from the txpool

PR #6548, opened 2024-02-09
uprev artifact name

jflo:


macfarla (approved):

siladu (approved):
> - Adds timestamp to enable Cancun upgrade on mainnet

nbd but this is under bugfixes in the changelog

PR #6555, opened 2024-02-09
Database metadata refactor

### LINKED ISSUE #6408: DATABASE_METADATA.json enhancement
What do you think about migrating our DATABASE_METADATA.json content to something more talkative? 

I mean now we have:
`{"version":2} `
wouldn't it clearer to have something like:
`{"format": "BONSAI", "version": 1}`?
Where format field is BONSAI or FOREST and version is the actual iteration of the format.

I am suggesting that, since I need to introduce a new version of the db, to identify the variables storage upgrade, and since it applies to both formats, with the current metadata, I should introduce 3 and 4 respectively from Forest and Bonsai, but I will prefer to have:
- `{"format": "FOREST", "version": 2}`
- `{"format": "BONSAI", "version": 2}`

Since the storage is implemented via plugins, I think it makes sense to also add a field to identify the plugin itself, e.g. `{"plugin":"rocksdb"}`

Changing the DATABASE_METADATA.json format is a breaking change so we also need to define a migration strategy, and how to version the DATABASE_METADATA itself.

We can consider the current DATABASE_METADATA format, the `v1`, and it has 2 fields: 
- `version`
- `privacyVersion`

we could introduce `v2`, changing the filename to DATABASE_METADATA_v2.json, example of its content:
```
{
  "v2": {
    "plugin": "rocksdb",
    "format": "BONSAI",
    "version": 2
  }
}
```

fab-10:
## PR description

Besu storage could have different formats, currently Forest or Bonsai, and each format can have different versions. We used to have only one identifier for these two properties, but this is not optimal, for clarity and does not play well with the fact that the storage is implemented via plugins.

Besu should only know about the format, and tell the plugin which is the required format, then the version of the format is something internal to the plugin itself.

So I moved the `DataStorageFormat` to the plugin API project, and this is the reason for the many file changed, most of them are just changes to reflect this move.
Then inside the RocksDB plugin I introduced the new class `VersionedStorageFormat` to managed the versioning of RocksDB databases.

The database metadata file, reflects this change, having an entry for the format and one for the version, for example, the default Bonsai storage will have this metadata (note the versioning of the metadata itself `v2` object, so we can introduce a `v3` if needed in future): 
```
{
  "v2": {
    "format": "BONSAI",
    "version": 2
  }
}
```

To correct the missed introduction of the new version, when the variables storage format was introduced, now for both Forest and Bonsai, a new version has been introduced, the version `2` for each format, being `1` the version pre variables storage.

At startup the existing metadata, that we call `v1` format, is automatically translated to the new format.

Check have been added to detect not managed (up|down)grades, so we can inform the user of the right steps, if the process in not automated.

It is possible to downgrade, running the subcommand  `storage revert-metadata v2-to-v1` before installing the previous binaries.

Test to perform:

- [x] Vanilla new installation, verify that the default format and version is used and written in DATABASE_METADATA.json 
- [x] New installation with custom format selection, verify that the selected format and version is used and written in DATABASE_METADATA.json 
- [x] Upgrade existing installation >=23.4.4, verify that DATABASE_METADATA.json has been translated to the new format
- [x] Upgrade existing installation >=23.4.4 then downgrade, verify that after the downgrade Besu refuses to start since the metadata is not recognized
- [x] Upgrade existing installation >=23.4.4 then revert metadata, then downgrade, verify that after the downgrade Besu starts correctly
- [x] Upgrade existing installation <23.4.4, verify that variables storage migration is performed and DATABASE_METADATA.json has the new format
- [x] Upgrade existing installation <23.4.4 then downgrade, verify that after the downgrade Besu refuses to start since the metadata is not recognized
- [x] Upgrade existing installation <23.4.4 first revert variables-storage, then revert metadata, then downgrade, verify that after the downgrade Besu starts correctly
## Fixed Issue(s)

fixes #6408


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 24.1.3-SNAPSHOT
+
+### Breaking Changes
+- RocksDB database metadata format has changed to be more expressive, the migration of an existing metadata file to the new format is automatic at startup. Before prerforming a downgrade to a previous version then it is mandatory to revert to the original format using the subcommand `besu --data-path=/path/to/besu/datadir storage revert-metadata v2-to-v1`.
```
gfukushima, line 6:
```suggestion
- RocksDB database metadata format has changed to be more expressive, the migration of an existing metadata file to the new format is automatic at startup. Before performing a downgrade to a previous version it is mandatory to revert to the original format using the subcommand `besu --data-path=/path/to/besu/datadir storage revert-metadata v2-to-v1`.
```

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/BaseVersionedStorageFormat.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.hyperledger.besu.plugin.services.storage.DataStorageFormat;
+
+import java.util.OptionalInt;
+
+public enum BaseVersionedStorageFormat implements VersionedStorageFormat {
+  FOREST_ORIGINAL(DataStorageFormat.FOREST, 1),
+  FOREST_WITH_VARIABLES(DataStorageFormat.FOREST, 2),
+  BONSAI_ORIGINAL(DataStorageFormat.BONSAI, 1),
+  BONSAI_WITH_VARIABLES(DataStorageFormat.BONSAI, 2);
```
jframe, line 25: Do we need to bake in variable and non-variable variants of the datastorage format? The variable change was introduced several versions ago don't think we need to have special logic for versioning this now.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/BaseVersionedStorageFormat.java
```diff
@@ -0,0 +1,61 @@
...
+
```
jframe, line 26: If I wanted to add a new bonsai version, how would I add that? Would I add another value to the `BaseVersionedStorageFormat`?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueSegmentIdentifier.java
```diff
@@ -46,32 +49,37 @@ public enum KeyValueSegmentIdentifier implements SegmentIdentifier {
...
-    this(id, new int[] {0, 1, 2}, containsStaticData, eligibleToHighSpecFlag, false);
+    this(
+        id,
+        EnumSet.allOf(DataStorageFormat.class),
+        containsStaticData,
+        eligibleToHighSpecFlag,
+        false);
   }
 
-  KeyValueSegmentIdentifier(final byte[] id, final int[] versionList) {
```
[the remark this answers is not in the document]
fab-10: That's a good question, in theory the column families (identifiers) are a concept that should not be in the plugin API, since a storage plugin should be free to organized the data as it wants, we can decide to review also this part later.

For the moment the specific version of the format is hidden here, since it is an internal property of each plugin, but that should not limit too much, since if you want to change the column families a new version of Besu is released and the new version of the RocksDB embedded plugin can managed the upgrade/downgrade.

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/BaseVersionedStorageFormat.java
fab-10: The version 2 of each format are here for reference, basically running this version of Besu, as you said, means that each format is, or already was, migrated to the `_WITH_VARIABLES` version.
Does this answer your question?

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/BaseVersionedStorageFormat.java
fab-10: yes, for example:
```
BONSAI_CODE_BY_HASH(DataStorageFormat.BONSAI, 3)
```
then in `RocksDBKeyValueStorageFactory` you can decide what to do in case you detect an upgrade or downgrade, coding the logic in `handleVersion*` methods and provide `storage` subcommands in case are needed

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/DatabaseMetadata.java
```diff
@@ -147,15 +104,124 @@ private static File getDefaultMetadataFile(final Path dataDir) {
...
     try {
-      databaseMetadata = MAPPER.readValue(metadataFile, DatabaseMetadata.class);
+      try {
+        return tryReadAndMigrateV1(metadataFile);
+      } catch (DatabindException dbe) {
+        return tryReadV2(metadataFile);
+      }
     } catch (FileNotFoundException fnfe) {
-      databaseMetadata = new DatabaseMetadata(1, 1);
+      throw new StorageException(
```
gfukushima, line 114: This is just a comment, non blocking feedback, are we planning to automate a solution for this catastrophic case in the future? Maybe it's not worth trying and users will have to wipe db and resync....

plugins/rocksdb/src/test/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValuePrivacyStorageFactoryTest.java
```diff
@@ -48,15 +48,15 @@ public class RocksDBKeyValuePrivacyStorageFactoryTest {
...
     final Path tempDatabaseDir = temporaryFolder.resolve("db");
     final Path tempPrivateDatabaseDir = tempDatabaseDir.resolve("private");
     Files.createDirectories(tempPrivateDatabaseDir);
     Files.createDirectories(tempDataDir);
-    when(commonConfiguration.getStoragePath()).thenReturn(tempDatabaseDir);
-    when(commonConfiguration.getDataPath()).thenReturn(tempDataDir);
-    when(commonConfiguration.getDatabaseVersion()).thenReturn(DEFAULT_VERSION);
+    mockCommonConfiguration(tempDataDir, tempDatabaseDir);
+
+    Utils.createDatabaseMetadataV1Privacy(tempDataDir, 1, 1);
```
gfukushima, line 59: For context and readability we should keep the removed lines 40 and 41 and use them here rather than the numbers itself

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/DatabaseMetadata.java
fab-10: I do not think automatic recover is worth the effort, since the lack of that file means the user did an error, and recovering from it means to try to identify the format and version querying the db, and if we implement that, then the metadata itself is redoundant

plugins/rocksdb/src/test/java/org/hyperledger/besu/plugin/services/storage/rocksdb/RocksDBKeyValuePrivacyStorageFactoryTest.java
fab-10: good point, reused already present constants

gfukushima (approved):
lgtm

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/BaseVersionedStorageFormat.java
jframe: It does. So maybe BONSAI_ORIGINAL isn't needed but think it's still ok.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueSegmentIdentifier.java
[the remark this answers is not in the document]
jframe: Good point we do the versioning at the application level. I actually prefer that than we have right now.

jframe (approved):
LGTM

PR #6557, opened 2024-02-09
Move account warming from gas calculator to call operations

shemnon:
## PR description

The BerlinGasCalculator had a side effect of warming addresses when calculating call gas costs. By introducing a boolean we can keep the side effects in the operation instead of in the gas calculator. This makes gas cost estimation by downstream users of the API safer.


evm/src/main/java/org/hyperledger/besu/evm/gascalculator/BerlinGasCalculator.java
```diff
@@ -161,8 +192,8 @@ public long callOperationGasCost(
             outputDataLength,
             transferValue,
             recipient,
-            to);
-    final boolean accountIsWarm = frame.warmUpAddress(to) || isPrecompile(to);
+            to,
+            true);
```
macfarla, line 196: shouldn't this use the accountIsWarm param that's passed in?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/FrontierGasCalculator.java
```diff
@@ -239,6 +240,31 @@ public long callOperationGasCost(
...
+        frame,
+        stipend,
+        inputDataOffset,
+        inputDataLength,
+        outputDataOffset,
+        outputDataLength,
+        transferValue,
+        recipient,
+        to,
+        true);
```
macfarla, line 253: why default to true?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/BerlinGasCalculator.java
shemnon: No, we want the "warmed price" as we will charge for warming ourselves.  Added a comment to the code.

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/FrontierGasCalculator.java
shemnon: The notion of charging for "cold storage" and "cold accounts" was added in Berlin, so the default would be to get the price without this cold access surcharge.  For all calculators pre-berlin it's a distinction without difference, however, as all prices are the same.

macfarla (approved):

PR #6558, opened 2024-02-10
Remove uses of org.jetbrains annotations

shemnon:
## PR description

We have a mix of org.jetbrains.annotations.NotNull and javax.annotations.Nonnull. Change all to the latter. Add an errorprone check to enforce this.


macfarla (approved):
thanks @shemnon!

Issue #6301, opened 2023-12-14
Bad Block Comparator - Plugin


jflo:
In order to quickly diagnose bad blocks, it is necessary to compare block traces between two clients _at the same state_. Usually, this requires either quick action (most clients have limits on how much historical state they can maintain, since it is only needed for reorgs) or access to an archive node.

To solve the recency issue, it would be helpful to have a plugin in Besu that could be configured with a peer execution client, which could be asked for a trace of the same block as soon as a bad one is detected and added to the bad block manager. It could work something like this:

- when a bad block is detected, trace it and store the trace 
- contact the configured peer, and immediately trace the same block
- normalize the output, trace formats are not quite the same
- provide both versions of the raw trace, and a delta between their normal forms via debug apis.

This would automate a very large portion of the legwork needed to diagnose issues.

mbaxter:
Plugin development has been started in a new repo here: https://github.com/Consensys/besu-bad-block-plugin

macfarla:
Removing the `help wanted` label since @mbaxter is working on this - thanks!

macfarla:
Removing the `help wanted` label since @mbaxter is working on this - thanks!

non-fungible-nelson:
@jflo what shall we do with this one?

jflo:
Icebox seems like the right place

Issue #5153, opened 2023-03-03
Remove BadBlockManager from ProtocolSpec


siladu:
This is a bug because when you transition to a new fork schedule, you lose your existing bad block cache.

Could do with knowing why it was added there in the first place before we blindly remove it though.

siladu:
Probably not a bug so much as a feature with unintended consequences. 

It make sense to partition bad blocks by fork since they blocks may be valid/invalid depending on different fork rules.

It does mean that the previous fork's bad blocks are unavailable in the debug_getBadBlocks endpoint, however that's a very minor edge case, probably not worth the effort so closing this.

If we implemented the "archive" cache mentioned in #5460 and made it agnostic of fork/protocol schedule then that may fix this issue as well.

PR #6559, opened 2024-02-12
[Issue 6301] Refactor BadBlockManager

mbaxter:
## PR description
Refactor `BadBlockManager` handling - move it out of `ProtocolSpec` and into `ProtocolContext`.  This will make it simpler to expose bad block events through the plugins API.  

This refactor simplifies access to `BadBlockManager` and improves the clarity of the codebase. Accessing `BadBlockManager` through `ProtocolSpec` is misleading because it implies this object is changing depending on the current spec / milestone.  However, this is not the case - `BadBlockManager` is a global cache unrelated to the spec. 

Now, instead of accessing the `BadBlockManager` in a roundabout way by pulling the latest spec:
```
Blockchain chain = protocolContext.getBlockchain();
BlockHeader head = chain.getChainHeader()
BadBlockManager badBlockManager = protocolSchedule
        .getByBlockHeader(head)
        .getBadBlocksManager()
``` 
You can just grab it directly from `ProtocolContext.getBadBlockManager()`. 

**Notes for the reviewer**:
The first commit moves the instantiation of `BadBlockManager` from `ProtocolScheduleBuilder` to `BesuControllerBuilder`.  This introduces a lot of changes because `BadBlockManager` now has to be injected into `ProtocolScheduleBuilder`.  The second commit pushes `BadBlockManager` into `ProtocolContext`, which also introduces a lot of changes - mostly in tests.  The rest of the commits are much smaller and more manageable. 

## Fixed Issue(s)
Part of #6301


siladu:
Just wanted to note the following issue comment since removing BadBlockManager from ProtocolSpec was previously considered and decided against.

https://github.com/hyperledger/besu/issues/5153#issuecomment-1548878777

The most relevant point is:

> It make sense to partition bad blocks by fork since they blocks may be valid/invalid depending on different fork rules.

@daniellehrner may have some further thoughts about this.

At the time of #5153, the plugin use cases mentioned for this PR were not previously considered so that may changes things.

mbaxter:
> It make sense to partition bad blocks by fork since they blocks may be valid/invalid depending on different fork rules.

Isn't a block necessarily on a single deterministic fork?  When would a block move between forks?

siladu:
> Isn't a block necessarily on a single deterministic fork? When would a block move between forks?

Agree a block wouldn't move between forks. 
Trying to remember the context of that comment...`debug_getBadBlocks` isn't fork-aware, so for example it's potentially misleading to send a request after the Cancun timestamp has passed and get a response containing bad blocks that are no longer invalid according to Cancun spec. 
But think you're right...it's possible for the API user to filter based on blockNumber themselves so it's just a semantic change to the API, not a big deal.

siladu (approved):
LGTM. Thanks for the review-friendly commits and description 👍

PR #6563, opened 2024-02-13
Fix for tx incorrectly discarded when there is a timeout during block creation

fab-10:
## PR description

In case the selection of transactions during block creation goes in timeout, it is possible that some txs are incorrectly dropped from the txpool, since marked as invalid.
When the timeout occurs the block is created with all the txs selected until that moment, after that the world state is closed, while it is possible that a tx was still processing, in that case the processing tx validation could fail due fact that is no more possible to get the account from the closed world state, and the tx is wrongly tagged as invalid with UPFRONT_COST_EXCEED_BALANCE, that cause the tx to be dropped from the pool.

This PR, improves the handling of timeout and not selected txs, avoiding that them can be incorrectly dropped in case of timeout


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -437,17 +421,51 @@ private TransactionSelectionResult handleTransactionNotSelected(
 
     final var pendingTransaction = evaluationContext.getPendingTransaction();
 
-    transactionSelectionResults.updateNotSelected(
-        evaluationContext.getTransaction(), selectionResult);
-    pluginTransactionSelector.onTransactionNotSelected(evaluationContext, selectionResult);
+    final TransactionSelectionResult actualResult =
+        isTimeout.get() && !wasTimeout(selectionResult)
```
gauravahuja, line 425: The names of the conditions sound contradictory even though they might be what you want. The first one would mean there is a timeout but the next one would mean not a timeout. 

Maybe rename `wasTimeout` to `[in]correctTimeoutType`?

gauravahuja (commented):
lgtm, just the one comment.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
fab-10: I see what you mean, will find for a better wording, `!wasTimeout(selectionResult)` is there to identify if the timeout was already processed, in that case we do not need to reprocess it, so probably naming it `wasTimeoutAlreadyProcessed` reads better?

gauravahuja: Yes, definitely better

fab-10: rename done

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -437,17 +421,51 @@ private TransactionSelectionResult handleTransactionNotSelected(
...
+  }
+
+  private boolean transactionTookTooLong(final TransactionEvaluationContext evaluationContext) {
+    final var evaluationTimer = evaluationContext.getEvaluationTimer();
+    if (evaluationTimer.elapsed(TimeUnit.MILLISECONDS) > blockTxsSelectionMaxTime) {
+      LOG.atWarn()
+          .setMessage(
+              "Transaction {} is too late for inclusion, evaluated in {} that is over the max limit of {}ms"
+                  + ", removing it from the pool")
+          .addArgument(evaluationContext.getPendingTransaction()::toTraceLog)
```
macfarla, line 457: is toTraceLog() at warn level too verbose?

macfarla: what does this log line look like in reality?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -383,28 +383,12 @@ private TransactionSelectionResult handleTransactionSelected(
...
-        timeoutSelectionResult = TX_EVALUATION_TOO_LONG;
-      } else {
-        LOG.atTrace()
-            .setMessage("Transaction {} is too late for inclusion")
-            .addArgument(transaction::toTraceLog)
-            .addArgument(evaluationTimer)
-            .log();
-        timeoutSelectionResult = BLOCK_SELECTION_TIMEOUT;
-      }
+      final TransactionSelectionResult timeoutSelectionResult =
```
macfarla, line 388: this value isn't used below? and the same thing is evaluated again when getting `actualResult`

macfarla (commented):
I think it makes sense. couple of comments

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
fab-10: Like that, it should not be frequent, do you think it is better to only print the hash?
```
Transaction 0xee8fa4f7c1aa6ce9e06b70d7fb1984934fc9f09614f53f3ff26fd87874797d92={MessageCall, 386, 0x7be0af0785193c5b00672c70364d545aaf35b3d6, EIP1559, mf: 4.00 gwei, pf: 3.00 gwei, gl: 7532434, v: 0 wei, to: 0x69d35b2d44442824e6aa105674779eb8e5091cd2} is too late for inclusion, evaluated in 2.256 s that is over the max limit of 1800ms, removing it from the pool
```

macfarla: if it's not too frequent it's prob ok 👍

macfarla (approved):

PR #6570, opened 2024-02-14
remove useless persist during backward sync

matkt:
## PR description

During backward sync we call validateAndProcessBlock, which will process the block, update the state, and calculate the root hash.

If the block is validated, we will persist the block again for no reason, which will lead to a new unnecessary calculation of rootHash. Removing this second call should really significantly improve performance.


ahamlat (approved):
Nice finding 👍

garyschulte (approved):
🚢

PR #6572, opened 2024-02-14
add curl in base openjdk-latest dockerfile

### LINKED ISSUE #6571: openjdk latest debug flavor
--Improvement /  QoL--
As a user of the besu docker image I would like to have curl available in the running container.

siibars:
## PR description
As a user of the besu docker image I would like to have curl available in the running container.

NB: teku include curl in their image https://github.com/Consensys/teku/blob/b544b9ebe589e78fa2f075ae389b041dc2871ea2/docker/jdk21/Dockerfile#L17

## Fixed Issue(s)
fixes #6571


.circleci/config.yml
```diff
@@ -301,6 +301,10 @@ jobs:
           name: hadoLint_openjdk_latest
           command: |
             docker run --rm -i hadolint/hadolint < docker/openjdk-latest/Dockerfile
+      - run:
+          name: hadoLint_openjdk_latest
+          command: |
+            docker run --rm -i hadolint/hadolint < docker/openjdk-latest-debug/Dockerfile
```
garyschulte, line 307: @jflo we are no longer using circle-ci correct?

docker/openjdk-latest-debug/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN apt-get update && \
+ apt-get install --no-install-recommends -q --assume-yes curl=7* openjdk-21-jre-headless=21* libjemalloc-dev=5.* adduser=3* && \
+ apt-get clean  && \
+ rm -rf /var/lib/apt/lists/*  && \
+ # Ubuntu 23.10 comes with an "ubuntu" user with uid 1000. We need 1000 for besu.
+ userdel ubuntu 2>/dev/null || true && rm -rf /home/ubuntu && \
+ # Ensure we use a stable UID for besu, as file permissions are tied to UIDs.
+ adduser --uid 1000 --disabled-password --gecos "" --home /opt/besu besu && \
+    chown besu:besu /opt/besu && \
+    chmod 0755 /opt/besu
```
garyschulte, line 14: is there a reason apart from curl that we want a latest debug image?  debug images have a larger surface area in terms of exposed ports and functionality.  If we really just want `curl` present in operational images, I think we should just add it rather than creating a debug image that would be used operationally just to have curl available

garyschulte (changes requested):

.circleci/config.yml
siibars: okay, I removed the circleci changes . @jflo  let me know If you want me to add it back.

docker/openjdk-latest-debug/Dockerfile
siibars: You are right.
So for this debug flavor to make sense I added few other packages (curl,wget,jq,net-tools) and exposed 5005 (used by remote debuggers), 9545 (metrics), 1098 (Java JMX monitoring), similarly to the openjdk-17-debug flavor.
With this feature scope, would you consider the openjdk-latest-debug ?

garyschulte: As a debug image it seems fine, but the issue and pr description made it sound like the motivation was only to have these tools available operationally.  

I am not opposed to an openjdk-latest-debug image.  Rather I would suggest we add some of these quality of life tools to the base Dockerfiles also, like curl, wget, jq so that we don't end up running debug images out of convenience.

siibars: Ok changes suggested added in the PR 👍

jframe:
We no longer support debug specific images, you can always add this to your own dockerfile if need this for your environment.

PR #6574, opened 2024-02-14
Apply Reference tests v13.1

shemnon:
## PR description

Apply v 13.1 of the reference tests, includes test related fixes around initing the KZG precompile, uint64s in timestamps and blob gas, and some formatting cleanup.


evm/src/main/java/org/hyperledger/besu/evm/internal/Words.java
```diff
@@ -213,4 +215,23 @@ static Bytes longBytes(final long value) {
...
+      radix = 16;
+      parsable = number.substring(2);
+    } else if (!number.matches("\\d+")) {
+      // presume naked hex
+      radix = 16;
+    }
+
+    BigInteger bi = new BigInteger(parsable, radix);
+    if (bi.bitCount() > 64) {
+      throw new NumberFormatException("Nuber larger than uint64");
```
macfarla, line 232:
```suggestion
      throw new NumberFormatException("Number larger than uint64");
```

macfarla (approved):
one small nit in error message

PR #6576, opened 2024-02-15
Remove duplicated addCreate call

gfukushima:
## PR description
Remove duplicated call to the addCreate method. The method is already called here  https://github.com/hyperledger/besu/blob/5c295c5aa6c277287bf3c8d6679a9cca36427838/evm/src/main/java/org/hyperledger/besu/evm/processor/ContractCreationProcessor.java#L125


shemnon (approved):
this is the older version.  125 is the newer call erroneously added.

PR #6540, opened 2024-02-08
rocksdb usage - Replace Column Size with more accurate Total Size

siladu:
Replace "Column Size" with "Total Size" where Total Size = SST Files Size + Blob Files Size

`besu --data-path=... storage rocksdb usage`

e.g. with a BONSAI + X_SNAP node...

BEFORE THIS PR

```
| Column Family                  | Keys            | Column Size  | SST Files Size  |
|--------------------------------|-----------------|--------------|-----------------|
| BLOCKCHAIN                     | 2355141414      | 905 GiB      | 166 GiB         |
| VARIABLES                      | 26              | 120 KiB      | 240 KiB         |
| ACCOUNT_INFO_STATE             | 9634454         | 294 MiB      | 496 MiB         |
| ACCOUNT_STORAGE_STORAGE        | 24041432        | 977 MiB      | 1 GiB           |
| CODE_STORAGE                   | 37703864        | 10 GiB       | 12 GiB          |
| TRIE_BRANCH_STORAGE            | 1885032116      | 110 GiB      | 138 GiB         |
| TRIE_LOG_STORAGE               | 267301          | 14 GiB       | 17 GiB          |
```

Column Size Total =  905 + 0.294 + 0.977 + 10 + 110 + 14 = 1042 GB

AFTER THIS PR

```
| Column Family                  | Keys            | Total Size  | SST Files Size  | Blob Files Size  |
|--------------------------------|-----------------|-------------|-----------------|------------------|
| BLOCKCHAIN                     | 2355141414      | 933 GiB     | 166 GiB         | 767 GiB          |
| VARIABLES                      | 26              | 240 KiB     | 240 KiB         | 0 B              |
| ACCOUNT_INFO_STATE             | 9634454         | 496 MiB     | 496 MiB         | 0 B              |
| ACCOUNT_STORAGE_STORAGE        | 24041432        | 1 GiB       | 1 GiB           | 0 B              |
| CODE_STORAGE                   | 37703864        | 12 GiB      | 12 GiB          | 0 B              |
| TRIE_BRANCH_STORAGE            | 1885032116      | 138 GiB     | 138 GiB         | 0 B              |
| TRIE_LOG_STORAGE               | 267301          | 17 GiB      | 17 GiB          | 0 B              |
|--------------------------------|-----------------|-------------|-----------------|------------------|
| ESTIMATED TOTAL                | 4311820607      | 1104 GiB    | 337 GiB         | 767 GiB          |
```

1104 GB is much closer to the on disk result:

[...]

PR #6577, opened 2024-02-15
[MINOR] Use better savings estimate for x-trie-log prune

siladu:
Following on from #6540, use `total-sst-files-size + total-blob-file-size` instead of `estimate-live-data-size` for x-trie-log prune estimate.


gfukushima (approved):
lgtm

Issue #6573, opened 2024-02-14
Discovery is not taking NATManager portMapping into account


daanporon:
### Description
This bug report is related to the limitations [described here](https://besu.hyperledger.org/private-networks/tutorials/kubernetes#limitations). But I wanted to write down my findings, not sure these can be resolved or maybe I am missing something.

We are encountering an issue while setting up Besu networks on different Kubernetes clusters with the aim of enabling discovery between nodes across clusters. 

We setup one node on cluster A and a second node on cluster B. When starting the node on cluster B we configured the enode from the node on cluster A as a bootnode. Both nodes have a LoadBalancer configured with a different discovery port for udp than the rlpx port for tcp as described in the limitations. 
Next to this we also configured the  `KubernetesNATManager` which discovers the [portMapping from the LoadBalancer](https://github.com/hyperledger/besu/blob/main/nat/src/main/java/org/hyperledger/besu/nat/kubernetes/KubernetesNatManager.java#L125). All of this seems to be working correct, if i use `admin_nodeInfo` it gives me the right [enode url with the discport which was found by the `KubernetesNATManager`](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AdminNodeInfo.java#L148). 

The problems are starting to happen when the discovery mechanism starts to do its thing. The PING/PONG mechanism is not taking the portForwarding from the `KubernetesNATManager` into account. You see the node on cluster B sending a PING with the "wrong" udpPort configured. I'm putting wrong between "", because it is the udpPort configured on the besu service itself, but not the one from the LoadBalancer. 

```
[...]

PR #6578, opened 2024-02-15
fix: Discovery not taking nat manager port-mapping into account

daanporon:
## PR description
Proposal to fix the issue i reported in #6573 
* `PingPacketData` should take into account the NATManager portMapping for the discoveryPort.
* Take into account the `udpPort` from the `PingPacketData` if the `host` is also taken from there.

## Fixed Issue(s)
fixes: https://github.com/hyperledger/besu/issues/6573


daanporon:
This is my first PR and i haven't tested it yet .. at the moment it's more a proposal on how i think it can be fixed.

daanporon:
This PR is ready for review. Only had one question. I saw that even when discovery is disabled we are still starting the `peerDiscoveryAgent`, does this make sense? 

I noticed, because in the tests `enodeUrlShouldHaveAdvertisedHostWhenDiscoveryDisabled` discovery is disabled but still the enodeUrl returned a discport because the discoveryAgent was started and so we fetched the discoveryPort from the NatManager. The NatManager was enabled because on the CI we are running in Docker and by default natMode == AUTO.

matthew1001:
@daanporon this seems like a reasonable change.

Re. the `PeerDiscoveryAgent` always starting, I'm not sure I'm afraid. It may be that disabling discovery doesn't prevent "outbound" UDP-based discovery via bootnodes, in which case I guess the agent is still needed?

matthew1001:
> @matthew1001 is something else needed in order to get this approved?

It looks OK to me, but I think it would be sensible to get @pinges to take a look and do the actual approval.

pinges (commented):
These changes look good to me.
@daanporon what scenarios have you tested that in? What kind of nodes have you started to test this?

daanporon:
> These changes look good to me. @daanporon what scenarios have you tested that in? What kind of nodes have you started to test this?

I have only run the tests to see if it works and started a local node using: https://wiki.hyperledger.org/display/BESU/Building+from+source#running-developer-builds, but this doesn't use the NAT manager. 

If possible it would be good if i can tests this in my k8s environment. Any idea what would be the easiest to build this and push it to a container registy so that i can easily test this?

macfarla:
@daanporon develop builds (updated every push to main) are here https://github.com/hyperledger/besu/releases/tag/develop - can you work with that?

daanporon:
> @daanporon develop builds (updated every push to main) are here https://github.com/hyperledger/besu/releases/tag/develop - can you work with that?

but that will only work once it is merged in main i think? but i will try to trigger the workflow in my own repo and push it to a container registry i own so that i can test it.

daanporon:
I have been testing this on AWS and Azure and the change is working as expected. I still have [other issues with discovery](https://github.com/hyperledger/besu/issues/6573) on kubernetes but at least this works already.

jframe:
@pinges Can you take another look at this PR

pinges:
Looks like this is needed in some environments, I think the only question is how this can be tested. Would be great to have an "AT" that shows that it is working as it should. I guess setting up an environment that we could use to test that automatically is a lot of work. 
We could just do some manual testing ourselves, or rely on Daanporon?
@jflo @jframe

daanporon:
I was picking this back up today, and resynced the PR ... because without the PR i'm not able to get it to work. The PONG will always be send to the host/udp-port combination.

macfarla:
Hi @daanporon 

We have made the decision to [deprecate k8s with nat](https://github.com/hyperledger/besu/pull/7534). So we don't want to add to this section of the code base. We've got solutions in place that bypass nat manager for k8s and thats working for private, and public nodes that we run. Our preference is to keep it simple and clean for the long term and not more band aid patches for another service mechanism or cloud provider. 

We recognize that you have gone through a lot of trial and error to create this PR but that's kinda on us for leaving this as is and not making the call earlier - and is also a symptom of the issue that we besu developers aren't experts in this part of the code base, it's not our core expertise.

If you want to keep this functionality and extend, you are welcome to fork and maintain - or consider converting the functionality to a plugin long-term?

I'm closing this PR as we won't be merging it, hope you understand the maintainers point of view.

daanporon:
@macfarla thanks for the clarification. We already worked around this by using NodePorts in stead of the Loadbalancers. But the last time i checked i was still having issues with the P2P discovery, and it trying to communicate back on the wrong UDP port. But i'll check again, and maybe create a partial PR with this fix: https://github.com/hyperledger/besu/pull/6578/files#diff-b10aec51aeaa23631d517c3107f0ab25c7f2afda83a8535e95d302b5c825a85cR298, because this was not only for the NAT manager.

Issue #6291, opened 2023-12-13
Expose runtime values of minGasPrice and minPriorityFee as metrics


fab-10:
With #6080 and #6097, is now possible to change the values of the `--min-gas-price` and `--min-priority-fee` options at runtime, for this is convenient to expose the runtime values of both options as metrics so they can be tracked and monitored

Brindrajsinh-Chauhan:
@fab-10 Would be happy to work on this if this not something you are actively working on

fab-10:
Hi @Brindrajsinh-Chauhan, I am not working on it at the moment, so feel free to work on it.

Brindrajsinh-Chauhan:
@fab-10 Opened a PR (mentioned above) for the same, would be grateful to get your reviews/feedback on it.

Thank you

matthew1001:
It sounds like this can be closed as PR https://github.com/hyperledger/besu/pull/6587 has now merged.

PR #6587, opened 2024-02-19
Mining Parameters Metrics

Brindrajsinh-Chauhan:
## PR description
This PR adds metrics to the Mining Parameters `minGasPrice` and `minPriorityFee` as `ethereum_min_gas_price` and `ethereum_min_priority_fee` respectively to expose runtime Values of those properties set via cli arguments or RPC methods as a gauge.

## Fixed Issue(s)

https://github.com/hyperledger/besu/issues/6291


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2122,11 +2125,19 @@ private MiningParameters getMiningParameters() {
     if (miningParameters == null) {
       miningOptions.setGenesisBlockPeriodSeconds(
           getGenesisBlockPeriodSeconds(getActualGenesisConfigOptions()));
+      miningOptions.setMiningParametersMetricsSystem(getMiningParametersMetrics());
```
fab-10, line 2128: just as an example on how you can simplify the PR, avoiding to move around the `MiningParameterMetrics` object, is to create a gauge like this:

```
      metricsSystem.get().createGauge(BesuMetricCategory.ETHEREUM,
              MIN_GAS_PRICE_GAUGE,
              "Gauge to measure the runtime value of min-priority-fee",
              () -> miningParameters.getMinTransactionGasPrice().toBigInteger().doubleValue());
```

since you can directly use the `miningParameters` object as source for the value, without duplicating the values in another class `MiningParameterMetrics` and having to change all the files to pass it around

fab-10 (changes requested):
Hi @Brindrajsinh-Chauhan thanks for working on this.
Your solution works, but there should be a simpler way to achieve the same, touching very few files and avoiding keeping 2 copies of the values. 
See my suggestion for a simple snippet of code as example of how to implement the simpler solution, and let me know if it is not clear

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
Brindrajsinh-Chauhan: Thank you for the feedback. This makes complete sense. While understanding the other code sections, i thought of this but was not sure if it worked that way or not. Thank you for clarifying this. 

If my understanding is correct, this way i would not have to take miningParameterMetrics to the JSONRPC methods right and could just use the miningParameters to get the value. 
Please correct me if I am wrong.
Thank you

Brindrajsinh-Chauhan: Will work on the feedback and push changes as soon as possible. Thank you

Brindrajsinh-Chauhan: Updated the PR with feedback. would be grateful to get a review. Thank you

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/MiningParametersMetrics.java
```diff
@@ -0,0 +1,45 @@
...
+    metricsSystem.createGauge(
+        BesuMetricCategory.ETHEREUM,
+        MIN_PRIORITY_FEE_GAUGE,
+        "Gauge to measure the runtime value of min-priority-fee",
+        () -> miningParameters.getMinPriorityFeePerGas().toBigInteger().doubleValue());
+  }
+
+  public MetricsSystem getMetricsSystem() {
+    return metricsSystem;
+  }
```
fab-10, line 44: is this needed?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/MiningParametersMetrics.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.hyperledger.besu.ethereum.core;
+
+import org.hyperledger.besu.metrics.BesuMetricCategory;
+import org.hyperledger.besu.plugin.services.MetricsSystem;
+
+public class MiningParametersMetrics {
+  public static final String MIN_GAS_PRICE_GAUGE = "min_gas_price";
+  public static final String MIN_PRIORITY_FEE_GAUGE = "min_priority_fee";
+  private final MetricsSystem metricsSystem;
```
fab-10, line 23: it seems not necessary to keep this reference, since everything is done in the constructor

fab-10 (commented):
LGTM, just some code that I think could be removed

fab-10 (approved):

PR #6593, opened 2024-02-19
Increment private nonce even if transaction failed.

### LINKED ISSUE #1942: Nonce of private transaction not incrementing when private transaction reverts
[Raised in RocketChat](https://chat.hyperledger.org/channel/besu?msg=3gsQR3svKFEp5gq4i):

Hi community,
We have noticed that if a private tx reverts (e.g, when a require condition fails), then the private nonce is not incremented. On the other hand, the public nonce, that of the privacy marker tx, do increment. So my question is: ¿is there any way to change this behaviour, allowing the private nonce to increment? We have found this situation problematic when leading with concurrent private tx, because if tx n fails, then n+1, n+2, etc will also fail.
Thank you very much in advance community! 

- [x] Confirm behaviour is inconsistent with public transaction behaviour 
- [ ] Document current behaviour as a known issue
- [ ] (longer term) If yes, fix.  Fix will require a CLI option or preferably something better (something in the database possibly) to not break backwards compatibility.  -> this might be support for forks in privacy - and this is also a way to deprecate privacy methods in future

gtebrean:
## Fixed Issue(s)

Fixes #1942 

Classes to follow:

BesuCommand
PrivacyParameters
ForestMutableWorldState
PrivateState
PrivacyPrecompiledContract

Currently private nonce update applies only to Forest State


gtebrean:
This was raised and discussed in Besu Contributors call on April 16th, and some of the concerns were the following:
1. Aligns with Ethereum standards.
2. Security concerns if we submit a private nonce of a failing transaction.
3. Breaking current functionality.

I did some researches and the current approach applies at Private Transaction Processor level and will directly save only the private nonce in the world state without affecting other things, further more this will be enabled only when required by genesis file or flag.

Concerns would be also if someone would like to query for tx using that nonce.

pinges (commented):
A few comments:

- I don't think there is a safety issue with incrementing the nonce even if the tx fails. The behaviour is different, but as this has to be enabled it should be OK.
- I thought that we would only need to know about this feature in the privacy pre compile contracts. Privacy parameters are passed in to these pre compile contracts, so we should not have to pass that flag into protocol specs.
- I assume that the logging at info level will be removed later, or at least upgraded to debug or higher.

gtebrean:
@pinges @macfarla please take a look again to my implementation.

Implementation is done and works for Forest state, if you agree with the approach I will do it also for Bonsai.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
```diff
@@ -181,14 +185,23 @@ public PrecompileContractResult computePrecompile(
         processPrivateTransaction(
             messageFrame, privateTransaction, privacyGroupId, privateWorldStateUpdater);
 
+    if (!result.isSuccessful() && incrementPrivateNonce) {
+      privateWorldStateUpdater.commitPrivateNonce();
+      disposablePrivateState.persist(null);
+
+      storePrivateMetadata(
```
gtebrean, line 192: the reason for which I'm calling the storePrivateMetadata here, which is storing also  the metadata of the transaction even if is null, is because is not enough only to persist the new nonce on the trieState. In order for the nonce to be available for next iteration the lastRootHash needs to be updated with the new state in privateWorldStateArchive, and by calling only  updatePrivacyGroupHeadBlockMap would have create incomplete states, resulting  validation failing.

pinges (changes requested):
Please have a look at the comments.

ethereum/core/src/integration-test/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContractIntegrationTest.java
```diff
@@ -205,6 +206,7 @@ public void testSendAndReceive() {
             new PrivateStateRootResolver(privateStateStorage),
             new PrivateStateGenesisAllocator(
                 false, (privacyGroupId, blockNumber) -> Collections::emptyList),
+            false,
```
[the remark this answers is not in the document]
gtebrean: I tried to fix it, but Spotless is rearranging them like this.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -715,6 +715,12 @@ static class PrivacyOptionGroup {
         names = {"--privacy-flexible-groups-enabled"},
         description = "Enable flexible privacy groups (default: ${DEFAULT-VALUE})")
     private final Boolean isFlexiblePrivacyGroupsEnabled = false;
+
+    @Option(
+        names = {"--privacy-nonce-incrementation-enabled"},
```
pinges, line 720: What about renaming that to "--privacy-always-increment-nonce" ?
The original name does not really say what this is doing, as we are incrementing the nonce when the tx is successful, just not when it is invalid or not failed.
You might have a better idea for a name :-)

pinges (approved):

macfarla (commented):
needs a changelog entry :)

PR #6595, opened 2024-02-19
Refactor `TransactionSelectionService`

fab-10:
## PR description

Scope of this refactor is to move the `TransacationSelectionService` from protocol context to mining parameters where it is more pertinent.
The relevant changes themselves are small, but there are a lot of touched files due to the update protocol context


siladu (approved):
Good change, net deletion of code is winning :)
Also like the fact this removes something from ProtocolContext +1

Issue #6591, opened 2024-02-19
Inconsistent Trace Details and Missing Error Messages in `trace_block` Method Output


achmand:
### Description
As a developer working with Ethereum node trace methods, I want the `trace_block` method to accurately reflect trace details, including `subTraces` and `error` messages, so that I can reliably analyze transaction execution and failures.

### Acceptance Criteria
- The `trace_block` method should accurately report the number of `subTraces` for each trace.
- Error messages, such as "out of gas", should be correctly reported in the trace details when applicable.
- The trace output from `trace_block` should be consistent with what is returned by `debug_traceBlockByNumber` when using the `callTracer`.

### Steps to Reproduce (Bug)
- Execute a `trace_block` request for block `0x20F6DF` (block number 2160351) using the provided JSON-RPC request.
- Note the `subTraces` count and lack of `error` message for the delegate call trace.
- Execute a `debug_traceBlockByNumber` request for the same block using the `callTracer`.
- Compare the results, noting discrepancies in the `subTraces` count (only in parity style traces) and the presence of an "out of gas" error in the `debug_traceBlockByNumber` output but not in the `trace_block` output.

**Request Example:**
```
{"jsonrpc":"2.0","method":"trace_block","params": ["0x20F6DF"],"id":0}
```

**Response Example (trace_block):**
```
[
    {
        "jsonrpc": "2.0",
        "id": 0,
        "result": [
            {
                "action": {
                    "callType": "call",
                    "from": "0xd8422593f16cd6d65c3f4cfa28b5b5a862037123",
                    "gas": "0x1a18",
                    "input": "0x5b7d7482000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000004034313465376364326235656636313337656166383661396134653064666430333536313063323562323536333031633662396361356361313531653233373965000000000000000000000000000000000000000000000000000000000000004034646163313738646333376331303736346235643862373862633135323635366437316662343865626537396361613835343437336530663031313132313732",
                    "to": "0xd1a3abf42f9e66be86cfdea8c5c2c74f041c5e14",
                    "value": "0x0"
                },
                "blockHash": "0x2345afc92f072449294eba81a33116da5d014e74ecbe5db53e5d72b40145976f",
                "blockNumber": 2160351,
                "result": {
                    "gasUsed": "0x0",
                    "output": "0x"
                },
                "subtraces": 0,
[...]

PR #6599, opened 2024-02-21
Ensure halt reasons in TraceFrames in `DebugOperationTracer.tracePostExecution`

Gabriel-Trintinalia:
## PR description

This PR fixes the debug trace series generation for `opcodes` that do not set an explicit `ExceptionalHaltReason` in the `OperationResult`. Previously, operations like `dup5` could lead to an exceptional halt (e.g., due to running out of gas) without this being accurately captured in the trace output, as the halt reason was not set explicitly. This situation affected the `FlatTraceGenerator's` ability to accurately reflect halt conditions and `subtrace` counts.

The `DebugOperationTracer.tracePostExecution` method has been updated to check the `MessageFrame` for an exceptional halt reason if it's absent in the `OperationResult`. This adjustment aligns with the existing approach used by the `StandardJsonTracer`

The trace output for the transaction referenced in the [issue](https://github.com/hyperledger/besu/issues/6591) now accurately displays the correct halt reason and appropriately counts subtraces.

```diff
{
   "jsonrpc":"2.0",
   "id":1,
   "result":[
      {
         "action":{
            "callType":"call",
            "from":"0xd8422593f16cd6d65c3f4cfa28b5b5a862037123",
            "gas":"0x1a18",
            "input":"0x5b7d7482000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000004034313465376364326235656636313337656166383661396134653064666430333536313063323562323536333031633662396361356361313531653233373965000000000000000000000000000000000000000000000000000000000000004034646163313738646333376331303736346235643862373862633135323635366437316662343865626537396361613835343437336530663031313132313732",
            "to":"0xd1a3abf42f9e66be86cfdea8c5c2c74f041c5e14",
            "value":"0x0"
         },
         "blockHash":"0x2345afc92f072449294eba81a33116da5d014e74ecbe5db53e5d72b40145976f",
         "blockNumber":2160351,
-        "result": {
-           "gasUsed": "0x0",
-           "output": "0x"
-        },
-        "subtraces": 0,
+       "error":"Out of gas",
+       "subtraces":1,
        "traceAddress":[
            
         ],
         "transactionHash":"0xea89b8082ab1e9e8f0a025ead197ce8e2cdf4b1212d4ec13a91e97570485a053",
         "transactionPosition":0,
         "type":"call"
      },
      {
         "action":{
            "callType":"delegatecall",
            "from":"0xd1a3abf42f9e66be86cfdea8c5c2c74f041c5e14",
            "gas":"0x6dd",
            "input":"0x5b7d7482000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000004034313465376364326235656636313337656166383661396134653064666430333536313063323562323536333031633662396361356361313531653233373965000000000000000000000000000000000000000000000000000000000000004034646163313738646333376331303736346235643862373862633135323635366437316662343865626537396361613835343437336530663031313132313732",
            "to":"0x564477025731ee7197eecf2c4a0d0106cc3e4572",
            "value":"0x0"
         },
         "blockHash":"0x2345afc92f072449294eba81a33116da5d014e74ecbe5db53e5d72b40145976f",
         "blockNumber":2160351,
-        "result": {
-           "gasUsed": "0xffffffffffffece0",
-            "output": "0x"
-         },
+       "error":"Out of gas",
         "subtraces":0,
         "traceAddress":[
            0
         ],
         "transactionHash":"0xea89b8082ab1e9e8f0a025ead197ce8e2cdf4b1212d4ec13a91e97570485a053",
         "transactionPosition":0,
         "type":"call"
      }
   ]
}

```


shemnon (approved):

shemnon:
(a) tests would be nice

(b) A word of caution updating the trace_ series calls, they were extensively tested to be identical to Parity's trace so we could be a drop in replacement.  But parity's been dead for a number of years, so it may not matter.  Any change that changes a parity trace result should be done with this use case in mind and potential side effects understood.  Not a blocker however, especially in the face of user requests.

macfarla (approved):
one line change. love it.

agree would be nice to add a test - happy for this to be a separate PR though

MRabenda:
@shemnon one comment from my end as we found this bug. All other "parity" clients including `trace_block` and `debug_trace` with `flatCallTracer` have a proper response (we are doing many validation checks on top). Only Besu had a problem in this case

shemnon:
That's great to hear that this brings us in line with other traces.  There should be a test written to ensure we don't face an accidental regression in the future then.

PR #6622, opened 2024-02-27
[#6301] Track bad block cause

mbaxter:
## PR description
Track bad block causes in `BadBlockManager` in preparation for exposing plugin events.  This data is useful because a consumer may want to take a different action depending on the cause.  For example, for #6301, we probably want to trace blocks with an actual validation error and ignore blocks that just descend from a bad block. 

Summary of changes:
- Require a `BadBlockCause` when pushing blocks to `BadBlockManager`
- Simplify bad block checks by adding utility method `BadBlockManager.isBadBlock`
- Add some basic tests for `BadBlockManager`
- In event handler `MergeCoordinator.onBadChain`, only push descendants of the bad block to the`BadBlockManager`.  The bad block itself will already have been pushed into the `BadBlockManager` when processed in `MainnetBlockValidator` and the `MergeCoordinator` does not have the context on why the bad block was marked as bad.
- Remove unused API `MergeMiningCoordinator.addBadBlock`
- Fix / add tests to `MainnetBlockValidatorTest` 
   - Add tests for `fastBlockValidation()`
   - Set all mocks upfront to ensure every test executes from a baseline of success. Previously several tests were passing artificially.  The mocks were not set up correctly so some tests were not hitting the cases that were intended. 
- Create some factory methods on `InvalidBlockException` to handle cases where (previously required) block data is not available
- Disambiguate a few different types of errors in `DownloadHeaderSequenceTask` and add tests

## Fixed Issue(s)
Part of #6301


mbaxter:
@siladu - I have already pinged @garyschulte and @jflo for a review if you are busy

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/BadBlockManagerTest.java
```diff
@@ -0,0 +1,101 @@
...
+  public void addBadBlock_ignoresInternalError(final String caseName, final Exception err) {
+    BadBlockManager badBlockManager = new BadBlockManager();
+    final BadBlockCause cause = BadBlockCause.fromProcessingError(err);
+    badBlockManager.addBadBlock(block, cause);
+
+    assertThat(badBlockManager.getBadBlocks()).isEmpty();
+  }
+
+  @Test
+  public void addBadBlock_doesNotIgnoreRuntimeException() {
```
jflo, line 64: This could be a problem. If a runtime problem is repeatable, and occurs for any out-of-protocol reason/violation, we will get a chain halt. The CL will continue to try to add the block, as it is canonical, and besu will refuse since it is already present in the bad block manager. We ran into this a lot building out v3 engine apis. I don't know how to better answer the question "did this RuntimeException reflect a protocol violation".

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/DownloadHeaderSequenceTask.java
```diff
@@ -199,41 +202,38 @@ private CompletableFuture<List<BlockHeader>> processHeaders(
...
-                            "Received invalid headers from peer (BREACH_OF_PROTOCOL), disconnecting from: {}",
-                            headersResult.getPeer());
-                        headersResult.getPeer().disconnect(DisconnectReason.BREACH_OF_PROTOCOL);
-                        future.completeExceptionally(
-                            new InvalidBlockException(
-                                "Header failed validation.",
-                                invalidBlock.getNumber(),
-                                invalidBlock.getHash()));
-                      });
+            final boolean foundChild = child != null;
```
jflo, line 206: This whole class is a lot more readable now, thx

jflo (changes requested):
Overall big improvement in observability and readability! I think we need to think through the treatment of any RuntimeException as a bad block, that is a lot broader than known application exceptions like StorageException and MerkleTrieException. If that gets thrown unexpectedly and a protocol-compliant block is marked as bad, there will be a chain halt.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/BadBlockManagerTest.java
mbaxter: 100% Agree.  I have a follow-up PR coming to fix this.  I wanted to isolate behavior changes in a separate PR.

jflo (approved):

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/BadBlockManagerTest.java
mbaxter: To be clear - this is a test validating the current behavior (which was previously untested).  I agree this behavior is problematic and should be fixed.  I will push fixes / functional changes in a follow-up PR.

mbaxter:
> I think we need to think through the treatment of any RuntimeException as a bad block, that is a lot broader than known application exceptions like StorageException and MerkleTrieException. If that gets thrown unexpectedly and a protocol-compliant block is marked as bad, there will be a chain halt.

See inline response [here](https://github.com/hyperledger/besu/pull/6622#discussion_r1511822925).

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/BadBlockManagerTest.java
mbaxter: Here is the follow-up: https://github.com/hyperledger/besu/pull/6678

PR #6625, opened 2024-02-28
prevent startup if PoA and snap/checkpoint

### LINKED ISSUE #6385: behave better if BFT and SNAP specified
Since SNAP is not supported with PoA chains currently. Today if you try to use X_SNAP with e.g. QBFT the node looks like it's going to start but never actually works.

(no reason why it wouldn't work but you probably want the entire history which is effectively FAST sync)

~~if it's simple - just fall back to fast sync~~

macfarla:
Fixes #6385 

Docs - can't see anywhere in the docs where we mention snap or checkpoint on the private networks side, but may be worth adding an explicit note that you must use FAST or FULL if using PoA consensus.


matthew1001 (approved):
Thanks @macfarla, looks good to me, will help avoid confusion.

PR #6640, opened 2024-02-28
Snap server rebase

### LINKED ISSUE #3165: Implement snapsync as a server

garyschulte:
## PR description

Initial implementation of snap server using streaming flat database.  Obseletes #5887 

This PR adds snap server functionality behind a feature flag `--Xsnapsync-server-enabled` which enables besu to serve snap/1 protocol fully.

Snap server has a specific set of storage requirements:
* `--data-storage-format=BONSAI`, snap server only works with bonsai storage
* `--Xbonsai-code-using-code-hash-enabled`, snap server nodes need to store code  by code hash in order to serve snap code requests
* `--Xsnapsync-synchronizer-flat-db-healing-enabled`, Snap server nodes must be snap synced with full flat db enabled, (or the node has to have been full-synced to have a fully-flattened db)

Given two of these requrirements (code-by-code-hash and flat-db-healing), most nodes will have to resync in order to serve snap data, with these parameters in the config file:

```
data-storage-format="BONSAI"
Xsnapsync-server-enabled=true
Xbonsai-code-using-code-hash-enabled=true
Xsnapsync-synchronizer-flat-db-healing-enabled=true
```

### Performance Impact
Testing the initial implementation has shown serving snap data has slight impact to execution performance in normal network conditions, on the order of 10% 

Screenshot 2024-03-22 at 5 12 25 PM

### Load Test
To get an idea of the cost to execution performance in less than ideal conditions, besu was tested with the load of 2 static syncing peers (besu serving all of the sync load to both clients simulatenously).  In this case besu showed roughly 2x the block execution times of besu nodes which were not serving snap data.  

Normally besu would only serve some fraction of the load of syncing clients.  We could push this further with more static syncing clients, but this was sufficient to get signal about the performance impact.

Once the snap clients completed their worldstate sync, besu performance returned to the mean immediately

2 client sync perf

Read vs write during this time:

Screenshot 2024-03-22 at 5 06 39 PM

### Testing Clients
Besu 23.4.x
Geth 13.14
Nethermind 1.25.4

### What's Next
* Optimizations.  This is an initial implementation of snap server and now that we are satisfied in its correctness, we will turn to optimizing to minimize its impact on block execution times.

* A follow-on PR will add a storage subcommand to convert existing databases to fully flattened and code-by-code-hash storage, without the need for a resync.

## Fixed Issue(s)

fixes #3165 


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateStorageCoordinator.java
```diff
@@ -62,14 +62,23 @@ public Optional<Bytes> getAccountStorageTrieNode(
...
   public <STRATEGY extends WorldStateKeyValueStorage> STRATEGY getStrategy(
       final Class<STRATEGY> strategyClass) {
     return (STRATEGY) worldStateKeyValueStorage;
   }
 
   public boolean isMatchingFlatMode(final FlatDbMode flatDbMode) {
-    if (getDataStorageFormat().equals(DataStorageFormat.BONSAI)) {
+    Optional<DataStorageFormat> storageFormat =
+        Optional.ofNullable(getDataStorageFormat())
+            .filter(format -> format.equals(DataStorageFormat.BONSAI));
```
matkt, line 79: why we need that ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/EthProtocolConfiguration.java
```diff
@@ -24,7 +24,7 @@
 public class EthProtocolConfiguration {
 
   public static final int DEFAULT_MAX_MESSAGE_SIZE = 10 * ByteUnits.MEGABYTE;
-  public static final int DEFAULT_MAX_GET_BLOCK_HEADERS = 192;
+  public static final int DEFAULT_MAX_GET_BLOCK_HEADERS = 512;
```
matkt, line 27: why we need to increase that ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateStorageCoordinator.java
garyschulte: IIRC there are a lot of test cases that do not fully setup the coordinator, and datastorageformat was coming back nullable.  I will remove this Optional and see which tests fail and fix those.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/EthProtocolConfiguration.java
garyschulte: geth requests headers in blocks of 512.  their rationale is here: https://discord.com/channels/595666850260713488/1014177210560696330/1171097165473521754

If geth cannot get a full set of 512 headers, it will disconnect the peer it is trying to sync from

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/bonsai/cache/CachedWorldStorageManager.java
```diff
@@ -192,14 +203,50 @@ public Optional<BonsaiWorldState> getHeadWorldState(
...
+    for (long i = head; i > Math.max(0, head - numEntries); i--) {
+      blockchain
+          .getBlockHeader(i)
+          .ifPresent(header -> stateRootToBlockHeaderCache.put(header.getStateRoot(), header));
+    }
+  }
+
+  /**
+   * Returns the worldstate for the supplied root hash. If the worldstate is not already in cache,
+   * this method will attempt to fetch it and add it to the cahce. synchronized to prevent
```
macfarla, line 226:
```suggestion
   * this method will attempt to fetch it and add it to the cache. synchronized to prevent
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/StorageFlatDatabaseHealingRangeRequest.java
```diff
@@ -56,7 +57,7 @@ public class StorageFlatDatabaseHealingRangeRequest extends SnapDataRequest {
   private final Bytes32 storageRoot;
   private final Bytes32 startKeyHash;
   private final Bytes32 endKeyHash;
-  private TreeMap<Bytes32, Bytes> slots;
+  private NavigableMap<Bytes32, Bytes> slots;
```
macfarla, line 60: did error prone complain about the TreeMaps?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,616 @@
...
+
+          // at startup, prime the latest worldstates by roothash:
+          cachedStorageManager.primeRootToBlockHashCache(blockchain, PRIME_STATE_ROOT_CACHE_LIMIT);
+
+          var flatDbStrategy =
+              ((BonsaiWorldStateKeyValueStorage)
+                      worldStateStorageCoordinator.worldStateKeyValueStorage())
+                  .getFlatDbStrategy();
+          if (!flatDbStrategy.isCodeByCodeHash()) {
+            LOGGER.warn("SnapServer requires code stored by codehash, but it is not enabled");
```
matkt, line 175: what we are doing in this case ? not sure to understand if we are stopping the snapserver ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,616 @@
...
+        var cachedStorageManagerOpt =
+            bonsaiArchive.map(archive -> archive.getCachedWorldStorageManager());
+
+        if (cachedStorageManagerOpt.isPresent()) {
+          var cachedStorageManager = cachedStorageManagerOpt.get();
+          this.worldStateStorageProvider = cachedStorageManager::getStorageByRootHash;
+
+          // when we start we need to build the cache of latest 128 worldstates
+          // trielogs-to-root-hash:
+          var blockchain = protocolContext.map(ProtocolContext::getBlockchain).orElse(null);
```
matkt, line 165: why we are not also preloading the bonsai worldstate for this headers ?

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -116,28 +124,67 @@ public Optional<NearestKeyValue> getNearestTo(
...
-              .filter(entry -> entry.getValue().isPresent())
-              .map(
-                  bytesEntry ->
-                      Pair.of(bytesEntry.getKey().toArrayUnsafe(), bytesEntry.getValue().get()))
-          // since we are layered, concat a parent stream filtered by our map entries:
-          ,
-          parent.stream(segmentId).filter(e -> !ourLayerState.containsKey(Bytes.of(e.getLeft()))));
-    } finally {
-      lock.unlock();
+    if (ourLayerState == null) {
```
matkt, line 129: is it needed ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/snap/GetStorageRangeMessage.java
```diff
@@ -116,7 +116,7 @@ public StorageRange range(final boolean withRequestId) {
     }
     if (input.nextIsNull()) {
       input.skipNext();
-      range.endKeyHash(Hash.ZERO);
+      range.endKeyHash(Hash.LAST);
```
matkt, line 119: 🥇

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -116,28 +124,67 @@ public Optional<NearestKeyValue> getNearestTo(
   @Override
   public Stream<Pair<byte[], byte[]>> stream(final SegmentIdentifier segmentId) {
     throwIfClosed();
+    var ourLayerState = hashValueStore.computeIfAbsent(segmentId, s -> newSegmentMap());
```
matkt, line 128: why not keeping the 
```java
final Lock lock = rwLock.readLock();
    lock.lock();
```
in this method ?

matkt (commented):
Added some questions but in general I like the implementation 💯

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,616 @@
...
+          resp.getSize(),
+          maxResponseBytes);
+      return resp;
+    } catch (Exception ex) {
+      LOGGER.error("Unexpected exception serving bytecodes request", ex);
+      return EMPTY_BYTE_CODES_MESSAGE;
+    }
+  }
+
+  MessageData constructGetTrieNodesResponse(final MessageData message) {
```
matkt, line 466: I think we need to limite the number of trie nodes to serve by request as Geth. Because it seems to be a possible dos vector

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,616 @@
...
+    LOGGER
+        .atTrace()
+        .setMessage("Receive get bytecodes message for {} hashes")
+        .addArgument(codeHashes.hashes()::size)
+        .log();
+
+    try {
+      List<Bytes> codeBytes = new ArrayDeque<>();
+      for (Bytes32 codeHash : codeHashes.hashes()) {
+        Optional<Bytes> optCode = worldStateStorageCoordinator.getCode(Hash.wrap(codeHash), null);
```
matkt, line 443: In case a peer is asking an empty code hash we can directly return Bytes.EMPTY instead of reading database for nothing

matkt: I also think we can limit the number of code requests because a node could ask a huge number of empty code hash and spaming our node

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/heal/StorageFlatDatabaseHealingRangeRequest.java
garyschulte: NavigableMap is the superclass of both TreeMap and the streaming class I used in the flat streaming impl (IIRC ConcurrentSkipListMap).  I have been slowly changing TreeMap references to NavigableMap where necessary.  The current diff for this PR doesn't show this as a diff but I wanted to address your q 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
garyschulte: for now, snap server will operate as normal with just a warning.  Essentially we will just serve empty code requests if the storage hasn't migrated.  If we think it is necessary we can refuse to start snap server if we are not in code-by-code hash storage mode.  I was thinking that we would address this corner case via the storage migration subcommand PR

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
garyschulte: we can, and I had it that way for a while to pass hive tests.  But actually loading the last 128 worldstates delays startup a bit and feels a little clunky.  Just priming the blockhash to roothash cache seems to be a happy medium

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
garyschulte: This might be a rebase snafu - I am looking closer

garyschulte: Looking back, I was probably thinking about Layered only in the context of frozen states.  I believe that is the case anywhere that we are streaming - snap server and snap client.  AFAIK only snap server would stream a layered worldstate.  

But you are right, it is feasible we could have a concurrent modification while we are streaming and perhaps some future code might not expect this.  Perhaps snap server and snap client could use an unsafeStream() method that does not do locking, and we can make stream() do rw locking.  

I suspect that rw lock will be fairly expensive to the performance of snap server, since we want to stream requests concurrently.  AFAIK the only code path that could potentially stream a layer is snap server.  I am open to suggestions here.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
garyschulte: If we adopt your suggestion about Bytes.EMPTY return for empty code hash, then the response size limit will be decent DoS'd protection.

It seems like p2p rate limiting should be handled at a lower level since we could be dos'd by regular eth messages also 🤔

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
garyschulte: I will 👀 what geth is doing here.  Just to clarify, what about trie nodes queries makes you think they are disproportionately expensive?  I would think range requests would represent more resource consumption, especially considering we are not directly using an iterator but rather collecting in memory to serve the requests.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,619 @@
...
+                  }
+
+                  if (!statefulPredicate.shouldGetMore()) {
+                    break;
+                  }
+                }
+
+                var resp = StorageRangeMessage.create(collectedStorages, proofNodes);
+                LOGGER.debug(
+                    "returned in {} storage {} to {} range {} to {} with {} storages and {} proofs, resp size {} of max {}",
```
macfarla, line 405: nit: this log message is duplicated, could be refactored to a method

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,619 @@
...
+      return EMPTY_BYTE_CODES_MESSAGE;
+    }
+    StopWatch stopWatch = StopWatch.createStarted();
+
+    final GetByteCodesMessage getByteCodesMessage = GetByteCodesMessage.readFrom(message);
+    final GetByteCodesMessage.CodeHashes codeHashes = getByteCodesMessage.codeHashes(true);
+    final int maxResponseBytes = Math.min(codeHashes.responseBytes().intValue(), MAX_RESPONSE_SIZE);
+    LOGGER
+        .atTrace()
+        .setMessage("Receive get bytecodes message for {} hashes")
```
macfarla, line 439:
```suggestion
        .setMessage("Received get bytecodes message for {} hashes")
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,619 @@
...
+      return EMPTY_TRIE_NODES_MESSAGE;
+    }
+    StopWatch stopWatch = StopWatch.createStarted();
+
+    final GetTrieNodesMessage getTrieNodesMessage = GetTrieNodesMessage.readFrom(message);
+    final GetTrieNodesMessage.TrieNodesPaths triePaths = getTrieNodesMessage.paths(true);
+    final int maxResponseBytes = Math.min(triePaths.responseBytes().intValue(), MAX_RESPONSE_SIZE);
+    LOGGER
+        .atTrace()
+        .setMessage("Receive get trie nodes message of size {}")
```
macfarla, line 480:
```suggestion
        .setMessage("Received get trie nodes message of size {}")
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -14,66 +14,619 @@
...
 import kotlin.collections.ArrayDeque;
+import org.apache.commons.lang3.time.StopWatch;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import org.apache.tuweni.units.bigints.UInt256;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+/** See https://github.com/ethereum/devp2p/blob/master/caps/snap.md */
 @SuppressWarnings("unused")
```
macfarla, line 66: can this annotation be removed now

macfarla (approved):
few nits mostly for logging

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
matkt: If the answer is empty we will never reach the max limit. Imagine if I ask you for 3000 Empty Code Hash. How can the size limit be reached?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
matkt: I ask you 30000 trie nodes that does not exist. You will still have to read in base to find out if the trie node exists. So I can spam you as much as I want. The Range is not the same I ask you for a range and the peers will control what it returns in this range according to the number of leaves. To test it is an intuition that I have

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
matkt: If there is no choice for performance. I think UnsafeStream should be good

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
garyschulte: Removed 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
garyschulte: apparently yes 👍

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
garyschulte: I am going to skip this one for now since all of the return messages at least subtly different.

PR #6641, opened 2024-02-28
Feature/fleet mode rebase

garyschulte:
## PR description
rebase onto main of the services refactor for fleet mode, see original #6003

### DIFF
--- a/plugin-api/build.gradle
+++ b/plugin-api/build.gradle
@@ -70,7 +70,7 @@ Calculated : ${currentHash}
 tasks.register('checkAPIChanges', FileStateChecker) {
   description = "Checks that the API for the Plugin-API project does not change without deliberate thought"
   files = sourceSets.main.allJava.files
-  knownHash = 'W3h2jfQ0dVHlmzD9cDnn3S6keqPikCOXGaVSdr2D78Y='
+  knownHash = 'wMhttXj2aWFgpN9msxHz/FwQVovpYRxNysnZEpwZYxg='
 }
 check.dependsOn('checkAPIChanges')
 


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -393,6 +393,7 @@ private void cacheBlockData(final Block block, final List<TransactionReceipt> re
         cache -> cache.put(block.getHash(), block.getHeader().getDifficulty()));
   }
 
+  @SuppressWarnings("unused")
```
matkt, line 396: I think we can remove it

matkt (commented):
did we tried follower+leader with this last commit ?

matkt:
@jframe could you check the block pruning modification we are doing in this PR to be sure it's correct. normally we did tests but I prefer to have your feedback

garyschulte:
> @jframe could you check the block pruning modification we are doing in this PR to be sure it's correct. normally we did tests but I prefer to have your feedback

I inadvertentlty tested this feature on the captain of a captain-follower sepolia pair, and block pruning is definitely working.  

> did we tried follower+leader with this last commit ?
I am resyncing the follower now to ensure it works correctly, since it failed to find the pruned captain blocks on the prior test 😂

garyschulte:
> did we tried follower+leader with this last commit ?

Yes, confirmed the follower syncs and follows with only a leader.  

Regarding the pruner on the follower, I see it configured, but I am not seeing any executions of the pruner triggered.

jframe:
> @jframe could you check the block pruning modification we are doing in this PR to be sure it's correct. normally we did tests but I prefer to have your feedback

Is the changes in the ChainDataPruner to only start pruning after the initial sync? Think it's better not to delay this until after the initial sync to keep the storage used small. Ideally, you would have Besu configured with `--Xcheckpoint-post-merge-enabled` so there isn't anything to prune anyway. Probably need to think more about how these and sync all interact.

besu/VERSION_METADATA.json
```diff
@@ -0,0 +1 @@
+{"besuVersion":"24.2.0-dev-c990209a"}
```
garyschulte, line 1: need to add this file to .gitignore

matkt (approved):
if we had time to do a quick test with the fleet plugin I think we can merge this PR

non-fungible-nelson:
🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢

Issue #6674, opened 2024-03-04
`block export` (or any sub command) starts a BFT mining coordinator


matthew1001:
### Description
When running the `block export` or `block import` commands, a BFT mining coordinator is started. If the node is the only validator is will continue to produce new blocks even during `export`/`import`.

### Acceptance Criteria
Don't start a mining coordinator when running any of the sub-commands

### Steps to Reproduce (Bug)
Run `besu --genesis-file=../genesis.json --data-path=./data blocks export --to=blocks.bin`. If the node is the only validator it will produce new blocks during the export

PR #6675, opened 2024-03-04
Don't start the BFT mining coordinator when it is created, just enable it

matthew1001:
## PR description
When creating a BFT mining coordinator, if the sync phase is already complete then enable the coordinator but don't start it.

Start will be done either during starting of the ethereum main loop, or on receiving an event notification that syncing has completed.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/6674


jframe (commented):
Changes look good, it would be good add a changelog entry as well

jframe (approved):

PR #6596, opened 2024-02-20
Clique block period transition

siladu:
Add [BFT-style transitions](https://besu.hyperledger.org/private-networks/how-to/configure/consensus/qbft#transitions) to Clique, modelled with `ForksSchedule<CliqueConfigOptions>`

CliqueMiningAcceptanceTest heavily inspired by [BftBlockRewardPaymentAcceptanceTest](https://github.com/hyperledger/besu/blob/main/acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/bft/BftBlockRewardPaymentAcceptanceTest.java)

Preparatory story to https://github.com/hyperledger/besu/issues/6290 - a `createemptyblocks` transition will be added in the next PR.

Create genesis config transitions:
```
  "config": {
    ...
    "clique": {
      "blockperiodseconds": 3,
      "epochlength": 30,
      "requesttimeoutseconds": 6,
      "createemptyblocks": true
    },
    "transitions": {
      "clique": [
        {
          "block": 3,
          "blockperiodseconds": 1
        },
        {
          "block": 6,
          "blockperiodseconds": 2
        },
      ]
    }
  },
```

PR #6676, opened 2024-03-04
Make block txs selection max time aware of PoA transitions

fab-10:

- [ ] locally run all unit tests via: `./gradlew build`

## PR description

As a follow up of #6596 this PR makes the block txs selection max time, correctly  aware of the PoA transitions, while before only the block period from the genesis was used, while that value can change according to the transitions defined in the genesis file. To support that the `genesisBlockPeriodSeconds` has been renamed to `blockPeriodSeconds` and the configuration is now modifiable at runtime, according to the transitions schedule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Transaction call object to accept both `input` and `data` field simultaneously if they are set to equal values [#6702](https://github.com/hyperledger/besu/pull/6702)
 
 ### Bug fixes
+- Make block transaction selection max time aware of PoA transictions [#6676](https://github.com/hyperledger/besu/pull/6676)
 
 ### Download Links
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 - Transaction call object to accept both `input` and `data` field simultaneously if they are set to equal values [#6702](https://github.com/hyperledger/besu/pull/6702)
 
 ### Bug fixes
-- Make block transaction selection max time aware of PoA transictions [#6676](https://github.com/hyperledger/besu/pull/6676)
+- Make block transaction selection max time aware of PoA transitions [#6676](https://github.com/hyperledger/besu/pull/6676)
 
 ### Download Links
 
--- a/consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/blockcreation/CliqueBlockScheduler.java
+++ b/consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/blockcreation/CliqueBlockScheduler.java
@@ -20,6 +20,7 @@ import org.hyperledger.besu.consensus.common.validator.ValidatorProvider;
 import org.hyperledger.besu.datatypes.Address;
 import org.hyperledger.besu.ethereum.blockcreation.DefaultBlockScheduler;
 import org.hyperledger.besu.ethereum.core.BlockHeader;
+import org.hyperledger.besu.ethereum.core.MiningParameters;
 
 import java.time.Clock;
 import java.util.Collection;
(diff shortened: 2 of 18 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Update commons-compress to 1.26.0 [#6648](https://github.com/hyperledger/besu/pull/6648)
 
 ### Bug fixes
+- Make block transaction selection max time aware of PoA transictions [#6676](https://github.com/hyperledger/besu/pull/6676)
```
siladu, line 22:
```suggestion
- Make block transaction selection max time aware of PoA transitions [#6676](https://github.com/hyperledger/besu/pull/6676)
```

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/blockcreation/CliqueBlockScheduler.java
```diff
@@ -43,19 +44,24 @@ public class CliqueBlockScheduler extends DefaultBlockScheduler {
...
-                    .getFork(parentHeader.getNumber() + 1)
-                    .getValue()
-                    .getBlockPeriodSeconds(),
+        parentHeader -> {
+          final var blockPeriodSeconds =
+              forksSchedule
+                  .getFork(parentHeader.getNumber() + 1)
+                  .getValue()
+                  .getBlockPeriodSeconds();
+          miningParameters.setBlockPeriodSeconds(blockPeriodSeconds);
```
siladu, line 62: Feels slightly odd to pass this into the constructor just to have this value set. 

I think I can see why you've done it though - the alternative of querying the forksSchedule at the point where we read from the miningParameters could be quite messy.

siladu (approved):
Wiring runtime values into MiningParameters is making me think twice, but seems there is a precedent for it already. 

Also the way the block period is set on MiningParameters but I don't feel too strongly about it.

Other than that, PR looks good :)

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/blockcreation/CliqueBlockScheduler.java
jframe: Does seem like a bit hack putting it here to keep the MinerParameters synced with the current blockPeriodSeconds.

Rather than adding it here, I think we could add a block added listener in the CliqueBesuControllerBuilder to update the block period on the MiningParameters when we move to a new block height.

i.e. something like this in CliqueBesuControllerBuilder.createMiningCoordinator
```
protocolContext.getBlockchain().observeBlockAdded(o -> miningParameters.setBlockPeriodSeconds(forksSchedule.getFork(o.getBlock().getHeader().getNumber()).getValue().getBlockPeriodSeconds()));
```

fab-10: I was not happy with the solution too, so I have applied your suggestion, tnx

CHANGELOG.md

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/BftBlockCreator.java
```diff
@@ -83,6 +84,14 @@ private static MiningBeneficiaryCalculator miningBeneficiaryCalculator(
         forksSchedule.getFork(blockNum).getValue().getMiningBeneficiary().orElse(localAddress);
   }
 
+  private void setBlockPeriodSeconds(
```
jframe, line 87: Moving this into the controller same way you did with Clique would be nicer

fab-10: done, please re-review

jframe (approved):

PR #6678, opened 2024-03-04
Bugfix - only track intrinsically bad blocks

mbaxter:
## PR description
This is a follow-up to PR https://github.com/hyperledger/besu/pull/6622.  Rework bad block handling so that we are only tracking blocks that are intrinsically bad due to a spec-related validation error.  We should not be marking blocks as bad if we are missing data locally or encounter some unspecified exception during block processing.  

## Fixed Issue(s)
Tangentially related to https://github.com/hyperledger/besu/issues/6301


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -157,53 +160,50 @@ public BlockProcessingResult validateAndProcessBlock(
...
+        LOG.info("Invalid block {}", failedBlock.toLogString());
+      }
+
+      if (shouldRecordBadBlock) {
+        // Result.errorMessage should not be empty on failure, but add a default to be safe
+        String description = result.errorMessage.orElse("Unknown cause");
+        final BadBlockCause cause = BadBlockCause.fromValidationFailure(description);
+        badBlockManager.addBadBlock(failedBlock, cause);
+      } else {
+        LOG.debug("Invalid block {} not added to badBlockManager ", failedBlock.toLogString());
```
macfarla, line 205:
```suggestion
        LOG.debug("Failed block {} not added to badBlockManager ", failedBlock.toLogString());
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockCause.java
```diff
@@ -64,20 +53,8 @@ public String getDescription() {
...
-    return "BadBlockCause{"
-        + "reason="
-        + reason
-        + ", description='"
-        + description
-        + '\''
-        + ", exception="
-        + exception
-        + '}';
+    return "BadBlockCause{" + "reason=" + reason + ", description='" + description + '\'' + '}';
```
macfarla, line 58:
```suggestion
    return "BadBlockCause{reason=" + reason + ", description='" + description + "\}";
```

mbaxter: Yeah - this method is kind of inconsistent and not very readable.  Regenerated using `MoreObjects`.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
mbaxter: The block here actually is invalid (failed validation), so I think its useful to label it that way.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockCause.java
```diff
@@ -31,29 +31,20 @@ public enum BadBlockReason {
...
   }
 
   public static BadBlockCause fromValidationFailure(final String failureMessage) {
-    return new BadBlockCause(
-        BadBlockReason.SPEC_VALIDATION_FAILURE, failureMessage, Optional.empty());
+    return new BadBlockCause(BadBlockReason.SPEC_VALIDATION_FAILURE, failureMessage);
   }
 
-  private BadBlockCause(BadBlockReason reason, String description, Optional<Throwable> exception) {
+  private BadBlockCause(final BadBlockReason reason, final String description) {
```
mbaxter, line 45: I'm not sure why the missing `final`'s here only caused a compilation error / failed unit tests ([here](https://github.com/hyperledger/besu/actions/runs/8159257090/job/22303389097?pr=6678) and [here](https://github.com/hyperledger/besu/actions/runs/8159257090/job/22303387743?pr=6678)) when I pushed the toString change?

jflo (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockCause.java
mbaxter: Added an issue to investigate [here](https://github.com/hyperledger/besu/issues/6685).

PR #6698, opened 2024-03-07
Support pruned chain history in peer validators

garyschulte:
## PR description
Change the behavior of peer block validator to allow block absence to be tolerated for some rules, specifically DaoForkPeerValidator, in support of:
* https://eips.ethereum.org/EIPS/eip-4444
* https://github.com/ethereum/EIPs/pull/8266
* and besu checkpoint sync'd nodes


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/peervalidation/AbstractPeerBlockValidator.java
```diff
@@ -86,12 +86,20 @@ public CompletableFuture<Boolean> validatePeer(
...
+                if (blockIsRequired()) {
+                  // If no headers are returned, fail
+                  LOG.debug(
+                      "Peer {} is invalid because required block ({}) is unavailable.",
+                      ethPeer,
+                      blockNumber);
+                  return false;
+                } else {
+                  LOG.debug(
+                      "Peer {} deemed valid despite block ({}) being unavailable.",
```
macfarla, line 98:
```suggestion
                      "Peer {} deemed valid because unavailable block ({}) is not required.",
```

macfarla (approved):
minor comment on log message. worth a change log entry?

PR #6699, opened 2024-03-07
proposed "develop" build versions

garyschulte:
## PR description
As part of our recent build and release changes, we no longer need to explicitly maintain the build version in gradle.properties file.

This PR is a proposal to change our "develop" builds to use git commit and HEAD details to specify a dynamic build version.

for example this PR, prior to committing the changes, produced a build named:
`besu-24.3.7-develop-610927511e.tar.gz`

After committing the gradle changes and rebuilding, it produced a build named:
`besu-24.3.7-develop-600d812d89.tar.gz`

Open to suggestions, feedback, and identifying corner cases

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,7 +119,7 @@ allprojects {
   apply plugin: 'net.ltgt.errorprone'
   apply from: "${rootDir}/gradle/versions.gradle"
 
-  version = rootProject.version
+  version = calculateVersion()
 
   jacoco {
     toolVersion = '0.8.8'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -1038,42 +1038,40 @@ def buildTime() {
...
-// with the git commit version.
 @Memoized
 def calculateVersion() {
-  String version = rootProject.version
-  if (version.endsWith("-SNAPSHOT")) {
-    version = version.replace("-SNAPSHOT", "-dev-" + getCheckedOutGitCommitHash())
-  }
-  return version
+  def gitDetails = getGitCommitDetails(10) // Adjust length as needed
+  return "${gitDetails.date}-develop-${gitDetails.hash}"
```
garyschulte, line 1044: @jflo thinking about the impact of this on docker builds.  we would probably have a build explosion if we published versions like this on dockerhub.  Does the new CI pipeline explicitly push versions into snapshot builds on `main` ?  or is that something we need to account for here?

garyschulte: I disabled the duplicate circle-ci workflows in this PR that are causing us to publish docker images to dockerhub using snapshot build tags

garyschulte: reverted the change in favor of just disabling the circle-ci webhook

jflo (changes requested):
Love it, but we do need to still honor `-Pversion` when it is used by the release pipeline.

build.gradle
```diff
@@ -119,7 +119,7 @@ allprojects {
   apply plugin: 'net.ltgt.errorprone'
   apply from: "${rootDir}/gradle/versions.gradle"
 
-  version = rootProject.version
```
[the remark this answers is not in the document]
garyschulte: added a provision for build version to be supplied 👍  .  I also force it to be semver or semver-something

jflo (approved):

build.gradle
jflo: you can still have build artifacts, just not docker images:  https://github.com/hyperledger/besu/releases/tag/develop

PR #6702, opened 2024-03-08
Transaction call object to accept both input and data field if equal

### LINKED ISSUE #6697: RPC call parameters do not accept input and data field simultaneously
Hi!

https://github.com/hyperledger/besu/pull/6094 introduced the option to provide the calldata via the `input` field instead of the deprecated `data` field. To remain compatible with chains that have not yet implemented the updated spec, the Chainlink node will set both fields, which is working fine with all other chains so far but fails on Besu due to the strict check in the aforementioned PR:
```
    if (input != null && data != null) {
      throw new IllegalArgumentException("Only one of 'input' or 'data' should be provided");
    }
```

Is there a specific resons why this check is this strict, or could it be relaxed to the following:
```
    if (input != null && data != null && !input.equals(data)) {
      throw new IllegalArgumentException("Only one of 'input' or 'data' should be provided");
    }
```

friedemannf:

- [X] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?

- [X] locally run all unit tests via: `./gradlew build`
- [X] locally run all integration tests via: `./gradlew integrationTest`

## PR description
https://github.com/hyperledger/besu/pull/6094 introduced the option to provide the calldata via the input field instead of the deprecated data field. It also introduced a check that prevents both field from being set. This seems overly strict as it prevents clients from setting both fields to remain backward-compatible. This PR relaxes the check to only fail if both fields are set and they are not equal.

## Fixed Issue(s)
fixes #6697


friedemannf:
We might want to adjust the documentation here to specify this behavior: 
https://besu.hyperledger.org/public-networks/reference/api/objects#transaction-call-object

macfarla (approved):
LGTM. test cases make it clear. Thanks for your contribution @friedemannf

macfarla:
> We might want to adjust the documentation here to specify this behavior: https://besu.hyperledger.org/public-networks/reference/api/objects#transaction-call-object

Thanks for this - added `doc-change-required` label for that purpose.

Issue #6685, opened 2024-03-05
Investigate CI incorrectly passing code with compilation errors


mbaxter:
### Description
For context, see the PR comment [here](https://github.com/hyperledger/besu/pull/6678#discussion_r1513151554).  In summary, CI has been passing, a change is pushed up, and it starts failing on some unrelated code.  The failing code was pushed in a previous PR, and there is a sequence of passing CI checks before it finally fails here due to the [MethodInputParametersMustBeFinal](https://github.com/hyperledger/besu/blob/dbc128f1002d931344c6a94d6d82253f664d92e8/errorprone-checks/src/main/java/org/hyperledger/errorpronechecks/MethodInputParametersMustBeFinal.java#L39-L39) errorprone check.

I tried running the failing gradle task locally at the git commit where it finally failed in CI.  When running locally, the task passed without failure.  It looks like the `-WError` flag should always be applied (see [build.gradle](https://github.com/hyperledger/besu/blob/2b8d44ec77f19c3a0b06406d75c16f073d1eea86/build.gradle#L225-L225)) so not sure why this is passing locally.

We should investigate this CI flakiness and make sure that we're actually enforcing all of the rules in CI consistently.

non-fungible-nelson:
@macfarla for visibility

usmansaleem:
@non-fungible-nelson I've picked up this issue. This resulted in taking out Besu's custom errorprone into its own repo. Also, the class involved, looks like our errorprone check is not always able to pick it, so I've make a small refactoring in it. Will create a PR to cover this and other bits that are uncovered due to upgrading the errorprone. See https://github.com/hyperledger/besu/pull/6692 as well.

PR #6716, opened 2024-03-12
Refactor BadBlockReason enum out of BadBlockCause

usmansaleem:

## PR description
BadBlockReason was an non-static inner enum in BadBlockCause and it was used in static context. This caused one of our custom errorprone checks to not pick up the "method param not final argument" correctly. Refactoring it out as an static top level class. Now all the checks are applied correctly.

## Fixed Issue(s)

#6685

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockReason.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockReason.java
@@ -17,8 +17,6 @@ package org.hyperledger.besu.ethereum.chain;
 public enum BadBlockReason {
   // Standard spec-related validation failures
   SPEC_VALIDATION_FAILURE,
-  // When an unexpected exception occurs during block processing
-  EXCEPTIONAL_BLOCK_PROCESSING,
   // This block is bad because it descends from a bad block
   DESCENDS_FROM_BAD_BLOCK,
 }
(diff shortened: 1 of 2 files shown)


macfarla (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockReason.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.chain;
+
+public enum BadBlockReason {
+  // Standard spec-related validation failures
+  SPEC_VALIDATION_FAILURE,
+  // When an unexpected exception occurs during block processing
+  EXCEPTIONAL_BLOCK_PROCESSING,
```
mbaxter, line 21: (optional) While you're here - this value is not used anywhere and can be removed.  If you want to leave it for now, I can clean this up in another PR.

mbaxter:
Does the custom errorprone check need to be updated to avoid this in the future?  Or do we need another rule that enforces static inner classes?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/BadBlockReason.java
usmansaleem: done

usmansaleem:
@mbaxter I am extracting out Besu errorprone in its own repo. https://github.com/hyperledger/besu-errorprone-checks/ . I tried to reproduce this error via test case, however, I wasn't able to reproduce it: https://github.com/hyperledger/besu-errorprone-checks/blob/main/src/test/resources/org/hyperledger/errorpronechecks/ConstructorParameterMustBeFinalPositiveCases.java

I suspect in addition of slightly buggy custom errorprone check, we have some other complexity involved here as well (perhaps dagger compiler?).

PR #6727, opened 2024-03-13
Suppress ComparisonOutOfRange errorprone warning in AbstractGasLimitSpecification

usmansaleem:
## PR description
Suppress ComparisonOutOfRange error-prone warning in AbstractGasLimitSpecification. This was reported during errorprone plugin update. Add a unit test to cover the max limit as well.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractGasLimitSpecification.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractGasLimitSpecification.java
@@ -47,7 +47,6 @@ public abstract class AbstractGasLimitSpecification {
    * @return true if within bounds
    */
   public static boolean isValidTargetGasLimit(final long targetGasLimit) {
-    // only check the lower bound as the upper bound is Long.MAX_VALUE
-    return targetGasLimit >= DEFAULT_MIN_GAS_LIMIT;
+    return DEFAULT_MIN_GAS_LIMIT <= targetGasLimit && DEFAULT_MAX_GAS_LIMIT >= targetGasLimit;
   }
 }
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractGasLimitSpecification.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractGasLimitSpecification.java
@@ -46,6 +46,7 @@ public abstract class AbstractGasLimitSpecification {
    * @param targetGasLimit the target gas limit to validate
    * @return true if within bounds
    */
+  @SuppressWarnings("ComparisonOutOfRange")
   public static boolean isValidTargetGasLimit(final long targetGasLimit) {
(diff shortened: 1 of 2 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractGasLimitSpecification.java
```diff
@@ -40,7 +40,14 @@ public static long deltaBound(final long currentGasLimit) {
...
+  /**
+   * Verify that the target gas limit is within the allowed bounds.
+   *
+   * @param targetGasLimit the target gas limit to validate
+   * @return true if within bounds
+   */
   public static boolean isValidTargetGasLimit(final long targetGasLimit) {
-    return DEFAULT_MIN_GAS_LIMIT <= targetGasLimit && DEFAULT_MAX_GAS_LIMIT >= targetGasLimit;
+    // only check the lower bound as the upper bound is Long.MAX_VALUE
+    return targetGasLimit >= DEFAULT_MIN_GAS_LIMIT;
```
siladu, line 51: Can you point to which tests this is covered by please?

usmansaleem: ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/TargetingGasLimitCalculatorTest.java

siladu: Thanks. I find this errorprone check odd - wouldn't this code break if we error changed the DEFAULT_MAX to be something other than Long.MAX_VALUE?

macfarla: agree it seems safer to suppress this warning

usmansaleem: To counter-argue, if we jump from long to BigInteger for example, then the targetGasLimit would also change from long to BigInteger. What is the value `DEFAULT_GAS_MAX_LIMIT` as per the spec? Here it seems to be bounded by the variable type we are using i.e. Long.MAX_VALUE.

I'll close this PR and will add the suppression when we bring in the updated errorprone library dependency.

siladu: Can't find a max value in the spec, was looking here: https://ethereum.github.io/execution-specs/src/ethereum/shanghai/fork.py.html#ethereum.shanghai.fork.check_gas_limit:0

It's UInt so specified as arbitrary size https://ethereum.github.io/execution-specs/src/ethereum/base_types.py.html

>  if we ever changed the DEFAULT_MAX to be something other than Long.MAX_VALUE

I was more thinking if a future spec introduced a specific max value to enforce something in the protocol.

usmansaleem: Agreed. I'll add a unit test instead to cover the max value.

macfarla (approved):

siladu (approved):
👍

PR #6734, opened 2024-03-14
Change eth trace from 'Message not expected' to 'Request message'

matthew1001:
## PR description
Change the trace entry for messages that aren't `response`-type messages.

## Fixed Issue(s)
N/A


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeer.java
```diff
@@ -425,7 +425,7 @@ Optional<RequestManager> dispatch(final EthMessage ethMessage, final String prot
         localRequestManager -> localRequestManager.dispatchResponse(ethMessage),
         () -> {
           LOG.trace(
-              "Message {} not expected has just been received for protocol {}, peer {} ",
+              "Request message has just been received for protocol {}, peer {} ",
```
macfarla, line 428: you're missing a set of {}

matthew1001: Thanks @macfarla, just fixed and moved the PR out of draft.

Issue #6749, opened 2024-03-18
Backward sync stuck in a loop


pinges:
### Description
The backward sync (BWS) is getting stuck in a loop with these recurring log messages:
```
{"@timestamp":"2024-03-05T13:57:59,199","level":"INFO","thread":"EthScheduler-Timer-0","class":"BackwardSyncContext","message":"Current backward sync session failed, it will be restarted","throwable":""}
{"@timestamp":"2024-03-05T13:58:01,114","level":"INFO","thread":"vert.x-worker-thread-0","class":"BackwardSyncContext","message":"Starting a new backward sync session","throwable":""}
```
Enough peers are present.

Restarting the node fixes the problem.

### Reason
We receive an fcu containing the block hash of a head block. This block is added to the hashesToAppend queue. The block get's reorged and when we try to retrieve that block in the BWS from out peers none of them is able to provide it to us. This causes the BWS to fail, and when we receive the next fcu, a new hash might be added to the queue, but a new BWS will be started, trying to retrieve the same block that we have unsuccessfully tried to retrieve before.

This happened on 7 out of 8 nodes I started based on 24.2.0-RC4:
dev-elc-besu-teku-mainnet-dev-stefan-rc4-(1,2,4)
dev-elc-besu-teku-mainnet-dev-stefan-ss-(1,2,3,4)

The block reorged had the hash 0x4550b82492bf1738af79efb6140770c5443d368b9512ae8551583909554a040f.

Link to Kibane should work for about another 3 weeks:
[...]

PR #6756, opened 2024-03-19
Fix "Backward sync stuck in a loop" #6749

pinges:
## PR description
This PR is fixing issue #6749, trying to make minimal changes to the existing implementation.


ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContextTest.java
```diff
@@ -438,4 +438,31 @@ public void shouldFailAfterMaxNumberOfRetries() {
...
+        assertThat(backwardSyncException.getMessage())
+            .contains("Max number of retries " + NUM_OF_RETRIES + " reached");
+      }
+    }
+
+    // represents subsequent FCU with reorged version of the same block
+    final CompletableFuture<Void> secondFuture =
+        context.syncBackwardsUntil(getBlockByNumber(REMOTE_HEIGHT - 1).getHash());
+    respondUntilFutureIsDone(secondFuture);
+    secondFuture.get();
```
macfarla, line 466: is there any useful assertion to make at this point? or just that it didn't error out

gradle.properties
```diff
@@ -14,3 +14,4 @@ org.gradle.jvmargs=-Xmx4g \
 --add-opens jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
 # Could be moved to sonar properties after https://sonarsource.atlassian.net/browse/SONARGRADL-134
 systemProp.sonar.gradle.skipCompile=true
+version=24.3.1-SNAPSHOT
```
garyschulte, line 17: we removed version from gradle.properties intentionally.  Is this an errant add?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncAlgorithm.java
```diff
@@ -61,7 +63,28 @@ public CompletableFuture<Void> executeBackwardsSync(final Void unused) {
...
-      return executeSyncStep(firstHash.get())
+      final CompletableFuture<Void> completableFuture = new CompletableFuture<>();
+      executeSyncStep(firstHash.get())
+          .whenComplete(
+              (result, error) -> {
+                if (error != null) {
+                  if (error instanceof CompletionException
+                      && error.getCause() instanceof MaxRetriesReachedException) { // &&
+                    // context.getEthContext().getEthPeers().peerCount() >
+                    // (context.getEthContext().getEthPeers().getMaxPeers() / 2)
```
garyschulte, line 74: it probably is wise to at least assert we have peers before we drop hashes, but making a qualitative judgement about the peers, when all we have to go by is a hash, is not straightforward.  

IMO dropping the hash is straightforward and not catastrophic, since we will get a subsequent fcU.  As long as we do not give up either:
* too easily 
* in the absence of peers


I think a simple peer count check with a task retry backoff will suffice :+1:

garyschulte (changes requested):
Good 'light touch' approach.  good to merge once those two minor issues are resolved

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContextTest.java
siladu: it's still in dev, will tidy up this test now I know it works for the impl

gradle.properties
pinges: I had to add it to be able to use the Jenkins build from source. Sally found that problem in the scripts and I think it should get fixed soon. Once that is done I can get rid of it.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncAlgorithm.java
pinges: We do know that we have at least one peer, otherwise we would not get the max retries exception.

siladu:
TODO: probably needs another test for the case where there are two entries in the queue at the point where the head is removed.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContextTest.java
siladu: Done

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContextTest.java
```diff
@@ -83,7 +83,7 @@ public class BackwardSyncContextTest {
   public static final int REMOTE_HEIGHT = 50;
   public static final int UNCLE_HEIGHT = 25 - 3;
 
-  public static final int NUM_OF_RETRIES = 100;
+  public static final int NUM_OF_RETRIES = 1;
```
matkt, line 86: only one retry ? in case of bad peers it will not be a problem? not sure how backwardsync is working so just asking

siladu: This is test code, prod code has 2 retries (which is actually just one retry: two "tries")

gradle.properties
pinges: removed

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncAlgorithm.java
pinges: Even if we do remove a hash from the list that is on the chain, the BWS still works.

garyschulte (approved):
LGTM, non-blocking feedback

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncContextTest.java
```diff
@@ -438,4 +441,53 @@ public void shouldFailAfterMaxNumberOfRetries() {
       }
     }
   }
+
+  @SuppressWarnings("BannedMethod")
```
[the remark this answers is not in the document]
pinges: Had to add the version back in so I can use the Jankins build from source. Will fix it :-)

Issue #6760, opened 2024-03-19
QBFT/Shanghai blocks failing withdrawal validation on import


matthew1001:
### Description
Running 4 QBFT validators with `shanghaiTime: 0` the nodes mine blocks correctly and accept transactions correctly.

However, when a new validator joins and syncs with the chain it fails to validate the imported blocks because the block body `withdrawals` element is `null` (represented as `Optional.empty()` in the code)

In the QBFT/Shanghai PR (https://github.com/hyperledger/besu/pull/6353) I disabled `withdrawal` validation for all BFT protocol schedules (see https://github.com/matthew1001/besu/blob/97c50417787f21bca8acf0a8b7d4e74d65f9f19e/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilder.java#L120) since withdrawals aren't a thing for BFT chains.

In hindsight I think the correct way to treat them would have been to always include an empty withdrawals list, rather than no list at all. This would pass `withdrawal` validation by `shanghai` and onwards protocol schedule validators.

Removing the BFT protocol schedule override:

```
.withdrawalsValidator(
            new WithdrawalsValidator
                .ProhibitedWithdrawals()) // QBFT/IBFT doesn't support withdrawals
```

means that BFT nodes will use whichever withdrawal validator is used by the given protocol schedule.

### Versions (Add all that apply)
* Software version: `24.3.0`

matthew1001:
Removing the withdrawal validator override does indeed resolve the import issue, but I see an interesting QBFT issue if QBFT validators set an empty `withdrawals` list in the blocks they propose. The recipients of the proposed blocks seem to receive a block containing a `null` withdrawals field, instead of the empty list set by the proposer. I've drilled all the way down to the RLP encoding layer and back up, and can't identify where the `withdrawals` field is being nulled out during the multicast to other nodes. It's possible the RLP decode is removing it, rather than the encode not including it (scratches head)

matthew1001:
It turns out there are two code BFT code-paths to create a new block and they both needed updating to create a correct block with withdrawals enabled.

PR #6765, opened 2024-03-19
Ensure empty withdrawal lists are set in BFT blocks when the protocol spec is shanghai or higher

matthew1001:
## PR description
If the protocol spec is as `shanghai` or higher, a valid (if empty) withdrawals list must be present in the blocks proposed.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/6760


consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/statemachine/IbftRound.java
```diff
@@ -118,7 +121,20 @@ public ConsensusRoundIdentifier getRoundIdentifier() {
    * @param headerTimeStampSeconds the header time stamp seconds
    */
   public void createAndSendProposalMessage(final long headerTimeStampSeconds) {
-    final Block block = blockCreator.createBlock(headerTimeStampSeconds).getBlock();
+
+    // Determine if we're at a protocol spec that requires withdrawals (even if empty) to be present
+    ProtocolSpec protocolSpec =
```
shemnon, line 126: This is duplicated several times.  It feels like we can do this in BftBlockCreator by overriding createBlock, checking the withdrawal validator, and switching on that.  If the sequence number is truly needed then making a new createBlock method with both args.  May need to also do it for PkuQbftBlockCreator since there is no shared parent classes.

matthew1001: Good thought. I've done some refactoring as `BftBlockCreator` already has access to the parent header required to get the next block number. So it was a little easier than I thought. Now `QbftRound` and `IbftRound` can just continue to call `createBlock(...)` as they did before, and `BftBlockCreator.createBlock(...)` can decide which type of block to actually create. There's a little repetition in `PkiQbftBlockCreator` like you say, but it's definitely tidier now.

shemnon (approved):

cryptohazard:
hello,
could you confirm that this update only work for network that will now go through a Shanghai update and not for previous networks that already activated Shanghai with an older version of Besu.
I'm having an issue related to that https://github.com/hyperledger/besu/issues/7036#issuecomment-2113635177

shubham-edx:
+following

PR #6775, opened 2024-03-21
Refactor BackwardsSyncAlgorithm.pickNextStep for readability

siladu:
Simple series of Extract Method operations.

Tidy up couple of things from #6749

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncAlgorithm.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncAlgorithm.java
@@ -120,27 +120,32 @@ public class BackwardSyncAlgorithm implements BesuEvents.InitialSyncCompletionLi
 
   private void handleSyncStepSuccess(
       final Block result, final Hash firstHash, final CompletableFuture<Void> syncStep) {
-    LOG.atDebug()
-        .setMessage("Backward sync target block is {}")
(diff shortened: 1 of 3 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncAlgorithm.java
```diff
@@ -137,6 +104,54 @@ public CompletableFuture<Void> pickNextStep() {
...
+              } else {
+                handleSyncStepSuccess(result, firstHash, syncStep);
+              }
+            });
+    return syncStep;
+  }
+
+  private void handleSyncStepSuccess(
+      final Block result, final Hash firstHash, final CompletableFuture<Void> syncStep) {
+    LOG.atDebug()
```
pinges, line 123:
```suggestion
    if (result == null) {
      syncStep.completeExceptionally(new RuntimeException("Unexpected null result in handleSyncStep"));
    }
    LOG.atDebug()
```

pinges (approved):
I think we do not need the null check for the result that I have suggested, but might be a good idea.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/backwardsync/BackwardSyncAlgorithm.java
siladu: Good spot, done

PR #6808, opened 2024-03-26
Set noisy TransactionLogBloomCacher debug log to trace

siladu:
Example log:
> Caching logs bloom for block 0x127804

During sync with DEBUG enabled, this is extremely noisy...printed nearly 600K times during a holesky sync in < 1 hour...
Screenshot 2024-03-26 at 4 09 35 PM


jframe (approved):
LGTM

PR #6811, opened 2024-03-26
Update Web3j dependencies

fab-10:


settings.gradle
```diff
@@ -17,6 +17,10 @@
 pluginManagement {
   repositories {
     gradlePluginPortal()
+    maven {
```
shemnon, line 20: Is this needed?  There are no plugins from the `tech.pegasys` namespace.

fab-10: it is needed because new web3j plugin has `tech.pegasys:jc-kzg-4844` as dependency

shemnon: That's not good.  Web3j cannot be loaded entirely from maven central?  Consensys should invest time in getting these published to Maven Central.

fab-10: I know someone is already working on that, so this will not be required in a future version, in the meantime what you prefer, using version `4.10.0` of the plugin or temporarily keeping the extra repo here?

shemnon: We do have the repository in our general listings.  I was hoping we could keep the plugins section minimal.  So this is acceptable since a long term solution is progressing.

PR #6830, opened 2024-03-27
Logging improvements for Bonsai historical queries 

non-fungible-nelson:
## PR description
Makes the logs more clear about historical queries in Bonsai when you exceed the `bonsai-historical-block-limit`

## Fixed Issue(s)
None


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/DiffBasedWorldStateProvider.java
```diff
@@ -132,7 +132,7 @@ public Optional<MutableWorldState> getMutable(
       if (chainHeadBlockHeader.getNumber() - blockHeader.getNumber()
           >= trieLogManager.getMaxLayersToLoad()) {
         LOG.warn(
-            "Exceeded the limit of back layers that can be loaded ({})",
+            "Exceeded the limit of historical blocks that can be loaded ({}). If you need to make older historical queries, configure your `--bonsai-historical-block-limit`.",
```
ahamlat, line 135: I was wondering how we can add a warning that Bonsai isn't designed to deserve historical data. Si if the user configure--bonsai-historical-block-limit to 100000 for example, it will cause performance issues and it could even crash the node

ahamlat (approved):
Just one small comment

PR #6833, opened 2024-03-28
Retry to get peers best block head

pinges:
## PR description
In the ChainHeadTracker we are trying to retrieve the "best" block from a peer that we have added to our EthPeers.
Originally we were disconnecting the peer if the request for the header was empty or an error was reported.
This PR adds 4 retries for getting the header before diconnection.


macfarla:
closing as we're going with a different approach

PR #6847, opened 2024-03-29
Snap client fixes

garyschulte:
## Fixed Issue(s)

snap client fixes discovered during besu snap-client -> besu snap-server testing.
* correctly handle empty range responses
* skip stack trie commits without keys
* adjust remote peer connection limit arithmetic to cover cases where max-peers is 1

rip-coderpal
RIP coder buddy

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
@@ -140,8 +140,8 @@ public class RequestDataStep {
                 /*
                  * Checks if the response represents an "empty range".
                  *
-                 * An "empty range" is defined as a response where either no slots are present,
-                 * or the first slot is empty, but at least one proof exists
+                 * An "empty range" is defined as a response where at least one proof exists
+                 * and either no slots are present, or the first slot is empty
                  */
                 try {
                   final boolean isEmptyRange =
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
@@ -41,6 +41,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableMap;
 import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
@@ -133,7 +134,7 @@ public class RequestDataStep {
             (response, error) -> {
               if (response != null) {
                 downloadState.removeOutstandingTask(getStorageRangeTask);
-                final ArrayDeque<TreeMap<Bytes32, Bytes>> slots = new ArrayDeque<>();
+                final ArrayDeque<NavigableMap<Bytes32, Bytes>> slots = new ArrayDeque<>();
                 // Check if we have an empty range
 
                 /*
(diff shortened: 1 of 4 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
```diff
@@ -131,15 +134,36 @@ public CompletableFuture<List<Task<SnapDataRequest>>> requestStorage(
...
+                /*
+                 * Checks if the response represents an "empty range".
+                 *
+                 * An "empty range" is defined as a response where either no slots are present,
+                 * or the first slot is empty, but at least one proof exists
+                 */
+                try {
+                  final boolean isEmptyRange =
+                      (response.slots().isEmpty() || response.slots().get(0).isEmpty())
+                          && !response.proofs().isEmpty();
```
macfarla, line 149: should this be ((A || B) && !C)
or
(A || (B & !C))

does it make sense for the response.slots to be empty if response.proofs is not empty

garyschulte: yes, in that case it would be a proof of exclusion 👍

garyschulte: || takes precedent over &&, so this should be ((A ||B) && !C).  🙏

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
```diff
@@ -131,15 +134,36 @@ public CompletableFuture<List<Task<SnapDataRequest>>> requestStorage(
...
-                      response.slots().get(i),
-                      i < response.slots().size() - 1 ? new ArrayDeque<>() : response.proofs());
+                final ArrayDeque<NavigableMap<Bytes32, Bytes>> slots = new ArrayDeque<>();
+                // Check if we have an empty range
+
+                /*
+                 * Checks if the response represents an "empty range".
+                 *
+                 * An "empty range" is defined as a response where either no slots are present,
+                 * or the first slot is empty, but at least one proof exists
```
macfarla, line 144:
```suggestion
                 * and either no slots are present, or the first slot is empty
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
```diff
@@ -131,15 +134,36 @@ public CompletableFuture<List<Task<SnapDataRequest>>> requestStorage(
...
-                      worldStateProofProvider,
-                      response.slots().get(i),
-                      i < response.slots().size() - 1 ? new ArrayDeque<>() : response.proofs());
+                final ArrayDeque<NavigableMap<Bytes32, Bytes>> slots = new ArrayDeque<>();
+                // Check if we have an empty range
+
+                /*
+                 * Checks if the response represents an "empty range".
+                 *
+                 * An "empty range" is defined as a response where either no slots are present,
```
macfarla, line 143:
```suggestion
                 * An "empty range" is defined as a response where at least one proof exists
```

macfarla (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/RequestDataStep.java
macfarla: added a suggested rewrite for the comment

PR #6848, opened 2024-03-29
[Issue-6301] Add bad block events

mbaxter:
## PR description
Add bad block events and expose them via the plugin API (BesuEvents).

## Fixed Issue(s)
Part of https://github.com/hyperledger/besu/issues/6301


plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BadBlockCause.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * A more descriptive explanation for why the block was marked bad
+   *
+   * @return the description
+   */
+  String getDescription();
+
+  /** An enum representing the reason why a block is marked bad */
+  enum BadBlockReason {
```
mbaxter, line 36: @usmansaleem - I didn't completely follow why you had to move the nested enum in PR https://github.com/hyperledger/besu/pull/6716.  Tagging you here in case you think this is a problem?

usmansaleem: @mbaxter This should not be a problem in the interface as it will be treated as `static final` member. The reason for #6716 was weird errorprone issues where it was tripping and applied incorrect warnings (and only in that class).

usmansaleem (approved):
LGTM

PR #6858, opened 2024-04-02
[MINOR] Use 7 chars instead of 10 for source build hash

siladu:
Think 7 is the standard length...at least it matches what's in https://github.com/hyperledger/besu/commits/main/ which makes cross referencing a deployment to a commit easier.


usmansaleem (approved):
LGTM

garyschulte (approved):

PR #6859, opened 2024-04-02
Remove develop prelease

siladu:
Create noise in release notifications, can use "build from source" alternatives


siladu:
replaced by https://github.com/hyperledger/besu/pull/6860

PR #6860, opened 2024-04-02
Remove develop prelease workflow

siladu:
Creates noise in release notifications, can use "build from source" alternatives


jflo (approved):

garyschulte:
FYI, hive is defaulting/expecting `develop` to be the [latest unstable](https://github.com/ethereum/hive/blob/master/clients/besu/Dockerfile#L3) version of besu to test nightlies with.  What do we propose to use as a floating 'latest' target?

PR #6866, opened 2024-04-02
use event name instead of ref, which may be non-deterministic

jflo:
## PR description

An interesting subtlety of the github api is that when referring to a ref, there may be multiple and they are sorted alphabetically. So if we assume that the version to use in the build is what we tagged it as, we will only get the right value if the second tag is alphabetically earlier.

This fixes it by using whatever name is given to the release. Don't be clever, just use calver for now!

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -27,7 +27,7 @@ jobs:
           cache-disabled: true
       - name: assemble release
         run:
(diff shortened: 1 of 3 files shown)


siladu (approved):
Let's try it! Is there a safe way of testing this without doing a release?

jflo:
> Let's try it! Is there a safe way of testing this without doing a release?

Yes, and I did so in my fork.

PR #6883, opened 2024-04-04
Implemented Validator Exit contract helper

lucassaldanha:
- **EIP-7002: Add exits/exits_root to block/block_header and update execution payload for engineV4 methods**
- **Pending changes**
- **Fix BlockHeaderTestFixture**
- **Fix tests**
- **Fix another error**
- **Updated plugin api hash**
- **Fix AT**
- **Remove 7002 precompile entry**
- **Validator exit contract helper**
- **ValidatorExitsValidator tests**
- **Fix ValidatorExit toString()**
- **Adding exits to created blocks and processing exits when importing blocks**
- **Fix missing headers**
- **Added comment**
- **Spotless**
- **Fix padding on validtor exit contract storage**
- **Fixing tests after updating padding**


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ValidatorExitsValidator.java
```diff
@@ -40,17 +49,78 @@ public boolean validateValidatorExitParameter(
...
+        return false;
+      }
+
+      if (block.getBody().getExits().isEmpty()) {
+        LOG.warn("Block {} must contain exits (even if empty list)", blockHash);
+        return false;
+      }
+
+      final List<ValidatorExit> exitsInBlock = block.getBody().getExits().get();
+      // TODO Do we need to allow for customization? (e.g. if the value changes in the next fork)
```
Gabriel-Trintinalia, line 98: Suggestion: We could rename it to PrageExitsValidator, and make it a standalone class. If the next fork needs to change this we can easily abstract a class override what we need.

Gabriel-Trintinalia: nit: Also, giving the validations, the semantics here are more like a `RequiredExitsValidator` than an `AllowedExistsValidator`.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -197,6 +197,12 @@ public BlockProcessingResult processBlock(
       }
     }
 
+    final ValidatorExitsValidator exitsValidator = protocolSpec.getExitsValidator();
+    if (exitsValidator instanceof ValidatorExitsValidator.AllowedExits) {
```
Gabriel-Trintinalia, line 201: I think that the verification to perform the popping should belong to the validator/processor itself and not the blockprocessor. Maybe an extra method allowExists()?

Gabriel-Trintinalia (commented):
Nice testing, PR looks good. I have added some comments.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ValidatorExitsValidator.java
```diff
@@ -40,17 +49,78 @@ public boolean validateValidatorExitParameter(
...
+
+    @Override
+    public boolean validateExitsInBlock(
+        final Block block, final List<ValidatorExit> expectedExits) {
+      final Hash blockHash = block.getHash();
+
+      if (block.getHeader().getExitsRoot().isEmpty()) {
+        LOG.warn("Block {} must contain exits_root", blockHash);
+        return false;
+      }
```
Gabriel-Trintinalia, line 90: This piece should be probably in a headervalidationrules instead of here

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ValidatorExitsValidator.java
```diff
@@ -40,17 +49,78 @@ public boolean validateValidatorExitParameter(
...
+
+      // Validate exits_root
+      final Hash expectedExitsRoot = BodyValidation.exitsRoot(exitsInBlock);
+      if (!expectedExitsRoot.equals(block.getHeader().getExitsRoot().get())) {
+        LOG.warn(
+            "Block {} exits_root does not match expected hash root for exits in block", blockHash);
+        return false;
+      }
```
Gabriel-Trintinalia, line 110: s.a

Gabriel-Trintinalia (approved):
LGTM

PR #6894, opened 2024-04-06
Default bonsai to fully flat db and code storage by codehash 

garyschulte:
## PR description

This PR changes the default storage mode of bonsai to use code storage by code hash, and uses flat db by default.  This supports the future direction of bonsai, and simplifies the configuration necessary for besu to serve snap protocol (see description on #6640, now just `--Xsnapsync-server-enabled` is necessary).

changes:
* defaults bonsai to using a fully flattened db
* defaults bonsai to using code stored by code hash
* deprecates snap sync full flat config param
* creates --Xbonsai-full-flat-db-enabled parameter
* FlatDbStrategyProvider defers config if db is empty and persists a flat db strategy config

tested with:
* full sync, correctly creates full flattened db
* snap sync, correctly creates full flattend db, correctly heals flat db
* existing holesky sync, with FULL flat db  and code by CODE_HASH
  * starting with config for partial db and code by account hash correctly defers to db config and logs actual
* existing 24.3.0 holesky with PARTIAL flat db and code by ACCOUNT_HASH
  * starting with configs for full and code by code hash correctly defers to db config and logs actual

### DIFF
--- a/besu/src/test/resources/everything_config.toml
+++ b/besu/src/test/resources/everything_config.toml
@@ -219,6 +219,7 @@ receipt-compaction-enabled=true
 Xsecp256k1-native-enabled=false
 Xaltbn128-native-enabled=false
 Xsnapsync-server-enabled=true
+Xbonsai-full-flat-db-enabled=true
 
 # compatibility flags
 compatibility-eth64-forkid-enabled=false
--- a/besu/src/test/resources/everything_config.toml
+++ b/besu/src/test/resources/everything_config.toml
@@ -218,6 +218,7 @@ receipt-compaction-enabled=true
 # feature flags
 Xsecp256k1-native-enabled=false
 Xaltbn128-native-enabled=false
+Xsnapsync-server-enabled=true
 
 # compatibility flags
 compatibility-eth64-forkid-enabled=false
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProvider.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProvider.java
@@ -148,7 +148,6 @@ public class FlatDbStrategyProvider {
   public void upgradeToFullFlatDbMode(final SegmentedKeyValueStorage composedWorldStateStorage) {
     final SegmentedKeyValueStorageTransaction transaction =
         composedWorldStateStorage.startTransaction();
-    // TODO: consider ARCHIVE mode
     transaction.put(
         TRIE_BRANCH_STORAGE, FLAT_DB_MODE, FlatDbMode.FULL.getVersion().toArrayUnsafe());
     transaction.commit();
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProvider.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProvider.java
@@ -16,6 +16,7 @@ package org.hyperledger.besu.ethereum.trie.diffbased.common.storage.flat;
 
 import static org.hyperledger.besu.ethereum.storage.keyvalue.KeyValueSegmentIdentifier.CODE_STORAGE;
 import static org.hyperledger.besu.ethereum.storage.keyvalue.KeyValueSegmentIdentifier.TRIE_BRANCH_STORAGE;
+import static org.hyperledger.besu.ethereum.trie.diffbased.common.storage.DiffBasedWorldStateKeyValueStorage.WORLD_ROOT_HASH_KEY;
 
 import org.hyperledger.besu.ethereum.trie.diffbased.bonsai.storage.flat.FullFlatDbStrategy;
 import org.hyperledger.besu.ethereum.trie.diffbased.bonsai.storage.flat.PartialFlatDbStrategy;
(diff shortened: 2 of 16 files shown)


garyschulte:
It is worth noting there is a side effect of this PR on worldstate fault heal/recovery.   If a db with a `FULL` flat db strategy encounters an inconsistent worldstate db, as part of fixing the worldstate, the database will be left with a `PARTIAL` flat db strategy after fixing the worldstate.  

We can re-enable flat db after the healing, but this side effect might actually be desirable.

besu/src/test/resources/everything_config.toml
```diff
@@ -217,6 +217,7 @@ receipt-compaction-enabled=true
 # feature flags
 Xsecp256k1-native-enabled=false
 Xaltbn128-native-enabled=false
+Xsnapsync-server-enabled=true
```
matkt, line 220: we should not add the new full flat db flag ?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiSnapshotIsolationTests.java
```diff
@@ -152,6 +152,8 @@ public void assertCloseDisposesOfStateWithoutCommitting() {
       assertThat(res.isSuccessful()).isTrue();
       shouldCloseSnapshot.persist(oneTx.getHeader());
 
+      // TODO: full flat strategy will not be updated with these values, only the trie
+      //       determine how we want to make this assertion for full flat
```
matkt, line 156: I think we can use an updater in order to fix the test

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProvider.java
```diff
@@ -71,12 +72,36 @@ public void loadFlatDbStrategy(final SegmentedKeyValueStorage composedWorldState
...
+    final var existingTrieData =
+        composedWorldStateStorage.get(TRIE_BRANCH_STORAGE, WORLD_ROOT_HASH_KEY).isPresent();
+
     var flatDbMode =
         FlatDbMode.fromVersion(
             composedWorldStateStorage
                 .get(TRIE_BRANCH_STORAGE, FLAT_DB_MODE)
                 .map(Bytes::wrap)
-                .orElse(FlatDbMode.PARTIAL.getVersion()));
+                .orElseGet(
```
matkt, line 88: 👍

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProvider.java
```diff
@@ -71,12 +72,36 @@ public void loadFlatDbStrategy(final SegmentedKeyValueStorage composedWorldState
 
   @VisibleForTesting
   FlatDbMode deriveFlatDbStrategy(final SegmentedKeyValueStorage composedWorldStateStorage) {
+    final FlatDbMode requestedFlatDbMode =
+        dataStorageConfiguration.getUnstable().getBonsaiFullFlatDbEnabled()
```
matkt, line 76: is this flag still in unstable ?

matkt (commented):
added small comments

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProvider.java
garyschulte: it is, but the generated have to be rebuilt to see it.  Pretty much all of the bonsai configs are in unstable, we should move them all out of unstable when we change the default to bonsai :+1:

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiSnapshotIsolationTests.java
garyschulte: I thought about it and realized it will degrade the performance of the 'normal' use case for frozen/cached worldstates.  I think this note here should be sufficient to explain why we are using PARTIAL for the test - removed from TODO.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldStateDownloader.java
```diff
@@ -183,16 +183,6 @@ public CompletableFuture<Void> run(
       } else {
         // start from scratch
         worldStateStorageCoordinator.clear();
-        // we have to upgrade to full flat db mode if we are in bonsai mode
-        if (snapSyncConfiguration.isFlatDbHealingEnabled()) {
```
matkt, line 187: we don't have to downgradeToPartialFlatDbMode  if a user disable the full flat flag ? as the default is full now

garyschulte: that is a good point.  I hadn't considered that anybody would want to run in a partial mode

matkt: We can prevent user from doing that. It would not be logic

PR #6903, opened 2024-04-08
Expose transaction count by type metrics for the layered txpool

fab-10:
## PR description

Expose transaction count by type metrics for each layer of the layered txpool

The new metric name is `besu_transaction_pool_number_of_transactions_by_type` and has `layer` and `type` as labels.

Example panel 
image

Besu Full Grafana dashboard will be updated to render this metrics once the PR will be merged.

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolMetrics.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolMetrics.java
@@ -14,6 +14,7 @@
  */
 package org.hyperledger.besu.ethereum.eth.transactions;
 
+import org.hyperledger.besu.datatypes.TransactionType;
 import org.hyperledger.besu.ethereum.transaction.TransactionInvalidReason;
 import org.hyperledger.besu.metrics.BesuMetricCategory;
 import org.hyperledger.besu.metrics.ReplaceableDoubleSupplier;
@@ -27,6 +28,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.function.DoubleSupplier;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractTransactionsLayer.java
```diff
@@ -91,6 +93,11 @@ protected AbstractTransactionsLayer(
     metrics.initSpaceUsed(this::getLayerSpaceUsed, name());
     metrics.initTransactionCount(pendingTransactions::size, name());
     metrics.initUniqueSenderCount(txsBySender::size, name());
+    Arrays.stream(TransactionType.values())
+        .forEach(
+            type ->
+                metrics.initTransactionCountByType(
+                    () -> txCountByType[type.ordinal()], name(), type));
```
shemnon, line 100: using ordinal to track the external feels fragile, but if experimental tx types start showing up they may leave large gaps.  However... it's what we express externally, so this looks like it will work.  See my other comment on the metric.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolMetrics.java
```diff
@@ -97,6 +102,14 @@ public TransactionPoolMetrics(final MetricsSystem metricsSystem) {
             "The number of transactions currently present in the layer",
             "layer");
 
+    transactionCountByType =
+        metricsSystem.createLabelledGauge(
+            BesuMetricCategory.TRANSACTION_POOL,
+            "number_of_transactions_by_type",
+            "The number of transactions, of a specified type, currently present in the layer",
+            "layer",
+            "type");
```
shemnon, line 111: In addition to the type name, should we expose the EIP-2718 type number?  We shouldn't use ordinal and should encode it in TransactionType but that has some weirdness for frontier types.  May be best for a future PR.

shemnon (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolMetrics.java
fab-10: yes I think it makes sense to explicitly add the EIP-2718 type number to the tx type, and expose it in the metric label.
About the use of `type.ordinal()`, it is only used internally to index the array of int for the counters, and not exposed outside.

PR #6905, opened 2024-04-08
Refactor Gas Calculator Create methods

shemnon:
## PR description

Prepatory for EOF.  Refactor the gas calculation logic so that it does
not depend on specific stack entries, and break out the pieces of the
calculation for readability.


evm/src/main/java/org/hyperledger/besu/evm/internal/Words.java
```diff
@@ -170,7 +206,7 @@ static int readBigEndianU16(final int index, final byte[] array) {
     if (index + 1 >= array.length) {
       throw new IndexOutOfBoundsException();
     }
-    return ((array[index] & 0xff) << 8) | (array[index + 1] & 0xff);
+    return ((array[index] << 8) & 0xff00) | (array[index + 1] & 0xff);
```
jflo, line 209: This seems like the same result as before but masks the full 16 bits before or'ing the lower half. Whats the preference behind the change?

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/ConstantinopleGasCalculator.java
```diff
@@ -39,12 +37,24 @@ public class ConstantinopleGasCalculator extends ByzantiumGasCalculator {
 
   private static final long EXTCODE_HASH_COST = 400L;
 
+  /**
+   * Returns the amount of gas the CREATE2 operation will consume.
+   *
+   * @param frame The current frame
+   * @return the amount of gas the CREATE2 operation will consume
+   * @deprecated Compose the operation cost from {@link #txCreateCost()}, {@link
```
jflo, line 45: Perhaps a mention of the Create2Opcode.cost() method since it is implemented there?

evm/src/main/java/org/hyperledger/besu/evm/operation/Create2Operation.java
```diff
@@ -40,20 +46,34 @@ public Create2Operation(final GasCalculator gasCalculator, final int maxInitcode
...
   @Override
-  public long cost(final MessageFrame frame) {
-    return gasCalculator().create2OperationGasCost(frame);
+  protected Code getInitCode(final MessageFrame frame, final EVM evm) {
+    final long inputOffset = clampedToLong(frame.getStackItem(1));
+    final long inputSize = clampedToLong(frame.getStackItem(2));
+    final Bytes inputData = frame.readMemory(inputOffset, inputSize);
+    // Never cache CREATEx initcode. The amount of reuse is very low, and caching mostly
+    // addresses disk loading delay, and we already have the code.
+    return evm.getCode(null, inputData);
```
jflo, line 77: Had to peek in here to remind myself why the comment was necessary, and maybe evm.getCode() should be separated into getCode() and createCode()? with getCode delegating when necessary. Then this opcode can be a little more clear about its intent.

evm/src/main/java/org/hyperledger/besu/evm/internal/Words.java
```diff
@@ -94,6 +107,29 @@ static long clampedToLong(final Bytes uint) {
...
   }
 
+  /**
+   * The value of the bytes as though it was representing an unsigned integer, however if the value
+   * exceeds Integer.MAX_VALUE then Integer.MAX_VALUE will be returned.
+   *
+   * @param uint the unsigned integer
+   * @return the least of the integer value or Integer.MAX_VALUE
+   */
+  static int clampedToInt(final Bytes uint) {
```
jflo, line 117: I think this warrants some test coverage.

jflo (changes requested):
Would like to see basic test coverage of Words.clampedToInt, and it is likely time to break evm.getCode() into 2 methods.

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/ConstantinopleGasCalculator.java
shemnon: Also added to javadoc in for createOperationGasCost.

evm/src/main/java/org/hyperledger/besu/evm/internal/Words.java
shemnon: From different drafts of the EOF fork I was working on.  This fromulation has less issues with sign extension when you do the masking last, and makes the byte position clear. More of an issue with right shifts but masking last has reduced errors in other code I've worked on.

evm/src/main/java/org/hyperledger/besu/evm/internal/Words.java
shemnon: Vectors added

diega (approved):
looks really good to me

evm/src/test/java/org/hyperledger/besu/evm/internal/WordsTest.java
```diff
@@ -46,7 +48,61 @@ Collection<Object[]> unsignedMinLongTestVector() {
...
+            Bytes.fromHexString(
+                "0x0000000000000000000000000000000000000000000000000000000080000001"),
+            Integer.MAX_VALUE
+          },
+        });
+  }
+
+  @ParameterizedTest
+  @MethodSource("clampedToLongTestVector")
+  void clampedToLongTest(final Bytes theBytes, final int theExpectedInt) {
```
jflo, line 105:
```suggestion
  void clampedToIntTest(final Bytes theBytes, final int theExpectedInt) {
```

jflo (approved):
looks great. includes suggestion to fix typo.

PR #6916, opened 2024-04-09
Fix and reformat produced block timing log

fab-10:

### DIFF
--- a/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockMiner.java
+++ b/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockMiner.java
@@ -168,7 +168,7 @@ public class BlockMiner<M extends AbstractBlockCreator> implements Runnable {
   }
 
   private void logProducedBlock(final Block block, final BlockCreationTiming blockCreationTiming) {
-    String log =
+    LOG.info(
         String.format(
             "Produced #%,d / %d tx / %d om / %,d (%01.1f%%) gas / (%s) in %01.3fs",
             block.getHeader().getNumber(),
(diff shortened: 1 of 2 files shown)


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockMiner.java
```diff
@@ -177,13 +177,9 @@ private void logProducedBlock(final Block block, final BlockCreationTiming block
...
             block.getHash(),
-            blockCreationTiming.end("log").getSeconds() / 1000.0);
+            blockCreationTiming.end("log").toMillis() / 1000.0));
 
-    if (LOG.isDebugEnabled()) {
-      log = log + ", timing [" + blockCreationTiming + "]";
-    }
-
-    LOG.info(log);
+    LOG.debug("Timing #{} / {}", block.getHeader().getNumber(), blockCreationTiming);
```
ahamlat, line 182: You don't test here if Debug log are enabled ?

fab-10: in this case there is no need, because, if debug is not enabled there is no extra wasted computation done, since `blockCreationTiming.toString()` is not called, and `block.getHeader().getNumber()` is just accessing a var

ahamlat: Yes, that's right, but if someone change it in the future without paying attention to the attributes computation, it will generate an overhead

ahamlat (approved):

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/BlockMiner.java
ahamlat: It is not blocking though.

PR #6943, opened 2024-04-12
Fix named network default disabled for downgrade protection

garyschulte:
## PR description
Address downgrade default for named networks.  Previously the check for named network was only considering command line params, and not toml config files.

Also implemented Comparable on VersionMetadata, and added explicit version test cases for a variety of build versions.

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -2679,7 +2679,8 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   private Boolean getDefaultVersionCompatibilityProtectionIfNotSet() {
     // Version compatibility protection is enabled by default for non-named networks
     return Optional.ofNullable(versionCompatibilityProtection)
-        .orElse(commandLine.getParseResult().hasMatchedOption("network") ? false : true);
+        // if we have a specific genesis file or custom network id, we are not using a named network
+        .orElse(genesisFile != null || networkId != null);
   }
 
   private String generateConfigurationOverview() {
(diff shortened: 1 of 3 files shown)


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2679,7 +2679,8 @@ private SyncMode getDefaultSyncModeIfNotSet() {
   private Boolean getDefaultVersionCompatibilityProtectionIfNotSet() {
     // Version compatibility protection is enabled by default for non-named networks
     return Optional.ofNullable(versionCompatibilityProtection)
-        .orElse(commandLine.getParseResult().hasMatchedOption("network") ? false : true);
+        // if we have a specific genesis file or custom network id, we are not using a named network
+        .orElse(genesisFile != null || networkId != null);
```
non-fungible-nelson, line 2683: does this handle the case where no network is set, but the user is targeting mainnet?

non-fungible-nelson (commented):
one nit

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
garyschulte: Yes, network is by default MAINNET, unless otherwise specified.  network-id and genesis-file are parameters that are necessary if you are not using a named network.  So if those are not specified, we are returning false (version protection disabled)

non-fungible-nelson: ahh gotcha - so this isn't looking at the config `--network` but instead parsing Besu's Named Network from runtime ?

garyschulte: We are not looking at `network` because it is always something (defaults to mainnet).  What we were doing previously is looking at the command line parameters for `--network`, but that does not take into account toml config options.  

So now instead we are inverting the condition and looking at `genesisFile` and `networkId` which are populated from both the cli parameters and the toml config options.

non-fungible-nelson:
LGTM 🚢  Cannot provide approval

macfarla (approved):

PR #6948, opened 2024-04-15
Attempt to speed up ATs

siladu:
Attempt to revert long running AT times from 7e46889


fab-10:
The commit that I did, is actually fixing the intent of the test, that was to stop the cluster after each single unit tests, and start with a new one every time, that intended behavior was broken, and had the side effect of keeping previously started Besu processes around for the duration of the test class, and so had the nice side effect that for all the test methods after the first a cluster was already up to respond to requests.
So instead of reverting to the wrong but faster state, I will prefer that we understand if the initial intends of restarting the cluster after each test methods was really needed and instead replace it with a restart each test class.
It is also worth to note that ATs in CI could be re-organized to balance the time.

fab-10 (approved):
after a quick analysis, I found there are some steps needed to first port all the tests to Junit5, then identify test classes that do not need to have a cluster restart for every method and probably also a refactor of the base classes, so for the moment we can revert to the previous buggy but faster code, while implementing these pre tasks.

PR #6952, opened 2024-04-15
optimize node startup speed and memory allocation

lyfsn:
## PR description

In summary, if a user employs a large genesis file and activates the `--genesis-state-hash-cache-enabled` parameter, their node can start in under **10 seconds**, requiring less than **2GB of memory**. This is the essence of what this PR accomplishes.

1. Regarding the process of serializing the string into a GenesisConfigFile, many parts of the code follow a similar logic pattern: `GenesisConfigFile.fromConfig(genesisContent).getConfigOptions()`, which means obtaining config options from the GenesisConfigFile. In fact, it does not care about the rest of the content, especially the accounts information, while large genesis files often have accounts occupying a large portion.

    Therefore, this PR introduces a new method, `fromConfigWithoutAccounts`, for all places where accounts are not needed. During the serialization of the string into a GenesisConfigFile, it ignores the accounts information. This change is should be safe because the only time accounts information is needed is during the calculation of the genesis state hash.

2. Based on the `-genesis-state-hash-cache-enabled` parameter added in https://github.com/hyperledger/besu/pull/6758, this PR further enhances the performance improvements that the parameter can offer.

    Originally, the parameter's purpose was to cache and skip the calculation of the genesis state hash value. Building upon the previous improvements, this PR determines that if this parameter is enabled and there is a cached genesis state hash in the database, then it will directly ignore the accounts content when reading the genesis file.

The table below outlines the comparison of node startup time and total memory reclaimed before and after the optimizations made in this PR. All tests utilized a genesis file of 1.1GB in size.

|  | Time Consuming - Before optimization | Memory Allocated - Before optimization | Time Consuming - Optimized (this PR) | Memory Allocated - Optimized (this PR) |
| --- | --- | --- | --- | --- |
| --genesis-state-hash-cache-enabled=false && block number=0 | 7min37s | 184.95GB | 7min24s | 158.83GB |
| --genesis-state-hash-cache-enabled=false && block number=1 | 2min53s | 125.47GB | 2min36s | 96.53GB |
| --genesis-state-hash-cache-enabled=true && block number=0 | 7min35s | 183.33GB | 7min34s | 154.68GB |
| --genesis-state-hash-cache-enabled=true && block number=1 | 0min50s | 66.75GB | 0min9s | 1.31GB |

Without the `--genesis-state-hash-cache-enabled` parameter enabled, the startup time reduced from **2 minutes and 53 seconds** to **2 minutes and 36 seconds**, and the total memory reclaimed decreased from **125.47GB** to **96.53GB**. The optimization effect here is modest.

With the `--genesis-state-hash-cache-enabled` parameter enabled, the startup time can be reduced from **50 seconds** to **9 seconds**, and the total memory reclaimed from **66.75GB** to **1.31GB**, marking a significant optimization effect.

This is the unit test coverage report for the PR: [Unit Test Coverage Report](https://lyfsn.github.io/besu-starter/pr2/).

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -91,6 +91,7 @@ import org.hyperledger.besu.components.BesuComponent;
 import org.hyperledger.besu.config.CheckpointConfigOptions;
 import org.hyperledger.besu.config.GenesisConfigFile;
 import org.hyperledger.besu.config.GenesisConfigOptions;
+import org.hyperledger.besu.config.JsonUtil;
 import org.hyperledger.besu.config.MergeConfigOptions;
 import org.hyperledger.besu.consensus.qbft.pki.PkiBlockCreationConfiguration;
 import org.hyperledger.besu.consensus.qbft.pki.PkiBlockCreationConfigurationProvider;
@@ -117,6 +118,7 @@ import org.hyperledger.besu.ethereum.api.jsonrpc.ipc.JsonRpcIpcConfiguration;
 import org.hyperledger.besu.ethereum.api.jsonrpc.websocket.WebSocketConfiguration;
 import org.hyperledger.besu.ethereum.api.query.BlockchainQueries;
 import org.hyperledger.besu.ethereum.chain.Blockchain;
+import org.hyperledger.besu.ethereum.chain.VariablesStorage;
 import org.hyperledger.besu.ethereum.core.MiningParameters;
 import org.hyperledger.besu.ethereum.core.MiningParametersMetrics;
 import org.hyperledger.besu.ethereum.core.PrivacyParameters;
@@ -1598,7 +1600,8 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   private GenesisConfigOptions readGenesisConfigOptions() {
 
     try {
-      final GenesisConfigFile genesisConfigFile = GenesisConfigFile.fromConfig(genesisConfig());
+      final GenesisConfigFile genesisConfigFile =
+          GenesisConfigFile.fromConfigWithoutAccounts(genesisConfig());
       genesisConfigOptions = genesisConfigFile.getConfigOptions(genesisConfigOverrides);
     } catch (final Exception e) {
       throw new ParameterException(
@@ -2352,16 +2355,46 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   }
 
   private GenesisConfigFile getGenesisConfigFile() {
-    return GenesisConfigFile.fromConfig(genesisConfig());
+    return GenesisConfigFile.fromConfigWithoutAccounts(genesisConfig());
   }
 
+  private String genesisConfigString = "";
+
   private String genesisConfig() {
-    try {
-      return Resources.toString(genesisFile.toURI().toURL(), UTF_8);
-    } catch (final IOException e) {
-      throw new ParameterException(
-          this.commandLine, String.format("Unable to load genesis URL %s.", genesisFile), e);
+    if (!genesisConfigString.isEmpty()) {
+      return genesisConfigString;
     }
+    if (genesisStateHashCacheEnabled) {
+      // If the genesis state hash is present in the database, we can use the genesis file without
+      pluginCommonConfiguration.init(
+          dataDir(),
+          dataDir().resolve(DATABASE_PATH),
+          getDataStorageConfiguration(),
+          getMiningParameters());
+      final KeyValueStorageProvider storageProvider = keyValueStorageProvider(keyValueStorageName);
+      if (storageProvider != null) {
+        boolean isGenesisStateHashPresent;
+        try {
+          // A null pointer exception may be thrown here if the database is not initialized.
+          VariablesStorage variablesStorage = storageProvider.createVariablesStorage();
+          Optional<Hash> genesisStateHash = variablesStorage.getGenesisStateHash();
+          isGenesisStateHashPresent = genesisStateHash.isPresent();
+        } catch (Exception ignored) {
+          isGenesisStateHashPresent = false;
+        }
+        if (isGenesisStateHashPresent) {
+          genesisConfigString = JsonUtil.getJsonFromFileWithout(genesisFile, "alloc");
+        }
+      }
+    }
+    if (genesisConfigString.isEmpty()) {
+      try {
+        genesisConfigString = Resources.toString(genesisFile.toURI().toURL(), UTF_8);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }
+    return genesisConfigString;
   }
 
   private static String genesisConfig(final NetworkName networkName) {
@@ -2607,7 +2640,7 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
     return Optional.ofNullable(genesisConfigOptions)
         .orElseGet(
             () ->
-                GenesisConfigFile.fromConfig(
+                GenesisConfigFile.fromConfigWithoutAccounts(
                         genesisConfig(Optional.ofNullable(network).orElse(MAINNET)))
                     .getConfigOptions(genesisConfigOverrides));
   }
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -2358,12 +2358,9 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
     return GenesisConfigFile.fromConfigWithoutAccounts(genesisConfig());
   }
 
-  private String genesisConfigString = "";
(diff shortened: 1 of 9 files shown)


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2352,16 +2355,46 @@ private EthNetworkConfig updateNetworkConfig(final NetworkName network) {
   }
 
   private GenesisConfigFile getGenesisConfigFile() {
-    return GenesisConfigFile.fromConfig(genesisConfig());
+    return GenesisConfigFile.fromConfigWithoutAccounts(genesisConfig());
   }
 
+  private String genesisConfigString = "";
```
fab-10, line 2361: [A `Suppliers.memoize` ](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Suppliers.html#memoize(com.google.common.base.Supplier)) seems better suited for this field

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2352,16 +2355,46 @@ private EthNetworkConfig updateNetworkConfig(final NetworkName network) {
...
+
   private String genesisConfig() {
```
fab-10, line 2363: now this method does not only return the genesis config, so the name is no more descriptive of its behavior

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2352,16 +2355,46 @@ private EthNetworkConfig updateNetworkConfig(final NetworkName network) {
...
+          VariablesStorage variablesStorage = storageProvider.createVariablesStorage();
+          Optional<Hash> genesisStateHash = variablesStorage.getGenesisStateHash();
+          isGenesisStateHashPresent = genesisStateHash.isPresent();
+        } catch (Exception ignored) {
+          isGenesisStateHashPresent = false;
+        }
+        if (isGenesisStateHashPresent) {
+          genesisConfigString = JsonUtil.getJsonFromFileWithout(genesisFile, "alloc");
+        }
+      }
```
fab-10, line 2388: please explain why this code is needed here now, so we can find a better organization for it

fab-10 (commented):
I understand Besu startup code is not easy, so added some initial comments to try to improve the code in `BesuCommand`

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
lyfsn: In the original code flow, `genesis file (disk)` --(1)--> `genesis string` --(2)--> `genesis struct`.

Here, the code is positioned at step (1).

If the node has the `--genesis-state-hash-cache-enabled` parameter enabled, the value of `state root hash` will be saved in `variablesStorage`.

The intent of this code is to determine if the value of `state root hash` already exists in the node's variablesStorage, and if so, during step (1), it will directly ignore the `alloc` field in the gensis.json file.

Because if the `state root hash` value already exists, it indicates that the account information in the `genesis.json` file has already been written into the state database, and there is no need to calculate the `state root hash` again later.

config/src/main/java/org/hyperledger/besu/config/JsonUtil.java
```diff
@@ -466,4 +511,94 @@ private static boolean validateInt(final JsonNode node) {
...
+      while ((token = parser.nextToken()) != null) {
+        if (token == JsonToken.START_OBJECT) {
+          jsonBuilder.append(handleObject(parser, excludedFieldName));
+        }
+      }
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+    return jsonBuilder.toString();
+  }
```
fab-10, line 541: Have you checked if Jackson natively support excluding fields without having to implement the parsing methods?

lyfsn: Yes, I have checked it, and this is the fastest method with the least memory usage. I will check it again in a moment.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
lyfsn: Do you mean the method "genesisConfig()"?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
lyfsn: I made a commit to use Suppliers.memoize: https://github.com/hyperledger/besu/pull/6952/commits/8ed9e4c32f203f8bb75479015d3d92e49e6b98b6

ahamlat:
@lyfsn Great work on the PR, I have few questions if you don't mind :

- Is there a way to reproduce the numbers on my side, by sharing the big (fake) genesis file, and the configuration you used for your test ? 
- Is "requiring less than 2GB of memory" a strong prerequisite for your use case ? 
- Can you share the -Xmx value used during your test?
- Have you checked GC activity during the test period ? 
- Have you done some profiling to see what is taking more than 7 minutes in the first and third test with this PR's code ?

lyfsn:
> @lyfsn Great work on the PR, I have few questions if you don't mind :
> 
> * Is there a way to reproduce the numbers on my side, by sharing the big (fake) genesis file, and the configuration you used for your test ?
> * Is "requiring less than 2GB of memory" a strong prerequisite for your use case ?
> * Can you share the -Xmx value used during your test?
> * Have you checked GC activity during the test period ?
> * Have you done some profiling to see what is taking more than 7 minutes in the first and third test with this PR's code ?

- You can use genesis.json from this repository (though it's smaller than the one I used). [LInk](https://github.com/OpenFusionist/network_config)
- This repository has 8 `jfr` files; they are about memory allocation records. [LInk](https://github.com/lyfsn/besu-starter?tab=readme-ov-file#jfr)
- Also, in this repository, there are some records about profiler records. [Link](https://github.com/lyfsn/besu-starter?tab=readme-ov-file#empty-data-directory-on-startup-startup-time-7-minutes)
- I run the binary Besu directly, so I do not set -Xmx explicitly. I run this on a 64GB server, so the default should be 16GB.

You can refer to this information first, and if you have more specific needs, please feel free to let me explain!

fab-10:
While reviewing your PR, I saw the need for a refactor of how the genesis file and the genesis options are managed, since with the time some tech debt has accumulated, and with that I had the idea of making the load of the accounts lazy, with the hope of moving that logic in the class that manage the genesis file, will let you know when I have something ready to test.

ahamlat:
@lyfsn thanks for the troubleshooting data, awesome work.
I was able to reproduce some of the number with this PR, for `--genesis-state-hash-cache-enabled=true && block number=0`, but then I did the test on a previous release, 23.10.2 (without both PRs), and it was faster for the block 0 test :

- `block number=0` : 5 minutes
- `block number=1` : 4 minutes 17 seconds 

I'm currently analyzing the regression as we know what code is causing it.
I would like to know what was the test you shared in you blog where Besu took more than 100 minutes to start

lyfsn:
> @lyfsn thanks for the troubleshooting data, awesome work. I was able to reproduce some of the number with this PR, for `--genesis-state-hash-cache-enabled=true && block number=0`, but then I did the test on a previous release, 23.10.2 (without both PRs), and it was faster for the block 0 test :
> 
> * `block number=0` : 5 minutes
> * `block number=1` : 4 minutes 17 seconds
> 
> I'm currently analyzing the regression as we know what code is causing it. I would like to know what was the test you shared in you blog where Besu took more than 100 minutes to start

I guess there are two main reasons:
- At that time, we used our devnet file, which is about 1.1GB, a bit larger than the mainnet file, which is about 800MB. The mainnet file was provided to you for convenience in using publicly available files for testing.
- The server's hardware was limited at that time, with only 16GB of RAM and a set 16GB swap space. The swap space obviously slowed down the processing speed.

These were the parameters for the test at that time:
|Client|Startup command|Version|OS|Arch|CPU core|RAM|Swap space|
|--|--|--|--|--|--|--|--|
|Besu|env BESU_OPTS=-Xmx16g ./besu --(....)|24.3.0/openjdk17|ubuntu 22.04|x64|2|16GB|16GB|

Additionally, are you referring to this blog article? The article also detailed the testing conditions:
https://devlog.fusionist.io/posts/ethereum-execution-clients-memory-comparison-in-endurance/

fab-10:
are your genesis files public? I would like to test them on the refactor, if not I will try to generate one with random data

lyfsn:
> are your genesis files public? I would like to test them on the refactor, if not I will try to generate one with random data

I sent an email to your public email address; it contains the file access document.

fab-10:
@lyfsn could you test [this PR](https://github.com/hyperledger/besu/pull/6977)? 
The allocations are lazy loaded only when needed, and there is reduced memory usage when parsing it, with less object copies and allocations streaming.

fab-10:
Closing this as superseded by #6977

PR #6955, opened 2024-04-16
Fix docker goss test

### LINKED ISSUE #6929: GHA: buildDocker (self-hosted, ARM64) step failing
Last success was 2 weeks ago: https://github.com/hyperledger/besu/actions/runs/8487812388

Discovered during the 24.3.2 release process -
https://github.com/hyperledger/besu/actions/runs/8639188327/job/23686748688#step:9:478

It actually failed for 24.4.0-RC2 as well last week: https://github.com/hyperledger/besu/actions/runs/8500195660

```

INFO: Found goss_wait.yaml, waiting for it to pass before running tests
ERROR: goss_wait.yaml never passed
INFO: Deleting container

> Task :testDocker FAILED

* Where:
Build file '/home/runner/_work/besu/besu/build.gradle' line: 798

* What went wrong:
Execution failed for task ':testDocker'.
> Process 'command '/bin/bash'' finished with non-zero exit value 1
```

siladu:
Fixes docker.yml github workflow on self-hosted runner version 2.315.0 (docker ~25.0.4+~ 26.0.0?)

Since docker 26.0.0, IPv6 is used by default but our goss_wait test is expecting ports to be exposed on IPv4.
Disabling IPv6 should work for all docker versions, at least while the runners are on various versions of docker.

Once all runners are on 26+ then we could probably change goss_wait.yml to:

```
# runtime docker tests for interfaces & ports
port:
  tcp6:30303:
    listening: true
    ip:
    - '::'
  udp6:30303:
    listening: true
    ip:
    - '::'
  tcp6:8545:
    listening: true
    ip:
    - '::'
  tcp6:8546:
    listening: true
    ip:
    - '::'
  tcp6:8547:
    listening: true
    ip:
    - '::'
```

and reenable IPv6

Fixes #6929

### DIFF
--- a/docker/test.sh
+++ b/docker/test.sh
@@ -20,7 +20,7 @@ if [[ $i != 0 ]]; then exit $i; fi
 # we test that things listen on the right interface/port, not what interface the advertise
 # hence we dont set p2p-host=0.0.0.0 because this sets what its advertising to devp2p; the important piece is that it defaults to listening on all interfaces
 GOSS_FILES_PATH=$TEST_PATH/01 \
-bash $TEST_PATH/dgoss run $DOCKER_IMAGE \
+bash $TEST_PATH/dgoss run -p 8545:8545 -p 8546:8546 -p 8547:8547 -p 30303:30303 $DOCKER_IMAGE \
 --network=dev \
 --rpc-http-enabled \
 --rpc-ws-enabled \
--- a/docker/test.sh
+++ b/docker/test.sh
@@ -20,7 +20,7 @@ if [[ $i != 0 ]]; then exit $i; fi
 # we test that things listen on the right interface/port, not what interface the advertise
 # hence we dont set p2p-host=0.0.0.0 because this sets what its advertising to devp2p; the important piece is that it defaults to listening on all interfaces
 GOSS_FILES_PATH=$TEST_PATH/01 \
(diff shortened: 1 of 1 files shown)


siladu:
Doesn't actually fix the issue

siladu:
Seems to be introduced by this docker 26.0.0 feature...

https://docs.docker.com/engine/release-notes/26.0/#bug-fixes-and-enhancements-1

> Always attempt to enable IPv6 on a container's loopback interface, and only include IPv6 in /etc/hosts if successful. [moby/moby#47062](https://github.com/moby/moby/pull/47062)

> Note

> By default, IPv6 will remain enabled on a container's loopback interface when the container is not connected to an IPv6-enabled network. For example, containers that are only connected to an IPv4-only network now have the ::1 address on their loopback interface.

> To disable IPv6 in a container, use option --sysctl net.ipv6.conf.all.disable_ipv6=1 in the create or run command, or the equivalent sysctls option in the service configuration section of a Compose file.

> If IPv6 is not available in a container because it has been explicitly disabled for the container, or the host's networking stack does not have IPv6 enabled (or for any other reason) the container's /etc/hosts file will not include IPv6 entries.

jframe (approved):
Seems like a sensible approach until all the runners can be updated

siladu:
> Seems like a sensible approach until all the runners can be updated

@jframe spoke to Josh - this is the preferred approach. Most users should remain on ipv4

Issue #6884, opened 2024-04-05
Chain Download Halt


siladu:
Seen on 24.3.0 and 24.4.0-RC2

On fresh SNAP/CHECKPOINT sync on a variety of networks, block import sometimes halts and requires a restart, despite having a full set of peers and world state downloading finishing.

Example:
```
{"@timestamp":"2024-04-04T05:48:35,877","level":"INFO","thread":"EthScheduler-Services-37 (importBlock)","class":"ImportBlocksStep","message":"Block import progress: 4030400 of 5625239 (71%)","throwable":""}
{"@timestamp":"2024-04-04T05:48:39,175","level":"INFO","thread":"EthScheduler-Services-26 (batchPersistTrieNodeData)","class":"SnapSyncMetricsManager","message":"Healed 221585 world state trie nodes, Peer count: 6","throwable":""}
{"@timestamp":"2024-04-04T05:49:42,821","level":"INFO","thread":"EthScheduler-Services-36 (requestCompleteTask)","class":"SnapWorldDownloadState","message":"Pausing world state download while waiting for sync to complete","throwable":""}

// and then nothing until restart
```

siladu:
Tested this logging on a Sepolia node and it hung at 99% (note in this case, it looks like worldstate didn't finish)...

It gets stuck in a loop where it repeatedly prints this log for the same peers every few seconds.

```
...
// last world state log
2024-04-06 02:52:10.313+00:00 | EthScheduler-Services-3 (batchPersistAccountData) | INFO  | SnapSyncMetricsManager | Worldstate download progress: 49.64%, Peer count: 7
2024-04-06 02:52:41.596+00:00 | EthScheduler-Services-43 (importBlock) | INFO  | ImportBlocksStep | Block import progress: 3779220 of 5638048 (67%)
...
2024-04-06 07:18:36.991+00:00 | EthScheduler-Services-43 (importBlock) | INFO  | ImportBlocksStep | Block import progress: 5633620 of 5638048 (99%)
[...]

PR #6969, opened 2024-04-19
fix chain download halt

pinges:
## PR description
In some cases a syncing Besu node stopped downloading the chain data. The reason for that was that sometimes peers that were selected to download headers that are needed to allow the parallel download of the next batch of headers only returned one header instead of 200 headers. 
There was a mechanism that restarted the pipeline for the chain download, but if the same peer was selected to do that initial download the same would happen again. In some cases that happened for a long time, making it noticeable, but it turned out that this happens for short periods of time as well (still slowing down the chain download process).
This PR adds a check that disconnects the peer only returning one block to make sure that a different peer is selected for the next restart of the pipeline.

## Fixed Issue(s)
#6884


garyschulte:
great find.  We recently discovered that the geth snap sync pipeline does something very similar.

macfarla:
closing since this got merged into a different PR (already in)

PR #6973, opened 2024-04-22
add log-file flag

Brindrajsinh-Chauhan:
## PR description
This PR adds a new cli flag `--log-file` to configure the logs to be written to the provided file instead of the console. If this parameter is not specified, by default the logs are written to console.

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -798,6 +798,11 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   private static Boolean colorEnabled = null;
 
   @Option(
+      names = {"--log-file"},
+      description = "Log file to output logs")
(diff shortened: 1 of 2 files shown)


shemnon:
Is there a reason shell redirection is insufficient?  (`> logfile`)

This change would increase use of Log4j specific APIs, which is something I am hesitant to do after Log4Shell

fab-10:
I agree with @shemnon, and this seems also a redundant feature, since you can obtain the same with the output redirection for simple use case, or specifying a log4j2 configuration for production use cases where you want more control on the output, like rolling the file etc..., using the `log4j2.configurationFile` system properties or the `LOG4J_CONFIGURATION_FILE` env var.

It will be nice to go in the direction moving all log related configurations outside of Besu's code, so it is log framework agnostic.

Brindrajsinh-Chauhan:
Thank you for the feedback. This makes sense to me now. Will close this PR for now based on the comments and feedback.
Thank you again

PR #6978, opened 2024-04-22
Allow API configuration to override supplied gasLimit

garyschulte:
## PR description

Allow API configuration to override supplied gasLimit.  Originally this commit was part of #6641.  Moved here for visibility and easier review.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -235,11 +235,8 @@ public Optional<TransactionSimulatorResult> processWithWorldUpdater(
             ? callParams.getGasLimit()
             : blockHeaderToProcess.getGasLimit();
     if (rpcGasCap > 0) {
-      final long gasCap = rpcGasCap;
-      if (gasCap < gasLimit) {
-        gasLimit = gasCap;
-        LOG.info("Capping gasLimit to " + gasCap);
-      }
+      gasLimit = rpcGasCap;
+      LOG.info("Capping gasLimit to " + rpcGasCap);
```
macfarla, line 239: so no longer comparing gasCap to gasLimit, just if > zero?

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -565,8 +565,9 @@ public void shouldCapGasLimitWhenOriginalTransactionExceedsGasCap() {
   }
 
   @Test
-  public void shouldKeepOriginalGasLimitWhenCapIsHigherThanOriginalValue() {
-    // generate a transaction with a gas limit that is lower than the gas cap
+  public void shouldUseRpcGasCapWhenCapIsHigherThanGasLimit() {
```
macfarla, line 568: what if cap is lower than gasLimit

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
garyschulte: yes, it is essentially an override value from the cli.  @matkt can speak to the specifics about why we would raise the cap beyond the requested limit, but I think it has something to do with simulating transactions that will exceed the anticipated gas, but not exceed some predefined hard cap (DoS vector).

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
garyschulte: yeah, this test can really just be renamed to something like when present rpc gas cap supersedes request gas limit

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
matkt: yes it is need in order to allow user setting a high gasLimit for ethcall

matkt (approved):
Lgtm

Issue #6531, opened 2024-02-06
EIP-2935: Serve historical block hashes from state 


matkt:
### Description

For verkle we need to implement the EIP 2935 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2935.md

in order to save historical block hashes in the state. There is already an implementation in the verkle branch

Gabriel-Trintinalia:
@matkt do you want help extracting it from the Verkle branch into Besu?

matkt:
yes feel free to do it 👍 

most important is here 
- Write historical block history 
https://github.com/hyperledger/besu/blob/verkle/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HistoricalBlockHashProcessor.java
- Allow to add exception in order to write empty account in the trie
https://github.com/hyperledger/besu/blob/verkle/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ClearEmptyAccountStrategy.java
- Define the clear empty account in the strategy 
https://github.com/hyperledger/besu/blob/verkle/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java#L143
https://github.com/hyperledger/besu/blob/verkle/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java#L296
- Add clear empty account with exception for the new fork 
https://github.com/hyperledger/besu/blob/verkle/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java#L743
- Add historical block hash processor for the new fork 
https://github.com/hyperledger/besu/blob/verkle/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java#L769
- Call it during block processor 
https://github.com/hyperledger/besu/blob/verkle/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java#L115

etc..

PR #6991, opened 2024-04-23
EIP-2935

Gabriel-Trintinalia:
## PR description

fix https://github.com/hyperledger/besu/issues/6531

- [x] Extract EIP-2935 from the verkle branch
- [x] Fix Engine Prague Acceptance tests
- [x] Implement tests for BlockHashOperation and HistoricalBlockHashProcessor


evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
```diff
@@ -29,45 +32,93 @@
 /** The Block hash operation. */
 public class BlockHashOperation extends AbstractFixedCostOperation {
 
-  private static final int MAX_RELATIVE_BLOCK = 255;
+  /** The HISTORICAL_BLOCKHASH_ADDRESS */
+  public static final Address HISTORICAL_BLOCKHASH_ADDRESS =
+      Address.fromHexString("0xfffffffffffffffffffffffffffffffffffffffe");
+
+  private static final int MAX_RELATIVE_BLOCK = 256;
```
jflo, line 39: Suggest that we just reuse constant names from the EIP.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HistoricalBlockHashProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+ * Processes and stores historical block hashes in accordance with EIP-2935. This class is
+ * responsible for managing the storage of block hashes to support EIP-2935, which introduces
+ * historical block hash access in smart contracts.
+ */
+public class HistoricalBlockHashProcessor {
+
+  public static final Address HISTORICAL_BLOCKHASH_ADDRESS =
+      Address.fromHexString("0xfffffffffffffffffffffffffffffffffffffffe");
+
+  private static final long HISTORY_SAVE_WINDOW = 8192;
```
jflo, line 38: Suggest that we just reuse constant names from the EIP. s/HISTORY_SAVE_WINDOW/HISTORY_SERVE_WINDOW.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HistoricalBlockHashProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.tuweni.units.bigints.UInt256;
+
+/**
+ * Processes and stores historical block hashes in accordance with EIP-2935. This class is
+ * responsible for managing the storage of block hashes to support EIP-2935, which introduces
+ * historical block hash access in smart contracts.
+ */
+public class HistoricalBlockHashProcessor {
+
+  public static final Address HISTORICAL_BLOCKHASH_ADDRESS =
```
jflo, line 35: Suggest that we just reuse constant names from the EIP. EIP has been updated with actual address: 0x25a219378dad9b3503c8268c9ca836a52427a4fb

jflo (changes requested):
This is good, the requested changes are naming and contract addresses.

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
```diff
@@ -29,45 +32,93 @@
...
+    }
+  }
+
+  private Bytes32 readBlockHashFromState(final MessageFrame frame, final long soughtBlock) {
+    Hash blockHash =
+        Hash.wrap(
+            frame
+                .getWorldUpdater()
+                .get(HISTORY_STORAGE_ADDRESS)
+                .getStorageValue(UInt256.valueOf(soughtBlock % BLOCKHASH_OLD_WINDOW)));
```
jframe, line 107: I think we should be doing `soughtBlock % HISTORY_SERVE_WINDOW` not `soughtBlock % BLOCKHASH_OLD_WINDOW`

jflo: I think jason is right here, we may need to add a test where the block being sought is > BLOCKHASH_OLD_WINDOW but < HISTORY_SERVE_WINDOW

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HistoricalBlockHashProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+      final WorldUpdater worldUpdater,
+      final ProcessableBlockHeader currentBlockHeader) {
+
+    final MutableAccount account = worldUpdater.getOrCreate(HISTORY_STORAGE_ADDRESS);
+
+    if (currentBlockHeader.getNumber() > 0) {
+      storeParentHash(account, currentBlockHeader);
+
+      BlockHeader ancestor =
+          blockchain.getBlockHeader(currentBlockHeader.getParentHash()).orElseThrow();
```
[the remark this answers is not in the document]
matkt: normally it's not possible to try import a block without having the parent it will be a critical issue. maybe we need to test to be sure the behavior is correct in this case and be sure we are considering the block as problematic to allow a retry ?

matkt (commented):
It seems to be fine for me as it's similar to the verkle implementation that is currently working in a devnet.

We just need to be sure we are catching correctly the case where we don't have the parent header and try this feature with reference tests when it will be available

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
matkt: The last information I had was that this EIP was not supposed to serve more than 256 block hashes with the block hash operation, but this may have changed and needs to be verified. @Gabriel-Trintinalia

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HistoricalBlockHashProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+ * historical block hash access in smart contracts.
+ */
+public class HistoricalBlockHashProcessor {
+
+  public static final Address HISTORY_STORAGE_ADDRESS = BlockHashOperation.HISTORY_STORAGE_ADDRESS;
+
+  private static final long HISTORY_SERVE_WINDOW = 8192;
+
+  private final long forkTimestamp;
+  private final long historySaveWindow;
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: I don't think we should use `DEFAULT_HISTORY_SERVE_WINDOW` here. Default implies that this can have a different value, so we have a default. The value cannot be changed

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
Gabriel-Trintinalia: `BLOCKHASH_OLD_WINDOW` is the serving window pre-prague.  `HISTORY_SERVE_WINDOW`  is the new serving window. There was an error in that part. Fixed now

jflo (approved):
great work!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HistoricalBlockHashProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.tuweni.units.bigints.UInt256;
+
+/**
+ * Processes and stores historical block hashes in accordance with EIP-2935. This class is
+ * responsible for managing the storage of block hashes to support EIP-2935, which introduces
+ * historical block hash access in smart contracts.
+ */
+public class HistoricalBlockHashProcessor {
+
+  public static final Address HISTORY_STORAGE_ADDRESS = BlockHashOperation.HISTORY_STORAGE_ADDRESS;
```
jframe, line 38: nit: Why not just use the constant from `BlockHashOperation` directly? It's already public

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
```diff
@@ -29,45 +32,105 @@
...
     return successResponse;
   }
+
+  private boolean isBlockWithinServeWindow(final long soughtBlock, final long currentBlockNumber) {
+
+    final long historyServeWindow =
+        isServingFromState() ? HISTORY_SERVE_WINDOW : BLOCKHASH_OLD_WINDOW;
+
+    return soughtBlock >= Math.max(currentBlockNumber - historyServeWindow, 0)
+        && soughtBlock < currentBlockNumber;
```
jframe, line 98: The serve window should include the currentBlockNumber as we used to do that from what I can see of the previous code. So this should be changed from `soughtBlock < currentBlockNumber` to `soughtBlock <= currentBlockNumber` reflecting the negation of the previous condition `soughtBlock > mostRecentBlockNumber`.

jframe (changes requested):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/HistoricalBlockHashProcessor.java
jflo: Because it is not available from some places it is needed, like MainnetProtocolSpecs, without introducing a circular dependency.

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
```diff
@@ -29,45 +32,105 @@
 /** The Block hash operation. */
 public class BlockHashOperation extends AbstractFixedCostOperation {
 
-  private static final int MAX_RELATIVE_BLOCK = 255;
+  /** The HISTORY_STORAGE_ADDRESS */
```
shemnon, line 35: I really don't like this implementation, as it forces design decisions onto downstream users of the library.

This logic should live in a new BlockHashLookup function that is wired in in the mainnet protocol schedule, rather than re-writing the operation based on a flag.

shemnon (changes requested):
EVM implementation will cause problems with downstream users of the library.  It should be refactored so the logic remains within the Block Hash Lookup function passed into the EVM, rather than bleeding into block hash operation logic.  This is why the function was created.

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
jflo: current block is supposed to be a zero hash return, per the yellowpaper.

jframe: yep your right it is defined that way

evm/src/main/java/org/hyperledger/besu/evm/operation/BlockHashOperation.java
shemnon: Here's a WIP of how I think it should work - https://github.com/hyperledger/besu/compare/main...shemnon:besu:eip-2935-danno

* Move all block hash lookup out of the operation, the operation only delegates to the data rpovider
* the data provider is part of the protocol spec
* merge in the beacon and block hash processing into the same class
* MainnetProtocolSpecs is where it all gets loaded in.

shemnon:
Changes are needed, it damages the utility of the standalone EVM library.  Here's my WIP to retain this utility - https://github.com/hyperledger/besu/compare/main...shemnon:besu:eip-2935-danno

shemnon:
Here's a more fully worked version: https://github.com/hyperledger/besu/pull/7080
I'll clean up the FIXMEs tomorrow.  Acceptance and reference tests worked locally.

shemnon:
PR #7080 is ready to review.  It includes the changes in this PR but also containing the changes the EVM needs to remain a standalone library. @Gabriel-Trintinalia

Gabriel-Trintinalia:
Closed in favour of #7080

PR #7011, opened 2024-04-29
Run apt-get in docker builds without using any cache. 

### LINKED ISSUE #7009: Docker build fails because of transparent caching problems
When executing `./gradlew distDocker` it can fail if a transparent cache contains faulty data. The output of one such example can be seen here where `apt-get instal fails because of this`:

```
Get:6 http://ports.ubuntu.com/ubuntu-ports mantic-updates/main arm64 libexpat1 arm64 2.5.0-2ubuntu0.1 [72.9 kB]
Fetched 70.5 MB in 35s (2015 kB/s)
E: Failed to fetch http://ports.ubuntu.com/ubuntu-ports/pool/main/d/dbus/libdbus-1-3_1.14.10-1ubuntu1_arm64.deb  Hash Sum mismatch
   Hashes of expected file:
    - SHA512:30bae536562368b33aff85ebfacf3ccc6546f4842a11687d5c4b33ab1192310bdd2678e988aaaf549b2a355bbd175596c813f043e9ad91d59da0df1d256f6ff2
    - SHA256:17824410384cb3cf3f3f8453df4d619d6e2b09d53a17ddd17227d720a92c3754
    - SHA1:d4a9de66cd87c1601e8bfccbf2eb8fdcc01663d1 [weak]
    - MD5Sum:28c01bfffb282cbbb7e0923a700ab39f [weak]
    - Filesize:207670 [weak]
   Hashes of received file:
    - SHA512:e4b0231445ecc92b02332c05df33894919109f773f4d8bce4c75cce6c5ba3758acc8797dc0d9065109f9cfc5c905d3f3489a8e67c611634e65e1f63b935f35ce
    - SHA256:3e02aecd59db347d6f6d37dbf112066e69cdce46bbb42e64abd0e812f05fef04
    - SHA1:4273c162aee6d9edf230c318ae297fa021041d3b [weak]
    - MD5Sum:554fe1c38c0a3c4a3b72b1164b3e40cc [weak]
    - Filesize:207670 [weak]
   Last modification reported: Fri, 15 Sep 2023 13:15:15 +0000
E: Failed to fetch http://ports.ubuntu.com/ubuntu-ports/pool/main/k/krb5/libkrb5-3_1.20.1-3ubuntu1_arm64.deb  Hash Sum mismatch
   Hashes of expected file:
    - SHA512:33bcf16f850c5a27465d87c8e934b5a43a3b66ba56a2435c47d603e493f0ca7c88b457ea762f8dcf4f0ed15b4141863f7682ce8bc4d3b2dee541ff7f03e85fb4
    - SHA256:e3784e42e7db7c9a7564e35dbed0a4efcfa8fb5212ab7cd57f5fe5a6d177839d
    - SHA1:cf1f157bbbd69da692f10773ef8ae5464351e828 [weak]
    - MD5Sum:265a74275a2239a500cbd35832d5ccb3 [weak]
    - Filesize:345166 [weak]
   Hashes of received file:
    - SHA512:cfd51bc6b6e6eff56b5d900e88fc4447db8c752439b34d5aec0e2e4191260d94692635f702cb92100e953cd0ae8e2dabb0e9b0de6b714b5436e5f93a51bcca9d
    - SHA256:e668220906b7f23f61fe5a6b4e1edb7688e833258ea1d0978d4cd2fda3c6cc5a
    - SHA1:706454689416a9ab86879a52a726a9cd87b9dad5 [weak]
    - MD5Sum:d02154045be6a547822f9ceeab2c3899 [weak]
    - Filesize:345166 [weak]
   Last modification reported: Thu, 24 Aug 2023 19:09:36 +0000
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
/bin/sh: 1: adduser: not found
```

To avoid this `apt-get` needs to be set to avoid any transparent caching from an ISP or other network nodes in between.

daniellehrner:
## PR description

The  docker build step can fail if `apt-get` gets it's dependencies from a transparent cache (e.g. from an ISP) and this cache serves corrupted or outdated packages. This PR clears all the apt-get cache first and afterwards executes the next steps while making sure that no cache whatsoever is used.

I have further split the the `RUN` command into several steps to make debugging easier, because like that the exact command that is failing while be shown. Right now we have several commands chained together and it's not always clear which of the several commands was failing.

## Fixed Issue(s)
fixes #7009 


fab-10 (commented):
Which is the impact, of splitting the `RUN` command, in term of bigger image size?

docker/Dockerfile
```diff
@@ -2,16 +2,27 @@
...
- # Ubuntu 23.10 comes with an "ubuntu" user with uid 1000. We need 1000 for besu.
- userdel ubuntu 2>/dev/null || true && rm -rf /home/ubuntu && \
- # Ensure we use a stable UID for besu, as file permissions are tied to UIDs.
- adduser --uid 1000 --disabled-password --gecos "" --home /opt/besu besu && \
-    chown besu:besu /opt/besu && \
-    chmod 0755 /opt/besu
+# Clean apt cache
+RUN apt-get clean
+RUN rm -rf /var/cache/apt/archives/* /var/cache/apt/archives/partial/*
+RUN rm -rf /var/lib/apt/lists/*
```
fab-10, line 8: are these really required? we should assume that the official Ubuntu base image has not corrupted caches

docker/Dockerfile
```diff
@@ -2,16 +2,27 @@
...
+
+# Update and install dependencies without using any cache
+RUN apt-get update -o Acquire::BrokenProxy="true" \
```
fab-10, line 11: is it possible to set these apt options as env vars?

docker/Dockerfile
```diff
@@ -1,17 +1,21 @@
...
- # Ensure we use a stable UID for besu, as file permissions are tied to UIDs.
- adduser --uid 1000 --disabled-password --gecos "" --home /opt/besu besu && \
-    chown besu:besu /opt/besu && \
-    chmod 0755 /opt/besu
+ENV NO_PROXY_CACHE="-o Acquire::BrokenProxy=true -o Acquire::http::No-Cache=true -o Acquire::http::Pipeline-Depth=0"
+
+# Clean apt cache
+RUN apt-get clean  && \
+  rm -rf /var/cache/apt/archives/* /var/cache/apt/archives/partial/*  && \
+  rm -rf /var/lib/apt/lists/*  && \
```
fab-10, line 9: since we should assume that the base ubuntu image is good, and the following commands have the cache disable, the cleanup could be moved at the end of the command list, so the image will be smaller.

daniellehrner:
> Which is the impact, of splitting the `RUN` command, in term of bigger image size?

@fab-10 I misunderstood how docker images are built and did not realize the impact in size of several run commands. I've reverted the change and use only one run command to save disk size.

docker/Dockerfile
daniellehrner: You are right I've changed it to do it after the installation

docker/Dockerfile
daniellehrner: Good idea, I did that

docker/Dockerfile
daniellehrner: I've moved it to after the installation. It was already there before, I assume to save disk size

PR #7012, opened 2024-04-29
Refactor genesis config file and options

fab-10:
## PR description

While working on optimizing the performance of the startup phase, specifically in regard to genesis file management, I found that there is a bit of tech debt accumulated in the way we manage genesis file and options, that makes it difficult to work with it, and makes the code fragile, so before posting the optimizations, this PR refactors the genesis file management to make it more uniform and also make sure that overrides are always applied, before trying to get any option.

Most of the changed files are related to renaming of change of types, while the core of the refactor is in `BesuCommand`, `BesuController` and `BesuControllerBuilder`.

In `BesuCommand` the genesis file and options are now provided by a supplier, that loads, and caches, them on first request, applying any CLI overrides at load time, and this simplify the workflow, since there is no more ad-hoc code to apply them in different places. Since `GenesisConfigOptions` are always accessible via `GenesisConfigFile` then we can just pass this to the `BesuControllor` simplifying the builder.


jflo (approved):
Overall this is an improvement. My only notes are with regard to the object model; I find extending/overriding/mutating the GenesisConfigFile with GenesisConfigOptions to be slightly confusing. I think readers may find it more clear to introduce a GenesisConfiguration class that combines GenesisConfigFile and GenesisConfigOptions.

This is an aesthetic opinion, loosely held.

config/src/test/java/org/hyperledger/besu/config/GenesisConfigFileTest.java
```diff
@@ -59,7 +59,7 @@ void shouldLoadMainnetConfigFile() {
 
   @Test
   void shouldLoadDevelopmentConfigFile() {
-    final GenesisConfigFile config = GenesisConfigFile.development();
+    final GenesisConfigFile config = GenesisConfigFile.fromResource("/dev.json");
```
garyschulte, line 62: this is less discoverable and brittle - wrapping the resource location with a reusable and easier-to-discover method name is cleaner IMO

fab-10: I removed the `GenesisConfigFile.development()` method since it is only meant for testing, and so does not fit well being part of production code.

fab-10:
> Overall this is an improvement. My only notes are with regard to the object model; I find extending/overriding/mutating the GenesisConfigFile with GenesisConfigOptions to be slightly confusing. I think readers may find it more clear to introduce a GenesisConfiguration class that combines GenesisConfigFile and GenesisConfigOptions.
> 
> This is an aesthetic opinion, loosely held.

Indeed this was also confusing me at the beginning, and I was thinking of introducing a single class, but then saw that `GenesisConfigFile` is more related to building the genesis block, while `GenesisConfigOptions` is related to consensus and protocol schedule related stuff, so I kept them separated, but I agree that better naming or further refactoring could be done to make it clearer.

PR #7042, opened 2024-05-03
Count all exceptions as retries when getting the range headers

pinges:
## PR description

In a previous PR we made the change to disconnect the peer that was selected to provide the range headers that we need as input for the chain downloading pipeline for useless responses after a number of retires. To make sure that these useless peers are disconnected this PR increments the retry count for any exception that happens while trying to get these headers.

In addition this PR throws an exception after disconnecting the useless peer so the pipeline is immediately  restarted with a new target peer.

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/SyncTargetRangeSource.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/SyncTargetRangeSource.java
@@ -31,6 +31,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


jframe (commented):
Changes make sense. Just need to see if it improves the chain halt issue by testing on some main nodes.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/SyncTargetRangeSource.java
```diff
@@ -150,13 +150,16 @@ private SyncTargetRange getRangeFromPendingRequest() {
       this.pendingRequests = Optional.empty();
       if (newHeaders.isEmpty()) {
         retryCount++;
-        if (retryCount >= retriesPermitted) {
+        if (retryCount > retriesPermitted) {
```
macfarla, line 153: why strictly > here

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/range/SyncTargetRangeSource.java
```diff
@@ -150,13 +150,16 @@ private SyncTargetRange getRangeFromPendingRequest() {
...
           LOG.atDebug()
               .setMessage(
-                  "Disconnecting target peer for providing useless or empty range header: {}.")
+                  "Disconnecting target peer for not providing useful range headers after {} retries: {}.")
+              .addArgument(retriesPermitted)
```
macfarla, line 157: since retryCount > retriesPermitted

should this log be retryCount

PR #7045, opened 2024-05-03
handle RPC execution error

### LINKED ISSUE #3603: Unhandled exception error with no details
I've noticed some logs like this:
2022-03-17 23:32:34.706+00:00 | vert.x-eventloop-thread-2 | ERROR | RoutingContext | Unhandled exception in router

There's no error details and can be confusing to node operators (is it critical or not? how do we solve it?)

Unclear how to replicate at this stage we might wanna look in where the exception is logged instead.

macfarla:
## PR description
* call handleJsonRpcError in one extra spot
* and some extra logging about where the error occurred

## Fixed Issue(s)
fixes #3603 


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcExecutorHandler.java
```diff
@@ -58,11 +58,13 @@ public static Handler<RoutingContext> handler(
                   } catch (IOException e) {
                     final String method = executor.getRpcMethodName(ctx);
                     LOG.error("{} - Error streaming JSON-RPC response", method, e);
-                    throw new RuntimeException(e);
+                    handleJsonRpcError(ctx, null, RpcErrorType.INTERNAL_ERROR);
```
garyschulte, line 61: This seems like a good practice, but afaict IOException is going to be thrown by execute in the same thread and the runtime exception would be handled by the outer try/catch.  So I don't think it is strictly necessary.

The change seems ok, but I am curious if Devin gave any justification

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcExecutorHandler.java
```diff
@@ -58,11 +58,13 @@ public static Handler<RoutingContext> handler(
...
                   }
                 },
                 () -> handleJsonRpcError(ctx, null, RpcErrorType.PARSE_ERROR));
       } catch (final RuntimeException e) {
+        final String method = ctx.get(ContextKey.REQUEST_BODY_AS_JSON_OBJECT.name());
+        LOG.error("Unhandled exception in JSON-RPC executor for method {}", method, e);
```
macfarla, line 67: this change will log the method that was called so we're getting extra context

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcExecutorHandler.java
macfarla: think you're right on this one, we're not strictly gaining anything except uniformity in the code

macfarla: from Devin - It replaces throwing a new RuntimeException with a call to a method named handleJsonRpcError, which includes the error message from the exception.

usmansaleem (approved):
LGTM

PR #7061, opened 2024-05-07
build: Add default constructor and javadoc for :besu

usmansaleem:
## PR description
build: Add default constructor and javadoc for Java 21 for besu subproject. In order for build to pass javadoc lint, default constructor and their javadoc needs to be added. 


besu/src/main/java/org/hyperledger/besu/cli/options/stable/NodePrivateKeyFileOption.java
```diff
@@ -22,6 +22,8 @@
 
 /** The Node private key file Cli option. */
 public class NodePrivateKeyFileOption {
+  /** Default constructor. */
+  NodePrivateKeyFileOption() {}
```
siladu, line 26: Why are some of them package-private?

siladu (commented):
Wondering how the public vs package-private decision was made?

Also wondering if any of these should be private constructors, e.g. in case of static factories? I haven't checked if there are any static factories here, I guess I'm asking if you have :)

besu/src/main/java/org/hyperledger/besu/cli/options/stable/NodePrivateKeyFileOption.java
usmansaleem: The private constructors should be "injected" in factories which only has static methods.

usmansaleem: In other words, if I am confident that the constructor should be declared private, I declare it as such. Otherwise if its an abstract class, protected constructor makes sense. package level access constructors are suitable where those constructors are utilized in picocli/unit tests.

macfarla (approved):

PR #7079, opened 2024-05-08
Split unit tests by time

fab-10:
## PR description

This PR reduces the time of the `pre-review` workflow from ~30min to <20min, splitting unit tests by time, and removing the dependency between `compile` and `unit tests` since the Gradle cache is disabled and so test tasks need to recompile anyway.
Also reduced the number of ATs runners to 10 since the workflow takes only ~10min to complete.
Next step is to see if reference tests can be split by time as well.

### DIFF
--- a/.github/workflows/pre-review.yml
+++ b/.github/workflows/pre-review.yml
@@ -118,13 +118,13 @@ jobs:
         uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3
         if: matrix.runner_index == 0
         with:
-          name: acceptance-tests-timing
+          name: unit-tests-timing
           path: 'tmp/timing.tsv'
       - name: Upload Lists
         uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3
         if: matrix.runner_index == 0
         with:
-          name: acceptance-tests-lists
+          name: unit-tests-lists
           path: 'tmp/*.list'
       - name: Upload gradle test tasks
         uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3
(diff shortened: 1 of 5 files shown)


.github/workflows/pre-review.yml
```diff
@@ -116,22 +94,54 @@ jobs:
...
+          name: 'unit-.*-test-results'
+          path: tmp/junit-xml-reports-downloaded
+          if_no_artifact_found: true
+      - name: Split tests
+        run: .github/workflows/splitTestsByTime.sh tmp/junit-xml-reports-downloaded "tmp/junit-xml-reports-downloaded/unit-.*-test-results" "build/test-results" ${{env.total-runners}} ${{ matrix.runner_index }} > testList.txt
+      - name: Upload Timing
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3
+        if: matrix.runner_index == 0
+        with:
+          name: acceptance-tests-timing
```
macfarla, line 121:
```suggestion
          name: unit-tests-timing
```

.github/workflows/pre-review.yml
```diff
@@ -116,22 +94,54 @@ jobs:
...
+          path: 'tmp/timing.tsv'
+      - name: Upload Lists
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3
+        if: matrix.runner_index == 0
+        with:
+          name: acceptance-tests-lists
```
macfarla, line 127:
```suggestion
          name: unit-tests-lists
```

macfarla (approved):
assuming this doesn't change how failures are presented?

fab-10:
> assuming this doesn't change how failures are presented?

there should be no difference in that

PR #7089, opened 2024-05-10
Peering - disconnect peers that have multiple discovery ports

### LINKED ISSUE #6805: Low Peer Numbers
Multiple reports of peering issues, this is a placeholder to track and gather info.

Anecdotally seems like issues began after Dencun fork on March 13th and are worse during initial sync.

* @garyschulte suspected a Dencun fork id issue on March 22nd related to geth during snap server testing...it ultimately turned out to be an old version of geth, but there were still a lack geth peers on our canary nodes.

* Other devs internally have noticed issues connecting to geth peers while trying to sync Holesky.
  * > This node syncs basically using 2 Geth nodes and disconnects all non Geth nodes, because they are useless. After the sync, when we stop sending snap requests we do find peers, but out of a max of 21, I have 2 Geth nodes, 1 reth node, 1 besu node, and the rest is nethermind. Something is not right here!

* Issue reported on geth https://github.com/ethereum/go-ethereum/issues/29312

### User reports from EthStaker #besu-helpdesk Discord
 
* GGubignou — 17/03/2024 19:51
> Hi, I updated to the latest Besu via eth-docker and it now seems to struggle finding peers. I was previously at 25 peers constantly but now it oscillates between 8-13 peers. Has anything been changed with regards to port forwarding in the latest version?
> (FYI my consensus client Teku finds peers as usual and I have otherwise not changed anything to my setup) 
  * GGubignou — 18/03/2024 16:49
> Hi @siladu, after a couple hours of struggle it eventually managed to go back to 25 peers and it seems fine now. I upgrade frequently so I was likely on the previous release before that.
  * ktmkancha — 23/03/2024 01:05
> Hi Siladu- I've been having trouble with Peer connections. I hadn't updated my node prior to the Dencun fork for few months so I was probably few updates behind. I have been hovering around 9-11 peers in the last few hours. It hasn't really climbed up or down. Any advice? 
  * ktmkancha — 26/03/2024 05:28
> Since the Dencun update, my peer count has been single digits. I used to get consistently 25 peers prior to this, any advise? Thank you
  * ktmkancha — 28/03/2024 02:27
> Hi @siladu Just wanted to report that after the besu node sync up with the top of the chain- the peer count is back up to 15+.  It's better than single digits, but compared to dencun update- it's still 10 less peers. I used to get 25 peers consistenly. Will keep an eye out of the any besu updates! thanks
  * MrPrezident — 24/03/2024 03:08
> I noticed that while syncing besu, it is consistently using only 1 peer, whereas teku currently has 24 peers. Is that because I'm doing a checkpoint sync?
  * MrPrezident — 24/03/2024 06:32
 > teku is connected to 28 peers, so I don't think it is having any problems. Only besu is having issues with peers. Both teku and besu are being run on the same host using ethdocker
The only firewall should be my pfsense router.
but I have portforwarding turned on in pfsense
  * MrPrezident — 24/03/2024 14:26
> Just for testing purposes, I reconfigured my ethdocker to use geth/prysm and restarted syncing to see if it would have the same problem as besu/teku. It did finish syncing on both execution and consensus clients, but I noticed that the consensus client (prysm) is hovering at about 70 peers and the execution client (geth) is hovering at about 5 peers. That seems like a huge discrepancy. That is sort of in line with what I was seeing with besu/teku. Besu was only getting 1 or 2 peers vs about 20 peers with Teku. Is this normal to have less peers with the execution client? How many peers should I expect to see with the execution client?
  * MrPrezident — Yesterday at 00:51
> This morning I looked at my dashboard and now I see 33 execution peers and 72 consensus peers (this is with geth/prysm). I didn't change anything. I just let it run overnight. I assume this is a normal range? It would be good to know how many peers I should expect. (this is on holesky)
  * scott — 26/03/2024 05:32
> Also happening here, seems to be a known issue with an unknown root cause right now.
  * All Might — 26/03/2024 06:01
> Same here, my peers have been at 5 the entire day. Has happened several times in the last 5 days and I don't know how to fix it
  * RemiRaznor — 28/03/2024 00:01
> I'm having the same issue with low to no peer count. I'm trying to rebuild my Besu database after switching from goerli to holesky. For 2 days, I have gotten "Unable to find sync target. Currently checking 0 peers for usefulness" messages. Once or twice it found 1 or 2 peers enough for the download progress to be just 13.49%. I have restarted besu 4 or 5 times in the last 2 days with no success.   This validator was working fine in goerli. All ports are open. 
  * RemiRaznor — 28/03/2024 06:10
> I got it to sync finally by adding the --p2p-port=30303  card to the Besu service. After restarting the EL, I was able to connect with one peer just enough to get the download started again and then I got one more peer which got me to 57% and then I lost the peers again and stalled. I started adding the other ports, rpc port, engine port and metrics port one by one until after the metrics port the sync started up again.  Not sure if this will help but good luck...

pinges:
## PR description
In discovery we seemed to see a lot of "misconfigured" nodes that share the same IP address and TCP port, but use different discovery ports. This PR checks for these nodes and removes them from the peer table.

Fixes #6805


macfarla:
very cool find @pinges - are there more tests you wanted to add?

macfarla:
11 holesky nodes - 10 of them get to 100% peers within 2h
Screenshot 2024-05-10 at 2 43 23 PM

the 11th one eventually gets to 100% peers (although it does take ~7h)
Screenshot 2024-05-10 at 2 44 21 PM

Current behaviour on holesky is > 90% of nodes get stuck at 0 peers indefinitely (require setting static peers to sync) 

See #6805

macfarla (commented):
needs a changelog entry

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
```diff
@@ -51,6 +54,10 @@ public class PeerTable {
   private final Map<Bytes, Integer> distanceCache;
   private BloomFilter<Bytes> idBloom;
   private int evictionCnt = 0;
+  private final LHMWithMaxSize<String, Integer> ipAddressCheckMap =
+      new LHMWithMaxSize<>(DEFAULT_BUCKET_SIZE * N_BUCKETS);
+  private final CircularFifoQueue<String> invalidIPs =
+      new CircularFifoQueue<>(DEFAULT_BUCKET_SIZE * N_BUCKETS);
```
macfarla, line 60: curious - why these specific collection types?

macfarla (approved):
just needs a changelog entry

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
```diff
@@ -51,6 +54,10 @@ public class PeerTable {
   private final Map<Bytes, Integer> distanceCache;
   private BloomFilter<Bytes> idBloom;
   private int evictionCnt = 0;
+  private final LHMWithMaxSize<String, Integer> ipAddressCheckMap =
```
lucassaldanha, line 57: I think we should rename `LHMWithMaxSize` as `LinkedHashMapWithMaxSize` to be clearer.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
pinges: CircularFifo is a simple queue with a maximum size.
LHMWithMaxSize is a private class, it is a LinkedHashMap that I have extended to have a maximum size.
Lukas complained about the name of the class and I will rename it :-)

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerTable.java
pinges: Will do :-)

PR #7101, opened 2024-05-14
Small optimization on eth_sendRawTransaction

ahamlat:
## PR description
This PR is a small improvement on eth_sendRawTransaction, as it notifies the listeners on added transactions asynchronously. 
The PR also pre-calculates the sender of the transaction which is more a mitigation of the synchronized on `LayeredPendingTransactions.addTransaction`, the sender can be calculated by one the thread of ForkjoinPool.commonPool while the main thread is locking on syncronized.

image


fab-10 (approved):
LGTM both changes make sense, so for me it is fine to ship.
I will just keep thinking if the asynchronous notification, could potentially create a race condition in case a tx is added and suddenly removed, and the remove notification can be delivered before the added one, not sure this can happen before a review better that part, and in case we can introduce a queue for notifications so we are sure they are delivered in order.

PR #7102, opened 2024-05-14
Make `eth_gasPrice` aware of the base fee market

fab-10:
## PR description

The `eth_gasPrice` is not taking in account the base fee market values, when the network supports them, so it is possible that it returns values that are not good for pricing a tx for inclusion in the next block.
Specifically the calculation of gas price lower bound only rely on the runtime value of `minGasPrice`, but on base fee market networks, it could be that `baseFee` for the next block is higher than the `minGasPrice`, for example for devnets using the default, the `minGasPrice=1kwei` and the genesis sets `baseFee=1gwei` and the result is that senders thinks the can just price the gas at 1kwei to the the tx included, which is not the case.

This PR takes in consideration the baseFee value when calculating the lower bound, along with the fix I did some refactoring to move the gas price calculation into `BlockchainQueries` and continue to directly use `MiningParameters` to get the gas market info.

### DIFF
--- a/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGasPriceTest.java
+++ b/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGasPriceTest.java
@@ -41,6 +41,7 @@ import org.hyperledger.besu.ethereum.core.MiningParameters;
 import org.hyperledger.besu.ethereum.core.Transaction;
 import org.hyperledger.besu.ethereum.mainnet.ProtocolSchedule;
 import org.hyperledger.besu.ethereum.mainnet.ProtocolSpec;
+import org.hyperledger.besu.ethereum.mainnet.feemarket.CancunFeeMarket;
 import org.hyperledger.besu.ethereum.mainnet.feemarket.LondonFeeMarket;
 import org.hyperledger.besu.evm.log.LogsBloomFilter;
 
@@ -63,6 +64,8 @@ public class EthGasPriceTest {
   private static final String ETH_METHOD = "eth_gasPrice";
   private static final long DEFAULT_BLOCK_GAS_LIMIT = 100_000;
   private static final long DEFAULT_BLOCK_GAS_USED = 21_000;
+  private static final Wei DEFAULT_MIN_GAS_PRICE = Wei.of(1_000);
+  private static final Wei DEFAULT_BASE_FEE = Wei.of(100_000);
 
   @Mock private ProtocolSchedule protocolSchedule;
   @Mock private Blockchain blockchain;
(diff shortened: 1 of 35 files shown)


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGasPriceTest.java
```diff
@@ -163,104 +192,142 @@ public void shouldReturnLimitedPriceWhenUpperBoundIsPresent() {
...
+    long upperBoundCoefficient = 200;
+    verifyGasPriceLimit(lowerBoundCoefficient, upperBoundCoefficient, gasPrice);
   }
 
+  //  @Test
+  //  public void shouldReturnConfiguredGasPriceLowerBoundAtGenesis() {
+  //    mockBaseFeeMarket();
+  //    mockBlockchain(0, 0);
+  //
+  //  }
```
macfarla, line 206: remove if not needed

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGasPriceTest.java
```diff
@@ -163,104 +192,142 @@ public void shouldReturnLimitedPriceWhenUpperBoundIsPresent() {
...
             request.getRequest().getId(), Wei.of(expectedGasPrice).toShortHexString());
 
-    when(blockchain.getChainHeadBlockNumber()).thenReturn(10L);
-    when(blockchain.getBlockByNumber(anyLong()))
-        .thenAnswer(invocation -> createFakeBlock(invocation.getArgument(0, Long.class)));
+    final var chainHeadBlockNumber = 10L;
+    mockBlockchain(chainHeadBlockNumber, 1);
+
+    // baseFee for next block = 0x7c61 (31_841)
+    // BlockchainQueries returns gasPrice = 0x5b8d80 (60_000)
```
macfarla, line 237: remove if not needed

macfarla (commented):
some commented out code hanging around, overall looks good

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGasPriceTest.java
fab-10: done

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGasPriceTest.java
fab-10: done

macfarla (approved):

Issue #7036, opened 2024-05-02
BAD Block in private QBFT chain


cryptohazard:
### Description
We ran a private chain of 4 nodes, with QBFT and all the milestones enabled. We added Shanghai once we migrated to 24.3.0.
One node, which is our JSON-RPC node fell out of sync. It showed a `BAD BLOCK` error and from there we couldn't make it join the network any more (we ran into #6760 then).

My question is: do you see a way a heavy (relatively) load on the JSON-RPC could end up with bad blocks? 
Looking through other issues with bad blocks here, I'm pretty unsure how we caused it.

MASDXI:
can you provide your configuration, did you use bonsai data structure?

macfarla:
@cryptohazard is this the same as your issue https://github.com/hyperledger/besu/issues/6680

cryptohazard:
> @cryptohazard is this the same as your issue #6680

 No it doesn't seem so. 

I will run some tests and share the errors I'm getting (annoyingly, I can't copy/paste everything due to work policy so I have to adapt/clean the logs, config...)

cryptohazard:
> can you provide your configuration, did you use bonsai data structure?

Forest, since it was the old default and our chain was already up before you changed the default to Bonsai.

cryptohazard:
the rroe:
```
WARN withdrawlsRoot must not be null when Withdrawals are activate
INFO Invalid block
```
I can also say that updating to the current version `24.5.1`after activating Shanghai on a previous version (`24.3.0`) doesn't fix the issue.

macfarla:
@matthew1001 
WARN withdrawlsRoot must not be null when Withdrawals are activate
INFO Invalid block
on QBFT, shanghai, 24.5.1 - any insights on this one?

matthew1001:
Yeah I'm just having a look now. There was a fix in `24.5.1` that is related to this area (https://github.com/hyperledger/besu/pull/6765) but I think it may not be complete enough. Going to look at this issue today.

matthew1001:
I've raised a PR which resolves the issue (https://github.com/hyperledger/besu/pull/7115). Tested by syncing `24.5.1` + fix against an existing `24.3.0` `shanghai` chain. Without the fix I hit the issue described. With the fix the node syncs and then continues mining blocks.

@cryptohazard do you have a Besu build environment? If you do and are able to test out https://github.com/matthew1001/besu/tree/qbft-ignore-withdrawals please do.

[...]

PR #7115, opened 2024-05-16
BFT ignore withdrawals

matthew1001:
### Short version:

Skips validation of `withdrawals` when importing BFT blocks since withdrawals don't apply to BFT chains

### Long version:

Unfortunately the original shanghai PR https://github.com/hyperledger/besu/pull/6353 which was included in `24.3.0` had a bug that caused block `withdrawals` to be set incorrectly for new `shanghai` blocks. Withdrawals are not applicable to BFT chains, but the block header still has a `withdrawals` field which is validated whenever a block is imported.

I fixed the issue in PR https://github.com/hyperledger/besu/pull/6765 which was released in `24.5.1`. However, any chain with `shanghai` enabled in `24.3.0` will have already created incorrect blocks, which `24.5.1` will now fail to import as the validation is now correct - but the created blocks are not.

Without requiring users to restore to a point before `24.3.0` blocks were produced, I think the most pragmatic approach is to disable withdrawal validation on BFT blocks.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/7036


macfarla (approved):
neat

PR #7116, opened 2024-05-16
Fix parsing `gasLimit` parameter when its value is > `Long.MAX_VALUE`

fab-10:
## PR description

When calling `eth_estimateGas` with the `gas` field set to the maximum value `0xffffffffffffffff` or anything greater than `0x7fffffffffffffff`, the endpoint returns and error says parameters are invalid, due to the fact that we use a `long` to fix the value, and it cannot fit these values.
This is not the correct behavior, so the fix is to cap the value to Long.MAX_VALUE, in case the input field has a greater value, since in practice gas limit will not reach such value.
The deserializer has been rename to `GasDeserializer` since it is now specific to gas and not generic long values.


pinges (approved):
LGTM

PR #7121, opened 2024-05-17
Pectra-devnet-0 discoveries

jflo:
- corrects gas checking in AUTH / AUTHCALL
- fixes bug in engine api v4
- corrects gas pricing in BLOCKHASH

PR #7122, opened 2024-05-17
Fix name of parameter for WithdrawalRequest

lucassaldanha:
## PR description

The name of the parameter should be `validatorPublicKey`. This change https://github.com/hyperledger/besu/pull/7121/commits/7e78da631a24a55f990a58dfdcb7f63ca95b0267 tried to fix it but only made the change in the @JsonProperty in the constructor, that is used for deserialization.

To fix this, we need to fix it on the getter method with the @JsonGetter annotation. This is what is used for serialization and infers the name of the field from the name of the getter method. So `getValidatorPublicKey` becomes `validatorPublicKey` when serializing to json.

I have also removed the method `asJsonObject` because it is not being used.


lucassaldanha:
PS: this needs to be backported to a pectra branch.

macfarla (approved):

Issue #7064, opened 2024-05-07
No JSON/RPC endpoint exists to determine which EVM version a permissioned chain is using


matthew1001:
### Description
The `admin_nodeInfo` JSON/RPC method returns the configuration of a node which includes its fork schedule, but you have to manually derive which schedule is in use by correlating it with the chain's block height and/or timestamp.

It would be useful to have a JSON/RPC method that returned the EVM version in use for the `latest` block.

One additional enhancement would be the ability to specify a block number and get back the EVM version in use at that block height, but the use-case for that is less clear than for `latest`

### Acceptance Criteria
A JSON/RPC method exists to determine the EVM version in use on the current/latest block height.

PR #7127, opened 2024-05-21
Include currently active EVM version in `admin_nodeInfo` response

matthew1001:
## PR description
Adds a new field `activeFork` to the response body of `admin_nodeInfo` JSON/RPC requests.

The JSON/RPC call already returns the fork configuration but that requires the caller to make multiple calls and compare chain heights and/or timestamps to infer the currently active fork. The new field makes it possible to determine accurately in a single call.

The value is whatever `getName()` returns for the active `ProtocolSpec`. This means it is sentence cased (upper-case first letter) but a caller would be expected to use a case-insensitive string compare so I'm not inclined to upper or lower case the string.

Example output with the `shanghai` fork EVM enabled:

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "enode": "enode://beea7aa3a6e1f862870a456239dbf9a232d9a69e58edfeda92193f679bdbf3960bb7f41bc6f0d574c4d0c6baa2324c7279045405ef8a240906bc13d3145ccc14@127.0.0.1:30303",
        "listenAddr": "127.0.0.1:30303",
        "activeFork": "Shanghai",    <<-- NEW FIELD
        "ip": "127.0.0.1",
        "name": "besu/v24.5-develop-4753b28/osx-aarch_64/openjdk-java-17",
        "id": "beea7aa3a6e1f862870a456239dbf9a232d9a69e58edfeda92193f679bdbf3960bb7f41bc6f0d574c4d0c6baa2324c7279045405ef8a240906bc13d3145ccc14",
        "ports": {
            "discovery": 30303,
            "listener": 30303
        },
        "protocols": {
            "eth": {
                "config": {
                    "chainId": 1337,
                    "berlinBlock": 0,
                    "londonBlock": 0,
                    "shanghaiTime": 1716294470,
                    "qbft": {
                        "epochLength": 30000,
                        "blockPeriodSeconds": 2,
                        "requestTimeoutSeconds": 4
                    },
                    "zeroBaseFee": true
                },
                "difficulty": 62,
                "genesis": "0xf1b9d4b087f35660b73ef0ad56d0f241009c422459500f0a1e8f421ffbe5c29c",
                "head": "0x67ed5c441d3a2eba53c97b8f3b88ccb217ac17084ad0ed722fe894963a49ba6f",
                "network": 1337
            }
        }
    }
}
```

I've tested the call before and after a fork to check that the value of `activeFork` changes as soon as the new fork is enabled.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/7064


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AdminNodeInfo.java
```diff
@@ -73,6 +77,8 @@ public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
     final Map<String, Object> response = new HashMap<>();
     final Map<String, Integer> ports = new HashMap<>();
 
+    System.out.println("MRW: Returning node info");
```
macfarla, line 80: remove debugging

macfarla (commented):
needs a changelog entry and there's a stray println, otherwise LGTM

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AdminNodeInfo.java
matthew1001: Bah, thanks, will remove now and add the changelog entry.

matthew1001: Both now sorted @macfarla

macfarla (approved):

PR #7140, opened 2024-05-24
Add experimental `--Xsnapsync-bft-enabled` which enables snap sync for BFT chains

matthew1001:
## PR description

This PR adds various changes/fixes that allow snap-sync and BFT to be used together.

Since this is a relatively new combination of configuration I have added an experimental flag `--Xsnapsync-bft-enabled` which must be set in order for Besu to start.

My general opinion for enterprise chains is that it is probably reasonable for a BFT chain to have `sync-min-peers` set to `1`. The trust requirements of a public chain are not as applicable to an enterprise chain, so the need to have e.g. `5` peers before you are willing to consider their pivot-block data is too restrictive. At the very least it makes it impossible to have a chain of 4 validators, or a chain of 6 validators with one unavailable. I've broadly outlined below the `snap` sync logic that this PR uses, some of which assumes that an enterprise user will set `--sync-min-peers=1` if their chain only has a small number of nodes. I think there's an argument for the `--profile=enterprise` defaults to set `--sync-min-peers=1`, but I'm going to save that for a separate PR.

The basic logic used for sync is as follows:
1. If a node has a best peer (i.e. they have found a node with pivot data) but the peer's current block height < `--Xsynchronizer-fast-sync-pivot-distance` (`50` by default) it skips any attempt at `snap` sync. Quitting the snap sync process allows the node to start contributing to BFT voting which ensures the chain isn't stalled waiting for this node. (The state for the most recent 50 blocks is obtained using `full` sync, which is existing behaviour)
3. If a node enters `snap` sync but believes itself to be the only validator it exits `snap` sync. As above, this ensures that the node can then start its BFT mining coordinator and continue producing new blocks, rather than get stuck in a loop of failed `snap` sync attempts which it will probably never get out of. This is particularly useful for the case where a single BFT validator node is started as a new chain, to which other validators will be added later on. Without this logic it would wait for at least 1 other peer, which may never come along.
4. If a node has `sync-min-peers` peers who are validators, but none of them has any usable pivot data, it quits the `snap` sync process, again to allow its BFT mining coordinator to start producing new blocks. This is necessary in a case where a number of new validators have been created for a brand new chain, but because they are all new and have no state to share they would get stuck in a loop of failed `snap` sync attempts. 

(It's also worth noting that I've continued the process of renaming the fast-sync internal variables to be more generic as they also apply to `snap` and `checkpoint` sync)

(Note: see related PR https://github.com/hyperledger/besu/pull/7204 which fixes a follow on issue)

### Documentation

The broad topic of sync-ing is covered in the public section of the docs here: https://besu.hyperledger.org/development/public-networks/get-started/connect/sync-node

I think the private section of the docs would do with its own sync-ing topic, which refers to the article above for the basic "how do the different sync modes work" but explains the permissioned-chain specific behaviours that I've implemented in this PR. I think the explanation in the description above is probably a reasonable starting point to write those docs, but I'm happy to review and edit the new topic once an initial draft is there.


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/BFTPivotSelectorFromPeers.java
```diff
@@ -0,0 +1,105 @@
...
+
+      return bestPeer.flatMap(this::fromBestPeer);
+    } else {
+      // Treat single-validator as a special case. We are the only node that can produce
+      // blocks so we won't wait to sync with a non-validator node that may or may not exist
+      if (validatorProvider.getValidatorsAtHead().size() == 1
+          && validatorProvider
+              .getValidatorsAtHead()
+              .contains(Util.publicKeyToAddress(nodeKey.getPublicKey()))) {
+        throw new NoSyncRequiredException();
```
garyschulte, line 82: 👍

garyschulte (commented):
This LGTM.  It is probably worth having @matkt have a look also, but this behavior seems well contained for *bft.  We might want to push the "snap-sync minimum distance" up into PivotSelectorFromPeers.  That is a worthwhile distinction to make on any consensus mechanism imo

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/BFTPivotSelectorFromPeers.java
```diff
@@ -0,0 +1,105 @@
...
+    final ValidatorProvider validatorProvider = bftContext.getValidatorProvider();
+    // See if we have a best peer
+    Optional<EthPeer> bestPeer = selectBestPeer();
+
+    if (bestPeer.isPresent()) {
+      // For a recently created permissioned chain we can skip snap sync until we're past the
+      // pivot distance
+      if (bestPeer.get().chainState().getEstimatedHeight()
+          <= syncConfig.getFastSyncPivotDistance()) {
+        throw new NoSyncRequiredException();
```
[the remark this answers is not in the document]
matkt: just not sure why we need snapsync for a new chain. a fullsync will work well

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
```diff
@@ -80,7 +80,7 @@ public CompletableFuture<FastSyncState> start() {
     if (!running.compareAndSet(false, true)) {
       throw new IllegalStateException("SyncDownloader already running");
     }
-    LOG.info("Starting sync");
+    LOG.info("Starting fast sync");
```
matkt, line 83: it's not clean but this class is also used by snapsync. I think it's why we removed the fast

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
```diff
@@ -94,7 +94,7 @@ protected CompletableFuture<FastSyncState> start(final FastSyncState fastSyncSta
           onBonsai.clearFlatDatabase();
           onBonsai.clearTrieLog();
         });
-    LOG.debug("Start sync with initial sync state {}", fastSyncState);
+    LOG.info("Start fast sync with initial sync state {}", fastSyncState);
```
matkt, line 97: same here

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
```diff
@@ -114,15 +114,17 @@ public CompletableFuture<FastSyncState> findPivotBlock(
...
       return CompletableFuture.failedFuture(error);
     } else if (rootCause instanceof StalledDownloadException) {
-      LOG.debug("Stalled sync re-pivoting to newer block.");
+      LOG.info("Stalled sync re-pivoting to newer block.");
       return start(FastSyncState.EMPTY_SYNC_STATE);
     } else if (rootCause instanceof CancellationException) {
       return CompletableFuture.failedFuture(error);
     } else if (rootCause instanceof MaxRetriesReachedException) {
-      LOG.debug(
+      LOG.info(
```
matkt, line 127: are we sure this log is not spamming during the sync. maybe we need to start a sync on mainnet to be sure we are not used debug to avoid log spamming

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapSyncDownloader.java
```diff
@@ -48,7 +48,7 @@ public SnapSyncDownloader(
 
   @Override
   protected CompletableFuture<FastSyncState> start(final FastSyncState fastSyncState) {
-    LOG.debug("Start snap sync with initial sync state {}", fastSyncState);
+    LOG.info("Start snap sync with initial sync state {}", fastSyncState);
```
matkt, line 51: the same here we need to be sure this log is not spamming on mainnet sync

non-fungible-nelson:
Please test these changes with a Mainnet or Holesky / sepolia node to look for unintended side-effects as called out by Karim. Or insane logging or strange behavior. A Holesky snap sync should complete in a few hours and be considered an acceptance test for this PR.

matthew1001:
Getting back into this PR now. I have some refactoring planned but pleased to see that the general direction seems to be OK.

@matkt I think there are benefits of snap sync to permissioned chains in the same way as public chains. Adding a new node (validator or otherwise) to a permissioned chain will take a long time to sync if the chain has been running for a while.

I'll look at your comments on the log changes I made. I found it clearer to see what was going on and to match that to the code if the log entries were more specific to the class they were coming from.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -242,16 +243,24 @@ private CompletableFuture<Void> handleSyncResult(final FastSyncState result) {
...
-        .ifPresent(
-            blockHeader -> protocolContext.getWorldStateArchive().resetArchiveStateTo(blockHeader));
-    LOG.info(
-        "Sync completed successfully with pivot block {}",
-        result.getPivotBlockNumber().getAsLong());
-    pivotBlockSelector.close();
-    syncState.markInitialSyncPhaseAsDone();
+
+    if (result instanceof NoSyncRequiredState) {
+      LOG.info("Sync completed (no sync required)");
```
matkt, line 248: I just wonder in terms of readability if we say that the sync is finished and then we have a backward sync or a fullsync which starts. I don't know if we could slightly modify the log, but honestly I haven't yet found a word that would allow snap/fast that download directly the state to be qualified differently from the other syncs that execute the blocks

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -256,7 +256,7 @@ public long getPropagationManagerGetBlockTimeoutMillis() {
 
   public static class Builder {
     private SyncMode syncMode = SyncMode.FULL;
-    private int fastSyncMinimumPeerCount = DEFAULT_FAST_SYNC_MINIMUM_PEERS;
+    private int syncMinimumPeerCount = DEFAULT_SYNC_MINIMUM_PEERS;
```
matkt, line 259: If we change the name, shouldn't we also create a new flag without "fast" and the old one will be depreciated with the idea of ​​removing it later

matthew1001: Yeah the external name was actually changed a few months ago. This is just me tidying up the name of the internal variable to match the new `--sync-min-peers` flag

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DefaultSynchronizer.java
matthew1001: Yeah I completely agree - I've been struggling to come up with a term that refers to all of the non-full sync methods. It would be really useful if we could come up with a consistent term that we can use. "Checkpoint" feels like a good common term for it, but since there's a specific sync method called "checkpoint" we can't really use that. Maybe "pivot-based sync" is a a reasonable description of them all? That would make the above log message "Sync completed (no pivot-based sync required)" or similar. Perhaps to make it clear that no sync took place I should change it to "Sync ended (no pivot-based sync required)". I think "ended" is clearer that nothing happened, where as "completed" suggests something did happen.

non-fungible-nelson: Is checkpoint sync appropriate for BFT networks? It should work based on a specified checkpoint but users would probably use a full snap sync in this case? Maybe "no snap sync required" is still valuable. If we are attempting to deprecate fast, we can likely avoid that language.  

[This page](https://besu.hyperledger.org/development/public-networks/get-started/connect/sync-node#run-a-full-node) lays out some language we could try to mirror.

non-fungible-nelson:
Completed a Holesky snap-sync with no unintended consequences (with the flag set). Some logging noise for cleanup.

matthew1001:
Firstly many thanks to @matkt for the help getting the snap-server issues fixed when there is an empty chain with `0` accounts, and for all the guidance on the codebase in that snap-sync area. Really enjoyed the collaboration on it.

@garyschulte I think the PR is ready for re-review now that we've got to the bottom of issues with `0` account chains among other things.

@non-fungible-nelson can you confirm if your sync with `holesky` was successful and that the PR hasn't regressed anything in that area?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2240,6 +2242,16 @@ private MiningParameters getMiningParameters() {
...
+
+      final ImmutableDataStorageConfiguration.Builder theBuilder =
+          ImmutableDataStorageConfiguration.builder().from(dataStorageConfiguration);
+      if (syncMode == SyncMode.FULL
+          && dataStorageOptions.toDomainObject().getDataStorageFormat()
+              == DataStorageFormat.BONSAI) {
+        logger.warn("Forcing bonsai-limit-trie-logs-enabled to false, since sync mode is FULL");
+        theBuilder.bonsaiLimitTrieLogsEnabled(false);
+      }
+      dataStorageConfiguration = theBuilder.build();
```
garyschulte, line 2254: this looks like a potential merge/rebase artifact ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -222,8 +224,8 @@ public float getFastSyncFullValidationRate() {
     return fastSyncFullValidationRate;
   }
 
-  public int getFastSyncMinimumPeerCount() {
-    return fastSyncMinimumPeerCount;
+  public int getSyncMinimumPeerCount() {
+    return syncMinimumPeerCount;
```
garyschulte, line 228: this really is a "fast sync" minimum peer count though right?  there is no minimum peer count for full sync AFAIK

garyschulte (commented):
generally LGTM.  Some compilation errors, questions about renaming and potential merge artifacts related to trielog pruning.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
matthew1001: This change was intentional, although it is a change to another PR that recently merged (https://github.com/hyperledger/besu/pull/7193) which stopped Besu starting if `FULL` and `Bonsai` were enabled together. I've tidied up the var name I used and added a comment explaining that we're just disabling `bonsai-limit-trie-logs-enabled` if that combination of sync mode and DB is being used.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/SynchronizerConfiguration.java
matthew1001: A number of the `fast` sync options also apply to `snap` and `checkpoint` sync, but because `fast` came long first a number of the variables (and things like the CLI arg `--fast-sync-min-peers`) have `fast` in the name. We've slowly been removing `fast` from them (so for example, the CLI options is now `--sync-min-peers`) but you're right, they're don't apply to `full` sync. I think it's OK that we remove `fast` to avoid confusion with them applying to `snap` and `checkpoint`, but I agree it's not as clear that they don't apply to `full`.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
matthew1001: @non-fungible-nelson did you see any issues with your holesky sync?

matthew1001:
> Completed a Holesky snap-sync with no unintended consequences (with the flag set). Some logging noise for cleanup.

I've reverted the `info` logs to `debug` logs in places that @matkt commented on, given that you said there was some logging noise.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapSyncDownloader.java
matthew1001: I've reverted it to a `debug` log

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
matthew1001: I've reverted it to a `debug` log

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
matthew1001: I've reverted it to a `debug` log

non-fungible-nelson:
Completed another holesky sync as of commit 31c0cd3 and it went great.

matthew1001:
@garyschulte @matkt think I've tidied up remaining loose ends:
 - [x] Re-tested a `FAST` sync to check that nothing has regressed there and it looks fine
 - [x] Public sync is looking good as per Matt's comment ^^
 - [x] I've added some additional tests for the BFTPivotSelectorFromPeers` class.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
garyschulte: I would prefer this to be in a different PR since it isn't related to BFT snapsync.  It isn't clear to me that we want to entirely prevent trie log pruning if the sync mode is full.  Linea in particular has some atypical configurations.  

Personally I think _ALL_ of these features should be wrapped in profiles rather than trying to have one-default-fits-all.  @matkt @non-fungible-nelson

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/SynchronizerConfiguration.java
garyschulte: IMO unless we can differentiate that the min peer count only applies to these 'fast' sync processes, we are not adding clarity, just changing things. 

 `--state-sync-min-peers` is less painful than `--weak-subjectivity-sync-min-peers`.

garyschulte:
I won't die on the `--fast-sync-min-peers` hill, but I would like to see the trie-log-pruning option change in a separate PR.

Would mark as :ship: with that minor change.

matthew1001:
Yeah that's reasonable. I'll refactor the `trie-log-pruning` change into its own PR

matkt:
> Completed another holesky sync as of commit [31c0cd3](https://github.com/hyperledger/besu/commit/31c0cd3d8e3cc63610bcd55bdb496d453bfaca9d) and it went great.

I think it will be better to test also a mainnet node just to be sure ( we have more pivot block swtich, more interesting usecases, etc). I don't think the modification will add regression but I will prefer to test (checkpoint is enough)

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -143,13 +143,6 @@ public static DataStorageOptions create() {
    */
   public void validate(final CommandLine commandLine, final SyncMode syncMode) {
     if (DataStorageFormat.BONSAI == dataStorageFormat && bonsaiLimitTrieLogsEnabled) {
-      if (SyncMode.FULL == syncMode) {
```
matkt, line 146: why we are removing this one ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/TrieLogPruner.java
```diff
@@ -99,7 +99,7 @@ private int preloadQueue() {
     }
   }
 
-  public void addToPruneQueue(final long blockNumber, final Hash blockHash) {
+  public synchronized void addToPruneQueue(final long blockNumber, final Hash blockHash) {
```
matkt, line 102: could you explain why this modification is needed ? I think if it was not needed on mainnet, I think it should be the same on qbft ?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
```diff
@@ -94,7 +94,7 @@ protected CompletableFuture<FastSyncState> start(final FastSyncState fastSyncSta
           onBonsai.clearFlatDatabase();
           onBonsai.clearTrieLog();
         });
-    LOG.debug("Start sync with initial sync state {}", fastSyncState);
+    LOG.debug("Start fast sync with initial sync state {}", fastSyncState);
```
matkt, line 97: maybe we can do the same as here https://github.com/hyperledger/besu/pull/7140/files#diff-c971b57a1d994b1c08fd5f43fd3cf282a41ccd524cafa8421c6eb1bad9087b08R83

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/TrieLogPruner.java
matthew1001: I think it is an existing bug that hasn't been hit on mainnet because the rate of receiving and handling new trie-logs from peers isn't fast enough to hit the threading issue.

I was getting `ConcurrentModificationExceptions` on the `trieLogBlocksAndForksByDescendingBlockNumber` map when a QBFT node was syncing from several peers due to `addToPruneQueue` modifying the map while `pruneFromQueue` was iterating over it.

Since it hasn't been hit on mainnet before, I'm assuming the making the methods `synchronized` isn't going to affect mainnet syncing performance at all.

matthew1001: If it's useful I can try recreating it to show the full `ConcurrentModificationException` stack, but it was definitely an intermittent (but reliable) exception once we had got the basic syncing logic working.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
matthew1001: In this case it genuinely is `fast` sync. The `start()` method is common to this class and the `SnapSyncDownload` subclass.

The `start(final FastSyncState fastSyncState)` method is actually doing `fast` sync in this class, and is doing `snap` sync in `SnapSyncDownloader.java` so in `SnapSyncDownloader` the log says "Start snap sync..."

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
matthew1001: This is going to get refactored into a separate PR (see Gary's comment https://github.com/hyperledger/besu/pull/7140#issuecomment-2192593451).

But the reason for removing it is I've changed the logic to set `--bonsai-limit-trie-logs-enabled=false` if `--sync-mode=FULL`, rather than force the user to set `--bonsai-limit-trie-logs-enabled=false`. See https://github.com/hyperledger/besu/blob/7bd09e3ee2ca05298469e7020b037091f3777931/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java#L2251.

It's a bit like how we automatically set `--tx-pool-price-bump=0` if `--min-gas-price=0`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
matkt: Ok missed that thanks 🙏

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/TrieLogPruner.java
matkt: Sounds good. I just ping @siladu who has coded this part and can verify that we are good

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
matkt: Ok thanks for the info 🙏

matthew1001:
> I won't die on the `--fast-sync-min-peers` hill, but I would like to see the trie-log-pruning option change move to a separate PR.
> 
> Would mark as 🚢 with that minor change.

The latest commit removes the trie-log change and I'll raise a separate PR for that shortly where any additional discussion can take place

garyschulte (approved):
GTG by me, 🚢

PR #7068, opened 2024-05-07
EIP-7685

Gabriel-Trintinalia:
###  General purpose execution layer requests [eip-7685](https://eips.ethereum.org/EIPS/eip-7685 )
This PR implements the generic execution layer requests.

Fixes #6985, #6986 and the last item of #6800

### Main changes:

- Removal of `depositsRoot` and `withdrawalRequestRoot` from the `blockheader`
- Removal of `deposit` and `withdrawal` requests from the block body
- Introduction of `requestsRoot` in the `blockHeader`
- Introduction of requests in the block body
- Introduction of Generic Request Type
- Removed `depoistValidator` and `withdrawalRequestValidator`
- Added a generic Request decoder/encoder
- Introduction of `RequestValidatorCoordinator` and `RequestProcessorCoordinator`
- Extract the processing logic of `deposits` and `withdrawalRequests` from block creation and block processing to new `DepositProcessor` and `WithdrawalRequestProcessor`
- Replaced of `depositValidatorProvider` and `withdrawalRequestValidatorProvider` by `RequestValidadorProvider`
- Change logic in the engine methods to use the correct validator to validate requests
- Adapts deposit existing tests and withdrawal existing tests to use the new classess
- Adds generic validation to the request list (Request root, request ordering, etc)

This PR does not ensure the logic of `Deposits` and `withdrawalRequests` but encapsulates them into a generic request list that is used to calculate the new `requestRoot`. The engine API has not changed and the requests are still split into `deposits` and `withdrawalRequests`.

## To do: 
- [ ] Add more unit tests

PR #7142, opened 2024-05-27
Minor improvements to EIP-7685

Gabriel-Trintinalia:
## PR description
Address feedback of #7068


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/RequestUtil.java
```diff
@@ -40,6 +42,23 @@ public static <T extends Request> List<T> filterRequestsOfType(
...
+
+  public static Optional<List<WithdrawalRequest>> getWithdrawalRequests(
+      final Optional<List<Request>> requests) {
+    return requests.map(r -> filterRequestsOfType(r, WithdrawalRequest.class));
+  }
+
+  /**
+   * Combines two optional lists of requests into a single optional list.
+   *
+   * @param maybeDeposits Optional list of deposit requests.
```
macfarla, line 57: nit - for consistency should this be renamed to maybeDepositsRequest

macfarla: actually on further consideration I see that matches the class names Deposit and WithdrawalRequest

macfarla: We have Withdrawal and WithdrawalRequest but Deposit is itself the request... it is a bit confusing but probably outside the scope of this PR. non-blocking.

macfarla (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/RequestUtil.java
Gabriel-Trintinalia: I will rename Deposit to DepositRequest but I wanted to do that in a different PR so it is easier to review it.

jframe (approved):

PR #7145, opened 2024-05-28
[MINOR] Rename deposit to depositRequest

Gabriel-Trintinalia:
## PR description
Rename `deposit` to `depositRequest`


macfarla (approved):
plugin API  hash needs updating.

but the only classes modified are internal

PR #7168, opened 2024-06-04
Subnet-Based Peer Permissions

### LINKED ISSUE #6620: `--netrestrict` feature to enable IP filtering for Besu nodes used in key infrastructure
### Description
As an infrastructure provider, I want to be able to restrict my node to certain IP networks so that I can prevent and filter unwanted peering. I also want parity with Geth nodes with this option to better take advantage of existing configs and tooling. 

Besu should enable restriction to peer-to-peer connectivity on an IP subnet. Doing so will further isolate the network and prevents cross-connecting with other blockchain networks in case the nodes are reachable from the Internet.

### Acceptance Criteria
* `--netrestrict` option exposed taking a string representing IP value 
* No external peering or connection to the internet is enabled
* If this option is set to a non-nil value, only hosts which match one of the IP networks contained in the list are considered.

### Reference code 
[Geth Codebase](https://github.com/search?q=repo:ethereum/go-ethereum%20netrestrict&type=code)

Gabriel-Trintinalia:
## PR description
This pull request introduces a new feature for managing peer permissions based on IP subnet configurations in the networking layer. The core addition is the `PeerPermissionSubnet` class, which extends the existing `PeerPermissions` class to allow for the specification of allowed IP subnets. Peers whose IP addresses fall within these allowed subnets are granted access, enhancing the network's security by ensuring that only peers from specific, trusted IP ranges can interact with the node.

## Fixed Issue(s)
Fixes #6620

## Key Changes:

- `PeerPermissionSubnet` Class: This class implements logic to control access based on IP subnet configurations. It applies restrictions to all peer actions, including but not limited to, sending and receiving messages in both the discovery protocol and the RLPx transport layer. If no subnets are specified, the default behaviour permits all peers.

- CLI Configuration: The introduction of the `--net-restrict` CLI option allows network administrators to specify allowed IP subnets directly through the CLI. This option accepts a comma-separated list of subnet specifications (e.g., --net-restrict=192.168.1.0/24,10.0.0.0/8), converting them into `SubnetInfo` objects using the `SubnetInfoConverter`. If this option is not utilised, no subnet-based peer permission restrictions will be applied, allowing the network to operate without these specific constraints.


ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/permissions/PeerPermissionsSubnetTest.java
```diff
@@ -0,0 +1,72 @@
...
+  @Test
+  public void peerInSubnetRangeShouldBePermitted() {
+    String subnet = "127.0.0.0/24";
+    List<SubnetInfo> allowedSubnets = List.of(new SubnetUtils(subnet).getInfo());
+    PeerPermissionSubnet peerPermissionSubnet = new PeerPermissionSubnet(allowedSubnets);
+    checkPermissions(peerPermissionSubnet, remoteNode, true);
+  }
+
+  @Test
+  public void peerOutSubnetRangeShouldNolBePermitted() {
```
jframe, line 43:
```suggestion
  public void peerOutSubnetRangeShouldNotBePermitted() {
```

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/permissions/PeerPermissionsSubnetTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.commons.net.util.SubnetUtils;
+import org.apache.commons.net.util.SubnetUtils.SubnetInfo;
+import org.junit.jupiter.api.Test;
+
+public class PeerPermissionsSubnetTest {
+
+  private final Peer remoteNode = createPeer();
+
+  @Test
+  public void peerInSubnetRangeShouldBePermitted() {
```
jframe, line 35: A test with multiple subnets to check would be good

jframe (approved):

PR #7171, opened 2024-06-04
Use github arm64 runners

### LINKED ISSUE #7026: Migrate from self-hosted hyp-ci-linux-arm64 to besu-arm64 runner
For arm docker builds we currently target the self-hosted `hyp-ci-linux-arm64` but now there is a github provided runner enabled

ryjones:
## Fixed Issue(s)
fixes #7026

github now offer arm64 runners.

The arm64 image is a plain ubuntu 22.04 server with nothing installed (no docker).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,9 +1,8 @@
-name: docker develop
-
+name: docker
 on:
-  push:
-    branches:
-      - main
+  release:
+    types:
+      - prereleased
 env:
   registry: docker.io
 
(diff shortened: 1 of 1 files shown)


ryjones:
@seanyoung is this the change you wanted?

seanyoung:
There should be an arm64 image with all the docker tools installed https://github.com/github-early-access/arm-runners-beta/blob/main/images/arm-ubuntu-22-image.md

That should make this PR much simpler.

Issue #6560, opened 2024-02-12
default for bonsai-limit-trie-logs-enabled=True


macfarla:
change default (currently false) for

bonsai-limit-trie-logs-enabled=True

macfarla:
this change breaks the minimalist-staker profile with the message  --bonsai-historical-block-limit minimum value is 512
minimalist-staker profile has it set to 128
check is done [here](https://github.com/hyperledger/besu/blob/61c7e18ab446a7383bafc2e10d96a88595cf1986/besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java#L113)
with trie-log-limit enabled = false (previous default), it doesn't do that check

can we lower the minimum value from 512 to 128 cc @non-fungible-nelson

macfarla:
actually went the other way in the PR - changed minimalist-staker limit from 128 to 512

macfarla:
(won't be a breaking change if we want to go the other way in the future)

macfarla:
nope, changed again. minimum = 128 and default = 512

non-fungible-nelson:
@siladu - any comments here? I know we had some interesting behaviors with minimums and defaults. What we would like to do is support a global default of 512, but a local default of 128 for the minimal staker profile.

siladu:
> @siladu - any comments here? I know we had some interesting behaviors with minimums and defaults. What we would like to do is support a global default of 512, but a local default of 128 for the minimal staker profile.

Discussed this with @macfarla.
I'm wondering what is the benefit of reducing from 512 -> 128? 
In terms of storing trie logs, there's only about 60MB difference between 512 vs 128 so not worth the extra logic IMO.

PR #7181, opened 2024-06-05
Enable --Xbonsai-limit-trie-logs-enabled by default

siladu:
## PR description

Enable `--Xbonsai-limit-trie-logs-enabled` by default, unless sync-mode=FULL.
More info: https://wiki.hyperledger.org/display/BESU/Limit+Trie+Logs+for+Bonsai

When sync-mode=FULL and data-storage-format=BONSAI you get a config validation error at startup...
```
besu --sync-mode=FULL --data-storage-format=BONSAI
2024-06-06 17:36:12.806+01:00 | main | INFO  | Besu | Starting Besu
2024-06-06 17:36:13.027+01:00 | main | ERROR | Besu | Failed to start Besu
...
Cannot enable --Xbonsai-limit-trie-logs-enabled with sync mode FULL
```

I will promote to a Stable option in a separate PR.

This option has been tested on our canaries and by the community since https://github.com/hyperledger/besu/releases/tag/24.3.0

Added a fallback value for the option otherwise specifying --Xbonsai-limit-trie-logs-enabled on the command line, without a value, toggles it back off again.

The `minimalist-profile` has been updated from `bonsai-historical-block-limit=128` to `bonsai-historical-block-limit=512` to avoid further code changes and because I believe there is only a ~60MB difference so using 128 has limited value, see https://github.com/hyperledger/besu/issues/6560#issuecomment-1953928253

I will conservatively set `Xbonsai-limit-trie-logs-enabled=false` in the `enterprise` profile for now as not sure it is desirable.

## Fixed Issue(s)

Fixes https://github.com/hyperledger/besu/pull/6560


besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -86,6 +86,8 @@ public static class Unstable {
     @CommandLine.Option(
         hidden = true,
         names = {BONSAI_LIMIT_TRIE_LOGS_ENABLED, "--Xbonsai-trie-log-pruning-enabled"},
+        defaultValue = "true",
```
jframe, line 89: Should this be enabled by default for all profiles? In the The enterprise-private profile probably makes more sense to not prune I think. @non-fungible-nelson What do you think?

siladu: Probably not. I thought it might be alright initially since it's scoped to BONSAI, but I know we have a follow-on story related to profiles somewhere...I will check if we need to play that at the same time.

siladu: This is relevant https://github.com/hyperledger/besu/issues/6560

I think this may have been the PR I was thinking about, which doesn't actually touch the profiles directly but had some impact on them https://github.com/hyperledger/besu/pull/6561

besu/src/test/java/org/hyperledger/besu/cli/BesuCommandTest.java
```diff
@@ -1244,8 +1244,13 @@ public void ethStatsContactOptionCannotBeUsedWithoutEthStatsServerProvided() {
   }
```
fab-10, line 1244: any test to verify it is enabled by default?

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
siladu: I disabled this in enterprise-profile 👍  - was surprised to find enterprise was defaulting to BONSAI, addressed that here: https://github.com/hyperledger/besu/pull/7186

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
```diff
@@ -69,14 +69,16 @@ default Unstable getUnstable() {
...
+    boolean DEFAULT_BONSAI_LIMIT_TRIE_LOGS_ENABLED = true;
     long MINIMUM_BONSAI_TRIE_LOG_RETENTION_LIMIT = DEFAULT_BONSAI_MAX_LAYERS_TO_LOAD;
     int DEFAULT_BONSAI_TRIE_LOG_PRUNING_WINDOW_SIZE = 30_000;
     boolean DEFAULT_BONSAI_FULL_FLAT_DB_ENABLED = true;
     boolean DEFAULT_BONSAI_CODE_USING_CODE_HASH_ENABLED = true;
 
     DataStorageConfiguration.Unstable DEFAULT =
-        ImmutableDataStorageConfiguration.Unstable.builder().build();
+        ImmutableDataStorageConfiguration.Unstable.builder()
+            .bonsaiLimitTrieLogsEnabled(true)
```
fab-10, line 80: this should be redundant since the default is true now, right?

fab-10 (approved):
LGTM, just a couple of nits

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/DataStorageConfiguration.java
siladu: removed here [`100bf43` (#7181)](https://github.com/hyperledger/besu/pull/7181/commits/100bf43dd97419a42f8e92bc7390ec651e466ea5)

non-fungible-nelson:
Commit LGTM - thanks @siladu for the last minute fixes.

jframe (approved):

matthew1001:
FYI @siladu in PR https://github.com/hyperledger/besu/pull/7140 I was planning to change the fail-on-start behaviour if `FULL` sync mode is enabled and `bonsai-trie-log-pruning-enabled` is set, and instead set `bonsai-trie-log-pruning-enabled` to false if sync mode is `FULL` (with a log message to state that the default has been overridden).

Let me know if you have any concerns about that change - maybe via comments in that PR?

PR #7190, opened 2024-06-07
Snapserver responses to return at least one response

jframe:
## PR description
The snap server responses must return at least one response item if the data can be retrieved according to the spec https://github.com/ethereum/devp2p/blob/master/caps/snap.md. This applies to all the request types.

After making this change all the GetByteCodes hive tests are passing. Note still need to make a change to hive testsuite to enable snapserver before this will be seen.

## Fixed Issue(s)

Part of #7197


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -456,8 +456,9 @@ MessageData constructGetBytecodesResponse(final MessageData message) {
         } else {
           Optional<Bytes> optCode = worldStateStorageCoordinator.getCode(Hash.wrap(codeHash), null);
           if (optCode.isPresent()) {
-            if (sumListBytes(codeBytes) + optCode.get().size() > maxResponseBytes
-                || stopWatch.getTime() > StatefulPredicate.MAX_MILLIS_PER_REQUEST) {
+            if (!codeBytes.isEmpty()
```
garyschulte, line 459: sumListBytes operates on the stream, with an orElse(0).  it shouldn't matter whether the list is empty or not right?  what am I missing here?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -608,11 +612,12 @@ public boolean test(final Pair<Bytes32, Bytes> pair) {
         return false;
       }
 
+      var hasNoRecords = recordLimit.get() == 0;
       var underRecordLimit = recordLimit.addAndGet(1) <= MAX_ENTRIES_PER_REQUEST;
       var underByteLimit =
           byteLimit.accumulateAndGet(0, (cur, __) -> cur + encodingSizeAccumulator.apply(pair))
               < maxResponseBytesFudgeFactor;
-      if (underRecordLimit && underByteLimit) {
+      if (hasNoRecords || (underRecordLimit && underByteLimit)) {
```
garyschulte, line 620: Is this a "pathological" hive test case where the byteLimit is set too low to allow any records through?  

👍 This seems a reasonable fix if the spec says there must always be at least one record in the response, but we should add a comment about why we are doing this, because it isn't obvious why we would do this otherwise.

garyschulte (approved):
LGTM, 
one question about the isEmpty() checks, and an ask for comments, but otherwise 🚢

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
jframe: The sumListBytes will work fine on an empty list. This change isn't to guard against calling sumListBytes with an empty list but to ensure we only do the `sumListBytes(codeBytes) > maxReponseBytes` check once we have at least one record in the response. So that we always send at least once record in the response.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
jframe: There wasn't any "pathological" test for the responses that use this predicate, well none obvious anyway. The "pathological" test was for getByteCodes with a limit too small to allow any response. But since the spec does state we need to return at least one response I made the changes anyway to be compliant.

PR #7203, opened 2024-06-11
Update Dockerfile - reuse NO_PROXY_CACHE env

tbenr:
Just noticed this.


usmansaleem (approved):
LGTM

docker/Dockerfile
```diff
@@ -6,7 +6,7 @@ ENV NO_PROXY_CACHE="-o Acquire::BrokenProxy=true -o Acquire::http::No-Cache=true
 # Update and install dependencies without using any cache
 RUN apt-get update $NO_PROXY_CACHE  && \
   # $NO_PROXY_CACHE must not be used here or otherwise will trigger a hadolint error
```
usmansaleem, line 8: nit: we can remove the above comment as well.

tbenr: there was an actual bug in my changes. Fixed and removed a redundant `rm -rf`

Issue #6680, opened 2024-03-05
Invalid block detected (BREACH_OF_PROTOCOL) | IBFT 2.0 


avenbreaks:
### Description
After the node stop and then tries to re-run it seems to get an `invalid block detected `error and then the peer disconnects. We build blockchain using besu for private networks. 

### Config 
```
besu/bin/besu \
     --api-gas-price-max=1000000000000000 \
     --min-gas-price=1000000000 \
     --data-path=Node-1/data \
     --data-storage-format=BONSAI \
     --network-id=6666 \
     --rpc-http-enabled=true \
     --rpc-http-host=0.0.0.0 \
     --rpc-http-port=8545 \
     --rpc-http-cors-origins="*" \
     --rpc-http-api=NET,WEB3,ETH,IBFT,ENGINE,TXPOOL,TRACE \
     --rpc-http-authentication-jwt-algorithm=RS256 \
     --rpc-ws-enabled=true \
     --rpc-ws-api=NET,WEB3,ETH,IBFT,ENGINE,TXPOOL,TRACE \
     --rpc-ws-host=0.0.0.0 \
     --rpc-ws-port=8546 \
     --rpc-ws-max-active-connections=100 \
     --rpc-ws-max-frame-size=1048576 \
     --genesis-file=genesis.json \
     --host-allowlist="*" \
     --engine-rpc-enabled=true \
     --engine-host-allowlist="*" \
     --engine-jwt-secret=jwtsecret.hex \
     --engine-rpc-port=8551 \
     --node-private-key-file=/Node-1/data/key \
     --max-peers=4 \
     --Xp2p-peer-lower-bound=4 \
     --sync-mode=FULL \
     --p2p-enabled=true \
     --p2p-host=XXX.XXX.XXX.XXX \
     --p2p-port=30303 \
     --logging=INFO \
     --metrics-enabled=true \
     --metrics-host=0.0.0.0 \
     --metrics-port=9545 \
     --metrics-protocol=PROMETHEUS \
     --metrics-category=BLOCKCHAIN,ETHEREUM,EXECUTORS,NETWORK,PEERS,PERMISSIONING,PRUNER,RPC,SYNCHRONIZER,TRANSACTION_POOL,STRATUM,JVM,PROCESS 

```
### Logs (if a bug)

```
2024-03-05 02:05:41.425+00:00 | BftProcessorExecutor-IBFT-0 | ERROR | AbstractBlockProcessor | failed persisting block
[...]

PR #7204, opened 2024-06-11
Don't persist BFT proposed blocks, only committed ones

matthew1001:
## PR description
This PR changes BFT block proposal validation to not persist the changes as a result of the proposed block. When a QBFT `commit` message or Ethereum `NewBlock` message is processed the changes will still be persisted at that point.

Prior to this PR I could only get a few 1000 transactions through a 4-validator QBFT chain before I hit `Invalid block` error. With this fix in place I can run any number of high TPS tests against a 4-validator chain and have yet to see an `Invalid block` error.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/6680

(It might fix this issue, or there could still be something to resolve regarding the `RocksDB busy` error. I suspect this PR will improve the above issue at the very least, but suggest closing it with this PR and re-opening if it still occurs after this has merged)


matkt (approved):
LGTM

jframe (approved):
LGTM - Nice change

PR #7206, opened 2024-06-11
[VERKLE] Preload stems 

matkt:
## PR description

Currently, trie keys generation will gather all transactions data and generate a single large batch of trie keys at once.
However, there are already mechanisms implemented in Besu to preload those trie keys in a background thread during block processing. 

This PR generates the different triekeys in the background in order to have a maximum of triekeys already calculated at the end of the block


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/cache/TrieKeyPreloader.java
```diff
@@ -0,0 +1,258 @@
...
+        .toList();
+  }
+
+  private Map<Bytes32, Bytes32> createAccountCache() {
+    return new ConcurrentHashMap<>();
+  }
+
+  private Map<Bytes32, Bytes32> getCachedTrieKeysForAccount(final Address address) {
+    return Optional.ofNullable(trieKeysCache.getIfPresent(address))
+        .orElseGet(ConcurrentHashMap::new);
```
garyschulte, line 250: it doesn't look like trieKeysCache can ever be null in this case

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/cache/TrieKeyPreloader.java
```diff
@@ -0,0 +1,258 @@
...
+
+  public void preLoadStorageSlot(final Address account, final StorageSlotKey slotKey) {
+    CompletableFuture.runAsync(() -> cacheSlotTrieKeys(account, slotKey));
+  }
+
+  public void preLoadCode(final Address account, final Bytes code) {
+    CompletableFuture.runAsync(() -> cacheCodeTrieKeys(account, code));
+  }
+
+  public Map<Bytes32, Bytes32> generateManyTrieKeyHashes(
```
garyschulte, line 118: do we really need to cache trieKeys?  seems like the only expensive thing to cache would be the 31 byte stem address, and we can just suffix it at-will.  It should reduce the cache size and might make for a better performing key generation on execution (since the stem would be cached and not regenerated if the slot key or code keys were not pre-calculated)

garyschulte (approved):
LGTM, a couple non-blocking suggestions

matkt:
I updated the code in order to save stem , this one depends on this besu-verkle-trie PR https://github.com/hyperledger/besu-verkle-trie/pull/64

jflo:
already implemented via another branch

PR #7222, opened 2024-06-13
Add build version option to prefix git hash with custom version property

matthew1001:
## PR description
Adds the option to provide a custom version prefix at build time that will be prefixed to the git hash. Currently a custom version can't include the githash. This follows on from some recent version refactoring in https://github.com/hyperledger/besu/pull/6699

Specifically it allows for `-Pversionprefix=24.5.2` to result in a version like `v24.5.2-16ff080`


build.gradle
```diff
@@ -935,6 +935,9 @@ def calculateVersion() {
 
   if (project.hasProperty('version') && (project.version =~ calVerPattern)) {
     return "${project.version}"
+  } else if (project.hasProperty('versionprefix') && (project.versionprefix =~ calVerPattern)) {
+    def gitDetails = getGitCommitDetails(7) // Adjust length as needed
+    return "${project.versionprefix}-${gitDetails.hash}"
```
garyschulte, line 940: instead of introducing versionprefix, could we instead add a 'force githash suffix' property?  The behavior would be more intuitive imo, instead of having `version` and `versionprefix`

garyschulte (changes requested):
This seems reasonable, but I think we can accomplish the stated goal with more clarity by just forcing githash suffix

build.gradle
matthew1001: Yeah I think that would be a good way of achieving the same thing. I'm on vacation tomorrow but I'll update the PR on Monday.

matthew1001: I've refactored along the lines you suggested @garyschult. I've also included commented-out examples of using them.

The example `version` field I've put in the file includes an example vendor name: `24.5.6-acme`. That's how we'll probably use the field, and I think other vendors might want to set it in a similar way if they package Besu into their own offerings, so their users can check that they're running a version from their vendor. Obviously just setting `24.5.6` works fine as well.

matthew1001:
@garyschulte any more thoughts on this or are you happy to approve in its new shape?

PR #6501, opened 2024-01-31
Deprecate --Xp2p-peer-lower-bound

macfarla:
Deprecate CLI option --Xp2p-peer-lower-bound and remove usage

Fixes #6376

PR #7247, opened 2024-06-20
Remove deprecation message for `--Xp2p-peer-lower-bound`

Gabriel-Trintinalia:
## PR description

Remove the deprecation message for `--Xp2p-peer-lower-bound`. This option has already been removed. 
See https://github.com/hyperledger/besu/pull/6501


macfarla (approved):
LGTM - prob need a changelog entry for final removal of this

PR #7245, opened 2024-06-19
Reconfigure how Protocol Specs are created WRT EVM condiguration

shemnon:
## PR description

Make the max code size and max initcode size a part of the EVM configuration. As part of the change we need to move the tasks CodeFactory once handled as a static class and move it into the EVM. This has a nice follow on effect that we don't need to pass in max EOF versions or max code sizes anymore.

This is also required to address a recent EOF change limiting the total size of containers.

### Thanks for sending a pull request! Have you done the following?

- [ ] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [ ] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [ ] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [ ] unit tests: `./gradlew build`
- [ ] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

PR #7250, opened 2024-06-21
Make code sizes a characteristic of the EVM version

shemnon:
## PR description

Make the max code size and max initcode size a part of the EVM
configuration. As part of the change we need to move the tasks
CodeFactory once handled as a static class and move it into the EVM.
This has a nice follow on effect that we don't need to pass in max EOF
versions or max code sizes anymore.


shemnon:
This is a variant of #7245 removing the ProtocolSpec configurations.  #7245  will remain as the PR where we alter protocol spec creation.

shemnon:
Main config is fine.  Closing and merging the parent PR.

PR #7252, opened 2024-06-22
Fix the wrong 'Identifier' and 'Synchronizer' usage

leniram159:
## PR description
Fix the wrong 'Identifier' and 'Synchronizer' usage


macfarla:
@leniram159 you'll need to fix the DCO (add signoff to your commits) before we can accept your PR

macfarla (approved):

PR #7257, opened 2024-06-25
Add bootnodes to the maintained peer list

matthew1001:
## PR description
This PR adds the bootnodes to the maintained-peer list in a permissioned chain. Currently the maintained node list only contains static nodes and nodes added with `admin_addPeer`. Adding bootnodes to the maintained list ensures that if a bootnode is offline when the node starts, it will retry its connection once a minute rather than requiring a restart.

Bootnode connectivity in a public-chain node remains unchanged.

## Fixed issues
Fixes https://github.com/hyperledger/besu/issues/7261

- [ ] ~For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests~

Based on the outcome of this PR, I plan to raise a docs PR to describe how connections are maintained to bootnodes and static nodes since there isn't much detail covering the behaviour today. I'll wait for the outcome of this PR before raising the docs changes though.

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
@@ -795,7 +795,10 @@ public class RunnerBuilder {
       // In a permissioned chain Besu should maintain connections to both static nodes and
       // bootnodes, which includes retries periodically
       maintainedPeers =
-          Stream.concat(sanitizePeers(network, staticNodes), sanitizePeers(network, bootnodes));
+          sanitizePeers(
+              network,
+              Stream.concat(staticNodes.stream(), bootnodes.stream()).collect(Collectors.toList()));
+      LOG.debug("Added bootnodes to the maintained peer list");
     } else {
       // In a public chain only maintain connections to static nodes
       maintainedPeers = sanitizePeers(network, staticNodes);
--- a/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
+++ b/besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
@@ -790,18 +790,19 @@ public class RunnerBuilder {
       LOG.debug("added ethash observer: {}", stratumServer.get());
     }
 
+    final Stream<EnodeURL> maintainedPeers;
     if (besuController.getGenesisConfigOptions().isPoa()) {
       // In a permissioned chain Besu should maintain connections to both static nodes and
       // bootnodes, which includes retries periodically
-      Stream.concat(sanitizePeers(network, staticNodes), sanitizePeers(network, bootnodes))
-          .map(DefaultPeer::fromEnodeURL)
-          .forEach(peerNetwork::addMaintainedConnectionPeer);
+      maintainedPeers =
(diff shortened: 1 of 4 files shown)


besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -790,9 +790,18 @@ public Runner build() {
       LOG.debug("added ethash observer: {}", stratumServer.get());
     }
 
-    sanitizePeers(network, staticNodes)
-        .map(DefaultPeer::fromEnodeURL)
-        .forEach(peerNetwork::addMaintainedConnectionPeer);
+    if (besuController.getGenesisConfigOptions().isPoa()) {
+      // In a permissioned chain Besu should maintain connections to both static nodes and
+      // bootnodes, which includes retries periodically
+      Stream.concat(sanitizePeers(network, staticNodes), sanitizePeers(network, bootnodes))
```
shemnon, line 796: nit: could the stream be factored out as a variable and the map and forEach statements are only written once? This insulates us from changes to those steps accidentally only impacting one path.

matthew1001: Sure, I'll make that change now and push a new commit

matthew1001: See latest changes @shemnon

besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -790,7 +790,17 @@ public Runner build() {
       LOG.debug("added ethash observer: {}", stratumServer.get());
     }
 
-    sanitizePeers(network, staticNodes)
+    final Stream<EnodeURL> maintainedPeers;
+    if (besuController.getGenesisConfigOptions().isPoa()) {
+      // In a permissioned chain Besu should maintain connections to both static nodes and
+      // bootnodes, which includes retries periodically
+      maintainedPeers =
+          Stream.concat(sanitizePeers(network, staticNodes), sanitizePeers(network, bootnodes));
```
macfarla, line 798: would it be more efficient to join the 2 collections and call sanitizePeers only once?

macfarla: suggest a debug log message that this is happening

macfarla (approved):
LGTM. I would just make sure also that the logging of static nodes, bootnodes etc at startup still makes sense

besu/src/main/java/org/hyperledger/besu/RunnerBuilder.java
matthew1001: Thanks @macfarla - both comments should now be addressed

matthew1001: > I would just make sure also that the logging of static nodes, bootnodes etc at startup still makes sense
I'm hoping that I haven't affected any existing maintained-peer logs. The maintained-peer list also contains any nodes added manually with `admin_addPeer` so it's not purely a static-nodes list today.

PR #7258, opened 2024-06-25
Implement getNearest methods

matkt:
## PR description

Add getNearestAfter method needed for verkle and fix getNearestBefore method


services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/KeyComparator.java
```diff
@@ -0,0 +1,52 @@
...
+   */
+  public static int compareKeyLeftToRight(final Bytes key1, final Bytes key2) {
+    int minLength = Math.min(key1.size(), key2.size());
+    for (int i = 0; i < minLength; i++) {
+      int compare = Byte.compareUnsigned(key1.get(i), key2.get(i));
+      if (compare != 0) {
+        return compare;
+      }
+    }
+    return Integer.compare(key1.size(), key2.size());
```
garyschulte, line 50: is Byte.compareUnsigned ignoring trailing bytes and/or zero's?

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -101,25 +102,72 @@ public Optional<byte[]> get(final SegmentIdentifier segmentId, final byte[] key)
...
+      return parentNearest;
+    } else {
+      final int ourCommonPrefixLength = ourNearest.get().key().commonPrefixLength(key);
+      final int parentCommonPrefixLength = parentNearest.get().key().commonPrefixLength(key);
+      if (ourCommonPrefixLength != parentCommonPrefixLength) {
+        return ourCommonPrefixLength > parentCommonPrefixLength ? ourNearest : parentNearest;
+      } else {
+        return (isAfter ^ (ourNearest.get().key().compareTo(parentNearest.get().key()) > 0))
+            ? ourNearest
+            : parentNearest;
```
garyschulte, line 166: this is very succinct 🥇, but I had to think hard about this.  It might be worth a comment to explain to our future selves

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
```diff
@@ -143,27 +145,70 @@ public Optional<byte[]> get(final SegmentIdentifier segmentIdentifier, final byt
...
+                // Depending on the useMin flag, find either the minimum or maximum entry according
+                // to key order
+                final Optional<Map.Entry<Bytes, Optional<byte[]>>> sortedStream =
+                    useMin
+                        ? filteredStream.min(Map.Entry.comparingByKey())
+                        : filteredStream.max(Map.Entry.comparingByKey());
+                // Map the found entry to a NearestKeyValue object
+                return sortedStream.map(
+                    entry -> new NearestKeyValue(entry.getKey(), entry.getValue()));
+              };
```
garyschulte, line 200: The only param from the method signature used in this function is useMin.  Can we define 2 static predicates, one for before and one for after, rather than creating a new function for every call to getNearest?    I would presume that allocating a new function for every call to getNearest is unnecessarily expensive

services/kvstore/src/test/java/org/hyperledger/besu/services/kvstore/AbstractSegmentedKeyValueStorageTest.java
```diff
@@ -29,52 +30,58 @@
...
+  protected SegmentedKeyValueStorage createSegmentedKeyValueStorageWithValues() throws IOException {
     try (final var store = this.createSegmentedStore()) {
-
-      // create 10 entries
       final SegmentedKeyValueStorageTransaction tx = store.startTransaction();
       IntStream.range(1, 10)
           .forEach(
               i -> {
                 final byte[] key = bytesFromHexString("000" + i);
+                System.out.println(Bytes.of(key));
```
garyschulte, line 40: ☠️

services/kvstore/src/test/java/org/hyperledger/besu/services/kvstore/AbstractSegmentedKeyValueStorageTest.java
```diff
@@ -29,52 +30,58 @@
...
                 final byte[] value = bytesFromHexString("0FFF");
                 tx.put(SEGMENT_IDENTIFIER, key, value);
                 // different common prefix, and reversed order of bytes:
                 final byte[] key2 = bytesFromHexString("010" + (10 - i));
+                System.out.println(Bytes.of(key2));
```
garyschulte, line 45: ☠️

garyschulte (changes requested):
Generally LGTM.  the rewrite of getNearest seems a little chat-gpt-y to me though

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/KeyComparator.java
matkt: Not sure to understand why you want to ignore zero in this case

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
matkt: Is it just this line that seems complicated or the method?

matkt:
> Generally LGTM. the rewrite of getNearest seems a little chat-gpt-y to me though

Thanks for the review.. Can you also tell me what bothers so that I can change. Is factorization too hard to understand ?

The use of Min Max on this case is more interesting because it avoids sorting that would be more expensive. This is why this proposal seems to me more interesting for this call. But I can understand that the overall factorization of the method is complex and I can try to simplify by duplicating more

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
matkt: I simplified this code and added explanation , feel free to tell me if it's better

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
matkt: we are also using segmentMap and this one is dynamic . not sure how you want to do, if you can give me an example to understand your idea

services/kvstore/src/test/java/org/hyperledger/besu/services/kvstore/AbstractSegmentedKeyValueStorageTest.java
matkt: removed

matkt:
@garyschulte did some clean in the PR, I also modified the tests in order to check more edgcases, the new tests are comparing directly with rocksdb instead of hardcoding the responses

thanks to that I fixed some new issues

garyschulte:
> The use of Min Max on this case is more interesting because it avoids sorting that would be more expensive.

that makes a lot of sense, min/max will be O(n). 👍

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/KeyComparator.java
garyschulte: I was curious why the fallback return of Integer.compare - thinking about it today, that is just delegating in case one has a length of 0 right

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
garyschulte: it might or might not be more expensive  to create a function for every call versus just having a static function we pass parameters to.  It would be an interesting question to answer with profiling

garyschulte (approved):
LGTM.  I am interested to know if there is measurable overhead for creating a function on every call to getNearest, but it is not worth blocking the PR for.

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/SegmentedInMemoryKeyValueStorage.java
matkt: maybe @ahamlat can help for that ? if you think it will impact performance I will modify this part

ahamlat: We can do micro-benchmarking with JMH on both implementations and enable Async profiler during the test. I just need to have an example of the usage of both methods, like a unit test.

macfarla:
@garyschulte is this GTG?

matkt:
yes will fix the PR

Issue #7265, opened 2024-06-26
Externalise profile loading


Gabriel-Trintinalia:
https://besu.hyperledger.org/public-networks/how-to/use-configuration-file/profile

Ability to load a custom profile. This should follow similar pattern as the plugin directory. There are internal plugins and then there are external plugins. Similarly, there are internal profiles, there should be an option to load external profiles from a default or custom location.

PR #7292, opened 2024-07-03
Enhance --profile to load external profiles

usmansaleem:
## PR description
Add support to load external profiles using `--profile` option. The external profiles are expected to be placed in `profiles` directory under besu home (similar to `plugins` directory). This can be overridden using system property `besu.profiles.dir`. Running `./besu --help` should report internal profiles along with external profiles - file names without .toml extension.

External profiles can be used to create custom Besu bundles with various plugins and their default options.

Example setup:
- Create `profiles` directory under besu home
- Create `profiles/custom_profile.toml` with configuration values to override.
- Use `./besu --profile=custom_profile ...` to load the custom profile

Docs link: https://besu.hyperledger.org/public-networks/how-to/use-configuration-file/profile

## Fixed Issue(s)

#7265 


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2764,7 +2765,8 @@ private String generateConfigurationOverview() {
     }
 
     if (profile != null) {
-      builder.setProfile(profile.toString());
+      // TODO: Do we need to convert _ to spaces?
```
pinges, line 2768: Is that for the profile name? I would say no, because that only make it harder. The name is the name!

Gabriel-Trintinalia (approved):
LGTM

PR #7296, opened 2024-07-05
Optimistic parallelization of transactions to improve performance

matkt:
## PR description

The idea of this PR is to create a system for parallelizing transactions in an optimistic way. What's the parallelization mechanism we're proposing?

We plan to execute all transactions within a block in parallel, without checking for conflicts between them. We're starting with the assumption that everything can be parallelized. Then, without waiting for the results of these background executions, we'll process the block sequentially.

During the sequential phase, we'll perform various checks:

## Step 1: Transaction Finalization Check 

Determine if each transaction has been finalized by background threads.

### If the transaction is finalized:
We check whether there are any modification conflicts with previously executed transactions.
- No Conflict: Directly apply the background-generated state modifications to the block, bypassing re-execution.
- Conflict Detected: Proceed to replay the transaction using a cache of background reads to enhance efficiency.
### If the transaction hasn't been finalized
The transaction is executed sequentially within the block, ensuring completion without relying on background threads.

# Conflict Detection

Currently, we're implementing a simple method to detect conflicts to avoid debugging complexities or more edgecases to manage. However, the simplicity of the code also allows for future enhancements to reduce false positives.

We're utilizing a Bonsai feature that identifies touched or modified addresses and slots during the execution of a block or transaction. We compare the list of touched accounts and slots from the transaction with that of the block. Each time a transaction is added to the block, its list is also added to the block. Therefore, before adding a new transaction, we check to ensure it hasn't interacted with an account modified by the block (i.e., by previous transactions). If the block only reads the data, it's not considered a conflict.

More complex verification mechanisms could be developed, such as identifying the exact nature of the modification (slot, balance, etc.). If the modifications involve different slots, it might be acceptable. Currently, I'm not doing that , and already, more than 40% of transactions don't need to be replayed. The extent to which we want to detect conflicts, considering the additional cost of more Bytes checks and comparison, remains to be determined.

# Key Benefits of this :

This approach maximizes CPU and disk usage. Traditional sequential execution does not use all of the available resources and can be slow. By processing transactions in parallel and importing the result sequentially, we make full use of the system's capabilities, speeding up the entire process. Without that, we'd only be using a small amount of the available resources.


garyschulte (commented):
Still reviewing the parallelism bit, but providing early feedback on the plumbing itself.  It seems to me that protocolschedule / spec is not the right place to plumb this.

~~I think adding these few configs to TransactionProcessorBuilder interface would limit the number of places we need to plumb these options.  We will still need some changes to AbstractBlockProcessor, but I think the changes would be minimal. ~~

Looking at how dissimilar MainnetTransactionProcessor and ParallelizedConcurrentTransactionProcessor are, we would probably need to implement a BlockProcessorBuilder and plumb it in as an attribute to ProtocolSpecBuilder, and move the TransactionProcessorBuilder into the BlockProcessorBuilder.  

That should limit the plumbing and the need to expose the mechanics of transaction processing in the protocol schedule.  Otherwise we could configure the block processor via dagger, and just pass a dagger context into the protocol spec builder.

ahamlat:
I like the idea @garyschulte, I think we still need to give metricsSystem object to MainnetBlockProcessor constructor as I would like to add more metrics on block processing time (PRs to come), when parallel transaction flag is enabled or not.

matkt:
> Still reviewing the parallelism bit, but providing early feedback on the plumbing itself. It seems to me that protocolschedule / spec is not the right place to plumb this.
> 
> ~I think adding these few configs to TransactionProcessorBuilder interface would limit the number of places we need to plumb these options. We will still need some changes to AbstractBlockProcessor, but I think the changes would be minimal~.
> 
> Looking at how dissimilar MainnetTransactionProcessor and ParallelizedConcurrentTransactionProcessor are, a clean way to not plumb those fields everywhere would be to add BlockProcessorBuilder to ProtocolSpecBuilder, and subclass MainnetBlockProcessor as ParallelMainnetBlockProcessor.
> 
> That should limit the plumbing and the need to expose the mechanics of transaction processing in the protocol schedule. It would be nice if we could configure the block processor via dagger, and just pass a dagger context into the protocol spec builder, so we can get away from plumbing all of these things via constructor args and builder vars. @jflo
> 
> here is a [commit](https://github.com/garyschulte/besu/commit/29dd4f590626592136eb378371e27b122b6d2ef0) to try to better explain what I mean

Good idea, I used your commit and I had to adapt it because I want to have a new instance of the ParallelizedTransactionProcessor for each processBlock

datatypes/src/main/java/org/hyperledger/besu/datatypes/StorageSlotKey.java
```diff
@@ -94,7 +94,7 @@ public boolean equals(final Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(slotHash.hashCode());
+    return slotHash.hashCode();
```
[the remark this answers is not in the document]
matkt: just a cleaning of the code because it seems to be useless

crypto/algorithms/src/main/java/org/hyperledger/besu/crypto/Blake2bfMessageDigest.java
```diff
@@ -112,6 +119,17 @@ public static class Blake2bfDigest implements Digest {
       v = new long[16];
     }
 
+    @Override
+    public Blake2bfDigest clone() throws CloneNotSupportedException {
```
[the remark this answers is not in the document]
matkt: we are using that here https://github.com/hyperledger/besu/blob/main/crypto/algorithms/src/main/java/org/hyperledger/besu/crypto/Hash.java#L61

evm/src/main/java/org/hyperledger/besu/evm/tracing/OperationTracer.java
```diff
@@ -85,6 +86,17 @@ default void tracePrepareTransaction(final WorldView worldView, final Transactio
...
+  /**
+   * Trace the end of a transaction just before mining reward.
+   *
+   * @param worldView an immutable view of the execution context
+   * @param tx the transaction that just concluded
+   * @param miningReward the reward that the mining beneficiary will receive.
+   */
+  default void traceTransactionBeforeMiningReward(
+      final WorldView worldView, final Transaction tx, final Wei miningReward) {}
+  ;
```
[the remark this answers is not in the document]
matkt: yes good point

ahamlat:
We observed that approximately 40% of the transactions were parallelized thanks to this PR, while the remaining transactions were conflicting. However, the execution results were cached to avoid reprocessing them a second time.
The conflict detection algorithm remains very simple at this stage, we can improve it in future PRs and thus improve the number of parallel executions.

image

We also notilced at least 25% improvement on block processing time, some CLs have even better results as the overhead on Besu block execution is less important (Both EL and CL run on the same machine and share the same resources).

image

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
```diff
@@ -122,6 +122,14 @@ public static class Unstable {
             "Enables code storage using code hash instead of by account hash. (default: ${DEFAULT-VALUE})")
     private boolean bonsaiCodeUsingCodeHashEnabled = DEFAULT_BONSAI_CODE_USING_CODE_HASH_ENABLED;
 
+    @CommandLine.Option(
+        hidden = true,
+        names = {"--Xbonsai-parallel-tx-enabled"},
+        arity = "1",
+        description =
+            "Enables parallelization of transactions to optimize processing speed by concurrently loading and executing necessary data in advance. (default: ${DEFAULT-VALUE})")
+    private Boolean isParallelTxEnabled = false;
```
fab-10, line 131: non blocking suggestion: s/ParallelTxEnabled/ParallelTxProcessingEnabled/
seems more appropriate to specify to what the parallelism apply

crypto/algorithms/src/test/java/org/hyperledger/besu/crypto/Blake2bfMessageDigestTest.java
```diff
@@ -124,6 +141,39 @@ public void throwsIfEmptyBufferUpdatedLargeByteArray() {
         .isInstanceOf(IllegalArgumentException.class);
   }
 
+  @SuppressWarnings("unchecked")
+  @Test
+  public void testDigestThreadSafety() throws ExecutionException, InterruptedException {
```
fab-10, line 146: I think this test is not deterministic, so even if it passes it is not sure that the code is thread safe.
Not sure how to explicitly test the thread safety, but using a special executors could work

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -484,6 +483,9 @@ public TransactionProcessingResult processTransaction(
       final Wei coinbaseWeiDelta =
           coinbaseCalculator.price(usedGas, transactionGasPrice, blockHeader.getBaseFee());
 
+      operationTracer.traceBeforeRewardTransaction(worldUpdater, transaction, coinbaseWeiDelta);
```
fab-10, line 486: this seems unrelated

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/parallelization/ParallelizedConcurrentTransactionProcessor.java
```diff
@@ -0,0 +1,268 @@
...
+ * Optimizes transaction processing by executing transactions in parallel within a given block.
+ * Transactions are executed optimistically in a non-blocking manner. After execution, the class
+ * checks for potential conflicts among transactions to ensure data integrity before applying the
+ * results to the world state.
+ */
+@SuppressWarnings({"unchecked", "rawtypes"})
+public class ParallelizedConcurrentTransactionProcessor {
+
+  private static final int NCPU = Runtime.getRuntime().availableProcessors();
+  private static final Executor executor = Executors.newFixedThreadPool(NCPU);
```
fab-10, line 55: will be nice to have this part of EthScheduler, and use MonitoredExcecutor for metrics

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/parallelization/TransactionCollisionDetector.java
```diff
@@ -0,0 +1,108 @@
...
+        getAddressesTouchedByTransaction(
+            transaction, Optional.of(parallelizedTransactionContext.transactionAccumulator()));
+    if (addressesTouchedByTransaction.contains(miningBeneficiary)) {
+      return true;
+    }
+    final Set<Address> addressesTouchedByBlock =
+        getAddressesTouchedByBlock(Optional.of(blockAccumulator));
+    final Set<Address> commonAddresses = new HashSet<>(addressesTouchedByTransaction);
+    commonAddresses.retainAll(addressesTouchedByBlock);
+    return !commonAddresses.isEmpty();
```
fab-10, line 54: little opt: you can avoid to create a new Set and just iterate on `addressesTouchedByTransaction` and return `true` asa one address if found in the other set

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/accumulator/DiffBasedWorldStateUpdateAccumulator.java
```diff
@@ -357,12 +471,6 @@ public void commit() {
                 pendingStorageUpdates.clear();
               }
 
-              // This is especially to avoid unnecessary computation for withdrawals and
```
fab-10, line 360: is this related to the feature? is so why?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/parallelization/ParallelizedConcurrentTransactionProcessor.java
ahamlat: That was my first intent, but I had to give up because of some cyclic dependencies. I wanted to make EthScheduler part of util module to be able to reference it from everywhere in the code, but then I realized that EthScheduler is also tied to EthTask. I think we need to review or split EthScheduler and have a thread pool framework that we can we can use/reuse from utility module like util.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
ahamlat: This was a hint used by Karim to handle the miningBeneficiary case. He used the tracing feature to access the context of miningBeneficiary in ParallelizedConcurrentTransactionProcessor.runTransaction, and check if the address of the miner was touched before executing the transaction.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/accumulator/DiffBasedWorldStateUpdateAccumulator.java
matkt: There was a side effect because we manipulate and commit the Accumulator a lot more because of the background thread. Thanks to this we saw a bug that impacted the performance of this PR and also Besu in general. This line of code prevents the purge of entries of the map which are useless because there is a clean that was no longer done. This results in larger map to check and an impact on performance

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
matkt: Yes it is necessary for the PR and for conflict detection. I explain it in the comments normally

crypto/algorithms/src/test/java/org/hyperledger/besu/crypto/Blake2bfMessageDigestTest.java
matkt: I had tested without and with the fix and I was able to reproduce it every time but if you have a safer solution it will be better

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/parallelization/TransactionCollisionDetector.java
matkt: Yes good point . @ahamlat If you can do it. I could review

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
matkt: In fact I use it to know if we touched the address of the beneficiary before paying the reward. If we touch it before it is a conflict if we touch it just for the reward it's ok

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/parallelization/TransactionCollisionDetector.java
ahamlat: Yes, will do

besu/src/main/java/org/hyperledger/besu/cli/options/stable/DataStorageOptions.java
ahamlat: done.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/parallelization/TransactionCollisionDetector.java
ahamlat: done.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
garyschulte: Sorry for the late feedback here, but wouldn't paying the reward to the coinbase affect consensus?  If one of the parallel transactions was _from_ the coinbase, couldn't its result be affected by interim transaction rewards if the coinbase submitted a transaction that invoked BALANCE opcode?

I think we need to flatly ban coinbase from parallelization for this reason.

garyschulte (changes requested):
I think transactions sent from a coinbase are a problem, and they should be banned from parallelization period.  

If I am wrong, I would love to see a test block that proved parallelized coinbase transactions which use BALANCE opcode are safe/deterministic.

matkt:
> I think transactions sent from a coinbase are a problem, and they should be banned from parallelization period.
> 
> If I am wrong, I would love to see a test block that proved parallelized coinbase transactions which use BALANCE opcode are safe/deterministic.

 A transaction that will interact with the coinbase to send, receive, or simply read or modify throughout the transaction will be in the accumulator at the time of this event https://github.com/hyperledger/besu/pull/7296#discussion_r1683250602.

 Therefore, before paying the reward, If we detect this coinbase access, it will necessarily be banned from parallelization, and that's why we have this code. We track the state of the accumulator before paying the reward because it allows us to detect these transactions that need the correct state of the coinbase, and we ban them. This allows us to differentiate a simple payment of transaction fees, which will not impact the transaction's behavior, from a transaction whose behavior will depend on the coinbase. Then we pay the rewards in the sequential part to have the correct coinbase balance for the start of each transaction. Thus, if a transaction is banned due to its access to the coinbase, it will no longer have a problem when processed sequentially because the balance of a coinbase has been correctly updated by previous transactions during the sequential part

Here is the part of the code where we ban transactions that have access to coinbase https://github.com/hyperledger/besu/pull/7296/files#diff-53805bb7138bc804441822c2d6acca3a1c72a26926563bc898b28eec5654c812R48

So unless I did not understand your remark for me it is already managed and was precisely an important element of the PR

garyschulte:
> So unless I did not understand your remark for me it is already managed and was precisely an important element of the PR

AFAIK, we do not need a transaction to directly interact with the coinbase in order to alter its state.

My thinking is that each transaction pays coinbase some bit at its conclusion.  In a serial execution context this just means that any future transactions in the block would see the coinbase with an increased balance.

If in that same block, the coinbase is the originator of a transaction, nothing else would conflict with it because we are not considering coinbase reward as a conflict.  If the coinbase transaction attempts to read its balance in isolation, without the tx rewards that might/would have accumulated before it executed its transaction, then a parallel coinbase tx would report a different value when executing the BALANCE opcode than if it had been executed serially.

garyschulte:
> > So unless I did not understand your remark for me it is already managed and was precisely an important element of the PR
> 
> AFAIK, we do not need a transaction to directly interact with the coinbase in order to alter its state.

talking oob with Karim, he pointed out that the reason I didn't see the coinbase conflict detection is because I was searching for "coinbase", and the check is using "miningBeneficiary":
https://github.com/hyperledger/besu/pull/7296/files#diff-53805bb7138bc804441822c2d6acca3a1c72a26926563bc898b28eec5654c812R48

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/parallelization/ParallelizedConcurrentTransactionProcessor.java
```diff
@@ -0,0 +1,268 @@
...
+                   * it means that the value is necessary for the proper execution of the transaction and will therefore be considered in collision detection.
+                   * If this is not the case, we can ignore this address during conflict detection.
+                   */
+                  if (transactionCollisionDetector
+                      .getAddressesTouchedByTransaction(
+                          transaction, Optional.of(roundWorldStateUpdater))
+                      .contains(miningBeneficiary)) {
+                    contextBuilder.isMiningBeneficiaryTouchedPreRewardByTransaction(true);
+                  }
+                  contextBuilder.miningBeneficiaryReward(miningReward);
```
garyschulte, line 171: Running gradle check with code coverage doesn't show this covered.  Definitely would be nice to have this covered in test

fab-10 (commented):
The idea behind this feature make sense, and as first step looks good to me, and I have nothing else to add, just waiting for the approval from @garyschulte since he has much more knowledge on this part.

garyschulte (approved):
Lets merge and continue to test :+1:

Issue #6612, opened 2024-02-26
remove deprecated sync modes X_SNAP and X_CHECKPOINT


macfarla:
First we need to get some external docs updated 

eg somer's guide

https://someresat.medium.com/guide-to-staking-on-ethereum-ubuntu-lighthouse-773f5d982e03

These sync modes are deprecated as of 24.2.0

non-fungible-nelson:
@macfarla

macfarla:
looks like somer's guide is updated - this would be a good first issue!

PR #7309, opened 2024-07-12
6612: Remove deprecated sync modes and related helper methods

Matilda-Clerke:
## PR description
Remove deprecated sync modes and related helper methods

## Fixed Issue(s)

https://github.com/hyperledger/besu/issues/6612


Matilda-Clerke:
Relevant docs needing changing: https://besu.hyperledger.org/public-networks/how-to/bonsai-limit-trie-logs#unable-to-change-the-sync-mode

pinges (approved):
LGTM

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -361,7 +359,7 @@ public BesuControllerBuilder fromGenesisFile(
       // wrap with TransitionBesuControllerBuilder if we have a terminal total difficulty:
       if (configOptions.getTerminalTotalDifficulty().isPresent()) {
         // Enable start with vanilla MergeBesuControllerBuilder for PoS checkpoint block
-        if (isCheckpointSync(syncMode) && isCheckpointPoSBlock(configOptions)) {
+        if (syncMode == SyncMode.CHECKPOINT && isCheckpointPoSBlock(configOptions)) {
```
siladu, line 362:
```suggestion
        if (SyncMode.CHECKPOINT == syncMode  && isCheckpointPoSBlock(configOptions)) {
```

nit: this way around avoids a null pointer if syncMode were ever to be null (which shouldn't happen to be honest)

jframe (commented):
Can you add an entry in the changelog, in the breaking changes section mentioning that the X_SNAP and X_CHECKPOINT options have now been removed and to use SNAP and CHECKPOINT instead

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
siladu, line 1: nit: same point about the NullPointers

siladu (approved):
LGTM - just a non-blocking comment about NPEs

pinges:
I guess a CHANGELOG entry would be good!!

siladu:
> I guess a CHANGELOG entry would be good!!

yeh, that's fine to do in another PR.

macfarla:
I think the docs changes have already been done for this, if so can remove the label

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
Matilda-Clerke: I thought if syncMode were null, it should just evaluate `null == SyncMode.CHECKPOINT`, which should be ok and return false as expected, right?

siladu: Sorry, I was getting confused with `nullableField.equals(value)` idiom throwing a null pointer. 
I should have suggested SyncMode.CHECKPOINT.equals(syncMode). 

You're right, both will return false so it's not a big deal, mostly stylistic.

bgravenorst:
Doc changes implemented here: https://github.com/hyperledger/besu-docs/commit/21271eb8341a8f1572427d4ddf72322befc446c9

PR #7310, opened 2024-07-12
removed PKI backed QBFT

### LINKED ISSUE #7163: remove PKI backed QBFT code
was officially deprecated in 24.5

Feature can be removed in July release - 24.7

https://github.com/hyperledger/besu/pull/6979

macfarla:
## Fixed Issue(s)
fixes #7163


fab-10 (approved):
🚮

acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/configuration/BesuNodeConfiguration.java
```diff
@@ -275,10 +271,6 @@ public Optional<KeyPair> getKeyPair() {
     return keyPair;
   }
 
-  public Optional<PkiKeyStoreConfiguration> getPkiKeyStoreConfiguration() {
-    return pkiKeyStoreConfiguration;
-  }
-
   public boolean isStrictTxReplayProtectionEnabled() {
```
jframe, line 274: The p2pTLSEnabled method can be removed and also the create TLS methods in BesuNodeFactory

jframe:
pki module can be removed as well

jframe:
Is this PR also removing the p2p tls support?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/QbftContext.java
```diff
@@ -18,38 +18,21 @@
...
-
   /**
    * Instantiates a new Qbft context.
    *
    * @param validatorProvider the validator provider
    * @param epochManager the epoch manager
    * @param blockInterface the block interface
-   * @param pkiBlockCreationConfiguration the pki block creation configuration
    */
   public QbftContext(
```
jframe, line 32: nit: QBFT could be changed to use the BftContext instead, there was only a QBFT specific context because of PKI block creation. But this would be in a lot place so probably doesn't make sense in this PR

macfarla:
> Is this PR also removing the p2p tls support?

I think if we're removing that, it could be a separate PR - this one is big enough already

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/QbftContext.java
macfarla: ok yeah that could be a separate PR also

PR #7312, opened 2024-07-12
Improve blob size transaction selector

fab-10:
## PR description

The main goal of this PR is to make clearer the reason a blob tx is not selected during block creation, with the introduction of 2 specifics not selected results:

- BLOBS_FULL = The block already contains the max number of allowed blobs.
- TX_TOO_LARGE_FOR_REMAINING_DATA_GAS = The transaction has not been selected since there is not enough remaining data gas in the block to fit the blobs of the tx.

and to achieve that it was clearer to extract the blob tx selection login in its own class `BlobSizeTransactionSelector`, while doing that a small bug was fixed, since data gas used was added to gas used when computing if a block was full, while these 2 gas types are independent, [see here for details](https://github.com/hyperledger/besu/pull/7312/files#r1677540274).

### DIFF
--- a/plugin-api/src/main/java/org/hyperledger/besu/plugin/data/TransactionSelectionResult.java
+++ b/plugin-api/src/main/java/org/hyperledger/besu/plugin/data/TransactionSelectionResult.java
@@ -53,6 +53,7 @@ public class TransactionSelectionResult {
   private enum BaseStatus implements Status {
     SELECTED,
     BLOCK_FULL(true, false),
+    BLOBS_FULL(false, false),
     BLOCK_OCCUPANCY_ABOVE_THRESHOLD(true, false),
     BLOCK_SELECTION_TIMEOUT(true, false),
     TX_EVALUATION_TOO_LONG(true, true),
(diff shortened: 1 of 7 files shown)


plugin-api/src/main/java/org/hyperledger/besu/plugin/data/TransactionSelectionResult.java
```diff
@@ -53,9 +53,10 @@ protected interface Status {
   private enum BaseStatus implements Status {
     SELECTED,
     BLOCK_FULL(true, false),
+    BLOBS_FULL(false, false),
     BLOCK_OCCUPANCY_ABOVE_THRESHOLD(true, false),
     BLOCK_SELECTION_TIMEOUT(true, false),
-    TX_EVALUATION_TOO_LONG(false, true),
```
fab-10, line 58: we can safely stop selected for this selection results, since it means that the tx took more that the block selection time out to evaluate, and so does not make sense to continue the evaluation

Gabriel-Trintinalia:
Can we use `blob gas` instead of `data gas`? for the variables and transaction selection reasons?

macfarla:
from reading the description this PR is doing 4 very excellent things, but would be easier to review as several smaller PRs

fab-10:
> from reading the description this PR is doing 4 very excellent things, but would be easier to review as several smaller PRs

Fair point, splitting in 3 PRs:

first one [Stop transaction selection on TX_EVALUATION_TOO_LONG](https://github.com/hyperledger/besu/pull/7330)

fab-10:
> Can we use the term `blob gas` instead of `data gas`? for the variables and transaction selection reasons?

I also like more *blob gas* because is more common, and I will rename to it, but in many other places we have *data gas* like the `DataGas` type, so it is up to another PR to align the naming everywhere

macfarla (approved):

PR #7313, opened 2024-07-12
Make maintained PoA bootnode connections configurable

matthew1001:
## PR description
This PR follows on from https://github.com/hyperledger/besu/pull/7257. Specifically it makes the behaviour configurable and defaults to `off`. Since the `25.7.0` release has just gone out this PR will technically be a change in behaviour, so I'll add a change log entry accordingly. However, assuming this PR is included in an upcoming release I don't imagine any users will be affected by the behaviour change.

In an enterprise network with a large number of nodes, maintaining connections to the bootnodes could be problematic for the bootnodes. Defaulting the behaviour to `off` avoids the need to specifically disable the behaviour on all of the nodes, but allows a smaller chain to be configured to still treat bootnodes like static nodes.


matthew1001:
After discussion with @pinges we've decided to close this and revert https://github.com/hyperledger/besu/pull/7257 (the latter has gone into `24.7.0` so will need a breaking change entry). https://github.com/hyperledger/besu/pull/7314 provides the behaviour change that is most important - namely that bootnodes can be retried during peer table refresh - so that PR will merge pending approval.

PR #7314, opened 2024-07-12
Add new config option to use bootnodes during any peer table refresh, not just the first one

matthew1001:
## PR description
This PR adds the option in PoA networks only to query the bootnodes whenever the peer table is being refreshed.

The behaviour today is that the bootnodes are queried on first startup, but if no peers are found on that occasion Besu will not use the bootnode again until it is restarted.

The new option is `on` by default. Setting `--poa-discovery-retry-bootnodes` disables it.

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
@@ -143,6 +143,7 @@ public class PeerDiscoveryController {
   private final AtomicBoolean peerTableIsDirty = new AtomicBoolean(false);
   private OptionalLong cleanTableTimerId = OptionalLong.empty();
   private RecursivePeerRefreshState recursivePeerRefreshState;
+  private final boolean includeBootnodesOnPeerRefresh;
 
   private PeerDiscoveryController(
       final NodeKey nodeKey,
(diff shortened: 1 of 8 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -483,7 +486,23 @@ RecursivePeerRefreshState getRecursivePeerRefreshState() {
...
+    final List<DiscoveryPeer> initialPeers;
+    if (includeBootnodesOnPeerRefresh) {
+      bootstrapNodes.stream()
+          .filter(p -> p.getStatus() != PeerDiscoveryStatus.BONDED)
+          .forEach(p -> p.setStatus(PeerDiscoveryStatus.KNOWN));
+
+      // If configured to retry bootnodes during peer table refresh, include them
+      // in the initial peers list.
+      initialPeers =
+          Stream.concat(
```
pinges, line 499: we could just do a `initialPeers.addAll(bootstrapNodes)` here if we were to do line 504 before the if.

ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -483,7 +486,23 @@ RecursivePeerRefreshState getRecursivePeerRefreshState() {
...
+                  bootstrapNodes.stream(),
+                  peerTable.nearestBondedPeers(Peer.randomId(), 16).stream())
+              .collect(Collectors.toList());
+    } else {
+      initialPeers = peerTable.nearestBondedPeers(Peer.randomId(), 16);
+    }
```
pinges, line 505:
```suggestion
    final List<DiscoveryPeer> initialPeers = peerTable.nearestBondedPeers(Peer.randomId(), 16);
    if (includeBootnodesOnPeerRefresh) {
      bootstrapNodes.stream()
          .filter(p -> p.getStatus() != PeerDiscoveryStatus.BONDED)
          .forEach(p -> p.setStatus(PeerDiscoveryStatus.KNOWN));

      // If configured to retry bootnodes during peer table refresh, include them
      // in the initial peers list.
      initialPeers.addAll(bootstrapNodes);
    }
```

pinges (approved):
I have added one suggestion. Otherwise LGTM.

PR #7336, opened 2024-07-17
Refactor - eliminate non-constant string concatenation from debug and trace

macfarla:
## PR description
use atDebug() and atTrace()


jframe (approved):
LGTM

besu/src/main/java/org/hyperledger/besu/Runner.java
```diff
@@ -253,7 +253,11 @@ private void waitForServiceToStop(final String serviceName, final SynchronousShu
     try {
       shutdown.await();
     } catch (final InterruptedException e) {
-      LOG.debug("Interrupted while waiting for service " + serviceName + " to stop", e);
+      LOG.atDebug()
```
fab-10, line 256: in this case, and in most of the cases below, you can use the more compact form:

```
LOG.debug("Interrupted while waiting for service {} to stop", serviceName, e);
```

that has the same effect and maybe better performance, since the `serviceName.toString()` is only called if DEBUG is enabled, the rule of thumb is to used `atDebug` and similar, when some computation needs to be done to get the string to log, otherwise the compact form is easier to read.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthScheduler.java
```diff
@@ -243,15 +243,18 @@ private <T> CompletableFuture<T> timeout(
...
       syncWorkerExecutor.shutdownNow();
       txWorkerExecutor.shutdownNow();
       scheduler.shutdownNow();
       servicesExecutor.shutdownNow();
       computationExecutor.shutdownNow();
       shutdown.countDown();
     } else {
-      LOG.trace("Attempted to stop already stopped " + getClass().getSimpleName());
+      LOG.atTrace()
+          .setMessage("Attempted to stop already stopped ")
```
fab-10, line 255:
```suggestion
          .setMessage("Attempted to stop already stopped {}")
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapProtocolManager.java
```diff
@@ -123,18 +126,23 @@ public void processMessage(final Capability cap, final Message message) {
               .dispatch(new EthMessage(ethPeer, requestIdAndEthMessage.getValue()))
               .map(responseData -> responseData.wrapMessageData(requestIdAndEthMessage.getKey()));
     } catch (final RLPException e) {
-      LOG.debug(
-          "Received malformed message {} , disconnecting: {}", messageData.getData(), ethPeer, e);
+      LOG.atDebug()
```
fab-10, line 129: this is a perfect example, there is no overhead in the old statement, and it is more compact and easier to read

fab-10 (commented):
The code is ok, but I suggest to use the compact form where possible, as described in the comments

PR #7340, opened 2024-07-17
[DO NOT MERGE] Use besu-untuned to execute the acceptance tests

ahamlat:
## PR description
Use besu-untuned to execute different tests that need to run a besu node in a different process.
Also, there is no need to apply the JVM tuning related to GC and continuous profiling to this kind of tests.

image


macfarla (approved):

macfarla:
It's different test failures each time but it seems specific to this PR (eg CLIOptions tests)

ahamlat:
Yes, it seems to be related to this PR, I tried different executions but I had different issues especially the timeout one. My interpretation is that with besu sh, the acceptance tests are running faster. The only difference between besu and besu-untuned is GC flags and JFR continuous profiling.
I will continue to investigate.

PR #7337, opened 2024-07-17
Make TrieLogPruner.preload async

siladu:
Signed-off-by: Simon Dudley <simon.dudley@consensys.net>

## Fixed Issue(s)

Fixes #7322

### Thanks for sending a pull request! Have you done the following?

- [ ] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [ ] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [ ] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [ ] unit tests: `./gradlew build`
- [ ] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

PR #7350, opened 2024-07-18
Optimise TrieLogPruner.preload

siladu:
These optimisations only seem to save <= 17% of the time.

Control node took ~6mins.
Test node took between 5 mins and 5 mins 30 seconds on a couple of runs.

---

The problem with https://github.com/hyperledger/besu/pull/7337 is that for the duration of the async load, block import time appears to be affected. This could lead to impaired attestation performance and a missed proposal (although admittedly better than the alternative of downtime).

Ultimately need another PR that used a reasonable timeout for this load+prune operation, maybe 30 seconds.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/TrieLogPruner.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/TrieLogPruner.java
@@ -95,20 +95,22 @@ public class TrieLogPruner implements TrieLogEvent.TrieLogObserver {
     try (final Stream<byte[]> trieLogKeys = rootWorldStateStorage.streamTrieLogKeys(loadingLimit)) {
       final AtomicLong count = new AtomicLong();
       final AtomicLong orphansPruned = new AtomicLong();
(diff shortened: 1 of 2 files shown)


siladu:
Wall clock profiling of 
* control - dev-clc-bu-tk-mainnet-mainnet-archive-simon-test-snapshot4
* test - dev-clc-bu-tk-mainnet-mainnet-archive-simon-test-snapshot5

[besu_7350_wallclock.tar.gz](https://github.com/user-attachments/files/16282397/besu_7350_wallclock.tar.gz)

ahamlat:
The parallelization didn't work very well. For example , the main thread didn't execute any of the tasks, it was just waiting for the other threads from ForkjoinPool.commonPool, which is not the expected behaviour. The main thread with (numCPU - 1) other threads from ForkjoinPool.commonPool should do the work in a proportional way.
In the profiling, in addition to the main thread, I see also that all the ForkjoinPool.commonPool threads are not used, ex [ForkJoinPool.commonPool-worker-2 tid=37631].
The iterator could be the issue, I think you need to review the way you access database, you don't need RocksDB iterator if you can all the keys.

siladu:
Superceded by https://github.com/hyperledger/besu/pull/7365

PR #7353, opened 2024-07-19
7732: Remove datagas related stuff in favour of blobgas

Matilda-Clerke:
## PR description
Remove datagas related stuff in favour of blobgas


fab-10:
Have the official specs changed? if so please add a link to it

Matilda-Clerke:
I took at look at some EIPs and found EIP-4844 only mentions blob gas, and doesn't mention data gas at all: https://eips.ethereum.org/EIPS/eip-4844#gas-accounting
I'm not sure on the history of such fields, but it appears this EIP is when they should have been introduced.

macfarla (approved):
looks like the field was renamed a while back, but we had the old names in there until corresponding reference tests were updated. If the ref tests are passing, we can safely remove the old names now.

Gabriel-Trintinalia:
@fab-10 
https://github.com/ethereum/EIPs/pull/7354

PR #7359, opened 2024-07-22
Penalize invalid transient pending transactions in the layered transaction pool

fab-10:
## PR description

This PR improve the layered txpool, avoiding that invalid transient* pending txs will be continuously evaluated for block inclusion, potentially preventing the evaluation of other valid pending txs that rank behind the former one.

The goal is simple, just penalize invalid transient pending txs, pushing then down in the rank, after all the other, so they will have a chance to be evaluated, but only after all the non-penalized, or less penalized, pending txs have been evaluated.
To achieve this, the `PendingTransaction` now has a score, that start with the max value (127) and it is decremented on every penalization, and that score is now taken in consideration when sorting pending transaction in the layers.

* invalid transient pending tx means that the tx cannot be included now in the block, but it could be in a future block, because it has a transient issue, for example not enough balance in the sender wallet, or gas price below the min and so on, all conditions that could resolve in the future.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractPrioritizedTransactions.java
```diff
@@ -137,6 +141,13 @@ protected void internalRemove(
     orderByFee.remove(removedTx);
   }
 
+  @Override
+  protected void internalPenalize(final PendingTransaction penalizedTx) {
+    orderByFee.remove(penalizedTx);
```
jflo, line 146: TreeSet is not synchronized, so multiple subclasses could experience race issues. Suggest either making it sync or making this method sync.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/LayeredPendingTransactions.java
```diff
@@ -314,55 +316,80 @@ public synchronized List<Transaction> getPriorityTransactions() {
...
+        () -> {
+          invalidTransactions.forEach(
+              invalidTx -> {
+                synchronized (this) {
+                  prioritizedTransactions.remove(invalidTx, INVALIDATED);
+                }
+              });
+          penalizedTransactions.forEach(
+              penalizedTx -> {
+                synchronized (this) {
```
jflo, line 388: If this is the preferred means to address the sync issue I outlined above, then I suppose that is fine. Might want a javadoc on that class indicating that it requires external syncronization.

fab-10: same as above, do you think the package javadoc is enough or we need also something here?

jflo: I do think we also need something here, there are not many good paths that will get a confused devs to the package level javadoc.

jflo (approved):

Issue #7322, opened 2024-07-15
TrieLogPruner.preloadQueue Performance Issue


siladu:
First alerted when Discord user **ktmkancha** mentioned they upgraded to Besu 24.6.0 (which enabled bonsai-limit-trie-log-enabled by default) and TrieLogPruner was taking 45mins+ to load. It is suspected this user hadn't pruned their trie log backlog and is it not known how much trie log data they have.

A second instance of this was reported to me by @gfukushima when upgrading his node, `dev-clc-bu-tk-mainnet-mainnet-archive-bkp` which contained 112 GiB (47 GiB of which is SST files, 64 GiB is BlobDB files):
```
| Column Family                  | Keys            | Total Size  | SST Files Size  | Blob Files Size  |
|--------------------------------|-----------------|-------------|-----------------|------------------|
| TRIE_LOG_STORAGE               | 1564331         | 112 GiB     | 47 GiB          | 64 GiB           |
```

The time taken was ~45 mins...

```
{"@timestamp":"2024-07-11T23:45:46,224","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loading first 30000 trie logs from database...","throwable":""}
{"@timestamp":"2024-07-12T00:32:30,543","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loaded 30000 trie logs from database","throwable":""}
```

Subsequent restarts of this node took ~2 mins to load...
```
{"@timestamp":"2024-07-13T03:16:19,846","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loading first 30000 trie logs from database...","throwable":""}
{"@timestamp":"2024-07-13T03:18:20,768","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loaded 30000 trie logs from database","throwable":""}
```

Note, that the first restart pruned ~30_000 trie logs so the total to load reduced by ~2% (from 1,564,331 keys to 1,534,331).

I did some further testing of another canary, `prd-elc-besu-lighthouse-mainnet-nightly-bonsai-snap` which contained 70 GiB of trie logs (majority of which is BlobDB files)...
```
|--------------------------------|-----------------|-------------|-----------------|------------------|
| Column Family                  | Keys            | Total Size  | SST Files Size  | Blob Files Size  |
|--------------------------------|-----------------|-------------|-----------------|------------------|
| TRIE_LOG_STORAGE               | 954612          | 70 GiB      | 42 MiB          | 70 GiB           |
[...]

PR #7365, opened 2024-07-23
TrieLogPruner preload with 30 second timeout

siladu:
## PR description

The TrieLogPruner preload operation executes during startup. It streams a batch of trie logs from rocksdb, looks up their block headers to get a block number and adds them to the prune queue. Finally it triggers a prune operation on the prune queue, which may prune some of these trie logs if they are eligible - note pruning from the queue is fast.

The preload is executed synchronously on the main thread. Under certain conditions, this can take a long time, particularly when the trie log column family contains a large number of keys.

This PR wraps the whole operation in a 30 second timeout to avoid startup hangs.

**It is still synchronous and blocks the main startup thread** because doing this async (as explored in https://github.com/hyperledger/besu/pull/7337) affects the performance (especially disk I/O) of the normal operation of the node: block import and potentially block proposal.

In addition, the default pruning window size is reduced from 30_000  to 5_000 which should be sufficient to cover any pruning gaps due to maintenance downtime, while reducing the impact of the performance issue.

The expectation is that if the user has a backlog of trie logs, they will follow the guide to use the prune subcommand. As such, a log has been added which will be the last log line printed before the 30 second timeout window expires.

Here's an example from before this PR of a 2.5minute load...
```
{"@timestamp":"2024-07-23T05:56:27,308","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loading first 5000 trie logs from database...","throwable":""}
{"@timestamp":"2024-07-23T05:59:01,297","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Loaded 5000 trie logs from database","throwable":""}
```

And now with this PR...
```
{"@timestamp":"2024-07-23T05:56:29,465","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Attempting to load first 5000 trie logs from database...","throwable":""}
{"@timestamp":"2024-07-23T05:56:29,466","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Trie log pruning will timeout after 30 seconds. If this is timing out, consider using `besu storage trie-log prune` subcommand, see https://besu.hyperledger.org/public-networks/how-to/bonsai-limit-trie-logs","throwable":""}
{"@timestamp":"2024-07-23T05:56:59,466","level":"WARN","thread":"pool-8-thread-1","class":"TrieLogPruner","message":"Timeout occurred while loading and processing 5000 trie logs from database","throwable":""}
{"@timestamp":"2024-07-23T05:56:59,536","level":"INFO","thread":"main","class":"TrieLogPruner","message":"Operation timed out, but still pruned 513 trie logs.","throwable":""}
```

## Fixed Issue(s)

https://github.com/hyperledger/besu/issues/7322


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/TrieLogPruner.java
```diff
@@ -83,38 +90,82 @@ public TrieLogPruner(
...
+          .log();
+
+      preloadQueue(timeoutOccurred, timeoutFuture);
+    }
+  }
+
+  private void preloadQueue(
+      final AtomicBoolean timeoutOccurred, final ScheduledFuture<?> timeoutFuture) {
+
     try (final Stream<byte[]> trieLogKeys = rootWorldStateStorage.streamTrieLogKeys(loadingLimit)) {
```
pinges, line 132: Would it make sense to start streaming at a random key? There is a method streamFromKey() in class KeyValueStorage that allows you to pass in the starting key.

pinges (approved):
I've left a question for you, but I think it looks good and adding the timeout is a good thing!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/TrieLogPruner.java
```diff
@@ -83,38 +90,82 @@ public TrieLogPruner(
...
+          "Added {} trie logs to prune queue. Commencing pruning of eligible trie logs...",
+          addToPruneQueueCount.intValue());
+      int prunedCount = pruneFromQueue();
+      LOG.atInfo().log("Pruned {} trie logs.", prunedCount);
     } catch (Exception e) {
-      LOG.error("Error loading trie logs from database, nothing pruned", e);
-      return 0;
+      if (e.getCause() != null && e.getCause() instanceof TimeoutException) {
+        int prunedCount = pruneFromQueue();
+        LOG.atInfo().log("Operation timed out, but still pruned {} trie logs.", prunedCount);
```
macfarla, line 165:
```suggestion
        LOG.atInfo().log("Pruned {} trie logs before stopping due to time limit.", prunedCount);
```

siladu: I've used the phrase "time out" in several logs (shown in description) so would like to keep it consistent

siladu: Logs without the metadata...

```
"TrieLogPruner","message":"Attempting to load first 5000 trie logs from database..."
"TrieLogPruner","message":"Trie log pruning will timeout after 30 seconds. If this is timing out, consider using `besu storage trie-log prune` subcommand, see https://besu.hyperledger.org/public-networks/how-to/bonsai-limit-trie-logs"
"TrieLogPruner","message":"Timeout occurred while loading and processing 5000 trie logs from database","throwable"
"TrieLogPruner","message":"Operation timed out, but still pruned 513 trie logs."
```

PR #7379, opened 2024-07-25
In process RPC service

fab-10:
## PR description

Expose a way to call RPC method in process, so they can be invoked by plugins without the overhead of the network (or native socket) and the json (de)serialization.

The interface is still fragile, because RPC method inputs and outputs are not typed, but just Object, so a next step will be to make them typed so plugins will discover at compile time that something is broken, but that will be address in another PR.

Do not be scared by the amount of file changed since many are due to the move of some classes in the plugin-api module.


usmansaleem (dismissed):
LGTM

fab-10:
replaced by #7395

PR #7390, opened 2024-07-29
Metrics for sync phases

pinges:
## PR description
This PR is adding metrics for the duration of the chain download and the world state download when doing snap or checkpoint sync.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java
```diff
@@ -142,7 +143,7 @@ public void stop() {
     synchronized (this) {
       if (running.compareAndSet(true, false)) {
         LOG.info("Stopping sync");
-        // Cancelling the world state download will also cause the chain download to be cancelled.
+        // Canceling the world state download will also cause the chain download to be cancelled.
```
macfarla, line 146: I had to look it up. US tends to favor the single "l" version ("canceling" and "canceled") but both are used. However, the double ll version is used in a few spots so if you want to do a find and replace I won't object but it could be a separate PR

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/FastWorldDownloadState.java
```diff
@@ -71,6 +74,10 @@ public synchronized boolean checkCompletion(final BlockHeader header) {
       // so they can give up waiting.
       notifyAll();
       LOG.info("Finished downloading world state from peers");
+
+      // stop the metrics timer for the world download
```
macfarla, line 78: I feel like this comment could be incorporated into either the log message above or the method call below

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -205,6 +215,9 @@ public TrailingPeerRequirements calculateTrailingPeerRequirements() {
   public CompletableFuture<Void> start() {
     if (running.compareAndSet(false, true)) {
       LOG.info("Starting synchronizer.");
+
+      overallSyncTimingContext = overallSyncTimer.startTimer();
```
siladu, line 219: Not sure the prometheus metrics are designed to time such a long task, could have implications on memory for example.

Was thinking a metric to record the events "sync started" and "sync finished" would be better, then can derive the duration after the fact, e.g. in the dashboard panel. Also, makes the metrics more flexible in case we want to combine the various events, e.g. for different sync phases.

siladu: Am starting to wonder if metrics is the right way to do this - maybe parsing the logs with Loki is a better way to go? Might be a bit more effort to set up though. Can you see any long term advantage of metrics over logs?

pinges: In my opinion it is better to use code (Metrics) than logs, because people might change logs when they are not aware that they are used in Grafana.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/FastWorldDownloadState.java
pinges: this has been removed

ethereum/eth/src/jmh/java/org/hyperledger/besu/ethereum/eth/sync/worldstate/WorldStateDownloaderBenchmark.java
```diff
@@ -120,7 +121,8 @@ public void setUpUnchangedState() {
             syncConfig.getWorldStateMaxRequestsWithoutProgress(),
             syncConfig.getWorldStateMinMillisBeforeStalling(),
             Clock.fixed(Instant.ofEpochSecond(1000), ZoneOffset.UTC),
-            metricsSystem);
+            metricsSystem,
+            SyncDurationMetrics.NO_METRICS_SYNC_DURATION_METRICS);
```
garyschulte, line 125: there seems to be quite a few of these NO_METRICS_SYNC_DURATION_METRICS, it might be useful to have an overloaded constructor that delegates to the main constructor with NO_METRICS...

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -205,6 +213,9 @@ public TrailingPeerRequirements calculateTrailingPeerRequirements() {
   public CompletableFuture<Void> start() {
     if (running.compareAndSet(false, true)) {
       LOG.info("Starting synchronizer.");
+
+      syncDurationMetrics.startTimer(SyncDurationMetrics.Labels.TOTAL_SYNC_DURATION);
```
garyschulte, line 217: does startTimer work as intended if it has been stopped?  Is it additive, or does it replace the prior metric values?

Thinking about the worldstate heal re-use of the synchronizer

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
```diff
@@ -271,7 +279,8 @@ public synchronized void reloadTrieHeal() {
   }
 
   public synchronized void startFlatDatabaseHeal(final BlockHeader header) {
-    LOG.info("Initiating the healing process for the flat database");
+    syncDurationMetrics.startTimer(SyncDurationMetrics.Labels.FLAT_DB_HEAL);
+    syncDurationMetrics.stopTimer(SyncDurationMetrics.Labels.CHAIN_DOWNLOAD_DURATION);
```
garyschulte, line 283: this is a surprising place to find this imo.  If a node starts and does not perform a worldstate sync/heal, will the chain download duration count upwards endlessly?  IIRC the chain download pipeline doesn't ever stop until the node stops.

Perhaps a better place to stop this timer would be as an InitialSyncListener in PipelineChainDownloader.

garyschulte (commented):
I had to look up the difference between Summary and Histogram timers, SimpleTimer seems to be a good name for this

Some comments, but generally LGTM

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DefaultSynchronizer.java
pinges: startTimer() only works the first time it is called. After the first start the call has no effect.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
pinges: You are right, that is surprising. The problem with the PipelineChainDownloader is that it is started and stopped multiple times. I have made clear in the PR description that these metrics are only for snap and cp sync. Fast sync is broken at the moment and we will probably deprecate it once snap works for private networks.
I have moved the stopTimer out of the startFlatDatabaseHeal, as you can still configure besu to not do it.

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.HashMap;
+
+/**
+ * This class manages the synchronization duration metrics for the Hyperledger Besu project. It
+ * provides methods to start and stop timers for various synchronization phases.
+ */
+public class SyncDurationMetrics {
+
+  /** A {@link SyncDurationMetrics} instance that does not record any metrics. */
+  public static final SyncDurationMetrics NO_METRICS_SYNC_DURATION_METRICS =
```
siladu, line 31:
```suggestion
  public static final SyncDurationMetrics NO_OP_SYNC_DURATION_METRICS =
```
?

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Creates a new {@link SyncDurationMetrics} instance.
+   *
+   * @param metricsSystem The {@link MetricsSystem} to use to record metrics.
+   */
+  public SyncDurationMetrics(final MetricsSystem metricsSystem) {
+    timer =
+        metricsSystem.createSimpleLabelledTimer(
+            BesuMetricCategory.SYNCHRONIZER, "sync_duration", "Time taken to sync", "name");
```
siladu, line 46:
```suggestion
            BesuMetricCategory.SYNCHRONIZER, "total_sync_duration", "Total time taken to sync", "name");
```

Would be nice to state in the description (or at least the comment) exactly what this is measuring in terms of the stages, e.g. chain download, heal, backwards sync, etc

siladu: What is the purpose of the "name" string here?

siladu: hmmm...is "sync_duration" actually a label for a group of metrics?

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+  public void stopTimer(final Labels label) {
+    OperationTimer.TimingContext context = timers.get(label.name());
+    if (context != null) {
+      context.stopTimer();
+    }
+  }
+
+  /** Enum representing the different synchronization phases. */
+  public enum Labels {
+    /** Total time taken to get into sync. */
```
siladu, line 72: Would be nice to state exactly what this is measuring in terms of the other stages, e.g. chain download, heal, backwards sync, etc. See other comment also

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+   * @param label The synchronization phase to stop the timer for.
+   */
+  public void stopTimer(final Labels label) {
+    OperationTimer.TimingContext context = timers.get(label.name());
+    if (context != null) {
+      context.stopTimer();
+    }
+  }
+
+  /** Enum representing the different synchronization phases. */
```
siladu, line 70: Would be helpful to indicate how these various durations might overlap as it could be quite confusing to interpret the metrics without this domain knowledge, see other comment about total_sync_duration.

I think this warrants user documentation as well - maybe a wider issue that explains the sync stages and how these metrics relate, an updated version of https://besu.hyperledger.org/public-networks/how-to/monitor/understand-metrics perhaps

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/PipelineChainDownloaderTest.java
```diff
@@ -69,13 +70,15 @@ public class PipelineChainDownloaderTest {
   public void setUp() {
     syncTarget = new SyncTarget(peer1, commonAncestor);
     syncTarget2 = new SyncTarget(peer2, commonAncestor);
+    final NoOpMetricsSystem noMetricsSystem = new NoOpMetricsSystem();
```
siladu, line 73:
```suggestion
    final NoOpMetricsSystem noOpMetricsSystem = new NoOpMetricsSystem();
```

ethereum/eth/src/jmh/java/org/hyperledger/besu/ethereum/eth/sync/worldstate/WorldStateDownloaderBenchmark.java
siladu: Agree.

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
pinges: actually sync_duration is the actual name of the metric, which then has multiple children, which we identify with a label.

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
pinges: I agree. I think I should update that for the sync duration timers!

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
pinges: It says what it is measuring :-) 
Total time to get into sync, From the start of the sync to the last thing that is happening (BWS).

siladu: That implies a knowledge of our sync stages which not everyone who reads this will have, but even then, it doesn't include BWS which is surprising, I think that's worth mentioning at least.

Most important this is explained in the user docs I guess, devs can always refer to that, it's just convenient when you're reading the code IMO

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+  public void stopTimer(final Labels label) {
+    OperationTimer.TimingContext context = timers.remove(label.name());
+    if (context != null) {
+      context.stopTimer();
+    }
+  }
+
+  /** Enum representing the different synchronization phases. */
+  public enum Labels {
+    /** Total time taken to get into sync. */
```
siladu, line 72:
```suggestion
    /** Total time taken to get into sync. It is useful for SNAP and CHECKPOINT sync-modes only. 
    
    Total sync duration includes the separate stages mentioned below, some of which occur in parallel. 
    
    Total sync duration excludes the backwards sync stage due to implementation challenges. The backwards sync should be a very short duration following the other sync stages.
      */
```

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+    TOTAL_SYNC_DURATION,
+    /** Time taken to download the chain data (headers, blocks, receipts). */
+    CHAIN_DOWNLOAD_DURATION,
+    /** Time taken to download the initial world state, before the healing step. */
+    SNAP_INITIAL_WORLD_STATE_DOWNLOAD_DURATION,
+    /** Time taken to heal the world state, after the initial download. */
+    SNAP_WORLD_STATE_HEALING_DURATION,
+    /** Time taken to do the fast world state download. */
+    FAST_WORLD_STATE_DOWNLOAD_DURATION,
```
siladu, line 81: unused now?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
```diff
@@ -102,60 +108,53 @@ public SnapWorldDownloadState(
...
+    metricsSystem.createLongGauge(
+        BesuMetricCategory.SYNCHRONIZER,
+        "snap_world_state_pending_trie_node_requests_current",
+        "Number of trie node pending requests for snap sync world state download",
+        pendingTrieNodeRequests::size);
+    syncDurationMetrics.startTimer(
+        SyncDurationMetrics.Labels.SNAP_INITIAL_WORLD_STATE_DOWNLOAD_DURATION);
+    LOG.info(
+        "startTimer SNAP_INITIAL_WORLD_STATE_DOWNLOAD_DURATION: {}",
+        LocalDateTime.now(ZoneId.systemDefault()));
```
pinges, line 157:
```suggestion
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
```diff
@@ -271,7 +280,7 @@ public synchronized void reloadTrieHeal() {
   }
 
   public synchronized void startFlatDatabaseHeal(final BlockHeader header) {
-    LOG.info("Initiating the healing process for the flat database");
+    syncDurationMetrics.startTimer(SyncDurationMetrics.Labels.FLAT_DB_HEAL);
```
pinges, line 283:
```suggestion
    LOG.info("Initiating the healing process for the flat database");
    syncDurationMetrics.startTimer(SyncDurationMetrics.Labels.FLAT_DB_HEAL);
```

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+      */
+    TOTAL_SYNC_DURATION,
+    /** Time taken to download the chain data (headers, blocks, receipts). */
+    CHAIN_DOWNLOAD_DURATION,
+    /** Time taken to download the initial world state, before the healing step. */
+    SNAP_INITIAL_WORLD_STATE_DOWNLOAD_DURATION,
+    /** Time taken to heal the world state, after the initial download. */
+    SNAP_WORLD_STATE_HEALING_DURATION,
+    /** Time taken to do the fast world state download. */
+    FAST_WORLD_STATE_DOWNLOAD_DURATION,
```
pinges, line 86:
```suggestion
```

metrics/core/src/main/java/org/hyperledger/besu/metrics/SyncDurationMetrics.java
pinges: Done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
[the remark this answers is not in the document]
pinges: Done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/SnapWorldDownloadState.java
```diff
@@ -271,7 +280,7 @@ public synchronized void reloadTrieHeal() {
   }
 
   public synchronized void startFlatDatabaseHeal(final BlockHeader header) {
-    LOG.info("Initiating the healing process for the flat database");
```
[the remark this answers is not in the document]
pinges: Done

siladu (approved):
LGTM - thanks for the changes!

PR #2754, opened 2021-09-13
Initial PoC for RPC end points via the plugin mechanism.

antonydenyer:
## PR description
re-implementation of https://github.com/PegaSysEng/pantheon/pull/1909/

## Changelog

- [ ] I thought about the changelog and included a [changelog update if required](https://wiki.hyperledger.org/display/BESU/Changelog).

PR #7395, opened 2024-07-29
In process RPC service

fab-10:
## PR description

Expose a way to call RPC method in process, so they can be invoked by plugins without the overhead of the network (or native socket) and the json (de)serialization.

The interface is still fragile, because RPC method inputs and outputs are not typed, but just Object, so a next step will be to make them typed so plugins will discover at compile time that something is broken, but that will be address in another PR.

Do not be scared by the amount of file changed since many are due to the move of some classes in the plugin-api module.

This feature is disabled by default and can be enabled setting the new experimental flag `Xin-process-rpc-enabled=true`, and to enabled specific RPC namespaces use the new experimental option `Xin-process-rpc-apis` with a comma separated list of namespaces.

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/Runner.java
+++ b/besu/src/main/java/org/hyperledger/besu/Runner.java
@@ -18,6 +18,7 @@ import org.hyperledger.besu.controller.BesuController;
 import org.hyperledger.besu.ethereum.api.graphql.GraphQLHttpService;
 import org.hyperledger.besu.ethereum.api.jsonrpc.EngineJsonRpcService;
 import org.hyperledger.besu.ethereum.api.jsonrpc.JsonRpcHttpService;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.JsonRpcMethod;
 import org.hyperledger.besu.ethereum.api.jsonrpc.ipc.JsonRpcIpcService;
 import org.hyperledger.besu.ethereum.api.jsonrpc.websocket.WebSocketService;
 import org.hyperledger.besu.ethereum.api.query.cache.AutoTransactionLogBloomCachingService;
@@ -39,6 +40,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
@@ -69,6 +71,7 @@ public class Runner implements AutoCloseable {
   private final Optional<EngineJsonRpcService> engineJsonRpc;
   private final Optional<MetricsService> metrics;
   private final Optional<JsonRpcIpcService> ipcJsonRpc;
+  private final Map<String, JsonRpcMethod> inProcessRpcMethods;
   private final Optional<Path> pidPath;
   private final Optional<WebSocketService> webSocketRpc;
   private final TransactionPoolEvictionService transactionPoolEvictionService;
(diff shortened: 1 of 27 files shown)


besu/src/main/java/org/hyperledger/besu/Runner.java
```diff
@@ -108,6 +112,7 @@ public class Runner implements AutoCloseable {
       final Optional<GraphQLHttpService> graphQLHttp,
       final Optional<WebSocketService> webSocketRpc,
       final Optional<JsonRpcIpcService> ipcJsonRpc,
+      final Map<String, JsonRpcMethod> inProcessRpcMethods,
```
siladu, line 115: Comparing to JsonRpcHttpService, I think in-process is missing metrics, e.g. requestTimer. 

Wonder If it's still useful to have separate metrics for these in-process RPC calls?

besu/src/main/java/org/hyperledger/besu/services/RpcEndpointServiceImpl.java
```diff
@@ -17,25 +17,42 @@
...
 import java.util.Locale;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /** The RPC endpoint service implementation. */
 public class RpcEndpointServiceImpl implements RpcEndpointService {
   private final Map<String, Function<PluginRpcRequest, ?>> rpcMethods = new HashMap<>();
+  private Map<String, JsonRpcMethod> inProcessRpcMethods;
```
siladu, line 42: Is this compatible with RPCs added via plugins as well, as per https://github.com/hyperledger/besu/pull/2754? 

I'm not 100% sure if this scenario makes sense, but perhaps we'd want to call a plugin's RPC method via another plugin, i.e.
* plugin1 registers rpc1 with besu
* plugin2 calls rpc1 in-process

besu/src/main/java/org/hyperledger/besu/services/RpcEndpointServiceImpl.java
```diff
@@ -48,6 +65,37 @@ public <T> void registerRPCEndpoint(
     rpcMethods.put(namespace + "_" + functionName, function);
   }
 
+  @Override
+  public PluginRpcResponse call(final String methodName, final Object[] params) {
```
siladu, line 69: Should we LOG.trace the methodName and params here?

siladu (approved):
LGTM. Minor comments and suggestions which could be done in follow up PRs if you decide.

Appreciate the good PR description, test coverage and the way you've broken up the PR 👍 

Great feature BTW, could be some interesting possibilities using this with the engine API.

besu/src/main/java/org/hyperledger/besu/Runner.java
fab-10: good points, I think it is useful to have metrics, will review this in a separate PR

besu/src/main/java/org/hyperledger/besu/services/RpcEndpointServiceImpl.java
fab-10: This a quick test, and found that it is possible to call a method defined in another plugin, but only if the plugin namespace is enabled via `rpc-http-api`, that's not optimal, since should be enough to enable it only using `Xin-process-rpc-apis`, will improve that on a following PR with proper tests

besu/src/main/java/org/hyperledger/besu/services/RpcEndpointServiceImpl.java
fab-10: done

PR #7389, opened 2024-07-28
5098: Add additional RpcErrorTypes for invalid parameters

Matilda-Clerke:
## PR description
Adds additional RpcErrorTypes to provide extra context around invalid parameters. I recommend starting review with changes in RpcErrorType and InvalidJsonRpcParameters, as these changes lay the foundation for the remainder of the changes.

## Fixed Issue(s)

https://github.com/hyperledger/besu/issues/5098

### Thanks for sending a pull request! Have you done the following?

- [ ] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [ ] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [ ] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [x] unit tests: `./gradlew build`
- [ ] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

PR #7405, opened 2024-07-30
5098 branch 4 update invalid address params

Matilda-Clerke:
## PR description
Update code to use RpcErrorType.INVALID_ADDRESS_PARAMS
## Fixed Issue(s)


consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/Discard.java
```diff
@@ -46,7 +48,13 @@ public String getName() {
   public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
     checkState(
         validatorProvider.getVoteProviderAtHead().isPresent(), "Clique requires a vote provider");
-    final Address address = requestContext.getRequiredParameter(0, Address.class);
+    final Address address;
+    try {
+      address = requestContext.getRequiredParameter(0, Address.class);
+    } catch (Exception e) { // TODO:replace with JsonRpcParameter.JsonRpcParameterException
+      throw new InvalidJsonRpcParameters(
+          "Invalid address parameter", RpcErrorType.INVALID_ADDRESS_PARAMS, e);
```
siladu, line 56: IIUC, there's two messages related to this error: 
1. The string here, "Invalid address parameter"
2. [The one hardcoded against INVALID_ADDESS_PARAMS (added the previous PR), "Invalid address params"](https://github.com/hyperledger/besu/blob/4e79c810a6c9e94db6168a46e769d82e747da226/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/RpcErrorType.java#L33)

Why do we need both and is there a reason we need all the extra RpcErrorTypes versus dynamically passing the string here?

consensus/clique/src/test/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/DiscardTest.java
```diff
@@ -121,7 +121,7 @@ public void discardWithoutAddress() {
     final Discard discard = new Discard(validatorProvider);
 
     assertThatThrownBy(() -> discard.response(requestWithParams()))
-        .hasMessage("Missing required json rpc parameter at index 0")
+        .hasMessage("Invalid address parameter")
```
siladu, line 124: Wondering if also knowing the parameter index could be useful for users in any of these RPC error? e.g. if you have multiple parameters with values that either are or could look like an Address (i.e. a hash)

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/Discard.java
```diff
@@ -46,7 +48,13 @@ public String getName() {
   public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
     checkState(
         validatorProvider.getVoteProviderAtHead().isPresent(), "Clique requires a vote provider");
-    final Address address = requestContext.getRequiredParameter(0, Address.class);
+    final Address address;
+    try {
+      address = requestContext.getRequiredParameter(0, Address.class);
+    } catch (Exception e) { // TODO:replace with JsonRpcParameter.JsonRpcParameterException
```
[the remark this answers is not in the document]
Matilda-Clerke: Yes, in order to replace these TODOs, it means throwing a checked exception from some common code, which would result in an extremely large PR like https://github.com/hyperledger/besu/pull/7389. One of the final PRs in the series will be to throw the checked exception, and action these TODOs.

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/Discard.java
Matilda-Clerke: There are times when the message can add some extra context, but for the most part it's redundant.

consensus/clique/src/test/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/DiscardTest.java
Matilda-Clerke: Hmm, could be useful yeah. I'll update for this PR and we can assess whether we want to apply it across all of the PRs

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/Discard.java
siladu: So how does this work, does one override the other or does it respond with both? 
Have you got an example where the message gives extra context?

Matilda-Clerke: To be honest, I don't really know the full context of how each message is used and where.

siladu: Updating for completeness...we figured out that this error message is displaying in a DEBUG log stack, whereas the one attached to INVALID_ADDRESS_PARAMS is what the user actually sees. At least in the example tested.

siladu (approved):
LGTM

PR #7406, opened 2024-07-31
5098 branch 5 update invalid auth params

Matilda-Clerke:
## PR description
Update code with usage of RpcErrorType.INVALID_AUTH_PARAMS
## Fixed Issue(s)


consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/Propose.java
```diff
@@ -57,7 +57,13 @@ public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
       throw new InvalidJsonRpcParameters(
           "Invalid address parameter (index 0)", RpcErrorType.INVALID_ADDRESS_PARAMS, e);
     }
-    final Boolean auth = requestContext.getRequiredParameter(1, Boolean.class);
+    final Boolean auth;
+    try {
+      auth = requestContext.getRequiredParameter(1, Boolean.class);
+    } catch (Exception e) { // TODO:replace with JsonRpcParameter.JsonRpcParameterException
+      throw new InvalidJsonRpcParameters(
+          "Invalid auth parameter (index 1)", RpcErrorType.INVALID_AUTH_PARAMS, e);
```
siladu, line 65: I believe this is the relevant RPC - https://besu.hyperledger.org/private-networks/reference/api#clique_propose

The second param is 
> proposal: boolean - true to propose adding signer or false to propose removing signer

Invalid auth seems unrelated at first glance. Will it used for anything else apart from this RPC? 
Maybe "Invalid proposal" instead?

Matilda-Clerke: Admittedly, I named the enumeration after the variable name storing the parameter. This makes much more sense. Thanks for the link too, I didn't realise these RPC calls were documented in an API.

siladu (approved):
LGTM

PR #7454, opened 2024-08-13
Correct default `--help` values

### LINKED ISSUE #7367: help for --rpc-max-trace-filter-range shows incorrect default
besu -h shows this 

      --rpc-max-trace-filter-range=<maxTraceFilterRange>
                             Specifies the maximum number of blocks for the trace_filter method. Must be >=0. 0 specifies no limit  (default: $DEFAULT-VALUE)

rather than the actual default value for --rpc-max-trace-filter-range

### LINKED ISSUE #7369: help does not show default value for `--genesis-state-hash-cache-enabled`
help does not show default value for `--genesis-state-hash-cache-enabled`

help message should shoe default value for this option as for other options when `besu -h` is called

7suyash7:
Fixes #7367 : Correct default value display for --rpc-max-trace-filter-range along with the requested unit test.

Fixes #7369 : Show default value for --genesis-state-hash-cache-enabled.


jframe (approved):
Thanks for your contribution.

Tested this locally and it's now displaying the default values correctly

PR #7457, opened 2024-08-13
5098 branch 16 update invalid enode and excess blob gas params

Matilda-Clerke:
## PR description
Update to use RpcErrorType.INVALID_ENODE_PARAMS and RpcErrorType.INVALID_EXCESS_BLOB_GAS_PARAMS
## Fixed Issue(s)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV2.java
```diff
@@ -64,7 +64,7 @@ protected ValidationResult<RpcErrorType> validateParameters(
     }
     if (payloadParameter.getExcessBlobGas() != null) {
       return ValidationResult.invalid(
-          RpcErrorType.INVALID_PARAMS, "non-null ExcessBlobGas pre-cancun");
+          RpcErrorType.INVALID_EXCESS_BLOB_GAS_PARAMS, "Missing excess blob gas field");
```
macfarla, line 67: do we want to keep the "pre-cancun" text

macfarla (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV2.java
Matilda-Clerke: Can do if you think it's beneficial? These messages are only seen in the logs though.

PR #7462, opened 2024-08-14
Precompile calls were not traced when insuficient gas

### LINKED ISSUE #7318: Erroneous stateDiff in API calls for failed transactions
### Description

The `stateDiff` field of the result of a pure transfer tx in `trace_replayBlockTransactions` call is `null` if the transaction failed. 

### Steps to Reproduce (Bug)

On an archive Besu node on the 

1. Have RPC access to an archive Besu node following Holesky
2. Replay [this tx's block](https://holesky.etherscan.io/tx/0xb457838013969202364745b586e4094c4edd1bd376244dac87c496eef777b6c7) and look at the result for the 16th tx with the following `curl`/`jq` command:

```
curl -X POST -H "Content-Type: application/json" --data '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "trace_replayBlockTransactions",
    "params": ["0x1D1921", ["stateDiff"]]
    }' http://localhost:8545 | jq ".result[15]"
```

**Expected behavior:** 

`stateDiff` should contain the balance changes for the tx sender and the block validator (here is the correct output as provided by QuickNode: 

```
{
  # This is the sender
  "0x5324ff21932b3ae69a6d46aaf29eca5bbfb86b73": {
    "balance": {
      "*": {
        "from": "0x5bb96a5487a77495",
        "to": "0x5bb9573b1619f045"
      }
    },
    "code": "=",
    "nonce": {
      "*": {
        "from": "0xe3",
        "to": "0xe4"
      }
    },
    "storage": {}
  },
  # This is the miner
  "0x9baa3244565d51d9c7897c0eb6679ed4890e536e": {
    "balance": {
      "*": {
        "from": "0xa39659f932cf1b4668b",
        "to": "0xa39659fa646633eb68b"
      }
    },
    "code": "=",
    "nonce": "=",
    "storage": {}
  }
}
```

**Actual behavior:** 

`stateDiff` is `null`.

**Frequency:**

100%

* Software version: freshly compiled from `main` branch, as well as 24.5.2
* Java version: 17 & 21
* OS Name & Version: Debian bookworm
* Consensus Client & Version if using Proof of Stake: Teku 24.4.0

It seems that this issue is triggered by simple transfer TXs running out of gas. The bug may be coming [from this line](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java#L41), where the generation of the state diff is entirely skipped if there are no frames in the tx execution, hence ignoring the balance changes to the sender & miner.

lu-pinto:
## PR description
This fixes a tracing problem while replaying a transaction that executes a precompiled contract but it runs out of gas. We were getting an empty `TraceFrame` array back from the tracer because the halt condition was never traced.

## Fixed Issue(s)

fixes #7318 

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java
@@ -38,7 +38,7 @@ public class StateDiffGenerator {
   public Stream<Trace> generateStateDiff(final TransactionTrace transactionTrace) {
     final List<TraceFrame> traceFrames = transactionTrace.getTraceFrames();
     if (traceFrames.isEmpty()) {
-      throw new RuntimeException("expected to have at least one processed frame");
+      return Stream.empty();
     }
 
     // This corresponds to the world state after the TX executed
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java
@@ -38,7 +38,7 @@ public class StateDiffGenerator {
   public Stream<Trace> generateStateDiff(final TransactionTrace transactionTrace) {
     final List<TraceFrame> traceFrames = transactionTrace.getTraceFrames();
     if (traceFrames.isEmpty()) {
-      return Stream.empty();
+      throw new RuntimeException("expected to have at least one processed frame");
     }
 
     // This corresponds to the world state after the TX executed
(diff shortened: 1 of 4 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/tracing/diff/StateDiffGenerator.java
```diff
@@ -38,7 +38,7 @@ public class StateDiffGenerator {
   public Stream<Trace> generateStateDiff(final TransactionTrace transactionTrace) {
     final List<TraceFrame> traceFrames = transactionTrace.getTraceFrames();
     if (traceFrames.isEmpty()) {
-      return Stream.empty();
+      throw new RuntimeException("expected to have at least one processed frame");
```
shemnon, line 41: This change in behavior is not essential to the precompile trace fix.  Please consider moving it to one of the other RPC PRs.

shemnon (approved):

PR #7465, opened 2024-08-14
Dagger plugin context

jflo:
Depends on PR #7464 , merge that one first.

Eliminates dependency from Controller on BesuComponent, replaces with a much narrower dependency onto BesuPluginContextImpl. This should make future dependency cycles much less likely.


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -865,40 +864,35 @@ static class PrivacyOptionGroup {
...
   private BesuConfigurationImpl pluginCommonConfiguration;
 
-  private BesuComponent besuComponent;
   private final Supplier<ObservableMetricsSystem> metricsSystem =
-      Suppliers.memoize(
-          () ->
-              besuComponent == null || besuComponent.getObservableMetricsSystem() == null
-                  ? MetricsSystemFactory.create(metricsConfiguration())
-                  : besuComponent.getObservableMetricsSystem());
+      Suppliers.memoize(() -> MetricsSystemFactory.create(metricsConfiguration()));
```
garyschulte, line 868: are we not getting metrics system from dagger anymore?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -865,40 +864,35 @@ static class PrivacyOptionGroup {
...
-      final BesuComponent besuComponent,
       final Supplier<RlpBlockImporter> rlpBlockImporter,
       final Function<BesuController, JsonBlockImporter> jsonBlockImporterFactory,
       final Function<Blockchain, RlpBlockExporter> rlpBlockExporterFactory,
       final RunnerBuilder runnerBuilder,
       final BesuController.Builder controllerBuilder,
       final BesuPluginContextImpl besuPluginContext,
-      final Map<String, String> environment) {
+      final Map<String, String> environment,
+      final Logger commandLogger) {
```
garyschulte, line 894: what is the motivation to inject this logger?

jflo: That one is configured [a little differently](https://github.com/hyperledger/besu/blob/a7e5548135d440bbbaa52441f24eef5dde9cb456/besu/src/main/java/org/hyperledger/besu/Besu.java#L55) than all other Loggers.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
jflo: Right, so this was what i meant by a "broad cycle" if BesuCommand depends on BesuComponent, it's gonna eventually depend on all the things, when in reality it just needed the plugin context.

There are four MetricsSystems we know how to create, Prometheus, OTEL and the NoOp and Stub systems used in all the testing. If a client is using a metrics system that is configurable, (Prometheus and OTEL) then Besu command MUST know how to create it, since the details needed come from BesuCommand.

So we never really needed that hack in the first place, and instead can just create a DaggerModule to provide the NoOp or Stub MetricsSystems should we need them in tests.

garyschulte (approved):
I don't have strong DI opinions, but this looks clean and straightforward. LGTM

PR #7468, opened 2024-08-15
Decouple block creators

jflo:
Builds on top of #7465 do not merge till that has been merged. 

decouples the parent header being built on from the construction of a block creator, instead supplying it as a parameter. There is no loss of immutability, and will allow for singleton block creators managed by Dagger in the future.

This will then lead to not needing BlockCreatorFactories, which were a pattern used to workaround this.


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreator.java
```diff
@@ -45,7 +45,7 @@ public PoWBlockCreator(
       final ProtocolContext protocolContext,
       final ProtocolSchedule protocolSchedule,
       final PoWSolver nonceSolver,
-      final BlockHeader parentHeader,
+      // final BlockHeader parentHeader,
```
macfarla, line 48: delete this comment?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreator.java
```diff
@@ -54,14 +54,15 @@ public PoWBlockCreator(
         transactionPool,
         protocolContext,
         protocolSchedule,
-        parentHeader,
+        // parentHeader,
```
macfarla, line 57: delete this comment?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/PoWMinerExecutor.java
```diff
@@ -93,7 +93,7 @@ public PoWBlockMiner createMiner(
                 protocolContext,
                 protocolSchedule,
                 solver,
-                parentHeader,
+                // parentHeader,
```
macfarla, line 96: delete this comment?

ethereum/referencetests/src/reference-test/external-resources
macfarla, line 1: accidental checkin?

ethereum/referencetests/build.gradle
```diff
@@ -232,7 +232,7 @@ tasks.register('validateReferenceTestSubmodule') {
   description = "Checks that the reference tests submodule is not accidentally changed"
   doLast {
     def result = new ByteArrayOutputStream()
-    def expectedHash = '9201075490807f58811078e9bb5ec895b4ac01a5'
+    def expectedHash = 'faf33b471465d3c6cdc3d04fbd690895f78d33f2'
```
macfarla, line 235: accidental checkin?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSchedule.java
```diff
@@ -116,7 +116,6 @@ public static ProtocolSchedule fromConfig(
    * @param config {@link GenesisConfigOptions} containing the config options for the milestone
    *     starting points
    * @param evmConfiguration size of
-   * @param miningParameters the mining parameters
```
macfarla, line 119: javadoc does not match method?

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeBlockCreator.java
```diff
@@ -61,7 +61,7 @@ public MergeBlockCreator(
         transactionPool,
         protocolContext,
         protocolSchedule,
-        parentHeader,
+        // parentHeader,
```
macfarla, line 64: delete this comment?

fab-10:
General question, could https://dagger.dev/dev-guide/assisted-injection.html be used to create new block creators passing the parent header?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -517,7 +534,7 @@ boolean rewardBeneficiary(
   }
 
   protected abstract BlockHeader createFinalBlockHeader(
-      final SealableBlockHeader sealableBlockHeader);
+      final SealableBlockHeader sealableBlockHeader, final Optional<BlockHeader> maybeParentHeader);
```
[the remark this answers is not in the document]
jflo: when it is not currently in the blockchain and needs to be requested from the network.

jflo:
> General question, could https://dagger.dev/dev-guide/assisted-injection.html be used to create new block creators passing the parent header?

yes, this is how prototype scoped instances should be created. I don't think it is appropriate in this case, but in another situation this could be a good move.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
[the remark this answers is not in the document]
fab-10: will review this scenario, since I want to have a better understanding of it

fab-10:
> > General question, could https://dagger.dev/dev-guide/assisted-injection.html be used to create new block creators passing the parent header?
> 
> yes, this is how prototype scoped instances should be created. I don't think it is appropriate in this case, but in another situation this could be a good move.

have you already tried to make block creators singletons? do you see any side effect?

jflo:
> > > General question, could https://dagger.dev/dev-guide/assisted-injection.html be used to create new block creators passing the parent header?
> > 
> > 
> > yes, this is how prototype scoped instances should be created. I don't think it is appropriate in this case, but in another situation this could be a good move.
> 
> have you already tried to make block creators singletons? do you see any side effect?

Yes, there will be a next branch based on this one that makes the block creators singletons, so that the mining parameters can be provided once and then retained.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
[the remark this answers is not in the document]
jflo: Yes it is unusual and unlikely, but i believe dictated by querying for the parent from the blockchain, which returns optional.

fab-10 (approved):

PR #7470, opened 2024-08-16
CLI option for disabling auto-registration of external plugins

Gabriel-Trintinalia:
## PR description

This pull request introduces a new flag to provide more granular control over plugin registration in Besu. While Besu currently allows specifying individual plugins to load, it does not offer an option to disable the automatic registration of external plugins. This new flag will enable users to disable external plugins, offering greater flexibility in managing plugin configurations.

### Usage

- **Enable external plugins (Default)**:
  ```bash
  besu --Xplugins-external-enabled=true
  ```

- **Disable external plugins**:
  ```bash
  besu --Xplugins-external-enabled=false
  ```

The new flag defaults to enabled, ensuring no breaking changes to existing setups. At this stage, the flag is kept hidden as it is experimental.


besu/src/main/java/org/hyperledger/besu/services/BesuPluginContextImpl.java
```diff
@@ -142,10 +142,11 @@ public void registerPlugins(final PluginConfiguration config) {
           matchAndValidateRequestedPlugins(requestedPlugins, detectedPlugins);
 
       registerPlugins(registeringPlugins);
-    } else {
-      // If no plugins were specified, register all detected plugins
+    } else if (config.isPluginAutoEnabled()) {
```
Matilda-Clerke, line 145: So, if pluginAutoEnable is true, but we also specify some requested plugins, it will only load the requested plugins? Could be a bit confusing if we don't explicitly explain this behavior in documentation.

Matilda-Clerke (approved):
LGTM assuming we're happy with the behavior I commented on.

besu/src/main/java/org/hyperledger/besu/services/BesuPluginContextImpl.java
Gabriel-Trintinalia: That is the flow:

```mermaid
graph TD
    B{--plugins specified?} -->|Yes| C[Load specified plugins]
    B -->|No| D{--Xplugin-auto-enabled?}
    D -->|Yes| E[Auto-register plugins]
    D -->|No| F[Do not register plugins]

File: besu/src/main/java/org/hyperledger/besu/cli/DefaultCommandValues.java [hunk 2]
```diff
@@ -128,6 +128,9 @@ public interface DefaultCommandValues {
   /** The constant DEFAULT_PLUGINS_OPTION_NAME. */
   String DEFAULT_PLUGINS_OPTION_NAME = "--plugins";
 
+  /** The constant DEFAULT_PLUGINS_AUTO_ENABLED_OPTION_NAME. */
+  String DEFAULT_PLUGINS_AUTO_ENABLED_OPTION_NAME = "--Xplugin-auto-enabled";
```
macfarla, line 132:
```suggestion
  String DEFAULT_PLUGINS_AUTO_ENABLED_OPTION_NAME = "--Xplugins-auto-enabled";
```

macfarla (approved):
minor non-blocking suggestion on option name

besu/src/main/java/org/hyperledger/besu/cli/DefaultCommandValues.java
jframe, line 132: The naming of this is a bit confusing, maybe --Xplugins-auto-loading-enabled?

besu/src/main/java/org/hyperledger/besu/services/BesuPluginContextImpl.java
jframe: If --plugins are specified and --Xplugin-auto-enabled are specified then I think there should be an error as you have specified plugins but disabled the mechanism to do so.

Gabriel-Trintinalia: Do you prefer the option to disable plugins at all?

jframe (approved):

PR #7473, opened 2024-08-16
Correctly release txpool save and restore lock in case of exceptions

fab-10:
## PR description

The disk access lock used when saving or restoring the txpool from disk, was not released in case of an exception, with the effect of stalling the sync


garyschulte (approved):
LGTM, I will move the changelog entry into release-next

PR #7475, opened 2024-08-16
Bonsai archive feature

matthew1001:
## PR description

Introduces a new (experimental) "Bonsai Archive" DB mode which creates a full archive of the chain it syncs with. This allows JSON/RPC calls to be made with historic blocks as context, for example `eth_getBalance` to get the balance of an account at a historic block, or `eth_call` to simulate a transaction at a given block in history.

The PR is intended to provide part of the function currently offered by the (now deprecated) `FOREST` DB mode. Specifically it allows state to be queried at an arbitrary block in history, but does not currently offer `eth_getProof` for said state. A subsequent PR will implement `eth_getProof` for historic blocks.

### Summary of the overall design & changes

 - More detailed design is covered in https://lf-hyperledger.atlassian.net/wiki/spaces/BESU/pages/22156895/Bonsai+archive+Kikori+DB+feature)
 - Also see discord discussion https://discord.com/channels/905194001349627914/1261348506648707133/1261348712047706243

This PR builds on PR https://github.com/hyperledger/besu/pull/5865 which proved the basic concept of archiving state in the Bonsai flat DB by suffixing entries with the block in which they were changed.

For example the state for account `0x0e79065B5F11b5BD1e62B935A600976ffF3754B9` at block `37834` is stored as 

```
<account-hash><block-num-hex> = 0x9ab656e8fa2a1029964289c9a189083db258ca4b46ebaa374477e069b8f47dec00000000000093ca
```

In order to minimise performance degradation over time, historic state and storage entries in the DB are "archived" by moving them into a separate DB segment.

Where account state is stored in segment `ACCOUNT_INFO_STATE`, state that has been archived is stored in `ACCOUNT_INFO_STATE_ARCHIVE`. Likewise where storage is held in segment `ACCOUNT_STORAGE_STORAGE`, archived storage entries are stored in `ACCOUNT_STORAGE_ARCHIVE`.

An example Rocks DB query to retrieve the state of the example account above would be:

```
ldb --db=. get --column_family=ACCOUNT_INFO_STATE_ARCHIVE --key_hex --value_hex 0x9ab656e8fa2a1029964289c9a189083db258ca4b46ebaa374477e069b8f47dec00000000000093ca
```

### Creating a Bonsai Archive node

The PR introduces an entirely new data storage format (as opposed to making it a configuration option of the existing `BONSAI` storage format.

To create a bonsai archive node simply set `--data-storage-format=x_bonsai_archive` when creating it.

An existing `FOREST` or `BONSAI` node cannot be migrated to `BONSAI_ARCHIVE` mode.

### Storage requirements

An archive node intrinsically requires more storage space than a non-archive node. Every state update is retained in the archive DB segments as outlined above. An archive node for the `holesky` testnet as of the raising of this PR requires approximately `160Gi` of storage.

### Sync time

In order to create an archive of an entire chain, `FULL` sync mode must be used. This PR does not prevent `SNAP` syncing an archive node, but this will result in only a partial archive of the chain.

While the node is performing a `FULL` sync with the chain it is also migrating entries from the regular DB segments to the archive DB segments. Overall this increases the time to create the archive node. For a public chain this might require 1 week or more to complete syncing and archiving.


garyschulte:
Reviewing now.  What networks have you synced with this method?  Do you have block execution performance metrics?

I will kick off a full sync of mainnet alongside a regular bonsai full sync so we can get a signal about performance comparison. 👍

matthew1001:
So far I've `full` synced `holesky` to completion and I'm currently `full` syncing `mainnet`. Mainnet is taking a long time, but I don't know if it's longer than it should be or if my Besu/Teku setup is sub-optimal.

Here's a snapshot of the current sync state with `mainnet`: 
image

Yellow line at the top is number of blocks synced. Blue line is number of blocks archived.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiArchiver.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class manages the archiving of historic state that is still needed to satisfy queries but
+ * doesn't need to be in the main DB segment for. Doing so would degrade block-import performance
```
macfarla, line 42:
```suggestion
 * doesn't need to be in the main DB segment. Doing so would degrade block-import performance
```

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/BaseVersionedStorageFormat.java
```diff
@@ -44,7 +44,17 @@ public enum BaseVersionedStorageFormat implements VersionedStorageFormat {
    * Current Bonsai version, with receipts using compaction, in order to make Receipts use less disk
    * space
    */
-  BONSAI_WITH_RECEIPT_COMPACTION(DataStorageFormat.BONSAI, 3);
+  BONSAI_WITH_RECEIPT_COMPACTION(DataStorageFormat.BONSAI, 3),
+  /**
+   * Current Bonsai version, with blockchain variables in a dedicated column family, in order to
```
macfarla, line 49:
```suggestion
   * Current Bonsai archive version, with blockchain variables in a dedicated column family, in order to
```

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/configuration/BaseVersionedStorageFormat.java
```diff
@@ -44,7 +44,17 @@ public enum BaseVersionedStorageFormat implements VersionedStorageFormat {
...
+   * make BlobDB more effective
+   */
+  BONSAI_ARCHIVE_WITH_VARIABLES(DataStorageFormat.X_BONSAI_ARCHIVE, 1),
+  /**
+   * Current Bonsai version, with receipts using compaction, in order to make Receipts use less disk
```
macfarla, line 54:
```suggestion
   * Current Bonsai archive version, with receipts using compaction, in order to make Receipts use less disk
```

trie.txt
macfarla, line 1: is this file used in tests?

garyschulte:
Not stale, mainnet archive  regression testing just takes a long time

garyschulte:
Regression testing on mainnet got far enough (well past cancun, ~19mil) to assert there is no regression in the base bonsai case w/o archive enabled.  Functionally, this PR is good-to-merge IMO

garyschulte (dismissed):
I am good with this PR, but it would be wise to get a second approval from @matkt

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -72,12 +76,51 @@ public BonsaiWorldStateProvider(
...
 
+  @Override
+  public Optional<MutableWorldState> getMutable(
+      final BlockHeader blockHeader, final boolean shouldPersistState) {
+    if (shouldPersistState) {
+      return getMutable(blockHeader.getStateRoot(), blockHeader.getHash());
+    } else {
+      if (this.worldStateKeyValueStorage.getFlatDbStrategy() instanceof BonsaiArchiveFlatDbStrategy
+          && trieLogManager.getTrieLogLayer(blockHeader.getBlockHash()).isPresent()) {
+        var contextSafeCopy =
```
matkt, line 94: I don't quite understand this modification. If we don't want to persist the state, the logic is to use `cachedWorldStorageManager`, which will handle returning the safe version of the `keyValueStorage` (wrapped or not) to ensure the database's safety. Using `worldStateKeyValueStorage` from `WorldStateProvider` without wrapping it when we don't want to persist seems dangerous and could create inconsistencies, in my opinion. I would suggest keeping it to ensure we have a safe copy of the key-value storage.

```java
cachedWorldStorageManager
    .getWorldState(blockHeader.getHash())
    .or(() -> cachedWorldStorageManager.getHeadWorldState(blockchain::getBlockHeader))
```

And then call this on the copy:

```java
var contextSafeCopy = ((BonsaiWorldStateKeyValueStorage) worldStateKeyValueStorage).getContextSafeCopy();
contextSafeCopy.getFlatDbStrategy().updateBlockContext(blockHeader);
```

Because this method is not that safe:

```java
public BonsaiWorldStateKeyValueStorage getContextSafeCopy() {
    return new BonsaiWorldStateKeyValueStorage(
        flatDbStrategyProvider.contextSafeClone(), composedWorldStateStorage, trieLogStorage);
}
```

It just closes the db strategy but not the columns. With the code below, it will be really safe:

Otherwise, we can also ensure that we always have a copy of the strategy when wrapping a key-value storage (see the constructor of `BonsaiSnapshotWorldStateKeyValueStorage`), and thus we won't need:

```java
var contextSafeCopy = ((BonsaiWorldStateKeyValueStorage) worldStateKeyValueStorage).getContextSafeCopy();
```

And we will just have:

```java
Optional<DiffBasedWorldState> worldState = cachedWorldStorageManager
    .getWorldState(blockHeader.getHash())
    .or(() -> cachedWorldStorageManager.getHeadWorldState(blockchain::getBlockHeader))

worldState.ifPresent(ws -> {
    ws.getFlatDbStrategy().updateBlockContext(blockHeader);
});
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,406 @@
...
+    return Arrays.concatenate(
+        naturalKey,
+        context
+            .getBlockHeader()
+            .map(BlockHeader::getNumber)
+            .map(Bytes::ofUnsignedLong)
+            .map(Bytes::toArrayUnsafe)
+            .orElseGet(
+                () -> {
+                  // TODO: remove or rate limit these warnings
```
matkt, line 391: is it something to manage in another PR ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,406 @@
...
+                  LOG.atWarn().setMessage("Block context not present, using default suffix").log();
+                  return orElseSuffix;
+                }));
+  }
+
+  @Override
+  public void updateBlockContext(final BlockHeader blockHeader) {
+    context.setBlockHeader(blockHeader);
```
matkt, line 399: the logic with Bonsai is that the `WorldState` does not have any notion of block context, etc., but that the information comes from its `KeyValueStorage`, which provides the head and the data.

I wonder why the `blockHeader` would not be given to the `KeyValueStorage` instead of the strategy, which has nothing to do with it.

What I mean is:

I want to trace block 7 and the head is 9.

I do:

```java
cachedWorldStorageManager
    .getWorldState(blockHeader.getHash())
    .or(() -> cachedWorldStorageManager.getHeadWorldState(blockchain::getBlockHeader))
```

This will take the `KeyValueStorage` of head 9 and wrap it with a `LayeredKeyValueStorage`. We then give the `blockHeader`, which is 7, to this layered storage returned by this call. Thus, we have a safe copy that knows its context, and if we have other calls at the same time to trace another block, we will do the same. Each call will have its own copy and will not impact the keyvalue storage or other calls. This also remains in the spirit of the initial code, which is to say that it is the `KeyValueStorage` class that will know all the context.

idk what do you think @garyschulte

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -337,28 +341,34 @@ private void clearStorage(
...
-                    .computeIfAbsent(
-                        storageSlotKey, key -> new DiffBasedValue<>(slotValue, null, true))
-                    .setPrior(slotValue);
-              });
-          entriesToDelete.keySet().forEach(storageTrie::remove);
-          if (entriesToDelete.size() == 256) {
-            entriesToDelete = storageTrie.entriesFrom(Bytes32.ZERO, 256);
-          } else {
-            break;
+        if (storageToDelete != null) {
```
matkt, line 344: did you tried the proposition I shared on discord https://discord.com/channels/905194001349627914/1261348506648707133/1296138719115345980

regarding my comment here https://discord.com/channels/905194001349627914/1261348506648707133/1294238512157032491

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/StorageRangeDataRequest.java
```diff
@@ -108,15 +108,24 @@ protected int doPersist(
...
-        });
+    if (!worldStateStorageCoordinator.isMatchingFlatMode(FlatDbMode.PARTIAL)) {
+      // we have a flat DB only with Bonsai
+      flatDatabaseUpdater.set(
+          (key, value) ->
+              ((BonsaiWorldStateKeyValueStorage.Updater) updater)
+                  .putStorageValueBySlotHash(
+                      accountHash, Hash.wrap(key), Bytes32.leftPad(RLP.decodeValue(value))));
+    } else {
+      worldStateStorageCoordinator.applyOnMatchingFlatMode(
```
matkt, line 119: I think that will throw an exception with FOREST and SNAPSYNC ?

matkt (changes requested):
I have mainly ensured that we do not introduce any regression in Bonsai and that we also respect the logic of Bonsai.

trie.txt
matthew1001: No I was using that for debugging, not sure how I ended up committing it. I've removed it in the latest commit.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
matthew1001: I found on first startup of a new archive node I saw a flurry of these warnings, and then they didn't happen again. I think for now I'd be happier leaving in the warnings without any rate limiting, and we can see if we get reports of them being too noisy. If it's OK I'd like to leave the TODO comment in place for now.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiWorldStateProvider.java
garyschulte: What you are suggesting is actually dangerous for the archive node because it would allow mutating the 'global' block context. When we query the flat db, that would change what the 'nearest' blocknumber suffix could be for any given block, and the worldstate would give incorrect results and corrupt the flatdb archive worldstate.

The problem is that blockContext is a bit of state that is specific to head in archive mode. We need to make sure it does not get mutated by copies of the worldstate that are mutating non-head worldstate. They need their own copy of blockcontext so they can get correct/consistent results and don't end up values from the future. In the initial design the flat db strategy had a bonsai context with the block context. Since all of the archive functionality is localized to the flat db, using a block number suffix, it just made sense to have it there.

It isn't ideal to have to copy the flat db strategy when returning a non-persisting cached worldstate, but I don't think it is unsafe. It is essentially the same getMutable(shouldPersist=false) behavior we have had with non-archive bonsai, it just does it with a copy of the flatdbstrategy so we can give it a block context that is both correct, and isolated.

A better solution would likely be to have each bonsai archive worldstate take have block context supplier/mutator that the ArchiveFlatDbStrategy can defer to. When we hand out cached/frozen worldstates, we give it an explicit supplier/mutator that is at the correct block height and isolated from any other worldstates (including head)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
garyschulte: I think this is similar to what I was alluding to in the above comment.  The challenge is how can we make bonsai archive not leak its implementation details into the non-archive worldstate interfaces.  Bonsai, Verkle, Forest, etc - none of them care what the associated block number is, but for archive we have to get it there somehow, and update it when we get a new block.

matthew1001: I'm doing some refactoring after a quick chat with @matkt on Friday. Hoping to have new commits ready today. I think/hope they address this topic in general

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiWorldStateProvider.java
matthew1001: Yeah I had a chat with @matkt on Friday where we went over the code in this area. I've got a new commit almost ready which refactors the approach to how bonsai context and archive world states are managed. In general I'm much happier with where the code is in my local commit, so hopefully it addresses this comment.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
matthew1001: Yes I did try it, and I think I then got side tracked by hitting some unrelated issues. So I've reinstated your proposed implementation of `clearStorage()` in the latest commits.

matthew1001:
@garyschulte @matkt I've added some commits that refactor the way archive world state and bonsai context work together. I think it's generally much cleaner now.

There are some small things I think that could still be improved. I've updated `BonsaiContext` to allow either block number or block hash to be provided - perhaps that should just be one or the other but I'm inclined to leave it as a pragmatic solution for now. But I'm generally pretty happy with the code structure now.

garyschulte:
> @garyschulte @matkt I've added some commits that refactor the way archive world state and bonsai context work together. I think it's generally much cleaner now.
> 
> There are some small things I think that could still be improved. I've updated `BonsaiContext` to allow either block number or block hash to be provided - perhaps that should just be one or the other but I'm inclined to leave it as a pragmatic solution for now. But I'm generally pretty happy with the code structure now.

...

Is there any time pressure for this feature?  With the latest refactor, I'd like to at minimum archive and non-archive sync a network that predates cancun (where we nerfed SELFDESTRUCT).  

Ideally we would have time to full sync mainnet without archive to ensure no regression, but that is a months long process on commodity hardware.  

Either way - I will keep :eyes: on this PR and kick off regression tests when you signal that it is ready again with your latest commits.

matthew1001:
> > @garyschulte @matkt I've added some commits that refactor the way archive world state and bonsai context work together. I think it's generally much cleaner now.
> > There are some small things I think that could still be improved. I've updated `BonsaiContext` to allow either block number or block hash to be provided - perhaps that should just be one or the other but I'm inclined to leave it as a pragmatic solution for now. But I'm generally pretty happy with the code structure now.
> 
> ...
> 
> Is there any time pressure for this feature? With the latest refactor, I'd like to at minimum archive and non-archive sync a network that predates cancun (where we nerfed SELFDESTRUCT).
> 
> Ideally we would have time to full sync mainnet without archive to ensure no regression, but that is a months long process on commodity hardware.
> 
> Either way - I will keep 👀 on this PR and kick off regression tests when you signal that it is ready again with your latest commits.

I think almost all of the latest refactoring is isolated to archive-specific logic or classes. The `clearStorage()` implementation may be the one exception to that, which I'd be happy to revert if we want to merge the logic that we've tested with the full `main` merge. My concern with holding off too long is that the code-base is evolving and requires a fair amount of effort to keep rebasing/merging the PR with main. Unless we have big concerns that Bonsai (non-archive) has regressed I'd be keen to get this one merged sooner rather than later.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
```diff
@@ -165,8 +166,23 @@ public void persist(final BlockHeader blockHeader) {
 
     boolean success = false;
 
-    final DiffBasedWorldStateKeyValueStorage.Updater stateUpdater =
-        worldStateKeyValueStorage.updater();
+    // Bonsai archive uses context set in the world state storage for the specified block.
+    // Update the block context before putting entries to storage via calculateRootHash()
+    // TODO - rename calculateRootHash() to be clearer that it updates state, it doesn't just
+    // calculate a hash
+    if (worldStateKeyValueStorage.getFlatDbStrategy() instanceof BonsaiArchiveFlatDbStrategy
```
matkt, line 173: I prefer to put it all the time, it doesn't really have a huge impact and avoids an additional condition

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -78,12 +81,36 @@ public BonsaiWorldStateProvider(
...
     this.worldStateHealerSupplier = worldStateHealerSupplier;
+    this.evmConfiguration = evmConfiguration;
     provideCachedWorldStorageManager(bonsaiCachedWorldStorageManager);
     loadPersistedState(
         new BonsaiWorldState(
             this, worldStateKeyValueStorage, evmConfiguration, defaultWorldStateConfig));
   }
 
+  @Override
+  public Optional<MutableWorldState> getMutable(
```
matkt, line 92: it seems to be useless ? it's already in DiffBasesWorldStateProvider class

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,404 @@
...
+                                    .value()
+                                    .get())
+                            .trimLeadingZeros())));
+  }
+
+  /*
+   * Puts the account data for the given account hash and block context.
+   */
+  @Override
+  public void putFlatAccount(
```
matkt, line 231: we cannot add direlctly the blocknumber in the method fields instead of passing a context ? like that the implementation decide ot use it or not ? just a proposition

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/DiffBasedWorldStateProvider.java
```diff
@@ -153,7 +155,7 @@ public synchronized Optional<MutableWorldState> getMutable(
     return rollMutableStateToBlockHash(persistedState, blockHash);
   }
 
-  Optional<MutableWorldState> rollMutableStateToBlockHash(
+  protected Optional<MutableWorldState> rollMutableStateToBlockHash(
```
matkt, line 158: I think we can go back to private

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiWorldStateProvider.java
matthew1001: Agreed - I suspect I needed it before the refactoring I did. I've removed it in the latest commit.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
matthew1001: The latest refactoring is cleaner in this regard I believe.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -1942,9 +1942,14 @@ private PrivacyParameters privacyParameters() {
         throw new ParameterException(
             commandLine, String.format("%s %s", "Checkpoint sync", errorSuffix));
       }
-      if (getDataStorageConfiguration().getDataStorageFormat().equals(DataStorageFormat.BONSAI)) {
+      if (getDataStorageConfiguration().getDataStorageFormat() == DataStorageFormat.BONSAI) {
```
matkt, line 1945: is it needed ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/common/GenesisWorldStateProvider.java
```diff
@@ -57,7 +61,7 @@ public static MutableWorldState createGenesisWorldState(
    *
    * @return a mutable world state for the Genesis block
    */
-  private static MutableWorldState createGenesisBonsaiWorldState() {
+  private static MutableWorldState createGenesisBonsaiWorldState(final boolean archiveMode) {
```
matkt, line 64: nit : why not passing directly DataStorageConfiguration instead of the boolean ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,434 @@
...
+    Optional<BonsaiContext> context = getStateArchiveContextForRead(storage);
+    return Optional.of(
+        context.orElseThrow(
+            () -> new IllegalStateException("World state missing archive context")));
+  }
+
+  private Optional<BonsaiContext> getStateArchiveContextForRead(
+      final SegmentedKeyValueStorage storage) {
+    // For Bonsai archive get the flat DB context to use for reading archive entries
+    Optional<byte[]> archiveContext = storage.get(TRIE_BRANCH_STORAGE, WORLD_BLOCK_NUMBER_KEY);
```
matkt, line 90: so we cannot find a way to cache it and not read it everytime ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/DiffBasedWorldStateProvider.java
```diff
@@ -141,7 +143,7 @@ public Optional<MutableWorldState> getMutable(
       return cachedWorldStorageManager
           .getWorldState(blockHeader.getHash())
           .or(() -> cachedWorldStorageManager.getNearestWorldState(blockHeader))
-          .or(() -> cachedWorldStorageManager.getHeadWorldState(blockchain::getBlockHeader))
+          .or(() -> cachedWorldStorageManager.getWorldState(chainHeadBlockHeader.getHash()))
```
matkt, line 146: why removing 'head' ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
```diff
@@ -157,8 +165,21 @@ public void persist(final BlockHeader blockHeader) {
...
+    // calculate a hash
+    DiffBasedWorldStateKeyValueStorage.Updater stateUpdater = worldStateKeyValueStorage.updater();
+    stateUpdater
+        .getWorldStateTransaction()
+        .put(
+            TRIE_BRANCH_STORAGE,
+            WORLD_BLOCK_NUMBER_KEY,
+            Long.toHexString(blockHeader == null ? 0L : blockHeader.getNumber())
+                .getBytes(StandardCharsets.UTF_8));
+    stateUpdater.commit();
```
matkt, line 180: I will prefer to remove this one and use the commit ihere https://github.com/hyperledger/besu/pull/7475/files#diff-a914e260cf83b36ae4032f8bb562fd1a794ef1432cd8642e5483cbb7270121c4R235
like that we are sure to have only one commit for everything and we avoid having mix if the next step is failing

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/ArchiveFlatDbReaderStrategyTest.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.bonsai.storage.flat;
+
+public class ArchiveFlatDbReaderStrategyTest {}
```
matkt, line 17: I think we can remove , it's empty

matkt (changes requested):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/DiffBasedWorldStateProvider.java
matthew1001: I found a race condition where `getHeadWorldState` can return `empty` because `blockchain.getBlockHeader()` returns empty for the world state block hash in `rootWorldStateStorage`. For archive it's better that we get the latest block that `blockchain` can provide, even if that's not the absolute latest world state. If you think it's better for `DiffBasedWorldStateProvider` to return `Optional.empty` rather than that, I'm happy to revert the change in `DiffBasedWorldStateProvider`.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
matthew1001: I'm not sure why that got changed. I'll revert it to the previous code.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
matthew1001: I don't believe we can use a single DB commit because the `PUT` of the `WORLD_BLOCK_NUMBER_KEY` needs committing before subsequent state access (in `calculatedRootHash`) can do a `GET `of that key.

The latest commit just updates it to create an updater, commit it, then recreate it.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/bonsai/storage/flat/ArchiveFlatDbReaderStrategyTest.java
matthew1001: Yeah I can't remember if there was originally going to be an archive flat DB reader but yeah, I'll remove it.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
matthew1001: We possibly could. Maybe something to improve upon in the next PR(s) in this area?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
matkt: The calculateRootHash should read the head from the database, so it should be the last block number persisted and not the new one we are trying to persist. The old block number should already be saved in the storage if we have committed the block previously. Therefore, I don't think we need to do it again before calculating the root hash.

matthew1001: > The calculateRootHash should read the head from the database

Agreed and it still does. It doesn't use `WORLD_BLOCK_NUMBER_KEY` to determine which block/state to read. Possibly we should rename `WORLD_BLOCK_NUMBER_KEY` -> `WORLD_ARCHIVE_BLOCK_KEY` and reserve it for archive-related use? In the future we could add `WORLD_BLOCK_NUMBER_KEY` if we thought it was useful and it wouldn't be updated at the same time as `WORLD_ARCHIVE_BLOCK_KEY`.

> The old block number should already be saved in the storage if we have committed the block previously.

That's true, the old block number should already be persisted. But for the new block & flat-db archive keys we take `WORLD_BLOCK_NUMBER_KEY` and use that as the suffix - so it can't be the previous block number when we call `calculateRootHash`. For persisting a new block the number in storage needs to be the new block number, not the previous one. So I see this code change as "this is the block number we're about to persist, use it for archive flat-db suffixes". Again, perhaps `WORLD_ARCHIVE_BLOCK_KEY` would help ensure it isn't misinterpreted elsewhere in the codebase?

matkt: so it's for persisting the new keys in the database only ?

matkt: if it's the case why not passing this block number to the updater ?

matthew1001: Yeah I understand that concern. I think the current design is OK - I don't think we should pass it into the updater - but perhaps I need to change how we use the value of that key:

 - When using `WORLD_BLOCK_NUMBER_KEY` for archive **read** activity (i.e. lookup this account in the archive for this state) just use the value of that key from the DB/state
 - When using `WORLD_BLOCK_NUMBER_KEY` for archive **write** activity (i.e. put this new account) we use `WORLD_BLOCK_NUMBER_KEY+1`
   - E.g. we have currently the state for block 10, we're creating block 11, read WORLD_BLOCK_NUMBER_KEY (which is 10) and put archive entries as `WORLD_BLOCK_NUMBER_KEY+1` (11)

I've already introduced `getStateArchiveContextForRead()` and `getStateArchiveContextForWrite()` in `BonsaiArchiveFlatDbStrategy` because there are reasons for treating the cases differently (particularly error handling). So I think the above changes would meet your requirement for `WORLD_BLOCK_NUMBER_KEY` matching what the world state is.

matkt: WORLD_BLOCK_NUMBER_KEY+1 but in case of reorg for example we need to go back from the head 11 to 10 ? I think it will be a problem maybe ? because will write to 12 instead of 10 .

I am not 100% confident about this modification, but I will not block the PR it if you think it is good, knowing that it will only impact the Bonsai archive. I would like to have @garyschulte 's opinion on this.

I would prefer not to commit before and only do it in case of success, but I will go with the majority's opinion if Gary thinks this approach is ok.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/DiffBasedWorldStateProvider.java
matkt: I understand the need. The problem is that with this modification, if your head is not in the cache, it won't work. So I think it's a bit risky. When exactly do you have the race condition? Who is calling the getMutable at that moment? Added that the head is there as a fallback normally if nothing is in the cache. We take the persisted state.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
garyschulte: If we want to set the block context optimistically and avoid a `commit()`, we could just remove the commit and  read-through -the-transaction whenever we are looking for the archive block number.  

i.e. add a `.get(..)` method to SegmentedKeyValueStorageTransaction.  RocksDBSnapshotTransaction already implements get(), it just isn't exposed in the interface.  

I think it would be trivial to add it to SegmentedInMemoryTransaction, where we read from the updated value map and if not present, defer to the parent SegmentedInMemoryKeyValueStorage.get().  There are a few other variants of this, but these are the two main flavors - rocksdb transactions and in-memory maps.

We'd just need to ensure that when we are accessing the "context" block number, we always read through the transaction.

That being said, in this solution we are using the storage layer to communicate what I think are ephemeral query-time parameters.  I don't really like persisting a 'latest query parameter' block number to storage.  It is brittle and a potential source of contradiction that we have to handle in all of the corner cases like reorgs, block production, and now querying.  IMO we have the block number when we need to produce a block, and we have the block number when we need to query a block.  Persisting this somewhere and handing its update seems like an unnecessary complication to an already complex storage mechanism.

garyschulte (commented):
I didn't get through the whole review due to time, but wanted to provide feedback on what I did get to.

garyschulte:
> My concern with holding off too long is that the code-base is evolving and requires a fair amount of effort to keep rebasing/merging the PR with main. Unless we have big concerns that Bonsai (non-archive) has regressed I'd be keen to get this one merged sooner rather than later.

Agreed.  This is a big PR and not only is it a nightmare to keep up to date with an evolving storage subsystem, all those merges are potential sources of error.  I am keen to get this tested and merged also

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
garyschulte: reading-through-the-transaction should be all in memory and relatively cheap.  Pretty much cached by default.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServer.java
```diff
@@ -145,8 +144,9 @@ public synchronized SnapServer start() {
     if (!isStarted.get() && snapServerEnabled) {
       // if we are bonsai and full flat, we can provide a worldstate storage:
       var worldStateKeyValueStorage = worldStateStorageCoordinator.worldStateKeyValueStorage();
-      if (worldStateKeyValueStorage.getDataStorageFormat().equals(DataStorageFormat.BONSAI)
-          && worldStateStorageCoordinator.isMatchingFlatMode(FlatDbMode.FULL)) {
+      if (worldStateKeyValueStorage.getDataStorageFormat().isBonsaiFormat()
+          && (worldStateStorageCoordinator.isMatchingFlatMode(FlatDbMode.FULL)
+              || worldStateStorageCoordinator.isMatchingFlatMode(FlatDbMode.ARCHIVE))) {
```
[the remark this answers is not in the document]
matthew1001: No I've only tried it with the test cases I've written in the PR and local QBFT nodes. I could switch my own archive sync to have snap-server enabled in case that shows anything up, might not be a quick thing to hit though

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/trie/diffbased/common/trielog/ArchiverTests.java
```diff
@@ -0,0 +1,1132 @@
...
+  private BonsaiWorldStateKeyValueStorage worldStateStorage;
+  private Blockchain blockchain;
+  private TrieLogManager trieLogManager;
+  private final Consumer<Runnable> executeAsync = Runnable::run;
+  private BonsaiWorldState bonsaiWorldState;
+
+  @SuppressWarnings("BannedMethod")
+  @BeforeEach
+  public void setup() {
+    Configurator.setLevel(LogManager.getLogger(ArchiverTests.class).getName(), Level.TRACE);
```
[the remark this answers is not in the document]
matthew1001: No I think this can go

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/StorageRangeDataRequest.java
```diff
@@ -108,15 +108,24 @@ protected int doPersist(
...
-    worldStateStorageCoordinator.applyOnMatchingFlatMode(
-        FlatDbMode.FULL,
-        bonsaiWorldStateStorageStrategy -> {
-          flatDatabaseUpdater.set(
-              (key, value) ->
-                  ((BonsaiWorldStateKeyValueStorage.Updater) updater)
-                      .putStorageValueBySlotHash(
-                          accountHash, Hash.wrap(key), Bytes32.leftPad(RLP.decodeValue(value))));
-        });
+    if (!worldStateStorageCoordinator.isMatchingFlatMode(FlatDbMode.PARTIAL)) {
```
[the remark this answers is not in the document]
matthew1001: Yeah let me take another look. I think I made this change fairly early on in the PR, let me see if I can remember the intent of the change.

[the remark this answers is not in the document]
matthew1001: I think this was left over from an early commit on the PR that I meant to come back and remove the old logic. The latest commit(s) remove the if/else and use a new `applyOnMatchingModes()` I've added.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
matthew1001: I've decided to rework the logic so that the archive strategy code takes the world state block number and increases it by 1 when writing new keys, but uses it verbatim when reading. The `WORLD_BLOCK_NUMBER_KEY` is then persisted in the same commit as the calculated block hash once it has been calculated. This should avoid @matkt 's concern about the possibility of world state & the block hash/number ever being incorrect.

I did need to add some specific logic to handle the genesis block case, where there's no 
 initial `WORLD_BLOCK_NUMBER_KEY-1` DB entry for the archive context. I think it's OK to have that one special case, let me know what you think.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,451 @@
...
+    // we're working with the latest world state so putting new flat DB keys requires us to +1 to it
+    Optional<byte[]> archiveContext = storage.get(TRIE_BRANCH_STORAGE, WORLD_BLOCK_NUMBER_KEY);
+    if (archiveContext.isPresent()) {
+      try {
+        return Optional.of(
+            // The context for flat-DB PUTs is the block number recorded in the specified world
+            // state, + 1
+            new BonsaiContext(
+                Long.decode("0x" + (new String(archiveContext.get(), StandardCharsets.UTF_8)))
+                    + 1));
```
matkt, line 90: I think you can use something like that to decode and encode in the database

encode to byte[] before writing into the database
    byte[] bytes = Bytes.ofUnsignedLong(100000).toArrayUnsafe();

decode from the database
Bytes.wrap(bytes.toArrayUnsafe()).toLong();

matkt, line 90: I think you can use something like that to decode and encode in the database

encode to byte[] before writing into the database
    byte[] bytes = Bytes.ofUnsignedLong(100000).toArrayUnsafe();

decode from the database
Bytes.wrap(bytes.toArrayUnsafe()).toLong();

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/DiffBasedWorldStateKeyValueStorage.java
```diff
@@ -122,6 +136,12 @@ public Optional<Hash> getWorldStateBlockHash() {
         .map(Hash::wrap);
   }
 
+  public Optional<Long> getWorldStateBlockNumber() {
+    return composedWorldStateStorage
+        .get(TRIE_BRANCH_STORAGE, WORLD_BLOCK_NUMBER_KEY)
+        .map(bytes -> Long.decode("0x" + new String(bytes, StandardCharsets.UTF_8)));
```
matkt, line 142: I think you can use something like that to decode and encode in the database

encode to byte[] before writing into the database
 >  byte[] bytes = Bytes.ofUnsignedLong(100000).toArrayUnsafe();

decode from the database
> Bytes.wrap(bytes).toLong();

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
```diff
@@ -223,6 +237,14 @@ public void persist(final BlockHeader blockHeader) {
       stateUpdater
           .getWorldStateTransaction()
           .put(TRIE_BRANCH_STORAGE, WORLD_ROOT_HASH_KEY, calculatedRootHash.toArrayUnsafe());
+
+      stateUpdater
+          .getWorldStateTransaction()
+          .put(
+              TRIE_BRANCH_STORAGE,
+              WORLD_BLOCK_NUMBER_KEY,
+              Long.toHexString(blockHeader == null ? 0L : blockHeader.getNumber())
```
matkt, line 246: same as the other one

I think you can use something like that to decode and encode in the database

encode to byte[] before writing into the database
 >  byte[] bytes = Bytes.ofUnsignedLong(100000).toArrayUnsafe();

decode from the database
> Bytes.wrap(bytes.toArrayUnsafe()).toLong();

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
matkt: How we manage a reorg in this case if we add everytime +1 ? maybe I'm missing something

matkt (commented):
I think we're close to merge. I just have one small thing I'd like to fix before merging, which is the encode/decode of the block number. After that, I think we can proceed with the first merge of this PR since I don't see any regression in the non-archive part. I'll let @garyschulte  do a second check before we merge.

I have a question that doesn't necessarily need to be addressed in this PR, but I'd like some feedback on it. It's about the fact that we're always doing `+1`. How do we handle the case where there's a reorg and the node needs to go back? we can chat about that if it's not clear @matthew1001 

Have you had the chance to run a checkpoint sync test on mainnet and let it run? I can do it if needed.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
matthew1001: > I think you can use something like that to decode and encode in the database
> 
> encode to byte[] before writing into the database byte[] bytes = Bytes.ofUnsignedLong(100000).toArrayUnsafe();
> 
> decode from the database Bytes.wrap(bytes.toArrayUnsafe()).toLong();

Yeah I'll look at rewriting that in the couple of places it's used, thanks

matthew1001:
> Have you had the chance to run a checkpoint sync test on mainnet and let it run? I can do it if needed.

No, so far my syncs have been full syncs. Do you mean with archive enabled, or just to check regular bonsai?

matkt:
> > Have you had the chance to run a checkpoint sync test on mainnet and let it run? I can do it if needed.
> 
> No, so far my syncs have been full syncs. Do you mean with archive enabled, or just to check regular bonsai?

No just a regular bonsai node that did a checkpoint sync and importing during the weekend. just to be more confident

matkt:
Did a checkpoint sync on mainnet during the weekend with this branch and it's ok. I think we are good to merge it

is it ok for you @ahamlat @garyschulte ?

ahamlat:
I would like to have a look to the performance, can we hold it until the end of the week ?

matthew1001:
> I would like to have a look to the performance, can we hold it until the end of the week ?

It would be really good to get it merged soon, just to minimise the effort keeping it synced with `main`. But I don't mind holding out a bit longer if you want to check for any regressions introduced by the changes.

ahamlat:
I see pretty slow performances, more than 1 second block time, where block execution even recent block is around ~100/200 ms.
I did profiling and noticed that it is related to seekForPrev method, this call on getAccount and getCode takes most of the time in transaction execution. cc @matkt @garyschulte 

image

matthew1001:
Thanks for the analysis @ahamlat It makes sense that that's where a large chunk of time is spent. We could look at what could be improved there, although it's somewhat fundamental to the current design so would need some thought.

> I think we need to use the non archive BonsaiFlatDbStrategy to update the head and only use BonsaiAchiveFlatDbStrategy for RPC request . like that we will not have performance impact related to seekForPrev in order to sync the head

Are you thinking store the current chain head flat DB entries without archive context, so they can be accessed directly for any chain-head updates (and I guess also things like snap range requests)?

matkt:
> Thanks for the analysis @ahamlat It makes sense that that's where a large chunk of time is spent. We could look at what could be improved there, although it's somewhat fundamental to the current design so would need some thought.
> 
> > I think we need to use the non archive BonsaiFlatDbStrategy to update the head and only use BonsaiAchiveFlatDbStrategy for RPC request . like that we will not have performance impact related to seekForPrev in order to sync the head
> 
> Are you thinking store the current chain head flat DB entries without archive context, so they can be accessed directly for any chain-head updates (and I guess also things like snap range requests)?

Not write, I think for importing a new from the head, we don't need to read using seekForPrev we can do a normal get I think.  maybe not . is the last data (head) are also saved with the blocknumber prefix ?

matthew1001:
> is the last data (head) are also saved with the blocknumber prefix

Currently yes - https://github.com/hyperledger/besu/blob/657999079eb11562e0d022a21d1450a0070976e0/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java#L273

matkt:
> > is the last data (head) are also saved with the blocknumber prefix
> 
> Currently yes -
> 
> https://github.com/hyperledger/besu/blob/657999079eb11562e0d022a21d1450a0070976e0/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java#L273

Is it necessary? As I understand it, the key is moved to the archive column with the prefix when we change the head, but we don’t do that for the head. Maybe I misunderstood.

matthew1001:
> Is it necessary? As I understand it, the key is moved to the archive column with the prefix when we change the head, but we don’t do that for the head. Maybe I misunderstood. if we do like that we can avoid using seekForPrev when we import a new block

Yeah currently it always uses the archive key suffix, and moves keys to the archive DB segment with the same key. That's just where the first implementation ended I think. I agree we could look at also storing the current state without a suffix. Just not sure if we need to do that for the first experimental version.

ahamlat:
> > Is it necessary? As I understand it, the key is moved to the archive column with the prefix when we change the head, but we don’t do that for the head. Maybe I misunderstood. if we do like that we can avoid using seekForPrev when we import a new block
> 
> Yeah currently it always uses the archive key suffix, and moves keys to the archive DB segment with the same key. That's just where the first implementation ended I think. I agree we could look at also storing the current state without a suffix. Just not sure if we need to do that for the first experimental version.

I think it depends on the effort, if it is not too much effort, it make sens to store the current state and the ones we have trie logs in memory (512 by default) without a suffix and add it async when moving the keys.
Because currently it has a big impact on performance, as even SLOAD operation uses seekForPrev. On recent block, this will take a very long time

image

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiArchiver.java
```diff
@@ -0,0 +1,243 @@
...
+    initialize();
+    final Optional<Long> blockNumber =
+        Optional.of(addedBlockContext.getBlock().getHeader().getNumber());
+    blockNumber.ifPresent(
+        blockNum -> {
+          // Since moving blocks can be done in batches we only want
+          // one instance running at a time
+          executeAsync.accept(
+              () -> {
+                if (archiveMutex.tryLock()) {
```
ahamlat, line 233: With `synchronized` in `moveBlockStateToArchive`, do you need to add this lock here. I guess the idea is to be sure that we finished moving the block N state to archive before block N+1, `moveBlockStateToArchive` should handle that case.

matthew1001:
> I think it depends on the effort, if it is not too much effort, it make sens to store the current state and the ones we have trie logs in memory (512 by default) without a suffix and add it async when moving the keys.

Yeah I agree it's probably more efficient to do that. I'll see how much work it is to rework the current behaviour.

matthew1001:
> I will prefer to have only BONSAI with a archive flag, but maybe we can do that in another PR.

Yeah I think there was some discussion about that early on in the archive work and IIRC Gary favoured a specific data storage format rather than a config option on BONSAI.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/GenesisState.java
```diff
@@ -151,6 +155,27 @@ public Block getBlock() {
...
+        // Bonsai archive nodes need the block number for the current state in order to persist flat
+        // DB keys. For every other block this is automatically set to the previous block number +
+        // 1. Since this is not possible for the genesis block we manually set it.
+        SegmentedKeyValueStorageTransaction genesisStateTX =
+            ((PathBasedWorldState) genesisState)
+                .getWorldStateStorage()
+                .getComposedWorldStateStorage()
+                .startTransaction();
+        genesisStateTX.put(
+            TRIE_BRANCH_STORAGE, WORLD_BLOCK_NUMBER_KEY, Bytes.ofUnsignedLong(0).toArrayUnsafe());
```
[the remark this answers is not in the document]
matthew1001: I'm not sure the code currently supports starting an already-synced node with a new data storage format. If the node previously started with `--data-storage-format=BONSAI` then starting it with `X_BONSAI_ARCHIVE` would fail to start. So I believe this is safe for any node that is using the new storage format.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,448 @@
...
+        storage
+            .getNearestBefore(ACCOUNT_INFO_STATE, keyNearest)
+            .filter(found -> accountHash.commonPrefixLength(found.key()) >= accountHash.size());
+
+    // If there isn't a match look in the archive DB segment
+    if (nearestAccount.isEmpty()) {
+      accountFound =
+          storage
+              .getNearestBefore(ACCOUNT_INFO_STATE_ARCHIVE, keyNearest)
+              .filter(found -> accountHash.commonPrefixLength(found.key()) >= accountHash.size())
```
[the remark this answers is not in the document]
matthew1001: You mean generate `accountHash.commonPrefixLength(found.key())` before calling `.getNearestBefore`, so the code becomes something like:

```
int comPrefLength = accountHash.commonPrefixLength(found.key());
...

      .filter(found -> comPrefLength >= accountHash.size())
```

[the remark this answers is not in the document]
matthew1001: Just realised that's not possible since `found` doesn't exist until we're streaming the results. So I'm not quite sure what change you're suggesting?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,448 @@
...
+      if (accountFound.isPresent()) {
+        getAccountFromArchiveCounter.inc();
+      }
+    } else {
+      accountFound =
+          nearestAccount
+              .filter(
+                  found ->
+                      !Arrays.areEqual(
+                          DELETED_ACCOUNT_VALUE, found.value().orElse(DELETED_ACCOUNT_VALUE)))
```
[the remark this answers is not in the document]
matthew1001: > it is important to keep the deleted value if we have ACCOUNT_INFO_STATE_ARCHIVE ?

No, I think that's a good point. I've updated the logic in a new commit so the deleted key filter is applied regardless of which DB table the entry is retrieved from.

matkt (approved):
Just waiting for Gary's approve also before merging

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/GenesisState.java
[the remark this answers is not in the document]
matkt: I'm talking about a node started with bonsai and restarting with Bonsai and this PR. it seems we are calling ensureGenesisArchiveContext also in bonsai mode . no ?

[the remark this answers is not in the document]
matthew1001: I see what you mean. I don't think we'd be calling `writeAccountsTo()` on genesis state for an already-synced Bonsai node would we?

matthew1001:
> Just waiting for Gary's approval also before merging

@garyschulte what do you think?

app/src/main/java/org/hyperledger/besu/cli/subcommands/storage/RevertMetadataSubCommand.java
```diff
@@ -118,6 +118,7 @@ public void run() {
             switch (dataStorageFormat) {
               case FOREST -> 1;
               case BONSAI -> 2;
+              case X_BONSAI_ARCHIVE -> 3;
```
garyschulte: this is probably ok, but we should never have a bonsai archive node with the prior metadata format, since the metadata change predates the bonsai archive feature.  It might make sense to just throw here.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/storage/flat/BonsaiFlatDbStrategyProvider.java
```diff
@@ -56,9 +61,16 @@ protected FlatDbMode alternativeFlatDbModeForExistingDatabase() {
...
-        TRIE_BRANCH_STORAGE, FLAT_DB_MODE, FlatDbMode.FULL.getVersion().toArrayUnsafe());
+    if (dataStorageConfiguration.getDataStorageFormat() == DataStorageFormat.BONSAI) {
+      LOG.info("setting FlatDbStrategy to FULL");
+      transaction.put(
+          TRIE_BRANCH_STORAGE, FLAT_DB_MODE, FlatDbMode.FULL.getVersion().toArrayUnsafe());
+    } else if (dataStorageConfiguration.getDataStorageFormat()
+        == DataStorageFormat.X_BONSAI_ARCHIVE) {
+      LOG.info("setting FlatDbStrategy to ARCHIVE");
+      transaction.put(
+          TRIE_BRANCH_STORAGE, FLAT_DB_MODE, FlatDbMode.ARCHIVE.getVersion().toArrayUnsafe());
```
garyschulte, line 72: Do we actually want to do this?  if we are transitioning from partial to archive, we implicitly do not have a real archive node.  Perhaps there could be some value in running an archive node "since" a block height, but I would imagine an archive sync strategy would be a better fit than a snap sync strategy that transitions to archive when complete 🤔

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,448 @@
...
+            .getNearestBefore(ACCOUNT_INFO_STATE, keyNearest)
+            .filter(found -> accountHash.commonPrefixLength(found.key()) >= accountHash.size());
+
+    // If there isn't a match look in the archive DB segment
+    if (nearestAccount.isEmpty()) {
+      accountFound =
+          storage
+              .getNearestBefore(ACCOUNT_INFO_STATE_ARCHIVE, keyNearest)
+              .filter(found -> accountHash.commonPrefixLength(found.key()) >= accountHash.size());
+
```
garyschulte, line 145: if we are looking in the context of MAX_BLOCK_SUFFIX, we shouldn't have to defer to archive right?  current/unchanged account values should not be moved to the archive unless something has gone wrong with archiving right?  if we can skip this step we should see substantially faster "zero read" cases non-historical queries

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
```diff
@@ -0,0 +1,448 @@
...
+    // Find the nearest storage for this address, slot key hash, and block context
+    Optional<SegmentedKeyValueStorage.NearestKeyValue> nearestStorage =
+        storage
+            .getNearestBefore(ACCOUNT_STORAGE_STORAGE, keyNearest)
+            .filter(
+                found -> Bytes.of(naturalKey).commonPrefixLength(found.key()) >= naturalKey.length);
+
+    // If there isn't a match look in the archive DB segment
+    if (nearestStorage.isEmpty()) {
```
garyschulte, line 334: same thing as above comment regarding zero-read performance.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/common/worldview/PathBasedWorldState.java
```diff
@@ -188,6 +195,8 @@ public void persist(final BlockHeader blockHeader) {
       final Hash calculatedRootHash;
 
       if (blockHeader == null || !worldStateConfig.isTrieDisabled()) {
+        // TODO - rename calculateRootHash() to be clearer that it updates state, it doesn't just
+        // calculate a hash
```
garyschulte, line 199: this is only half true, since an Optional.empty() state updater would mean no update.  but yeah it is kinda side-effect-y

garyschulte (approved):
Looks really great.  I left comments and questions regarding feature/function choices, but all of the changes seem to be compartmentalized to the new experimental archive storage mode, so we can address here or in a future PR IMO.  

For the record, I think we need a separate storage format for archive, **as you have it in this PR**.  The archive segments, metadata, and behavior are significantly different to BONSAI. If we don't differentiate between them as separate formats the potential to corrupt a very-expensive-to-sync archive is just one accidental config change away.  

I have my misgivings about communicating the block context via a worldstate storage key (rather than treating it as an ephemeral parameter), BUT it seems the implications of that behavior are limited only to X_BONSAI_ARCHIVE.  We can discover how this works testing operationally.  

FWIW, I have a couple test nodes using this PR - a full sync bonsai node that is right around 12 mil, and an archive node started at the same time that is around 6 mil.   There are probably opportunities for performance improvement in future PRs.

Thanks for keeping this PR updated and waiting for prague 🙏

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/storage/flat/BonsaiFlatDbStrategyProvider.java
matthew1001: Yeah, the archive + proofs PR https://github.com/hyperledger/besu/pull/8669 does some refactoring in this area because I also wasn't totally happy with the way partial/full/archive were handled here. So for now I think I'm going to elect to leave it unchanged in this PR because I think the follow on PR much improves things.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/storage/flat/BonsaiArchiveFlatDbStrategy.java
matthew1001: Yes agreed. I think `calculateArchiveKeyWithMaxSuffix()` needs renaming as it _may_ use `MAX_BLOCK_SUFFIX`, but only if an archive context block isn't set. So I think some additional logic to check if we ended up defaulting to `MAX_BLOCK_SUFFIX` and then skipping the archive lookup is what's needed (plus the possible function rename). If it's OK I'd like to start getting other people looking at the code, and tracking archive bugs etc. as standalone issues, so for this and the other comment I'm going to raise an issue and not fix in the PR.

garyschulte: Yup, definitely reasonable to defer to a subsequent issue.  I think not only can we skip archiving for MAX_BLOCK_SUFFIX, but also for block heights that are above the high water mark for being archived.  If we expose the current archive block number via bonsai context, we can make that decision very cheaply, and get a significant boost for zero reads 👍

matthew1001: I've raised https://github.com/hyperledger/besu/issues/8850 to track this fix

PR #7481, opened 2024-08-19
5098 branch 24 throw checked exception to remove todos

Matilda-Clerke:
## PR description
Final PR in the epic series of PRs. Throw checked exception at low level, and catch to remove TODOs. Also catch in a few places missed in previous reviews.
## Fixed Issue(s)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetUncleCountByBlockHash.java
```diff
@@ -42,7 +43,7 @@ public JsonRpcResponse response(final JsonRpcRequestContext requestContext) {
     final Hash hash;
     try {
       hash = requestContext.getRequiredParameter(0, Hash.class);
-    } catch (Exception e) { // TODO:replace with JsonRpcParameter.JsonRpcParameterException
+    } catch (JsonRpcParameter.JsonRpcParameterException e) {
```
macfarla, line 46: would it be nicer to modify getRequiredParameter() to throw the ex 

hash = requestContext.getRequiredParameterOrElseThrow(0, Hash.class,  "Invalid block hash parameter (index 0)", RpcErrorType.INVALID_BLOCK_HASH_PARAMS)

but there's a couple of spots where we do something different

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/JsonRpcParameter.java
```diff
@@ -60,11 +57,9 @@ public <T> T required(final Object[] params, final int index, final Class<T> par
    * @param <T> The type of parameter.
    * @return Returns the parameter cast as T if available.
    */
-  // TODO: update to throw JsonRpcParameterException as a checked exception, forcing callers to
-  // handle it to supply appropriate context
   @SuppressWarnings("unchecked")
```
macfarla, line 60: should this suppression be removed now?

consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/jsonrpc/methods/CliqueGetSigners.java
```diff
@@ -71,7 +72,7 @@ private Optional<BlockHeader> determineBlockHeader(final JsonRpcRequestContext r
     final Optional<BlockParameter> blockParameter;
     try {
       blockParameter = request.getOptionalParameter(0, BlockParameter.class);
-    } catch (Exception e) { // TODO:replace with JsonRpcParameter.JsonRpcParameterException
+    } catch (JsonRpcParameter.JsonRpcParameterException e) {
```
jframe, line 75: Be nicer to static import JsonRpcParameter.JsonRpcParameterException so don't have the qualified exception here

Matilda-Clerke: Can do

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetUncleCountByBlockHash.java
Matilda-Clerke: That would be another way to get the details we want into the exception, but in general, I'd prefer to keep low level code as simple as possible, and let higher level code handle the specifics in the appropriate context.

As a middle ground, I could implement a set of methods in JsonRpcRequestContext like 
`public Hash getRequiredHashParameter(int index)` which would clean up the higher level code, and still add detail to exceptions in the appropriate context.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/JsonRpcParameter.java
Matilda-Clerke: This was unrelated to the changes I made. It seems we're retrieving an Object from the request context and performing an unchecked cast to a parameterized type. i.e.
`
final T param;
    final Object rawParam = params[index];
    if (paramClass.isAssignableFrom(rawParam.getClass())) {
      // If we're dealing with a simple type, just cast the value
      param = (T) rawParam;
    } 
`
Looks like even though we're checking the type of T is appropriate for the object, the compiler (or intelliJ) is still complaining about the unchecked cast. We can get around this by using `paramClass.cast(rawParam)`, which should have the same effect.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetUncleCountByBlockHash.java
macfarla: it would simplify slightly - I have no strong opinion on that - will leave up to you

Matilda-Clerke: I think given the length of time it would take, it's probably not worth doing right now

ethereum/stratum/src/main/java/org/hyperledger/besu/ethereum/stratum/Stratum1Protocol.java
```diff
@@ -175,18 +176,24 @@ private void handleMiningSubmit(final JsonRpcRequest message, final Consumer<Str
...
     Hash mixHash = null;
     try {
       mixHash = Hash.fromHexString(message.getRequiredParameter(4, String.class));
-    } catch (Exception e) { // TODO:replace with JsonRpcParameter.JsonRpcParameterException
+    } catch (JsonRpcParameterException e) {
       throw new InvalidJsonRpcParameters(
           "Invalid mix hash parameter (index 4)", RpcErrorType.INVALID_MIX_HASH_PARAMS, e);
     }
-    Bytes powHash = Bytes.fromHexString(message.getRequiredParameter(3, String.class));
+    Bytes powHash = null;
```
jframe, line 190: nit: could removed the redundant initialiser to null here

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetUncleCountByBlockHash.java
jframe: nice suggestion, it would make the code more readable removing a lot of try/catch blocks. but agree that it's probably not worth doing right now

jframe (approved):

PR #7486, opened 2024-08-19
CI Fix: Preprocess release title in gha

garyschulte:
## PR description

pre-process release title in release actions to ensure it conforms to a besu release version and trim any leading or trailing spaces

Tested on forked repo with a variety of whitespace and failure modes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,17 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - name: Pre-process Release Name
+        id: pre_process_release_name
+        run: |
+            RELEASE_NAME="${{ github.event.release.name }}"
+            # strip all whitespace
+            RELEASE_NAME="${RELEASE_NAME//[[:space:]]/}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -19,6 +19,17 @@ jobs:
...
       - name: checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - name: Pre-process Release Name
+        id: pre_process_release_name
+        run: |
+            RELEASE_NAME="${{ github.event.release.name }}"
+            # strip all whitespace
+            RELEASE_NAME="${RELEASE_NAME//[[:space:]]/}"
+            if [[ ! "$RELEASE_NAME" =~ ^[0-9]+\.[0-9]+(\.[0-9]+)?(-.*)?$ ]]; then
+              echo "Release name does not conform to a valid besu release format YY.M.v[-suffix], e.g. 24.8.0-RC1."
```
siladu, line 29: Wonder if we should make this strictly full release since this workflow is triggered by the `released` type which doesn't include pre-releases.

siladu (approved):
Nice solution 👍 

Think regex could be stricter but this is a clear improvement

.github/workflows/release.yml
siladu: Would these be published as full releases rather than a pre-release then?

PR #7496, opened 2024-08-20
BFT soak test - use both db modes

matthew1001:
## PR description
Update the BFT soak acceptance test to use both BONSAI and FOREST DB modes


siladu (approved):
LGTM

Issue #7503, opened 2024-08-21
debug_traceCall fails with "gasPrice is less than the current BaseFee" while `eth_call` works


fab-10:
Reported on Eth R&D.

Example call that fails
```
curl --data '{
      "jsonrpc": "2.0",
      "id": 0,
      "method": "debug_traceCall",
      "params": [ {
            "to": "0x0aae40965e6800cd9b1f4b05ff21581047e3f91e",
            "data": "0x000000000000000000000000000000000000000000000000000000000001A00E"
        }, "latest"]
    }'  -H "Content-Type: application/json"  -X POST

{"jsonrpc":"2.0","id":0,"error":{"code":-32603,"message":"Internal error","data":"gasPrice is less than the current BaseFee"}} 
```

garyschulte:
for reference https://discord.com/channels/595666850260713488/892088344438255616/1278303112951496779

PR #7510, opened 2024-08-23
Fix debug_traceCall to handle underpriced transactions

7suyash7:
Fixes: #7503 where `debug_traceCall` was failing due to gas price checks, while `eth_call` was succeeding.

Output from `debug_traceCall`
```
curl --data '{               
  "jsonrpc": "2.0",
  "id": 0,
  "method": "debug_traceCall",
  "params": [ {
    "to": "0x0aae40965e6800cd9b1f4b05ff21581047e3f91e",
    "data": "0x000000000000000000000000000000000000000000000000000000000001A00E"
  }, "latest"]
}' -H "Content-Type: application/json" -X POST http://127.0.0.1:8545
{"jsonrpc":"2.0","id":0,"result":{"gas":21164,"failed":false,"returnValue":"","structLogs":[{"pc":0,"op":"STOP","gas":7978836,"gasCost":0,"depth":1,"stack":[],"memory":[],"storage":{},"reason":null}]}}
```


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceCall.java
```diff
@@ -69,6 +70,44 @@ protected TraceOptions getTraceOptions(final JsonRpcRequestContext requestContex
     }
   }
 
+  protected Object resultByBlockHeader(
```
jframe, line 73: This method doesn't appear to be used anywhere in DebugTraceCall

7suyash7: Hi, yeah, I just removed it, I also removed the `errorResponse` method. Weirdly, the compiler caught that `errorResponse` was not being used anywhere when I tried compiling and building after removing `resultByBlockHeader` right now but didn't catch `resultByBlockHeader`...

fab-10 (commented):
This fix the issue with `debug_traceCall`, but since it part of a group of RPC methods extending the `AbstractTraceByBlock` class, it is better to review if the issue affects other methods in the group, and compare the behavior with other clients to be sure there are not other differences.

7suyash7:
> This fix the issue with `debug_traceCall`, but since it part of a group of RPC methods extending the `AbstractTraceByBlock` class, it is better to review if the issue affects other methods in the group, and compare the behavior with other clients to be sure there are not other differences.

@fab-10 What do you expect the next steps to be here?

fab-10:
@7suyash7 first I suggest to review if  `trace_call`, `trace_callMany` and `trace_rawTransaction` that extend `AbstractTraceByBlock` like `debug_traceCall` have a similar issue and depending on that uniform the fix across all of them.

7suyash7:
> Thanks @7suyash7 for checking, I think that is enough and the change can be limited to this call, could you just add a test case and a changelog entry to complete the fix?

@fab-10 I just did, can you please check?

fab-10 (approved):

fab-10:
LGTM

Issue #6471, opened 2024-01-26
engine_getClientVersionV1 - expose web3_clientVersion via EngineAPI


jflo:
### Description
As an Consensus Client, I want access to my execution clients name and version over the Engine API so that I can include it in my proposed block grafitti.

https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md

### Acceptance Criteria
* Proposed blocks from any CL include the besu version by default.

PR #7512, opened 2024-08-23
add engine_getClientVersionV1

Matilda-Clerke:
## PR description
Add the engine_getClientVersionV1 to the engine RPC API
## Fixed Issue(s)

https://github.com/hyperledger/besu/issues/6471


ethereum/api/src/integration-test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcTestMethodsFactory.java
```diff
@@ -64,7 +64,9 @@
 /** Provides a facade to construct the JSON-RPC component. */
 public class JsonRpcTestMethodsFactory {
 
-  private static final String CLIENT_VERSION = "TestClientVersion/0.1.0";
+  private static final String CLIENT_NODE_NAME = "TestClientVersion/0.1.0";
```
Matilda-Clerke, line 67: Renaming this to better match its source in BesuInfo

besu/src/main/java/org/hyperledger/besu/BesuInfo.java
```diff
@@ -27,6 +29,21 @@ public final class BesuInfo {
   private static final String VERSION = BesuInfo.class.getPackage().getImplementationVersion();
   private static final String OS = PlatformDetector.getOS();
   private static final String VM = PlatformDetector.getVM();
+  private static final String COMMIT;
+
+  static {
+    if (VERSION == null) {
+      COMMIT = null;
```
usmansaleem, line 36: instead of NULL, would it be better to return empty String `""`?

usmansaleem: Also, is there a test that covers NULL commit? How would it be reported?

Matilda-Clerke: Unfortunately, it's difficult to test the value of COMMIT because BesuInfo pulls data from the class' package which isn't typically set outside of a proper build.

As for null vs empty String, philosophically I'd rather it be null. The api doesn't really specify what should be supplied if a value isn't available.

besu/src/main/java/org/hyperledger/besu/BesuInfo.java
```diff
@@ -27,6 +29,21 @@ public final class BesuInfo {
...
+    } else {
+      Pattern pattern = Pattern.compile("v?\\d*\\.\\d*\\.\\d*-\\w+-(?<commit>[0-9a-fA-F]{8})");
+      Matcher matcher = pattern.matcher(VERSION);
```
jframe, line 39: Only the develop builds include the commit hash in the version at the moment, so this won't work for github release versions.

I tested locally using `./gradlew -Prelease.releaseVersion=99.9.0 -Pversion=99.9.0 assemble installDist` and the commit field in the response was null.

Matilda-Clerke: Yeah, I'm not sure how else to get the commit hash. The purpose of the field is to give a specific commit at which the application was run, and for github releases there seems to be an accompanying tag, so the version itself can be used to find the exact commit if desired.

Do you think it would be worthwhile finding some other way to get the commit field populated?

pinges (approved):
LGTM

PR #7532, opened 2024-08-28
refactor SimpleWorld to use Optional for storing accounts and extra null checks

### LINKED ISSUE #7524: Bug in `SimpleWorld.commit()` ?
### Description
Currently, `SimpleWorld.commit()` is defined as follows:

```java
@Override
  public void commit() {
    accounts.forEach(
        (address, account) -> {
          if (!account.updateParent()) {
            parent.accounts.put(address, account);
          }
        });
  }
```

However, this seems to be in conflict with this:

```java
  @Override
  public Collection<Address> getDeletedAccountAddresses() {
    return accounts.entrySet().stream()
        .filter(e -> e.getValue() == null)
        .map(Map.Entry::getKey)
        .toList();
  }
```

It seems to me that we want `account == null || !account.updateParent()` as the criteria in `commit()` above.

Thoughts?

lu-pinto:
## PR description
Fix several NPEs in `SimpleWorld` and I also decided to use `Optional` for accounts since the return from `get`/`getAccount` was ambiguous between an account that has been deleted or has never existed. Now there's a clear distinction between deleted (`Optional.empty()`) or has never existed (`null`). Also `getTouchedAccounts` returned null values when there were deleted accounts within which can be dangerous to the caller.

Note: `SimpleWorld` seems to only be used by the test framework, so this is not critical.

## Fixed Issue(s)

fixes #7524

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleAccount.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleAccount.java
@@ -214,21 +214,4 @@ public class SimpleAccount implements MutableAccount {
       return false;
     }
   }
-
-  /**
-   * Push changes into the parent account, if one exists
-   *
-   * @return true if a parent account was updated, false if not (this indicates the account should
-   *     be inserted into the parent contact).
-   */
-  public boolean updateParent() {
-    if (parent instanceof SimpleAccount simpleAccount) {
-      simpleAccount.balance = balance;
-      simpleAccount.nonce = nonce;
-      simpleAccount.storage.putAll(storage);
-      return true;
-    } else {
-      return false;
-    }
-  }
 }
--- a/evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
@@ -29,10 +29,10 @@ import java.util.Optional;
 public class SimpleWorld implements WorldUpdater {
 
   /** The Parent. */
-  SimpleWorld parent;
+  private final SimpleWorld parent;
 
   /** The Accounts. */
-  Map<Address, SimpleAccount> accounts = new HashMap<>();
+  private Map<Address, Optional<SimpleAccount>> accounts = new HashMap<>();
 
   /** Instantiates a new Simple world. */
   public SimpleWorld() {
@@ -55,13 +55,15 @@ public class SimpleWorld implements WorldUpdater {
 
   @Override
   public Account get(final Address address) {
+    Optional<SimpleAccount> account = Optional.empty();
     if (accounts.containsKey(address)) {
-      return accounts.get(address);
+      account = accounts.get(address);
     } else if (parent != null) {
-      return parent.get(address);
-    } else {
-      return null;
+      if (parent.get(address) instanceof SimpleAccount accountFromParent) {
+        account = Optional.of(accountFromParent);
+      }
     }
+    return account.orElse(null);
   }
 
   @Override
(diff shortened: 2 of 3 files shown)


evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
```diff
@@ -120,12 +123,16 @@ public void revert() {
...
   @Override
   public void commit() {
-    accounts.forEach(
-        (address, account) -> {
-          if (!account.updateParent()) {
-            parent.accounts.put(address, account);
-          }
-        });
+    accounts.entrySet().stream()
+        .filter(e -> e.getValue().isPresent())
```
shemnon, line 127: Shouldn't a commit to the parent also commit all deletes?  Doesn't this filter make all deletes transient?  EIP-6780 compatible, but that should be advertised in the docs that SimpleWorld is not suitable for pre-shanghai EVMs.

lu-pinto: Well spotted @shemnon I would argue that it wasn't committing deletes before as it would throw an NPE, so I was unsure of the right behavior. But now that you mention it, not committing doesn't make sense at all - I'll change.

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
```diff
@@ -55,13 +55,15 @@ public WorldUpdater updater() {
...
     if (accounts.containsKey(address)) {
-      return accounts.get(address);
+      account = accounts.get(address);
     } else if (parent != null) {
-      return parent.get(address);
-    } else {
-      return null;
+      if (parent.get(address) instanceof SimpleAccount accountFromParent) {
+        account = Optional.of(accountFromParent);
+      }
```
garyschulte, line 64: Unless I am missing something, since parent is SimpleWorld, it can only return a SimpleAccount. is this a different flavor of not null check?

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
```diff
@@ -29,10 +29,10 @@
 public class SimpleWorld implements WorldUpdater {
 
   /** The Parent. */
-  SimpleWorld parent;
+  private final SimpleWorld parent;
 
   /** The Accounts. */
-  Map<Address, SimpleAccount> accounts = new HashMap<>();
+  private Map<Address, Optional<SimpleAccount>> accounts = new HashMap<>();
```
garyschulte, line 35: 👍

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
```diff
@@ -70,45 +72,46 @@ public MutableAccount createAccount(final Address address, final long nonce, fin
...
   @Override
   public void deleteAccount(final Address address) {
-    accounts.put(address, null);
+    accounts.put(address, Optional.empty());
   }
 
   @Override
   public Collection<? extends Account> getTouchedAccounts() {
-    return accounts.values();
+    return accounts.values().stream().filter(Optional::isPresent).map(Optional::get).toList();
```
garyschulte, line 108: similar comment regarding deleted accounts.  a deleted account is a touched account.  

I am not sure if we want to indicate an eip-6780 transient account delete is a touched account though 🤔

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
lu-pinto: It's needed there for recursion and  since `parent.get()` returns an `Account`, even though it's always backed by a `SimpleAccount`. The `instanceof` does the null check and the cast in one go.

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
lu-pinto: Not sure what you mean in this particular method - is it about not returning the same collection size anymore? Apart from that I don't get why returning nulls in the collection is useful

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleAccount.java
```diff
@@ -214,21 +214,4 @@ public boolean commit() {
...
     }
   }
-
-  /**
-   * Push changes into the parent account, if one exists
-   *
-   * @return true if a parent account was updated, false if not (this indicates the account should
-   *     be inserted into the parent contact).
-   */
-  public boolean updateParent() {
```
lu-pinto, line 224: Same as `commit` above and unused. Let me know if it's used externally

evm/src/main/java/org/hyperledger/besu/evm/fluent/SimpleWorld.java
shemnon: 6780 transient deletes should never hit the simple world. The stacked or journaled updates above the simple world will elide it out.  Only OG SELFDESTRUCTSs will be seen through to the SimpleWorld.  Given the fluent API supports all forks, we need to be able to support that.

garyschulte: After looking at the usages of  getTouchedAccounts(), it seems we only use this interface in the context of accounts which will continue to exist after the commit, those which would be deleted (including eip-6780 self destructs) are always accessed via getDeletedAccounts.  The interface comment is misleading (or overloads the meaning for 'touched')

In the context of bonsai, trie logs, linea, etc - all touched accounts includes read, updated, and deleted.  But that isn't the context here.  

In short, LGTM :)

garyschulte (approved):

PR #7535, opened 2024-08-28
Fix incorrect key filtering in `LayeredKeyValueStorage` stream

Gabriel-Trintinalia:
## PR description

This PR addresses an issue in the `LayeredKeyValueStorage` where the `stream` method incorrectly filters out keys with empty entries in the current layer, leading to incorrect results by falling back to the parent layer.

In the `LayeredKeyValueStorage`, the stream method is designed to merge key-value pairs from the current layer and its parent layer. However, the current implementation filters out entries with empty values in the current layer using the following line:

```
ourLayerState.entrySet().stream()
    .filter(entry -> entry.getValue().isPresent())
    .iterator()
```

This behaviour is problematic because If a key exists in the parent layer but has an empty entry in the current layer, filtering out null values causes the method to ignore the empty entry marker and incorrectly return the key from the parent layer.

### Example Scenario
```
Parent Layer:
Key1 -> Value1
Key2 -> Value2

Current Layer:
Key1 -> empty
Key3 -> Value3
```

With filtering stream: (incorrect)
```
Key1 -> Value1 (from the parent layer)
Key2 -> Value2
Key3 -> Value3
```

Without filtering stream: (correct)
```
Key2 -> Value2
Key3 -> Value3
```

## Changes Made

- Removed the filtering step that excludes entries with null values in the stream method. 
- Updated the logic to correctly handle null values, ensuring that empty entries in the current layer are respected. 

This change resolves the issue where the stream method returned more accounts than expected in cases where the state was rolled back.


services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -178,61 +178,58 @@ public Stream<Pair<byte[], byte[]>> stream(final SegmentIdentifier segmentId) {
...
+      this.parentIterator = parentIterator;
+    }
+
+    @Override
+    public boolean hasNext() {
+      return ourIterator.hasNext() || parentIterator.hasNext();
+    }
+
+    private Pair<byte[], byte[]> mapEntryToPair(final Map.Entry<Bytes, Optional<byte[]>> entry) {
+      byte[] value = entry.getValue().orElse(null);
```
jframe, line 210: why does this now return null if they value is empty instead of new byte[0]?

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -178,61 +178,58 @@ public Stream<Pair<byte[], byte[]>> stream(final SegmentIdentifier segmentId) {
...
+
+      if (ourPeek == null || parentPeek == null) {
+        return ourPeek == null ? parentIterator.next() : mapEntryToPair(ourIterator.next());
+      }
+
+      int comparison = ourPeek.getKey().compareTo(Bytes.wrap(parentPeek.getKey()));
+      if (comparison < 0) {
+        return mapEntryToPair(ourIterator.next());
+      } else if (comparison == 0) {
+        parentIterator.next();
```
jframe, line 227: would be clear if the comment ```// skip dupe key from parent, return ours:``` was left in

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -178,61 +178,58 @@ public Stream<Pair<byte[], byte[]>> stream(final SegmentIdentifier segmentId) {
...
+    @Override
+    public Pair<byte[], byte[]> next() {
+      var ourPeek = ourIterator.peek();
+      var parentPeek = parentIterator.peek();
+
+      if (ourPeek == null || parentPeek == null) {
+        return ourPeek == null ? parentIterator.next() : mapEntryToPair(ourIterator.next());
+      }
+
+      int comparison = ourPeek.getKey().compareTo(Bytes.wrap(parentPeek.getKey()));
```
jframe, line 223: would be clear if the comment ```// otherwise compare:``` was left in

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -178,61 +178,58 @@ public Stream<Pair<byte[], byte[]>> stream(final SegmentIdentifier segmentId) {
...
-                .iterator());
-
-    final PeekingIterator<Pair<byte[], byte[]>> parentIterator =
+    PeekingIterator<Map.Entry<Bytes, Optional<byte[]>>> ourIterator =
+        new PeekingIterator<>(ourLayerState.entrySet().stream().iterator());
+    PeekingIterator<Pair<byte[], byte[]>> parentIterator =
         new PeekingIterator<>(parent.stream(segmentId).iterator());
 
     return StreamSupport.stream(
-        Spliterators.spliteratorUnknownSize(
```
jframe, line 192: Like the change to move the iterator out into it's own class

matkt:
could you add some tests :

Parent Layer 0:
Key1 -> empty
Key2 -> Value2

Parent Layer 1:
Key1 -> Value1
Key2 -> Value2

Current Layer:
Key1 -> empty
Key3 -> Value3

and 

Parent Layer 0:
Key1 -> Value
Key2 -> Value2

Parent Layer 1:
Key1 -> empty
Key2 -> Value2

Current Layer:
Key1 -> Value
Key3 -> Value3


and 

Parent Layer 0:
Key1 -> empty
Key2 -> Value2

Parent Layer 1:
Key1 -> empty
Key2 -> Value2

Current Layer:
Key1 -> Value
Key3 -> Value3

and 

Parent Layer 0:
Key1 -> empty
Key2 -> Value2

Current Layer:
Key1 -> Value
Key3 -> Value3

Just to be sure there are no regression

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
```diff
@@ -178,61 +178,60 @@ public Stream<Pair<byte[], byte[]>> stream(final SegmentIdentifier segmentId) {
     throwIfClosed();
     var ourLayerState = hashValueStore.computeIfAbsent(segmentId, s -> newSegmentMap());
 
-    // otherwise, interleave the sorted streams:
-    final PeekingIterator<Map.Entry<Bytes, Optional<byte[]>>> ourIterator =
-        new PeekingIterator<>(
-            ourLayerState.entrySet().stream()
-                .filter(entry -> entry.getValue().isPresent())
```
garyschulte, line 185: Removing this is the core change of this PR, right

This is a generic keyvalue layered storage facade, but LayeredKV storage exists essentially exclusively for bonsai.  I cannot think of a situation where we would want to filter out empty here and defer to the parent in the case of worldstate.

services/kvstore/src/main/java/org/hyperledger/besu/services/kvstore/LayeredKeyValueStorage.java
Gabriel-Trintinalia: Since the empty entries have already been filtered out in the current code, we won't encounter an `orElse` scenario here, right? This code should never happen. 

Since removing the initial filter, I'm setting it to null so I can remove it in the final stream.

jframe: Thanks, that makes sense now

plugins/rocksdb/src/test/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/LayeredKeyValueStorageTest.java
```diff
@@ -0,0 +1,339 @@
...
+
+    List<Pair<byte[], byte[]>> resultList = result.toList();
+    assertEquals(1, resultList.size());
+    assertArrayEquals(key, resultList.get(0).getKey());
+    // Layer value should be returned
+    assertArrayEquals(layerValue, resultList.get(0).getValue());
+  }
+
+  @Test
+  void shouldNotStreamKeyIfLayerKeyIsEmpty() {
```
lu-pinto, line 157: could you also create a test where `parent` also contains empty values to make sure they are also filtered out?

matkt: https://github.com/hyperledger/besu/pull/7535#issuecomment-2317016065

I think it was in the test list that I proposed to add? Or is it another case?

Gabriel-Trintinalia: yeah, `shouldStreamWithMultipleLayersAndCurrentLayerOverridesWithValues` tests this scenario

jframe (approved):

PR #7538, opened 2024-08-29
Layered txpool: fix for unsent drop notifications on remove

### LINKED ISSUE #7504: Layered txpool not sending drop notification when an invalid tx is removed during block creation
### Description

Currently the layered txpool only sends drop notifications when a tx is removed from the pool because it is full or when it is replaced, while it should also send this notifications when a tx is found invalid during block creation and when it is removed because it belongs to new imported block, so it is confirmed

fab-10:
## PR description

In the layered txpool, drop notifications were only sent in case the pending tx was evicted due to the txpool being full, but not in case it was confirmed in an imported block, or removed for other reasons, like when found invalid during the block creation process.

`LayersTest` has been expanded to also verify the expected add and drop notifications are sent.

## Fixed Issue(s)

fixes #7504 


pinges (approved):
LGTM

fab-10:
Thanks @pinges for the review, this is still in draft since it depends on https://github.com/hyperledger/besu/pull/7539 it you want to check it too

PR #7539, opened 2024-08-29
Layered txpool: do not send notifications when moving tx between layers

fab-10:
## PR description

Notifications for tx added to the layered txpool were incorrectly sent also when the tx was already added and was just internally moved between layers, potentially resulting in duplicated notifications.

Tests for this change are part of the following PR #7538 that is also fixing the drop notifications.

### DIFF
--- a/ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/layered/LayeredPendingTransactionsTest.java
+++ b/ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/layered/LayeredPendingTransactionsTest.java
@@ -20,6 +20,7 @@ import static org.hyperledger.besu.ethereum.eth.transactions.TransactionAddedRes
 import static org.hyperledger.besu.ethereum.eth.transactions.TransactionAddedResult.ALREADY_KNOWN;
 import static org.hyperledger.besu.ethereum.eth.transactions.TransactionAddedResult.NONCE_TOO_FAR_IN_FUTURE_FOR_SENDER;
 import static org.hyperledger.besu.ethereum.eth.transactions.TransactionAddedResult.REJECTED_UNDERPRICED_REPLACEMENT;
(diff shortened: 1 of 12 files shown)


ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/layered/LayeredPendingTransactionsTest.java
```diff
@@ -207,14 +208,14 @@ public void addRemoteTransactions() {
         createRemotePendingTransaction(transaction0), Optional.empty());
     assertThat(pendingTransactions.size()).isEqualTo(1);
 
-    assertThat(getAddedCount(REMOTE, NO_PRIORITY, layers.prioritizedTransactions.name()))
+    assertThat(getAddedCount(REMOTE, NO_PRIORITY, NEW, layers.prioritizedTransactions.name()))
```
pinges, line 211: I noticed that you are only using the reason "NEW" in this file. Is that because other reasons don't make a difference here? Do the tests in LayersTest cover all the functionality that has changed with the "reasons"?

pinges (approved):
Just one question, otherwise LGTM.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/transactions/layered/LayeredPendingTransactionsTest.java
fab-10: Using only `NEW` here because we are only adding new txs to the pool, and not moving existing txs between layers. Will review if more tests are needed to cover everything.

fab-10: added tests to cover also the move of txs between layers

PR #7544, opened 2024-08-29
Disable EOF CALL opcodes in legacy

shemnon:
## PR description

When the EOF call opcodes are added to the EVM, ensure legacy cannot it.


evm/src/main/java/org/hyperledger/besu/evm/operation/AbstractExtCallOperation.java
```diff
@@ -95,6 +95,11 @@ public long gasAvailableForChildCall(final MessageFrame frame) {
 
   @Override
   public OperationResult execute(final MessageFrame frame, final EVM evm) {
+    Code callingCode = frame.getCode();
+    if (callingCode.getEofVersion() == 0) {
+      return InvalidOperation.INVALID_RESULT;
```
jframe, line 100: Would be good to add some tests showing that an INVALID_RESULT is returned for legacy with EXTCALL, EXTDELEGATECALL and EXTSTATICCALL

shemnon: https://github.com/ethereum/execution-spec-tests/pull/768/files

Tests all legacy-invalid opcodes, not just EXT*CREATE.

garyschulte: LGTM, we can pull in the latest execution-spec tests when they get published :+1:

garyschulte (approved):
🚢

evm/src/main/java/org/hyperledger/besu/evm/operation/AbstractExtCallOperation.java
shemnon: https://github.com/hyperledger/besu/pull/7544/commits/f19e24279753b2a99af03b5f4f645961b632adca also should address @jframe's concern

PR #7547, opened 2024-08-30
Use updated calculateImplementationVersion for Implementation-Version in jar manifests

Matilda-Clerke:


Matilda-Clerke:
Went in a different direction. Closing this PR as no longer needed

PR #7548, opened 2024-08-30
Reintroduce engine get client version v1 with commit in manifest

### LINKED ISSUE #6471: engine_getClientVersionV1 - expose web3_clientVersion via EngineAPI

Matilda-Clerke:
re-fixes #6471 

for reference:
https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md


Matilda-Clerke:
Note: CHANGELOG.md may need to be updated if this is merged after the next release

config/build.gradle
```diff
@@ -22,7 +22,8 @@ jar {
       'Specification-Title': archiveBaseName,
       'Specification-Version': project.version,
       'Implementation-Title': archiveBaseName,
-      'Implementation-Version': calculateVersion()
+      'Implementation-Version': calculateVersion(),
+      'Commit-Hash': getGitCommitDetails(40).hash
```
garyschulte, line 26: are we doing this everywhere as a matter of consistency?  It seems like we only currently need this in the `:besu` artifact, but I don't see a reason not to be consistent across all

build.gradle
```diff
@@ -823,7 +823,7 @@ task distDocker {
         dockerPlatform = "--platform ${project.getProperty('docker-platform')}"
         println "Building for platform ${project.getProperty('docker-platform')}"
       }
-      def gitDetails = getGitCommitDetails(7)
+      def gitDetails = getGitCommitDetails()
```
garyschulte, line 826: 8 seems fine, but if we are going to generate a sha that is different from the build artifact name, we may as well use all 40 here

garyschulte (approved):
LGTM, nothing blocking (except pesky failing info regex tests)

config/build.gradle
Matilda-Clerke: Yes, I updated all of the gradle sub-modules where we're setting Implementation-Version to also set a Commit-Hash for consistency. Currently, it's only used in the besu module

PR #7551, opened 2024-09-01
Set docker VCS_REF back to a 7 char hash

Matilda-Clerke:
## PR description
Noticed in a PR for a hasty fix (https://github.com/hyperledger/besu/pull/7548) that the VCS_REF passed to docker build was unintentionally lengthened to 8 chars, when it should have stayed 7 chars.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -819,7 +819,7 @@ task distDocker {
         dockerPlatform = "--platform ${project.getProperty('docker-platform')}"
         println "Building for platform ${project.getProperty('docker-platform')}"
       }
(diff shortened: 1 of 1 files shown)


siladu (approved):

siladu:
Will this impact the release? Wondering if we might need to cherry pick it into 24.9.0?

Matilda-Clerke:
I don't believe this will break anything. I tested distDocker and testDocker locally and they both executed fine. It should be simple enough to cherry pick it to the release if desired, but nothing seems to be broken by having 8 chars instead of 7... this is just to revert the accidental change.

Issue #7445, opened 2024-08-12
engine_getBlobsV1


macfarla:
engine_getBlobsV1 . 

It seems pretty certain at this point that it will go into the hard fork. 

https://github.com/ethereum/execution-apis/pull/559

usmansaleem:
@pinges FYI

PR #7553, opened 2024-09-02
implement engine_getBlobsV1

pinges:
## PR description
Implements the engine RPC engine_getBlobsV1 as described in https://github.com/ethereum/execution-apis/pull/559/

## Fixed Issue(s)
#7445


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlobAndProofV1.java
```diff
@@ -0,0 +1,43 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.api.jsonrpc.internal.results;
+
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+/**
+ * BlobAndProofV1: title: Blob and proof object V1 type: object required: - blob - proof properties:
+ * blob: title: Blob $ref: '#/components/schemas/bytes' proof: title: proof $ref:
+ * '#/components/schemas/bytes48'
```
macfarla, line 22: simplify this comment

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1.java
```diff
@@ -0,0 +1,155 @@
...
+
+    BlobsWithCommitments.BlobQuad blobQuad = blobMap.get(vh);
+    if (blobQuad == null) {
+      blobQuad = blobCache.get(vh);
+    }
+
+    return blobQuad;
+  }
+
+  public void onTransactionAdded(final Transaction transaction) {
```
macfarla, line 134: idk if the RPC should be itself managing the blob cache?

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1Test.java
```diff
@@ -0,0 +1,285 @@
...
+
+    assertThat(blobAndProofV1s.size()).isEqualTo(versionedHashesList.size());
+    // for loop to check each blob and proof
+    for (int i = 0; i < versionedHashes.length; i++) {
+      assertThat(blobAndProofV1s.get(i)).isNull();
+    }
+  }
+
+  @Test
+  public void shouldRetrunEmptyResponseForEmptyRequest() {
```
macfarla, line 214:
```suggestion
  public void shouldReturnEmptyResponseForEmptyRequest() {
```

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreatorTest.java
```diff
@@ -371,7 +372,8 @@ private TransactionPool createTransactionPool(
             mock(TransactionBroadcaster.class),
             ethContext,
             new TransactionPoolMetrics(metricsSystem),
-            poolConf);
+            poolConf,
+            new BlobCache());
```
macfarla, line 376: @jflo can we daggerise the BlobCache or does it need other infra first?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1.java
pinges: I was wondering where that should live. The other option would be the TransactionPool. The BlobCache is taken care of in the TransactionPool, because these blobs are needed in case we have a reorg.
The blobMap is not needed by the TransactionPool. Here we just keep track of the blobs that we have available in our transaction pool and we have to update the map when blob transactions are added or removed from the pool.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/BlobAndProofV1.java
pinges: Done :-)

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreatorTest.java
pinges: It is kind of daggerised. It is available in BesuComponent, but I did not know how to use that in the engine RPC. Maybe @jflo can tell us!

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1.java
```diff
@@ -0,0 +1,155 @@
...
+    return new BlobAndProofV1(
+        bq.blob().getData().toHexString(), bq.kzgProof().getData().toHexString());
+  }
+
+  private BlobsWithCommitments.BlobQuad getBlobQuad(final VersionedHash vh) {
+
+    BlobsWithCommitments.BlobQuad blobQuad = blobMap.get(vh);
+    if (blobQuad == null) {
+      blobQuad = blobCache.get(vh);
+    }
```
fab-10, line 129: in theory this fallback should not be necessary, and removing the need to have the `blobCache` will simplify a lot the change, removing the need to pass it around, or am I missing something?

fab-10 (commented):
LGTM, see the note on the of passing `blobCache`, then add a changelog entry and doc-change-required label

fab-10:
@pinges note that https://github.com/hyperledger/besu/pull/7539 and https://github.com/hyperledger/besu/pull/7538 are a prerequisite for this one, otherwise it will not work correctly, so please do not merge it before the other 2 are in.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1.java
jflo: I also don't understand. It seems to me that blobMap is redundant, and this RPC should just use the same instance of the blobCache that the transactionpool does. Management of the cache contents should be ignored by the RPC and treated as a "read only" concern.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1.java
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * <p>4. Client software **MAY** return an array of all `null` entries if syncing or otherwise
+ * unable to serve blob pool data.
+ *
+ * <p>5. Callers **MUST** consider that execution layer clients may prune old blobs from their pool,
+ * and will respond with `null` if a blob has been pruned.
+ */
+public class EngineGetBlobsV1 extends ExecutionEngineJsonRpcMethod {
+
+  private final Map<VersionedHash, BlobsWithCommitments.BlobQuad> blobMap = new HashMap<>();
```
jflo, line 67: I think naming this something more specific would help me understand why it is necessary. As is, it reads like a synonym for blobCache, and seems like duplication.

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/PoWBlockCreatorTest.java
jflo: To make this a more complete approach, the RPC and the TransactionPool would need to be daggerized as well, which would allow all intermediate references to be removed.

jflo (changes requested):
Need to remove the responsibility of managing blobs from the RPC method.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1.java
pinges: The blob cache does contain all the blobs that have been put into a blob recently. These blobs are used to re-add the blob transactions in case of a reorg, as the blobs are not part of the block. After a 3 epochs these blobs are removed from the cache.
The blob map keeps track of all the blobs that are part of transactions that are in our transaction pool. These are the blobs that the CL will most likely ask for. We do keep them in the map for as long as their transactions are in the transaction pool. 
@fab-10 @jflo

fab-10: According to the specification of the new method, the CL should be interested in blobs that are in the txpool, and not blobs that have been already included in a block, so `blobMap` should be enough, because in case of a reorg a new add notification will be sent.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -104,6 +107,7 @@ public class TransactionPool implements BlockAddedObserver {
   private final SaveRestoreManager saveRestoreManager = new SaveRestoreManager();
   private final Set<Address> localSenders = ConcurrentHashMap.newKeySet();
   private final EthScheduler.OrderedProcessor<BlockAddedEvent> blockAddedEventOrderedProcessor;
+  private final Map<VersionedHash, BlobsWithCommitments.BlobQuad> blobMap = new HashMap<>();
```
fab-10, line 110: what about a more talking name, like `liveBlobs` or `pendingBlobs`?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineGetBlobsV1.java
pinges: I have moved that logic into the TransactionPool

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -647,6 +653,38 @@ public CompletableFuture<Void> setDisabled() {
...
+  }
+
+  public BlobsWithCommitments.BlobQuad getBlobQuad(final VersionedHash vh) {
+    BlobsWithCommitments.BlobQuad blobQuad = blobMap.get(vh);
+    if (blobQuad == null) {
+      blobQuad = blobCache.get(vh);
+    }
+    return blobQuad;
+  }
+
```
fab-10, line 687: it is better to have the logic here, and I am tempted to suggest to go a step further and extending the current `BlobCache` to manage both the confirmed blobs, as it does now, plus those _pending_ blobs. (However not a blocking request, just a suggestion to manage blobs in a single place)

pinges: I was thinking about that as well, but it felt like these are two different concerns and I think that the "map" should be part of the Transaction pool, as it points to the blobs of transactions in the pool.

fab-10:
> Much better. The only suggestion I have would be to consider using another instance of BlobCache internally for the pooled transaction, instead of just a map. That would afford some protection from unexpected cache growth, but I think trusting the add/drop tx events is a safe assumption.

That's a good point, and in the related PRs that are fixing the notifications in the layered txpool, I will make sure they are reliable, but I think it is worth to export a metrics with the size of the map, so we can monitor its growth, but this could be part of a following PR.

fab-10 (approved):

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - Performance optimzation for ECMUL (2 of 2) [#7543](https://github.com/hyperledger/besu/pull/7543)
 - Include current chain head block when computing `eth_maxPriorityFeePerGas` [#7485](https://github.com/hyperledger/besu/pull/7485)
 - Remove (old) documentation updates from the changelog [#7562](https://github.com/hyperledger/besu/pull/7562)
+- Add `engine_getBlobsV1` method to the Engine API [#7553](https://github.com/hyperledger/besu/pull/7553)
```
fab-10, line 29: Need to move this to the **Unreleased** section

fab-10 (changes requested):
LGTM, just move the changelog entry to the right place, and add `doc-change-required` label

fab-10:
FYI: task for adding the metrics https://github.com/hyperledger/besu/issues/7578

CHANGELOG.md
pinges: ups! Fixed!

CHANGELOG.md
```diff
@@ -40,7 +40,7 @@
 - Correctly drops messages that exceeds local message size limit [#5455](https://github.com/hyperledger/besu/pull/7507)
 - **DebugMetrics**: Fixed a `ClassCastException` occurring in `DebugMetrics` when handling nested metric structures. Previously, `Double` values within these structures were incorrectly cast to `Map` objects, leading to errors. This update allows for proper handling of both direct values and nested structures at the same level. Issue# [#7383](https://github.com/hyperledger/besu/pull/7383)
 - `evmtool` was not respecting the `--genesis` setting, resulting in unexpected trace results. [#7433](https://github.com/hyperledger/besu/pull/7433)
-- The genesis config override `contractSizeLimit` was not wired into code size limits [#7557](https://github.com/hyperledger/besu/pull/7557)
+- The genesis config override `contractSizeLimit`q was not wired into code size limits [#7557](https://github.com/hyperledger/besu/pull/7557)
```
fab-10, line 43: typo

fab-10 (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -640,6 +650,38 @@ public CompletableFuture<Void> setDisabled() {
...
+  private void mapBlobsOnTransactionAdded(
+      final org.hyperledger.besu.datatypes.Transaction transaction) {
+    final Optional<BlobsWithCommitments> maybeBlobsWithCommitments =
+        transaction.getBlobsWithCommitments();
+    if (maybeBlobsWithCommitments.isEmpty()) {
+      return;
+    }
+    final List<BlobsWithCommitments.BlobQuad> blobQuads =
+        maybeBlobsWithCommitments.get().getBlobQuads();
+    blobQuads.forEach(bq -> mapOfBlobsInTransactionPool.put(bq.versionedHash(), bq));
```
karalabe, line 662: This can be used my a malicious transaction to corrupt the mapping by sending the same blobs as a benign tx. Then send a replacement to nuke out all blobs, and the original correct mapping disappears, making the blob non-retrievable any more.

PR #7571, opened 2024-09-04
Update Java and Gradle dependencies

fab-10:
## Fixed Issue(s)

2 small changes are required after the update, to adapt to the change in the way ` assertThat(response).usingRecursiveComparison()` works, and a new field in the `Suppliers::memoize` class that increment the memory size of pending txs.


ethereum/api/src/integration-test/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/fork/frontier/EthCreateAccessListIntegrationTest.java
```diff
@@ -236,7 +236,12 @@ private void assertAccessListExpectedResult(
         new JsonRpcSuccessResponse(null, new CreateAccessListResult(accessList, gasUsed));
 
     final JsonRpcResponse response = method.response(request);
-    assertThat(response).usingRecursiveComparison().isEqualTo(expectedResponse);
+    assertThat(response)
```
jflo, line 239: Is this required for the dependency upgrade somehow?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransaction.java
```diff
@@ -31,8 +31,8 @@
 public abstract class PendingTransaction
     implements org.hyperledger.besu.datatypes.PendingTransaction {
   static final int NOT_INITIALIZED = -1;
-  static final int FRONTIER_AND_ACCESS_LIST_SHALLOW_MEMORY_SIZE = 888;
-  static final int EIP1559_AND_EIP4844_SHALLOW_MEMORY_SIZE = 1000;
+  static final int FRONTIER_AND_ACCESS_LIST_SHALLOW_MEMORY_SIZE = 904;
```
jflo, line 34: surprised to see this in this pr

jflo (approved):
Couple of surprising files to be modified in here

ethereum/api/src/integration-test/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/fork/frontier/EthCreateAccessListIntegrationTest.java
fab-10: it is required because with the update, `hashCode` fields are compared, and some tuweni classes compute the field lazily, so even if the object are equals, the `hashCode` field in the `expectedResponse` is `null`, so for those classes instead of a field by field comparison we use the `equals`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransaction.java
fab-10: This is because the new guava version is adding a new field to the class returned by `Suppliers::memoize`

PR #7575, opened 2024-09-04
[#4292] Fix mounted data path directory permissions for besu user

### LINKED ISSUE #4292: Read file permission denied when starting besu from docker image
### Description

Hello, I'm looking for help with running besu from docker image. The problem I am running into is mounting a named volume (which i named "besu") to a path i specified with `--data-path`. The container errors out saying:

```
Failed to retrieve the RocksDB database meta version: /var/lib/besu/DATABASE_METADATA.json (Permission denied) could not be found. You may not have the appropriate permission to access the item.
```

I suspect that since user is changed from [root to "besu"](https://github.com/hyperledger/besu/blob/332ca416ba6c995605cad3b25b3fc1441879407c/docker/graalvm/Dockerfile#L8), the user "besu" does not have permissions to write in the `--data-path` dir that i specified.

i can work around this by setting the docker run config `--user=0` to switch back to root, but this obviously doesn't seem like the most optimal solution.

### Acceptance Criteria
* I can mount a volume to the besu `--data-path` config without overriding the user in the container
or
* volume mount is automatically done by the container

### Steps to Reproduce (Bug)
1. Create a named volume called "besu": `docker volume create besu`
2. run `docker run --name besu  -v besu:/var/lib/besu hyperledger/besu:latest --data-path=/var/lib/besu`

**Expected behavior:** 

expected to run besu without error (i.e. container user has permissions to read/write to specified volume)

**Actual behavior:**

container errors out with message seen in the Description
100%

* Software version: hyperledger/besu:latest (docker image)
* OS Name & Version: Ubuntu 22.04.1 LTS
* Kernel Version: 5.15.0-46-generic
* Docker Version: 20.10.17

Edit: Updated issue description with given layout (sorry didn't see when clicking "Reference in new issue" from code line!)

pullurib:
## PR description

- Fix data path directory permissions for besu user by adding an entry script to update the permissions depending on the data path parameter passed while running the docker image 
- Script is run as root user and switches to besu user at the end before launching besu 

Looking to get approval for this approach at this point . Has more work todo 
- Manually copied the entry script to install path for now - this should be part of build. 
- These checks and updates in entry script can be moved to main besu launch script 
- More checks should be added to get and validate data path as described in the entry script comments 

## Fixed Issue(s)
fixes #4292


fab-10:
Your approach looks fine, and I would like to review a more elegant and future proof way to detect which are the paths that need to be owned by Besu, I was first thinking of a kind of Docker _introspection_ where from inside the container it is possible to detect which are the mounted volumes, but then I thought could be better to delegate this directly to Besu, I mean having a dedicated subcommand, or option, that given the passed options, or configuration, just prints the data dir path and exit. WDYT?

Also will be nice to have a test that reproduce the use case, so we can add it to our CI and detect possible regressions.

pullurib:
Thanks for taking a look. Besu subcommand approach sounds good. Since the options can be set through CLI, env vars or config file, we can let the subcommand consume options passed , check read/write access for required paths (data path and few more )  and output the paths which need adjusting the permissions . 
The entry script would first run besu with this subcommand as a child, as besu user, get the list of paths and adjust permissions for besu users and finally run besu 

list of options to include in the check now (which can be extended in future) 
- --data-path
-  --genesis-file
- --rpc-ws-authentication-credentials-file
- --tx-pool-save-file
- --config-file

Does this look ok?

I'll add the test case

fab-10:
@pullurib yes it looks ok, not sure at the moment if it is better to do this via a subcommand or via a special option like `--print-paths-and-exit` but probably you will get a better idea working on it.

pullurib:
@fab-10 went with --print-paths-and-exit option as the config has to be resolved from options passed , env vars and config files which is already done in besu main command . Added a test case in docker tests

fab-10 (commented):
Looks good, just some comments to make the options more generic for other possible use cases, and other little things.

pullurib:
@fab-10 , thanks for the valuable suggestions. I made the review changes along with CHANGELOG addition. I see that doc changes would go into a different repo. Once this is merged, I can update the docs and link that PR here . Or should we wait to merge both at about same time ?

CHANGELOG.md
```diff
@@ -11,8 +11,13 @@
 - Remove privacy test classes support [#7569](https://github.com/hyperledger/besu/pull/7569)
 
 ### Bug fixes
+- Layered txpool: do not send notifications when moving tx between layers [#7539](https://github.com/hyperledger/besu/pull/7539)
+- Layered txpool: fix for unsent drop notifications on remove [#7538](https://github.com/hyperledger/besu/pull/7538)
+- Honor block number or tag parameter in eth_estimateGas and eth_createAccessList [#7502](https://github.com/hyperledger/besu/pull/7502)
```
fab-10, line 16: Removing entries that were released on 24.9.1

```suggestion
```

fab-10 (commented):
just some final touches before approval, check the suggested change for the CHANGELOG, and since this options seems to be platform dependent, it works on Linux, not sure on MacOs, but for sure will not work on Linux, then I suggest to check the OS if the option is passed and report an error message on not supported OSes.
You can use the already provided dependency for system info https://www.oshi.ooo/oshi-core-java11/apidocs/com.github.oshi/oshi/SystemInfo.html#getCurrentPlatform()

fab-10:
> I see that doc changes would go into a different repo. Once this is merged, I can update the docs and link that PR here . Or should we wait to merge both at about same time ?

You can do the update on your own, or you can wait that someone else will pick up the doc task.

pullurib:
Added platform checks. The option works fine on MacOS too, so checking for Linux or Mac for now to be able to use the option. Thanks for the CHANELOG correction, included it

fab-10 (approved):

fab-10:
@pullurib I have merged the PR, but on main the CI report this error https://github.com/hyperledger/besu/actions/runs/10919521049/job/30307346592, could you check?

pullurib:
Raised a small PR https://github.com/hyperledger/besu/pull/7637 for hadolint failure fix, please take a look

m4sterbunny:
@pullurib Corresponding documentation issue raised: https://github.com/hyperledger/besu/issues/7667

PR #7576, opened 2024-09-04
Remove tx from pool when its score is lower than a configured value

### LINKED ISSUE #7572: Remove a tx from the txpool when its score is below a configurable value
### Description

#7359 introduced a score for txs in the poll, but txs are never removed even if their score reached the min value, in some cases it could be useful to have a way to tell to the txpool to just remove txs after their score is below a configurable value, that basically means that the tx was tried for block inclusion a certain number of times without success, so removing it could prevent future resource waste.

fab-10:
## PR description

Introduce a new option `--tx-pool-min-score`, that takes values between `-128` and `127`, that instruct the txpool to remove pending txs when their score is below the option value.

By default the value is `-128`, that is also the lowest score a pending tx can have, and basically means that the pending tx will not be removed and will stay in the pool with the lowest score, so it will be selected after all the other pending txs

## Fixed Issue(s)

fixes #7572 


besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -196,6 +197,15 @@ static class Layered {
             "Max number of future pending transactions allowed for a single sender (default: ${DEFAULT-VALUE})",
         arity = "1")
     Integer txPoolMaxFutureBySender = TransactionPoolConfiguration.DEFAULT_MAX_FUTURE_BY_SENDER;
+
+    @CommandLine.Option(
+        names = {TX_POOL_MIN_SCORE},
+        paramLabel = "<Byte>",
+        description =
+            "Keep a pending transaction in the txpool until its score is greater than or equal to this value."
```
pinges, line 205:
```suggestion
            "Remove a pending transaction from the txpool if its score is lower than this value."
```

pinges (approved):
One proposal, otherwise LGTM

PR #6965, opened 2024-04-18
Implementing support for emptyBlockPeriodSeconds in QBFT (Issue #3810)

amsmota:
## PR description
Implementing emptyblockperiodseconds for producing empty blocks at a specific interval independently of the value of the existing blockperiodseconds setting

## Fixed Issue(s)
https://github.com/hyperledger/besu/issues/3810

### Thanks for sending a pull request! Have you done the following?

- [x] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [ ] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [ ] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [x] unit tests: `./gradlew build`
- [ ] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

PR #7588, opened 2024-09-09
Dev/test option for short BFT block periods

matthew1001:
## PR description

This is a test-only option to mine BFT blocks more quickly than once a second.

It is not designed to be used in production and I've added warnings to the logs when it is set.

The reason for adding it is for Web3 frameworks wanting to run tests against a Besu node, possibly as part of a CI/CD pipeline, where lots of transactions & receipts need round-tripping during the test. 1 second block periods mean a test could take a very long time to complete.

I've tested it down to `50ms` blocks and a single node is stable enough to mine transactions into blocks at that speed. I've also tested it in conjunction with draft PR https://github.com/hyperledger/besu/pull/6965 with `emptyblockperiodseconds` set to 10s (so while the test isn't driving TXN traffic Besu isn't spinning creating huge numbers of empty blocks) and the two work fine together - see the following logs with `300ms` blocks:

image


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BlockTimer.java
```diff
@@ -24,15 +24,53 @@
...
 
+  private static final Logger LOG = LoggerFactory.getLogger(BlockTimer.class);
+
   private final ForksSchedule<? extends BftConfigOptions> forksSchedule;
   private final BftExecutors bftExecutors;
   private Optional<ScheduledFuture<?>> currentTimerTask;
   private final BftEventQueue queue;
   private final Clock clock;
 
+  // For testing only, it can be useful to set BFT block periods less than a second. This allows
```
jframe, line 41: I know this is only for testing, but I still don't like the environment variable parsing in the main code here. Adding an experimental config option to QbftConfig would be much nicer. It's quite easy to pass the value into this class too since we create this in the QbftBesuControllerBuilder.

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/RoundTimer.java
```diff
@@ -45,7 +45,12 @@ public RoundTimer(
     this.queue = queue;
     this.bftExecutors = bftExecutors;
     this.currentTimerTask = Optional.empty();
-    this.baseExpiryMillis = baseExpirySeconds * 1000;
```
jframe, line 48: If the baseExpiryMillis type was changed to a duration, then we could handle both milliseconds and seconds without any special logic

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/headervalidationrules/TimestampMoreRecentThanParent.java
```diff
@@ -45,7 +51,11 @@ private boolean validateTimestamp(final long timestamp, final long parentTimesta
   private boolean validateHeaderSufficientlyAheadOfParent(
       final long timestamp, final long parentTimestamp) {
     final long secondsSinceParent = timestamp - parentTimestamp;
-    if (secondsSinceParent < minimumSecondsSinceParent) {
+    if (usingExperimentalShortBFTBlockPeriod) {
```
jframe, line 54: It would be better to disable this in the QbftBlockHeaderValidationRulesetFactory based on the config from QbftConfigOptions

matthew1001:
Thanks for the comments @jframe, agree with you on all counts. I'll update the PR shortly

matthew1001:
I've done some refactoring to the PR as follows:

- Removed the env var `BESU_X_DEV_BFT_PERIOD_MS` and created `xblockperiodmilliseconds` as a QBFT genesis config option
- If `xblockperiodmilliseconds` is set `>0` then `blockperiodseconds` is ignored
- Refactored several areas of block and round handling to pass `Duration` rather than seconds as an `int`

matthew1001:
Thanks for the comments @jframe See the latest commit & comment which hopefully resolves your comments

jframe (approved):

PR #7594, opened 2024-09-10
Make access event keys eip 6800 compatible

lu-pinto:
## PR description
This PR encodes balance, nonce, version and code size into a single leaf in Verkle, while code hash has its own leaf, as mandated by EIP-6800.
I'm not doing fixes to gas costs for EIP-4762 yet, most changes are just to superficially not charge for the same leaf twice and to make code compile. I will implement EIP-4762 gas costs in a subsequent PR.

## Fixed Issue(s)

This is needed for fixing gas costs for EIP-4762


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -206,124 +196,75 @@ protected Hash internalCalculateRootHash(
...
-                            + updatedAccount.getBalance());
-                    stateTrie.put(bytes, bytes2);
-                  });
-          maybeStateUpdater.ifPresent(
-              bonsaiUpdater ->
-                  bonsaiUpdater.putAccountInfoState(hashAndSavePreImage(accountKey), accountValue));
-        }
-      }
+    if (accountUpdate.isUnchanged()) {
+      return;
```
matkt, line 209: we cannot use if, else if , else if ? I think it will be better

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -206,124 +196,75 @@ protected Hash internalCalculateRootHash(
...
+          bonsaiUpdater -> bonsaiUpdater.putCode(accountHash, codeHash, codeUpdate.getUpdated()));
     }
   }
 
-  private boolean codeIsEmpty(final Bytes value) {
-    return value == null || value.isEmpty();
-  }
-
-  private void updateAccountStorageState(
+  private void generateStorageValue(
```
matkt, line 260: generateStorageValue-S ?

ethereum/verkletrie/src/main/java/org/hyperledger/besu/ethereum/verkletrie/TrieKeyPreloader.java
```diff
@@ -40,22 +40,16 @@ public class TrieKeyPreloader {
   public TrieKeyPreloader() {
     this.hasher = new PedersenHasher();
     trieKeyAdapter = new TrieKeyBatchAdapter(hasher);
-    trieKeyAdapter.versionKey(
-        Address.ZERO); // TODO REMOVE is just to preload the native library for performance check
   }
 
   public List<Bytes32> generateAccountKeyIds() {
```
matkt, line 45: we still need a lit for this one ?

matkt (commented):
I like the proposed modification, I just have some remarks and we must not forget to add the equals method in VerkleAccount.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/VerkleAccount.java
```diff
@@ -37,6 +37,7 @@
 
 public class VerkleAccount extends DiffBasedAccount {
   private Hash storageRoot; // TODO REMOVE AS USELESS
+  private int hashCode;
```
garyschulte, line 40: should we remove storageRoot while we are here

lu-pinto: good shout. @matkt mentioned he will remove it in future PR so I didn't want to create conflicts unnecessarily

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -366,6 +292,69 @@ private void updateAccountStorageState(
...
+        .forEach(
+            (key, value) -> {
+              System.out.println("add key " + key + "  leaf value " + value);
+              stateTrie.put(key, value);
+            });
+    verkleEntryFactory
+        .getStorageKeyValuesForUpdate()
+        .forEach(
+            (storageSlotKey, pair) -> {
+              var storageAccountUpdate = worldStateUpdater.getStorageToUpdate().get(accountKey);
```
matkt, line 340: I don't think this one is still needed as we already checked here https://github.com/hyperledger/besu/pull/7594/files#diff-a8c0068954d44abe33b85a9a6c2b70309adb959255e417350abe54c858574b62R267 but maybe I missing something ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -206,124 +196,77 @@ protected Hash internalCalculateRootHash(
...
+      verkleEntryFactory.generateAccountKeysForRemoval(accountKey);
+      final Hash addressHash = hashAndSavePreImage(accountKey);
+      maybeStateUpdater.ifPresent(
+          bonsaiUpdater -> bonsaiUpdater.removeAccountInfoState(addressHash));
+      return;
+    }
+    final Bytes priorValue =
+        accountUpdate.getPrior() == null ? null : accountUpdate.getPrior().serializeAccount();
+    final Bytes accountValue = accountUpdate.getUpdated().serializeAccount();
+    if (!accountValue.equals(priorValue)) {
```
matkt, line 221: do we still need this check if we have now https://github.com/hyperledger/besu/pull/7594/files#diff-a8c0068954d44abe33b85a9a6c2b70309adb959255e417350abe54c858574b62R208 ?

matkt (approved):
Added comment but approve to unblock

lu-pinto:
I noticed that adding `VerkleAccount.equals/hashcode` changes the state root in `LogRollingTests`. I'm going to remove that commit for now and should be investigated next week

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
lu-pinto: Not sure I understand. So the link you posted is when the key values are generated. Here we commit them to the trie in one go.
On generation we only have the `storageSlotKey` and corresponding key/value. We don't have the `storageUpdate` ref where we should save the prior value

matkt: I was talking about this check (==null) 
https://github.com/hyperledger/besu/pull/7594/files/2d12b659a7478f155f9a7be52e1dfed6dfcb3c7d#diff-a8c0068954d44abe33b85a9a6c2b70309adb959255e417350abe54c858574b62R341

when we are generating the key we already check if the collection is empty, so imo when we enter this loop we cannot have a null

lu-pinto: yes you're right that we if there are elements we can't have null. I added it based on a defensive approach since we don't know how the programmer will generate keys and values. What if it's based on another object?
If someone changes the code above, then we will not break here. I would prefer to leave it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -355,6 +358,16 @@ private void updateState(
             });
   }
 
+  public Set<Address> getAddressesToPersist(
+      final DiffBasedWorldStateUpdateAccumulator<?> accumulator) {
+    Set<Address> mergedAddresses =
+        new HashSet<>(accumulator.getAccountsToUpdate().keySet()); // accountsToUpdate
+    mergedAddresses.addAll(accumulator.getCodeToUpdate().keySet()); // codeToUpdate
+    mergedAddresses.addAll(accumulator.getStorageToClear()); // storageToClear
+    mergedAddresses.addAll(accumulator.getStorageToUpdate().keySet()); // storageToUpdate
```
ahamlat, line 362: We need to check the performances of these merges. Depending on the number of addresses, it can have a non negligeable overhead.

matkt (approved):
LGTM

PR #7599, opened 2024-09-10
CI Fix, propagate release name preprocessing

garyschulte:
## Fixed Issue(s)

Fixes pre-processed release name propagation in the release workflow jobs via a dedicated preprocessing job.

Post-24.8.0 release, preprocessing of release names was added to ensure they are well formed.  The envar method of propagation is limited to a single job however.  Subsequent release workflow jobs do not have this value and are failing.

This PR:
* moves release name preprocessing to its own job
* makes release_name a job output
* for each job that require an explicit release version:
  * adds a dependency on `pre_process_release_name`
  * declares an envar RELEASE_NAME from the output of the preprocess_release_name job

Verified on a [fork](https://github.com/garyschulte/besu/actions/runs/10801154312/job/29961599252) that release name is propagated to the dependent jobs: 
- [x] artifacts
- [x] buildDocker
- [x] artifactoryPublish
- [x] multiArch
- [x] amendNotes
- [x] dockerPromoteX64
- [x] verifyContainer

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,8 +9,28 @@ env:
   GRADLE_OPTS: "-Dorg.gradle.parallel=true -Dorg.gradle.caching=true"
 
 jobs:
+  preprocess_release:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Pre-process Release Name
+        id: pre_process_release_name
+        run: |
+          RELEASE_NAME="${{ github.event.release.name }}"
+          # strip all whitespace
+          RELEASE_NAME="${RELEASE_NAME//[[:space:]]/}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -9,8 +9,28 @@ env:
   GRADLE_OPTS: "-Dorg.gradle.parallel=true -Dorg.gradle.caching=true"
 
 jobs:
+  preprocess_release:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Pre-process Release Name
+        id: pre_process_release_name
```
siladu, line 16: nit: preprocess vs pre_process

...only because it feels ripe for a future bug in this case!

siladu (approved):
Nice fix. The variable context scope is such a common error...seems to happen almost every time we touch these workflows 😓

Issue #7578, opened 2024-09-05
Export new metrics to report the size of blob cache and map


fab-10:
### Description
As an devops, I want 2 metrics with the size of the `TransactionPool.cacheForBlobsOfTransactionsAddedToABlock` and `TransactionPool.mapOfBlobsInTransactionPool` so that I can monitor how they grow.

KoxyG:
Hi @fab-10 inwould love to work on this issue. Kindly pls assign this to me

fab-10:
Thanks @KoxyG for your interest in contributing, but it seems @7suyash7 has already a PR, if you want to can review it to compare with your idea or search for some other tasks.

macfarla:
fixed by #7622

PR #7622, opened 2024-09-16
Add BlobMetrics

7suyash7:
Add support for metrics, besu_blob_storage_cache_size: `TransactionPool.cacheForBlobsOfTransactionsAddedToABlock` and besu_blob_storage_blob_map_size: `TransactionPool.mapOfBlobsInTransactionPool`

image

#7578


7suyash7:
> Since these metrics are related to the txpool, you can add them to the existing `TransactionPoolMetrics` class, that should also make the whole PR much smaller, since you do not need to pass around the metrics system.

@fab-10 I just updated it, can you please take a look.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -109,6 +109,8 @@ public class TransactionPool implements BlockAddedObserver {
   private final EthScheduler.OrderedProcessor<BlockAddedEvent> blockAddedEventOrderedProcessor;
   private final Map<VersionedHash, BlobsWithCommitments.BlobQuad> mapOfBlobsInTransactionPool =
       new HashMap<>();
+  private final ConcurrentHashMap<VersionedHash, BlobsWithCommitments.BlobQuad> blobCacheTracker =
+      new ConcurrentHashMap<>();
```
fab-10, line 113: what is this for? it seems to have no real usage

metrics/core/src/main/java/org/hyperledger/besu/metrics/BesuMetricCategory.java
```diff
@@ -55,7 +55,9 @@ public enum BesuMetricCategory implements MetricCategory {
   /** Stratum besu metric category. */
   STRATUM("stratum"),
   /** Block processing besu metric category. */
-  BLOCK_PROCESSING("block_processing");
+  BLOCK_PROCESSING("block_processing"),
+  /** Blob storage besu metric category. */
+  BLOB_STORAGE("blob_storage");
```
fab-10, line 60: not used

fab-10 (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
7suyash7: It's being used to maintain the count of blobs in the cache...

fab-10: better to add a `size` method to `BlobCache`, it is simpler and avoid keeping this extra map

fab-10 (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
7suyash7: Okay, I was trying to avoid doing that... I have added it now and removed the extra map

fab-10 (commented):
LGTM, just add a CHANGELOG entry and its done

7suyash7:
> LGTM, just add a CHANGELOG entry and its done

Done!

fab-10 (approved):

PR #7628, opened 2024-09-17
7311: add peertask foundation code

Matilda-Clerke:
## PR description
This PR adds the foundational code for the new peer task system. For details about how this is used, please see the[ spike PR](https://github.com/hyperledger/besu/pull/7602) proving the concept is workable.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+        executorResult =
+            new PeerTaskExecutorResult<>(null, PeerTaskExecutorResponseCode.INVALID_RESPONSE);
+
+      } catch (ExecutionException e) {
+        executorResult =
+            new PeerTaskExecutorResult<>(null, PeerTaskExecutorResponseCode.INTERNAL_SERVER_ERROR);
+      }
+    } while (--triesRemaining > 0
+        && executorResult.getResponseCode() != PeerTaskExecutorResponseCode.SUCCESS
+        && executorResult.getResponseCode() != PeerTaskExecutorResponseCode.PEER_DISCONNECTED
```
pinges, line 125: If you don't care about which peer is used you would try with another peer ...

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+        && sleepBetweenRetries(WAIT_TIME_BEFORE_RETRY[triesRemaining]));
```
pinges, line 126: could be handled by the peer manager

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+  private boolean sleepBetweenRetries(final long sleepTime) {
+    try {
+      Thread.sleep(sleepTime);
+      return true;
+    } catch (InterruptedException e) {
+      return false;
+    }
+  }
+
+  private static boolean isPeerUnused(
```
pinges, line 145: should be handled by the peer manager

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+      final EthPeer ethPeer, final Collection<EthPeer> usedEthPeers) {
+    return !usedEthPeers.contains(ethPeer);
+  }
+
+  private static boolean isPeerHeightHighEnough(final EthPeer ethPeer, final long requiredHeight) {
```
pinges, line 150: should be handled by the peer manager

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+    return ethPeer.chainState().getEstimatedHeight() >= requiredHeight;
+  }
+
+  private static boolean isPeerProtocolSuitable(final EthPeer ethPeer, final String protocol) {
```
pinges, line 154: should be handled by the peer manager

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskRequestSender.java
```diff
@@ -0,0 +1,55 @@
...
+import org.hyperledger.besu.ethereum.p2p.rlpx.connections.PeerConnection;
+import org.hyperledger.besu.ethereum.p2p.rlpx.wire.MessageData;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class PeerTaskRequestSender {
+  private static final long DEFAULT_TIMEOUT_MS = 20_000;
```
pinges, line 28: I think at the moment we have a timeout of 5s

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskRequestSender.java
```diff
@@ -0,0 +1,55 @@
...
+          InterruptedException,
+          TimeoutException {
+    ResponseStream responseStream =
+        ethPeer.send(requestMessageData, subProtocol, ethPeer.getConnection());
+    final CompletableFuture<MessageData> responseMessageDataFuture = new CompletableFuture<>();
+    responseStream.then(
+        (boolean streamClosed, MessageData message, EthPeer peer) -> {
+          responseMessageDataFuture.complete(message);
+        });
+    return responseMessageDataFuture.get(timeoutMs, TimeUnit.MILLISECONDS);
```
pinges, line 53: I like that!

pinges (commented):
Left some comments

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerManager.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Predicate;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** "Manages" the EthPeers for the PeerTaskExecutor */
+public class PeerManager {
```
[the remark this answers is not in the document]
Matilda-Clerke: An interface to allow easy swapping of different implementations is a fine, but I'm not sure how we'd implement retries in the peer manager... the peer manager is completely separated from the PeerTasks and has no idea what tasks are being executed and whether they succeed or fail.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+    this.requestSender = requestSender;
+    this.protocolSpecSupplier = protocolSpecSupplier;
+    requestTimer =
+        metricsSystem.createLabelledTimer(
+            BesuMetricCategory.PEERS, "Peer Task Executor Request Time", "", "Task Class Name");
+  }
+
+  public <T> PeerTaskExecutorResult<T> execute(final PeerTask<T> peerTask) {
+    PeerTaskExecutorResult<T> executorResult;
+    int triesRemaining =
```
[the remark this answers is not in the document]
Matilda-Clerke: I'm not sure I understand how that would work. The peer manager doesn't have any connection to the tasks being run

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: This is in the `executeAgainstPeer()` method, so at this point in the code we do care which peer is used. If the calling code doesn't care which peer a task is run against, it can call the `execute()` method instead, which contains the logic for peer switching.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: I'm not sure how exactly. The peer manager doesn't (and shouldn't) know anything about whether a peer is requested for a retry or for a first attemp.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: These methods are here because they are used in the predicate specifically to request peers for the peer task executor. They may not be applicable to all calls to `peerManager.getPeer`.

They certainly could be moved to PeerManager, but that more tightly couples the peer manager to this one specific use case. Maintaining separation is what gives and maintains the simplicity of this code, as compared with the old system.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskRequestSender.java
Matilda-Clerke: Thanks, I'll update it to 5s for now. In future, it might be good to make this configurable somehow, so end users can tune it to work best in their environment. But that's definitely out of scope for this work.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+        && executorResult.getResponseCode() != PeerTaskExecutorResponseCode.SUCCESS);
+
+    return executorResult;
+  }
+
+  public <T> CompletableFuture<PeerTaskExecutorResult<T>> executeAsync(final PeerTask<T> peerTask) {
+    return CompletableFuture.supplyAsync(() -> execute(peerTask));
+  }
+
+  public <T> PeerTaskExecutorResult<T> executeAgainstPeer(
```
siladu, line 87: private?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+        executorResult =
+            new PeerTaskExecutorResult<>(null, PeerTaskExecutorResponseCode.NO_PEER_AVAILABLE);
+      }
+    } while (--triesRemaining > 0
+        && executorResult.getResponseCode() != PeerTaskExecutorResponseCode.SUCCESS);
+
+    return executorResult;
+  }
+
+  public <T> CompletableFuture<PeerTaskExecutorResult<T>> executeAsync(final PeerTask<T> peerTask) {
```
siladu, line 83: unused, better IMO to add it in as needed.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+      }
+    } while (--triesRemaining > 0
+        && executorResult.getResponseCode() != PeerTaskExecutorResponseCode.SUCCESS
+        && executorResult.getResponseCode() != PeerTaskExecutorResponseCode.PEER_DISCONNECTED
+        && sleepBetweenRetries(WAIT_TIME_BEFORE_RETRY[triesRemaining]));
+
+    return executorResult;
+  }
+
+  public <T> CompletableFuture<PeerTaskExecutorResult<T>> executeAgainstPeerAsync(
```
siladu, line 131: unused

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
+
+/** Manages the execution of PeerTasks, respecting their PeerTaskBehavior */
+public class PeerTaskExecutor {
+  private static final long[] WAIT_TIME_BEFORE_RETRY = {0, 20000, 5000};
```
siladu, line 35: What's the thinking behind these retry timings? Warrants a comment I think.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
siladu: Why do we need to sleep between retries?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+      EthPeer peer;
+      try {
+        peer =
+            peerManager.getPeer(
+                (candidatePeer) ->
+                    isPeerUnused(candidatePeer, usedEthPeers)
+                        && (protocolSpecSupplier.get().isPoS()
+                            || isPeerHeightHighEnough(
+                                candidatePeer, peerTask.getRequiredBlockNumber()))
+                        && isPeerProtocolSuitable(candidatePeer, peerTask.getSubProtocol()));
```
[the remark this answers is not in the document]
siladu: Also struggling to find test coverage for this logic.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskRequestSenderTest.java
```diff
@@ -0,0 +1,73 @@
...
+    CompletableFuture<MessageData> actualResponseMessageDataFuture =
+        CompletableFuture.supplyAsync(
+            () -> {
+              try {
+                return peerTaskRequestSender.sendRequest(subprotocol, requestMessageData, peer);
+              } catch (Exception e) {
+                throw new RuntimeException(e);
+              }
+            });
+    Thread.sleep(500);
```
siladu, line 64: Why do we need this in the test?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutorTest.java
siladu, line 1: Nice coverage!

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/peertask/DefaultPeerManagerTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public DefaultPeerManager peerManager;
+
+  @BeforeEach
+  public void beforeTest() {
+    peerManager = new DefaultPeerManager();
+  }
+
+  @Test
+  public void testGetPeer() throws NoAvailablePeerException {
```
siladu, line 41: would be nice to test the exceptional case too

siladu (commented):
Ended up making quite a few comments as I went through.

TL;DR: No major objections. The way peers are selected needs some more consideration IMO as it seems a bit spread out across the classes.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: We don't strictly need to sleep between retries, but if a peer is unable to respond before timing out, it's likely just busy or we're experiencing network issues, so waiting a moment before retrying can allow things a chance to improve.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerManager.java
```diff
@@ -0,0 +1,57 @@
...
+package org.hyperledger.besu.ethereum.eth.manager.peertask;
+
+import org.hyperledger.besu.ethereum.eth.manager.EthPeer;
+import org.hyperledger.besu.ethereum.p2p.peers.PeerId;
+
+import java.util.Optional;
+import java.util.function.Predicate;
+
+/** "Manages" the EthPeers for the PeerTaskExecutor */
+public interface PeerManager {
```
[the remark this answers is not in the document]
Matilda-Clerke: Yep, I agree. PeerSelector more accurately describes the classes responsibility.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerManager.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Gets the highest reputation peer matching the supplies filter
```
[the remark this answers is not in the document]
Matilda-Clerke: That's fair. Originally, there was no interface, so the javadoc described the implementation. I'll reword it

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
siladu: My understanding was this is a refactor so I think we should match the existing behaviour. Not sure how the retries, possible it is 1 second?  https://github.com/hyperledger/besu/blob/main/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingPeerTask.java#L154

If we can later improve the peering performance with different timeouts that would be a good experiment to test out once the refactor is stable.

Instead of Thread.sleep, is there a reason not to use ethScheduler.scheduleFutureTask as before?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/DefaultPeerManager.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * This is a simple PeerManager implementation that can be used the default implementation in most
+ * situations
+ */
+public class DefaultPeerManager implements PeerManager {
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultPeerManager.class);
+
+  // use a synchronized map to ensure the map is never modified by multiple threads at once
+  private final Map<PeerId, EthPeer> ethPeersByPeerId =
+      Collections.synchronizedMap(new HashMap<>());
```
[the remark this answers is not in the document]
Matilda-Clerke: Nice, never use that one before.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: The idea was to have a progressive back-off timer on retries. As discussed elsewhere, it would be better to stay consistent with the existing implemention (appears to be a 1 second delay), and make this sort of change later, so I'll change that shortly.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/peertask/DefaultPeerManagerTest.java
Matilda-Clerke: Good call, I'll add another test

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
siladu: > progressive back-off timer on retries

0 -> 20 seconds -> 5 seconds? Should it `{0, 5000, 20000};` instead?

but yeh, think we should avoid changing any functionality as much as possible

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: #Instead of Thread.sleep, is there a reason not to use ethScheduler.scheduleFutureTask as before?
Mainly because this particular method shouldn't be concerned with scheduling tasks. Its singular concern is to implement task execution with optional retries.

In addition, methods in the EthSchedule class seem to operate on either Runnables or EthTasks, which really doesn't combine well with the entirely different paradigm at play with PeerTasks and the PeerTaskExecutor. A better solution would be to ensure we have similar metrics set up in the PeerTaskExecutor.

siladu: > Mainly because this particular method shouldn't be concerned with scheduling tasks. 
If we use it, EthScheduler would be the one concerned with scheduling tasks.

I think we need to maintain the metrics as they are, that's quite a big loss in functionality otherwise. Would much rather we use EthScheduler since a number of maintainers are trying to centralise thread execution here so we can reuse the metrics and testing support in a consistent way across Besu.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+  public <T> CompletableFuture<PeerTaskExecutorResult<T>> executeAsync(final PeerTask<T> peerTask) {
+    return CompletableFuture.supplyAsync(() -> execute(peerTask));
+  }
+
+  public <T> PeerTaskExecutorResult<T> executeAgainstPeer(
+      final PeerTask<T> peerTask, final EthPeer peer) {
+    MessageData requestMessageData = peerTask.getRequestMessage();
+    PeerTaskExecutorResult<T> executorResult;
+    int triesRemaining =
+        peerTask.getPeerTaskBehaviors().contains(PeerTaskBehavior.RETRY_WITH_SAME_PEER) ? 3 : 1;
```
jframe, line 91: Bit unsure about this behaviour. If you specify RETRY_WITH_SAME_PEER and RETRY_WITH_OTHER_PEERS what would happen? It would select a peer and try with that peer up to 3 times and if that fails repeat the process up to 3 times. So it could potentially be trying the request up to 9 times?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+      return false;
+    }
+  }
+
+  private static boolean isPeerUnused(
+      final EthPeer ethPeer, final Collection<EthPeer> usedEthPeers) {
+    return !usedEthPeers.contains(ethPeer);
+  }
+
+  private static boolean isPeerHeightHighEnough(final EthPeer ethPeer, final long requiredHeight) {
```
jframe, line 150: This method doesn't need to be static

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+    return ethPeer.chainState().getEstimatedHeight() >= requiredHeight;
+  }
+
+  private static boolean isPeerProtocolSuitable(final EthPeer ethPeer, final String protocol) {
```
jframe, line 154: This method doesn't need to be static

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+    try {
+      // sleep for 1 second to match implemented wait between retries in AbstractRetryingPeerTask
+      Thread.sleep(1000);
+      return true;
+    } catch (InterruptedException e) {
+      return false;
+    }
+  }
+
+  private static boolean isPeerUnused(
```
jframe, line 145: This method doesn't need to be static

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutorResult.java
```diff
@@ -0,0 +1,35 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.eth.manager.peertask;
+
+import java.util.Optional;
+
+public class PeerTaskExecutorResult<T> {
```
jframe, line 19: This would be simpler as a record

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: Alright, I'll see if I can do a bit of rework to utilise EthScheduler to capture some metrics.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: Oh, it was using the retry countdown to select it. First retry, the countdown would be 2, so `WAIT_TIME_BEFORE_RETRY[triesRemaining]` would be 5000

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTask.java
```diff
@@ -0,0 +1,63 @@
...
+   * @throws InvalidPeerTaskResponseException if the response messageData is invalid
+   */
+  T parseResponse(MessageData messageData) throws InvalidPeerTaskResponseException;
+
+  /**
+   * Gets the Collection of behaviors this task is expected to exhibit in the PeetTaskExecutor
+   *
+   * @return the Collection of behaviors this task is expected to exhibit in the PeetTaskExecutor
+   */
+  Collection<PeerTaskBehavior> getPeerTaskBehaviors();
```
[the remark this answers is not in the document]
Matilda-Clerke: Yep true

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTask.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param <T> The type of the result of this PeerTask
+ */
+public interface PeerTask<T> {
+  /**
+   * Returns the SubProtocol used for this PeerTask
+   *
+   * @return the SubProtocol used for this PeerTask
+   */
+  String getSubProtocol();
```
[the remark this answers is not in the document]
Matilda-Clerke: Iirc I looked at using the SubProtocol type and decided against it. I can't remember why, so I'll take another look.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskBehavior.java
```diff
@@ -0,0 +1,20 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.eth.manager.peertask;
+
+public enum PeerTaskBehavior {
+  RETRY_WITH_SAME_PEER,
+  RETRY_WITH_OTHER_PEERS
```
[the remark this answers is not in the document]
Matilda-Clerke: Non-retry behaviour is just the absence of retry behaviours.

I wanted to leave this as PeerTaskBehavior to allow for other behaviors to be added later. They aren't strictly just for describing desired retry behavior, but there aren't currently any other behaviors.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: Yes, that's correct

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskBehavior.java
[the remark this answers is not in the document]
siladu: I think if the goal is for the code to be more readable, I would go with more specific/descriptive naming and let whoever may (or may not) add more behaviours in later worry about the appropriate name then.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskRequestSenderTest.java
Matilda-Clerke: Because of the nature of RequestSender, the actual test has to be performed on another thread than the test thread. The code below this line must be execute after RequestSender has submitted it's callback method to responseStream.then. Since it doesn't make sense to expose a signal just for testing, we can just add a small delay to ensure the code is executed in the required sequence. Obviously, for production code, this is a dirty hack and proper inter-thread signalling should be used.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,157 @@
...
+    this.protocolSpecSupplier = protocolSpecSupplier;
+    requestTimer =
+        metricsSystem.createLabelledTimer(
+            BesuMetricCategory.PEERS, "Peer Task Executor Request Time", "", "Task Class Name");
+  }
+
+  public <T> PeerTaskExecutorResult<T> execute(final PeerTask<T> peerTask) {
+    PeerTaskExecutorResult<T> executorResult;
+    int triesRemaining =
+        peerTask.getPeerTaskBehaviors().contains(PeerTaskBehavior.RETRY_WITH_OTHER_PEERS) ? 3 : 1;
```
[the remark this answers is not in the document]
Matilda-Clerke: Added constants for now. I think I'll hold off on making them configurable until I get a better idea of how/when/why different values are used.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutorResult.java
Matilda-Clerke: It means losing the nice Optional wrapping to guarantee the optional result is always set, but yeah in general it should be a record.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskBehavior.java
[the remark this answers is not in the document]
Matilda-Clerke: Alright, that makes sense

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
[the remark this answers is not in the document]
Matilda-Clerke: So, the idea here is that the calling code supplies a Predicate<EthPeer> to effectively tell the PeerSelector what kind of peer is needed. I suppose if we move the logic to PeerSelector we can just expose additional getPeer methods for different use cases.

siladu (commented):
Few more comments, some of which were discussed on the call.

Thanks for your descriptive commit messages, really help with the re-review!

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/DefaultPeerSelector.java
```diff
@@ -0,0 +1,102 @@
...
+        .filter(filter)
+        .max(Comparator.naturalOrder())
+        .orElseThrow(NoAvailablePeerException::new);
+  }
+
+  @Override
+  public EthPeer getPeer(
+      final Collection<EthPeer> usedEthPeers,
+      final long requiredPeerHeight,
+      final SubProtocol requiredSubProtocol)
```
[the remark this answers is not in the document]
Matilda-Clerke: I think exposing a method to supply a `Predicate<EthPeer>` in PeerTask can represent the requirements of the task, plus something similar in the executor to represent the requirements of the executor. Stay tuned and I'll update this soon.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,153 @@
...
+            "className");
+  }
+
+  public <T> PeerTaskExecutorResult<T> execute(final PeerTask<T> peerTask) {
+    PeerTaskExecutorResult<T> executorResult;
+    int triesRemaining =
+        peerTask.getPeerTaskBehaviors().contains(PeerTaskRetryBehavior.RETRY_WITH_OTHER_PEERS)
+            ? RETRIES_WITH_OTHER_PEER
+            : NO_RETRIES;
+    final Collection<EthPeer> usedEthPeers = new HashSet<>();
```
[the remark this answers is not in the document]
Matilda-Clerke: That list represents the peers used for this particular task execution. Other tasks executing on the same peer shouldn't have any affect on each other (except through the normal peer reputation feedback mechanism)

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,153 @@
...
+        && executorResult.responseCode() != PeerTaskExecutorResponseCode.SUCCESS
+        && executorResult.responseCode() != PeerTaskExecutorResponseCode.PEER_DISCONNECTED
+        && sleepBetweenRetries());
+
+    return executorResult;
+  }
+
+  public <T> CompletableFuture<PeerTaskExecutorResult<T>> executeAgainstPeerAsync(
+      final PeerTask<T> peerTask, final EthPeer peer) {
+    return CompletableFuture.supplyAsync(() -> executeAgainstPeer(peerTask, peer));
```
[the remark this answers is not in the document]
Matilda-Clerke: Yep, I'll be looking at use EthScheduler for this I think

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+/** Manages the execution of PeerTasks, respecting their PeerTaskRetryBehavior */
+public class PeerTaskExecutor {
+
+  public static final int RETRIES_WITH_SAME_PEER = 3;
+  public static final int RETRIES_WITH_OTHER_PEER = 3;
+  public static final int NO_RETRIES = 1;
```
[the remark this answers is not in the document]
Matilda-Clerke: Reworking the loop should be fine.

I'm not sure I see any benefit to putting the number of retries on the PeerTaskRetryBehavior though, especially considering it enumerates different retry behaviors, and shouldn't really provide implementation details of how those enumerations should be implemented.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTask.java
```diff
@@ -0,0 +1,64 @@
...
+   * @throws InvalidPeerTaskResponseException if the response messageData is invalid
+   */
+  T parseResponse(MessageData messageData) throws InvalidPeerTaskResponseException;
+
+  /**
+   * Gets the Collection of behaviors this task is expected to exhibit in the PeetTaskExecutor
+   *
+   * @return the Collection of behaviors this task is expected to exhibit in the PeetTaskExecutor
+   */
+  Collection<PeerTaskRetryBehavior> getPeerTaskBehaviors();
```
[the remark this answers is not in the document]
Matilda-Clerke: IntelliJ refactor must have missed this one. I'll update it

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: Fair enough, it's easy to remove it and bring it back later. What I'm finding is that the parallelisation seems to work well when done over a whole set of actions anyway.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: It could be private for our existing uses, but afaik, there are some places that the calling code definitely cares about which peer it's requesting data from (particularly around pivot blocks, I think)

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -465,6 +468,18 @@ public void setTrailingPeerRequirementsSupplier(
     this.trailingPeerRequirementsSupplier = tprSupplier;
   }
 
+  // Part of the PeerSelector interface, to be split apart later
+  @Override
+  public EthPeer getPeer(final Predicate<EthPeer> filter) {
+    return streamBestPeers().filter(filter).findFirst().orElseThrow(NoAvailablePeersException::new);
+  }
```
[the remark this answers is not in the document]
Matilda-Clerke: Ok, I'll update

siladu (commented):
After discussion and further review, it sounds like this code is bypassing some existing functionality in EthPeers. That might be fine if done intentionally, but I'd like to understand why it is fine to skip it.

For example, I think it is missing an equivalent to `EthPeer.hasAvailableRequestCapacity`.

Also `EthPeer.executePeerRequest` has some logic around actualMinBlockNumber and maintains a synchronized `Collection<PendingPeerRequest> pendingRequests` which also has associated metrics - is there an equivalent to these in the new system? Maybe they are superfluous?

Without a good understanding of the old system, in particular EthPeers, it is hard to understand from the new system what is being bypassed, and whether it is intentional or something that was missed. I will have to defer to you @Matilda-Clerke to point these kind of changes out, or maybe for @pinges to spot them.

Matilda-Clerke:
> I think it is missing an equivalent to EthPeer.hasAvailableRequestCapacity
I've added this check in getPeer, but it does mean we can't just reuse the bestPeerMatchingCriteria method as you suggested.

> EthPeer.executePeerRequest
There definitely can be an equivalent in the new system. Alternative question though: Is that metric actually useful to anyone? The only metric I can find is just monitoring the size of the List, i.e. monitoring how many PeerRequests are "in flight".
The logic around minimum chain height, I did have implemented in PeerTaskExecutor when selecting a peer, but when we moved to the PeerTask providing a Predicate to filter EthPeers, it wasn't included. I'll have to find a way to get the required information into the places utilising GetReceiptsFromPeerTask so it can be passed in when the task is created.

We definitely need as many eyes on these changes as we can get. I think all that's really missing at the moment is the in-flight requests metric mentioned above.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,144 @@
...
+            : NO_RETRIES;
+    final Collection<EthPeer> usedEthPeers = new HashSet<>();
+    do {
+      Optional<EthPeer> peer =
+          peerSelector.getPeer(
+              (candidatePeer) ->
+                  peerTask.getPeerRequirementFilter().test(candidatePeer)
+                      && !usedEthPeers.contains(candidatePeer));
+      if (peer.isEmpty()) {
+        executorResult =
```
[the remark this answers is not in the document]
Matilda-Clerke: Yeah I noticed that. Seems very weird to me that we'd just sit and wait for a peer, but only for 5 seconds. Imo, we should just let the task/pipeline fail so it can be retried later.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,144 @@
...
+            new PeerTaskExecutorResult<>(
```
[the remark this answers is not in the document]
Matilda-Clerke: No, disconnecting failed peers shouldn't be done here (or in AbstractRetryingSwitchingPeerTask). It's just not one of the responsibilities of a class which is managing the sending of a request message and parsing of the response.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,144 @@
...
+    int retriesRemaining =
+        peerTask.getPeerTaskRetryBehaviors().contains(PeerTaskRetryBehavior.RETRY_WITH_OTHER_PEERS)
+            ? RETRIES_WITH_OTHER_PEER
+            : NO_RETRIES;
+    final Collection<EthPeer> usedEthPeers = new HashSet<>();
+    do {
+      Optional<EthPeer> peer =
+          peerSelector.getPeer(
+              (candidatePeer) ->
+                  peerTask.getPeerRequirementFilter().test(candidatePeer)
```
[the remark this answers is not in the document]
Matilda-Clerke: Yes, all of a tasks requirements should be included in the getPeerRequirementFilter Predicate, and any additional requirements should be added here by the PeerTaskExecutor

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,144 @@
...
+        && executorResult.responseCode() != PeerTaskExecutorResponseCode.PEER_DISCONNECTED
+        && sleepBetweenRetries());
+
+    return executorResult;
+  }
+
+  private boolean sleepBetweenRetries() {
+    try {
+      // sleep for 1 second to match implemented wait between retries in AbstractRetryingPeerTask
+      Thread.sleep(1000);
```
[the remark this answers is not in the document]
Matilda-Clerke: It's possible. In practice, this is only called on a worker thread anyway

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,144 @@
...
+            : NO_RETRIES;
+    do {
+      try {
+        T result;
+        try (final OperationTimer.TimingContext ignored =
+            requestTimer.labels(peerTask.getClass().getSimpleName()).startTimer()) {
+          MessageData responseMessageData =
+              requestSender.sendRequest(peerTask.getSubProtocol(), requestMessageData, peer);
+
+          result = peerTask.parseResponse(responseMessageData);
```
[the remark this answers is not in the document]
Matilda-Clerke: Currently, it's not. Looks like it's typically just checking that the response Collection/Map/etc is empty, which should be easy to add to PeerTask. In the old code, all it's doing is resetting the retry count when a partial success is detected, so I'm not sure it's really necessary to handle.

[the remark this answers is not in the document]
Matilda-Clerke: Should be handled appropriately now

siladu:
> I think all that's really missing at the moment is the in-flight requests metric mentioned above.

Is an equivalent of this logic needed too?https://github.com/hyperledger/besu/blob/aac7c63b0866d78edccdabace956c32947172090/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java#L268-L275

Matilda-Clerke:
@siladu As that's a requirement of the individual task, I've handled in the EthPeer filter of each PeerTask. You can see an example in https://github.com/hyperledger/besu/pull/7638/files#diff-65888fb86ed73e694cc6171b8f99c9150241acb7edc94a575a0ea83b4b91849dR111-R115

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,201 @@
...
+      final PeerTaskRequestSender requestSender,
+      final MetricsSystem metricsSystem) {
+    this.peerSelector = peerSelector;
+    this.requestSender = requestSender;
+    requestTimer =
+        metricsSystem.createLabelledTimer(
+            BesuMetricCategory.PEERS,
+            "PeerTaskExecutor:RequestTime",
+            "Time taken to send a request and receive a response",
+            "className");
```
jframe, line 61: Using label `taskName` would be better and matches with how label used elsewhere

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,201 @@
...
+  public PeerTaskExecutor(
+      final PeerSelector peerSelector,
+      final PeerTaskRequestSender requestSender,
+      final MetricsSystem metricsSystem) {
+    this.peerSelector = peerSelector;
+    this.requestSender = requestSender;
+    requestTimer =
+        metricsSystem.createLabelledTimer(
+            BesuMetricCategory.PEERS,
+            "PeerTaskExecutor:RequestTime",
```
jframe, line 59: Think `request_time` would be better no need for the PeerTaskExecutor: prefix

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,201 @@
...
+  private final Map<String, AtomicInteger> inflightRequestCountByClassName;
+
+  public PeerTaskExecutor(
+      final PeerSelector peerSelector,
+      final PeerTaskRequestSender requestSender,
+      final MetricsSystem metricsSystem) {
+    this.peerSelector = peerSelector;
+    this.requestSender = requestSender;
+    requestTimer =
+        metricsSystem.createLabelledTimer(
```
jframe, line 57: Can we just these metric names to use snakecase so they are consistent with naming we are using elsewhere for metrics

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,201 @@
...
+            BesuMetricCategory.PEERS,
+            "PeerTaskExecutor:RequestTime",
+            "Time taken to send a request and receive a response",
+            "className");
+    partialSuccessCounter =
```
jframe, line 62: For counters the convention we are using generally and there are probably exceptions is to end with the total suffix with snakecase.

So I think this would be better as `partial_success_total`

metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusGauge.java
```diff
@@ -47,17 +47,20 @@ public PrometheusGauge(
...
+    validateLabelsCardinality(labelValues);
     if (observationsMap.putIfAbsent(List.of(labelValues), valueSupplier) != null) {
       final String labelValuesString = String.join(",", labelValues);
       throw new IllegalArgumentException(
           String.format("A gauge has already been created for label values %s", labelValuesString));
     }
   }
 
+  @Override
+  public boolean isLabelsObserved(final String... labelValues) {
```
jframe, line 59: Couldn't see where this new method is used? Is it in the other PR?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
```diff
@@ -0,0 +1,201 @@
...
+      try {
+        T result;
+        try (final OperationTimer.TimingContext ignored =
+            requestTimer.labels(taskClassName).startTimer()) {
+          inflightRequestCountForThisTaskClass.incrementAndGet();
+
+          MessageData responseMessageData =
+              requestSender.sendRequest(peerTask.getSubProtocol(), requestMessageData, peer);
+
+          result = peerTask.parseResponse(responseMessageData);
```
pinges, line 143: It might be a good idea to not do the parsing here. I don't think that would be hard to change. It would make the PeerTaskExecutor almost trivial (no peer handling, no retries). The tasks themselves would have to handle that, but I guess they have to handle the partial success anyways. 
Could we have an AbstractTask that does most of that work?
Maybe something we can discuss tomorrow?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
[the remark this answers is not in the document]
pinges: With regard to the 5 second wait for a peer, I think there are cases where you want to do that retry, e.g. if all the matching peers are busy. But I think that could probably be decided by the caller?

plugin-api/build.gradle
```diff
@@ -71,7 +71,7 @@ Calculated : ${currentHash}
 tasks.register('checkAPIChanges', FileStateChecker) {
   description = "Checks that the API for the Plugin-API project does not change without deliberate thought"
   files = sourceSets.main.allJava.files
-  knownHash = '4jVaj9yW88nHbX0KmTR3dPQRvj9x8Pvh5E9Ry7KRT6w='
+  knownHash = 'VN2JB2HPpEUDQaDvd7QcMkmmgedasVChfA8tnSf1GHU='
```
pinges, line 74: what has changed in the plugin API?

metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusGauge.java
Matilda-Clerke: It should be used in PeerTaskExecutor.executeAgainstPeer to add labels for tasks we haven't seen before

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskExecutor.java
Matilda-Clerke: The whole point of the executor is handle retries and peer switching for a task. The task classes themselves don't actively do anything, they're more like a collection of methods that do small specific actions or supply values, like generating the request message. Changing this would be a pretty substantial change at this stage, and creating an AbstractPeerTask class is starting to sound extremely similar to the old system we're trying to move away from.

plugin-api/build.gradle
Matilda-Clerke: Added a method to LabelledGauge to allow us to check whether a label is observed or not

siladu: Can we add a default method (returns a dummy false value?) to avoid the breaking change please?

Matilda-Clerke: Yep, good idea

jframe (approved):

pinges (approved):
LGTM

Issue #7629, opened 2024-09-17
`IBFT2` fails block validation with `BONSAI` DB


matthew1001:
This log was captured from the `acceptanceTestBftSoak` acceptance test. The test was recently updated to use a combination of `FOREST` and `BONSAI` DB nodes - see PR https://github.com/hyperledger/besu/pull/7496. I believe it passed successfully before the PR was merged, but it seems to be semi-reliably failing now.

```
 shouldBeStableDuringLongTest | miner3 | 2024-09-17 08:54:27.703+00:00 | EthScheduler-Services-5 (importBlock) | INFO  | MainnetBlockValidator | Invalid block 305 (0x6a23d1450d82a1cf5431db18ddc02f80384c80c7d2a3356eb23726929c4ba9ad): Optional[Unable to process block because parent world state 0x305cc70172d6f210434ae4c650d3ac02629eb2f126ea094f921aa9692fc972f1 is not available]
    shouldBeStableDuringLongTest | miner3 | 2024-09-17 08:54:27.703+00:00 | EthScheduler-Services-5 (importBlock) | INFO  | Pipeline | Unexpected exception in pipeline. Aborting.
    shouldBeStableDuringLongTest | miner3 | Throwable summary: org.hyperledger.besu.ethereum.eth.sync.tasks.exceptions.InvalidBlockException: Failed to import block: Invalid block 305 (0x6a23d1450d82a1cf5431db18ddc02f80384c80c7d2a3356eb23726929c4ba9ad)
    shouldBeStableDuringLongTest | miner3 |     at: org.hyperledger.besu.ethereum.eth.sync.tasks.exceptions.InvalidBlockException.fromInvalidBlock(InvalidBlockException.java:36)
    shouldBeStableDuringLongTest | miner3 | 2024-09-17 08:54:27.703+00:00 | EthScheduler-Services-5 (importBlock) | WARN  | PipelineChainDownloader | Invalid block detected (BREACH_OF_PROTOCOL). Disconnecting from sync target. Failed to import block: Invalid block 305 (0x6a23d1450d82a1cf5431db18ddc02f80384c80c7d2a3356eb23726929c4ba9ad)
[...]

PR #7631, opened 2024-09-17
Don't persist IBFT2 proposal blocks, just validate them

matthew1001:
## PR description
This PR fixes a bug with IBFT2 chains where blocks are persisted at block-proposal time instead of block-import time. With Bonsai DB this causes issues updating the flat DB because if a round change happens the original proposal isn't actually the block that's accepted for that round. Since a flat-DB entry already exists for the proposed (but not accepted) block the imported block isn't actually persisted to the DB leaving the world state corrupted.

I've manually tested this with both `BONSAI` and `FOREST` DBs  to check that state updates are still persisted correctly by doing the following in both DB modes:

- Start a node with a fresh DB
- Deploy a contract, invoke it a few times
- Stop the node
- Restart the node
- Check the sender's transaction count is correct
- Check the state of the smart contract is correct
- Re-invoke it a few times to check it continues incrementing from the previous state

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/7629


jframe (approved):

jframe:
Changes make sense. You're right we shouldn't have been persisting the block during the validation.

PR #7633, opened 2024-09-17
7311: Add feature toggle for enabling use of the peertask system where available

Matilda-Clerke:
## PR description
Adds a feature toggle to enable use of the peer task system. To be utilised in future PRs implementing usage of the peer task system


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -842,6 +843,12 @@ static class PrivacyOptionGroup {
       description = "Specifies the number of last blocks to cache  (default: ${DEFAULT-VALUE})")
   private final Integer numberOfblocksToCache = 0;
 
+  @Option(
+      names = {"--peertask-system-enabled"},
```
macfarla, line 847: should this be a --X hidden option since it's temporary?

Matilda-Clerke: I didn't know we prefixed with X to denote hidden/temporary. I'll add it now

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -842,6 +843,12 @@ static class PrivacyOptionGroup {
       description = "Specifies the number of last blocks to cache  (default: ${DEFAULT-VALUE})")
   private final Integer numberOfblocksToCache = 0;
 
+  @Option(
+      names = {"--Xpeertask-system-enabled"},
```
macfarla, line 847: hidden = true,

macfarla: hidden means it won't show up in help besu -h

macfarla (approved):

besu/src/main/java/org/hyperledger/besu/cli/options/stable/P2PTLSConfigOptions.java
```diff
@@ -103,6 +104,13 @@ public class P2PTLSConfigOptions {
           "Whether to send a SNI header in the TLS ClientHello message (default: ${DEFAULT-VALUE})")
   private final Boolean p2pTlsClientHelloSniHeaderEnabled = false;
 
+  @Option(
```
jframe, line 107: This is in the wrong place, this option doesn't have anything to do with TLS. Better place for this would be `SynchronizerOptions`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskFeatureToggle.java
```diff
@@ -0,0 +1,59 @@
...
+package org.hyperledger.besu.ethereum.eth.manager.peertask;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Temporary class to allow easy access to the PeerTask feature toggle. This class can be removed
+ * once we've properly tested the PeerTask system and want to enable it permanently
+ */
+public class PeerTaskFeatureToggle {
```
jframe, line 24: Don't think this class is needed, this value could go into the appropriate domain object and be used from there e.g.  in `SynchronizerConfiguration`

jframe (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/PeerTaskFeatureToggle.java
Matilda-Clerke: I wanted to avoid passing around yet another configuration object just for a temporary feature toggle

besu/src/main/java/org/hyperledger/besu/cli/options/stable/P2PTLSConfigOptions.java
Matilda-Clerke: Sure, I'll move it over

jframe (approved):
LGTM

siladu (approved):
Toggle code looks good.

It would be good to get a rough idea about how it will be used in the full shebang https://github.com/hyperledger/besu/pull/7602

The POC touches a lot of existing code so could get messy (it's mostly wiring no doubt!)

Matilda-Clerke:
@siladu It's a little messy, but as you said it's just wiring. You can see how it'll be used in https://github.com/hyperledger/besu/pull/7638 (still some bugs to work out before that's ready for review)

PR #7640, opened 2024-09-19
Default value for min peers before starting sync from 5 to 1

macfarla:
## PR description
Change default value for minimum number of peers required before starting sync (only used by non-PoS networks) from 5 to 1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Additions and Improvements
 - Remove privacy test classes support [#7569](https://github.com/hyperledger/besu/pull/7569)
+- Set default value for sync-min-peers to 1 
 
 ### Bug fixes
 - Fix mounted data path directory permissions for besu user [#7575](https://github.com/hyperledger/besu/pull/7575)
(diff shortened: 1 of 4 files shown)


matthew1001:
I think I'm in favour of this change. The `enterprise-profile.toml` already sets `sync-min-peers` to 1 so it makes sense to just change the default.

matthew1001 (approved):
LGTM

macfarla:
> since it could make sense for private networks, for public ones that could increase the risk of syncing to the wrong fork, so could be possible to condition this on the basis if the network is public or private? probably looking at the presence or not of a custom genesis file?

ok I think to be on the safe side, we can leave this default as is. The "Private network" tutorials have been updated to use the enterprise profile so users won't strike this issue when following tutorials which is how it came up.

macfarla:
> > since it could make sense for private networks, for public ones that could increase the risk of syncing to the wrong fork, so could be possible to condition this on the basis if the network is public or private? probably looking at the presence or not of a custom genesis file?
> 
> ok I think to be on the safe side, we can leave this default as is. The "Private network" tutorials have been updated to use the enterprise profile so users won't strike this issue when following tutorials which is how it came up.

public PoA networks will still be using a custom genesis, right? so that wouldn't be a differentiator

PR #7641, opened 2024-09-19
Fix 7702 signature bound checks

daniellehrner:
## PR description

7702 signatures have different bound checks than regular transaction signatures. This PR create a new signature class for them with their own checks.

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GRADLE_OPTS: "-Xmx6g"
+  GRADLE_OPTS: "-Xmx7g"
   total-runners: 12
 
 jobs:
--- a/crypto/algorithms/src/main/java/org/hyperledger/besu/crypto/CodeDelegationSignature.java
+++ b/crypto/algorithms/src/main/java/org/hyperledger/besu/crypto/CodeDelegationSignature.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright contributors to Hyperledger Besu.
+ *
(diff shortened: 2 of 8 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GRADLE_OPTS: "-Xmx6g"
+  GRADLE_OPTS: "-Xmx7g"
```
fab-10, line 14: if you want to reduce the failure rate of ATs, then increase the number of total runners, maybe you can move 2 from reference-tests, that are faster, here

fab-10 (approved):

crypto/algorithms/src/main/java/org/hyperledger/besu/crypto/CodeDelegationSignature.java
```diff
@@ -0,0 +1,58 @@
...
+   * @param s the s part of the signature
+   * @param yParity the parity of the y coordinate of the public key
+   * @return the new CodeDelegationSignature
+   */
+  public static CodeDelegationSignature create(
+      final BigInteger r, final BigInteger s, final long yParity) {
+    checkNotNull(r);
+    checkNotNull(s);
+
+    if (r.compareTo(TWO_POW_256) >= 0) {
```
pinges, line 48: We could be a bit more precise here and check whether r and s are smaller than the prime of the curve, which is p = 2^256-2^32-977. But that still does not guarantee that (r, s) is a point on the curve.

pinges: Looking at the code this could be any curve that is being used here, so 2^256 might be better for an initial check?

PR #7646, opened 2024-09-19
Disable body validation for POS networks during sync

jframe:
## PR description
For POS networks this disables the body validation during sync import. This is unnecessary since we trust the chain since the CL has given us the tip of the chain to use. We still validate the transactions root, ommers root, receipts root, and withdrawals root during the download of the block bodies and transactions to verify we have the correct block bodies and receipts.

For non-POS networks body validation now skips validating transaction and receipts root during the sync as this is already done when downloading the block bodies and transaction receipts.

When body validation is used outside of syncing, the default original full validation mode is being used. This PR only affects the behaviour of the body validation during syncing.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/BlockValidator.java
```diff
@@ -83,22 +84,24 @@ BlockProcessingResult validateAndProcessBlock(
       final boolean shouldRecordBadBlock);
 
   /**
-   * Performs fast block validation with the given context, block, transaction receipts, requests,
-   * header validation mode, and ommer validation mode.
+   * Performs fast block validation appropriate for use during syncing skipping transaction receipt
+   * roots and receipts roots are these are done during the download of the blocks.
```
pinges, line 88:
```suggestion
   * roots and receipts roots as these are done during the download of the blocks.
```

pinges (commented):
would be nice to have a couple of tests for BodyValidationMode.NONE and BodyValidationMode.LIGHT
Otherwise good!

pinges (approved):
LGTM

PR #7654, opened 2024-09-20
Docker: Only switch user if the current user is root

skylenet:
One might be running the container without root (e.g. `docker run --user youruser ... `), if that's the case you'll get an error like this when running `su`:

```
Password: su: Authentication failure
```


besu/src/main/scripts/besu-entry.sh
```diff
@@ -45,5 +45,10 @@ done
 # Construct the command as a single string
 COMMAND="/opt/besu/bin/besu $@"
 
-# Switch to the besu user and execute the command
-exec su -s /bin/bash $BESU_USER_NAME -c "$COMMAND"
+# Check if current user is root
```
fab-10, line 48: Make sense, and probably if not root we can skip the all the part that fixes path permission, so I suggest to move this check at the very beginning of the script and if root just exec Besu, otherwise run the rest of the script, wdyt?

FYI @pullurib

pullurib: Yes, I agree, a check at very beginning of the script and if not root, just exec'ing Besu should be good.

skylenet: I've updated the PR and moved the user check to the beginning

fab-10 (approved):

PR #7662, opened 2024-09-23
Force besu to stop on plugin initialization errors

Gabriel-Trintinalia:
### PR Description

This PR changes the initialization behaviour of plugins in the `BesuPluginContextImpl` class by introducing a `pluginContinueOnError` flag. By default, Besu will not allow plugins to fail during registration and other startup lifecycle stages. 

### Changes Made

**pluginContinueOnError Flag:**

- Introduced the `--plugin-continue-on-error` CLI option (default to false).
- Added a `pluginContinueOnError` flag to the `PluginConfiguration` class.

**Lifecycle Methods:**

- Updated the `registerPlugins`, `beforeExternalServices`, `startPlugins`, and `afterExternalServicesMainLoop` methods to check the `pluginContinueOnError` flag. If a plugin fails and the flag is set to true, Besu will log the error and continue running. If the flag is set to false, Besu will throw a `RuntimeException` and halt.

### Usage Example

```sh
besu --plugin-continue-on-error=true
```

In this example, if any plugin fails during registration, before external services, startup, or after external services post main loop, Besu will log the error and continue running. If the `--plugin-continue-on-error` option is set to `false` (default), Besu will throw a `RuntimeException` and halt gracefully, ensuring that plugins cannot fail registration.


besu/src/main/java/org/hyperledger/besu/cli/options/stable/PluginsConfigurationOptions.java
```diff
@@ -44,9 +45,18 @@ public class PluginsConfigurationOptions implements CLIOptions<PluginConfigurati
...
       converter = PluginInfoConverter.class,
-      arity = "1..*")
+      arity = "1")
   private List<PluginInfo> plugins;
 
+  @CommandLine.Option(
+      names = {DEFAULT_HALT_ON_PLUGIN_ERROR_OPTION_NAME},
+      description =
+          "Prevent Besu from starting if any plugins fail to initialize correctly (default: ${DEFAULT-VALUE})",
+      hidden = true,
```
fab-10, line 55: probably better to make it visible

fab-10 (commented):
LGTM, just add a CHANGELOG entry and `doc-change-required` label.

Then I am thinking, that even if it will be a breaking changes, to set the default to `true`, at least to me it seems the normal behavior to fail in case a plugin has an issue. WDYT?

Gabriel-Trintinalia:
> LGTM, just add a CHANGELOG entry and `doc-change-required` label.
> 
> Then I am thinking, that even if it will be a breaking changes, to set the default to `true`, at least to me it seems the normal behavior to fail in case a plugin has an issue. WDYT?

I agree with stopping the startup when the plugin fails by default. That is the right behaviour in the majority of setups.

besu/src/main/java/org/hyperledger/besu/cli/DefaultCommandValues.java
```diff
@@ -129,7 +129,7 @@ public interface DefaultCommandValues {
   String DEFAULT_PLUGINS_OPTION_NAME = "--plugins";
 
   /** The constant DEFAULT_HALT_ON_PLUGIN_ERROR_OPTION_NAME. */
-  String DEFAULT_HALT_ON_PLUGIN_ERROR_OPTION_NAME = "--halt-on-plugin-error";
+  String DEFAULT_HALT_ON_PLUGIN_ERROR_OPTION_NAME = "--plugin-continue-on-error";
```
fab-10, line 132: need to rename also the constant name and javadoc

fab-10 (commented):
LGTM, just a last rename

fab-10 (approved):

PR #7665, opened 2024-09-23
Make inInSynch accessible to plugins

lu-pinto:


plugin-api/src/main/java/org/hyperledger/besu/plugin/services/sync/SynchronizationService.java
```diff
@@ -57,6 +57,8 @@ public interface SynchronizationService extends BesuService {
    */
   boolean isInitialSyncPhaseDone();
 
+  boolean isInSync(long tolerance);
```
fab-10, line 60: missing javadoc

fab-10 (commented):
Looks good, have you also reviewed what is already provided by `BesuEvents` that fire event on sync status change?

please add a CHANGELOG entry

lu-pinto:
> Looks good, have you also reviewed what is already provided by `BesuEvents` that fire event on sync status change?
> 
> please add a CHANGELOG entry

I hadn't thanks for that! I think I'll drop the PR then.
I was thinking, and given that this will fulfill RPC calls that needs an immediate reply, that for events there might be a case where an event was not fired since the client startup and so we will not be able to fulfill the RPC request. Though in the case of syncing we always start as out of sync and so that can be the returned default, unless a synced event occurs.

fab-10:
I think you can assume not in sync when the plugin starts and then wait for the in sync event.
Please note that if you need to specify a custom `tollerance` then `BesuEvents` does not support it.

fab-10:
> Isn't the `tolerance` not the same as doing in this case:
> 
> ```
> syncStatus.getHighestBlock() - syncStatus.getCurrentBlock() <= tolerance
> ```

yes, you are right

lu-pinto:
Closing as there's already a way to provide this info to plugins

PR #7673, opened 2024-09-24
Interrupt pending transaction processing on block creation timeout

fab-10:
## PR description

This PR continues on the track of making the block production more robust, in case some txs takes too much time to process, specifically it introduce a new feature, that forces the interruption of the tx that is still processing after the block creation timeout occurred. 

To trigger the interruption, a special operation tracer is used, that wraps and delegate all the calls to the original operator tracer, but in case the thread that is processing the tx is interrupted, just stop the processing of the tx.

To avoid penalizing perfectly valid and timely txs, a grace time, equals to the block creation max time, is given to the tx that was processing when a block creation timeout occurs, before forcing the interrupt, so that we correctly penalize only txs that are taking more that time.


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -422,14 +466,16 @@ private TransactionSelectionResult handleTransactionNotSelected(
     final var pendingTransaction = evaluationContext.getPendingTransaction();
 
     // check if this tx took too much to evaluate, and in case it was invalid remove it from the
-    // pool, otherwise penalize it.
+    // pool, otherwise penalize it. Not synchronized since there is no state change here.
     final TransactionSelectionResult actualResult =
```
Filter94, line 470: This becomes quite complex to read. Can we put this into a method with a clear name?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -422,14 +466,16 @@ private TransactionSelectionResult handleTransactionNotSelected(
...
         isTimeout.get()
             ? transactionTookTooLong(evaluationContext, selectionResult)
                 ? selectionResult.discard()
                     ? INVALID_TX_EVALUATION_TOO_LONG
                     : TX_EVALUATION_TOO_LONG
-                : BLOCK_SELECTION_TIMEOUT
+                : selectionResult.discard()
+                    ? BLOCK_SELECTION_TIMEOUT_INVALID_TX
```
Filter94, line 477: Isn't this supposed to be a dead branch? On line 473 we evaluated that `selectionResult.discard()` is false and only in this case we can go to evaluation on line 476 or am I misinterpreting this?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/InterruptibleOperationTracer.java
```diff
@@ -0,0 +1,143 @@
...
+
+  @Override
+  public void traceContextExit(final MessageFrame frame) {
+    checkInterrupt();
+    delegate.traceContextExit(frame);
+  }
+
+  private void checkInterrupt() {
+    if (Thread.interrupted()) {
+      throw new RuntimeException(new InterruptedException("Transaction execution interrupted"));
```
Filter94, line 140: What's the reason to wrap it into a RuntimeException?

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
```diff
@@ -1081,27 +1100,28 @@ public void subsetOfInvalidPendingTransactionsIncludedWhenTxSelectionMaxTimeIsOv
...
+          final boolean processingTooLate,
+          final boolean postProcessingTooLate) {
 
     internalBlockSelectionTimeoutSimulationInvalidTxs(
         isPoa,
         preProcessingTooLate,
         processingTooLate,
         postProcessingTooLate,
         900,
-        INVALID_TX_EVALUATION_TOO_LONG,
```
Filter94, line 1103: I think `INVALID_TX_EVALUATION_TOO_LONG` becomes missing from the tests. In which scenario we can receive this error?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
Filter94: I figured it out. I misinterpreted the indentation

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
fab-10: this is a very edge case, that could happen when the tx is invalid and complete the processing while at the same time the interrupt occurs, and it is not possible to cover without first making this test deterministic.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/InterruptibleOperationTracer.java
```diff
@@ -0,0 +1,143 @@
...
+import org.hyperledger.besu.plugin.data.ProcessableBlockHeader;
+import org.hyperledger.besu.plugin.services.tracer.BlockAwareOperationTracer;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import org.apache.tuweni.bytes.Bytes;
+
+public class InterruptibleOperationTracer implements BlockAwareOperationTracer {
```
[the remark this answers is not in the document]
fab-10: I only check for interruption in the methods that are called during the execution of opcodes, since is there where the tx could take too much time, other methods should be more or less constant in time.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
fab-10: yes will rework to make it easier to read

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
Filter94: From my examination of the code with the same config used to limit block building and transaction execution, I believe it can't happen at all. But it might make sense to separate these two in the future

testutil/src/main/resources/log4j2-test.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Configuration status="WARN">
   <Properties>
-    <Property name="root.log.level">INFO</Property>
+    <Property name="root.log.level">DEBUG</Property>
```
jflo, line 4: You sure about this one?

jflo (approved):
Might wanna double check running at DEBUG logging to make sure you really want all tests to be impacted.

testutil/src/main/resources/log4j2-test.xml
fab-10: no, but could be useful to enable more logging for the unit tests in another PR

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
fab-10: I have rework it a bit, and added javadoc, hope it is clearer now

PR #7677, opened 2024-09-25
[#1273] Expose revert reason field through GraphQL

### LINKED ISSUE #1273: Expose the revert reason in the GraphQL receipt
### Description

Revert reasons can be exposed via the JSON RPC interface according to https://besu.hyperledger.org/en/stable/HowTo/Send-Transactions/Revert-Reason/

But they are not exposed int he GraphQL interface

### Acceptance Criteria
* Expose the revert reason in the schema when Besu is started with the parameter.

pullurib:
## PR description
Expose revert reason through GraphQL queries 
- When Besu is configured to enable revert reasons, successful transactions would have null for revert reason and failed transactions would have it populated.
- When Besu is configured to not enable revert reasons, all transactions would have a null for revert reason. 

## Fixed Issue(s)
Fixes #1273


shemnon:
Can you get the official spec updated?  Kind of like how it was done in this PR? https://github.com/ethereum/execution-apis/pull/471/files

jflo (approved):

PR #7682, opened 2024-09-26
Simplifying BesuCommand step 1

jflo:
- pulls out P2P settings into its own options, and produces a record for downstream users to use, agnostic of picocli. this will eventually be moved alltogether into the relevant subsystem module.
- same for EngineRPC, since it re-uses some of those settings.
- moves DataStorageOptions up from the stable package since it also contains unstable options.
- puts P2PTLSOptions in unstable, since it is.
- overloads the JsonRpcConfiguration creation method to allow for no-arg creation.
- intentionally leaves in some TODOs because this is the first of many steps.


besu/src/main/java/org/hyperledger/besu/components/BesuCommandModule.java
```diff
@@ -53,7 +56,7 @@ BesuCommand provideBesuCommand(final @Named("besuCommandLogger") Logger commandL
             new BesuPluginContextImpl(),
             System.getenv(),
             commandLogger);
-    besuCommand.toCommandLine();
+    // besuCommand.toCommandLine();
```
fab-10, line 59: remove

fab-10 (commented):
look good just one commented line to remove.

For the future, this PR could have been split on a list of smaller PR according to the description

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
fab-10, line 1: since a lot of code has been moved out of `BesuCommand`, should also be possible to move some code out of `BesuCommandTest`

besu/src/test/java/org/hyperledger/besu/cli/CascadingDefaultProviderTest.java
```diff
@@ -105,7 +105,8 @@ public void overrideDefaultValuesIfKeyIsPresentInConfigFile(final @TempDir File
 
     final MetricsConfiguration metricsConfiguration =
         MetricsConfiguration.builder().enabled(false).host("8.6.7.5").port(309).build();
-
+    // push-host is defaulting to p2p-host, instead of localhost.
```
macfarla, line 108: should this be a TODO

macfarla (approved):

PR #7691, opened 2024-09-30
Fix: rpc gas price option should not incoditionally override smaller …

fab-10:
…tx gas limit


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -230,16 +230,23 @@ public Optional<TransactionSimulatorResult> processWithWorldUpdater(
...
-    long gasLimit =
+    final long txGasLimit =
         callParams.getGasLimit() >= 0
             ? callParams.getGasLimit()
             : blockHeaderToProcess.getGasLimit();
+
+    final long simulationGasLimit;
     if (rpcGasCap > 0) {
-      gasLimit = rpcGasCap;
+      simulationGasLimit = Math.min(txGasLimit, rpcGasCap);
```
ahamlat, line 240: There is a real need to make RPC gas limit above block size limit, especially on eth_call. I wonder if we cannot make eth_estimateGas using a different variante of the transaction simulator

fab-10: using different transaction simulators is doable, let me try to implement that way

fab-10: @ahamlat I have implemented your suggestion and now `eth_call` behaves differently of other call in regards of how `rpc-gas-cap` is applied

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -380,4 +374,87 @@ public Optional<Boolean> doesAddressExist(
...
+        final long simulationGasLimit = Math.min(txGasCap, rpcGasCap);
+        LOG.trace(
+            "Gas limit capped at {} for transaction simulation, tx gas limit={}, block gas limit={}, RPC gas cap={}",
+            simulationGasLimit,
+            txGasLimit,
+            blockGasLimit,
+            rpcGasCap);
+        return simulationGasLimit;
+      }
+      return txGasLimit;
```
Gabriel-Trintinalia, line 395: I'm trying to understand the logic here. If `rpcGasCap == 0` we return txGasLimit and do not consider `txGasCap`? If so, can we move the` final long txGasCap = txGasLimit >= 0 ? txGasLimit : blockGasLimit` to be inside the if?

Also, could you please add a summary to the PR description.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -380,4 +374,87 @@ public Optional<Boolean> doesAddressExist(
...
+
+    public Builder rpcGasCap(final long rpcGasCap, final RpcGasCapMode mode) {
+      this.rpcGasCap = rpcGasCap;
+      this.rpcGasCapMode = mode;
+      return this;
+    }
+
+    public TransactionSimulator build() {
+      final SimulationGasLimitCalculator simulationGasLimitCalculator;
+      if (rpcGasCap > 0 && rpcGasCapMode.equals(RpcGasCapMode.UNBOUNDED)) {
```
Filter94, line 446: a `long` + enum encodes same amount of information as `Long` or Optional<Long> null or `Optional.empty()` may mean no bound. Introduction of `RpcGasCapMode` is verbose, but maybe it makes it more obvious

Filter94: However possibility to set it to (0, BOUNDED) makes it errorprone. Optional or a nullable Long is less errorprone from this perspective

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -380,4 +374,87 @@ public Optional<Boolean> doesAddressExist(
 
     return Optional.of(worldState.get(address) != null);
   }
+
+  @FunctionalInterface
+  public interface SimulationGasLimitCalculator {
+    long calculate(final long txGasLimit, final long blockGasLimit);
+
+    static long bounded(final long txGasLimit, final long blockGasLimit, final long rpcGasCap) {
```
Filter94, line 382: I think if we replace 0 values with `Long.MAX_VALUE` the logic would be as simple as taking a min over 3 values. Would it be simpler to read?

PR #7692, opened 2024-09-30
dont' activate flight recorder dumps by default, RIP my ssd

jflo:
Removes enabling .jfr capture on every process started by acceptanceTest targets. This can still be added via any other target, or even explicitly via command line per-run.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -690,7 +690,6 @@ startScripts {
     "-XX:G1ConcRefinementThreads=2",
     "-XX:G1HeapWastePercent=15",
     "-XX:MaxGCPauseMillis=100",
-    "-XX:StartFlightRecording,dumponexit=true,settings=default.jfc",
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -690,7 +690,6 @@ startScripts {
     "-XX:G1ConcRefinementThreads=2",
     "-XX:G1HeapWastePercent=15",
     "-XX:MaxGCPauseMillis=100",
-    "-XX:StartFlightRecording,dumponexit=true,settings=default.jfc",
```
ahamlat, line 693: I recommend to keep it but remove only dumponexit=true, to not dump the file when the process stops or crashes.

ahamlat (changes requested):

pinges (approved):
LGTM

macfarla (approved):

ahamlat (approved):

PR #7703, opened 2024-10-01
Align gas cap calculation for transaction simulation to Geth approach

fab-10:
## PR description

A recent PR (https://github.com/hyperledger/besu/pull/6978) made the optional `rpc-gas-cap` config to always override the gas cap used for tx simulation, when specified, but while this could be good for `eth_call`, it does not work well with calls like `eth_estimateGas` when the gas cap could be based on what use provided or the sender balance, and it also diverging from what Geth does.

This PR align the gas cap calculation used during tx simulation, to what Geth does, with only a little difference, the algorithm is the following:
- if the user specific the gas limit, then use it unless it is bigger than the `rpc-gas-cap`, in which case we return the latter
- if there is not user specified gas limit, then use `rpc-gas-cap` if configured, otherwise use the block gas limit^.

^ in this case Geth use `MaxUint64 / 2`

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
@@ -230,16 +230,9 @@ public class TransactionSimulator {
     final Account sender = updater.get(senderAddress);
     final long nonce = sender != null ? sender.getNonce() : 0L;
 
-    long gasLimit =
-        callParams.getGasLimit() >= 0
-            ? callParams.getGasLimit()
-            : blockHeaderToProcess.getGasLimit();
-    if (rpcGasCap > 0) {
-      gasLimit = rpcGasCap;
-      LOG.trace(
-          "Gas limit capped at {} for transaction simulation due to provided RPC gas cap.",
-          rpcGasCap);
-    }
+    final long simulationGasCap =
+        calculateSimulationGasCap(callParams.getGasLimit(), blockHeaderToProcess.getGasLimit());
+
     final Wei value = callParams.getValue() != null ? callParams.getValue() : Wei.ZERO;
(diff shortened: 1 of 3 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -291,6 +283,38 @@ public Optional<TransactionSimulatorResult> processWithWorldUpdater(
     return Optional.of(new TransactionSimulatorResult(transaction, result));
   }
 
+  private long calculateSimulationGasCap(
+      final CallParameter callParams, final BlockHeader blockHeaderToProcess) {
```
Filter94, line 287: Sorry to come back to you on this, but rather than passing whole `CallParameter` and `BlockHeader`, we can pass gas limits only, it would make it even more focused

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -291,6 +283,38 @@ public Optional<TransactionSimulatorResult> processWithWorldUpdater(
...
+    final long simulationGasCap;
+
+    // when not set gas limit is -1
+    if (callParams.getGasLimit() >= 0) {
+      if (rpcGasCap > 0 && callParams.getGasLimit() > rpcGasCap) {
```
ahamlat: I wonder if it is more clear to use  `rpcGasCap > 0` as the first test.
Basically, we start by asking if the user have a configured rpcGasCap, and depending on this user choice, we can check other options.

ahamlat (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
fab-10: I understand it could be subjective, I prefer that user provided gas limit drives the choice.

PR #7711, opened 2024-10-02
Changes to release workflow

cdivitotawela:
## PR description
- Workflow environment variable changed to RELEASE_VERSION instead of RELEASE_NAME to make it more meaningful

- Workflows draft-release and docker-promote previously trigger workflow container-verify. This can result in trigger is successful but actual verification workflow fails. This will not be reflected in the draft-release and docker-promote workflows. Container verify code is embedded in the draft-release and docker-promote workflows to avoid this confusion


.github/workflows/docker-promote.yml
```diff
@@ -58,24 +58,49 @@ jobs:
...
+            runner: ubuntu-22.04
+          - tag: latest
+            platform: ''
+            runner: ubuntu-22.04
+          - tag: ${{ needs.validate.outputs.release_version }}-arm64
+            platform: ''
+            runner: besu-arm64
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
```
siladu, line 90: Not directly relevant to your changes but wondering if we really need to checkout the whole repo here? 
Is it just to get the script? I assume it pulls from our dockerhub so doesn't need the source to verify. Wonder if we can get hold of just the script instead.

.github/workflows/docker-promote.yml
```diff
@@ -58,24 +58,49 @@ jobs:
...
+          - tag: ${{ needs.validate.outputs.release_version }}
+            platform: ''
+            runner: ubuntu-22.04
+          - tag: ${{ needs.validate.outputs.release_version }}-amd64
+            platform: 'linux/amd64'
+            runner: ubuntu-22.04
+          - tag: latest
+            platform: ''
+            runner: ubuntu-22.04
+          - tag: ${{ needs.validate.outputs.release_version }}-arm64
```
siladu, line 85: Haven't we already checked the release_version variants in draft-release?

Think we should check latest, latest-amd64, latest-arm64 here

siladu (approved):
PR looks sound, my comments are beyond what the PR was intended for - up to you if you want to merge this one first.

.github/workflows/docker-promote.yml
cdivitotawela: Parse checkout is possible. Wonder it is slower than the full checkout. Let me check this and switch to parse checkout only the file.

cdivitotawela: Looks like parse checkout is faster. Let me update this.

.github/workflows/docker-promote.yml
cdivitotawela: Good point. I will add latest-amd64 and latest-arm64

.github/workflows/docker-promote.yml
cdivitotawela: Updated

.github/workflows/BesuContainerVerify.sh
```diff
@@ -57,7 +57,7 @@ else
 fi
 
 # For the latest tag check the version match
-if [[ ${TAG} == "latest" && ${CHECK_LATEST} == "true" ]]
+if [[ ${TAG} =~ "latest" && ${CHECK_LATEST} == "true" ]]
```
siladu, line 60: nit, could tighten up the regex slightly: 
```suggestion
if [[ ${TAG} =~ "^latest" && ${CHECK_LATEST} == "true" ]]
```

cdivitotawela: I tried that did not work. Let me check again whether I missed something

cdivitotawela: Found it. I should not use quotes around the pattern. Will fix it.

cdivitotawela: Updated

PR #7717, opened 2024-10-02
Warm up to address at tx start if account is delegated, restrict auth nonce

### LINKED ISSUE #7706: Devnet-3: Besu forked off at block 110500
Initial review shows a wrong gas calculation, Besu is charging 2500 gas more for a `CALL` than the other clients. It seems the problem is:

The tx causing it is `0x01911cedc5083ed9342846193bc1fbf187616f2366df7ba8606e49f39f2b7b93`

Analysis
* Tx is sent to account `3093` 
* `3093` is an account which has delegated code
* At some point the contract make a `CALL` to itself, 2500 gas is charged too much
* Whenever an OP code needs to resolve an account with delegated code, we either charge 2600 (cold read) or 100 (warm read) for the resolution. The difference between the 2 is exactly 2500.

Conclusion:
Other clients seem to warm up the `to` address when the account has delegated code, even though no other additional gas costs are charged for this specific resolution at the beginning of the transaction.

Solution:
Besu needs to warm the `to` address in this case as well.

daniellehrner:
## PR description

Fixes a missing gas calculation when the `to` address of a tx has delegated code. The address needs to be warmed up at the beginning of the tx.

The auth of the nonce was before resticted to be less or equal than 2**64-1, but now must be less than 2**64-1 to avoid overflows. That is the same restriction that tx nonces already have.

## Fixed Issue(s)
fixes #7706 


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -439,11 +438,17 @@ public TransactionProcessingResult processTransaction(
...
                 .code(code)
+                .accessListWarmAddresses(warmAddressList)
                 .build();
       } else {
         @SuppressWarnings("OptionalGetWithoutIsPresent") // isContractCall tests isPresent
         final Address to = transaction.getTo().get();
         final Optional<Account> maybeContract = Optional.ofNullable(evmWorldUpdater.get(to));
+
+        if (maybeContract.isPresent() && maybeContract.get().hasDelegatedCode()) {
+          warmAddressList.add(to);
```
macfarla, line 449: would be great to have a test for this

PR #7718, opened 2024-10-02
7536 use head for snap sync

### LINKED ISSUE #7536: Use head block instead of safe block for snap sync
Change snap sync to use head block instead of safe block for the snap sync.

Using the head allows for larger download windows between two pivot blocks. With larger windows, we will have fewer pivot blocks and therefore more time for the heal. This could allow us to finish faster. If we download the storage of an account and then change the pivot block, we need to request the account again to get the new storage root of the account in the block, as this can cause a lot of storage requests to fail.

Acceptance Criteria
* Verify that this has the improvement on healing and syncing that we are expecting

kingnhcomcast:
## Fixed Issue(s)

fixes #7536

Default is sync to head, added config option 'Xsnapsync-to-head-enabled' which can be set to false to revert to older sync to safe block.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromBlock.java
```diff
@@ -0,0 +1,166 @@
...
+                            }))
+            .orElse(0L),
+        localChainHeight);
+  }
+
+  @Override
+  public Optional<FastSyncState> selectNewPivotBlock() {
+    final Optional<ForkchoiceEvent> maybeForkchoice = forkchoiceStateSupplier.get();
+    if (maybeForkchoice.isPresent() && maybeForkchoice.get().hasValidSafeBlockHash()) {
+      // return Optional.of(selectBlockAsPivot(maybeForkchoice.get().getSafeBlockHash()));
```
garyschulte, line 128: nit: remove old commented code

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromBlock.java
```diff
@@ -0,0 +1,166 @@
...
+                              return null;
+                            }))
+            .orElse(0L),
+        localChainHeight);
+  }
+
+  @Override
+  public Optional<FastSyncState> selectNewPivotBlock() {
+    final Optional<ForkchoiceEvent> maybeForkchoice = forkchoiceStateSupplier.get();
+    if (maybeForkchoice.isPresent() && maybeForkchoice.get().hasValidSafeBlockHash()) {
```
garyschulte, line 127: hasValidSafeBlockHash() isn't necessary for the head block implementation.  I would suggest making getPivotHash return Optional<Hash>, and you can do the safe block check there and return Optional.empty() if it doesn't exist.

this is a very corner case, where we would be trying to sync a network very near genesis, before a block has been marked 'safe'.  The prior implementation of safe-only would have the same issue.  So this isn't a huge deal, but fixing it now might save a headache in the future for sync tests or something like that where we would be syncing a brand new chain.

garyschulte (changes requested):
LGTM, good tests.  One nit and one suggestion

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromBlock.java
kingnhcomcast: code removed

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromBlock.java
kingnhcomcast: code removed

garyschulte (approved):
🚢

PR #7720, opened 2024-10-03
Add support for `chainId` in `CallParameters`

fab-10:


Filter94:
I think we should add a testcase to acceptance tests for this not to ever miss this again

fab-10:
> I think we should add a testcase to acceptance tests for this not to ever miss this again

Added tests that verify that in case the chainId does not match, then an error is returned, then added the chainId in some successful tests. Worth to know that in `eth_estimateGas` the chainId was added by default if the tx type requires it, this feature was not present on Linea and will be added.

macfarla (approved):

PR #7723, opened 2024-10-04
Keep track of blobs that are part of multiple transactions

pinges:
## PR description
The same blob can be part of multiple transactions. This PR makes sure that these blobs can be returned by `eth_getBlobsV1` even when only one transaction containing that blob is still in the transaction pool.


fab-10 (commented):
Basically what you need is to count how many txs reference this blob, so it will be simpler and more performant to just keep a counter instead of a list

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -107,8 +108,8 @@ public class TransactionPool implements BlockAddedObserver {
   private final SaveRestoreManager saveRestoreManager = new SaveRestoreManager();
   private final Set<Address> localSenders = ConcurrentHashMap.newKeySet();
   private final EthScheduler.OrderedProcessor<BlockAddedEvent> blockAddedEventOrderedProcessor;
-  private final Map<VersionedHash, BlobsWithCommitments.BlobQuad> mapOfBlobsInTransactionPool =
-      new HashMap<>();
+  private final Map<VersionedHash, List<BlobsWithCommitments.BlobQuad>>
```
fab-10, line 111: This need to support concurrency, since notifications are sent async by potentially multiple threads

fab-10 (changes requested):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
siladu: Maybe switch to https://guava.dev/releases/3.0/api/docs/com/google/common/collect/Multimaps.html#synchronizedListMultimap(com.google.common.collect.ListMultimap) ?

pinges: The access to the map was synchronized, but using a map implementation that does the synchronization for me is probably a better idea!

pinges:
> Basically what you need is to count how many txs reference this blob, so it will be simpler and more performant to just keep a counter instead of a list

My first idea was to have a reference count, but this has the drawback that you potentially keep an additional blob in memory, while the list has the drawback of using a little more storage and it is a little slower.
I also like that there is an implementation of a multi threading safe map that contains a list as value.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -107,8 +110,10 @@ public class TransactionPool implements BlockAddedObserver {
   private final SaveRestoreManager saveRestoreManager = new SaveRestoreManager();
   private final Set<Address> localSenders = ConcurrentHashMap.newKeySet();
   private final EthScheduler.OrderedProcessor<BlockAddedEvent> blockAddedEventOrderedProcessor;
-  private final Map<VersionedHash, BlobsWithCommitments.BlobQuad> mapOfBlobsInTransactionPool =
-      new HashMap<>();
+  private final ListMultimap<VersionedHash, BlobsWithCommitments.BlobQuad>
+      mapOfBlobsInTransactionPool =
+          Multimaps.synchronizedListMultimap(
+              Multimaps.newListMultimap(new HashMap<>(), ArrayList::new));
```
pinges, line 116:
```suggestion
              Multimaps.newListMultimap(new HashMap<>(), () -> new ArrayList<>(1)));
```

fab-10 (approved):

Issue #6113, opened 2023-11-02
Nullpointer during snapsync "because taskElement is null"


matkt:
### Description

There is an issue when we receive data but the proof is invalid. We still try to retrieve the child tasks even though the stacktrie has not been modified due to the invalid proof. We should not only verify that we have retrieved data, but also check that the data is valid before trying to retrieve child request.

```
2023-11-01 18:39:01.848+00:00 | EthScheduler-Services-43 (importBlock) | INFO  | FastImportBlocksStep | Block import progress: 12158784 of 18477974 (65%)

2023-11-01 18:39:34.237+00:00 | EthScheduler-Services-13 (batchPersistLargeStorageData) | INFO  | Pipeline | Unexpected exception in pipeline. Aborting.

java.lang.NullPointerException: Cannot invoke "org.hyperledger.besu.ethereum.eth.sync.snapsync.StackTrie$TaskElement.proofs()" because "taskElement" is null

	at org.hyperledger.besu.ethereum.eth.sync.snapsync.request.StorageRangeDataRequest.getChildRequests(StorageRangeDataRequest.java:166)

	at org.hyperledger.besu.ethereum.eth.sync.snapsync.PersistDataStep.persist(PersistDataStep.java:51)

	at org.hyperledger.besu.ethereum.eth.sync.snapsync.SnapWorldStateDownloadProcess$Builder.lambda$build$13(SnapWorldStateDownloadProcess.java:314)

	at org.hyperledger.besu.services.pipeline.MapProcessor.processNextInput(MapProcessor.java:31)

	at org.hyperledger.besu.services.pipeline.ProcessingStage.run(ProcessingStage.java:38)

	at org.hyperledger.besu.services.pipeline.Pipeline.lambda$runWithErrorHandling$3(Pipeline.java:169)

	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)

	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)

	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)

	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)

	at java.base/java.lang.Thread.run(Thread.java:833)

[...]

PR #7724, opened 2024-10-04
Add null pointer check and testcase for snapsync tasks

pullurib:
## PR description
Added a null pointer check where a null pointer exception was thrown as described in the issue 
Added a testcase to verify the case is handled gracefully

taskElement shouldn't be null ideally as it's only added to trie after validating proofs and we get to this point after checking if data is available and isValidProof is true.  
- A race condition between calling addResponse(requestDataStep) and getChildRequests(persistDataStep) seems unlikely as they are sequential in pipeline and synchronized by setting isValidProof after adding taskElement and checked before accessing taskElement.  
- Clear() on the request also sets isValidProof to false before removing the element from Trie.

## Fixed Issue(s)
#6113


lu-pinto (approved):
LGTM

matkt:
is it something we already had during a sync ?

matkt:
For me, doing this is not a good idea because if we are searching for the children, it is typically because we intend to persist the parent. If there is an error of this type and we return empty, we will persist the request and never search for the child requests again, which will lead to an incomplete tree. I don’t think this is a serious issue because I believe this case doesn’t happen often, but I consider the fix to be invalid.

pullurib:
@matkt Based on my understanding, a null pointer failure should be handled similarly to an invalid proof by adding the task to the retry queue. I noticed that invalid proofs are managed in the addResponse function. Handling a null taskElement by adding it to the retry queue—much like adding it to the healing list—seems appropriate. Please let me know if this approach makes sense. I will look into this further and follow up with a fix.

matkt:
In your case, the response is marked as received because the proof is valid. If the request is therefore marked as valid, we will not retry it, and we will not fetch the subsequent requests to complete the range (see https://github.com/pullurib/besu/blob/9ea174bd1c2de13709f2e1168af10272b606269a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/request/StorageRangeDataRequest.java#L160 and https://github.com/pullurib/besu/blob/9ea174bd1c2de13709f2e1168af10272b606269a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/snapsync/CompleteTaskStep.java#L50).

In the case where the proof is invalid, it will be marked as false, and CompleteTaskStep will put it back into the pipeline.

PR #7729, opened 2024-10-04
Improve signature verification performance

2fehee:
* Efficient signature verification using tx in Tx pool
* This feature can enabled with a flag --tx-pool-tx-fast-verify-signature-enabled=true
* There was a performance improvement effect of about 30%
* No need to add transaction sender cache space

## PR description
In the proposed block (received block) verification step, when verifying the signature for the Tx contained in the received block, check whether the same Tx is already in the Tx-pool of the node to improve the efficiency of unnecessary signature verification work.

* Basically, Tx, both Local Tx and Remote Tx, undergo signature verification before being placed in the Tx pool, and only Txs that pass verification are placed in the Tx pool.
* At this time, the sender information, which is the signing subject of the Tx, is derived through the signature verification process for the Tx and this is stored in the Transaction object information.

* If the sender information exists in the Transaction object, no additional signature verification occurs because the signature verification has already been completed. However, if the sender information is null, the signature verification task is performed.

* When creating a block, no additional signature verification logic is performed because the node in the propose stage selects a Tx from its own Tx pool to create the block.

* However, other nodes that perform the proposed block (received block) verification step must perform the verification logic again because they cannot trust the received block.

  * Perform signature verification on Txs within received blocks

  * Perform EVM execution process for Txs within received blocks

* In the proposed block (received block) verification stage, the Txs in the received block are in the stage before the signature verification process is performed, so the sender information in the Transaction object is null.

* The Tx in the Tx pool of each node performing the proposed block (received block) verification step is a Tx that has already completed the signature verification process, so the sender information exists in the Transaction object.

* Therefore, if the Tx to be verified exists in the Tx pool of the node, filling in the sender information of the Tx in the Tx pool to the sender information in the Transaction object of the Tx to be verified will result in the verification effect without performing the signature verification task.

  *  Why this task is applicable (key features)

     * To check whether the Tx you want to verify exists in the Tx pool of the node, check whether the same Tx-hash value as the Tx-hash value of the Tx you want to verify exists in the Tx pool.

     * If (data value and signature value in Tx1) are the same as (data value and signature value in Tx2), the Tx-hash values ​​of the two Txs must be the same.

     * The Tx registered in the Tx pool has already had its signature verified, and this can be confirmed by checking whether the sender information is present in the Transaction object.

     * Since Txs for each node are synchronized through P2P communication in a short period of time, most of the Txs in the block to be verified are likely to exist in the Tx pool of each node (performance increases as the number of Txs exists).
       * If the Txs in the block to be verified do not exist in the Tx-pool of each node, the previous signature verification procedure is executed.

* Since we use the signature verification result information (sender value) through txpool, we do not need to add transaction sender cache space.
  * Therefore, Uses less memory than previous code.

* Key Option Conditions
  * --tx-pool-tx-fast-verify-signature-enabled=true
  * --tx-pool=sequenced
  * --min-gas-price=0
  * (The code applies to both IBFT and QBFT)

### Performance test result [--tx-pool-tx-fast-verify-signature-enabled=false]
* Performance Test Conditions and Environment
  * besu Version : 24.9-develop
  * Applying GraalVM : No
  * Consensus Algorithm : QBFT
  * Infrastructure Spec : AWS c5.xlarge (3.5Ghz, 4Core, 8GB), Disk (EBS)
  * Test Tool : JMeter (Distributed Testing Env, 1 Controller Node, 4 Worker nodes)
  * Smart Contract : ERC20
  * Number of EOA accounts : 8 (from EOA : 4, to EOA : 4)
    * As the number of accounts within a smart contract increases, performance may degrade 
  * Test time : 10 minutes
  * TPS Calculation
    * Total number of transactions successfully recorded in the distributed ledger / Total test time (seconds)
    * Total test time : (The final response time when all transactions were recorded in the distributed ledger and a response was returned) - (the time when the JMeter load started)
  * Test Result (TPS) : about 1730 TPS
    * Summary of Results
      * image
    * TPS
      * image
    * Response Time
      * image
    * Besu Log
      * image
 
### Performance test result [--tx-pool-tx-fast-verify-signature-enabled=true]
* Performance Test Conditions and Environment
  * besu Version : 24.9-develop
  * Applying GraalVM : No
  * Consensus Algorithm : QBFT
  * Infrastructure Spec : AWS c5.xlarge (3.5Ghz, 4Core, 8GB), Disk (EBS)
  * Test Tool : JMeter (Distributed Testing Env, 1 Controller Node, 4 Worker nodes)
  * Smart Contract : ERC20
  * Number of EOA accounts : 8 (from EOA : 4, to EOA : 4)
    * As the number of accounts within a smart contract increases, performance may degrade 
  * Test time : 10 minutes
  * TPS Calculation
    * Total number of transactions successfully recorded in the distributed ledger / Total test time (seconds)
    * Total test time : (The final response time when all transactions were recorded in the distributed ledger and a response was returned) - (the time when the JMeter load started)
  * Test Result (TPS) : about 2250 TPS
    * Summary of Results
      * image
    * TPS
      * image
    * Response Time
      * image
    * Besu Log
      * image


fab-10:
Hi and thanks for the contribution. I will review it in the next days.

FYI: @ahamlat you could be interested reviewing the performance side of this

ahamlat:
Thanks for the contribution @2fehee, I have few comments/questions.

> Since we use the signature verification result information (sender value) through txpool, we do not need to add transaction sender cache space.

The cache is a way to store the sender of each single transaction, that Besu had to recover the sender from signature. For example, we use it in [AbstractEngineNewPayload](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java#L243) to improve sender recovery on block execution for Proof Of Stake. The cache will store also all the senders of the transactions that are in the mempool as recovering the sender the first time in Besu will store it in the cache. So using the transactions from the transaction pool to get the sender is already covered by the cache. I wonder where the performance gain does come from as the cache already cover transaction pool transactions ? It would be interesting to have CPU profiling or Wall clock profiling for both cases.

> Since we use the signature verification result information (sender value) through txpool, we do not need to add transaction sender cache space.
> 
> Therefore, Uses less memory than previous code.


Especially for PoS use case, I don't think we should disable the sender cache in the Transaction class. It is pretty small and consumes less than 20 MiB, more information [here](https://github.com/hyperledger/besu/pull/6375) .

> If the sender information exists in the Transaction object, no additional signature verification occurs because the signature verification has already been completed. However, if the sender information is null, the signature verification task is performed.



In [MainnetTransactionValidator.validate](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java#L87) , there is more than just validating the sender of the transaction especially in PoS, I wonder if this is really safe, especially in this case.

2fehee:
> Hi and thanks for the contribution. I will review it in the next days.
> 
> FYI: @ahamlat you could be interested reviewing the performance side of this

2fehee:
> > Hi and thanks for the contribution. I will review it in the next days.
> > FYI: @ahamlat you could be interested reviewing the performance side of this

@fab-10 , @ahamlat 
Thanks for reviewing my code. 

I will check your comments/questions and reply to you

2fehee:
@ahamlat
Thanks for reviewing my code.

> The cache is a way to store the sender of each single transaction, that Besu had to recover the sender from signature. For example, we use it in [AbstractEngineNewPayload](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java#L243) to improve sender recovery on block execution for Proof Of Stake. The cache will store also all the senders of the transactions that are in the mempool as recovering the sender the first time in Besu will store it in the cache. So using the transactions from the transaction pool to get the sender is already covered by the cache. I wonder where the performance gain does come from as the cache already cover transaction pool transactions ? It would be interesting to have CPU profiling or Wall clock profiling for both cases.

First, I will explain the first comment you gave me among my PR sources.

My PR code excluded the senderCache area of ​​the Transaction.java file from [Add a sender cache to Transaction #6375]. Therefore, if the sender is null, the sender recovery process (signature verification process) is performed, and the recovered sender value is not stored in the senderCache, but rather stored in the sender variable in the Transaction object. (This part is the same as the code before applying [Add a sender cache to Transaction #6375])

If you run it with the old code (version 24.1 or lower), the pre-calculated signature verification result value will be stored in the sender variable in each transaction object, and transactions that include the recovered sender value will be located in the tx pool. The senderCache is similar to the search space required when reusing recovered sender values, which can be replaced by the sender of a transaction within the tx pool. A case where the recovered sender value within a transaction is reused is, for example, the verification step for transactions in the proposed block when executing the Proof of Authority consensus algorithm (IBFT2.0, QBFT). In this stage, since the proposed block (the received block) is not trustworthy, signature verification for the transactions in the proposed block must be performed. At this time, the recovered sender value of the transaction in the tx pool was reused, and When this logic was applied, the performance was improved by about 30% compared to when it was not. (Please note that the senderCache area is excluded in this PR code.)

In my estimate, the performance improvement is expected to be similarly 30% when comparing the case where senderCache is applied and the case where senderCache is not applied, excluding my PR code. This is because although there is a difference in where(senderCache or Txpool) the sender information is found, the effect of not having to perform signature verification work by reusing the recovered sender value is the same in both cases. However, if you use transactions in the tx pool, you can reduce the space by the size of the senderCache (Of course, 20MiB is not a large size).

> For example, we use it in AbstractEngineNewPayload to improve sender recovery on block execution for Proof Of Stake
> Especially for PoS use case, I don't think we should disable the sender cache in the Transaction class. It is pretty small and consumes less than 20 MiB, more information https://github.com/hyperledger/besu/pull/6375 .

Currently, I have applied the code only to the PoA (IBFT2.0, QBFT) area, not PoS, and completed the performance test.
What consensus algorithms are provided and used by Besu for PoS, not PoA (IBFT2.0, QBFT)?
If the case where transaction.getSender() is executed in the AbstractEngineNewPayload.java file falls under the case below, there seems to be no problem in performing it with the proposed PR code.

1) If it is the first time to recover the sender, there seems to be no problem because the recovered sender value can be stored in the sender variable in the Transaction object even if there is no senderCache.
Even if this signature verification is processed in parallel, there seems to be no problem because the recovered sender value can be stored in the sender variable.

2) If transaction.getSender() is used in the transaction verification step for the proposed block (received block), the transaction sender value stored in the tx pool is retrieved, so there seems to be no problem.
Even if this signature verification is processed in parallel, there seems to be no problem because it reads the transaction sender value stored in the tx pool.

Even though this signature verification process is processed in parallel, it seems there is no problem because it reads the transaction sender value stored in the tx pool.

Are there any other problematic cases besides the ones above?
	
> In [MainnetTransactionValidator.validate](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java#L87) , there is more than just validating the sender of the transaction especially in PoS, I wonder if this is really safe, especially in this case.

In addition to verifying the transaction signature, MainnetTransactionValidator.validate also checks whether the transaction type information, code size when creating a contract, gas price, fee value, etc. are valid.

However, this verification is ultimately done with the information in the transaction object to be verified, and the transactions existing in the tx pool are transactions that have verified all of these remaining elements in addition to the signature verification.

Therefore, since the recovered sender value was found in the tx pool by comparing the transaction hash value, we can judge that all the information in the transaction object is the same and all values ​​are verified, so we can skip the MainnetTransactionValidator.validate process

ahamlat:
Thanks @2fehee for the detailed reply.

> This is because although there is a difference in where(senderCache or Txpool) the sender information is found, the effect of not having to perform signature verification work by reusing the recovered sender value is the same in both cases. However, by using the restored sender value of the transaction in the tx pool, we can reduce the space by the size of the senderCache (Of course, 20MiB is not a large size).

When adding a transaction to the transaction pool with [TransactionPool.addTransaction](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java#L234),  the content of the transaction is validated with [TransactionPool.validateTransaction](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java#L249). This method calls the[ validate method](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java#L422) on the returned TransactionValidator. If we choose the case of MainnetTransactionValidator, this means that adding a transaction to the transaction pool will trigger [MainnetTransactionValidator.validate ](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java#L87)method, which calls [validateTransactionSignature](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java#L338) and then[ transaction.getSender](https://github.com/hyperledger/besu/blob/21e95a9efbdb95e7f70dfe5d89e76956c642c06e/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionValidator.java#L368).
This is the current implementation of transaction.getSender
```
public Address getSender() {
    if (sender == null) {
      Optional<Address> cachedSender = Optional.ofNullable(senderCache.getIfPresent(getHash()));
      sender = cachedSender.orElseGet(this::computeSender);
    }
    return sender;
  }
```
So, adding a transaction to the transaction pol triggers recovering the sender from signature and storing it in the Transaction class cache. 
Based on your explanation and this path execution, I still don't see how can we have 30% improvement, maybe I'm missing something here, if you can give more insights here. As I said in the comment above, we need to understand where the improvement come from, by generating CPU profiling.

> Currently, I have applied the code only to the PoA (IBFT2.0, QBFT) area, not PoS, and completed the performance test.

The point here is that during Consensus Layer newpayload calls, we call in async and concurrent way Transaction.getSender on all the block's transactions during block RLP decoding. This is a way to precalculate the sender or get it from the cache before the block starts to execute

> What consensus algorithms are provided and used by Besu for PoS, not PoA (IBFT2.0, QBFT)?

In PoS, Besu doesn't do consensus anymore, it only executes what the Consensus client sends as requests, basically moving the chain ahead (newPayload calls) and giving it the right chain (Fork chain update/FCU calls).

2fehee:
Thanks @ahamlat for the detailed reply.

I think my explanation was lacking. I will try to explain it again.
This is the current implementation of transaction.getSender (besu 24.3.0 or later version)

```
public Address getSender() {
    if (sender == null) {
      Optional<Address> cachedSender = Optional.ofNullable(senderCache.getIfPresent(getHash()));
      sender = cachedSender.orElseGet(this::computeSender);
    }
    return sender;
  }
```

Below is the implementation of my PR code (transaction.getSender)
```
public Address getSender() {
  if (sender == null) {
    final SECPPublicKey publicKey =
        signatureAlgorithm
            .recoverPublicKeyFromSignature(getOrComputeSenderRecoveryHash(), signature)
            .orElseThrow(
                () ->
                    new IllegalStateException(
                        "Cannot recover public key from signature for " + this));
    sender = Address.extract(Hash.hash(publicKey.getEncodedBytes()));
  }
  return sender;
}
```
  
this code(my PR code) is same Besu 24.1.1 code
The reason I used Besu 24.1.1 code in my PR code is because I can see performance improvements even without senderCache.
When adding a transaction to the transaction pool with TransactionPool.addTransaction
By performing sender = Address.extract(Hash.hash(publicKey.getEncodedBytes()));
Each transaction object stores and retains recovered sender in the sender variable after signature verification

The above has the same effect as calling the computeSender() function in the besu 24.3.0 or later version code to store pre-computed sender information in the senderCache.
Of course, senderCache has a <Hash, Address> structure, so if we need the calculatedSender value, we can find the calculatedSender value using the hash of the transaction as the Key.
Similarly, in my PR code, we can find the transaction we want to find in the Tx pool by entering the Hash value of the transaction you want to search for, and find the sender value stored in the transaction object.
The sender value found in this way is used in the block consensus stage of the PoA consensus algorithm (IBFT2.0 or QBFT) to improve performance.

> When adding a transaction to the transaction pool with TransactionPool.addTransaction, the content of the transaction is validated with TransactionPool.validateTransaction. This method calls the validate method on the returned TransactionValidator. If we choose the case of MainnetTransactionValidator, this means that adding a transaction to the transaction pool will trigger MainnetTransactionValidator.validate method, which calls validateTransactionSignature and then transaction.getSender.
This is the current implementation of transaction.getSender

Yes, that's right. What you explained is exactly the process of signature verification that is executed before a transaction that has entered the node is added to the transaction pool.
My PR code does not improve performance here. The performance improvement of my code occurs, for example, in the block consensus stage (proposed block verification stage) of the PoA consensus algorithm (IBFT2.0 or QBFT).
So the main improvement codes required for performance improvement are applied to the IbftRound.java and QBftRound.java files. I will describe this part in more detail later.

> Based on your explanation and this path execution, I still don't see how can we have 30% improvement, maybe I'm missing something here, if you can give more insights here.

Let me explain in which area a 30% performance improvement is possible.

This performance improvement does not occur when transactions are added to the transaction pool. The node in the proposal turn will create a block to propose a block, and the created block will be broadcast to each validator node for consensus with other nodes. Each validator node will verify the received block (the proposed block) and verify all transactions in the received block. Transaction verification goes through several procedures, and among them, the areas where many performance bottlenecks occur are the transaction signature verification area and the transaction EVM execution area. Transaction EVM execution is executed only after transaction signature verification is completed. The reason why the transaction signature verification process is necessary at this time is because transactions in blocks received from other nodes cannot be trusted in security. Therefore, transaction signature verification must be performed at this stage, and and the transaction object cannot have a pre-computed sender value. (It is correct that there is no pre-calculated sender value in terms of security.)
Transactions that require verification will need to perform signature verification since they do not have a pre-computed sender value, but we know that the transaction pool of each validator node already contains the information about the verified transactions. Therefore, if we can confirm that the transaction is identical to a transaction in the transaction pool of the validator node, there is no need to perform signature verification logic.
To do this, if we find a transaction (ex tx2) with the same hash value in the transaction pool by inputting the hash value of the transaction (ex tx1), we can obtain the signature verification effect by filling in the pre-calculated sender value of tx2 into the empty sender variable of tx1.

The process described above can be confirmed with the contents of IbftRound.java and QbftRound.java files of my PR code implemented below.
I have processed it so that the sender value can be filled before executing updateStateWithProposedBlock.

```
public void handleProposalMessage(final Proposal msg) {
  LOG.debug(
      "Received a proposal message. round={}. author={}",
      roundState.getRoundIdentifier(),
      msg.getAuthor());
  final Block block = msg.getSignedPayload().getPayload().getProposedBlock();

  if (transactionPool.getConfigTxFastVerifySignatureEnabled()) {
    final AbstractPendingTransactionsSorter pendingTransactions;
    pendingTransactions =
        (AbstractPendingTransactionsSorter) (transactionPool.getPendingTransactionsObject());

    final List<Transaction> blockTransactions = block.getBody().getTransactions();

    for (final Transaction blockTransaction : blockTransactions) {

      final Transaction pendingTransaction =
          pendingTransactions.getTransactionByHash(blockTransaction.getHash()).orElse(null);

      if (pendingTransaction != null) {
        blockTransaction.setSender(pendingTransaction.getSender());
      }
    }
  }

  if (updateStateWithProposedBlock(msg)) {
    sendPrepare(block);
  }
```

Through this process, if we have confirmed that the sender value is filled in the signature verification step before executing the EVM for the transactions in the proposed block, we can skip the signature verification process.
The improvement effect is better in high-load situations because the signature verification process can be skipped a number of times that is almost equal to the number of transactions to be verified.
Of course, if the performance load exceeds the threshold value, the Tx pool will be full or the consensus round will be changed, which will degrade the performance, I tested it to the threshold value.


The actual performance improvement occurs in the code below. The processTransaction() function is the area that verifies the signature and runs EVM for the transaction.
Through the above processes, transactions with the sender value filled in can skip the signature verification step before EVM execution.
I implemented the code so that original signature verification logic is performed when a transaction with the same hash value cannot be found in the Tx pool due to P2P network delay.
However, when performing a performance test with 4 nodes in an AWS environment, there was no case where a transaction with the same hash value was not found in the Tx pool.
```
public TransactionProcessingResult processTransaction(...)
...
  final var transactionValidator = transactionValidatorFactory.get();
  LOG.trace("Starting execution of {}", transaction);
  ValidationResult<TransactionInvalidReason> validationResult;
  if (transaction.getSender() == null) {
	validationResult =
		transactionValidator.validate(
			transaction,
			blockHeader.getBaseFee(),
			Optional.ofNullable(blobGasPrice),
			transactionValidationParams);
	// Make sure the transaction is intrinsically valid before trying to
	// compare against a sender account (because the transaction may not
	// be signed correctly to extract the sender).
	if (!validationResult.isValid()) {
	  LOG.warn("Invalid transaction: {}", validationResult.getErrorMessage());
	  return TransactionProcessingResult.invalid(validationResult);
	}
  }
...
```

> As I said in the comment above, we need to understand where the improvement come from, by generating CPU profiling

The current performance comparison work was performed by comparing TPS values ​​in the same performance test environment. I confirmed an improvement of about 30% in the TPS value difference (1730TPS -> 2300TPS). Please refer to the PR description for detailed performance test conditions and results.

Added besu option to run performance improvement code and compare performance results.
--tx-pool-tx-fast-verify-signature-enabled=false (1730TPS)
--tx-pool-tx-fast-verify-signature-enabled=true (2300TPS)

Although the explanation is insufficient, I have explained why performance improvement is possible and the code area where performance is improved.
If you think that the explanation is not sufficient, please reply. Then, I will try to check the improvement point through CPU profiling this time.
However, I have no experience with verifying this through CPU profiling yet, so it will probably take some time to learn the method and upload the information.


> The point here is that during Consensus Layer newpayload calls, we call in async and concurrent way Transaction.getSender on all the block's transactions during block RLP decoding. This is a way to precalculate the sender or get it from the cache before the block starts to execute
In PoS, Besu doesn't do consensus anymore, it only executes what the Consensus client sends as requests, basically moving the chain ahead (newPayload calls) and giving it the right chain (Fork chain update/FCU calls).

This PR code has been tested and improved only for PoA (IBFT2.0, QBFT) suitable for enterprise blockchain.
Based on the explanation above, is the PR code not working properly for PoS? If you could give me your opinion, it would be helpful for analyzing the PoS part of the code.
What do you think about setting --tx-pool-tx-fast-verify-signature-enabled=true and running it only when operating with PoA (IBFT2.0, QBFT)?

ahamlat:
What I was saying before, is that the cache should give the same result as your current PR, and you shouldn't disable the cache anyway as we're using it to optimize Block processing for PoS. The same code is used for PoS and PoA networks, optimizing one network shouldn't have an impact on the other one.
I would expect to have exactly the same throughput as this PR if you run only the modification below and keeping the same code as now (before our PR).
transaction.getSender() should return a the sender from the cache if the transaction is transaction pool. This is because, each time we add a transaction to the transaction pool, we recover the sender and add it to the cache. 

```
if (transaction.getSender() == null) {
	validationResult =
		transactionValidator.validate(
			transaction,
			blockHeader.getBaseFee(),
			Optional.ofNullable(blobGasPrice),
			transactionValidationParams);
	// Make sure the transaction is intrinsically valid before trying to
	// compare against a sender account (because the transaction may not
	// be signed correctly to extract the sender).
	if (!validationResult.isValid()) {
	  LOG.warn("Invalid transaction: {}", validationResult.getErrorMessage());
	  return TransactionProcessingResult.invalid(validationResult);
	}
  }
 ```

>  What do you think about setting --tx-pool-tx-fast-verify-signature-enabled=true and running it only when operating with PoA (IBFT2.0, QBFT)?

I'm still not convinced to be honest for two reasons :
- Even if is enabled with a flag, removing the cache will be definitive and not depend on the flag, and therefore will have an impact on PoS block processing time.
- As I explained above, I still don't see how this PR could improve the throughput. I'm not saying that it doe not improve the throughput, I just don't understand yet, and profiling would be great here. It could be that the cache is not as efficient as I think if there're a lot of evictions

PR #7731, opened 2024-10-04
Add jna native setup to besu and evmtool

garyschulte:
## PR description

PR to normalize jna path for native library loading, so that os identification is not a problem when loading besu-native libs with JNA.  

For example, an x86_64 linux system might identify as either `linux-gnu-x86_64` or `linux-x86_64`.  Rather than duplicating the library files in besu-native so that the native loader will find them, this PR just ensures we look in both locations.


garyschulte:
seems at least some jvms ignore this config when extracting libs from jars like besu-native.

PR #7737, opened 2024-10-06
Optimize Dockerfile

c0rv4x:
## PR description

Optimized Dockerfile by combining RUN commands to reduce image layers.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,20 +3,16 @@ ARG VERSION="dev"
 ENV NO_PROXY_CACHE="-o Acquire::BrokenProxy=true -o Acquire::http::No-Cache=true -o Acquire::http::Pipeline-Depth=0"
 
 # Update and install dependencies without using any cache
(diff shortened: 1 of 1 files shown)


macfarla:
@cdivitotawela could you review this one

cdivitotawela:
@c0rv4x thanks for the PR. All RUN commands are currently chained together so all in a single image layer. I do not see how the proposed changes reduce any image layers. Anything specific you trying to achieve with the proposed change? 

I will also have another check to see anything I have missed.

cdivitotawela:
@c0rv4x I have checked further and could not see a change in number of image layers nor reduction in image size with the proposed change. I built the image using the existing Dockerfile and proposed Dockerfile. Compared the images checking the history and size. Could not see any benefit with the proposed change.

cdivitotawela:
Proposed change does not reduce number of image layers nor the size of the layer. Closing the PR as its intended purpose is not achieved.

PR #7738, opened 2024-10-07
Replace custom required options check with standard PicoCLI features

fab-10:
## PR description

Replace the custom made `validateRequiredOptions`, that check if a required option is also not empty, with PicoCLI standard features and validation using JSR-380 BeanValidation, because the custom made validation breaks some feature of PicoCLI like https://picocli.info/#_mutually_dependent_options


gradle/versions.gradle
```diff
@@ -157,6 +157,8 @@ dependencyManagement {
 
     dependency 'org.fusesource.jansi:jansi:2.4.1'
 
+    dependency 'org.hibernate.validator:hibernate-validator:8.0.1.Final'
```
jflo, line 160: surprised to see this, can we narrow this dependency down to just jakarta?

jflo (commented):
love this! note on the dependency on hibernate-validator, seems overly broad?

gradle/versions.gradle
fab-10: jakarta is only the API, we need an implementation and it is provided by `hibernate-validator`

usmansaleem (approved):
Looks straight forward.

PR #7739, opened 2024-10-07
Sonar cloud build failure

jflo:
- updates jacoco plugins to aggregate reports across sub-projects correctly.
- excludes testutils subproject from analysis since it triggers a bug in sonar.
- only runs block data generation for import/export tests if an existing database doesn't already exist.


gradle/verification-metadata.xml
```diff
@@ -3593,6 +3593,19 @@
...
+         </artifact>
+         <artifact name="gradle-taskinfo-2.2.0.module">
+            <sha256 value="6352b4fcc722de060c073ed02b6efa1615324af9e727d252a189d9d7c27614dc" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
+      <component group="org.barfuin.gradle.taskinfo" name="org.barfuin.gradle.taskinfo.gradle.plugin" version="2.2.0">
+         <artifact name="org.barfuin.gradle.taskinfo.gradle.plugin-2.2.0.pom">
+            <sha256 value="917266d280b586c5f73133e50768405f9f4368a0537b69e4923504e249317e7e" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
```
garyschulte, line 3608: new transitive deps?

garyschulte (commented):
I think the verification metadata is unnecessary, otherwise lgtm

garyschulte (approved):

PR #7750, opened 2024-10-09
Fix eth_feeHistory rewards when bounded by configuration

fab-10:
## PR description

The issue is that, when bounding rewards returned by `eth_feeHistory` by configuration, the `minPriorityFeePerGas` runtime value is used, but it could not be great enough to return acceptable values in order to estimate which `priorityFeePerGas` to use in a tx to ensure it is paying enough to be mineable.
To fix that a _a mineable lower bound_ for the priority fee is calculated and used in place of `minPriorityFeePerGas` conf value, when it is greater that the latter, so we are reasonably sure that we are returning value that can be used to estimate the priority fee when submitting new txs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Additions and Improvements
 
 ### Bug fixes
+- Fix eth_feeHistory rewards when bounded by configuration [#7750](https://github.com/hyperledger/besu/pull/7750)
 
 ## 24.10.0
 
--- a/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
+++ b/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
@@ -39,6 +39,7 @@ import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.RpcErrorType;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.results.FeeHistory;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.results.ImmutableFeeHistory;
 import org.hyperledger.besu.ethereum.api.jsonrpc.internal.results.ImmutableFeeHistoryResult;
+import org.hyperledger.besu.ethereum.api.query.BlockchainQueries;
 import org.hyperledger.besu.ethereum.blockcreation.MiningCoordinator;
 import org.hyperledger.besu.ethereum.chain.Blockchain;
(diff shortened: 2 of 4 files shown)


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java
```diff
@@ -179,14 +187,51 @@ public void shouldBoundRewardsCorrectly() {
...
+    final var blockchainQueries = mockBlockchainQueries(blockchain, Wei.of(10));
+    when(miningCoordinator.getMinPriorityFeePerGas()).thenReturn(Wei.ZERO);
+
+    EthFeeHistory ethFeeHistory =
+        new EthFeeHistory(null, blockchainQueries, miningCoordinator, apiConfiguration);
+
+    List<Wei> rewards = ethFeeHistory.computeRewards(rewardPercentiles, block, Wei.of(8));
+
+    // Define the expected bounded rewards for each percentile
+    List<Wei> expectedBoundedRewards = Stream.of(4, 4, 4, 4, 5, 6, 6, 6, 6).map(Wei::of).toList();
```
Filter94, line 233: I don't get this test. If lower bound is set to 10 Wei, why are we getting values lower than that?

fab-10: in this case `minGasPrice` is 10wei, and since `baseFee` is 8wei, the calculated min priority fee (necessary to get the tx minable) is 2wei that is greater than the configured `minPriorityFeePerGas=0` so the former is used.

Then the rewards (effective priority fees) are bounded by 200 and 300 coefficients, so the rewards must be in the interval 4wei and 6wei

Filter94: Ok, I'm following it now, thanks!

macfarla (approved):

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 - Fix an unhandled PeerTable exception [#7733](https://github.com/hyperledger/besu/issues/7733)
 - Fix RocksDBException: Busy leading to MerkleTrieException: Unable to load trie node value [#7745](https://github.com/hyperledger/besu/pull/7745)
 - If a BFT validator node is syncing, pause block production until sync has completed [#7657](https://github.com/hyperledger/besu/pull/7657)
+- Fix eth_feeHistory rewards when bounded by configuration [#7750](https://github.com/hyperledger/besu/pull/7750)
```
macfarla, line 34: prob needs moving to next release section

PR #7751, opened 2024-10-09
Fix BASIC_DATA_LEAF update on account with existing code size value

lu-pinto:
## PR description
Spotted an issue while running one of the reference tests when an account already exists with code and only the account data is updated in the trie. In that case, since there won't be any updates for code, the current code will not be added to the BASIC_DATA_LEAF.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -233,6 +233,23 @@ private void generateAccountValues(
...
+      final VerkleEntryFactory verkleEntryFactory,
+      final DiffBasedValue<VerkleAccount> accountUpdate) {
+    final VerkleAccount priorAccount = accountUpdate.getPrior();
+    final VerkleAccount updatedAccount = accountUpdate.getUpdated();
+    if (priorAccount == null) {
+      verkleEntryFactory.generateCodeHashKeyValueForUpdate(
+          accountKey, updatedAccount.getCodeHash());
+      return;
+    }
+    Bytes currentCode = get(accountKey).getCode();
```
matkt, line 247: why we are not using accumulator to retrieve the code instead of reading from database ?

matkt (approved):

PR #7753, opened 2024-10-10
don't do too much work when disabled

pinges:
## PR description
This PR reduces the work that is done when the NewPooledTransactionHashesMessageHandler and the TransactionsMessageHandler are disabled.


Gabriel-Trintinalia (approved):
lgtm

PR #7754, opened 2024-10-10
Devnet4: 7702 spec changes

### LINKED ISSUE #7706: Devnet-3: Besu forked off at block 110500

daniellehrner:
## PR description

* Warm target of delegated account if a delegated account is the transaction destination
* restrict authorization nonce to 2**64 -1
* fix gas calculation for code delegation resolution if target is precompile
* verify auth nonce less than 2**64-1 during auth processing
* auth to zero address deletes delegation
* auth to precompile returns empty code
* auth nonce < 2**8
* increase auth base cost to 12500

## Fixed Issue(s)

fixes #7706


daniellehrner:
Closing in favor of #7809

Issue #7761, opened 2024-10-11
Remove deprecated --host-whitelist option


jflo:
The option `--host-whitelist` has been deprecated since 2020, and can be removed now.

-- host-allowlist already exists as an alias for this option, so it's only removing the name, not the functionality

Add to the upcoming changelog to indicate it will be removed soon.

TC1211:
Could me and @janetmo take this issue?

macfarla:
@janetMo you need to comment on here and then can assign you

JanetMo:
Yes I would work with @TC1211 on the issue, thanks!

jflo:
assigned to you two, enjoy.

pr9t:
Hi @macfarla please assign this issue to me.

PR #7762, opened 2024-10-11
communicate out option removal

jflo:
addresses but doesn't yet complete #7761


macfarla (approved):
approved but we need to do a thorough audit of all nodes we run to ensure they use the new config!

Issue #7763, opened 2024-10-14
Sync Halt with Best peer has wrong pivot block


siladu:
Sync hasn't progressed and repeatedly gets this error...
`{"@timestamp":"2024-10-14T04:14:06,834","level":"WARN","thread":"nioEventLoopGroup-3-4","class":"SyncTargetManager","message":"Best peer has wrong pivot block (#20939063) expecting 0x1e84b312979f4e6f2708e31fcc52bfbcc5a0aac7047565cbfd06fcb16a8623b8 but received 0x7466118286794e63be64f59b31665db8297d640432ea084c3d201d5097d44908.  Disconnect: PeerId: 0xdef2fc1486c65961... PeerReputation score: 102, timeouts: {}, useless: 0, validated? true, disconnected? false, client: Geth/v1.13.14-stable-2ecaf439/linux-amd64/go1.22.2, [Connection with hashCode -1435576429 inboundInitiated? false initAt 1728879241331], enode://def2fc1486c65961da0fdd8545dd55d8d3b4b23c855523b6f142a27d1c7a3f02365a25724c4b607ed778e68a1cad68dc5ce84c6c325ea77671a5d74dd9f55ba5@57.129.54.20:30000, isServingSnap false, has height 20961355, connected for 5503 ms","throwable":""}`

When this error started:
```
{"@timestamp":"2024-10-11T01:21:09,424","level":"INFO","thread":"EthScheduler-Services-692 (importBlock)","class":"ImportBlocksStep","message":"Block import progress: 17339357 of 20938939 (82%), Peer count: 25","throwable":""}
{"@timestamp":"2024-10-11T01:21:31,278","level":"INFO","thread":"EthScheduler-Services-771","class":"PivotSelectorFromHeadBlock","message":"Returning head block hash 0x1e84b312979f4e6f2708e31fcc52bfbcc5a0aac7047565cbfd06fcb16a8623b8 as pivot","throwable":""}
{"@timestamp":"2024-10-11T01:21:31,279","level":"INFO","thread":"EthScheduler-Services-771","class":"PivotSelectorFromBlock","message":"Selecting new pivot block: Optional[0x1e84b312979f4e6f2708e31fcc52bfbcc5a0aac7047565cbfd06fcb16a8623b8]","throwable":""}
[...]

PR #7764, opened 2024-10-14
Disable --Xsnapsync-to-head-enabled by default

siladu:
Due to this bug: https://github.com/hyperledger/besu/issues/7763


jframe (approved):
LGTM

PR #7778, opened 2024-10-17
stem based flat db for verkle

matkt:
## PR description

This PR introduces a new strategy for managing Verkle accounts using a stem-based flat database structure. Instead of saving by account, slot, and code like in the legacy approach, this implementation stores data by stem and code. This shift allows for direct access to the leaves of the Verkle tree, significantly reducing data duplication and improving storage efficiency.

The stem-based flat database strategy is designed to align better with Verkle tree structures, enabling more streamlined storage and access patterns. This should make it more effective in scenarios that require quick access to account data without the overhead of managing redundant information.

Key Changes:

Introduces the VerkleStemFlatDbStrategy class, which handles storage using a stem-based approach.

Reduces duplication by focusing on storing data at the stem level, improving overall efficiency.

Next Steps:

Once this PR is merged, the optimization of Pedersen hash calculations will be handled in a subsequent PR.

Testing will be conducted after the pre-image is available to determine the usability of this implementation in its current state.

Considerations:

This new approach is expected to be necessary for stateless clients, but further verification will be needed for regular nodes to ensure that it meets the requirements effectively.


besu/src/main/java/org/hyperledger/besu/cli/options/storage/VerkleSubStorageOptions.java
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Validates the data storage options
+   *
+   * @param commandLine the full commandLine to check all the options specified by the user
+   * @param dataStorageFormat the selected data storage format which determines the validation rules
+   *     to apply.
+   */
+  public void validate(final CommandLine commandLine, final DataStorageFormat dataStorageFormat) {
+    // no op
+  }
```
lu-pinto, line 73: remove method?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/VerkleSubStorageConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+  @Value.Default
+  default VerkleUnstable getUnstable() {
+    return VerkleUnstable.DEFAULT;
+  }
+
+  @Value.Immutable
+  interface VerkleUnstable {
+
+    VerkleSubStorageConfiguration.VerkleUnstable DEFAULT =
+        ImmutableVerkleSubStorageConfiguration.VerkleUnstable.builder().build();
```
lu-pinto, line 34: DEFAULT is essentially == DISABLED right? So why have both? Just do:
```
  @Value.Default
  default VerkleUnstable getUnstable() {
    return VerkleUnstable.DISABLED;
  }
```
And remove the DEFAULT field.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/common/AbstractStateTrieAccountValue.java
```diff
@@ -76,6 +78,16 @@ public Hash getStorageRoot() {
...
 
+  /**
+   * The size of the EVM bytecode associated with this account.
+   *
+   * @return the size of the account code (which may be {@link Optional#empty()}).
+   */
+  @Override
+  public Optional<Long> getCodeSize() {
+    return Optional.empty();
+  }
```
lu-pinto, line 89: how can `codeHash != Hash.EMPTY` and code size be empty? It doesn't seem to make sense

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -464,7 +465,9 @@ public Hash frontierRootHash() {
     return calculateRootHash(
         Optional.of(
             new BonsaiWorldStateKeyValueStorage.Updater(
-                noOpSegmentedTx, noOpTx, worldStateKeyValueStorage.getFlatDbStrategy())),
+                noOpSegmentedTx,
+                noOpTx,
+                (BonsaiFlatDbStrategy) worldStateKeyValueStorage.getFlatDbStrategy())),
```
lu-pinto, line 470: do you really need to cast it?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/accumulator/preload/CodeConsumingMap.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * A map that stores Ethereum addresses and their associated code values, with a consumer to process
+ * updates.
+ *
+ * <p>Each time a new value is added or updated, the consumer processes the code value.
+ *
+ * <p>This class uses a thread-safe {@link ConcurrentMap} to store data, allowing concurrent
+ * modifications.
+ */
+public class CodeConsumingMap extends ForwardingMap<Address, DiffBasedValue<Bytes>> {
```
lu-pinto, line 36: It looks like `AccountConsumingMap` is almost a complete copy of this class and you can achieve the same reusing that class. Why not renaming `AccountConsumingMap` -> `ConsumingMap` and declare it like:
```
codeToUpdate = new ConsumingMap<Bytes>(new ConcurrentHashMap<>(), codePreloader);
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/VerkleAccount.java
```diff
@@ -94,48 +94,58 @@ public static VerkleAccount fromRLP(
...
   }
 
   @Override
   public void writeTo(final RLPOutput out) {
     out.startList();
 
-    out.writeLongScalar(nonce);
     out.writeUInt256Scalar(balance);
-    out.writeBytes(storageRoot);
+    out.writeLongScalar(nonce);
```
lu-pinto, line 142: why did you swap nonce with balance for both reading and writing? Any reason in particular?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/VerkleAccount.java
```diff
@@ -94,48 +94,58 @@ public static VerkleAccount fromRLP(
...
+
+  @Override
+  public void setCode(final Bytes code) {
+    super.setCode(code);
+    if (code == null || code.isEmpty()) {
+      this.codeSize = 0L;
+    } else {
+      this.codeSize = code.size();
+    }
+  }
```
lu-pinto, line 120: So there is a kind of code smell to me here. When you construct `VerkleAccount` you have a `codeSize != 0` but `getCode() == null` this is not clear at all unless you are totally familiar with Verkle. And even then...

I understand that `code` and `codeSize` are 2 different regions of the DB and so is `codeHash` and that performance wise we should avoid touching slots unnecessarily but why not having a code provider with a lambda function instead of the field `Code code` that only populates it and goes to the DB if someone calls `getCode`? That would be more intuitive to me and consistent. Don't you know at least a way to get the code from the DB when constructing a `VerkleAccount` ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -440,7 +429,15 @@ public Map<Bytes32, Bytes> getAllAccountStorage(final Address address, final Has
   }
 
   private VerkleTrie createTrie(final NodeLoader nodeLoader, final Bytes32 rootHash) {
```
lu-pinto, line 431: `rootHash` unused var, can be removed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/VerkleWorldStateKeyValueStorage.java
```diff
@@ -42,34 +46,50 @@
...
+      final StemPreloader stemPreloader,
+      final DataStorageConfiguration dataStorageConfiguration,
+      final MetricsSystem metricsSystem) {
     super(composedWorldStateStorage, trieLogStorage);
-    this.flatDbStrategy = flatDbStrategy;
+    this.metricsSystem = metricsSystem;
+    this.dataStorageConfiguration = dataStorageConfiguration;
+    this.flatDbStrategyProvider =
+        new VerkleFlatDbStrategyProvider(metricsSystem, dataStorageConfiguration);
+    flatDbStrategyProvider.loadFlatDbStrategy(composedWorldStateStorage);
```
lu-pinto, line 86: Why is that the FlatDbStrategyProvider not calling `loadFlatDbStrategy` ? Is the object of any use without anything loaded on it?

I actually found out then that the `getFlatDbStategy` needs the same argument as the load and I found that weird. After checking the getter I can see that the getter calls `loadFlatDbStrategy`. Any particular reason for that? Why can't we call load in the constructor and make sure `flatDbStrategy is never null inside? Do we need lazy initialization? Only if it makes sense to create a provider and not use it...

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/VerkleWorldStateKeyValueStorage.java
```diff
@@ -164,6 +207,19 @@ public Updater putAccountInfoState(final Hash accountHash, final Bytes accountVa
...
+          composedWorldStateTransaction, accountHash, slotHash, storage);
+      return this;
+    }
+
+    public synchronized void removeStorageValueBySlotHash(
+        final Hash accountHash, final Hash slotHash) {
+      flatDbStrategy.removeFlatAccountStorageValueByStorageSlotHash(
+          composedWorldStateTransaction, accountHash, slotHash);
+    }
+
```
lu-pinto, line 222: `putStateTrieNode` and `removeStateTrieNode` are unused

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/VerkleWorldStateKeyValueStorage.java
```diff
@@ -164,6 +207,19 @@ public Updater putAccountInfoState(final Hash accountHash, final Bytes accountVa
       return this;
     }
 
+    public synchronized Updater putStorageValueBySlotHash(
```
lu-pinto, line 210: This Updater has a bunch of methods `synchronized` but then others are not. Does it make sense to have `synchronized` at all? If so others are also writing to the same object as these

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
```diff
@@ -453,7 +450,9 @@ public Hash frontierRootHash() {
     return calculateRootHash(
         Optional.of(
             new VerkleWorldStateKeyValueStorage.Updater(
-                noOpSegmentedTx, noOpTx, worldStateKeyValueStorage.getFlatDbStrategy())),
+                noOpSegmentedTx,
+                noOpTx,
+                (VerkleStemFlatDbStrategy) worldStateKeyValueStorage.getFlatDbStrategy())),
```
lu-pinto, line 455: Do you need to define `frontierRootHash`? From what I can see this is only used to define transaction receipts compliant with Frontier but Berlin is already overriding that. I can't see why we need to have verkle with Frontier.

evm/src/main/java/org/hyperledger/besu/evm/frame/MessageFrame.java
```diff
@@ -929,6 +929,17 @@ public boolean wasCreatedInTransaction(final Address address) {
...
+   * Checks if there is an account for this address in the world state. This is different from
+   * `wasCreatedInTransaction` since an account could have been created in another transaction.
+   *
+   * @param address the address to check
+   * @return true if account already exists in world state
+   */
+  public boolean accountExists(final Address address) {
+    return worldUpdater.get(address) != null;
+  }
+
```
lu-pinto, line 942: Unused, I would remove

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/flat/VerkleLegacyFlatDbStrategy.java
```diff
@@ -0,0 +1,186 @@
...
+ * <p>This strategy handles accounts, slots, and code similar to the legacy approach, ensuring
+ * consistency and compatibility with the pre-Verkle structure.
+ */
+public class VerkleLegacyFlatDbStrategy extends FlatDbStrategy {
+
+  protected final Counter getAccountNotFoundInFlatDatabaseCounter;
+
+  protected final Counter getStorageValueNotFoundInFlatDatabaseCounter;
+
+  public VerkleLegacyFlatDbStrategy(
```
lu-pinto, line 53: Could use some unit tests to make sure metrics work as expected at least

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/flat/VerkleStemFlatDbStrategy.java
```diff
@@ -0,0 +1,231 @@
...
+import org.apache.tuweni.bytes.Bytes32;
+
+/**
+ * Strategy for managing Verkle accounts in the flat database using a stem-based structure.
+ *
+ * <p>Unlike the legacy approach of saving by account, slot, and code, this strategy saves by stem
+ * and code. This allows direct access to the leaves of the Verkle trie for stems, reducing
+ * duplication of data.
+ */
+public class VerkleStemFlatDbStrategy extends FlatDbStrategy {
```
lu-pinto, line 53: nit: could use some unit tests to make sure metrics work as expected at least

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/trielog/TrieLogFactoryImpl.java
```diff
@@ -107,7 +108,21 @@ public static void writeTo(final TrieLog layer, final RLPOutput output) {
...
+                new VerkleStateTrieAccountValue(
+                        verkleAccount.getNonce(),
+                        verkleAccount.getBalance(),
+                        verkleAccount.getCodeHash(),
+                        verkleAccount.getCodeSize())
+                    .writeTo(o);
+              } else {
+                sta.writeTo(o);
+              }
+            });
```
lu-pinto, line 125: could you add a unit test?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/VerkleSubStorageConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+
+    VerkleSubStorageConfiguration.VerkleUnstable STEM_MODE =
+        ImmutableVerkleSubStorageConfiguration.VerkleUnstable.builder()
+            .stemFlatDbEnabled(true)
+            .build();
+
```
lu-pinto, line 40: DISABLED is not used and it seems `DEFAULT == DISABLED`. Maybe remove `DEFAULT` and just do:
```
@Value.Default
  default VerkleUnstable getUnstable() {
    return VerkleUnstable.DISABLED;
  }
```

lu-pinto (commented):
No blockers but it could use some cleanups

besu/src/main/java/org/hyperledger/besu/cli/options/storage/DataStorageOptions.java
```diff
@@ -76,6 +79,7 @@ public static DataStorageOptions create() {
    */
   public void validate(final CommandLine commandLine) {
     diffBasedSubStorageOptions.validate(commandLine, dataStorageFormat);
+    verkleSubStorageOptions.validate(commandLine, dataStorageFormat);
```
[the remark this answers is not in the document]
matkt: It's just to ensure that all the logic is coded in the same way so that if tomorrow someone adds a flag that requires a check, they don’t have to wonder why it doesn’t work or how to do it. They just need to follow the same approach as in other parts of the code and simply implement the check in `validate`. But if you think it's better to remove why not

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/common/AbstractStateTrieAccountValue.java
matkt: Optional.empty is not == 0 it's a way to say that the codesize is not available but I modified the code and this method is now abstract  and in **PmtStateTrieAccountValue** I throw an exception 

```java 
@Override
  public Optional<Long> getCodeSize() {
    throw new UnsupportedOperationException("Retrieving code size from state trie account value is not possible with Patricia Merkle Trie");
  }
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
matkt: fixed that

besu/src/main/java/org/hyperledger/besu/cli/options/storage/DataStorageOptions.java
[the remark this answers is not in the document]
lu-pinto: no strong opinions, you can leave it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/accumulator/preload/CodeConsumingMap.java
matkt: it's not completly the same, there is a difference in the consumer.process part (put method)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/VerkleAccount.java
matkt: yes it was to be compatible with the VerkleStateTrieAccountValue encoding part as I decided to implement it like that. no more reason than that I think I can revert this change

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/VerkleAccount.java
matkt: It's already working like that, it's using the same code as Bonsai, VerkleAccount is calling the context (MutableWorldstate) to retrieve the code when we call getCode 

See in DiffBasedAccount
```java
 public Bytes getCode() {
    if (code == null) {
      code = context.getCode(address, codeHash).orElse(Bytes.EMPTY);
    }
    return code;
  }
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/VerkleWorldStateKeyValueStorage.java
matkt: updated it was needed before but I think we had some refactoring of this part and it's not needed anymore in bonsai and verkle, I updated the code

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/VerkleWorldStateKeyValueStorage.java
matkt: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/VerkleWorldStateKeyValueStorage.java
matkt: I think it's just a copy/paste from Bonsai code, looking at the code I also don't see a need for a synchronized for the moment

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/trielog/TrieLogFactoryImpl.java
```diff
@@ -107,7 +108,21 @@ public static void writeTo(final TrieLog layer, final RLPOutput output) {
       if (accountChange == null || accountChange.isUnchanged()) {
         output.writeNull();
       } else {
-        writeRlp(accountChange, output, (o, sta) -> sta.writeTo(o));
+        writeRlp(
```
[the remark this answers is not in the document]
matkt: I tried to do some refactoring and avoid as much as possible duplication for this class

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
matkt: I prefer to keep it this way for now because I’d like to sync the mainnet with Verkle for testing purposes, and I might need this method. We can remove it later when it’s no longer useful.

evm/src/main/java/org/hyperledger/besu/evm/frame/MessageFrame.java
matkt: not sure why it's here 🤔 I will remove it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/trielog/TrieLogFactoryImpl.java
matkt: add tests for that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/flat/VerkleLegacyFlatDbStrategy.java
matkt: did

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/storage/flat/VerkleStemFlatDbStrategy.java
matkt: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
lu-pinto: Or are you building this for the future, to use more specialized methods? I don't get any compilation error if I do that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/accumulator/preload/CodeConsumingMap.java
lu-pinto: ah true! missed that, in `CodeConsumingMap` only the updated value is processed by the consumer. But is there any reason for the Consumer not to do that itself and only use the updated value, rather than doing it inside the `CodeConsumingMap` ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/VerkleAccount.java
lu-pinto: Yes I understand, this is not something just with Verkle but Bonsai also has the same problem. Maybe nothing to do for now, but it would be better to clarify.

Right now, there's so many ways you can set code in both: using a `context`, calling `setCode` directly, doing a copy constructor way. I feel this has potential to create problems and at least it's hard for me to deterministically say how the code is set into an account. Why can't you have code set always at construct time? Agree needs some thought though.

But maybe it's because we're saving on object allocation and reusing `*Account` objects?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/worldview/VerkleWorldState.java
lu-pinto: Maybe add a comment that you want it for that?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
matkt: Actually, I prefer the current method. everything related to Diffbased or the Bonsai/Verkle world. if I know we're in a Bonsai class, I directly specify it  and the subclasses should also be Bonsai as they will be required to follow that structure. This ensures that in the future, there's no need to question whether a specific Bonsai method needs to be called, and it also avoids others having to question it later if needed. The right types are already in place.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/accumulator/preload/CodeConsumingMap.java
matkt: not really , more because I didn't wanted to push the logic here

```java
public DiffBasedWorldStateUpdateAccumulator(
      final DiffBasedWorldView world,
      final Consumer<DiffBasedValue<ACCOUNT>> accountPreloader,
      final Consumer<StorageSlotKey> storagePreloader,
      final Consumer<Bytes> codePreloader,
      final EvmConfiguration evmConfiguration) {
    super(world, evmConfiguration);
    this.accountsToUpdate = new AccountConsumingMap<>(new ConcurrentHashMap<>(), accountPreloader);
    this.codeToUpdate = new CodeConsumingMap(new ConcurrentHashMap<>(), codePreloader);
    this.accountPreloader = accountPreloader;
    this.storagePreloader = storagePreloader;
    this.codePreloader = codePreloader;
    this.isAccumulatorStateChanged = false;
    this.evmConfiguration = evmConfiguration;
  ```
  but I don't have a strong opinion about that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/verkle/VerkleAccount.java
matkt: it's an optmization, we don't have to read a code if we don't need it . the idea is to read in the database only when we really want to access the code

also I don't see a potential problem with that. The context is used when you read the code from the database, after when you modify the code you will always pass via the account. the worldstate (via accumulator) will check the code in the account and save in the database

matkt: I think I will have in the futur to explain all of this logic in a doc (accumulator, worldstate, updater etc)

lu-pinto (approved):
LGTM

PR #7798, opened 2024-10-22
Refactoring around MiningParameters

jflo:
- optionalized some fields of ProtocolSchedule constructor, which will allow us to reduce number of constructors later.
- Refactored JsonBlockImporterTest so it can provide a BesuComponent
- ProtocolScheduleBuilder optionalizes the default chainid
- Blocks SubCommand is now Dagger aware
- Creates but rarely uses ProtocolSpec and ProtocolSchedule modules. Deeper adoption pending.
- Adds a Coinbase module with fixed values for use in testing.
- Introduces an EthereumCoreComponent, to be used as a subcomponent of BesuComponent in the future.
- Introduces a MiningParameters module with common static values used by tests. Removes MiningParameters static constructor.

tested by syncing to the head of ephemery testnet


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolScheduleModule.java
```diff
@@ -0,0 +1,100 @@
...
+   * @param evmConfiguration the EVM configuration
+   * @param badBlockManager the bad block manager
+   * @param isParallelTxProcessingEnabled whether parallel tx processing is enabled
+   * @param metricsSystem the metrics system
+   * @param miningParameters the mining parameters
+   * @return the protocol schedule builder
+   */
+  @Singleton
+  @Provides
+  public ProtocolScheduleBuilder provideProtocolSchedule(
```
garyschulte, line 58: probably rename to provideProtocolScheduleBuilder() to reflect the actual type

ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/RetestethContext.java
```diff
@@ -270,7 +270,7 @@ private boolean buildContext(
             syncState,
             transactionPoolConfiguration,
             new BlobCache(),
-            MiningParameters.newDefault());
+            miningParameters);
```
garyschulte, line 273: Not related to this PR, but IDK if retesteth is a thing anymore.  We should find out if EF testing is using it.  It would be nice to get rid of some more unused features.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilder.java
```diff
@@ -48,59 +48,12 @@ public class ProtocolScheduleBuilder {
...
-      final EvmConfiguration evmConfiguration,
-      final MiningParameters miningParameters,
-      final BadBlockManager badBlockManager,
-      final boolean isParallelTxProcessingEnabled,
-      final MetricsSystem metricsSystem) {
-    this(
-        config,
-        Optional.empty(),
-        protocolSpecAdapters,
-        privacyParameters,
```
garyschulte, line 94: if private chain protocol schedules/specs inherited from something other than mainnet, we could remove privacyParameters from mainnet entirely 🤔

garyschulte (commented):
Looks good.  Some comments, dagger questions, etc.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolScheduleModule.java
jflo: Yes it does. It's cool because in a test context, you would be able to create a component for each test, and the schedule is singleton to that component. If you wanted a test to cross 2 schedules, just create 2 components.

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilder.java
```diff
@@ -90,7 +91,7 @@ public BftProtocolSchedule createProtocolSchedule(
     final ProtocolSchedule protocolSchedule =
         new ProtocolScheduleBuilder(
                 config,
-                DEFAULT_CHAIN_ID,
+                Optional.of(DEFAULT_CHAIN_ID),
```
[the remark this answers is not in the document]
jflo: I don't follow, which constructor is overloaded? ProtocolScheduleBuilder only has 1 constructor.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolScheduleBuilder.java
jflo: oh hell yeah. that might be a good aspect to cleave this apart.

ethereum/retesteth/src/main/java/org/hyperledger/besu/ethereum/retesteth/RetestethContext.java
garyschulte: removed by #7833

besu/src/main/java/org/hyperledger/besu/controller/CliqueBesuControllerBuilder.java
```diff
@@ -132,7 +132,6 @@ protected ProtocolSchedule createProtocolSchedule() {
         genesisConfigOptions,
         forksSchedule,
         nodeKey,
-        privacyParameters,
```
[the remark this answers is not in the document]
garyschulte: why is it safe to use this constructor which uses default privacy parameters rather than the superclass' privacy parameters which may (or may not) have been set on the CLI

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilder.java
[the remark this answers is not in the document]
garyschulte: it is gone now 👍

garyschulte (commented):
LGTM, but I still have that question about why we are dropping privacy params for clique protocol scheduels

macfarla (approved):

jflo:
> LGTM, but I still have that question about why we are dropping privacy params for clique protocol scheduels

Sorry, my reply must've gotten dropped. The answer is that it was only ever passed in as [a hardcoded default](https://github.com/hyperledger/besu/blob/9e7b6aead7bd7114f628ba14c14ffbf18335d1bd/consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/CliqueProtocolSchedule.java#L155).

garyschulte:
> > LGTM, but I still have that question about why we are dropping privacy params for clique protocol scheduels
> 
> Sorry, my reply must've gotten dropped. The answer is that it was only ever passed in as [a hardcoded default](https://github.com/hyperledger/besu/blob/9e7b6aead7bd7114f628ba14c14ffbf18335d1bd/consensus/clique/src/main/java/org/hyperledger/besu/consensus/clique/CliqueProtocolSchedule.java#L155).


This still looks wrong imo.  The link you provides is just the alternate create() that doesn't take privacy parameters.  ❓ 

As I see it, BesuCommand is going to set privacy parameters on the controller builder, derived from CLI params.  For a clique network, that will be a CliqueBesuControllerBuilder, with the privacy params set by BesuCommand.  

In CliqueBesuControllerBuilder when we use the ProtocolSchedule.create method (the one with privacy params on 115) the privacy params from BesuCommand are used.  But if we use the other create (155), we are discarding the BesuCommand configured privacy params. 

It looks like the ProtocolScheduleBuilder that is return is indeed creating ProtocolSpec instances with the privacy params - so it isn't like they are being ignored.  What am I missing?

PR #7807, opened 2024-10-24
Data storage option refactoring 

matkt:
## PR description

The idea of this PR is to continue renaming and refactoring the classes common to both Bonsai and Verkle to facilitate the future integration of Verkle.


besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2127,26 +2127,26 @@ public DataStorageConfiguration getDataStorageConfiguration() {
 
     if (SyncMode.FULL.equals(getDefaultSyncModeIfNotSet())
         && DataStorageFormat.BONSAI.equals(dataStorageConfiguration.getDataStorageFormat())
-        && dataStorageConfiguration.getBonsaiLimitTrieLogsEnabled()) {
+        && dataStorageConfiguration.getDiffbasedLimitTrieLogsEnabled()) {
```
lu-pinto, line 2130: why is there a `Diffbased` prefix? It feels like this `DataStorageConfiguration` is doing more than one thing at the same time

besu/src/main/java/org/hyperledger/besu/cli/options/DataStorageOptions.java
```diff
@@ -52,45 +52,46 @@ public class DataStorageOptions implements CLIOptions<DataStorageConfiguration>
   private DataStorageFormat dataStorageFormat = DataStorageFormat.BONSAI;
 
   @Option(
-      names = {BONSAI_STORAGE_FORMAT_MAX_LAYERS_TO_LOAD, "--bonsai-maximum-back-layers-to-load"},
+      names = {DIFFBASED_STORAGE_FORMAT_MAX_LAYERS_TO_LOAD, "--bonsai-maximum-back-layers-to-load"},
```
lu-pinto, line 55: feels odd that the option is generic but then the cli option is specific to bonsai. Are you going to use the cli option to configure verkle too?

matkt: In the future, I will add a second option for Verkle, --verkle-maximum-back-layers-to-load, but in the end, it will be the same parameter in the code. I think for the user experience, it's better to have --verkle-maximum-back-layers-to-load and --bonsai-maximum-back-layers-to-load instead of --diffbased-maximum-back-layers-to-load because users don't know what that is.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/storage/BonsaiWorldStateKeyValueStorage.java
```diff
@@ -180,16 +177,17 @@ public void clear() {
   }
 
   @Override
-  public FlatDbStrategy getFlatDbStrategy() {
-    return flatDbStrategyProvider.getFlatDbStrategy(composedWorldStateStorage);
+  public BonsaiFlatDbStrategy getFlatDbStrategy() {
+    return (BonsaiFlatDbStrategy)
```
lu-pinto, line 181: suggestion: to avoid having to cast you could make `class FlatDbStrategy` -> `class FlatDbStrategy<T extends FlatDbMode>` or something and then `class BonsaiFlatDbStrategy<FlatDbMode.PARTIAL>`. You could then change `getFlatDbStrategy` return type to `FlatDbStrategy<? extends FlatDbMode>`.
But I'm not sure if that makes sense without getting deep into the code

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -2715,12 +2735,14 @@ private String generateConfigurationOverview() {
...
-          getDataStorageConfiguration().getBonsaiTrieLogPruningWindowSize());
+    if (DataStorageFormat.BONSAI.equals(getDataStorageConfiguration().getDataStorageFormat())) {
+      final DiffBasedSubStorageConfiguration subStorageConfiguration =
+          getDataStorageConfiguration().getDiffBasedSubStorageConfiguration();
+      if (subStorageConfiguration.getLimitTrieLogsEnabled()) {
+        builder.setLimitTrieLogsEnabled();
+        builder.setTrieLogRetentionLimit(subStorageConfiguration.getMaxLayersToLoad());
+        builder.setTrieLogsPruningWindowSize(subStorageConfiguration.getTrieLogPruningWindowSize());
+      }
     }
```
lu-pinto, line 2746: nit: just a suggestion - as you're only modifying the builder you could create a method called `ConfigurationOverviewBuilder.accept(DataStorageConfiguration)` where you move in all of this logic and maybe make setters private so that no one can do `setLimitTrieLogsEnabled` if not for a BONSAI storage.

matkt: The logic for modifying certain things only when in Bonsai mode is present in several places within BesuCommand. I noticed this while making my changes. I believe there should be another PR to possibly revisit the logic in BesuCommand at this level in order to avoid having `DataStorageFormat.BONSAI.equals(dataStorageConfiguration.getDataStorageFormat()` . I can try to create another PR to address this later, but I don't want to increase the size of this one

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/trie/diffbased/common/storage/flat/FlatDbStrategyProviderTest.java
```diff
@@ -43,8 +45,9 @@
 
 @ExtendWith(MockitoExtension.class)
 class FlatDbStrategyProviderTest {
```
lu-pinto, line 47: I think this class should be renamed too right? to `BonsaiFlatDbStrategyProviderTest` as you're only testing the Bonsai one

matkt: good catch I also need to move it

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServerTest.java
```diff
@@ -85,7 +85,8 @@ Bytes accountRLP() {
   // force a full flat db with code stored by code hash:
   final BonsaiWorldStateKeyValueStorage inMemoryStorage =
```
lu-pinto, line 86: nit: Something for a next PR and not this one - I would maybe prefer to see a faked/mocked/test version of `inMemoryStorage` instead of an implementation specific one because then we are kind of being biased towards that implementation and can clearly define dependencies

lu-pinto (approved):
Good stuff! really like the composition approach for the data storage.
I would just do the rename of the test but not strongly opinionated on the rest. LGTM

Issue #7605, opened 2024-09-11
EIP-7742 Uncouple blob count between CL and EL


daniellehrner:
Implement EIP-7742 for Pectra as specified here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7742.md

siladu:
@daniellehrner Are you planning to look at this yourself or does it need triaging?

daniellehrner:
The timeline is unclear on this. I think for the time being it's not urgent. So unfortunately I don't know yet if I'll work on it or not.

siladu:
Engine API changes: https://github.com/ethereum/execution-apis/pull/574

Incoming potential changes:
 - https://github.com/ethereum/EIPs/pull/8979

macfarla:
teku impl https://github.com/Consensys/teku/tree/electra-devnet-5

siladu:
This was removed from Pectra (essentially replaced by https://github.com/hyperledger/besu/issues/8026)

PR #7808, opened 2024-10-24
EIP-7742: Add target_blob_count to block header

siladu:
## PR description

First major piece of EIP-7742: Add target_blob_count to the block header.

Still todo in future PRs:
- Engine API
- Wiring engine API to block header
- block validation
- gas accounting 

## Fixed Issue(s)

Part of https://github.com/hyperledger/besu/issues/7605


acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/jsonrpc/ExecutionEnginePragueAcceptanceTest.java
```diff
@@ -20,8 +20,11 @@
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.params.provider.Arguments;
 
+// TODO SLD
+@Disabled("TODO SLD - Enable when Prague spec is finalized")
```
siladu, line 27: I propose we leave this disabled until the nearer the end of the Prague dev work. Anything that changes the block hash (genesis, block header, some engine API changes) causes a cascading change to all the JSON files and it's time consuming to update this with every PR. In the meantime, we will still have regression signal in the devnets and execution-spec/references tests.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/ProcessableBlockHeader.java
```diff
@@ -46,6 +47,8 @@ public class ProcessableBlockHeader
   protected final Bytes32 mixHashOrPrevRandao;
   // parentBeaconBlockRoot is included for Cancun
   protected final Bytes32 parentBeaconBlockRoot;
+  // TODO SLD Quantity or UInt64Value<UInt64> instead?
+  protected final UInt64 targetBlobCount;
```
siladu, line 51: Some other code for similar types use a subclass of BaseUInt256Value (which implements Quantity and UInt256Value). 
Withdrawals already uses UInt64 directly.

I think UInt64 is more direct than the tuweni generic base class construct and avoids yet another wrapper class.

Would like to get more opinions about the benefits of defining our own type, i.e. `TargetBlobCount extends BaseUInt256Value<TargetBlobCount> implements Quantity`

siladu (commented):
Added some of my own comments for the reviewer

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/ProcessableBlockHeader.java
siladu: Side note: Teku using UInt64 directly, as in this PR https://github.com/Consensys/teku/blob/110ce917f1fec942e2c2664c50a1bc3bb778ee07/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/schema/PayloadAttributesV4.java#L48

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/jsonrpc/ExecutionEnginePragueAcceptanceTest.java
jframe: I think that's fine, still it's a lot of effort to update. Once the testnets are done though we should re-enable this test

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/ProcessableBlockHeader.java
jframe: Fine with using UInt64 directly. Adding the new type doesn't add much value

config/src/main/java/org/hyperledger/besu/config/GenesisConfigFile.java
```diff
@@ -265,6 +265,17 @@ public String getParentBeaconBlockRoot() {
         "0x0000000000000000000000000000000000000000000000000000000000000000");
   }
 
+  /**
+   * Gets target blob count.
+   *
+   * @return the target blob count
+   */
+  public Optional<String> getTargetBlobCount() {
```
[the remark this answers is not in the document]
siladu: yes, [the EIP ](https://eips.ethereum.org/EIPS/eip-7742) says 
> For a genesis block with no existing parent, the value should be set according to the agreed specification for the target blob count given by that genesis block's protocol rule set.

which I have interpreted as pectra-at-genesis networks specifying an initial targetBlobCount.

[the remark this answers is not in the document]
jframe: I'm not sure how to interrupt that statement in the EIP, but until we know any different happy to go with this

jframe (approved):

PR #7809, opened 2024-10-24
EIP-7702: devnet-4 changes

daniellehrner:
## PR description

* Warm target of delegated account if a delegated account is the transaction destination
* restrict authorization nonce to 2**64 -1
* fix gas calculation for code delegation resolution if target is precompile
* verify auth nonce less than 2**64-1 during auth processing
* auth to zero address deletes delegation
* auth to precompile returns empty code
* auth nonce < 2**8
* increase auth base cost to 12500


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/CodeDelegationProcessor.java
```diff
@@ -89,6 +94,22 @@ private void processAuthorization(
       return;
     }
 
+    if (codeDelegation.nonce() == MAX_NONCE) {
```
Matilda-Clerke, line 97:
```suggestion
    if (codeDelegation.nonce() >= MAX_NONCE) {
```
Would make more sense to me, given the trace log this causes

Matilda-Clerke (commented):
LGTM but I'm not familiar with this code or the required changes for devnet-4, so leaving unapproved for now

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/CodeDelegationProcessor.java
daniellehrner: That check is a bit confusing, but is correct. `nonce` is a `long` and `MAX_NONCE` (which is 2^64-1) is the maximum value that a `long` can be. So `nonce` can never be greater than `MAX_NONCE`, because it would overflow to 0 in that case.

macfarla (approved):

Issue #7804, opened 2024-10-24
Besu returns empty trace on block 21030627


dylanbutler1:
**Actual behavior**
Execute
curl -H "Content-Type: application/json" -d '{"id": 1,"jsonrpc": "2.0","method": "trace_block","params":["0x140e6e3"]}' localhost:8545

Result:
{"jsonrpc":"2.0","id":1,"result":[]}

**Expected behavior**
Expected response is a list of traces.

cc @joshuafernandes

ahamlat:
@dylanbutler1  Which besu node are you using to test ? You need to have a Besu forest archive node otherwise you will get an empty response as a Besu Bonsai full node can trace by default `chain height - 512 blocks`. You can test again and if you see this log
Limit of historical layers that can be loaded with BONSAI ...
It means that you're using Bonsai and it can not perform a `trace_block` call on a block that is older than `chain height - 512`

racket2000:
For visibility, there was a similar issue affecting [Erigon](https://github.com/erigontech/erigon/issues/12432) as well 

I am not sure how to resolve this at the moment.

macfarla:
This is expected behaviour for bonsai (default). You need to have a Besu forest archive node otherwise you will get an empty response as a Besu Bonsai full node can trace by default chain height - 512 blocks. You can test again and if you see this log
`Limit of historical layers that can be loaded with BONSAI ...`
It means that you're using Bonsai and it can not perform a trace_block call on a block that is older than chain height - 512

ahamlat:
I tested with set_newHead on a bonsai node, to chain the head near the block we want to trace, but it didn't work, the node had few issues
```

{"@timestamp":"2024-10-25T19:06:55,322","level":"WARN","thread":"vert.x-worker-thread-3","class":"DefaultBlockchain","message":"Chain Reorganization +0 new / -13956 old
       Old - hash: 0x48a7ab9c62ce82c50c2809475c6ff7077a9577a889e53517991b54ba98785b00, height: 21044611
       New - hash: 0x7bc8e8f0f548333aae54bc5a061b0dc02d931221ac13d6f0e3278aacc25e94e2, height: 21030655
  Ancestor - hash: 0x7bc8e8f0f548333aae54bc5a061b0dc02d931221ac13d6f0e3278aacc25e94e2, height: 21030655","throwable":""}
[...]

PR #7821, opened 2024-10-28
feature: Add optional worldstate move flag to debug_setHead

garyschulte:
## PR description

This PR adds an optional parameter to `debug_setHead` that allows one to move the head worldstate along with the blockchain.  This is a no-op for Forest, but for DiffBased worldstates like bonsai and verkle, this will roll the head worldstate to the specified block number / block hash.

This is a debug method only and __should  not__ be used when a consensus client is directing besu, or while the node is actively importing or proposing blocks.

This feature is a stop-gap measure until #7475 is complete.  This debug feature will allow users to move bonsai worldstate head far back into block history so that expensive operations like `debug_traceBlock` can be run on historical blocks.  Otherwise, trying to execute RPC calls on old historical states will inevitably encounter an OOM (out of memory) while trying to roll the worldstate back in-memory.

This is a debug feature, if used incorrectly (like while some concurrent process is trying to move or modify the worldstate) it will corrupt the database.  

## Fixed Issue(s)

related to #7804 

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugSetHead.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugSetHead.java
@@ -75,7 +75,11 @@ public class DebugSetHead extends AbstractBlockParameterOrBlockHashMethod {
       if (archive.isWorldStateAvailable(blockHeader.getStateRoot(), blockHeader.getBlockHash())) {
         // WARNING, this can be dangerous for a DiffBasedWorldstate if a concurrent
         //          process attempts to move or modify the head worldstate.
-        //          for Foreset worldstates, this is essentially a no-op.
+        //          Ensure no block processing is occuring when using this feature.
+        //          No engine-api, block import, sync, or mining should be running.
+        //          Even p2p transaction processing will be very expensive and should be disabled.
+        //
+        //          for Forest worldstates, this is essentially a no-op.
         archive.getMutable(blockHeader.getStateRoot(), blockHeader.getBlockHash());
       }
     }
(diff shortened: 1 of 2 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugSetHead.java
```diff
@@ -45,26 +64,94 @@ public String getName() {
...
+      var archive = blockchainQueries.getWorldStateArchive();
+
+      // Only DiffBasedWorldState's need to be moved:
+      if (archive instanceof DiffBasedWorldStateProvider diffBasedArchive) {
+        rollWorldStateIncrementally(maybeBlockHeader.get(), blockchain, diffBasedArchive);
+      }
+    }
+
+    // finally move the blockchain
+    blockchain.rewindToBlock(maybeBlockHeader.get().getBlockHash());
```
garyschulte, line 99: The implication of moving the chain after moving the state is that we will not be able to move the worldstate FORWARD more than 512 blocks in the same call as the blockchain.  It would have to be split up into 2 calls, move the chain without worldstate, then move the chain with worldstate.

This is just due to the nature of doing the incremental state rolling.  The canonical chain needs to be at the target hash before we can do the incremental rolling.  Since there could be forks, and there isn't a canonical block number without the chain having moved first, we have to roll the chain forward first.   

We could flip this consideration and allow for arbitrary incremental forward rolling by moving the blockchain first, but that just flips the problem around such that we cannot move incrementally backward.  

This is less than ideal, but it is worth it to enable us to safely move the worldstate without exhausting the memory in an OOM and/or db transaction memory for long state moves.

if we move the chain by hash 512 or fewer blocks at a time, this is not a concern.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugSetHead.java
```diff
@@ -45,26 +64,94 @@ public String getName() {
...
 
     return JsonRpcSuccessResponse.SUCCESS_RESULT;
   }
+
+  private void rollWorldStateIncrementally(
```
matkt, line 104: maybe we can move this logic directly in the bonsai archive class ? I feel that this method can be used in other part of the code

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugSetHead.java
garyschulte: thinking about it today, we should be able to move the chain in the same increments as the worldstate and make this not a problem.  If they move in-tandem, then we should always be able to resolve where we are moving the worldstate to, forward or backward.

matkt (approved):

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugSetHead.java
garyschulte: added chain and worldstate incremental move.

Issue #7818, opened 2024-10-28
Plugins can no longer create new Metrics Categories


jflo:
### Steps to Reproduce (Bug)
1. Create a new plugin, which adds its own instance of MetricCategory to the MetricsCategoryRegistry service that is provided in the BesuContext at plugin registry time. Give it a unique name.
2. Deploy the plugin to any instance of Besu.
3. Startup the instance, with your new metrics category specified in the list of `--metrics-categories` to collect data on.

**Expected behavior:** new metrics are produced under the new category.

**Actual behavior:** Startup fails, because there is no MetricCategoryConverter for the string provided as an option to `--metrics-categories`. Plugin registration happens AFTER those converters are setup and used to parse the supplied `--metrics-categories` options.

**Frequency:** always

### Logs (if a bug)
Please post relevant logs from Besu (and the consensus client, if running proof of stake) from before and after the issue. 

### Versions (Add all that apply)
* Software version: 24.10.0+
* Java version: 21

jflo:
still not working after retesting with this https://github.com/Consensys/besu-plugins/pull/new/metricsCategoryDemo

jflo:
retested correctly, works fine.

PR #7825, opened 2024-10-29
Fix registering new metric categories from plugins

fab-10:
## PR description

This PR fixes an issue with new metric categories registered by plugins and enabled via `metrics-categories` option, since when the plugin registers them it is already too late in the startup process, and Besu fails reporting an unknown metric category option.
To fix this, the validation of the metric categories have been postponed to the moment we build the metric configuration from the metrics options, because at that point all is set.


jflo:
fixes #7818

jflo (approved):
great move getting rid of the need for converters

PR #7840, opened 2024-10-31
EIP-7742 Engine API Changes

siladu:
## PR description

**This makes Pectra require EIP-7742 fields so will break any pectra devnets that exclude 7742.** @daniellehrner will that be an issue? I could temporarily disable the engine api validation to avoid this if so.

This makes Besu usable for interop (has been tested with lodestar so far), however there are still some missing pieces that may cause 7742 to break depending on how maxBlobsPerBlock is set:
* The max isn't wired through to block creation so that still assumes Cancun's 6 blobs max.
* The 4844 block/tx max blob validation (via newPayload) hasn't been skipped for Prague so a newPayload with > 6 blobs may fail

In other words, in order for Besu to work for a Pectra interop we would require a EIP-7742 compatible CL and a max blob value of <= 6.

This PR was big enough already and I think uncontentious, whereas the implementations for the points above may require more discussion.

Also block creation wiring was required in this PR in order to satisfy CodeDelegationTransactionAcceptanceTest

Part of https://github.com/hyperledger/besu/issues/7605

Spec: https://github.com/ethereum/execution-apis/pull/574

### DIFF
--- a/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/PayloadIdentifier.java
+++ b/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/PayloadIdentifier.java
@@ -63,7 +63,8 @@ public class PayloadIdentifier implements Quantity {
    * @param feeRecipient the fee recipient
    * @param withdrawals the withdrawals
    * @param parentBeaconBlockRoot the parent beacon block root
-   * @param targetBlobsPerBlock the target blobs per block
+   * @param targetBlobsPerBlock the target blobs per block EIP-7742
+   * @param maxBlobsPerBlock the max blobs per block EIP-7742
    * @return the payload identifier
    */
   public static PayloadIdentifier forPayloadParams(
@@ -73,9 +74,10 @@ public class PayloadIdentifier implements Quantity {
       final Address feeRecipient,
       final Optional<List<Withdrawal>> withdrawals,
       final Optional<Bytes32> parentBeaconBlockRoot,
-      final Optional<UInt64> targetBlobsPerBlock) {
+      final Optional<UInt64> targetBlobsPerBlock,
+      final Optional<UInt64> maxBlobsPerBlock) {
 
-    return new PayloadIdentifier(
+    long payloadId =
         timestamp
             ^ ((long) parentHash.toHexString().hashCode()) << 8
             ^ ((long) prevRandao.toHexString().hashCode()) << 16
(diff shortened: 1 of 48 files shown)


consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/PayloadIdentifier.java
```diff
@@ -71,23 +72,26 @@ public static PayloadIdentifier forPayloadParams(
...
-            ^ ((long) parentBeaconBlockRoot.hashCode()) << 40);
+                    withdrawals
+                        .map(
+                            ws ->
+                                ws.stream()
+                                    .sorted(Comparator.comparing(Withdrawal::getIndex))
+                                    .map(Withdrawal::hashCode)
+                                    .reduce(1, (a, b) -> a ^ (b * 31)))
+                        .orElse(0)
+                << 32
```
siladu, line 92: Note, added a shift for withdrawals to keep the hashcodes evenly distributed by using the same pattern as the other fields.

daniellehrner:
@siladu Breaking existing testnets isn't an issue. Mekong will not be update, it's spec stays frozen, and devnet-4 has been shut down already

consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/PayloadIdentifier.java
macfarla: is it worth extracting this to a method? it's a bit hard to read indented over so many lines

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineForkchoiceUpdated.java
```diff
@@ -318,6 +320,10 @@ private void logPayload(final EnginePayloadAttributesParameter payloadAttributes
               .setMessage(message)
               .addArgument(() -> payloadAttributes.getParentBeaconBlockRoot().toHexString());
     }
+    if (payloadAttributes.getTargetBlobsPerBlock() != null) {
+      message += ", targetBlobsPerBlock: {}";
+      builder = builder.setMessage(message).addArgument(payloadAttributes::getTargetBlobsPerBlock);
+    }
```
macfarla, line 326: why not log getMaxBlobsPerBlock here too

siladu:
Removed from Pectra

PR #7842, opened 2024-10-31
Feat/contract info

FlorianHuc:
## PR description

Add info related to contracts created before London with nonce 0

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/frame/MessageFrame.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/frame/MessageFrame.java
@@ -246,6 +246,9 @@ public class MessageFrame {
   /** The mark of the undoable collections at the creation of this message frame */
   private final long undoMark;
 
+  /** Field to collect extra information while the frame is processed. */
+  private List<String> extraData;
+
   /**
    * Builder builder.
    *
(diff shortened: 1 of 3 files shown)


evm/src/main/java/org/hyperledger/besu/evm/frame/MessageFrame.java
```diff
@@ -1169,6 +1172,27 @@ public Optional<ExceptionalHaltReason> getExceptionalHaltReason() {
     return exceptionalHaltReason;
   }
 
+  /**
+   * Gets extra data colelcted during the frame procesing.
```
macfarla, line 1176:
```suggestion
   * Gets extra data collected during the frame processing.
```

evm/src/main/java/org/hyperledger/besu/evm/processor/ContractCreationProcessor.java
```diff
@@ -87,6 +88,17 @@ private static boolean accountExists(final Account account) {
...
   }
 
+  /**
+   * check is it is empty as per London contract spec strictly.
+   *
+   * @param account to check
+   * @return true if it has a nonce == 0, empty code and non-empty storage. It means it's non-empty
+   *     and was created before London.
+   */
+  public static boolean contractHasNonceZeroEmptyCodeAndNonEmptyStorage(final Account account) {
```
macfarla, line 98: needs a unit test

macfarla (commented):
@FlorianHuc is there a spec or similar for this change? is this for linea to interop with geth?

FlorianHuc:
Hi @macfarla we have discussed this PR with Fabio and I'll do differently. Closing it. Thank you.

PR #7845, opened 2024-10-31
Bonsai accounts don't have equals() or hashCode() implementations

matthew1001:
## PR description
Unlike `StateTrieAccountValue`, `BonsaiAccount` is missing overrides for `equals()` and `hashCode()`. This is causing unnecessary entries in the trie logs for accounts that have been loaded but which haven't changed.

An example is invoking a method that only emits an event but which doesn't update storage. The account is loaded, the method is invoked, but nothing in the account has changed. It is then (unnecessarily) recorded in the trie-log for that block.


garyschulte:
We might actually rely on this unintended behavior for linea-shomei zk-trielog plugin.  The primary difference between the TrieLogFactoryImpl and ZkTrieLogFactory implementation is that we add block number and zero reads to the trielog.  But the zk state needs to know about updates and reads.  

The value of having reads in the trielogs goes beyond just having a diff to apply, but rather is a log of all state access as well as modifications.  This is useful for historical block tracing (very cheap historical/archive worldstate).

I don't know that adding equals and hashCode implementations will break this.  Before we merge we should probably either assert that either this doesn't break anything or make explicit provision for persisting reads (and zero reads) if we relying on unintended consequences of not having these methods
@matkt  for visibility

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiAccount.java
```diff
@@ -173,6 +173,22 @@ public String toString() {
         + '}';
   }
 
+  @Override
+  public boolean equals(final Object o) {
```
garyschulte, line 177: We should move as much of this up into DiffBasedAccount as possible so that other implementations inherit this (like Verkle).  perhaps implement equals in DiffBasedAccount and delegate to an abstract method that implementations are forced to provide.

matthew1001:
Thanks for the comments @garyschulte I did wonder if the existing behaviour was in any way intentional.

In the case I've given in the PR description, there isn't even a state access, just an event. I'm not sure how realistic that is but it does need considering in terms of whether we want try logs for those types of invocation.

If the answer is "this is intentional/needed/useful" then I'm happy to close the PR with that explanation. I think I'll need to take your call on whether that's the case?

matkt:
It's something we saw while coding Verkle, but we decided not to include it for now until we have time to test that it won't have any side effects on the trielog shipping part. I think before we merge it, we will need to conduct some tests.

matkt (changes requested):
we need more tests around trielog shipping and consensus parts before merge

matthew1001:
I've addressed the code comment re. moving most of the `equals()` implementation into `DiffBaseAccount`. Regarding any additional testing required to prove that the change hasn't regressed anything or caused problems for verkle implementation, I think I'll need to wait for feedback from @matkt

matthew1001:
@garyschulte nudging on this one in case you have a view on whether this is OK to merge?

matkt:
I haven't been able to test yet, I had a lot of work. If it's a blocking issue for you I can try to prioritize the test

matthew1001:
No massive hurry @matkt, just useful to know at some point if we think the direction is to remove non-changed accounts from trie logs or not

Issue #7852, opened 2024-11-04
Access account information without triggering changes


FlorianHuc:
### Description
In order to generate traces of the EVM, the arithmetization requires access to the accounts. 

Current available method:

```
 MutableAccount getAccount(Address address);
```
  
gives the required information, but the implementation in the abstractWolrdUpdated as a side effect adds it to the updatedAccounts list which ends in state modification later during the processing. To avoid this unwanted modification, we need a method returning the same information without the side effect.

### Acceptance Criteria
* new method available to get data without tracking the account.

PR #7853, opened 2024-11-04
Add feature in to retrieve account info without tracking it

FlorianHuc:
## PR description
Add methods getAccountNoTrack in  WorldUpdater that returns account details but don't add it for tracking to avoid states changes.

## Fixed Issue(s)
fixes [7852](https://github.com/hyperledger/besu/issues/7852)


matkt:
I think we need to check but If I remember correctly https://github.com/hyperledger/besu/blob/main/evm/src/main/java/org/hyperledger/besu/evm/worldstate/WorldView.java#L32 is not tracking account. Maybe you can use this method instead of getAccount

FlorianHuc:
> I think we need to check but If I remember correctly https://github.com/hyperledger/besu/blob/main/evm/src/main/java/org/hyperledger/besu/evm/worldstate/WorldView.java#L32 is not tracking account. Maybe you can use this method instead of getAccount

it's indeed what i need. thank you.

PR #7864, opened 2024-11-06
adds action to manage stale issues

jflo:
## PR description

Adopts https://github.com/marketplace/actions/close-stale-issues action to manage old issues and prs.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,27 @@
+name: Close inactive issues
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


garyschulte (approved):
🏅

garyschulte:
One question - what does this acton do to re-opened issues?  does it restart the clock at the point of re-open?

jflo:
> One question - what does this acton do to re-opened issues? does it restart the clock at the point of re-open?

I'm assuming that is considered an update, and the remove-stale-when-updated flag being true would cause that to reset the clock.

PR #7869, opened 2024-11-07
Move EIP-4762 gas charging for contract account creation from code su…

lu-pinto:
## PR description

This fixes gas charges when creating contracts. While other clients charge the contract account creation while writing the nonce, besu was charging when the contract code was written to storage. A mismatch occurs if the code fails to deploy but the nonce is still written. In that case BASIC_DATA and CODE_HASH (empty contract hashcode) are still written to, despite code not being deployed.

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/gascalculator/stateless/Eip4762AccessWitness.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/gascalculator/stateless/Eip4762AccessWitness.java
@@ -40,11 +40,6 @@ public class Eip4762AccessWitness implements AccessWitness {
 
   private static final Logger LOG = LoggerFactory.getLogger(Eip4762AccessWitness.class);
   private static final TrieKeyAdapter TRIE_KEY_ADAPTER = new TrieKeyAdapter(new PedersenHasher());
-  private static final long WITNESS_BRANCH_COST = 1900;
-  private static final long WITNESS_CHUNK_COST = 200;
-  private static final long SUBTREE_EDIT_COST = 3000;
-  private static final long CHUNK_EDIT_COST = 500;
-  private static final long CHUNK_FILL_COST = 6200;
 
   private static final UInt256 zeroTreeIndex = UInt256.ZERO;
   private final Map<LeafAccessKey, Integer> leaves;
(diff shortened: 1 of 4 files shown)


evm/src/main/java/org/hyperledger/besu/evm/gascalculator/stateless/Eip4762AccessWitness.java
```diff
@@ -288,6 +288,7 @@ public long touchAddressAndChargeGas(
       final int accessMode) {
     final short accessEvents = touchAddress(address, treeIndex, subIndex, accessMode);
     long gas = 0;
+    LOG.atDebug().log(">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>");
```
matkt, line 291: I think we can remove it

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/stateless/Eip4762AccessWitness.java
```diff
@@ -368,7 +369,7 @@ public long touchAddressAndChargeGas(
                   + " "
                   + gasView);
     }
-
+    LOG.atDebug().log("<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<");
```
matkt, line 372: same here

evm/src/main/java/org/hyperledger/besu/evm/processor/ContractCreationProcessor.java
```diff
@@ -101,7 +102,17 @@ public void start(final MessageFrame frame, final OperationTracer operationTrace
...
       long statelessGasCost =
           evm.getGasCalculator().proofOfAbsenceCost(frame, contract.getAddress());
+      if (handleInsufficientGas(
+          frame,
+          statelessGasCost,
+          () ->
+              String.format(
+                  "Not enough gas to cover proof of absence fee for %s: remaining gas = %d < %d = creation fee",
+                  frame.getContractAddress(), frame.getRemainingGas(), statelessGasCost))) {
+        return;
```
matkt, line 112: do we need to return or set the state of the frame as we do here ?
frame.setExceptionalHaltReason(Optional.of(ExceptionalHaltReason.ILLEGAL_STATE_CHANGE));
frame.setState(MessageFrame.State.EXCEPTIONAL_HALT);

lu-pinto: is it Illegal state change? I think it is insufficient gas. It's already done inside `handleInsufficientGas`

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/stateless/Eip4762AccessWitness.java
lu-pinto: To me it is useful for debugging because it tells me what has been charged in a single touch call, because we could have multiple charges in one call.
Maybe we can do a pretty print at the end to show this information at once instead. I'll have a look

PR #7871, opened 2024-11-09
Refactor Transaction rlp encoding

Gabriel-Trintinalia:
## PR Description

This pull request introduces refactoring to simplify the encoding and decoding of transactions by utilizing a new class `RlpProvider`. The changes replace the direct usage of `TransactionEncoder` and `TransactionDecoder` with methods provided by `RlpProvider`.

Key updates include decoupling the encoding and decoding logic from the `Transaction` class itself, resulting in the removal of the `readFrom()` and `writeTo()` methods from the `Transaction` class.

Static calls for encoding and decoding have been replaced with calls to a singleton object, which can be configured to use different encoding classes if necessary. The next pull request will extend the RLP service to allow plugins to register different transaction encoding and decoding objects. This will enable networks with different RLP serialisation to replace the mainnet RLP serialisation.

Additionally, this pull request removes the encoding context from the transaction decoders/encoders to align them with other decoders. Some code cleanup was also performed by utilising the `encoded()` method of the `Transaction` class.

### Key Changes:

- **Refactoring to use `RlpProvider`:** Replaced `TransactionEncoder.encodeOpaqueBytes` and `TransactionDecoder.decodeOpaqueBytes` with `RlpProvider.transaction().encodeOpaqueBytes` and `RlpProvider.transaction().decodeOpaqueBytes` respectively in various files.

- **Decoupling from `Transaction` class:** Removed `readFrom()` and `writeTo()` methods from the `Transaction` class.

- **Singleton Object for Encoding/Decoding:** Replaced static calls with calls to a singleton object, allowing for configurable encoding classes.

- **Code Cleanup:** Utilized the `encoded()` method of the `Transaction` class for cleaner code.

**Future Extensions:**
* Prepared the RLP service for future extensions to allow plugins to register different transaction encoding and decoding objects.

- Use the same approach for `Block` and `Blockhead`


jflo (approved):
Thanks, this will help future Daggerization

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/registry/RlpRegistry.java
```diff
@@ -0,0 +1,76 @@
...
+  private TransactionEncoder pooledTransactionEncoder;
+
+  private RlpRegistry() {
+    transactionDecoder = new MainnetTransactionDecoder();
+    pooledTransactionDecoder = new PooledMainnetTransactionDecoder();
+    transactionEncoder = new MainnetTransactionEncoder();
+    pooledTransactionEncoder = new PooledMainnetTransactionEncoder();
+  }
+
+  protected static RlpRegistry getInstance() {
```
jframe, line 38: Why does this need to be a singleton? It's only usage is from the RlpProvider.

Gabriel-Trintinalia: The next pull request will extend the RLP service to allow plugins to register different transaction encoding and decoding objects

ahamlat (changes requested):
Can you test on mainnet with parallel transaction execution enabled ?

Issue #6613, opened 2024-02-26
Private blockchain stalling when less than 1/3 of validators are down - BftProcessorExecutor-QBFT-0 | INFO  | ProposalValidator | Invalid Proposal Payload: Latest Prepared Metadata blockhash does not align with proposed block


kkeaton-rcp:
### Description
I have a private blockchain with 7 validators, I take down 2 validators and the chain seems to freeze and is unable to produce blocks. 

### Acceptance Criteria
* Blockchain is able to handle up to 1/3 of the validators going down (should be able to handle 2 validators going offline) 

### Steps to Reproduce (Bug)
I have tested this a few times and it is inconsistent. Sometimes the chain does not stall, other times the chain does stall.

Shut down two validators and then trigger events to happen on the blockchain.

**Expected behavior:** [What you expect to happen]
Blockchain works with 2 validators down. 

**Actual behavior:** [What actually happens]
The blockchain will stall and gives `BftProcessorExecutor-QBFT-0 | INFO  | ProposalValidator | Invalid Proposal Payload: Latest Prepared Metadata blockhash does not align with proposed block` errors along with `EthScheduler-Timer-0 | DEBUG | WaitForPeerTask | Waiting for new peer connection. 9 peers currently connected.`

**Frequency:** [What percentage of the time does it occur?]
This happens about half the times I have run testing.

### Logs (if a bug)
```
2024-02-23 20:03:53.753+00:00 | nioEventLoopGroup-3-6 | DEBUG | EthPeer | handleDisconnect - EthPeer PeerId 0x9ee..., reputation PeerReputation 109, validated? true, disconnected? true, client: besu/v23.10.0/linux-x86_64/openjdk-java-17, connection [Connection with hashCode 131.... with peer 0x9ee... inboundInitiated false initAt 1708716292348], enode enode://9ee...
2024-02-23 20:03:53.753+00:00 | nioEventLoopGroup-3-6 | DEBUG | EthPeers | Disconnected EthPeer PeerId 0x9ee..., reputation PeerReputation 109, validated? true, disconnected? true, client: besu/v23.10.0/linux-x86_64/openjdk-java-17, connection [Connection with hashCode 1311626227 with peer 0x9ee... inboundInitiated false initAt 170...], enode enode://9e...
[...]

Issue #6053, opened 2023-10-18
BftProcessorExecutor-QBFT-0 | ERROR | AbstractBlockProcessor | failed persisting block org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Busy


kkeaton-rcp:
### Description
I have a private blockchain where all the validator nodes crashed and stopped processing blocks. They all seemed to have the error(s) attached below. This is extremely similar to [https://github.com/hyperledger/besu/issues/5441] which happened on besu version 23.4.0. However, the solution says it was an underlying hardware issue which I feel is unlikely as this happened to my nodes running on AWS. Any further discussion on this issue would be greatly appreciated, thank you!

### Steps to Reproduce
I am unsure how to reproduce this error as it seems to be an error with rocksdb. Using v23.1.2, ubuntu 22.04 running on t2.large aws instances.

**Expected behavior:** 
Expected the system to fix the error and start processing blocks again. 

**Actual behavior:** 
Error continues and seems to cause additional errors for the blockchain that cause the blockchain to stop processing blocks.

**Frequency:** 
This has only happened once so far on October 12th. 

### Versions 
Software version: besu v23.1.2
Java version: 17.0.8.1
OS Name & Version: Ubuntu 22.04.2 LTS
Kernel Version: x86_64
Cloud VM, type, size: AWS t2-large

### Logs
```
2023-10-12 00:13:21.013+00:00 | BftProcessorExecutor-QBFT-0 | ERROR | AbstractBlockProcessor | failed persisting block
org.hyperledger.besu.plugin.services.exception.StorageException: org.rocksdb.RocksDBException: Busy
        at org.hyperledger.besu.plugin.services.storage.rocksdb.segmented.RocksDBColumnarKeyValueStorage$RocksDbTransaction.commit(RocksDBColumnarKeyValueStorage.java:408)
        at org.hyperledger.besu.services.kvstore.SegmentedKeyValueStorageTransactionTransitionValidatorDecorator.commit(SegmentedKeyValueStorageTransactionTransitionValidatorDecorator.java:59)
        at org.hyperledger.besu.services.kvstore.SegmentedKeyValueStorageAdapter$1.commit(SegmentedKeyValueStorageAdapter.java:137)
        at org.hyperledger.besu.ethereum.storage.keyvalue.WorldStateKeyValueStorage$Updater.commit(WorldStateKeyValueStorage.java:216)
[...]

PR #7875, opened 2024-11-11
QBFT: Fix validation of proposal based on older round's prepared block

pullurib:
## PR description

From the logs, this particular occurence of the bug happened when transaction selection reached the time limit and then the round ran out of time before commit.

```
14:35:35.191+00:00 | BftProcessorExecutor-QBFT-0 | WARN  | BlockTransactionSelector | Interrupting transaction selection since it is taking more than the max configured time of 3000ms
14:35:35.191+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | TransactionSelectionResults | Selection stats: Totals[Evaluated=9, Selected=9, NotSelected=0, Discarded=0]; Detailed[]
```

 

There was a block proposed and prepared but not committed so it was propagated to new round . 

```
14:35:36.512+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftRound | Sending prepare message. round=ConsensusRoundIdentifier{Sequence=804237, Round=1}
14:35:38.468+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftRound | Received a prepare message. round=ConsensusRoundIdentifier{Sequence=804237, Round=1}. author=0x7f8843b0d49735fa174f202e7f18cc5f3142625f
14:35:39.957+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftRound | Received a prepare message. round=ConsensusRoundIdentifier{Sequence=804237, Round=1}. author=0xbd8c15e765f79f9fb1c5944e29ffcdb964170528
14:35:39.957+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftRound | Sending commit message. round=ConsensusRoundIdentifier{Sequence=804237, Round=1}
14:35:39.961+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftRound | Received a commit message. round=ConsensusRoundIdentifier{Sequence=804237, Round=1}. author=0xbd8c15e765f79f9fb1c5944e29ffcdb964170528
14:35:40.189+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftBlockHeightManager | Round has expired, creating PreparedCertificate and notifying peers. round=ConsensusRoundIdentifier{Sequence=804237, Round=1}
```

The new proposer identifies the Round Change (RC) message with the highest prepared round and bases a new proposal on the attached prepared message for that highest round. The proposal’s block header reflects the current round number, but the prepared message still references the prior round. In the code, an attempt is made to match the proposal’s block header round with the highest prepared round for validation by comparing the hashes of the proposed block and the underlying prepared block. However, the hashing method in BlockHeaderFunctions used is incorrect, as it replaces the round with 0.

As a result, validators continue sending RC messages with their already prepared blocks, causing this error to repeat and stall the network, preventing consensus progress. This issue occurs only when the highest prepared round is greater than 0—meaning consensus has moved to round 2 or beyond, using a prepared message from round 1 or later.

```
14:37:34.878+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftBlockHeightManager | Received sufficient RoundChange messages to change round to targetRound=ConsensusRoundIdentifier{Sequence=804237, Round=5}
14:37:34.878+00:00 | BftProcessorExecutor-QBFT-0 | DEBUG | QbftRound | Sending proposal from PreparedCertificate. round=ConsensusRoundIdentifier{Sequence=804237, Round=5}
14:37:36.211+00:00 | BftProcessorExecutor-QBFT-0 | INFO  | ProposalValidator | Invalid Proposal Payload: Latest Prepared Metadata blockhash does not align with proposed block
```

So made change to use the right hashing method which doesn’t touch the round number. 

I was able to repro the error by forcing one of the nodes to not send commit message in round 0 and round 1 .

## Fixed Issue(s)
#6732 and may be #6613 and may be #6053


consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftRound.java
```diff
@@ -332,7 +332,7 @@ private void peerIsPrepared(final Prepare msg) {
   private void peerIsCommitted(final Commit msg) {
     final boolean wasCommitted = roundState.isCommitted();
     roundState.addCommitMessage(msg);
-    if (wasCommitted != roundState.isCommitted()) {
+    if (!wasCommitted && roundState.isCommitted()) {
```
matthew1001, line 335: Is this logic identical to the previous logic, but you changed it to be clearer?

matthew1001: If so that's fine, just checking

matthew1001 (commented):
Is it easy to write a test for this? Would be nice to add a regression test if possible. Will add an approval once I've ready through the `forCommittedSeal` vs `forOnchainBlcok` code a bit more thoroughly.

matthew1001:
Just reading through the code and the detailed explanation you've outlined in the PR, the one thing I wanted to check related to this:

> The new proposer identifies the Round Change (RC) message with the highest prepared round and bases a new proposal on the attached prepared message for that highest round. The proposal’s block header reflects the current round number, but the prepared message still references the prior round.

Your fix sounds like it address this by effectively ignoring the round number. I'm wondering if there's an alternative fix which is to make sure the round number in the proposal message is correct for the new round?

@jframe I wonder if you've got time to take a look at this PR that Bhanu has raised to see what you think?

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftRound.java
pullurib: This correctly captures false to true transition of committed condition. I saw that blocks were committed without commit quorum before this change. I'll add more details about this change.

pullurib:
> Is it easy to write a test for this? Would be nice to add a regression test if possible. Will add an approval once I've ready through the `forCommittedSeal` vs `forOnchainBlcok` code a bit more thoroughly.

Yes I was thinking about the test , will check if it's relatively easy to add one.

pullurib:
> Just reading through the code and the detailed explanation you've outlined in the PR, the one thing I wanted to check related to this:
> 
> > The new proposer identifies the Round Change (RC) message with the highest prepared round and bases a new proposal on the attached prepared message for that highest round. The proposal’s block header reflects the current round number, but the prepared message still references the prior round.
> 
> Your fix sounds like it address this by effectively ignoring the round number. I'm wondering if there's an alternative fix which is to make sure the round number in the proposal message is correct for the new round?
> 
> @jframe I wonder if you've got time to take a look at this PR that Bhanu has raised to see what you think?

That's true, I got it incorrect about inclusion of round number . Currently (before this PR) prepared block and proposal both don't include round number in hash calculation . So , as far as validation goes , it should only be checked if prepared and proposed blocks hashes are equal (without any round number replacement) , which happens to be the case based on my experimentation. Prepared round and proposal round info are in metadata .

consensus/qbft/src/main/java/org/hyperledger/besu/consensus/qbft/statemachine/QbftRound.java
pullurib: reverted this change, don't see any issue now . Will fix it in a different PR if needed

pullurib:
Made changes to fix the other way as proposed above - adding round numbers to new block proposals and  updating the round number proposals based on prepared blocks to current round. updates are made in the block header .

pullurib:
Tested and verified the fix manually , will continue looking into adding a testcase

pullurib:
Figured that there's already a testcase which could have caught this error - https://github.com/hyperledger/besu/blob/40b27a15d2c4ee2f4f8442dbcc6da2c0d9a9c051/consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/test/RoundChangeTest.java#L172

It throws the validation error in logs 

```jsx

2024-11-14 16:02:16.999-05:00 | Test worker | INFO  | RoundChangeManager | Address: 0xb02c0093e94b417620f24d8857d6a1dd3d230272  Round: 4
2024-11-14 16:02:17.013-05:00 | Test worker | INFO  | RoundTimer | BFT round 3 expired. Moved to round 4 which will expire in 192 seconds
2024-11-14 16:02:17.029-05:00 | Test worker | INFO  | ProposalValidator | Invalid Proposal Payload: Latest Prepared Metadata blockhash does not align with proposed block
Task :consensus:qbft:integrationTest 
```

but doesn’t fail due to another bug in the code - while proposer handles their own proposal, it doesn’t check if the proposal is valid before sending prepare. Just fixing that made the above test fail. Then included the change to add round number in block header extradata, corresponding test case changes and verified all the testcases in consensus/qbft/src/integration-test/java/org/hyperledger/besu/consensus/qbft/test pass.

matthew1001:
This feels like a better fix to me @pullurib, thanks for tweaking it and for the extra tests.

matthew1001 (approved):

macfarla:
Since this fixes a number of issues, think it's worth adding a changelog entry @pullurib

pullurib:
> Since this fixes a number of issues, think it's worth adding a changelog entry @pullurib

Done, thanks

PR #7877, opened 2024-11-13
Consolidate all metric collectors into the metrics system

fab-10:
## PR description

~~Built on top of #7879~~ [MERGED]

This is the first PR of a series to refresh the metrics system, to pay some tech debt, and to migrate to the new Prometheus Java lib to version 1.x, from the current legacy Simpleclient version 0.x.

Along with the upgrade to the new Prometheus lib, these other tasks are planned to refresh the metrics system:
- Refactor the way the observation mechanism works to simplify it and make it more OO
- Replace our HTTP server implementation with the provided one, so we can benefit from the new export formats (like protobuf) without having to re-implement them
- Use Prometheus to provide OpenTelemetry metrics, and so remove the dedicated OpenTelemetry deps and code
- Add support for Histogram and Summary metric types.

The scope of this first PR is to consolidate all metric collectors in the metrics system, since with the time, something has spilled out of it, in particular RocksDB internal metrics and Guava cache metrics collectors were created outside the metric system, with the assumption to only support Prometheus as backend.

Both those collectors have been moved into the metrics system, and the ability to add custom collectors from outside the metrics system has been removed.

There are no functional changes or fixes in this PR only refactoring, so no support for the refactored collectors have been implemented for OpenTelemetry, since there was no support before, and since the idea is to use Prometheus to provide OpenTelemetry in a future PR.

### DIFF
--- a/metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusMetricsSystem.java
+++ b/metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusMetricsSystem.java
@@ -18,6 +18,7 @@ import org.hyperledger.besu.metrics.ObservableMetricsSystem;
 import org.hyperledger.besu.metrics.Observation;
 import org.hyperledger.besu.metrics.StandardMetricCategory;
 import org.hyperledger.besu.metrics.noop.NoOpMetricsSystem;
+import org.hyperledger.besu.plugin.services.metrics.ExternalSummary;
 import org.hyperledger.besu.plugin.services.metrics.LabelledGauge;
 import org.hyperledger.besu.plugin.services.metrics.LabelledMetric;
 import org.hyperledger.besu.plugin.services.metrics.MetricCategory;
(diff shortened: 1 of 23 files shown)


atoulme (approved):
thanks, that sounds like a good development

metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusMetricsSystem.java
```diff
@@ -170,7 +181,61 @@ public void createGauge(
...
+      registerCollector(category, collector);
+    }
+  }
+
+  @Override
+  public void trackExternalSummary(
+      final MetricCategory category,
+      final String name,
+      final String help,
+      final Supplier<ExternalSummary> summarySupplier) {
```
jflo, line 193: I don't truly understand why we need Supplier/Collectors in this method, but I'm guessing it has to do with pluggable metrics?

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/MetricsSystem.java
```diff
@@ -159,4 +164,45 @@ default void createLongGauge(
       final LongSupplier valueSupplier) {
     createGauge(category, name, help, () -> (double) valueSupplier.getAsLong());
   }
+
+  /**
+   * Track a summary that is computed externally to this metric system. Useful when existing
```
jflo, line 169: Thanks for elaborating in this comment, that answers my previous question :)

jflo (approved):

PR #7882, opened 2024-11-15
narrows dependencies of BlockchainService

jflo:
## PR description

Blockchain service does not need the entire ProtocolContext, just the Blockchain. Narrowing this will help us avoid circular dependencies.

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/services/BlockchainServiceImpl.java
+++ b/besu/src/main/java/org/hyperledger/besu/services/BlockchainServiceImpl.java
@@ -16,7 +16,6 @@ package org.hyperledger.besu.services;
 
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.datatypes.Wei;
-import org.hyperledger.besu.ethereum.ProtocolContext;
 import org.hyperledger.besu.ethereum.chain.MutableBlockchain;
 import org.hyperledger.besu.ethereum.core.Block;
 import org.hyperledger.besu.ethereum.mainnet.ProtocolSchedule;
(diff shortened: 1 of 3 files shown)


besu/src/main/java/org/hyperledger/besu/services/BlockchainServiceImpl.java
```diff
@@ -50,13 +48,12 @@ public BlockchainServiceImpl() {}
   /**
    * Initialize the Blockchain service.
    *
-   * @param protocolContext the protocol context
+   * @param blockchain the blockchain
    * @param protocolSchedule the protocol schedule
    */
-  public void init(final ProtocolContext protocolContext, final ProtocolSchedule protocolSchedule) {
-    this.protocolContext = protocolContext;
+  public void init(final MutableBlockchain blockchain, final ProtocolSchedule protocolSchedule) {
```
Matilda-Clerke, line 54: Any reason we are specifically requiring a MutableBlockchain and not just a Blockchain? It looks like all of the methods used below are present in Blockchain

Matilda-Clerke (approved):
LGTM, but please consider my comment regarding MutableBlockchain vs Blockchain

fab-10 (approved):

besu/src/main/java/org/hyperledger/besu/services/BlockchainServiceImpl.java
jflo: Sadly, this service does depend on the more specific MutableBlockchain in later, unmodified methods like storeBlock(), setFinalizedBlock() and setSafeBlock().

PR #7886, opened 2024-11-19
Pass correct updater to traceEndTransaction

Gabriel-Trintinalia:
## PR description
Provide the correct updater to `traceEndTransaction` to ensure consistency with other calls.


lu-pinto (approved):
LGTM, I actually think we should pass a fork of the `evmWorldUpdater` to every tracer call.
@daniellehrner since you made the changes to support account abstraction and `EvmWorldUpdater` was introduced there, it would be good to have your input

PR #7895, opened 2024-11-20
naming things is important

jflo:
BesuContext was not a helpful name for an interface that only manipulates services to be used by plugins.


plugin-api/src/main/java/org/hyperledger/besu/plugin/BesuContext.java
```diff
@@ -14,42 +14,6 @@
...
-   *
-   * @param serviceType the class defining the requested service.
-   * @param <T> the service type
-   * @return an optional containing the instance of the requested service, or empty if the service
-   *     is unavailable
-   */
-  <T extends BesuService> Optional<T> getService(Class<T> serviceType);
-}
+/** Deprecated in favor of the more precisely named ServiceManager interface. */
+@Deprecated
```
macfarla, line 18:
```suggestion
@Deprecated(since = "24.11.0", forRemoval = true)
```

macfarla (approved):

Gabriel-Trintinalia:
Yes, Besu will not startup: @jflo @matkt 

```
 java.lang.AbstractMethodError: Receiver class net.consensys.shomei.rpc.BesuShomeiRpcPlugin does not define or inherit an implementation of the resolved method 'abstract void register(org.hyperledger.besu.plugin.ServiceManager)' of interface org.hyperledger.besu.plugin.BesuPlugin.
besu-clique-verifier-1   | 	at org.hyperledger.besu.services.BesuPluginContextImpl.registerPlugin(BesuPluginContextImpl.java:208)
besu-clique-verifier-1   | 	at org.hyperledger.besu.services.BesuPluginContextImpl.registerPlugins(BesuPluginContextImpl.java:200)
besu-clique-verifier-1   | 	at org.hyperledger.besu.services.BesuPluginContextImpl.registerPlugins(BesuPluginContextImpl.java:154)
besu-clique-verifier-1   | 	at 
```

jflo:
yeah, the extension is backwards. BesuContext should be extending ServiceManger

PR #7896, opened 2024-11-21
Refactor Body Validation logic

Gabriel-Trintinalia:
## PR Description

This pull request introduces changes to the block validation process to improve its readability. The main change is that `BodyValidationMode` was moved to `validateBody` instead of `validateBodyLight` 

### Key changes include:

* Added a new validation mode `ROOT_ONLY` to the `BodyValidationMode` enum, which validates transactions, state, and receipts root without full validation. 
* Updated the `validateAndProcessBlock` method to include the new `BodyValidationMode.FULL` parameter for full body validation.
* Modified the `validateBlockForSyncing` method to throw an `UnsupportedOperationException` if `BodyValidationMode.FULL` is used, as full body validation is not supported for syncing blocks. 
* Added the `BodyValidationMode` parameter to the `validateBody` method and removed it from the `validateBodyLight` method.
* Refactored the `validateBody` method to handle different validation modes, including `NONE`, `LIGHT`, `ROOT_ONLY`, and `FULL`. 
* Consolidated the state root validation, transaction root validation, and receipts root validation into a new `validateBodyRoots` method for better code organisation and readability.

### Test Updates:
* Updated test cases to accommodate the validation mode and method signature changes. This includes adding new tests to ensure that the `UnsupportedOperationException` is thrown when full body validation is attempted during syncing.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -246,15 +251,24 @@ public boolean validateBlockForSyncing(
       final HeaderValidationMode headerValidationMode,
       final HeaderValidationMode ommerValidationMode,
       final BodyValidationMode bodyValidationMode) {
+
+    if (bodyValidationMode == BodyValidationMode.FULL) {
```
jframe, line 255: Why is full body validation not supported?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockBodyValidator.java
```diff
@@ -50,10 +50,39 @@ public boolean validateBody(
...
+        || bodyValidationMode == BodyValidationMode.FULL) {
+      return validateBodyRoots(block, receipts, worldStateRootHash);
+    }
+    return true;
+  }
+
+  @VisibleForTesting
+  protected boolean validateBodyRoots(
+      final Block block, final List<TransactionReceipt> receipts, final Hash worldStateRootHash) {
+    final BlockHeader header = block.getHeader();
```
jframe, line 76: Should withdrawals root also be included in this?

Gabriel-Trintinalia: withdrawals root is currently being validated into the light mode

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
Gabriel-Trintinalia: The only external path that calls `validateBodyLight` is `validateBlockForSyncing`, which will only be `NONE` or `LIGHT`, depending on PoS. In this path. As we do not have the `stateRoot`, I decided to enforce it so if something changes we will have to explicitly change this method.

jframe (approved):

PR #7914, opened 2024-11-25
Do not use the sync target to download the headers

pinges:
## PR description
The download headers step should be able to use any peer to download the headers


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/DownloadHeadersStep.java
```diff
@@ -99,7 +99,6 @@ private CompletableFuture<List<BlockHeader>> downloadHeaders(final SyncTargetRan
               range.getStart().getNumber(),
               headerRequestSize,
               metricsSystem)
-          .assignPeer(range.getSyncTarget())
```
jframe, line 102: For POS this is safe as the range start block number is used as the minimum required block. For non-pos chains not sure whether this is safe think might be an issue.

If you only skipped this when it's POS then I think it would be better.

jframe: Thinking through this more, it's already possible that other nodes might not have this header and we will fail to download the block. So I think this change is fine.

jframe (approved):

PR #7922, opened 2024-11-26
remove protocol name

### LINKED ISSUE #7810: Remove if not needed: getSupprtedProtocol() in BesuControllerBuilder always returns EthProtocol.NAME
### Description
Looks like this protocol name is mainly used in EthPeers and EthPeer, but as it is always the same it is probably not needed.

pinges:
## PR description
Removes the protocol name from EthPeers and EthPeer. The protocol name was always set to "eth".

Fixes #7810


fab-10 (approved):

macfarla (commented):
even SNAP peers were Eth protocol, right?

Issue #7835, opened 2024-10-31
plugin API - add tx permissioning hook


macfarla:
Similarly to `PermissioningService.registerNodePermissioningProvider()`
Need to add
`PermissioningService.registerTransactionPermissioningProvider()`

look at https://github.com/hyperledger/besu/blob/8ed4a904f15d67e67d34731ff95033f9fb2da369/ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/account/TransactionPermissioningProvider.java#L20 - the isPermitted() method there has the transaction object, which should have all the fields you need for tx permissioning v c

PR #7952, opened 2024-11-27
Add Transaction Permissioning Hook to PermissioningService Interface

vaidikcode:
Signed-off-by: vaidikcode <vaidikbhardwaj00@gmail.com>

## PR description
I’ve implemented the registerTransactionPermissioningProvider method to allow custom transaction permissioning logic by adding providers to a collection. A method to iterate over these providers and check if transactions are permitted has also been included.

Currently, other methods in the PermissioningService interface (like registerNodePermissioningProvider and registerNodeMessagePermissioningProvider) are left unimplemented, as they are not part of this PR's scope.

I may need guidance on whether you would like me to proceed with implementing the other methods, and any improvements or adjustments needed to integrate this into the broader permissioning system.
## Fixed Issue(s)

#7835 


macfarla (commented):
thanks for the contrib! 

a few comments. the logging would be better at info level. you can have extra logging in the plugin if needed, and/or you can use custom log4j config to turn this up for troubleshooting but don't want this permanently at info level. The one-off "registered plugin" is fine at info level.

there's also the TestPermissioningPlugin and PermissioningPluginTest - you could expand the test plugin and add a test to verify the tx permissioning callback works as expected

you'll also need to run the gradle target checkAPIChanges to calculate and update the api hash

and prob add a changelog entry

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/PermissioningService.java
```diff
@@ -14,6 +14,7 @@
  */
 package org.hyperledger.besu.plugin.services;
 
+import org.hyperledger.besu.ethereum.permissioning.account.TransactionPermissioningProvider;
```
macfarla, line 17: will need some refactoring to move this TransactionPermissioningProvider interface into the plugin package.

macfarla (changes requested):
will need some refactoring to move this TransactionPermissioningProvider interface into the plugin package.

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/PermissioningService.java
vaidikcode: Thank you for your feedback! Are you referring to this location: `plugin-api/src/main/java/org/hyperledger/besu/plugin/services/permissioning/`?

vaidikcode: I have a quick question: Since the TransactionPermissioningProvider is implemented by both the AccountLocalConfigPermissioningController and the TransactionSmartContractPermissioningController, what would be the best way to move forward with this? Your guidance would be greatly appreciated!

macfarla: I would have a look at how `NodeMessagePermissioningProvider` (functional interface) is used, since that's already exposed to plugins and used elsewhere in besu. It's a common pattern in besu to have an interface in the plugin-api package which is imported and implemented by classes where needed. `BlockHeader` is another good example but you can find lots of examples in the plugin-api package.

vaidikcode: There seems to be a circular dependency issue caused by importing the Transaction class into the TransactionPermissioningProvider interface.  introducing a lightweight abstraction (like a DTO or a subset of transaction properties) for permissioning logic might be a good solution. wdyt?

macfarla: does it help if the TransactionPermissioningProvider uses the org.hyperledger.besu.datatypes.Transaction interface instead

vaidikcode: Thank you very much. It works perfectly!.

macfarla:
@vaidikcode spotless!

macfarla:
@vaidikcode still getting compile errors - can you run compile and test tasks locally, your feedback loop will be waaaay faster

/Users/sm/workspace/b2/ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/TransactionSmartContractPermissioningController.java:205: warning: [rawtypes] found raw type: BaseUInt256Value
            .map(BaseUInt256Value::toBytes)
                 ^
  missing type arguments for generic class BaseUInt256Value<T>
  where T is a type-variable:
    T extends UInt256Value<T> declared in class BaseUInt256Value
error: warnings found and -Werror specified
1 error
1 warning

macfarla:
converting to draft as a signal it's not ready to review

ethereum/permissioning/src/main/java/org/hyperledger/besu/ethereum/permissioning/TransactionSmartContractPermissioningController.java
```diff
@@ -231,4 +237,16 @@ private static Bytes encodeBytes(final Bytes value) {
     final Bytes padding = Bytes.wrap(new byte[(32 - (value.size() % 32))]);
     return Bytes.concatenate(dynamicParameterOffset, length, value, padding);
   }
+
+  // Convert the Quantity value to Bytes
+  private static Bytes convertQuantityToBytes(final Quantity quantity) {
```
macfarla, line 242:
```suggestion
  private static Bytes encodeQuantity(final Quantity quantity) {
```

macfarla (commented):
couple of comments

also this test is failing when I run locally - PermissioningPluginTest.testGasLimitLogic

macfarla:
@vaidikcode couple of comments - it's getting there - but tests are still failing. Can you please run tests locally before saying it's ready to review? Your feedback loop will be like 100x faster

macfarla:
@vaidikcode checking if you're still planning on continuing working on this PR?

vaidikcode:
@macfarla , sorry for the delay. I won’t be able to work on this due to some reasons. Should I go ahead and close the PR?

PR #7965, opened 2024-12-02
Allow frontier simulation when Base Fee is present

Gabriel-Trintinalia:
## PR Description

This PR resolves an issue where simulating a legacy transaction in a post-frontier fork environment was impossible. The current logic defaults to using `maxFeePerGas` and `maxPriorityFeePerGas`, which incorrectly identifies the transaction type as EIP-1559.

### Changes in `TransactionSimulator`:
- Modified the `buildTransaction` method to conditionally set `maxFeePerGas` and `maxPriorityFeePerGas` only if they are provided. This ensures proper handling of both legacy transactions and EIP-1559 transactions.

### Changes in Test Cases:
- Introduced a new test case, `shouldSimulateLegacyTransactionWhenBaseFeeNotZero`, to confirm that the transaction simulator correctly processes legacy transactions when the base fee is not zero.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -437,18 +432,21 @@ private Optional<Transaction> buildTransaction(
       maxPriorityFeePerGas = callParams.getMaxPriorityFeePerGas().orElse(gasPrice);
       maxFeePerBlobGas = callParams.getMaxFeePerBlobGas().orElse(blobGasPrice);
     }
-    if (header.getBaseFee().isEmpty()) {
+
+    if (shouldSeGasPrice(callParams)) {
```
fab-10, line 436:
```suggestion
    if (shouldSetGasPrice(callParams)) {
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -489,4 +487,31 @@ public Optional<Boolean> doesAddressExist(
 
     return Optional.of(worldState.get(address) != null);
   }
+
+  private boolean shouldSeGasPrice(final CallParameter callParams) {
```
fab-10, line 491:
```suggestion
  private boolean shouldSetGasPrice(final CallParameter callParams) {
```

macfarla (commented):
changelog?

fab-10 (approved):

PR #7971, opened 2024-12-02
eip-7709 implement BLOCKHASH opcode from system contract state

### LINKED ISSUE #7811: Implement EIP-7709
EIP-7709: Read BLOCKHASH from storage and update cost

Implement BLOCKHASH operation by getting the hash of the block from system contract storage and stateless costs of touching the witness

More info in this link:
https://eips.ethereum.org/EIPS/eip-7709

lu-pinto:
## PR description
Refactor BlockHashLookup to store it in BlockHashProcessor so one can have different implementations for it. These changes are supposedly side-effect free so no changes to integration test assertions are required.
Also BlockHashLookup now takes the Frame so that it can lookup the hash in the system contract storage.
Implementation for EIP-7709 BlockHashLookup is provided but still with no updated costs.

Gas costs update and wiring will be done in the verkle branch.

## Fixed Issue(s)

fixes #7811


ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java
```diff
@@ -389,13 +388,28 @@ static T8nResult runTest(
...
+          // from them so need to
+          // add in a manual BlockHashLookup
+          blockHashLookup =
+              (__, blockNumber) -> {
+                if (referenceTestEnv.getNumber() - blockNumber > 256L
+                    || blockNumber >= referenceTestEnv.getNumber()) {
+                  return Hash.ZERO;
+                }
+                return referenceTestEnv.getBlockhashByNumber(blockNumber).orElse(Hash.ZERO);
+              };
```
Gabriel-Trintinalia, line 404: This happens because before the block range validation was done inside of the BlockHashOperation. This check was moved to the BlockHashLookup implementation so that is why we have to have the same block range check here. Is my understanding correct?

evm/src/main/java/org/hyperledger/besu/evm/blockhash/BlockHashLookup.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.hyperledger.besu.evm.blockhash;
+
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.evm.frame.MessageFrame;
+
+import java.util.function.BiFunction;
+
+/**
+ * Function that gets the block hash, passed in as part of TxValues.
```
Gabriel-Trintinalia, line 23:
```suggestion
 * Function that gets the block hash.
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/ContractBasedBlockHashLookup.java
```diff
@@ -0,0 +1,117 @@
...
+    this.historyServeWindow = historyServeWindow;
+    this.blockHashServeWindow = blockHashServeWindow;
+  }
+
+  @Override
+  public Hash apply(final MessageFrame frame, final Long blockNumber) {
+    final long currentBlockNumber = blockHeader.getNumber();
+    final long minBlockServe = Math.max(0, currentBlockNumber - blockHashServeWindow);
+    if (blockNumber >= currentBlockNumber || blockNumber < minBlockServe) {
+      LOG.debug("failed to read hash from system account for block {}", blockNumber);
```
Gabriel-Trintinalia, line 89: Maybe trace instead of debug?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/ContractBasedBlockHashLookup.java
```diff
@@ -0,0 +1,117 @@
...
+
+    final Hash cachedHash = hashByNumber.get(blockNumber);
+    if (cachedHash != null) {
+      return cachedHash;
+    }
+
+    final WorldUpdater worldUpdater = frame.getWorldUpdater();
+    Account account = worldUpdater.get(contractAddress);
+    if (account == null) {
+      LOG.error("cannot query system contract {}", contractAddress);
```
Gabriel-Trintinalia, line 101: This a valid scenario, correct? In case a network does not have this contract, should we log a warn/trace instead here?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/StateTestSubCommand.java
```diff
@@ -270,7 +270,10 @@ private void traceTestSpecs(final String test, final List<GeneralStateTestCaseEi
                 blockHeader,
                 transaction,
                 blockHeader.getCoinbase(),
-                blockNumber -> Hash.hash(Bytes.wrap(Long.toString(blockNumber).getBytes(UTF_8))),
+                (__, blockNumber) ->
+                    blockNumber >= blockHeader.getNumber()
+                        ? Hash.ZERO
+                        : Hash.hash(Bytes.wrap(Long.toString(blockNumber).getBytes(UTF_8))),
```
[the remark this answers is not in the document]
lu-pinto: it's because this logic was moved into `BlockHashLookup` now and no longer in `BlockHashOperation`

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/T8nExecutor.java
lu-pinto: yup, the same happens with https://github.com/hyperledger/besu/pull/7971/files/e310147d8efd8bf5115f231b050c2f6b20f1512a#r1878989411 but my comment is still correct. From what I remember the test is broken because the fixtures for the blockchain state are not valid. We should be able to build the state as these are integration tests and we have a fully working `BlockHashProcessor`. But I don't think there's a way to change them as they were generated with some tool and changing them would imply recalculating hashes, so we have to add a manual BlockHashLookup.

evm/src/main/java/org/hyperledger/besu/evm/blockhash/BlockHashLookup.java
lu-pinto: it's still passed as part of TxValues though: https://github.com/hyperledger/besu/blob/929945ad6c559d787a99c12e7f6c3dbcef6716f2/evm/src/main/java/org/hyperledger/besu/evm/frame/TxValues.java#L54

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/ContractBasedBlockHashLookup.java
lu-pinto: This would be a problem that we are not storing hashes in the system contract storage or we are storing in wrong account. I feel it should be an error as it will cause a consensus issue

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/ContractBasedBlockHashLookup.java
```diff
@@ -0,0 +1,117 @@
...
+   *
+   * @param currentBlock current block header being processed.
+   * @param contractAddress the address of the contract storing the history.
+   * @param historyServeWindow the number of blocks for which history should be saved.
+   * @param blockHashServeWindow the number of block for which contract can serve the BLOCKHASH
+   *     opcode.
+   */
+  @VisibleForTesting
+  ContractBasedBlockHashLookup(
+      final ProcessableBlockHeader currentBlock,
```
shemnon, line 74: Why pass the processable block header into the constructor? We have the MessageFrame and can get the current block number from that value.

Then we can re-use this object across multiple blocks, and even use it concurrently across multiple blocks.

shemnon (commented):
Just a sanity check, this isn't wired into any existing forks yet, right?  Is it still just a Verkle thing? Would it be better to work it out in the Verkle branch?

lu-pinto:
> Just a sanity check, this isn't wired into any existing forks yet, right? Is it still just a Verkle thing? Would it be better to work it out in the Verkle branch?

That's right, EIP is still in draft and is a hard dependency for Verkle. Don't think there's any case to have it sooner. We have a Verkle branch but have decided to start bringing stable parts of Verkle into `main` because of the high amount of conflicts while merging and time it takes to review these merges. But I'll defer to @matkt on that.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/ContractBasedBlockHashLookup.java
lu-pinto: In normal cases it should not occur as the `PragueBlockHashProcessor` that stores hashes in system contract storage should ship with Pectra and populate the contract storage within a day or so, but I would code defensively against it.

>- serve from memory or as per current designs if maintaining requisite history (full clients for e.g.)

There is always this option though. If I get that sentence right, it mentions that we could fallback to the previous approach of finding hashes from blockchain building. That was something I thought about actually but wasn't sure whether add the previous approach as fallback. But I agree with @shemnon, that this is not totally clear. By fatal, did you mean halting execution if this happens? I would agree with that

matkt:
> > Just a sanity check, this isn't wired into any existing forks yet, right? Is it still just a Verkle thing? Would it be better to work it out in the Verkle branch?
> 
> That's right, EIP is still in draft and is a hard dependency for Verkle. Don't think there's any case to have it sooner. We have a Verkle branch but have decided to start bringing stable parts of Verkle into `main` because of the high amount of conflicts while merging and time it takes to review these merges. But I'll defer to @matkt on that.

We're trying to merge things step by step into the main branch to reduce the size of the Verkle branch and also minimize futur conflicts, etc. The goal is to avoid having a branch that keeps growing if certain things can be merged without impacting Besu's codebase. Maybe we can add a comment to indicate that it's part of the Verkle work to prevent a future PR from trying to remove it as not used.

Gabriel-Trintinalia (approved):
lgtm

PR #7991, opened 2024-12-05
Fix possible incomplete txpool restore from dump file

fab-10:
## PR description

it could happen, especially at startup, that the node switch from in sync to out of sync quickly, and if that happens during the txpool restore from the dump file, the restore could be incomplete.

This is an log example of this issue:

```
1. Loading transaction pool content from file /data/besu/txpool.dump
2. Node out of sync, disabling transaction handling
3. Added 88 transactions of 442 loaded from file /data/besu/txpool.dump
4. Saving 88 transactions to file /data/besu/txpool.dump
5. Saved 88 transactions to file /data/besu/txpool.dump
6. Node is in sync, enabling transaction handling
7. Loading transaction pool content from file /data/besu/txpool.dump
8. Added 87 transactions of 88 loaded from file /data/besu/txpool.dump
```

So at the beginning the node was restoring the txpool (1.), 
but while this restore was still in progress, the node went out of sync and the txpool was disabled(2.), 
the issue manifest itself at this moment, since with the txpool disabled, restore txs are just ignored but technically the restore was completed successfully and the dump file removed, with only a subset of the txs restored (3.),
last part of disabling the txpool, consist in saving its content to the dump file (4.),
so the dump file is created with the current content of the txpool, that only has the subset of txs loaded from point 3 (5.)
Then after the node get in sync again, the process restart, but this time the dump file contains only a subset of the original txs set (6., 7., 8.)

To fix this issue, when disabling the txpool, we make sure that any in progress restore operation is cancelled correctly, removing from the dump file only the txs that have been restored, and changing the save operation to append to an existing dump file, so that no save txs are lost in case of a partial restore.

Below the new log with the fix
```
1. Loading transaction pool content from file /data/besu/txpool.dump
2. Node out of sync, disabling transaction handling
3. Added 94 transactions of 94 loaded from file /data/besu/txpool.dump, before operation was cancelled
4. Appending 94 transactions to file /data/besu/txpool.dump
5. Appended 94 transactions to file /data/besu/txpool.dump
6. Node is in sync, enabling transaction handling
7. Loading transaction pool content from file /data/besu/txpool.dump
8. Added 248 transactions of 348 loaded from file /data/besu/txpool.dump, deleting file
```

Note that restore operation is correctly stopped when the node goes out of sync (3.),
the save operation appends to the partially restored dump file (4., 5.)
and when the node goes in sync again the restore operation read the full set of txs from the dump, and deleted the file at the end (8.)

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
@@ -54,11 +54,11 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.IntSummaryStatistics;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
@@ -73,6 +73,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -645,15 +646,16 @@ public class TransactionPool implements BlockAddedObserver {
       isPoolEnabled.set(false);
       subscribeConnectId.ifPresent(ethContext.getEthPeers()::unsubscribeConnect);
       pendingTransactionsListenersProxy.unsubscribe();
-      final PendingTransactions pendingTransactionsToSave = pendingTransactions;
+      final CompletableFuture<Void> saveOperation =
+          saveRestoreManager
+              .saveToDisk(pendingTransactions)
+              .exceptionally(
+                  t -> {
+                    LOG.error("Error while saving transaction pool to disk", t);
+                    return null;
+                  });
       pendingTransactions = new DisabledPendingTransactions();
-      return saveRestoreManager
-          .saveToDisk(pendingTransactionsToSave)
-          .exceptionally(
-              t -> {
-                LOG.error("Error while saving transaction pool to disk", t);
-                return null;
-              });
+      return saveOperation;
     }
     return CompletableFuture.completedFuture(null);
   }
@@ -750,6 +752,7 @@ public class TransactionPool implements BlockAddedObserver {
     private final AtomicBoolean isCancelled = new AtomicBoolean(false);
 
     CompletableFuture<Void> saveToDisk(final PendingTransactions pendingTransactionsToSave) {
+      cancelInProgressReadOperation();
       return serializeAndDedupOperation(
           () -> executeSaveToDisk(pendingTransactionsToSave), writeInProgress);
     }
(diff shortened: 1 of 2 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -758,20 +761,26 @@ CompletableFuture<Void> loadFromDisk() {
       return serializeAndDedupOperation(this::executeLoadFromDisk, readInProgress);
     }
 
+    private void cancelInProgressReadOperation() {
+      if (!readInProgress.get().isDone()) {
+        LOG.debug("Cancelling in progress read operation");
+        isCancelled.set(true);
+        try {
+          readInProgress.get().get();
```
daniellehrner, line 769: It is not very clear in the code why this line is cancelling the read operation. Maybe it's worth it to put this call into a method with a descriptive name

fab-10: `readInProgress.get().get()` wait for the completable future to finish, introduced a method to describe its meaning

pinges (commented):
Is there a reason why you don't just keep the current dump if the restore is canceled?

fab-10:
> Is there a reason why you don't just keep the current dump if the restore is canceled?

The current dump is kept, but already restored txs are removed from it, to avoid duplicates.

In detail when a restore is cancelled, a subset of txs is added to the pool, along with other txs that have arrived via api/p2p, so when the restore is cancelled because the txpool is disabled, the restored txs are removed from the dump file, that is not deleted, and then the txpool is saved to disk, and in this operation the already restored and the new txs are appended to the existing dump file without duplications.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -791,12 +805,17 @@ private CompletableFuture<Void> serializeAndDedupOperation(
...
         final var allTxs = pendingTransactionsToSave.getPendingTransactions();
-        LOG.info("Saving {} transactions to file {}", allTxs.size(), saveFile);
+        LOG.info(
+            "{} {} transactions to file {}",
+            appending ? "Appending" : "Saving",
+            allTxs.size(),
+            saveFile);
 
-        final long savedTxs =
+        final long processedTxs =
```
macfarla, line 818:
```suggestion
        final long processedTxCount =
```

macfarla (commented):
maybe a changelog entry

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -645,15 +646,16 @@ public CompletableFuture<Void> setDisabled() {
       isPoolEnabled.set(false);
       subscribeConnectId.ifPresent(ethContext.getEthPeers()::unsubscribeConnect);
       pendingTransactionsListenersProxy.unsubscribe();
-      final PendingTransactions pendingTransactionsToSave = pendingTransactions;
+      final var saveOperation =
```
Matilda-Clerke, line 649: Would prefer to see a concrete type here rather than `var`. It's not entirely clear what type will be used.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -839,41 +864,77 @@ private void executeLoadFromDisk() {
           LOG.info("Loading transaction pool content from file {}", saveFile);
           try (final BufferedReader br =
               new BufferedReader(new FileReader(saveFile, StandardCharsets.US_ASCII))) {
-            final IntSummaryStatistics stats =
+            final var stats =
```
Matilda-Clerke, line 867: Would prefer to see the concrete type here rather than `var` as it's not clear what the type would be

Matilda-Clerke (commented):
Can't really comment on the meaningfulness of the changes, but LGTM, with some nits about `var` usage

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -758,20 +761,31 @@ CompletableFuture<Void> loadFromDisk() {
...
-            if (!operationInProgress.get().isDone()) {
-              isCancelled.set(true);
-              try {
-                operationInProgress.get().get();
-              } catch (ExecutionException ee) {
-                // nothing to do
-              }
-            }
 
             isCancelled.set(false);
```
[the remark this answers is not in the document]
macfarla: I guess yes because it's still used elsewhere

macfarla (approved):
LGTM

PR #8006, opened 2024-12-09
EIP-7756 Trace updates

shemnon:
## PR description

Update tracing for EIP-7756 changes, mostly swapping hex format and numbers for opcode and gas members.


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/processor/TransactionTracer.java
```diff
@@ -136,7 +136,9 @@ public List<String> traceTransactionToFile(
                             stackedUpdater,
                             transaction,
                             transactionProcessor,
-                            new StandardJsonTracer(out, showMemory, true, true, false),
+                            // FIXME // REVIEWME do we want to change the JSON
+                            // debug calls to use EIP-7756 values for gas and opcode?
+                            new StandardJsonTracer(out, showMemory, true, true, false, true),
```
shemnon, line 141: Do we want to extend this to JSON-RPC calls?  Currently this only impacts CLI tools (and those have a CLI override flag).

lu-pinto: I think having current behaviour with EIP-3155 is what makes most sense for RPC, just because we have seen  users complain about non conformity with other clients. For instance EIP-7756 makes `opName` mandatory which wasn't before and I don't think geth displays that.

Anyway it's best to go with no changes to RPC right now and we can just remove that flag once EOF goes in.

evm/src/main/java/org/hyperledger/besu/evm/tracing/StandardJsonTracer.java
```diff
@@ -186,20 +235,23 @@ public void tracePostExecution(
     if (eofContract) {
       sb.append("\"section\":").append(section).append(",");
     }
-    sb.append("\"op\":").append(opcode).append(",");
+    if (eip3155strict) {
+      sb.append("\"op\":").append(opcode).append(",");
+    } else {
+      sb.append("\"op\":\"").append(fastHexByte(opcode)).append("\",");
```
lu-pinto, line 241: Correct me if I'm wrong but `fastHexByte` is the same as `"0x" + Integer.toHexString(opcode)` why not use that?

shemnon:
@lu-pinto Ready for review.  The defaults are to do 3155 tracing data, but I am needing the new EOF elements for fuzzing in the main branch.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -199,6 +199,13 @@ void setBytes(final String optionValue) {
       scope = INHERIT)
   final Boolean showJsonAlloc = false;
 
+  @Option(
+      names = {"--eip-3155", "--trace.noeip-3155"},
```
lu-pinto, line 203: is `--trace.noeip-3155` really required or intencional? Isn't it `--trace.eip-3155` instead?
The negative version `--no-trace.noeip-3155` looks kinda weird

lu-pinto:
> @lu-pinto Ready for review. The defaults are to do 3155 tracing data, but I am needing the new EOF elements for fuzzing in the main branch.

Yeah no probs with that, current version looks alright to me. Just made a small remark regarding the 2nd name of the flag

evm/src/main/java/org/hyperledger/besu/evm/tracing/StandardJsonTracer.java
shemnon: `fastHexByte` only works on a single byte, and only does hex, and always keeps the leading zero, producing exactly two characters.  The Integer alternative may do up to 4 character and does not keep the leading zero.

lu-pinto:
> > @lu-pinto Ready for review. The defaults are to do 3155 tracing data, but I am needing the new EOF elements for fuzzing in the main branch.
> 
> Yeah no probs with that, current version looks alright to me. Just made a small remark regarding the 2nd name of the flag

I see you updated it to:
```
   @Option(
      names = {"--noeip-3155", "--trace.noeip-3155"},
      description = "Produce a trace with types strictly compatible with EIP-3155.",
```

But it doesn't feel very intuitive to me, I think you should have `--eip-3155` and `--trace.eip-3155` because that's what the variable actually means from the description there. From the documentation https://picocli.info/#_negatable_options it seems that it's expected to work that way. Is that a bug in picocli or something? Is it because you are not using the primitive `boolean` that the negation don't work as expected?

shemnon:
Whether it is a bug (as most people think) or a design choice (what PicoCLI calls it) "flag" type PicoCLI options need to have the value when not specified be the default value, and when specified be the parsed value.  Because the default is true we need to specify the negation as the PicoCLI value.  And once Fusaka ships I intend to flip the default, so keeping the API "--trace.eip-3155" to enable backwards compatibility mode is the goal.

Try changing the option names yourself and watch the tests break. PicoCLI is bizarre in this sense.  

You can see it works itself out in the help text.
```
Usage: evm state-test [-hV] [--[no]intrinsic-gas] [--json] [--json-alloc] [--
                      [no]memory] [--[no]eip-3155] [--[no]time] [--trace.[no]
                      stack] [--trace.[no]returndata] [--trace.[no]storage]
                      [--data-index=<dataIndex>] [--fork=<fork>]
                      [--fork-index=<forkIndex>] [--gas-index=<gasIndex>]
                      [--test-name=<testName>] [--value-index=<valueIndex>]
                      [<stateTestFiles>...]
Execute an Ethereum State Test.
      [<stateTestFiles>...]
      --data-index=<dataIndex>
                            Limit execution to one data variable.
      --fork=<fork>         Force the state tests to run on a specific fork.
      --fork-index=<forkIndex>
                            Limit execution to one fork.
      --gas-index=<gasIndex>
                            Limit execution to one gas variable.
  -h, --help                Show this help message and exit.
      --[no]intrinsic-gas   Calculate and charge intrinsic and tx content gas.
                              Default is not to charge.
      --json, --trace       Trace each opcode as a json object.
      --json-alloc          Output the final allocations after a run.
      --[no]memory, --trace.[no]memory
                            Show the full memory output in tracing for each op.
                              Default is not to show memory.
      --[no]eip-3155, --trace.[no]eip-3155
                            Produce a trace with types strictly compatible with
                              EIP-3155.
      --[no]time            Don't include time data in summary output.
      --test-name=<testName>
                            Limit execution to one named test.
      --trace.[no]stack     Show the operand stack in tracing for each op.
                              Default is to show stack.
      --trace.[no]returndata
                            Show the return data in tracing for each op when
                              present. Default is to show return data.
      --trace.[no]storage   Show the updated storage slots for the current
                              account. Default is to not show updated storage.
  -V, --version             Print version information and exit.
      --value-index=<valueIndex>
                            Limit execution to one value variable.
```

shemnon:
I would like to re-iterate this PR is needed for fuzz testing EOF in Fusaka.  All the tests pass and the options are all exercised. To the best of my ability to tweak it, it appears the PicoCLI documentation is leading to an incorrect understanding of what their code actually does.

lu-pinto (approved):

PR #8009, opened 2024-12-09
Replace waitforpeertask with ethpeers method

Matilda-Clerke:
## PR description
Add PeerSelector.waitForPeer, aiming to replace the EthTask based WaitForPeerTask and WaitForPeersTask classes


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromPeers.java
```diff
@@ -58,15 +55,19 @@ public Optional<FastSyncState> selectNewPivotBlock() {
...
                 new TrailingPeerRequirements(
-                    conservativelyEstimatedPivotBlock(), syncConfig.getMaxTrailingPeers()));
+                    estimatedPivotBlock, syncConfig.getMaxTrailingPeers()));
     trailingPeerLimiter.enforceTrailingPeerLimit();
 
-    return waitForPeers(syncConfig.getSyncMinimumPeerCount());
+    return ethContext
+        .getEthPeers()
+        .waitForPeer((peer) -> peer.chainState().getEstimatedHeight() >= estimatedPivotBlock)
+        .thenRun(() -> {});
```
Matilda-Clerke, line 70: This is the only major change in behavior. I couldn't figure out why on earth we would wait for more than one peer when we could just as easily get started with the first peer that's high enough to contain the pivot block we're looking for. Let me know what you think.

pinges: I think that in PoW we were originally picking a pivot block that had to be validated by 5 peers. Has that changed? Have we made that configurable?

jframe: Yeah we need that for POW, thought was configurable as sometimes we tell people to lower that if they are having trouble syncing.

We should keep the logic the same here so it still works with POW for now. With changes to syncing on POS we won't even use the PivotSelectorFromPeers.

Matilda-Clerke: Good news is, it'll still wait for the configured number of peers to confirm the pivot block. It'll just retry a few times until the peers are present.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -131,11 +130,17 @@ public CompletableFuture<FastSyncState> downloadPivotBlockHeader(
...
+      LOG.debug("Initial sync state {} already contains the block header", currentState);
       return completedFuture(currentState);
     }
 
-    return waitForPeers(1)
+    return ethContext
+        .getEthPeers()
+        .waitForPeer(
+            (peer) ->
+                currentState.getPivotBlockNumber().isEmpty()
```
jframe, line 141: Why is this checking for a missing pivot block number? The if statement at the top of this function ensures we only wait for a peer if we don't already have the pivot block header.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -131,11 +130,17 @@ public CompletableFuture<FastSyncState> downloadPivotBlockHeader(
...
+                    || peer.chainState().getEstimatedHeight()
```
jframe, line 142: We can't rely on the estimated height in pos networks as peers don't send the new block hash message in those networks. So either we guard this on POS or just remove the estimated height check here.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotBlockConfirmer.java
```diff
@@ -187,9 +187,14 @@ private CompletableFuture<BlockHeader> executePivotQuery(final long blockNumber)
       LOG.debug("No peer currently available to query for block {}.", blockNumber);
       pivotHeaderFuture =
           ethContext
-              .getScheduler()
-              .timeout(WaitForPeerTask.create(ethContext, metricsSystem), Duration.ofSeconds(5))
-              .handle((err, res) -> null) // Ignore result
+              .getEthPeers()
+              .waitForPeer(
+                  (peer) ->
+                      peer.chainState().getEstimatedHeight() >= blockNumber
```
jframe, line 193: Similar to the previous comment in FastSyncActions, the pivot block is still being used in the POS syncing and we can't rely on the estimated height. So should either get rid of that or guard with a isPOS check here.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncActions.java
Matilda-Clerke: Initially, I thought this too, but iirc when I tried to run the code, it would result in exceptions from the pivot block number being unset  but having getAsLong() called

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncActions.java
Matilda-Clerke: This check has been removed, so I'll focus discussion on the other related comment

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotBlockConfirmer.java
Matilda-Clerke: I think just removing it should be fine. If the peer isn't suitable when we try to use it, we'll end up back here again eventually anyway.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingPeerTask.java
```diff
@@ -129,13 +130,12 @@ protected void handleTaskError(final Throwable error) {
...
 
-      final WaitForPeerTask waitTask = WaitForPeerTask.create(ethContext, metricsSystem);
       executeSubTask(
           () ->
               ethContext
-                  .getScheduler()
-                  // wait for a new peer for up to 5 seconds
-                  .timeout(waitTask, Duration.ofSeconds(5))
+                  .getEthPeers()
+                  .waitForPeer((peer) -> true)
```
pinges, line 137: We do have the isSuitablePeer(peer) method on this class that we should use here.
```suggestion
                  .waitForPeer((peer) -> isSuitablePeer(peer))
```

pinges (changes requested):
We should make use of the isSuitablePeer method in AbstractRetryingPeerTask.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractRetryingPeerTask.java
Matilda-Clerke: Yeah, that sounds good

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromPeers.java
```diff
@@ -48,7 +46,6 @@ public PivotSelectorFromPeers(
     this.ethContext = ethContext;
```
jframe, line 46: The `metricsSystem` arg can be removed as well

pinges (approved):
LGTM

Matilda-Clerke:
Will wait one more day before merging for current test nodes to finish syncing

PR #8013, opened 2024-12-10
Improve equals performance on Address

ahamlat:
## PR description
We've noticed when profiling the SELFBALANCE during load testing that the throughput issue is related to the equals operation on the address class. 
image

This PR introduces a new implementation of the equals method in the Address class to replace the one inherited from AbstractBytes. The issue with equals in AbstractBytes is that it compares each byte from both addresses individually, and the get method used is megamorphic.


macfarla (commented):
do you need to override hashcode() because you've overridden equals()?

ahamlat:
The current implementation of the hashcode in AbstractByte is good IMO, it is also stored to avoid recalculation. That's why I didn't override it.

macfarla (approved):

PR #8017, opened 2024-12-11
Precompute authorities when importing blocks

fab-10:
## PR description

Like we do for senders, when can precompute the authorities for EIP-7702 transactions, to improve the performance during block import


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/AbstractEngineNewPayload.java
```diff
@@ -392,6 +379,47 @@ public JsonRpcResponse syncResponse(final JsonRpcRequestContext requestContext)
...
+  }
+
+  private void precomputeAuthorities(final Transaction transaction) {
+    final var codeDelegations = transaction.getCodeDelegationList().get();
+    int index = 0;
+    for (final var codeDelegation : codeDelegations) {
+      final var constIndex = index++;
+      mergeCoordinator
+          .getEthScheduler()
+          .scheduleTxWorkerTask(
```
ahamlat, line 409: Precompute is asynchronous here, are we sure that the authorities are not accessed before it is finished or there is a fallback to calculate them if they don't exist yet ?

ahamlat: I wonder if it is not better to precalculate all the senders than precompute all the autorities of all the transactions. Currently, we calculate sender of txn, then precomupte autorities of txn, then calculate sender of txn+1, then precomupte autorities of txn+1

fab-10: about your first question, if this async computation does not finish before the processing, then like for the sender it will be computed on [first usage ](https://github.com/hyperledger/besu/blob/ea04b2537e89eecf1cd0b7f8180b674956923b02/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/CodeDelegationProcessor.java#L113)

About the order, I do not know which is the best solution, because in serial mode, should be better to precompute the authorities to txn before the sender of txn+1, but in parallel mode not sure what is the best solution, what do you suggest?

ahamlat: I was thinking about this case, do we want to improve the time it takes to execute all the transaction by precalculating all the sender or improve a large one by precalculating all the authorizers.
I think it makes sens here to improve the worst case, so let's keep the current order.

ahamlat (approved):

daniellehrner (approved):

PR #8018, opened 2024-12-11
Support code delegations when purging confirmed blocks in the layered txpool

### LINKED ISSUE #7985: Add support for EIP-7702 transactions to the txpool
### Description

When an EIP-7702 transaction is confirmed, we need to look at its code delegations and extract the nonce for the authority and use that information to purge pending transaction in the txpool, like we do for the senders of all the confirmed txs.

fab-10:
## PR description

When an EIP-7702 transaction is confirmed, we need to look at its code delegations and extract the nonce for the authority and use that information to purge pending transaction in the txpool, like we do for the senders of all the confirmed txs.

## Fixed Issue(s)

fixes #7985 


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransaction.java
```diff
@@ -430,7 +430,7 @@ public interface MemorySize {
     int ACCESS_LIST_ENTRY_SHALLOW_SIZE = 248;
     int OPTIONAL_ACCESS_LIST_SHALLOW_SIZE = 40;
     int OPTIONAL_CODE_DELEGATION_LIST_SHALLOW_SIZE = 40;
-    int CODE_DELEGATION_ENTRY_SIZE = 472;
+    int CODE_DELEGATION_ENTRY_SIZE = 520;
```
daniellehrner, line 433: The chain id of the authorisation tuple has recently been changed from uint64 to uint256. Does the size need to be changed because of that?

daniellehrner (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/PendingTransaction.java
fab-10: the estimation trades the accuracy for speed, so the `chainId` is estimated on the base it is a long, which is true for named networks and probably in most of the other cases, so I think we can keep this estimation, also because for a EIP-7702 tx to be in the pool the chainIds in the authorization tuple must be 0 or match the networkId.

PR #8020, opened 2024-12-11
Clean up pivot selector from peers and unit test

Matilda-Clerke:
## PR description
Simplify the code in PivotSelectorFromPeers.


ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromPeersTest.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+public class PivotSelectorFromPeersTest {
+  private EthContext ethContext;
+  private EthPeers ethPeers;
```
macfarla, line 34:
```suggestion
  @Mock private EthPeers ethPeers;
```

macfarla: if you use the annotation you don't need to instantiate the mock in the before() method - it's a bit shorter and clearer

macfarla: you might need to add this for the class

Matilda-Clerke: Thanks, used to do something similar with junit 4 and couldn't get it to work with junit 5

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromPeers.java
```diff
@@ -86,41 +87,24 @@ protected Optional<FastSyncState> fromBestPeer(final EthPeer peer) {
...
-  private long countPeersThatCanDeterminePivotBlock() {
-    return ethContext
-        .getEthPeers()
-        .streamAvailablePeers()
-        .filter(this::canPeerDeterminePivotBlock)
-        .count();
-  }
-
-  private boolean canPeerDeterminePivotBlock(final EthPeer peer) {
-    LOG.debug(
```
macfarla, line 118: we're losing this log line, is that intentional?

macfarla (approved):
just a Q about logging

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromPeers.java
Matilda-Clerke: It's not a particularly good log as it doesn't include any context and is likely to end up listing peers. I can look to add something similar if you think it's worthwhile

macfarla: all good just checking it won't be missed

PR #8037, opened 2024-12-16
Revise the approach for setting level_compaction_dynamic_level_bytes

ahamlat:
## PR description
We noticed with the upgrade from RocksDB from 8.3.2 to 9.7.3 that the defaults on LeveledCompaction are changed to reduce space amplification. This changes triggered big compactions depending on the nodes when opening RocksDB databse. This is related to change in the value of this parameter `level_compaction_dynamic_level_bytes`, changed from false to true.
This PR modify the logic for determining `level_compaction_dynamic_level_bytes` 

1. If a node is already sync'd and validating : we keep the existing configuration (`level_compaction_dynamic_level_bytes=false` if the node is not upgraded yet to 24.12.0, 24.12.1 or 24.12.2) (i.e no compaction)
2. If the node is already ugraded to 24.12.0, 24.12.1 or 24.12.2, we keep the existing configuration, i.e `level_compaction_dynamic_level_bytes=true`, as the compaction already happened
3. If the node is a new node, sync'ing from scratch, we apply directly the default configuration, i.e : `level_compaction_dynamic_level_bytes=true`

**Tested with success** :

1. A node already sync'ed and not upgraded keeps `level_compaction_dynamic_level_bytes` set to false (no compaction)
2. A node already sync'ed and upgraded keeps `level_compaction_dynamic_level_bytes` set to true (compaction already happened, so no more overhead)
3. A node sync'ing from scratch, start with the recommended configuration, `level_compaction_dynamic_level_bytes` set to true

To detect similar issues in the future, we added a warning log if database opening exceeds 60 seconds.

```
{"@timestamp":"2024-12-16T15:35:11,767","level":"INFO","thread":"main","class":"DatabaseMetadata","message":"Lookup database metadata file in data directory: /data/besu","throwable":""}
{"@timestamp":"2024-12-16T15:35:11,791","level":"INFO","thread":"main","class":"RocksDBKeyValueStorageFactory","message":"Existing database at /data/besu. Metadata versionedStorageFormat=BaseVersionedStorageFormat{format=BONSAI, version=3}. Processing WAL...","throwable":""}
{"@timestamp":"2024-12-16T15:36:11,980","level":"WARN","thread":"pool-2-thread-1","class":"RocksDBOpener","message":"Opening RocksDB database is taking longer than 60 seconds... This may be due to prolonged RocksDB compaction. Please wait until the end of the compaction. No action is needed from the user.","throwable":""}
{"@timestamp":"2024-12-16T15:37:13,740","level":"INFO","thread":"main","class":"FlatDbStrategyProvider","message":"Flat db mode found FULL","throwable":""}

```


jflo (approved):

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -186,15 +185,47 @@ public RocksDBColumnarKeyValueStorage(
...
+      List<ColumnFamilyDescriptor> cfDescriptors = new ArrayList<>();
+
+      String latestOptionsFileName =
+          OptionsUtil.getLatestOptionsFileName(
+              configuration.getDatabaseDir().toString(), Env.getDefault());
+      LOG.trace("Latest OPTIONS file detected: " + latestOptionsFileName);
+
+      String optionsFilePath =
+          configuration.getDatabaseDir().toString() + "/" + latestOptionsFileName;
+      OptionsUtil.loadOptionsFromFile(configOptions, optionsFilePath, dbOptions, cfDescriptors);
```
matkt, line 201: you cannot use loadLatestOptions ? directly instead of calling getLatestOptionsFileName and loadOptionsFromFile ?

ahamlat: I think it is better for troubleshooting and check which file was used to load the options. You can have it if you enable trace logs.

macfarla (approved):
nice. changelog entry?

PR #8051, opened 2024-12-18
Return rpc error in eth_getLogs on unknown block hash

### LINKED ISSUE #8007: eth_getLogs must return "block unknown" when blockHash is not found
### Steps to Reproduce
```
curl -s -H 'content-type:application/json' --data '{"jsonrpc":"2.0","id":"53","method":"eth_getLogs","params":[{"topics":["0x55f4c645c36aa5cd3f443d6be44d7a7a5df9d2100d7139dfc69d4289ee072319"],"blockHash":"0x00000000000000000000000000000000000000000000000000000000000000ff"}]}' http://localhost:8445
```

**Expected behavior:**
If the node does not have the block shall return an error instead of saying "That block does not have logs". This can be problematic if the node is behind a load balancer and is not 100% in sync
```
{"jsonrpc":"2.0","id":"53","error":{"code":-32000,"message":"unknown block"}}
```

**Actual behavior:** 
```
{"jsonrpc":"2.0","id":"53","result":[]}
```

**Frequency:** Consystently

* Software version: besu 24.10

### Notes:
Geth and Infura return `unknown block`

jonson:
## Fixed Issue(s)

Fixes #8007 

The issue describes the problem well, so I won't repeat it here.

The root cause is `org.hyperledger.besu.ethereum.api.query.BlockchainQueries.matchingLogs` detects an unknown hash and returns an empty list.  This was the desired behaviour at some point in time, as there was even a test to validate it.

I have updated the method to throw an exception in this situation now, as IMO it really is a bad input if it's being called with an unknown hash.  This has potentially larger impacts as there are other callers of this function, but to my knowledge they are all passing in a valid hash (acquired via a `BlockHeader`).  An alternative approach could be to just patch `EthGetLogs` and check the block header prior to calling `matchingLogs`, which limits the impact.

I'm not sure if it's ok to be throwing an `InvalidJsonRpcParameters` from within the `org.hyperledger.besu.ethereum.api.query` package, there do not seem to be any other imports, so let me know if that should be cleaned up (I originally was going to raise a new exception type here, but I also noticed this package doesn't have any custom exceptions either).

Finally I am re-using the existing `RpcErrorType.BLOCK_NOT_FOUND` for this response, as I did not want to introduce a type to keep the change small.  But I can if that is better.

:information_source: I could not run the acceptance tests locally, they killed my (high-end) laptop.

### DIFF
--- a/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueriesTest.java
+++ b/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueriesTest.java
@@ -18,10 +18,13 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.hyperledger.besu.ethereum.core.InMemoryKeyValueStorageProvider.createInMemoryBlockchain;
 import static org.hyperledger.besu.ethereum.core.InMemoryKeyValueStorageProvider.createInMemoryWorldStateArchive;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import org.hyperledger.besu.datatypes.Address;
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.datatypes.Wei;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.exception.InvalidJsonRpcParameters;
+import org.hyperledger.besu.ethereum.api.jsonrpc.internal.response.RpcErrorType;
 import org.hyperledger.besu.ethereum.chain.MutableBlockchain;
 import org.hyperledger.besu.ethereum.core.Block;
 import org.hyperledger.besu.ethereum.core.BlockDataGenerator;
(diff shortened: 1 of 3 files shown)


macfarla:
change looks OK. see if you can follow the instructions to fix the DCO https://github.com/hyperledger/besu/pull/8051/checks?check_run_id=34624508989

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/query/BlockchainQueriesTest.java
```diff
@@ -402,12 +405,19 @@ public void logsShouldBeFlaggedAsRemovedWhenBlockIsNotInCanonicalChain() {
...
+
+    try {
+      queries.matchingLogs(Hash.ZERO, new LogsQuery.Builder().build(), () -> true);
+    } catch (final InvalidJsonRpcParameters e) {
+      assertThat(e.getMessage()).isEqualTo("Unknown block hash");
+      assertThat(e.getRpcErrorType()).isEqualTo(RpcErrorType.BLOCK_NOT_FOUND);
+      return;
+    }
+
+    fail("Expected InvalidJsonRpcParameters exception");
```
Gabriel-Trintinalia, line 420: `assertThrows` is used in other parts of the codebase to assert exceptions
```
    var exception =
        assertThrows(
            InvalidJsonRpcParameters.class,
            () -> queries.matchingLogs(Hash.ZERO, new LogsQuery.Builder().build(), () -> true));
    assertThat(exception.getMessage()).isEqualTo("Unknown block hash");
    assertThat(exception.getRpcErrorType()).isEqualTo(RpcErrorType.BLOCK_NOT_FOUND);
```

Gabriel-Trintinalia:
@jonson  I think we should try the alternative approach, which could be to patch `EthGetLogs` and check the block header before calling matching logs. `BlockchainQueries` seems to have this behaviour of returning empty for the majority of its methods. Another option could be to change the return type to `Optional<List<LogWithMetadata>` and return `Optional.empty` if the block does not exist

Gabriel-Trintinalia:
Hi @jonson are you planning to keep working on this issue?

PR #8057, opened 2024-12-20
Bls updates pectra devnet 5

garyschulte:
## PR description

PR against pectra devnet 5 branch, duplicate of #8008

reference tests are expected to fail


garyschulte:
@siladu any idea why CI is not running here?

PR #8058, opened 2024-12-20
8053: Add GetPooledTransactionsFromPeerTask and appropriate usage

Matilda-Clerke:
## PR description
Add GetPooledTransactionsFromPeerTask and appropriate usage replacing old EthTask with new peer task system


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/task/GetPooledTransactionsFromPeerTask.java
```diff
@@ -0,0 +1,79 @@
...
+    return GetPooledTransactionsMessage.create(hashes);
+  }
+
+  @Override
+  public List<Transaction> processResponse(final MessageData messageData)
+      throws InvalidPeerTaskResponseException {
+    final PooledTransactionsMessage pooledTransactionsMessage =
+        PooledTransactionsMessage.readFrom(messageData);
+    final List<Transaction> responseTransactions = pooledTransactionsMessage.transactions();
+    if (responseTransactions.size() != hashes.size()) {
```
jframe, line 56: This condition should be relaxed to `responseTransactions.size() > hashes.size()` as the PooledTransactions response message can return fewer transactions than the list of requested transaction hashes due to the peer not having all the transactions in its transaction pool. This is allowed by the spec.

Matilda-Clerke: Good catch, I'll update it

jframe (approved):

PR #8073, opened 2024-12-31
Fix: Extra Character in Git Command

brawncode:
### Subject: Fix Extra Character in Git Command

issue in the command used for filtering commits based on the "Signed-off-by" line.
there was an extra `--` symbol after the `--invert-grep` option in the following command.

this additional `--` was causing the command to behave incorrectly.

i've removed the extra symbol to ensure the command works as expected.

### DIFF
--- a/scripts/dco_check.sh
+++ b/scripts/dco_check.sh
@@ -26,7 +26,7 @@ do
     status=1
     commit_hash="$(echo "$results" | cut -d' ' -f1)"
     >&2 echo "$commit_hash is missing Signed-off-by line."
-  done < <(git log "$branch" --no-merges --pretty="%H %ae" --grep 'Signed-off-by' --invert-grep -- )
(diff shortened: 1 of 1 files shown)


shemnon:
Do not merge this, the script is fine and has been working great for 5 years.  These two characters are also an essential anti-hacking protection.

The `--` symbol separates command line options from a list of files, according to git's own help text when runing without the `--`
```
 % git log "$branch" --no-merges --pretty="%H %ae" --grep 'Signed-off-by' --invert-grep   
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

This line is part of the (process substitution)[https://tldp.org/LDP/abs/html/process-sub.htm] in a bash shell that will append the list of searched files.  The double dash is essential for script safety because without it a crafty PR could use file names to send various other grep options to maliciously disrupt the system that would be blocked by the shift to file-name only processing.

PR #8076, opened 2025-01-03
Improve debug_traceBlock calls performance

### LINKED ISSUE #5322: implement pipeline on debug_ tracing endpoints

ahamlat:
## PR description
This PR introduces several key improvements to the `debug_traceBlock`, `debug_traceBlockByNumber`, and `debug_traceBlockByHash` methods, significantly enhancing both performance and memory usage:

- The core change is the adoption of a pipeline architecture, where:
  - One thread handles transaction execution.
  - Other threads processes the execution results and generates the JSON output. This parallelization ensures better resource utilization and reduces overall execution time.

- Memory usage is optimized by the pipeline processing approach. Instead of retaining all trace frames for every transaction, the pipeline now keeps only the trace frames relevant to the current transaction, significantly reducing memory consumption.

- The transformation of trace frames into Java JSON representations now occurs in parallel. 

- Default trace options have been adjusted to align with the behavior of Geth and Nethermind:

  - disableStorage is set to false (enabling storage tracing).
  - disableMemory is set to true (disabling memory tracing).
  - disableStack is set to false (enabling stack tracing).
  
- The structure of the JSON output has been modified to match Geth's format:

  - Hexadecimal values are now displayed as short hex strings (e.g., 0x1, 0xabc), which is more compact and efficient.
  - The stack array is always displayed when the option is enabled, even when empty, to ensure consistency with other Ethereum clients. However, memory and storage are omitted if they are empty, which reduces unnecessary data.
  - Storage keys and values are now displayed without the 0x prefix, in line with Geth's behavior.

- Change the implementation of `toShortHexString`, by creating a new method `toCompactHex`. The main improvement is to not relay on Bytes, but on byte[] to avoid dynamic dispatch mechanism related to the use of methods `get(i) `and `size().` Those methods are not used anymore thanks to the switch to StringBuilder, even if it has its own cons. This method will replace in the future `toFastHex` implementation, once we have finished the work on Tuweni.

These changes have resulted in a 3X performance improvement, reducing the size of the output by 2 to 3 times, making previously unresponsive calls fast enough with comparable performance with other clients. The optimizations not only improve response times but also allow for smoother operation with lower memory usage, even when processing large traces.

Overall, the new approach ensures better performance, reduced resource usage, and more consistent output with other Ethereum clients, particularly Geth and Nethermind.

## Fixed Issue(s)

Fixes #5322 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Breaking Changes
 - `--host-whitelist` has been deprecated since 2020 and this option is removed. Use the equivalent `--host-allowlist` instead. 
+- Changed tracer API to include the mining beneficiary in BlockAwareOperationTracer::traceStartBlock [#8096](https://github.com/hyperledger/besu/pull/8096)
 
 ### Upcoming Breaking Changes
 - `MetricSystem::createLabelledGauge` is deprecated and will be removed in a future release, replace it with `MetricSystem::createLabelledSuppliedGauge`
--- a/besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
+++ b/besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
@@ -17,6 +17,7 @@ package org.hyperledger.besu.services;
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.hyperledger.besu.ethereum.mainnet.feemarket.ExcessBlobGasCalculator.calculateExcessBlobGasForParent;
 
+import org.hyperledger.besu.datatypes.Address;
 import org.hyperledger.besu.datatypes.BlobGas;
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.datatypes.Wei;
@@ -152,16 +153,13 @@ public class TraceServiceImpl implements TraceService {
             .toList();
     Tracer.processTracing(
         blockchainQueries,
-        blocks.get(0).getHash(),
+        blocks.getFirst().getHash(),
         traceableState -> {
           final WorldUpdater worldStateUpdater = traceableState.updater();
           final ChainUpdater chainUpdater = new ChainUpdater(traceableState, worldStateUpdater);
           beforeTracing.accept(worldStateUpdater);
           final List<TransactionProcessingResult> results = new ArrayList<>();
-          blocks.forEach(
-              block -> {
-                results.addAll(trace(blockchain, block, chainUpdater, tracer));
-              });
+          blocks.forEach(block -> results.addAll(trace(blockchain, block, chainUpdater, tracer)));
           afterTracing.accept(chainUpdater.getNextUpdater());
           return Optional.of(results);
         });
@@ -191,7 +189,9 @@ public class TraceServiceImpl implements TraceService {
     final ProtocolSpec protocolSpec = protocolSchedule.getByBlockHeader(block.getHeader());
     final MainnetTransactionProcessor transactionProcessor = protocolSpec.getTransactionProcessor();
     final BlockHeader header = block.getHeader();
-    tracer.traceStartBlock(block.getHeader(), block.getBody());
+    final Address miningBeneficiary =
+        protocolSpec.getMiningBeneficiaryCalculator().calculateBeneficiary(block.getHeader());
+    tracer.traceStartBlock(block.getHeader(), block.getBody(), miningBeneficiary);
 
     block
         .getBody()
--- a/besu/src/test/java/org/hyperledger/besu/services/TraceServiceImplTest.java
+++ b/besu/src/test/java/org/hyperledger/besu/services/TraceServiceImplTest.java
@@ -114,7 +114,9 @@ class TraceServiceImplTest {
 
     final Block tracedBlock = blockchain.getBlockByNumber(blockNumber).get();
 
-    verify(opTracer).traceStartBlock(tracedBlock.getHeader(), tracedBlock.getBody());
+    verify(opTracer)
+        .traceStartBlock(
+            tracedBlock.getHeader(), tracedBlock.getBody(), tracedBlock.getHeader().getCoinbase());
 
     tracedBlock
         .getBody()
@@ -163,7 +165,11 @@ class TraceServiceImplTest {
         .map(Optional::get)
         .forEach(
             tracedBlock -> {
-              verify(opTracer).traceStartBlock(tracedBlock.getHeader(), tracedBlock.getBody());
+              verify(opTracer)
+                  .traceStartBlock(
+                      tracedBlock.getHeader(),
+                      tracedBlock.getBody(),
+                      tracedBlock.getHeader().getCoinbase());
               tracedBlock
                   .getBody()
                   .getTransactions()
@@ -312,7 +318,8 @@ class TraceServiceImplTest {
     }
 
     @Override
-    public void traceStartBlock(final BlockHeader blockHeader, final BlockBody blockBody) {
+    public void traceStartBlock(
+        final BlockHeader blockHeader, final BlockBody blockBody, final Address miningBeneficiary) {
       if (!traceStartBlockCalled.add(blockHeader.getBlockHash())) {
         fail("traceStartBlock already called for block " + blockHeader);
       }
--- a/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
+++ b/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockCreator.java
@@ -221,7 +221,7 @@ public abstract class AbstractBlockCreator implements AsyncBlockCreator {
       final BlockAwareOperationTracer operationTracer =
           pluginTransactionSelector.getOperationTracer();
 
-      operationTracer.traceStartBlock(processableBlockHeader);
+      operationTracer.traceStartBlock(processableBlockHeader, miningBeneficiary);
       timings.register("preTxsSelection");
       final TransactionSelectionResults transactionResults =
           selectTransactions(
--- a/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/InterruptibleOperationTracer.java
+++ b/ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/InterruptibleOperationTracer.java
@@ -41,8 +41,9 @@ public class InterruptibleOperationTracer implements BlockAwareOperationTracer {
   }
 
   @Override
-  public void traceStartBlock(final BlockHeader blockHeader, final BlockBody blockBody) {
-    delegate.traceStartBlock(blockHeader, blockBody);
+  public void traceStartBlock(
+      final BlockHeader blockHeader, final BlockBody blockBody, final Address miningBeneficiary) {
+    delegate.traceStartBlock(blockHeader, blockBody, miningBeneficiary);
   }
 
   @Override
@@ -51,8 +52,9 @@ public class InterruptibleOperationTracer implements BlockAwareOperationTracer {
   }
 
   @Override
-  public void traceStartBlock(final ProcessableBlockHeader processableBlockHeader) {
-    delegate.traceStartBlock(processableBlockHeader);
+  public void traceStartBlock(
+      final ProcessableBlockHeader processableBlockHeader, final Address miningBeneficiary) {
+    delegate.traceStartBlock(processableBlockHeader, miningBeneficiary);
   }
 
   @Override
--- a/plugin-api/build.gradle
+++ b/plugin-api/build.gradle
@@ -71,7 +71,7 @@ Calculated : ${currentHash}
 tasks.register('checkAPIChanges', FileStateChecker) {
   description = "Checks that the API for the Plugin-API project does not change without deliberate thought"
   files = sourceSets.main.allJava.files
-  knownHash = 'By1EMWJvyXiRxNf52xZ0BUC2LpE6D4VlNHI0jYJryj0='
+  knownHash = 'b/u9Ety5B+Ni8UwGhvU8dq4jcZtulNczsVQZgG0Q5fw='
 }
 check.dependsOn('checkAPIChanges')
 
--- a/plugin-api/src/main/java/org/hyperledger/besu/plugin/services/tracer/BlockAwareOperationTracer.java
+++ b/plugin-api/src/main/java/org/hyperledger/besu/plugin/services/tracer/BlockAwareOperationTracer.java
@@ -14,6 +14,7 @@
  */
 package org.hyperledger.besu.plugin.services.tracer;
 
+import org.hyperledger.besu.datatypes.Address;
 import org.hyperledger.besu.evm.tracing.OperationTracer;
 import org.hyperledger.besu.plugin.data.BlockBody;
 import org.hyperledger.besu.plugin.data.BlockHeader;
@@ -37,8 +38,10 @@ public interface BlockAwareOperationTracer extends OperationTracer {
    *
    * @param blockHeader the header of the block which is traced
    * @param blockBody the body of the block which is traced
+   * @param miningBeneficiary the address of miner building the block
    */
-  default void traceStartBlock(final BlockHeader blockHeader, final BlockBody blockBody) {}
+  default void traceStartBlock(
+      final BlockHeader blockHeader, final BlockBody blockBody, final Address miningBeneficiary) {}
 
   /**
    * Trace the end of a block.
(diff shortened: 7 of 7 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/StructLog.java
```diff
@@ -153,4 +148,39 @@ public int hashCode() {
     result = 31 * result + Arrays.hashCode(stack);
     return result;
   }
+
+  public static String toCompactHex(final Bytes abytes, final boolean prefix) {
```
Matilda-Clerke, line 152: Any reason we can't use `Bytes.toHexString` or `Bytes.toFastHex`?

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceBlockByHashTest.java
```diff
@@ -15,125 +15,109 @@
...
-            invocation ->
-                invocation
-                    .<Function<MutableWorldState, Optional<? extends JsonRpcResponse>>>getArgument(
-                        1)
-                    .apply(mutableWorldState))
-        .when(blockchainQueries)
-        .getAndMapWorldState(any(), any());
-    when(blockchainQueries.getBlockHeaderByHash(any(Hash.class)))
-        .thenReturn(Optional.of(blockHeader));
+    MockitoAnnotations.openMocks(this);
```
Matilda-Clerke, line 63: Might be nicer to just add `@ExtendWith(MockitoExtension.class)` annotation to the class (and optionally `@MockitoSettings(strictness = Strictness.LENIENT)` if desired

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceBlockByNumberTest.java
```diff
@@ -14,131 +14,112 @@
...
+  @Mock private Blockchain blockchain;
+  @Mock private Block block;
+  @Mock private BlockHeader blockHeader;
+  @Mock private ProtocolSchedule protocolSchedule;
+  @Mock private ObservableMetricsSystem metricsSystem;
+  private DebugTraceBlockByNumber debugTraceBlockByNumber;
+
+  @BeforeEach
+  public void setUp() {
+    MockitoAnnotations.openMocks(this);
```
Matilda-Clerke, line 63: Here too

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceBlockTest.java
```diff
@@ -14,53 +14,106 @@
...
+  @Mock private BlockchainQueries blockchainQueries;
+  @Mock private Blockchain blockchain;
+  @Mock private ObservableMetricsSystem metricsSystem;
+  @Mock private WithdrawalsProcessor withdrawalsProcessor;
+  @Mock private TransactionValidatorFactory alwaysValidTransactionValidatorFactory;
+  private DebugTraceBlock debugTraceBlock;
+
+  @BeforeEach
+  public void setUp() {
+    MockitoAnnotations.openMocks(this);
```
Matilda-Clerke, line 73: Here too

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/StructLogTest.java
```diff
@@ -0,0 +1,235 @@
...
+public class StructLogTest {
+
+  @Mock private TraceFrame traceFrame;
+
+  private StructLog structLog;
+  private ObjectMapper objectMapper;
+
+  @BeforeEach
+  public void setUp() {
+    MockitoAnnotations.openMocks(this);
```
Matilda-Clerke, line 46: Here too

Matilda-Clerke (commented):
LGTM, just a few suggestions in comments. I'll leave unapproved to make sure someone more experienced in the area gets to review too.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
```diff
@@ -0,0 +1,157 @@
...
+  public AbstractDebugTraceBlock(
+      final ProtocolSchedule protocolSchedule,
+      final BlockchainQueries blockchainQueries,
+      final ObservableMetricsSystem metricsSystem) {
+    this.blockchainQueriesSupplier = Suppliers.ofInstance(blockchainQueries);
+    this.protocolSchedule = protocolSchedule;
+    this.outputCounter =
+        metricsSystem.createLabelledCounter(
+            BesuMetricCategory.BLOCKCHAIN,
+            "transactions_debugTraceblock_pipeline_processed_total",
```
siladu, line 65: nit: is the casing correct here?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
```diff
@@ -0,0 +1,157 @@
...
+                                  outputCounter,
+                                  false,
+                                  "debug_trace_block")
+                              .thenProcess("executeTransaction", executeTransactionStep)
+                              .thenProcessAsyncOrdered(
+                                  "debugTraceTransactionStep", debugTraceTransactionStep, 4)
+                              .andFinishWith("collect_results", tracesList::add);
+
+                      try {
+                        if (getBlockchainQueries().getEthScheduler().isPresent()) {
```
siladu, line 139: Is there a realistic scenario where EthScheduler isn't available?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
```diff
@@ -0,0 +1,157 @@
...
+                          getBlockchainQueries()
+                              .getEthScheduler()
+                              .get()
+                              .startPipeline(traceBlockPipeline)
+                              .get();
+                        } else {
+                          EthScheduler ethScheduler =
+                              new EthScheduler(1, 1, 1, 1, new NoOpMetricsSystem());
```
siladu, line 147: This is a little mysterious, might be good to name the variable something like "singleThreadedEthScheduler" perhaps? Don't we want metrics too?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceBlockByNumber.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright ConsenSys AG.
+ * Copyright contributors to Hyperledger Besu.
```
siladu, line 2: nit: think you used just "Besu" above - is that the latest template?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceBlockByNumber.java
```diff
@@ -75,22 +98,68 @@ protected Object resultByBlockNumber(
...
     }
+    Optional<Block> maybeBlock =
+        getBlockchainQueries().getBlockchain().getBlockByNumber(blockNumber);
 
-    return blockHash
+    return maybeBlock
         .flatMap(
-            hash ->
+            block ->
                 Tracer.processTracing(
```
siladu, line 107: Why can't this reuse the same call made in AbstractDebugTraceBlock (as ByHash appears to)?

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/StructLogTest.java
```diff
@@ -0,0 +1,235 @@
+/*
+ * Copyright contributors to Hyperledger Besu.
```
siladu, line 2: Older copyright?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/DebugOperationTracer.java
siladu, line 1: Can you explain what's happening with the changes in this file please?

siladu (commented):
Just an initial pass. Would like to dive deeper into the pipeline changes but feel free to merge if someone else approves

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/StructLog.java
ahamlat: The reason is explained in the description. We were using `toShortHexString` and it was one of the biggest bottelnecks. As we're currently migrating tuweni, I decided to create this method directly in the StructLog java class. It should replace in the future `Bytes.toFastHex` implementation.

> Change the implementation of toShortHexString, by creating a new method toCompactHex. The main improvement is to not relay on Bytes, but on byte[] to avoid dynamic dispatch mechanism related to the use of methods get(i) and size(). Those methods are not used anymore thanks to the switch to StringBuilder, even if it has its own cons. This method will replace in the future toFastHex implementation, once we have finished the work on Tuweni.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
```diff
@@ -0,0 +1,157 @@
...
+      final Optional<Block> maybeBlock) {
+    return maybeBlock
+        .flatMap(
+            block ->
+                Tracer.processTracing(
+                    getBlockchainQueries(),
+                    block.getHash(),
+                    traceableState -> {
+                      Collection<DebugTraceTransactionResult> tracesList =
+                          new CopyOnWriteArrayList<>();
```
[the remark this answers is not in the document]
ahamlat: Yes, you're right, Collections.synchronizedList is better choice here as CopyOnWriteArrayList is not suitable for heavy write loads, which is the case here. I tend to not use Collections.synchronized* but in this case it is a better choice.

ahamlat:
Just referencing here https://github.com/Consensys/protocol-misc/issues/1010, to be sure `toCompactHex` method will be migrated to Tuweni to replace `toFastHex`

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceBlockByNumber.java
ahamlat: Good question, this is because DebugTraceBlockByNumber already inherit from AbstractBlockParameterMethod, to handle flags related to block number, like latest, safe or finalized.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
ahamlat: This was inherited from Trace_block implementation. I changed both Trace* and DebugTrace* implementations to use the Besu ethScheduler, created when we start Besu. Now, we don't need anymore to check if it null or not.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
ahamlat: I changed the implementation and deleted this code https://github.com/hyperledger/besu/commit/6a541e5798cccfe8bcc580658c85f8809599ea03

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/DebugOperationTracer.java
ahamlat: The changes made in DebugOperationTracer are related to not displaying storage and memory in structlog json document if they are empty. This is to match Geth behaviour and reduce Json output size.

ahamlat:
> Just an initial pass. Would like to dive deeper into the pipeline changes but feel free to merge if someone else approves

To validate the changes with the pipeline, during the testing phase I created a new RPC method debug_traceBlock_with_pipeline without changing the existing one and checked on different examples that the output with both implementations is the same.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/DebugTraceBlockByNumber.java
ahamlat: It was auto-generated. I was aware that it changed. I modified all the filed related to this PR with the new template

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
```diff
@@ -0,0 +1,157 @@
...
+                          EthScheduler ethScheduler =
+                              new EthScheduler(1, 1, 1, 1, new NoOpMetricsSystem());
+                          ethScheduler.startPipeline(traceBlockPipeline).get();
+                        }
+                      } catch (InterruptedException | ExecutionException e) {
+                        throw new RuntimeException(e);
+                      }
+                      return Optional.of(tracesList);
+                    }))
+        .orElse(null);
```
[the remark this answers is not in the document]
ahamlat: This would change the rpc call output from
```
{"jsonrpc":"2.0","id":1,"result":null}
```
to 
```
{"jsonrpc":"2.0","id":1,"result":[]}
```
It was the original behaviour before this PR and would like to keep it as it is.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
```diff
@@ -136,17 +140,7 @@ protected Collection<DebugTraceTransactionResult> getTraces(
...
-                              .getEthScheduler()
-                              .get()
-                              .startPipeline(traceBlockPipeline)
-                              .get();
-                        } else {
-                          EthScheduler ethScheduler =
-                              new EthScheduler(1, 1, 1, 1, new NoOpMetricsSystem());
-                          ethScheduler.startPipeline(traceBlockPipeline).get();
-                        }
+                        ethScheduler.startPipeline(traceBlockPipeline).get();
```
Matilda-Clerke, line 143: Much cleaner, very nice

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceHostAllowlistTest.java
```diff
@@ -140,7 +141,8 @@ public void initServerAndClient() throws Exception {
                 vertx,
                 mock(ApiConfiguration.class),
                 Optional.empty(),
-                mock(TransactionSimulator.class));
+                mock(TransactionSimulator.class),
+                new EthScheduler(1, 1, 1, new NoOpMetricsSystem()));
```
Matilda-Clerke, line 145: We should consider using DeterministicEthScheduler as done in most other unit tests requiring an EthScheduler.

Matilda-Clerke (commented):
Definitely some good changes here

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcHttpServiceHostAllowlistTest.java
ahamlat: I noticed that DeterministicEthScheduler caused a performance issue on some trace calls unit tests on trace_block, trace_filter and trace_replayBlockTransactions. Unit tests were failing because of a timeout. I rollbacked that change on that specific test.

ahamlat:
@macfarla @siladu @Gabriel-Trintinalia Would you mind have a look to this PR ?

PR #8090, opened 2025-01-09
Fix guess type for delegate code transactions

Gabriel-Trintinalia:
## PR description
Since `DELEGATE_CODE` transactions have `maxFeePerGas` and `accessList`, the check for `codeDelegationAuthorizations` needs to come first; otherwise, we get the wrong type.


usmansaleem (approved):
LGTM

usmansaleem (approved):
lgtm

PR #8095, opened 2025-01-09
Precompile Caching MVP

garyschulte:
## PR description

PR adds precompile caching behavior for an MVP set of precompiles that are costly enough to benefit from caching.    Provision is added to disable caching via command line arg (for gas costing reasons), but it is enabled by default in besu, and disabled by default in evmtool and benchmark subcommand.  

Changes:
* add a static member and setter in AbstractPrecompiledContract used to control whether we want to cache results
* add precompile-specific LRU caches with rational size limits in each MVP precompile
* add a cli arg for precompile caching, defaulted to true

MVP precompiles include:
 * altbn128/bn254 precompiles for add, mul and pairing
 * ecrecover precompile
 * blake2 precompile
 * kzg point precompile
 * bls precompiles

Feedback welcome on the design choices:
* one cache per precompile contract (since each will have different input and output size characteristics)
* cache is <hashCode, input_and_result_tuple> in order to verify input is truly identical rather than just matching by hashCode (it is trivial to construct requests that have different inputs, but similar Bytes hashCode)

Parallel transaction execution should benefit from precompile caching when state conflicts are detected.  Attached are preliminary results from the nethermind gas-benchmarks suite which indicate performance does not seem to take a hit for cache checking and misses, and the caching itself is effective for repetitive/identical inputs

updated:
[ecmul_new.pdf](https://github.com/user-attachments/files/19396239/ecmul_new.pdf)
[ecrec_new.pdf](https://github.com/user-attachments/files/19396240/ecrec_new.pdf)
[blake2f_new.pdf](https://github.com/user-attachments/files/19396241/blake2f_new.pdf)

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -982,7 +982,14 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
       VersionMetadata.versionCompatibilityChecks(versionCompatibilityProtection, dataDir());
 
       configureNativeLibs(Optional.ofNullable(network));
-      AbstractPrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
+      if (enablePrecompileCaching
+          && getDataStorageConfiguration()
+              .getPathBasedExtraStorageConfiguration()
+              .getUnstable()
+              .isParallelTxProcessingEnabled()) {
+        // enable precompile caching if it is enabled and parallel tx processing is enabled:
+        configurePrecompileCaching();
+      }
 
       besuController = buildController();
 
@@ -1007,6 +1014,28 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
       logger.debug("Startup failure cause", e);
       throw new ParameterException(this.commandLine, e.getMessage(), e);
     }
+  }
+
+  private void configurePrecompileCaching() {
+    // enable precompile caching:
+    AbstractPrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
+
+    // set a metric logger
+    final var precompileCounter =
+        getMetricsSystem()
+            .createLabelledCounter(
+                BesuMetricCategory.BLOCK_PROCESSING,
+                "precompile_cache",
+                "precompile cache labeled counter",
+                "precompile_name",
+                "event");
+
+    // set a cache event consumer which logs a metrics event
+    AbstractPrecompiledContract.setCacheEventConsumer(
+        cacheEvent ->
+            precompileCounter
+                .labels(cacheEvent.precompile(), cacheEvent.cacheMetric().name())
+                .inc());
   }
 
   private void checkPermissionsAndPrintPaths(final String userName) {
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -983,7 +983,7 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
 
       configureNativeLibs(Optional.ofNullable(network));
       if (enablePrecompileCaching
-          && getDataStorageConfiguration()
+          || getDataStorageConfiguration()
               .getPathBasedExtraStorageConfiguration()
               .getUnstable()
               .isParallelTxProcessingEnabled()) {
@@ -1019,6 +1019,8 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   private void configurePrecompileCaching() {
     // enable precompile caching:
     AbstractPrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
+    // separately set KZG precompile to cache, it does not extend AbstractPrecompiledContract:
+    KZGPointEvalPrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
 
     // set a metric logger
     final var precompileCounter =
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -984,7 +984,6 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
 
       configureNativeLibs(Optional.ofNullable(network));
       if (enablePrecompileCaching) {
-        // enable precompile caching if it is enabled and parallel tx processing is enabled:
         configurePrecompileCaching();
       }
 
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -700,7 +700,7 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   @CommandLine.Option(
       names = {"--cache-precompiles"},
       description = "Specifies whether to cache precompile results (default: ${DEFAULT-VALUE})")
-  private final Boolean enablePrecompileCaching = true;
+  private final Boolean enablePrecompileCaching = false;
 
   // Plugins Configuration Option Group
   @CommandLine.ArgGroup(validate = false)
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -982,11 +982,7 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
       VersionMetadata.versionCompatibilityChecks(versionCompatibilityProtection, dataDir());
 
       configureNativeLibs(Optional.ofNullable(network));
-      if (enablePrecompileCaching
-          || getDataStorageConfiguration()
-              .getPathBasedExtraStorageConfiguration()
-              .getUnstable()
-              .isParallelTxProcessingEnabled()) {
+      if (enablePrecompileCaching) {
         // enable precompile caching if it is enabled and parallel tx processing is enabled:
         configurePrecompileCaching();
       }
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -139,6 +139,7 @@ import org.hyperledger.besu.ethereum.worldstate.ImmutableDataStorageConfiguratio
 import org.hyperledger.besu.ethereum.worldstate.ImmutablePathBasedExtraStorageConfiguration;
 import org.hyperledger.besu.ethereum.worldstate.PathBasedExtraStorageConfiguration;
 import org.hyperledger.besu.evm.precompile.AbstractAltBnPrecompiledContract;
+import org.hyperledger.besu.evm.precompile.AbstractPrecompiledContract;
 import org.hyperledger.besu.evm.precompile.BigIntegerModularExponentiationPrecompiledContract;
 import org.hyperledger.besu.evm.precompile.KZGPointEvalPrecompiledContract;
 import org.hyperledger.besu.metrics.BesuMetricCategory;
@@ -695,6 +696,11 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
       description = "Specifies the number of last blocks to cache  (default: ${DEFAULT-VALUE})")
   private final Integer numberOfBlocksToCache = 0;
 
+  @CommandLine.Option(
+      names = {"--cache-precompiles"},
+      description = "Specifies whether to cache precompile results (default: ${DEFAULT-VALUE})")
+  private final Boolean enablePrecompileCaching = true;
+
   // Plugins Configuration Option Group
   @CommandLine.ArgGroup(validate = false)
   PluginsConfigurationOptions pluginsConfigurationOptions = new PluginsConfigurationOptions();
@@ -976,6 +982,8 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
       VersionMetadata.versionCompatibilityChecks(versionCompatibilityProtection, dataDir());
 
       configureNativeLibs(Optional.ofNullable(network));
+      AbstractPrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
+
       besuController = buildController();
 
       besuPluginContext.beforeExternalServices();
--- a/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
@@ -139,6 +139,7 @@ import org.hyperledger.besu.ethereum.worldstate.ImmutableDataStorageConfiguratio
 import org.hyperledger.besu.ethereum.worldstate.ImmutablePathBasedExtraStorageConfiguration;
 import org.hyperledger.besu.ethereum.worldstate.PathBasedExtraStorageConfiguration;
 import org.hyperledger.besu.evm.precompile.AbstractAltBnPrecompiledContract;
+import org.hyperledger.besu.evm.precompile.AbstractBLS12PrecompiledContract;
 import org.hyperledger.besu.evm.precompile.AbstractPrecompiledContract;
 import org.hyperledger.besu.evm.precompile.BigIntegerModularExponentiationPrecompiledContract;
 import org.hyperledger.besu.evm.precompile.KZGPointEvalPrecompiledContract;
@@ -1015,8 +1016,10 @@ public class BesuCommand implements DefaultCommandValues, Runnable {
   private void configurePrecompileCaching() {
     // enable precompile caching:
     AbstractPrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
-    // separately set KZG precompile to cache, it does not extend AbstractPrecompiledContract:
+    // separately set KZG precompile caching, it does not extend AbstractPrecompiledContract:
     KZGPointEvalPrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
+    // separately set BLS precompiles caching, they do not extend AbstractPrecompiledContract:
+    AbstractBLS12PrecompiledContract.setPrecompileCaching(enablePrecompileCaching);
 
     // set a metric logger
     final var precompileCounter =
--- a/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
+++ b/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
@@ -18,6 +18,7 @@ import static org.hyperledger.besu.evmtool.BenchmarkSubCommand.COMMAND_NAME;
 import static picocli.CommandLine.ScopeType.INHERIT;
 
 import org.hyperledger.besu.BesuInfo;
+import org.hyperledger.besu.evm.precompile.AbstractPrecompiledContract;
 import org.hyperledger.besu.evmtool.benchmarks.AltBN128Benchmark;
 import org.hyperledger.besu.evmtool.benchmarks.BLS12Benchmark;
 import org.hyperledger.besu.evmtool.benchmarks.BenchmarkExecutor;
@@ -75,6 +76,13 @@ public class BenchmarkSubCommand implements Runnable {
       negatable = true)
   Boolean nativeCode;
 
+  @Option(
+      names = {"--use-precompile-cache"},
+      description = "Benchmark using precompile caching.",
+      scope = INHERIT,
+      negatable = true)
+  Boolean enablePrecompileCache = false;
+
   @Parameters(description = "One or more of ${COMPLETION-CANDIDATES}.")
   EnumSet<Benchmark> benchmarks = EnumSet.noneOf(Benchmark.class);
 
@@ -99,6 +107,7 @@ public class BenchmarkSubCommand implements Runnable {
   public void run() {
     LogConfigurator.setLevel("", "DEBUG");
     System.out.println(BesuInfo.version());
+    AbstractPrecompiledContract.setPrecompileCaching(enablePrecompileCache);
     var benchmarksToRun = benchmarks.isEmpty() ? EnumSet.allOf(Benchmark.class) : benchmarks;
     for (var benchmark : benchmarksToRun) {
       System.out.println("Benchmarks for " + benchmark);
--- a/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
+++ b/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
@@ -18,6 +18,7 @@ import static org.hyperledger.besu.evmtool.BenchmarkSubCommand.COMMAND_NAME;
 import static picocli.CommandLine.ScopeType.INHERIT;
 
 import org.hyperledger.besu.BesuInfo;
+import org.hyperledger.besu.evm.precompile.AbstractBLS12PrecompiledContract;
 import org.hyperledger.besu.evm.precompile.AbstractPrecompiledContract;
 import org.hyperledger.besu.evmtool.benchmarks.AltBN128Benchmark;
 import org.hyperledger.besu.evmtool.benchmarks.BLS12Benchmark;
@@ -108,6 +109,7 @@ public class BenchmarkSubCommand implements Runnable {
     LogConfigurator.setLevel("", "DEBUG");
     System.out.println(BesuInfo.version());
     AbstractPrecompiledContract.setPrecompileCaching(enablePrecompileCache);
+    AbstractBLS12PrecompiledContract.setPrecompileCaching(enablePrecompileCache);
     var benchmarksToRun = benchmarks.isEmpty() ? EnumSet.allOf(Benchmark.class) : benchmarks;
     for (var benchmark : benchmarksToRun) {
       System.out.println("Benchmarks for " + benchmark);
--- a/evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractBLS12PrecompiledContract.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractBLS12PrecompiledContract.java
@@ -147,9 +147,6 @@ public abstract class AbstractBLS12PrecompiledContract implements PrecompiledCon
       }
     }
 
-
-
-
     final byte[] result = new byte[LibGnarkEIP2537.EIP2537_PREALLOCATE_FOR_RESULT_BYTES];
     final byte[] error = new byte[LibGnarkEIP2537.EIP2537_PREALLOCATE_FOR_ERROR_BYTES];
 
@@ -159,19 +156,30 @@ public abstract class AbstractBLS12PrecompiledContract implements PrecompiledCon
         new IntByReference(LibGnarkEIP2537.EIP2537_PREALLOCATE_FOR_ERROR_BYTES);
 
     final int inputSize = Math.min(inputLimit, input.size());
-    final int errorNo = LibGnarkEIP2537.eip2537_perform_operation(operationId,
-        input.slice(0, inputSize).toArrayUnsafe(), inputSize, result, o_len, error, err_len);
+    final int errorNo =
+        LibGnarkEIP2537.eip2537_perform_operation(
+            operationId,
+            input.slice(0, inputSize).toArrayUnsafe(),
+            inputSize,
+            result,
+            o_len,
+            error,
+            err_len);
 
     if (errorNo == 0) {
-      res = new PrecompileInputResultTuple(enableResultCaching ? input.copy() : input,
-          PrecompileContractResult.success(Bytes.wrap(result, 0, o_len.getValue())));
+      res =
+          new PrecompileInputResultTuple(
+              enableResultCaching ? input.copy() : input,
+              PrecompileContractResult.success(Bytes.wrap(result, 0, o_len.getValue())));
     } else {
       final String errorMessage = new String(error, 0, err_len.getValue(), UTF_8);
       messageFrame.setRevertReason(Bytes.wrap(error, 0, err_len.getValue()));
       LOG.trace("Error executing precompiled contract {}: '{}'", name, errorMessage);
-      res = new PrecompileInputResultTuple(enableResultCaching ? input.copy() : input,
-          PrecompileContractResult.halt(null,
-              Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR)));
+      res =
+          new PrecompileInputResultTuple(
+              enableResultCaching ? input.copy() : input,
+              PrecompileContractResult.halt(
+                  null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR)));
     }
 
     if (enableResultCaching && cacheKey != null) {
@@ -227,5 +235,10 @@ public abstract class AbstractBLS12PrecompiledContract implements PrecompiledCon
     enableResultCaching = enablePrecompileCaching;
   }
 
+  /**
+   * get the presompile-specific cache.
+   *
+   * @return precompile cache.
+   */
   protected abstract Cache<Integer, PrecompileInputResultTuple> getCache();
 }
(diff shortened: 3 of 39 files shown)


evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128AddPrecompiledContract.java
```diff
@@ -74,11 +78,24 @@ public long gasRequirement(final Bytes input) {
   @Override
   public PrecompileContractResult computePrecompile(
       final Bytes input, @Nonnull final MessageFrame messageFrame) {
+
+    PrecompileInputResultTuple res;
+
+    if (enableResultCaching) {
+      res = bnAddCache.getIfPresent(input.hashCode());
```
ahamlat, line 85: As discussed before, hashCode is not a good key here as you may have collisions. We need to use either the whole input or a hashing function that assumes no collisions.

ahamlat (changes requested):
I think it makes sens to have a cache per precompile as we discussed. Also you need to change the key to use a hashing function that has no collisions.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
```diff
@@ -75,6 +76,13 @@ enum Benchmark {
       negatable = true)
   Boolean nativeCode;
 
+  @Option(
+      names = {"--use-precompile-cache"},
```
ahamlat, line 80: Nice to have a flag for this feature

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128MulPrecompiledContract.java
```diff
@@ -40,6 +42,8 @@ public class AltBN128MulPrecompiledContract extends AbstractAltBnPrecompiledCont
 
   private static final Bytes POINT_AT_INFINITY = Bytes.repeat((byte) 0, 64);
   private final long gasCost;
+  private static final Cache<Integer, PrecompileInputResultTuple> bnMulCache =
+      Caffeine.newBuilder().maximumSize(1000).build();
```
ahamlat, line 46: Not really urgent or mandatory but it would be nice to add metrics for it to have the cache hit ratio. Check : https://github.com/hyperledger/besu/blob/8cddcfd9550f37bd5a78888254bf7ac71bd3fdea/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/cache/BonsaiCachedMerkleTrieLoader.java#L43

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128AddPrecompiledContract.java
garyschulte: I was expecting and embracing collisions.  I think a cache limited to <=1000 items is unlikely to have collisions in the int space unless it is an intentional attempt to create collisions.  In that case, the equals() check will prevent using the cache and it will be overwritten with the new computed value.

If we use input Bytes, it seems we can attack the cache by filling it with collisions and cause a worst case performance where we have to use byte-by-byte equals() method to distinguish between the hashCode collisions.  

For that reason I think leaning into a 1 cache item per hash code is a safer strategy.   It will work well with few naturally occurring collisions, and won't result in the ability to attack the cache and cause slow blocks.

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128MulPrecompiledContract.java
garyschulte: Good idea.  I will add that

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128AddPrecompiledContract.java
daniellehrner: In another caching test PR I've used [xxHash](https://xxhash.com/) to create the cache key, which is supposed to be very fast.

Here's the PR where I've used it:
https://github.com/daniellehrner/besu/commit/c75f975c3753fba4bd8346345c04462a5948baf7#diff-006f4a4dd9869817685273bb0bc45923b06e6a926661e7ac84615564576e9060R146

ahamlat: Now that I understand that you have a second check on the input, I think it is a pretty smart method, I'm interested to see what the profiling show, basically what is the overhead of the equals on the input when there is a collision.
When there no collision, it is pretty good algorithm complexity. We can also play with the size of the cache to reduce collisions, and check if the memory overhead is not too big. This will depend on the input size, which not fixed I guess.

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128MulPrecompiledContract.java
garyschulte: Added hit/miss/false positive counters per precompile.  ECRECOVER has 0% false positive, but eip 196, 197, and blake2 seem to need a better hashcode implementation to limit the false positives.

garyschulte: for blake2 and altbn128 methods, the false positive vs hit ratio is not good:
image

garyschulte: ecrecover, which is hit much more frequently has zero false positives, and a good enough cache hit ratio to make it worthwhile:
image

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -696,6 +697,11 @@ static class PrivacyOptionGroup {
       description = "Specifies the number of last blocks to cache  (default: ${DEFAULT-VALUE})")
   private final Integer numberOfblocksToCache = 0;
 
+  @CommandLine.Option(
+      names = {"--cache-precompiles"},
+      description = "Specifies whether to cache precompile results (default: ${DEFAULT-VALUE})")
+  private final Boolean enablePrecompileCaching = true;
```
ahamlat, line 702: Can we disabled on default and enabled on demand, for a performance profile on block processing ?

evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractPrecompiledContract.java
```diff
@@ -53,4 +55,45 @@ protected GasCalculator gasCalculator() {
...
+   * @param enablePrecompileCaching boolean indicating whether to cache precompile results
+   */
+  public static void setPrecompileCaching(final boolean enablePrecompileCaching) {
+    enableResultCaching = enablePrecompileCaching;
+  }
+
+  /** enum for precompile cache metric */
+  public enum CacheMetric {
+    /** a successful cache hit metric */
+    HIT,
```
ahamlat, line 74: Can we make these lowercase, as they are used in the name of the metrics, may be implement a toString method.
Currently the name of besu metrics dans have uppercase names.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
```diff
@@ -976,6 +982,15 @@ public void run() {
       VersionMetadata.versionCompatibilityChecks(versionCompatibilityProtection, dataDir());
 
       configureNativeLibs(Optional.ofNullable(network));
+      if (enablePrecompileCaching
+          && getDataStorageConfiguration()
+              .getPathBasedExtraStorageConfiguration()
+              .getUnstable()
+              .isParallelTxProcessingEnabled()) {
```
garyschulte, line 989: right now this feature is only enabled if we have parallelism enabled

evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractPrecompiledContract.java
ahamlat: Checking the way the metric is exposed, this is part of the metric attributes, so it is fine to keep it as it is.
We get the metric value with this kind of formula `besu_block_processing_precompile_cache_total{instance =~"dev-elc-bu-tk-mainnet-01", event="HIT", precompile_name="ECREC"}`.

evm/src/main/java/org/hyperledger/besu/evm/precompile/BLAKE2BFPrecompileContract.java
```diff
@@ -77,6 +83,31 @@ public PrecompileContractResult computePrecompile(
...
+        cacheEventConsumer.accept(new CacheEvent(PRECOMPILE_NAME, CacheMetric.MISS));
+      }
+    }
+
+    res =
+        new PrecompileInputResultTuple(
+            input, PrecompileContractResult.success(Hash.blake2bf(input)));
+
+    if (cacheKey != null) {
+      blakeCache.put(input.hashCode(), res);
```
ahamlat, line 108: I suggest using cacheKey as done in other precompiles instead of calculating the hashcode  again, and it is not done the same way.

evm/src/main/java/org/hyperledger/besu/evm/precompile/ECRECPrecompiledContract.java
```diff
@@ -97,13 +128,22 @@ public PrecompileContractResult computePrecompile(
       final Optional<SECPPublicKey> recovered =
           signatureAlgorithm.recoverPublicKeyFromSignature(h, signature);
       if (recovered.isEmpty()) {
-        return PrecompileContractResult.success(Bytes.EMPTY);
+        res =
+            new PrecompileInputResultTuple(
+                nonMutatedInput, PrecompileContractResult.success(Bytes.EMPTY));
+        ecrecCache.put(nonMutatedInput.hashCode(), res);
```
ahamlat, line 134: We can calculate the hashcode only once above, as done in other precompiles.

ahamlat (changes requested):
Could you review the way the hashcode is calculate for some precompiles (see below) ? It is sometimes calculated twice where we can store it and use it to cache the result. Also, I wonder why you didn't add a cache for KZGPointEvalPrecompiledContract.

ahamlat:
Other than the small requested changes above, the PR is great, I think we can decouple it from parallel transaction execution as it can help even when parallel transaction execution is not enabled.
In terms of performance, in addition to what @garyschulte shared, profiling two nodes running this PR against two nodes running version 25.2.0 showed that the cache works fine, especially for EcRecover, and it reduce precompile execution, reducing the number of samples for a period of 300 second sampling from ~18 to 1 samples in both cases. 
A sample is collected each 11 ms. The profiling is done on the same blocks.

**Without this PR**
image

**With this PR**
image

ahamlat:
An interesting idea from this implementation, is that you created a cache where the eviction mechanism is based on hashcode collisions. It is like a hashmap where we keep only one node behind each bucket (hashcode index), the new value will alway replace the existing one.

evm/src/main/java/org/hyperledger/besu/evm/precompile/BLAKE2BFPrecompileContract.java
garyschulte: yeah this might be a problem if the input is getting mutated.  I missed this

garyschulte:
> Could you review the way the hashcode is calculate for some precompiles (see below) ? It is sometimes calculated twice where we can store it and use it to cache the result. Also, I wonder why you didn't add a cache for KZGPointEvalPrecompiledContract.

Will do, and will add similar caches for the pectra BLS precompiles.

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
garyschulte: I enabled it by default, with or without parallel processing.  Do you want to make it disabled by default?

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
garyschulte: removed this requirement

evm/src/main/java/org/hyperledger/besu/evm/precompile/ECRECPrecompiledContract.java
garyschulte: AbstractBytes.hashCode() caches but MutableBytes.hashCode() does not.  Several of the inputs I looked at were actually MutableBytes, so I made all of the precompiles pre-calculate and hang onto the cacheKey as you suggested 👍

evm/src/main/java/org/hyperledger/besu/evm/precompile/ECRECPrecompiledContract.java
```diff
@@ -78,6 +87,28 @@ public PrecompileContractResult computePrecompile(
...
+
+    if (enableResultCaching) {
+      res = ecrecCache.getIfPresent(nonMutatedInput.hashCode());
+
+      if (res != null) {
+        if (res.cachedInput().equals(nonMutatedInput)) {
+          cacheEventConsumer.accept(new CacheEvent(PRECOMPILE_NAME, CacheMetric.HIT));
+          return res.cachedResult();
+        } else {
+          LOG.info(
```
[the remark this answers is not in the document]
garyschulte: done.  I also added debug logging for all other false negatives, since they shouldn't happen now.

garyschulte:
OK, I found the issues with false positives.  It appears we are subsequently mutating the input bytes.  So the input value in our precompile result tuple was getting mutated after we cached it, and that was the source of the false positives.

What I have done for all precompiles is to copy the input in the precompile result tuple IF we have caching enabled.  This seems to be the sweet spot for removing false positives.  We should have little to no impact on precompile performance or overhead if caching is disabled.  

Since these changes, I have not seen any false positives:
image


If you would, please re-review at your leisure @ahamlat  🙏

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
ahamlat: Yes in the first place, we can make it default after some time of running it.

ahamlat: I understand that the feature will be enabled either with the parallelization flag, or when `--cache-precompiles` is set to true, correct ? 
So basically, even if `--cache-precompiles=false`, we can still have precompile caching enabled. I think it can be confusing for the users. Let's make it simple. 
I think `--cache-precompiles` should be the only flag that can define if this feature will enable it or not.

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128AddPrecompiledContract.java
```diff
@@ -91,16 +93,27 @@ public PrecompileContractResult computePrecompile(
...
         }
       } else {
         cacheEventConsumer.accept(new CacheEvent(PRECOMPILE_NAME, CacheMetric.MISS));
       }
     }
     if (useNative) {
-      res = new PrecompileInputResultTuple(input, computeNative(input, messageFrame));
+      res =
+          new PrecompileInputResultTuple(
+              enableResultCaching ? input.copy() : input, computeNative(input, messageFrame));
```
ahamlat, line 112: Will check the impact on performance of the copy

evm/src/main/java/org/hyperledger/besu/evm/precompile/ECRECPrecompiledContract.java
```diff
@@ -88,20 +87,22 @@ public PrecompileContractResult computePrecompile(
     }
 
     PrecompileInputResultTuple res;
-
+    Integer cacheKey = null;
     if (enableResultCaching) {
-      res = ecrecCache.getIfPresent(nonMutatedInput.hashCode());
+      cacheKey = input.hashCode();
```
ahamlat, line 92: can we use getCacheKey here ? why using hashcode instead of Arrays.hashCode(input.toArrayUnsafe());

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128MulPrecompiledContract.java
```diff
@@ -148,4 +162,8 @@ private static BigInteger extractParameter(
     final byte[] raw = Arrays.copyOfRange(input.toArrayUnsafe(), offset, offset + length);
     return new BigInteger(1, raw);
   }
+
+  private static Integer getCacheKey(final Bytes input) {
```
ahamlat, line 166: I suggest to define it only once in `AbstractPrecompiledContract.java` and reference it in all the cached precompiles.

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128AddPrecompiledContract.java
```diff
@@ -74,11 +81,44 @@ public long gasRequirement(final Bytes input) {
   @Override
   public PrecompileContractResult computePrecompile(
       final Bytes input, @Nonnull final MessageFrame messageFrame) {
+
+    PrecompileInputResultTuple res;
+    Integer cacheKey = null;
+
+    if (enableResultCaching) {
+      cacheKey = Arrays.hashCode(input.toArrayUnsafe());
```
ahamlat, line 89: Can we use getCacheKey here ?

evm/src/main/java/org/hyperledger/besu/evm/precompile/BLAKE2BFPrecompileContract.java
```diff
@@ -77,6 +84,42 @@ public PrecompileContractResult computePrecompile(
...
+          cacheEventConsumer.accept(new CacheEvent(PRECOMPILE_NAME, CacheMetric.FALSE_POSITIVE));
+        }
+      } else {
+        cacheEventConsumer.accept(new CacheEvent(PRECOMPILE_NAME, CacheMetric.MISS));
+      }
+    }
+
+    res =
+        new PrecompileInputResultTuple(
+            input.copy(), PrecompileContractResult.success(Hash.blake2bf(input)));
```
ahamlat, line 113: I think it is still worth checking if enableResultCaching is enabled before doing the copy.

evm/src/main/java/org/hyperledger/besu/evm/precompile/KZGPointEvalPrecompiledContract.java
```diff
@@ -112,6 +133,38 @@ public PrecompileContractResult computePrecompile(
       return PrecompileContractResult.halt(
           null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
     }
+
+    PrecompileInputResultTuple res;
+    Integer cacheKey = null;
+
+    if (enableResultCaching) {
+      cacheKey = Arrays.hashCode(input.toArrayUnsafe());
```
ahamlat, line 141: I suggest using getCacheKey static method here.

ahamlat (commented):
LGTM, just few comments.
I will approve once I checked the impact of the copy of the input.

ahamlat (approved):
Could you update the benchmarks in the description with the results from the last implementation ?
The CPU profiling shows an improvement on the nodes running this PR :
**Control node 1** / 11 samples for `MessageCallProcessor.executePrecompile`
image

**Control node 2** / 11 samples for `MessageCallProcessor.executePrecompile`
image

**Node 1 running this PR** / 4 samples for `MessageCallProcessor.executePrecompile`
image

**Node 2 running this PR** / 4 samples for `MessageCallProcessor.executePrecompile`
image

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128MulPrecompiledContract.java
garyschulte: Seems reasonable - I will define it there and override it for ECRECOVER.  I think that is the only one that is different

garyschulte: actually I think ECREC can use the same cache function

evm/src/main/java/org/hyperledger/besu/evm/precompile/ECRECPrecompiledContract.java
garyschulte: ECREC was the first precompile "hash collision" I looked into, hashCode was appropriate for it.  But I agree there is no reason not to use the Arrays.hashCode for it also

garyschulte:
Updated the pdf docs.  Input copying doesn't seem to be a big problem in this case.  The ecmul point-at-infinity optimization performs better without the cache - interestingly even when we run the optimization check before the cache check.  But otherwise the current implementation looks to be better all around.

garyschulte:
Added bls precompile caching also.  The gas-benchmarks suite doesn't have bls precompile tests, but evmtool gives some pretty dramatic results:

```
➜  besu git:(feature/precompile-caching-part1) ✗ build/install/besu/bin/evmtool benchmark --native --use-precompile-cache bls12      
besu/v25.3-develop-f4df214/osx-aarch_64/corretto-java-22
Benchmarks for Bls12
Bls12 G1 Add    375 avg gas @    1.1 µs /   332.3 MGps
Bls12 G1 MSM   501,672 total gas @   13.2 µs /38,046.9 MGps
Bls12 MapFpToG1  5,500 avg gas @    0.5 µs /10,890.2 MGps
Bls12 G2 Add    600 avg gas @    0.9 µs /   697.4 MGps
Bls12 G2 MSM   991,080 total gas @   16.6 µs /59,835.1 MGps
Bls12 MapFp2G1 23,800 avg gas @    0.1 µs /184,270.0 MGps
Bls12 Pairing 2,209,700 total gas @   20.7 µs /106,567.6 MGps
➜  besu git:(feature/precompile-caching-part1) ✗ build/install/besu/bin/evmtool benchmark --native --use-precompile-cache=false bls12
besu/v25.3-develop-f4df214/osx-aarch_64/corretto-java-22
Benchmarks for Bls12
Bls12 G1 Add    375 avg gas @    5.6 µs /    66.9 MGps
Bls12 G1 MSM   501,672 total gas @4,155.8 µs /   120.7 MGps
Bls12 MapFpToG1  5,500 avg gas @   47.9 µs /   114.7 MGps
Bls12 G2 Add    600 avg gas @    6.3 µs /    95.1 MGps
Bls12 G2 MSM   991,080 total gas @6,888.4 µs /   143.9 MGps
Bls12 MapFp2G1 23,800 avg gas @  226.9 µs /   104.9 MGps
Bls12 Pairing 2,209,700 total gas @22,388.5 µs /    98.7 MGps
```

ahamlat:
Also I don't see metrics for `BLS12_G1ADD`, `BLS12_G1MULTIEXP`, `BLS12_G2ADD`, `BLS12_G2MULTIEXP`, `BLS12_MAP_FIELD_TO_CURVE`, `BLS12_PAIRING`. I guess, it is because these precompiles are not called, but it is worth double checking.

garyschulte:
> Also I don't see metrics for `BLS12_G1ADD`, `BLS12_G1MULTIEXP`, `BLS12_G2ADD`, `BLS12_G2MULTIEXP`, `BLS12_MAP_FIELD_TO_CURVE`, `BLS12_PAIRING`. I guess, it is because these precompiles are not called, but it is worth double checking.

BLS doesn't go live until pectra, thus no stats yet...

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128PairingPrecompiledContract.java
```diff
@@ -32,13 +32,25 @@
...
-
+  private static final Logger LOG =
+      LoggerFactory.getLogger(AltBN128PairingPrecompiledContract.class);
   private static final int FIELD_LENGTH = 32;
   private static final int PARAMETER_LENGTH = 192;
+  private static final String PRECOMPILE_NAME = "AltBN128Pairing";
+
+  private static final Cache<Integer, PrecompileInputResultTuple> bnPairingCache =
+      Caffeine.newBuilder()
+          .maximumWeight(16_000_000)
```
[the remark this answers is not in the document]
garyschulte: the pairing inputs can be really large.  I should implement the weighting strategy for bls pairing, and msm precompiles also.  I just did a copy/paste of the cache definition 🤔

evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractBLS12PrecompiledContract.java
```diff
@@ -128,14 +167,25 @@ public PrecompileContractResult computePrecompile(
...
-          null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
+      res =
+          new PrecompileInputResultTuple(
+              enableResultCaching ? input.copy() : input,
+              PrecompileContractResult.halt(
+                  null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR)));
+    }
+
+    if (enableResultCaching && cacheKey != null) {
+      getCache().put(cacheKey, res);
```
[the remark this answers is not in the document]
garyschulte: the work the precompile has to do is pretty similar the same if there is an error versus not an error.  Subgroup checks are reasonably expensive for example.  I was thinking any result for a deterministic input should be cached, regardless of the result.

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128AddPrecompiledContract.java
```diff
@@ -74,11 +81,44 @@ public long gasRequirement(final Bytes input) {
   @Override
   public PrecompileContractResult computePrecompile(
       final Bytes input, @Nonnull final MessageFrame messageFrame) {
+
+    PrecompileInputResultTuple res;
+    Integer cacheKey = null;
+
+    if (enableResultCaching) {
+      cacheKey = getCacheKey(input);
```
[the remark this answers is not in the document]
garyschulte: this is a static method in AbstractPrecompiledContract

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128MulPrecompiledContract.java
```diff
@@ -85,11 +93,43 @@ public PrecompileContractResult computePrecompile(
           POINT_AT_INFINITY, false, MessageFrame.State.COMPLETED_SUCCESS, Optional.empty());
     }
 
+    PrecompileInputResultTuple res;
+    Integer cacheKey = null;
+    if (enableResultCaching) {
+      cacheKey = getCacheKey(input);
```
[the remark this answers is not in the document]
garyschulte: same, this is a static method in AbstractPrecompiledContract

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128PairingPrecompiledContract.java
```diff
@@ -99,11 +111,42 @@ public PrecompileContractResult computePrecompile(
       return PrecompileContractResult.halt(
           null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
     }
+    PrecompileInputResultTuple res;
+    Integer cacheKey = null;
+    if (enableResultCaching) {
+      cacheKey = getCacheKey(input);
```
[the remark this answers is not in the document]
garyschulte: same, this is a static method in AbstractPrecompiledContract

evm/src/main/java/org/hyperledger/besu/evm/precompile/BLAKE2BFPrecompileContract.java
```diff
@@ -77,6 +83,39 @@ public PrecompileContractResult computePrecompile(
       return PrecompileContractResult.halt(
           null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
     }
-    return PrecompileContractResult.success(Hash.blake2bf(input));
+    PrecompileInputResultTuple res = null;
+    Integer cacheKey = null;
+    if (enableResultCaching) {
+      cacheKey = getCacheKey(input);
```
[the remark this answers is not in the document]
garyschulte: same, this is a static method in AbstractPrecompiledContract

evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractBLS12PrecompiledContract.java
```diff
@@ -128,14 +167,25 @@ public PrecompileContractResult computePrecompile(
...
+              enableResultCaching ? input.copy() : input,
+              PrecompileContractResult.success(Bytes.wrap(result, 0, o_len.getValue())));
     } else {
       final String errorMessage = new String(error, 0, err_len.getValue(), UTF_8);
       messageFrame.setRevertReason(Bytes.wrap(error, 0, err_len.getValue()));
       LOG.trace("Error executing precompiled contract {}: '{}'", name, errorMessage);
-      return PrecompileContractResult.halt(
-          null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
+      res =
+          new PrecompileInputResultTuple(
```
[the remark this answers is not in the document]
garyschulte: I think we need to at least for the BLS precompiles, because the amount of work that might happen and still return an error is non-trivial.

garyschulte:
> Sharing the metrics on the configured precompiles on Ethereum mainnet, after ~40 minutes of executions. We're using counters, so this is the hit ratio for the 40 minutes of execution. From the metrics, we can reconsider at least enabling the cache on KZGPointEval, as the hit ratio is 0. It was suggestion from my side to enable caching on KZGPointEval, but the metrics show that it is not a good candidate for caching.

Over 4 days, I see a pretty low hit ratio:

nuc 8: 198 / 2665
nuc 14: 146 / 2594

how low of a ratio makes it not worth it ?

ahamlat:
> > Sharing the metrics on the configured precompiles on Ethereum mainnet, after ~40 minutes of executions. We're using counters, so this is the hit ratio for the 40 minutes of execution. From the metrics, we can reconsider at least enabling the cache on KZGPointEval, as the hit ratio is 0. It was suggestion from my side to enable caching on KZGPointEval, but the metrics show that it is not a good candidate for caching.
> 
> Over 4 days, I see a pretty low hit ratio:
> 
> nuc 8: 198 / 2665 nuc 14: 146 / 2594
> 
> how low of a ratio makes it not worth it ?

It depends on :
* The cost of checking if the entry is on the cache and adding it to the cache (i.e overhead of caching) 
**vs** 
* The cost of the precompile execution itself. So in the case of the metrics you shared, the cache has an overhead on 2665 calls and could avoid the execution of only 198 precompile calls. 

The overhead here is calculating the hashcode on the input byte array, and checking if the integer key (hashcode result) is in the cache. There is also the overhead of adding the new execution result to the cache. So in this case, we're improving 7% of the calls and generating an overhead on 93% of the calls. 
This kind of cache can still be interesting if the execution of the precompile is slow, let me get more metrics on `KZGPointEval`.

evm/src/main/java/org/hyperledger/besu/evm/precompile/AltBN128AddPrecompiledContract.java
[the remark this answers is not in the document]
ahamlat: Yes, but not visible from the caller point of view. I'm fine of keeping it as it is, but just we need to do the same everywhere, see `AbstractBLS12PrecompiledContract.java`

evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractBLS12PrecompiledContract.java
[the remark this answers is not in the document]
ahamlat: Ok understood, how about adding in the cache only when the result is not returned from the cache (a cache miss).

ahamlat:
> 
> It depends on :
> 
> * The cost of checking if the entry is on the cache and adding it to the cache (i.e overhead of caching)
>   **vs**
> * The cost of the precompile execution itself.
> 
> So in the case of the metrics you shared, the cache has an overhead on 2665 calls and could avoid the execution of only 198 precompile calls. The overhead here is calculating the hashcode on the input byte array, and checking if the integer key (hashcode result) is in the cache. There is also the overhead of adding the new execution result to the cache. So in this case, we're improving 7% of the calls and generating an overhead on 93% of the calls. This kind of cache can still be interesting if the execution of the precompile is slow, let me get more metrics on `KZGPointEval`.

So the execution time of KZGPointEval call is around 500 us on my laptop, which pretty slow. Will suggest a PR tomorrow to add it to the existing benchmarks. I think even if it is only 7 % hit ratio, we can keep it.

PR #8096, opened 2025-01-09
Pass miningBeneficiary address to BlockAwareOperationTracer::traceStartBlock calls

### LINKED ISSUE #8094: Pass miningBeneficiary to BlockAwareOperationTracer::traceStartBlock
### Description
As a developer, I want to have access to the mining beneficiary address, in this case the coinbase address from the tracer API, so that I can view it before any balances or nonces are updated.

Currently the only way to have access to this is through the `MessageFrame`, but at this stage it will be too late for the use case as the balance and the nonce of the sender would have already been modified.

### Acceptance Criteria
A change in the tracer API is required -> BlockAwareOperationTracer::traceStartBlock has a new parameter that is the mining beneficiary.

lu-pinto:
## PR description
This passes the miningBeneficiary, which in most cases should be the coinbase but for other consensus approaches could be different, to BlockAwareOperationTracer::traceStartBlock calls.

## Fixed Issue(s)

fixes #8094

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
+++ b/besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
@@ -17,6 +17,7 @@ package org.hyperledger.besu.services;
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.hyperledger.besu.ethereum.mainnet.feemarket.ExcessBlobGasCalculator.calculateExcessBlobGasForParent;
 
+import org.hyperledger.besu.datatypes.Address;
 import org.hyperledger.besu.datatypes.BlobGas;
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.datatypes.Wei;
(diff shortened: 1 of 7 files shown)


besu/src/main/java/org/hyperledger/besu/services/TraceServiceImpl.java
```diff
@@ -191,7 +191,7 @@ private List<TransactionProcessingResult> trace(
     final ProtocolSpec protocolSpec = protocolSchedule.getByBlockHeader(block.getHeader());
     final MainnetTransactionProcessor transactionProcessor = protocolSpec.getTransactionProcessor();
     final BlockHeader header = block.getHeader();
-    tracer.traceStartBlock(block.getHeader(), block.getBody());
+    tracer.traceStartBlock(block.getHeader(), block.getBody(), block.getHeader().getCoinbase());
```
Gabriel-Trintinalia, line 194: You will need to use `miningBeneficiaryCalculator` here otherwise the `coinbase` will still be zero for Click

Gabriel-Trintinalia (changes requested):

Gabriel-Trintinalia (approved):

PR #8101, opened 2025-01-10
Increase the max number of prioritized blob txs in the layered txpool for Pectra

fab-10:
## PR description

Pectra fork increase the max number of blobs in a block to 9, so this PR allows for a max of 9 blob txs in the layered txpool so Besu can fill the blob space when building a block.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolConfiguration.java
```diff
@@ -75,7 +75,7 @@ enum Implementation {
   long DEFAULT_PENDING_TRANSACTIONS_LAYER_MAX_CAPACITY_BYTES = 12_500_000L;
   int DEFAULT_MAX_PRIORITIZED_TRANSACTIONS = 2000;
   EnumMap<TransactionType, Integer> DEFAULT_MAX_PRIORITIZED_TRANSACTIONS_BY_TYPE =
-      new EnumMap<>(Map.of(TransactionType.BLOB, 6));
+      new EnumMap<>(Map.of(TransactionType.BLOB, 9));
```
garyschulte, line 78: It would be nice to have an accompanying  unit test assertion.

garyschulte (approved):

siladu (commented):
Consider loading from the blobSchedule config that was added here https://github.com/hyperledger/besu/commit/e830db72480838f6ad2d18adc5b32ac59b72bcf3?

fab-10:
> Consider loading from the blobSchedule config that was added here [e830db7](https://github.com/hyperledger/besu/commit/e830db72480838f6ad2d18adc5b32ac59b72bcf3)?

that is a good point, but requires the configuration to be aware of the protocol schedule and to be updated at runtime to switch to the new value when the fork happens, and that requires a bit of thinking and effort to achieve, so for the moment I prefer to do this simple update and evaluate if it is worth to link it to genesis conf or not, also taking in account the possibility that in future that value is driven by the CL.

PR #8103, opened 2025-01-10
Improve debug_traceBlock calls performance

ahamlat:
## PR description
This PR is the same as this PR https://github.com/hyperledger/besu/pull/8076, I closed by error.

It introduces several key improvements to the debug_traceBlock, debug_traceBlockByNumber, and debug_traceBlockByHash methods, significantly enhancing both performance and memory usage:

The core change is the adoption of a pipeline architecture, where:

One thread handles transaction execution.
Other threads processes the execution results and generates the JSON output. This parallelization ensures better resource utilization and reduces overall execution time.
Memory usage is optimized by the pipeline processing approach. Instead of retaining all trace frames for every transaction, the pipeline now keeps only the trace frames relevant to the current transaction, significantly reducing memory consumption.

The transformation of trace frames into Java JSON representations now occurs in parallel.

Default trace options have been adjusted to align with the behavior of Geth and Nethermind:

disableStorage is set to false (enabling storage tracing).
disableMemory is set to true (disabling memory tracing).
disableStack is set to false (enabling stack tracing).
The structure of the JSON output has been modified to match Geth's format:

Hexadecimal values are now displayed as short hex strings (e.g., 0x1, 0xabc), which is more compact and efficient.
The stack array is always displayed when the option is enabled, even when empty, to ensure consistency with other Ethereum clients. However, memory and storage are omitted if they are empty, which reduces unnecessary data.
Storage keys and values are now displayed without the 0x prefix, in line with Geth's behavior.
Change the implementation of toShortHexString, by creating a new method toCompactHex. The main improvement is to not relay on Bytes, but on byte[] to avoid dynamic dispatch mechanism related to the use of methods get(i) and size(). Those methods are not used anymore thanks to the switch to StringBuilder, even if it has its own cons. This method will replace in the future toFastHex implementation, once we have finished the work on Tuweni.

These changes have resulted in a 3X performance improvement, reducing the size of the output by 2 to 3 times, making previously unresponsive calls fast enough with comparable performance with other clients. The optimizations not only improve response times but also allow for smoother operation with lower memory usage, even when processing large traces.

Overall, the new approach ensures better performance, reduced resource usage, and more consistent output with other Ethereum clients, particularly Geth and Nethermind.

## Fixed Issue(s)

Fixes https://github.com/hyperledger/besu/issues/5322


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Retrieve all transaction receipts for a block in one request [#6646](https://github.com/hyperledger/besu/pull/6646)
 - Implement EIP-7840: Add blob schedule to config files [#8042](https://github.com/hyperledger/besu/pull/8042)
 - Allow gasPrice (legacy) and 1559 gasPrice params to be specified simultaneously for `eth_call`, `eth_createAccessList`, and `eth_estimateGas` [#8059](https://github.com/hyperledger/besu/pull/8059)
+- Improve debug_traceBlock calls performance and reduce output size [#8076](https://github.com/hyperledger/besu/pull/8076)
```
garyschulte, line 23: It might be worth mentioning the default trace option behavior change in the 'breaking changes' section.  Just so there are no surprises.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/StructLog.java
```diff
@@ -153,4 +148,39 @@ public int hashCode() {
     result = 31 * result + Arrays.hashCode(stack);
     return result;
   }
+
+  public static String toCompactHex(final Bytes abytes, final boolean prefix) {
+    if (abytes.isEmpty()) {
+      if (prefix) return "0x0";
+      else return "0";
+    }
+    final char[] hexChars = HEX_DIGITS.toCharArray();
```
garyschulte, line 157: I think this can be a constant char[] rather than allocating a new array on each invocation

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/StructLogTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+public class StructLogTest {
```
garyschulte, line 42: 👍

ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/debug/trace-call/debug_traceCall_all.json
```diff
@@ -28,271 +28,163 @@
         "op" : "PUSH1",
         "gas" : 16755910,
         "gasCost" : 3,
-        "depth" : 1,
-        "stack" : null,
-        "memory" : null,
-        "storage" : null,
-        "reason" : null
+        "depth" : 1
```
garyschulte, line 31: this seems to have been an issue before, but given the name wouldn't we expect stack, memory, and storage in this trace?  I get why this is as it is, but the name is misleading

garyschulte (approved):
LGTM.  Some easy to fix suggestions in comments.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractDebugTraceBlock.java
```diff
@@ -0,0 +1,151 @@
...
+  private final LabelledMetric<Counter> outputCounter;
+  private final Supplier<BlockchainQueries> blockchainQueriesSupplier;
+  private final EthScheduler ethScheduler;
+
+  public AbstractDebugTraceBlock(
+      final ProtocolSchedule protocolSchedule,
+      final BlockchainQueries blockchainQueries,
+      final ObservableMetricsSystem metricsSystem,
+      final EthScheduler ethScheduler) {
+    this.blockchainQueriesSupplier = Suppliers.ofInstance(blockchainQueries);
```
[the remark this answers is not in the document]
ahamlat: Good question, I actually started with a final member reference than I want to have the same behaviour between debug_traceblockByNumber and debug_traceBlockByHash. DebugTraceBlockByNumber extends AbstractBlockParameterMethod which uses a supplier, so I did the same for debug_traceBlock and debug_traceBlockByHash.

ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/debug/trace-call/debug_traceCall_all.json
ahamlat: Yes, I agree the name is a bit weird. I guess "all" here means that all the fields are disabled as the request parameters looks like
    `  "disableMemory": true, "disableStack": true,  "disableStorage": true`
Where debug_traceCall_disableMemory.json has only
`      "disableMemory":true
`

ahamlat: I renamed debug_traceCall_complete.json to debug_traceCall_default.json, as by default now, we're not tracing memory.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/results/StructLog.java
ahamlat: Done

CHANGELOG.md
ahamlat: Done

PR #8113, opened 2025-01-13
Refactor get mutable methods of the worldtstate provider

matkt:
## PR description

This PR refactors Bonsai/DiffBased and its provider to simplify the future addition of Verkle and the stateless part. The goal is to maintain a logical code structure, ensuring easy integration of the stateless code.


matkt:
test in progress, not merge yet

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/forest/ForestWorldStateArchive.java
```diff
@@ -66,13 +67,16 @@ public boolean isWorldStateAvailable(final Hash rootHash, final Hash blockHash)
...
-  public Optional<MutableWorldState> getMutable(
-      final BlockHeader blockHeader, final boolean isPersistingState) {
-    return getMutable(blockHeader.getStateRoot(), blockHeader.getHash());
+  public Optional<MutableWorldState> getWorldState(final WorldStateQueryParams queryParams) {
+    return getWorldState(queryParams.getStateRoot());
   }
 
   @Override
-  public Optional<MutableWorldState> getMutable(final Hash rootHash, final Hash blockHash) {
+  public MutableWorldState getWorldState() {
```
garyschulte, line 75: would it be reasonable to rename this to getEmptyWorldState() ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/WorldStateStorageConfig.java
```diff
@@ -0,0 +1,56 @@
...
+    this.isFrozen = builder.isFrozen;
+  }
+
+  public boolean isFrozen() {
+    return isFrozen;
+  }
+
+  public static WorldStateStorageConfig.Builder newBuilder() {
+    return new WorldStateStorageConfig.Builder();
+  }
```
garyschulte, line 31: A class with a single boolean member var (for now) having a builder seems like overkill IMO.  A single static constructor would probably be sufficient.  Will Verkle or stateless extend this class with its own configs?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
```diff
@@ -108,12 +113,33 @@ public Hash getWorldStateRootHash() {
...
+   *
+   * <p>This method checks if the given world state key-value storage is an instance that can modify
+   * the head of the node. The head of the node will never be of type {@link
+   * DiffBasedSnapshotWorldStateKeyValueStorage}. These instances are copies of the head or of the
+   * head at a previous block or other points in time.
+   *
+   * @param worldStateKeyValueStorage the world state key-value storage to check.
+   * @return {@code true} if the provided storage is modifying the head, {@code false} otherwise.
+   */
+  private boolean isHeadModifyingWorldState(
```
garyschulte, line 141: should probably be renamed to isModifyingHeadWorldState() since it differentiates between modifications to worldstate or a snapshot of worldstate.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java
```diff
@@ -0,0 +1,239 @@
...
+  public Hash getBlockHash() {
+    return blockHash;
+  }
+
+  /**
+   * Gets the state root.
+   *
+   * @return the state root
+   */
+  public Hash getStateRoot() {
```
garyschulte, line 89: IMO we should return Optional<Hash> using Optional.ofNullable(stateRoot) since it isn't a required field.  Same for the other fields which are effectively Optional.  returning null can be surprising

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/DiffBasedWorldStateProvider.java
```diff
@@ -120,40 +119,122 @@ public Optional<WorldState> get(final Hash rootHash, final Hash blockHash) {
...
+   *
+   * <p>The method follows these steps: 1. Check if the query parameters indicate that the world
+   * state should update the head. 2. If true, call {@link #getStatefulWorldStateFromHead(Hash)}
+   * with the block hash from the query parameters. 3. If false, call {@link
+   * #getStatefulWorldStateFromCache(BlockHeader)} with the block header from the query parameters.
+   *
+   * @param queryParams the query parameters
+   * @return the stateful world state, if available
+   */
+  private Optional<MutableWorldState> getStatefulWorldState(
```
garyschulte, line 178: I think having Stateful in these method names is redundant.  The worldstate is state, and hence should be stateful.  We are trying to signify whether or not it is backed by a trie or a flat db versus a witness/proof?  

Maybe FullWorldState vs SparseWorldState or CompleteWorldState vs WitnessWorldState ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
```diff
@@ -115,7 +116,7 @@ public BlockProcessingResult validateAndProcessBlock(
       final Block block,
       final HeaderValidationMode headerValidationMode,
       final HeaderValidationMode ommerValidationMode,
-      final boolean shouldPersist,
+      final boolean shouldUpdateHead,
```
garyschulte, line 119: I like this new naming.  It conveys how dangerous that param really is for diff based worldstates.  Maybe even shouldUpdateGlobalState or similar that conveys it mutates the root worldstate upon which all diffs apply

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/DiffBasedWorldStateProvider.java
```diff
@@ -120,40 +119,122 @@ public Optional<WorldState> get(final Hash rootHash, final Hash blockHash) {
   @Override
   public boolean isWorldStateAvailable(final Hash rootHash, final Hash blockHash) {
     return cachedWorldStorageManager.contains(blockHash)
-        || persistedState.blockHash().equals(blockHash)
+        || headWorldState.blockHash().equals(blockHash)
         || worldStateKeyValueStorage.isWorldStateAvailable(rootHash, blockHash);
   }
 
+  /**
+   * Gets the mutable world state based on the provided query parameters.
```
garyschulte, line 127: nitpick, should be `Gets a mutable world state`.  "The" to me would imply the head worldstate

garyschulte (commented):
naming is hard.  I see your intent :+1:  Only thing I am not keen on is over use of the builder pattern for small classes

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/forest/ForestWorldStateArchive.java
matkt: I had a hard time choosing the name for this class because, for Forest, it's the empty state, whereas for Bonsai, it's the head without rollback. So I left it as getWorldState because I couldn't figure out how to align the two. I think once Forest is deprecated, we can rename it or maybe remove it.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/DiffBasedWorldState.java
matkt: yes done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/DiffBasedWorldStateProvider.java
matkt: I think I will do a mix and use FullWorldState vs WitnessWorldState. Seems to be more clear, because SparseWorldState doesn't seems to be a good one

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
matkt: I think head is better, but I'm not opposed to global state.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/DiffBasedWorldStateProvider.java
matkt: yes clearly better like that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java
matkt: right updated it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/WorldStateStorageConfig.java
matkt: yes I think it will be needed for stateless. I can remove it later if I don't use it in the future

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
lu-pinto: I like as is now

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateArchive.java
```diff
@@ -34,11 +35,26 @@ public interface WorldStateArchive extends Closeable {
...
 
-  MutableWorldState getMutable();
+  /**
+   * Gets the head world state.
+   *
+   * <p>This method returns the head world state, which is the most recent state of the world.
+   *
+   * @return the head world state
+   */
+  MutableWorldState getWorldState();
```
lu-pinto, line 57: So from what I understand, there's no way to get an immutable WorldState correct?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/PrivacyPluginPrecompiledContract.java
```diff
@@ -81,7 +83,9 @@ public PrecompileContractResult computePrecompile(
         privateStateRootResolver.resolveLastStateRoot(privacyGroupId, privateMetadataUpdater);
 
     final MutableWorldState disposablePrivateState =
-        privateWorldStateArchive.getMutable(lastRootHash, null).get();
```
lu-pinto, line 84: I would prefer not passing `null`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java
```diff
@@ -0,0 +1,239 @@
...
+
+  /**
+   * Should return a worldstate instance with a state root, block hash, and should update the node
+   * head.
+   *
+   * @param stateRoot the state root
+   * @param blockHash the block hash
+   * @return an instance of WorldStateQueryParams
+   */
+  public static WorldStateQueryParams withStateRootAndBlockHashAndUpdateNodeHead(
```
lu-pinto, line 136: I don't think this is future proof enough because the more parameters you add to the method the longer will be the name and less readable it will be. I was thinking and what about postponing calling `build()` using a lambda?
You then also have the benefit of postponing constructing any object until you need it:
```
  public interface WorldStateQueryParamsSupplier extends Supplier<WorldStateQueryParams> {
    WorldStateQueryParams.Builder createBuilder();

    default WorldStateQueryParams get() {
      return createBuilder().build();
    }
  }
``` 
Usage:
```
        worldStateArchive
            .getWorldState(() -> newBuilder()
                .withStateRoot(lastRootHash)
                .withBlockHash(blockHash)
                .withShouldWorldStateUpdateHead(true))
```
Hence you won't need these custom methods and will have a clean constructor.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java
lu-pinto: I don't agree with Gary - they are not optional, e.g., if `shouldWorldStateUpdateHead == true` -> `blockhash != null` also if `shouldWorldStateUpdateHead == false` -> `blockHeader != null` at all times.

If any of these are null and someone tries to use them, it should be a developer error, hence getting an NPE is fine. It's not like we should handle the null.
About `stateRoot` there doesn't seem to be any usage so can speak for it.

lu-pinto (commented):
Some suggestions for improvement

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/worldstate/WorldStateArchive.java
matkt: There is one called Optional<WorldState> get(final Hash rootHash, final Hash blockHash), but it is used very rarely (only by Forest) and in test classes or nearly deprecated subcommands. Therefore, I am considering removing it in the future as well, but I would prefer to do it in a separate PR.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java
matkt: make sense I can try to do it and see what is the result

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/cache/NoOpBonsaiCachedWorldStorageManager.java
```diff
@@ -18,7 +18,7 @@
 import org.hyperledger.besu.ethereum.core.BlockHeader;
 import org.hyperledger.besu.ethereum.trie.diffbased.bonsai.storage.BonsaiWorldStateKeyValueStorage;
 import org.hyperledger.besu.ethereum.trie.diffbased.common.worldview.DiffBasedWorldState;
-import org.hyperledger.besu.ethereum.trie.diffbased.common.worldview.DiffBasedWorldStateConfig;
+import org.hyperledger.besu.ethereum.trie.diffbased.common.worldview.WorldStateSharedConfig;
```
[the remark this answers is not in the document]
matkt: I checked and WorldStateConfiguration is not used in main . I removed it.

WorldStateSharedConfig is the configuration shared across all world states. When we launch Besu, we provide a base configuration that specifies whether the tree is disabled or if it is stateful. All world states will share this configuration.

WorldStateStorageConfig is the configuration specific to the world state at the level of key-value storage. It determines, for instance, whether the storage in this world state will be frozen.

I’m not sure how to make this explanation clearer, but I’m open to any ideas you might have. I understand that this can be challenging to understand.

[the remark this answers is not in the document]
matkt: I think I can try to remove WorldStateStorageConfig if it's simplify the code

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/cache/DiffBasedCachedWorldStorageManager.java
```diff
@@ -92,7 +94,7 @@ public synchronized void addCachedLayer(
         Optional.ofNullable(this.cachedWorldStatesByHash.get(blockHeader.getBlockHash()));
     if (cachedDiffBasedWorldView.isPresent()) {
       // only replace if it is a layered storage
-      if (forWorldState.isPersisted()
+      if (forWorldState.isHeadModifyingWorldState()
```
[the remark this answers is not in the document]
matkt: It's not **isMutable**. Technically, all the world states returned by `getWorldState` will be mutable at their own level. The difference lies in whether the mutation can affect the **head** of the node (i.e., the node's primary world state), or if it's just a copy that only impacts itself and can be deleted when it's no longer needed.

For instance, in cases like transaction simulation or block building, the mutations would only affect the copy and not the main world state.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -305,7 +302,7 @@ private void updateAccountStorageState(
                         writeStorageTrieNode(
                             bonsaiUpdater, updatedAddressHash, location, key, value)));
         // only use storage root of the trie when trie is enabled
-        if (!worldStateConfig.isTrieDisabled()) {
+        if (!worldStateSpec.isTrieDisabled()) {
```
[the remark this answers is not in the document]
matkt: I can use worldStateConfig again

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/cache/NoOpBonsaiCachedWorldStorageManager.java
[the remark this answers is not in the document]
matkt: removed and added more comment to explain the goal of this flag

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/cache/DiffBasedCachedWorldStorageManager.java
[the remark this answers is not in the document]
matkt: added more logs to explain this method

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/worldview/BonsaiWorldState.java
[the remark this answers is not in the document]
matkt: changed and use again worldStateConfig

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/BonsaiWorldStateProvider.java
```diff
@@ -61,9 +61,9 @@ public BonsaiWorldStateProvider(
     provideCachedWorldStorageManager(
         new BonsaiCachedWorldStorageManager(
             this, worldStateKeyValueStorage, this::cloneBonsaiWorldStateConfig));
```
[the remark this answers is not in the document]
matkt: just checked, I think it was necessary in an old commit but not needed anymore, so I changed the code to remove that

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/WorldStateStorageConfig.java
matkt: removed the class

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java
matkt: I tested it, and I don't find it clearer. I understand that it's not future-proof, but I think that in the majority of cases, we won't need to pass more fields, except in a few calls (e.g., for stateless scenarios). In those cases, we could call the builder instead of the method, but in many cases, this method will be sufficient (e.g., for private parts). So, overcomplicating things for a somewhat custom call makes me doubt. I would suggest leaving it as is for now and revisiting in the future to see if your approach makes more sense.

example of 

```java
     privacyParameters
                  .getPrivateWorldStateArchive()
                  .getWorldState(
                      () ->
                          newBuilder()
                              .withStateRoot(
                                  privacyController
                                      .getStateRootByBlockNumber(
                                          privacyGroupId, enclaveKey, blockNumber)
                                      .get())
                              .withBlockHash(parentHash)
                              .withShouldUpdateHead()
                              .build())
                  .get();
  ```
  
  instead of 
  
  ```java
  .getWorldState(
                      withStateRootAndBlockHashAndUpdateNodeHead(
                          privacyController
                              .getStateRootByBlockNumber(privacyGroupId, enclaveKey, blockNumber)
                              .get(),
                          parentHash))
  ```
  
  what do you think ? @lu-pinto @garyschulte

Comment (lu-pinto) [reply ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java:117]: I just suggested because I imagine this might grow and then your 50 character method name will turn into 80, 100 as you add fields :D
But I'm also fine with keeping it as is for now and revisit later if needed.

File: ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/PrivacyPluginPrecompiledContract.java (hunk 9 above)
Comment (matkt) [reply ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/precompiles/privacy/PrivacyPluginPrecompiledContract.java:84]: updated

File: ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/cache/NoOpBonsaiCachedWorldStorageManager.java (hunk 11 above)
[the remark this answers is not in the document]
Comment (lu-pinto) [reply ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/cache/NoOpBonsaiCachedWorldStorageManager.java:21]: Thanks removing the other configs really simplifies the code. I'm still not sure about the `WorldStateSharedConfig` naming. This to me jumps out as a shared config at runtime, among different running entities (EL / CL) rather than a share of class definitions, storage types (Forest/Bonsai/Verkle). It implies that there is an exchange of configuration at runtime and each entity has a part that is common.
What about `WorldStateCommonConfig`? or just `WorldStateConfig`? In fact you would be extending `WorldStateConfig` and  having extra configs that are specific to each storage type, no?

File: ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java [hunk 15]
```diff
@@ -36,22 +37,6 @@ private WorldStateQueryParams(final Builder builder) {
...
-      if (!shouldWorldStateUpdateHead) {
-        throw new IllegalArgumentException(
-            "Either blockHash, stateRoot, or blockHeader must be provided, or shouldWorldStateUpdateHead must be true to return the head.");
-      }
-    } else if (blockHash == null && stateRoot == null) {
-      // If shouldWorldStateUpdateHead is false, either stateRoot or blockHash must be provided to
-      // retrieve the world state from the cache.
-      throw new IllegalArgumentException(
-          "Either blockHash or stateRoot must be provided to find the worldstate in the cache");
-    }
```
lu-pinto: Why did you remove these validation methods? I think that was good to validate before crashing on the spot where you're using them

lu-pinto (commented):
Looks generally good, just have the following queries open right now:
https://github.com/hyperledger/besu/pull/8113#discussion_r1920334022
https://github.com/hyperledger/besu/pull/8113#discussion_r1920323907

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/provider/WorldStateQueryParams.java
matkt: the condition was not good and I prefered to removed it because the condition depends a lot on the storage mode. I just added another  light check in the build method

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/bonsai/cache/NoOpBonsaiCachedWorldStorageManager.java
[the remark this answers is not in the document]
matkt: updated with WorldStateConfig, like that it will be easy to understand

lu-pinto (approved):
LGTM

matkt:
> will just run a bonsai node and forest node on top of this PR before merging

forest and bonsai are running correcly with this PR. So I will merge it

PR #8127, opened 2025-01-16
Plugin API - add methods returning actual RPC host/port values

macfarla:
Add methods `getConfiguredRpcHttpHost` and `getConfiguredRpcHttpPort` which return the actual values, which will always be populated since these options have defaults. 
And deprecate from `BesuConfiguration` methods `getRpcHttpHost` and `getRpcHttpPort` (which return Optionals) 

Not having to deal unnecessarily with Optionals makes life slightly easier in the plugins.


besu/src/main/java/org/hyperledger/besu/services/BesuConfigurationImpl.java
```diff
@@ -32,8 +32,8 @@ public class BesuConfigurationImpl implements BesuConfiguration {
 
   // defaults
   private MiningConfiguration miningConfiguration;
-  private Optional<String> rpcHttpHost = Optional.of("http://localhost");
-  private Optional<Integer> rpcHttpPort = Optional.of(8545);
+  private String rpcHttpHost = "http://localhost";
+  private Integer rpcHttpPort = 8545;
```
lu-pinto, line 36: Do we need these values set at all? At least use `JsonRpcConfiguration.DEFAULT_JSON_RPC_HOST` and `JsonRpcConfiguration.DEFAULT_JSON_RPC_PORT` instead of hardcoded

lu-pinto (approved):
LGTM functionally but would be good to look at the hardcoded vars

besu/src/main/java/org/hyperledger/besu/services/BesuConfigurationImpl.java
macfarla: good point. fixed

PR #8136, opened 2025-01-17
Check existence of hardfork milestone to decide if engine APIs should be added or not

daniellehrner:
## Fixed Issue(s)
fixes #8135 

### Thanks for sending a pull request! Have you done the following?

- [ ] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [ ] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [ ] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [ ] unit tests: `./gradlew build`
- [ ] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

Issue #8135, opened 2025-01-17
Activate engine APIs even when genesis timestamp of hard fork is set to 0


daniellehrner:
The following or similar genesis config are used on testnets or rollups:

```
{
  "config": {
    "chainId": 2151908,
    "homesteadBlock": 0,
    "eip150Block": 0,
    "eip155Block": 0,
    "eip158Block": 0,
    "byzantiumBlock": 0,
    "constantinopleBlock": 0,
    "petersburgBlock": 0,
    "istanbulBlock": 0,
    "muirGlacierBlock": 0,
    "berlinBlock": 0,
    "londonBlock": 0,
    "arrowGlacierBlock": 0,
    "grayGlacierBlock": 0,
    "mergeNetsplitBlock": 0,
    "shanghaiTime": 0,
    "cancunTime": 0,
    "bedrockBlock": 0,
    "regolithTime": 0,
    "canyonTime": 0,
    "ecotoneTime": 0,
    "fjordTime": 0,
    "graniteTime": 0,
    ...
}
```

Besu currently flattens those milestones, which leads to some engine API endpoints not being activated. See details in the following issue:

https://github.com/Consensys/op-besu/issues/4

We need to make sure that those endpoints work as expected with those configurations

PR #8138, opened 2025-01-18
Log WARN for failed RPC response

lu-pinto:
## PR description
While debugging an acceptance test on the verkle branch I noticed that the cause of failure was an RPC method that was not added to the collection of supported RPC methods [here](https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/ExecutionEngineJsonRpcMethods.java#L186-L205). 

However, without enabling TRACE logging it was impossible to see what was wrong as looking for WARN or ERROR was the first thing that came to mind. I think errors like this should not be hidden within TRACE and logging at least a warning when an RPC call fails is a must have for troubleshooting.

I don't think I'm missing any case where an `RpcResponseType.ERROR` might be legit, if so I can try limiting it to `RpcErrorType.METHOD_NOT_ENABLED`.
#### Before
```
2025-01-18 00:57:11.700+00:00 | pool-8-thread-1 | INFO  | SubProcessLog | 2025-01-18 00:57:11.700+00:00 | vert.x-worker-thread-6 | TRACE | AbstractJsonRpcExecutor | {"jsonrpc":"2.0","id":67,"error":{"code":-32604,"message":"Method not enabled"}}
```
#### After
```
2025-01-18 00:30:31.885+00:00 | pool-8-thread-1 | INFO  | SubProcessLog | 2025-01-18 00:30:31.885+00:00 | vert.x-worker-thread-6 | WARN  | AbstractJsonRpcExecutor | RPC error: {"jsonrpc":"2.0","id":67,"error":{"code":-32604,"message":"Method not enabled"}}
```

### DIFF
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcObjectExecutor.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcObjectExecutor.java
@@ -34,8 +34,11 @@ import io.opentelemetry.api.trace.Tracer;
 import io.vertx.core.http.HttpServerResponse;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.RoutingContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class JsonRpcObjectExecutor extends AbstractJsonRpcExecutor {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractJsonRpcExecutor.class);
   private final ObjectWriter jsonObjectWriter = createObjectWriter();
 
   public JsonRpcObjectExecutor(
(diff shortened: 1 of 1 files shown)


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcObjectExecutor.java
```diff
@@ -76,7 +79,11 @@ private void handleJsonObjectResponse(
       try (final JsonResponseStreamer streamer =
           new JsonResponseStreamer(response, ctx.request().remoteAddress())) {
         // underlying output stream lifecycle is managed by the json object writer
-        lazyTraceLogger(() -> getJsonObjectMapper().writeValueAsString(jsonRpcResponse));
+        if (jsonRpcResponse.getType() != RpcResponseType.ERROR) {
+          lazyTraceLogger(() -> getJsonObjectMapper().writeValueAsString(jsonRpcResponse));
+        } else {
+          LOG.warn("RPC error: {}", getJsonObjectMapper().writeValueAsString(jsonRpcResponse));
+        }
```
ahamlat, line 86: This could be pretty heavy on RPC nodes, I’m thinking of “Block not found” errors for example.

siladu: Agree, this could potentially blow up the logs for a lot of users, even if they are genuine issues.
Maybe we need some sort of log throttling mechanism to go with it?

Would favour the more specific METHOD_NOT_FOUND but unless we also have throttling, I think we'd still need some confidence about the impact of this on a variety of usecases/networks.

siladu (commented):
I agree with the change in principle but it's a potentially noisy, maybe even dangerous change (e.g. disk issues due to over-logging) for existing poorly configured systems.

Btw, I think this may fix the root cause of the RPC method not being enabled
https://github.com/hyperledger/besu/issues/8135

lu-pinto:
> Btw, I think this may fix the root cause of the RPC method not being enabled #8136

Hmm, but isn't the problem altogether about missing a definition that should be there? I don't understand how this could solve as it just moves the potential problem to a missing definition in the milestones now.

PR #8139, opened 2025-01-20
Simulation: Add nonce to call parameter

Gabriel-Trintinalia:
### PR Description: 

This pull request introduces the nonce parameter to the transaction call parameters. If a nonce is specified, the simulation will use the value provided. If not, it will default to the sender's nonce.

```java
// TransactionSimulator 

final long nonce =
    callParams
        .getNonce()
        .orElse(
            Optional.ofNullable(updater.get(senderAddress)).map(Account::getNonce).orElse(0L));
```

Additionally, this PR addresses and fixes the specification tests by either:
- Passing the correct nonce, or
- Removing the nonce from the request

This update ensures compatibility with Geth's transaction call object specifications: [Geth Documentation](https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#transaction-call-object).


usmansaleem (approved):
LGTM - most changes are in the .json files.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -399,8 +399,11 @@ public Optional<TransactionSimulatorResult> processWithWorldUpdater(
       }
     }
 
-    final Account sender = updater.get(senderAddress);
-    final long nonce = sender != null ? sender.getNonce() : 0L;
+    final long nonce =
+        callParams
+            .getNonce()
+            .orElse(
```
fab-10, line 405: For efficiency, should not be `orElseGet` to only compute it is the nonce is not present in the call params?

Gabriel-Trintinalia: > For efficiency, should not be `orElseGet` to only compute it is the nonce is not present in the call params?

@fab-10 https://github.com/hyperledger/besu/pull/8145

PR #8141, opened 2025-01-20
Improve error message for INVALID OPERATION opcode

matkt:
## PR description

This PR updates the behavior of the invalid operation opcode to return an error that includes the problematic opcode in the message, replacing the previous generic message. 

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/operation/InvalidOperation.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/operation/InvalidOperation.java
@@ -53,6 +53,12 @@ public class InvalidOperation extends AbstractOperation {
     return invalidOperationResult(getOpcode());
   }
 
+  /**
+   * Creates an {@link OperationResult} for an invalid opcode.
+   *
+   * @param opcode the invalid opcode encountered
+   * @return an {@link OperationResult} with zero gas cost and a description of the invalid opcode
+   */
   public static OperationResult invalidOperationResult(final int opcode) {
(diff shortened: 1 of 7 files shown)


matkt:
this is the evm specific PR @shemnon  for #8129

evm/src/main/java/org/hyperledger/besu/evm/operation/InvalidOperation.java
```diff
@@ -49,11 +46,37 @@ public InvalidOperation(final GasCalculator gasCalculator) {
...
+          @Override
+          public String getDescription() {
+            return "invalid opcode: 0x%02x".formatted(opcode);
+          }
+
+          @Override
+          public String toString() {
+            return name();
+          }
+        });
```
lu-pinto, line 80: Instead of having the factory method here, move this  `ExceptionalHaltReason` definition to its class, e.g.:
```
  static ExceptionalHaltReason newInvalidOperation(final long opcode) {
    return new ExceptionalHaltReason() {
      @Override
      public String name() {
        return INVALID_OPERATION.name();
      }

      @Override
      public String getDescription() {
        return "invalid opcode: 0x%02x".formatted(opcode);
      }

      @Override
      public String toString() {
        return name();
      }
    };
  }
```
And then call it like so here:
```
new OperationResult(0L, ExceptionalHaltReason.newInvalidOperation(getOpcode()));
```

matkt: not sure it's better but I don't have a strong opinion about that so I will do the change

lu-pinto (approved):

PR #8142, opened 2025-01-21
Qbft block type

jframe:
## PR description
Adds a specific QBFT block type, decoupling this type from the Ethereum block so that QBFT can be used for other blocks e.g. Could be used for a beacon block on a L2.

This PR still uses the Ethereum block header a follow on PR will change this to keep this PR smaller.

**Changes**
* Introduced a qbft block type replacing the Ethereum block type. Note the Ethereum block header is still being used for now.
* Created services/type interfaces in qbft core for operations that need to be performed on the block by qbft such as validation, encoding, importing, creating etc.
* Added adaptors to the main qbft module to allow the use of Besu classes and services as qbft core types for use in the qbft core module


consensus/qbft-core/src/main/java/org/hyperledger/besu/consensus/qbft/core/statemachine/QbftBlockHeightManager.java
```diff
@@ -155,11 +155,11 @@ private void buildBlockAndMaybePropose(
     }
 
     final long headerTimeStampSeconds = Math.round(clock.millis() / 1000D);
-    final Block block = qbftRound.createBlock(headerTimeStampSeconds);
-    final boolean blockHasTransactions = !block.getBody().getTransactions().isEmpty();
-    if (blockHasTransactions) {
+    final QbftBlock block = qbftRound.createBlock(headerTimeStampSeconds);
+    final boolean blockIsEmpty = block.isEmpty();
+    if (!blockIsEmpty) {
```
usmansaleem, line 160: nit: `if (!block.isEmpty())`

usmansaleem (approved):
LGTM - very minor nitpick which can be ignored.

consensus/qbft-core/src/main/java/org/hyperledger/besu/consensus/qbft/core/statemachine/QbftBlockHeightManager.java
jframe: done, also change the code back to use the existing `block.getBody().getTransactions().isEmpty()` check we doing before to determine if the block is empty. will spin up a quick testnet and verify that the less empty block flag works, too after the change.

PR #8143, opened 2025-01-21
Fix rlp validate to handle lists ending at the start of another list

Matilda-Clerke:
## PR description
RLP.validate attempts to traverse the supplied RLP, but when the end of a list is at the start of a new list it fails to correctly leave the previous list. In particular, this was noticed to occur when validating RLP representing an access list transaction in which both the access list entry and its list of storage keys ended on the start of the next access list entry.

### DIFF
--- a/ethereum/rlp/src/test/java/org/hyperledger/besu/ethereum/rlp/RLPTest.java
+++ b/ethereum/rlp/src/test/java/org/hyperledger/besu/ethereum/rlp/RLPTest.java
@@ -168,6 +168,14 @@ public class RLPTest {
         .hasMessageContaining("RLP item exceeds max supported size of 2147483647: 2147483648");
   }
 
+  @Test
+  public void testValidateWithEmptyListIncluded() {
+    Bytes validRlp =
+        Bytes.fromHexString(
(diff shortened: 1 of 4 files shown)


pinges (approved):
LGTM

ethereum/rlp/src/test/java/org/hyperledger/besu/ethereum/rlp/RLPTest.java
```diff
@@ -168,6 +168,14 @@ public void calculateSize_overflowMaxRLPStringLength() {
         .hasMessageContaining("RLP item exceeds max supported size of 2147483647: 2147483648");
   }
 
+  @Test
```
pinges, line 171: Is this testing the case where you have multiple lists end? If not, could you please add another test that does test that!

Matilda-Clerke: Yes, it is. This is testing the exact transaction that I have been investigating

ethereum/rlp/src/test/java/org/hyperledger/besu/ethereum/rlp/RLPTest.java
```diff
@@ -168,6 +168,14 @@ public void calculateSize_overflowMaxRLPStringLength() {
...
+  public void testValidateWithEmptyListIncluded() {
```
siladu, line 172: nit: Sounds like the issue is more specific than just an empty list? Would be nice to make the test name and RLP example precisely highlight the issue.

siladu (approved):

ethereum/rlp/src/test/java/org/hyperledger/besu/ethereum/rlp/RLPTest.java
Matilda-Clerke: Yep, I wrote the test back when I still thought the empty list was the problem

ahamlat:
This looks a pretty critical part of the code, that is running fine for years. Would you mind testing a sync from scratch to see if your change doesn't have any regression ?

Matilda-Clerke:
Fair point @ahamlat I'll be running full syncs for Matilda-Clerke:7582-avoid-unnecessary-rlp-encoding-during-sync with this change included

Issue #5788, opened 2023-08-23
Allow state override on eth_call and eth_estimateGas 


antonydenyer:
### Description
Currently, you can not perform an `eth_estimateGas` or an `eth_call` on something that isn't in the 'correct' state as the evm will, correctly, return an error. For example, on an erc20 you may wish to send an approval followed by a transfer call. 

You can estimate gas for the approve but not the transfer.
You can not check the result of a transfer without an approval first. 

As a network participant, I would like access to be able to drop in and replace Besu in certain networks/tooling/RPC stacks.

### Acceptance Criteria
Whilst it's not specified in the ethereum rpc spec the implementation should be based on the Geth rpc docs to allow a drop in replacement.

https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-call

adarcanum:
gm, anyone working on this? I recently had to implement this on client side for jvm wouldn't mind submitting a PR for besu.

macfarla:
Block overrides https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#block-overrides

State overrides https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#state-override-set

macfarla:
Created new issue for block overrides https://github.com/hyperledger/besu/issues/7822

macfarla:
#7801 implements this for eth_call, PR incoming to add the param parsing for eth_estimateGas

PR #8166, opened 2025-01-24
Simulation:  Add `state` to StateOverride (eth_call, eth_simulateV1).

Gabriel-Trintinalia:
## PR description

The state field functions similarly to stateDiff, but with the added functionality of clearing the storage before applying the diff map.

refs #5788


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -459,13 +459,21 @@ public Optional<TransactionSimulatorResult> processWithWorldUpdater(
...
+    override
+        .getState()
+        .ifPresent(
+            d -> {
+              account.clearStorage();
+              d.forEach(
+                  (key, value) ->
+                      account.setStorageValue(
+                          UInt256.fromHexString(key), UInt256.fromHexString(value)));
+            });
```
macfarla, line 476: maybe extract lines 472-475 to a method and call it for state and stateDiff - something like setStorageValues or applyStorageOverrides ...

macfarla: or maybe 470-475 and pass a boolean for whether to clear first

macfarla (approved):
non-blocking suggestion

Issue #8070, opened 2024-12-27
Pectra Execution Spec Tests


siladu:
Tests: https://github.com/ethereum/execution-spec-tests/releases/tag/pectra-devnet-5%40v1.1.0

siladu:
https://github.com/hyperledger/besu/pull/8080 adds adds the prereleased spec tests [`e2105b6` (#8080)](https://github.com/hyperledger/besu/pull/8080/commits/e2105b69ac164a016bf6464ffd9694b24274ce92)
to the draft devnet-5 branch

siladu:
Latest results for Prague fork using https://github.com/ethereum/execution-spec-tests/tree/pectra-devnet-5%40v1.3.0

### Running with EEST

`uv run fill --fork=Prague --evm-bin=../besu/build/install/besu/bin/evmtool --evm-dump-dir ./prague`

```
10 failed, 11062 passed, 18 skipped, 106 deselected, 16158 warnings, 65 errors in 4532.28s (1:15:32)
```

```
FAILED tests/cancun/eip4844_blobs/test_blob_txs.py::test_invalid_tx_max_fee_per_blob_gas_state[fork_Prague-insufficient_max_fee_per_blob_gas-state_test] - ethereum_test_specs.helpers.TransactionUnexpectedSuccessError:
FAILED tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py::test_block_hashes_history[fork_Prague-blockchain_test_engine-full_history_plus_one_check_blockhash_first] - requests.exceptions.ReadTimeout: HTTPConnectionPool(host='localhost', port=62579): Read timed out. (read timeout=5)
FAILED tests/prague/eip2935_historical_block_hashes_from_state/test_contract_deployment.py::test_system_contract_deployment[fork_CancunToPragueAtTime15k-blockchain_test-deploy_after_fork] - ethereum_test_types.types.Alloc.MissingAccountError: Account missing from allocation 0x0f792be4b0c0cb4dae440ef133e90c0ecd48cccc
FAILED tests/prague/eip2935_historical_block_hashes_from_state/test_contract_deployment.py::test_system_contract_deployment[fork_CancunToPragueAtTime15k-blockchain_test_engine-deploy_after_fork] - ethereum_test_types.types.Alloc.MissingAccountError: Account missing from allocation 0x0f792be4b0c0cb4dae440ef133e90c0ecd48cccc
[...]

PR #8167, opened 2025-01-24
Prague t8n and referenceTests fixes for EEST

siladu:
- Fix referenceTests bug with fallback EMPTY Address
- T8n support for excessBlobGas for both state tests and blockchain tests: State tests should use the given excessBlobGas rather than computing from the parent header

Part of https://github.com/hyperledger/besu/issues/8070


ethereum/referencetests/src/reference-test/java/org/hyperledger/besu/ethereum/vm/GeneralStateReferenceTestTools.java
```diff
@@ -252,7 +252,7 @@ private static void logWorldState(final ReferenceTestWorldState worldState) {
               if (!storageEntries.isEmpty()) {
                 accountJson.set("storage", storageJson);
               }
-              worldStateJson.set(account.getAddress().orElse(Address.wrap(Bytes.EMPTY)).toHexString(), accountJson);
+              worldStateJson.set(account.getAddress().orElse(Address.ZERO).toHexString(), accountJson);
```
matkt, line 255: is it better to do that 
```java
worldStateJson.set(account.getAddress().orElse(Bytes.EMPTY).toHexString()
```

siladu: Done [`e50e2fb` (#8167)](https://github.com/hyperledger/besu/pull/8167/commits/e50e2fb4b29c6343175dc0908dd6847ca2366ea9)

macfarla (approved):

PR #8174, opened 2025-01-27
Extend simulate transaction on pending block plugin API

### LINKED ISSUE #8173: Expose the peding block header to plugins
### Description
As a plugin dev, I want to have access to the pending block header so that I can better simulate transactions

fab-10:
## PR description

Extend the simulate transaction plugin API to allow for plugin to access the pending block header, and giving the possibility of allowing future nonce during the simulation, for example we accepting txs to the txpool, also the coinbase of the pending block header is now correctly populated, and removed deprecated methods.

## Fixed Issue(s)

fixes #8173 

### DIFF
--- a/besu/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
+++ b/besu/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
@@ -14,11 +14,15 @@
  */
 package org.hyperledger.besu.services;
 
+import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulator;
+import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulatorAllowExceedingBalance;
+import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulatorAllowExceedingBalanceAndFutureNonce;
+import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulatorAllowFutureNonce;
+
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.datatypes.StateOverrideMap;
 import org.hyperledger.besu.datatypes.Transaction;
 import org.hyperledger.besu.ethereum.chain.Blockchain;
-import org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams;
 import org.hyperledger.besu.ethereum.mainnet.ValidationResult;
 import org.hyperledger.besu.ethereum.processing.TransactionProcessingResult;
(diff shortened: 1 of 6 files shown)


besu/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
```diff
@@ -51,50 +56,67 @@ public void init(final Blockchain blockchain, final TransactionSimulator transac
...
-                : TransactionValidationParams.transactionSimulator(),
+                ? isAllowFutureNonce
+                    ? transactionSimulatorAllowExceedingBalanceAndFutureNonce()
+                    : transactionSimulatorAllowExceedingBalance()
+                : isAllowFutureNonce
+                    ? transactionSimulatorAllowFutureNonce()
+                    : transactionSimulator(),
             operationTracer,
-            transactionSimulator.simulatePendingBlockHeader())
+            (org.hyperledger.besu.ethereum.core.ProcessableBlockHeader) pendingBlockHeader)
```
garyschulte, line 119: If we change the import in TransactionSimulator and BlockHashProcessor to org.hyperledger.besu.plugin.data.ProcessableBlockHeader, we do not need this casst

besu/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
```diff
@@ -51,50 +56,67 @@ public void init(final Blockchain blockchain, final TransactionSimulator transac
...
+      final Hash blockHash,
       final OperationTracer operationTracer,
       final boolean isAllowExceedingBalance) {
 
     final CallParameter callParameter = CallParameter.fromTransaction(transaction);
 
-    if (maybeBlockHash.isPresent()) {
-      final Hash blockHash = maybeBlockHash.get();
+    final var maybeBlockHeader =
+        blockchain.getBlockHeader(blockHash).or(() -> blockchain.getBlockHeaderSafe(blockHash));
```
garyschulte, line 75: I think this can be just a call to getBlockHeaderSafe() since it defers to getBlockHeader if it is not in the cache

garyschulte (approved):
LGTM, non-blocking feedback

besu/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
fab-10: but `getBlockHeaderSafe` is always synchronized, so we can a bit doing first the non synchronized way

besu/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
fab-10: I tried but found much more places to change and gave up

PR #8183, opened 2025-01-28
Check if ENR is no more than 300 bytes when reading ENR in ENRResponsePacketData

Matilda-Clerke:
## PR description
Add a check that the ENR is no more than 300 bytes, per the [spec](https://eips.ethereum.org/EIPS/eip-778#rlp-encoding) before attempting to parse

### DIFF
--- a/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/ENRResponsePacketData.java
+++ b/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/ENRResponsePacketData.java
@@ -46,7 +46,10 @@ public class ENRResponsePacketData implements PacketData {
     in.enterList();
     final Bytes requestHash = in.readBytes();
     in.leaveListLenient();
-    final NodeRecord enr = NodeRecordFactory.DEFAULT.fromBytes(in.currentListAsBytes());
+
+    final Bytes rawEnrRlp = in.currentListAsBytes();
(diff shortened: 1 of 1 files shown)


ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/internal/ENRResponsePacketData.java
```diff
@@ -46,7 +46,10 @@ public static ENRResponsePacketData readFrom(final RLPInput in) {
     in.enterList();
     final Bytes requestHash = in.readBytes();
     in.leaveListLenient();
-    final NodeRecord enr = NodeRecordFactory.DEFAULT.fromBytes(in.currentListAsBytes());
+
+    final Bytes rawEnrRlp = in.currentListAsBytes();
+    checkArgument(rawEnrRlp.size() <= 300, "enr must be no more than 300 bytes");
+    final NodeRecord enr = NodeRecordFactory.DEFAULT.fromBytes(rawEnrRlp);
```
jframe, line 52: Think it would be better to make this change in the discovery library itself instead of Besu

Matilda-Clerke: Fair enough, I'll look into getting a local clone of the discovery repo

Matilda-Clerke:
Fixed in https://github.com/Consensys/discovery/pull/185

Issue #8097, opened 2025-01-09
Explicit error code in engine_newPayloadV4 for Prague


siladu:
https://github.com/ethereum/execution-apis/pull/622

Also these
https://github.com/ethereum/execution-apis/pull/623/files
https://github.com/ethereum/execution-apis/pull/624/files

siladu:
There's also this change https://github.com/ethereum/execution-apis/pull/624/files

PR #8187, opened 2025-01-29
Fix request parameter validation

pinges:
## PR description
The specification for engine_newPayloadV4 has slightly changed (https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md). This PR makes changes the validation of the executionRequests to make Besu conform to the specification.
Fixes issue #8097.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
```diff
@@ -41,20 +41,24 @@ public class MainnetRequestsValidator implements RequestsValidator {
    */
   @Override
   public boolean validate(final Optional<List<Request>> maybeRequests) {
-    if (maybeRequests.isEmpty()) {
-      LOG.warn("Must contain requests (even if empty list)");
+    if (containsRequestWithEmptyData(maybeRequests.get())) {
```
lu-pinto, line 44: should we discard everything if at least one has empty data? From [eip-7685](https://eips.ethereum.org/EIPS/eip-7685) it is said to discard the ones that are empty but still use the other ones:
> In order to compute the commitment, an intermediate hash list is first built by hashing all non-empty requests elements of the block requests list. Items with empty request_data are excluded, i.e. the intermediate list skips requests items which contain only the request_type (1 byte) and nothing else.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
```diff
@@ -41,20 +41,24 @@ public class MainnetRequestsValidator implements RequestsValidator {
    */
   @Override
   public boolean validate(final Optional<List<Request>> maybeRequests) {
```
lu-pinto, line 43: Why is this working with an `Optional<List<Request>>` ? Can't you just not use a `List`?
AFAICS returning `Collections.emptyList()` should handle what you need which is even more efficient than building an Optional.

lu-pinto:
I see filtering of requests with empty data is done in `BodyValidation::requestsHash` but why are they not filtered before in `AbstractEngineNewPayload::extractRequests` ? I don't see why we need them, even during the validation.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
pinges: The spec (https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md) says "Elements with empty request_data MUST be excluded from the list. If any element is out of order, has a length of 1-byte or shorter, or more than one element has the same type byte, client software **MUST** return `-32602: Invalid params` error.". 
I interpreted the "If any element ... has a length of 1-byte or shorter" as "if there is not data"

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
pinges: I think you are right in this case. I think it should be OK to change that interface here.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
lu-pinto: Hmm indeed it's confusing. The EIP mentions skipping hough, which is different from failing hard. The case of only a single byte is indeed the case where data is empty because the request type is 1 byte length.
Have you verified this against any test or client's code?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV4.java
```diff
@@ -78,9 +78,6 @@ protected ValidationResult<RpcErrorType> validateParameters(
       return ValidationResult.invalid(
           RpcErrorType.INVALID_PARENT_BEACON_BLOCK_ROOT_PARAMS,
           "Missing parent beacon block root field");
-    } else if (maybeRequestsParam.isEmpty()) {
```
jframe, line 81: I think we still need this check as we need to have execution requests, but they can be an empty list

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
jframe: as long as we can still tell whether the requests parameter is present at all we fail when it isn't present as at all for pectra then we can change the type

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
jframe: That section of the EIP is concerned with calculating the commitment, not the engine API. Still a good idea to check hive and other client's code though.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BodyValidation.java
```diff
@@ -97,14 +97,7 @@ public static Hash withdrawalsRoot(final List<Withdrawal> withdrawals) {
    */
   public static Hash requestsHash(final List<Request> requests) {
     List<Bytes> requestHashes = new ArrayList<>();
-    requests.forEach(
-        request -> {
-          // empty requests are excluded from the hash
-          if (!request.getData().isEmpty()) {
-            requestHashes.add(sha256(request.getEncodedRequest()));
-          }
```
jframe, line 105: I think we still need to filter out the empty requests when we are building our blocks

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/engine/EngineNewPayloadV4.java
pinges: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BodyValidation.java
pinges: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/MainnetRequestsValidator.java
pinges: I have checked and the way it is working at the moment is that there is a ProhibitedRequestValidator that checks that the Optional is empty. That validator is used for all forks, but Prague. That means we better keep it an optional.

PR #8207, opened 2025-01-30
fixes #8204 by using BadBlockManager from protocol context

### LINKED ISSUE #8204: Invalid Block Not Properly Tracked in Consensus Bugs
When a consensus bug occurs, the invalid block is not properly tracked. Although the block is added to the BadBlockManager, the instance used for at the block processing part is not the same instance used for the RPC tracing functionality.

As a result, when calling tracing APIs such as debug_standardTraceBadBlockToFile, the block is not found, preventing investigation.

To fix this issue, both components (the block processing and the RPC tracing) should share the same BadBlockManager instance

Also, it appears that adding a block to the BadBlockManager does not always happen. For instance, in the case of a state root mismatch, the execution goes through this part of the code without adding the block:
 [MainnetBlockValidator.java#L221](https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java#L221)

We should investigate why the decision was made not to add the block in this case and determine how to properly handle state root mismatches so that they are also tracked as bad blocks.

- Ensure that the BadBlockManager instance used for processing blocks is the same as the one used for RPC tracing.
- Investigate why state root mismatches are not added to the BadBlockManager and determine how to fix this behavior.

jflo:
fixes first part of #8204 by using BadBlockManager from protocol context

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - Upgrade spring-security-crypto to address CVE-2025-22228 [#8448](https://github.com/hyperledger/besu/pull/8448)
 - Fix restoring txpool from disk when blob transactions are present [#8481](https://github.com/hyperledger/besu/pull/8481)
 - Fix for bonsai db inconsistency on abnormal shutdown [#8500](https://github.com/hyperledger/besu/pull/8500)
+- Fix to add stateroot mismatches to bad block manager [#8207](https://github.com/hyperledger/besu/pull/8207)
 
 ## 25.3.0 
 
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/MainnetBlockValidator.java
@@ -15,7 +15,6 @@
 package org.hyperledger.besu.ethereum;
 
 import org.hyperledger.besu.ethereum.chain.BadBlockCause;
-import org.hyperledger.besu.ethereum.chain.BadBlockManager;
 import org.hyperledger.besu.ethereum.chain.MutableBlockchain;
 import org.hyperledger.besu.ethereum.core.Block;
 import org.hyperledger.besu.ethereum.core.BlockHeader;
@@ -56,9 +55,6 @@ public class MainnetBlockValidator implements BlockValidator {
   /** The BlockProcessor used to process blocks. */
   protected final BlockProcessor blockProcessor;
 
-  /** The BadBlockManager used to manage bad blocks. */
-  protected final BadBlockManager badBlockManager;
-
   /**
    * Constructs a new MainnetBlockValidator with the given BlockHeaderValidator, BlockBodyValidator,
    * BlockProcessor, and BadBlockManager.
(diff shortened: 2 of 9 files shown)


macfarla (dismissed):

matkt:
did you had time to test it with a stateroot mismatch issue ? I saw on my last test that with a stateroot mismatch we are passing here https://github.com/hyperledger/besu/pull/8207/files#diff-e390f1feab110aff3f324e7e36e681b676198067f3b8932ff3abe97f0793df50R215 and in this case we are not adding the bad block

garyschulte:
> did you had time to test it with a stateroot mismatch issue ? I saw on my last test that with a stateroot mismatch we are passing here https://github.com/hyperledger/besu/pull/8207/files#diff-e390f1feab110aff3f324e7e36e681b676198067f3b8932ff3abe97f0793df50R215 and in this case we are not adding the bad block

@matkt when you have a moment, pls check this commit for context: https://github.com/hyperledger/besu/pull/8207/commits/d9b05643d1da1647754ed0ae87185da84ce0d74b

matkt:
> > did you had time to test it with a stateroot mismatch issue ? I saw on my last test that with a stateroot mismatch we are passing here https://github.com/hyperledger/besu/pull/8207/files#diff-e390f1feab110aff3f324e7e36e681b676198067f3b8932ff3abe97f0793df50R215 and in this case we are not adding the bad block
> 
> @matkt when you have a moment, pls check this commit for context: [d9b0564](https://github.com/hyperledger/besu/commit/d9b05643d1da1647754ed0ae87185da84ce0d74b)

I see that you are catching a new execption for the stateroot mismatch but I don't see how it could fix the issue 

In my case the I had an invalid stateroot and we passed here https://github.com/hyperledger/besu/pull/8207/files#diff-e390f1feab110aff3f324e7e36e681b676198067f3b8932ff3abe97f0793df50L221. With your commit it seems to be the same or maybe I missing something 🤔

garyschulte:
> did you had time to test it with a stateroot mismatch issue ? I saw on my last test that with a stateroot mismatch we are passing here https://github.com/hyperledger/besu/pull/8207/files#diff-e390f1feab110aff3f324e7e36e681b676198067f3b8932ff3abe97f0793df50R215 and in this case we are not adding the bad block

can you give a test case?  previously bonsai stateroot mismatches were never being added as bad blocks, but were being treated as a database layer problem.  This PR should address that.  LMK if there is a reproducible case where a stateroot mismatch does not result in adding to the bad block manager.

matkt:
> > did you had time to test it with a stateroot mismatch issue ? I saw on my last test that with a stateroot mismatch we are passing here https://github.com/hyperledger/besu/pull/8207/files#diff-e390f1feab110aff3f324e7e36e681b676198067f3b8932ff3abe97f0793df50R215 and in this case we are not adding the bad block
> 
> can you give a test case? previously bonsai stateroot mismatches were never being added as bad blocks, but were being treated as a database layer problem. This PR should address that. LMK if there is a reproducible case where a stateroot mismatch does not result in adding to the bad block manager.

Don't have a unit test but it was pretty easy to reproduce the issue. I think if you force a stateroot mismatch by changing the gas cost of an operation and try to sync on mainnet you will be able to have it

matkt:
Checked and it's working, I saw that you are using  in case of a stateroot mismatch so it will work
```java
public BlockProcessingResult(
      final Optional<BlockProcessingOutputs> yield, final String errorMessage) {
    super(errorMessage);
    this.yield = yield;
    this.isPartial = false;
  }
  ```
  and not
  ```java
   public BlockProcessingResult(
      final Optional<BlockProcessingOutputs> yield, final Throwable cause) {
    super(cause.getLocalizedMessage(), cause);
    this.yield = yield;
    this.isPartial = false;
  }
  ```

matkt (approved):
LGTM

PR #8211, opened 2025-01-31
Prague Devnet5 - EIP2935 - Fix edge case when the contract is deployed after the fork

### LINKED ISSUE #8202: EIP-2935 Hive Failure
Failure is an edge case if the system contract doesn't exist by Pectra (which it will in reality), so this shouldn't block devnet-6/testnets

In https://eips.ethereum.org/EIPS/eip-2935

We don't do this...
> if no code exists at HISTORY_STORAGE_ADDRESS, the call must fail silently

We always create the account, https://github.com/hyperledger/besu/blob/ccd35826303fe38d4bba4c517ea4b0b830ecebdc/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/blockhash/PragueBlockHashProcessor.java#L71

Gabriel-Trintinalia:
## PR description
Fix edge case when the contract is deployed after the fork
Fixes #8202


siladu (approved):
LGTM

PR #8216, opened 2025-01-31
Add a tx selector to skip txs from the same sender after the first not selected

fab-10:
## PR description

During block creation, if a sender has more that one pending txs and one of them if not selected for any reason, then it is worthless to keep evaluation other pending txs from the same sender, since they will also be not selected due to the nonce gap. 
Before this PR, this check was only done in the layered txpool, but it is better to move it as a standard tx selector, so it applies also to the sequenced txpool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,12 @@
 ### Breaking Changes
 ### Upcoming Breaking Changes
 ### Additions and Improvements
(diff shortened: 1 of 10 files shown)


garyschulte (approved):
LGTM, but needs an updated plugin-api hash to pass CI

macfarla (approved):
changelog entry

Issue #7982, opened 2024-12-04
Dynamic Cache Sizes for Max Blobs


siladu:
https://eips.ethereum.org/EIPS/eip-7691 sets the max blobs using a blobSchedule in the genesis config. At a later date, perhaps EIP-7742 may configure this via the engine API instead.

Outside of the protocol changes, there are two places where we hardcoded max blobs that are impacted by this:
a. https://github.com/hyperledger/besu/blob/16713067490dc99ad36b21fa3d85d9dc639030ef/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/BlobCache.java#L37
b. https://github.com/hyperledger/besu/blob/16713067490dc99ad36b21fa3d85d9dc639030ef/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolConfiguration.java#L77-L78

They were updated manually for Pectra where max blob was increased to 9:
https://github.com/hyperledger/besu/pull/8222
https://github.com/hyperledger/besu/pull/8101

We need to be careful about driving these configs dynamically and it needs some performance analysis/capacity planning otherwise we could create a performance issue via the config.

from Mikhail:
> max blobs will not be higher than 16 during the next year.
> I think that the target for Danksharding was 128 blobs, and this is still the target for DAS. So, the max is likely targeted to 128

So for the txpool pool and blob cache, I think some implementation options are:
```
1. Dynamic via blobSchedule (and maybe via engine API if we play EIP-7742)
2. A staged config change, e.g. increase cache multiple to 16 now and plan to increase in the future
3. Go straight for 128 and hopefully don't have to think about it again
```

Sounds like Fabio is in favour of (3). I think I am too, but still considering (2) for blob cache depending on effort/performance impact. It would be nice not to increase Besu's memory requirement even more in the short term.

siladu:
EIP-7742 was replaced with EIP-7840 so these values can be read from the genesis config now instead...however, we could still preemptively oversize if we choose.

siladu:
@fab-10 Has tackled one of the pieces here: https://github.com/hyperledger/besu/pull/8101

Gabriel-Trintinalia:
Similar one, related to PeerDas: https://github.com/hyperledger/besu/issues/8749

PR #8222, opened 2025-02-03
Use 9 for max blobs when sizing BlobCache

siladu:
EIP-7691 increases max blobs to 9 in Prague

Increase memory impact:
```
(6 * 32 * 3) -> (9 * 32 * 3)
576 blobs -> 864 blobs
73.7MB -> 110.6MB
```

Part of https://github.com/hyperledger/besu/issues/7982


fab-10 (approved):

jflo (approved):
Should be sourced from protocol schedule since we can now configure it in the genesis file.

Doing that in another issue/pr is fine.

siladu:
> Should be sourced from protocol schedule since we can now configure it in the genesis file.

@jflo Agree, I've intentionally left that for now as I think it needs some performance analysis/planning otherwise we could create a performance issue via the config.

This is discussed on the parent issue: #7982 which I will reword and leave open now that we've done the necessaries for Pectra.

TL;DR - we could size according to the expected max (e.g. 128 blobs) and then hopefully not have to touch this again. Dynamic from config might be OK too, but we should probably cap it.

macfarla (approved):

PR #8224, opened 2025-02-03
Quorum/IBFT to  Besu/QBFT migration

pullurib:
## PR description

- Restored the removed IBFT code required for parsing and validating IBFT extra data.
- Added test cases for importing IBFT blocks into a Besu network that transitions to QBFT after the imported IBFT blocks.
- Added an end-to-end test case to simulate a 5-node Besu network and validate the block import process.


pullurib:
Discarding in lieu of https://github.com/hyperledger/besu/pull/8262

PR #8243, opened 2025-02-04
EIP2935 - Use system call instead of direct storage modification

Gabriel-Trintinalia:
## PR description
Changes `PragueBlockHashProcessor` to perform a system call to `HISTORY_STORAGE_ADDRESS` instead of changing the storage manually as defined here: https://github.com/ethereum/EIPs/pull/8816

Passing both `consume-rlp` and `consume-engine` execution-spec-tests in `pectra-devnet-6 v1.0.0 `


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/blockhash/PragueBlockHashProcessor.java
```diff
@@ -15,91 +15,47 @@
...
-        "Writing to {} {}=%{}", account.getAddress(), slot.toDecimalString(), value.toHexString());
-    account.setStorageValue(slot, value);
+  public Void process(final BlockProcessingContext context) {
+    super.process(context);
+    SystemCallProcessor processor =
+        new SystemCallProcessor(context.getProtocolSpec().getTransactionProcessor());
+
+    Bytes inputData = context.getBlockHeader().getParentHash();
+    processor.process(historyStorageAddress, context, inputData);
+    return null;
```
Gabriel-Trintinalia, line 59: This is the important change of this PR. The rest is refactoring.

matkt (approved):
LGTM

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/SystemCallProcessor.java
```diff
@@ -48,37 +49,33 @@ public SystemCallProcessor(final MainnetTransactionProcessor mainnetTransactionP
...
+   * @return The output of the system call.
    */
   public Bytes process(
-      final Address callAddress,
-      final WorldUpdater worldState,
-      final ProcessableBlockHeader blockHeader,
-      final OperationTracer operationTracer,
-      final BlockHashLookup blockHashLookup) {
+      final Address callAddress, final BlockProcessingContext context, final Bytes inputData) {
+    WorldUpdater updater = context.getWorldState().updater();
```
matkt, line 60: for more visibility I would have liked to see the commit of the updater at the same level as the creation of this one. Because we have to go down in the levels to understand that we are committing the updater in certain conditions

Gabriel-Trintinalia: Cool! Done!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/SystemCallProcessor.java
```diff
@@ -130,7 +119,7 @@ private MessageFrame createCallFrame(
         .type(MessageFrame.Type.MESSAGE_CALL)
         .address(callAddress)
         .contract(callAddress)
-        .inputData(Bytes.EMPTY)
+        .inputData(inputData)
         .sender(SYSTEM_ADDRESS)
```
lu-pinto, line 123: I would move `SYSTEM_ADDRESS` to `PragueBlockHashProcessor` and make it an instance field here set through the constructor. In the future why not reusing this `SystemCallProcessor` for other contract calls?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/BlockProcessingContext.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+package org.hyperledger.besu.ethereum.mainnet.systemcall;
+
+import org.hyperledger.besu.ethereum.core.MutableWorldState;
+import org.hyperledger.besu.ethereum.core.ProcessableBlockHeader;
+import org.hyperledger.besu.ethereum.mainnet.ProtocolSpec;
+import org.hyperledger.besu.evm.blockhash.BlockHashLookup;
+import org.hyperledger.besu.evm.tracing.OperationTracer;
+
+public class BlockProcessingContext {
```
lu-pinto, line 23: can be a `record`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/SystemCallProcessor.java
```diff
@@ -12,19 +12,19 @@
...
+import org.hyperledger.besu.ethereum.mainnet.MainnetTransactionProcessor;
 import org.hyperledger.besu.evm.account.Account;
 import org.hyperledger.besu.evm.blockhash.BlockHashLookup;
 import org.hyperledger.besu.evm.code.CodeV0;
 import org.hyperledger.besu.evm.frame.MessageFrame;
 import org.hyperledger.besu.evm.processor.AbstractMessageProcessor;
-import org.hyperledger.besu.evm.tracing.OperationTracer;
 import org.hyperledger.besu.evm.worldstate.WorldUpdater;
 
 import java.util.Deque;
```
lu-pinto, line 30: Why is this call just not called `SystemCall` instead of `SystemCallProcessor` ? What data is it processing? To me the name is a bit misleading.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/blockhash/Eip7709BlockHashProcessor.java
```diff
@@ -29,6 +29,6 @@ public class Eip7709BlockHashProcessor extends PragueBlockHashProcessor {
   @Override
   public BlockHashLookup createBlockHashLookup(
       final Blockchain blockchain, final ProcessableBlockHeader blockHeader) {
-    return new Eip7709BlockHashLookup(historyStorageAddress, historyServeWindow);
```
lu-pinto, line 32: It's better for unit tests to leave the parameter to the constructor no?

lu-pinto, line 32: Don't you think this could change in the future? Having the `BlockHashProcessor` have a handle on the window size makes it more extendable in another fork if the window changes

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/Eip7709BlockHashLookup.java
```diff
@@ -37,6 +37,7 @@
 public class Eip7709BlockHashLookup implements BlockHashLookup {
   private static final Logger LOG = LoggerFactory.getLogger(Eip7709BlockHashLookup.class);
   private static final long BLOCKHASH_SERVE_WINDOW = 256L;
+  private static final long HISTORY_SERVE_WINDOW = 8191;
```
lu-pinto, line 40: nit: yes, making these constants static here means that we cannot easily change these values in a future fork just by creating another instance with different values.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/SystemCallProcessor.java
Gabriel-Trintinalia: SYSTEM_ADDRESS is `0xfffffffffffffffffffffffffffffffffffffffe`. It is the same for all contracts and system calls and fills the frame's sender. The contract addresses, on the other hand, are indeed passed to the process as a parameter in the `callAddress`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/BlockProcessingContext.java
Gabriel-Trintinalia: `RequestProcessingContext` extends `BlockProcessingContext`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/Eip7709BlockHashLookup.java
Gabriel-Trintinalia: this constant was only moved from the processor class to the BlockHashLookup as it is not used by the Prague processor anymore

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/SystemCallProcessor.java
Gabriel-Trintinalia: `SystemCallProcessor` is similar to the `MessageCallProcessor`, it performs a message call to the contract address with some rules and default values

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/BlockProcessingContext.java
lu-pinto: Oh I didn't notice that. So why do you extend a POJO then? This looks a code smell to me.
Why doesn't `RequestProcessingContext` use composition instead of inheritance? This is a preferred best practice BTW.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/systemcall/SystemCallProcessor.java
lu-pinto: Well but you are not processing any message (which is what the `MessageProcessor` does) where an EOA sends a message to another EOA or a contract. In this case it is internal, you are creator of the call. See https://github.com/hyperledger/besu/pull/8243/files/6bc3fa87a7f3ab9f7ef69af600c67bf07fc63272#diff-1e7c7ef5c489c82232fb87d33704b19e4a7edb2c72a599a82293f22b79da32abL76-R76 a new `MessageProcessor` is created inside this function. Just the only we have to get the EVM to execute code right now, but it's an implementation detail.
Anyway I agree that it's subjective and it's a nit for sure.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/vm/Eip7709BlockHashLookup.java
lu-pinto: ok what about storing it in the `Eip7709BlockHashLookup` instead?

siladu (approved):
LGTM - just minor comments

PR #8248, opened 2025-02-04
Turned PostMergeContext to be a regular class instead of a singleton.…

Filter94:
… Addded some other improvements to the acceptance tests DSL

## Fixed Issue(s)
This fix allows `ThreadBesuNodeRunner` to spin up and tear down Besu instances multiple times during process's lifecycle.

### DIFF
--- a/acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/ThreadBesuNodeRunner.java
+++ b/acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/ThreadBesuNodeRunner.java
@@ -26,6 +26,7 @@ import org.hyperledger.besu.cli.config.EthNetworkConfig;
 import org.hyperledger.besu.cli.config.NetworkName;
 import org.hyperledger.besu.components.BesuComponent;
 import org.hyperledger.besu.config.GenesisConfig;
+import org.hyperledger.besu.consensus.merge.PostMergeContext;
 import org.hyperledger.besu.controller.BesuController;
 import org.hyperledger.besu.controller.BesuControllerBuilder;
 import org.hyperledger.besu.crypto.KeyPairUtil;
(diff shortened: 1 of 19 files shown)


acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/ThreadBesuNodeRunner.java
Filter94, line 1: I need a tip on this part. `ThreadBesuNodeRunner` can only create a Besu instance that doesn't have persistence, meaning it's losing all the state after restart. In-memory `KeyValueStorageProvider` is provided by `BesuControllerModule::provideKeyValueStorageProvider` called by dagger. I wonder if there's a non invasive way to invert this dependency and provide a different key value storage, namely I'd like to set it to a persistent version for my testing.

besu/src/main/java/org/hyperledger/besu/controller/TransitionBesuControllerBuilder.java
```diff
@@ -205,6 +206,8 @@ protected ConsensusContext createConsensusContext(
       final Blockchain blockchain,
       final WorldStateArchive worldStateArchive,
       final ProtocolSchedule protocolSchedule) {
+    preMergeBesuControllerBuilder.postMergeContext(postMergeContext);
+    mergeBesuControllerBuilder.postMergeContext(postMergeContext);
```
Filter94, line 210: Please note this change. I couldn't find a better way to get around this and I wonder if there's a better way

acceptance-tests/dsl/src/main/java/org/hyperledger/besu/tests/acceptance/dsl/node/ThreadBesuNodeRunner.java
Filter94, line 1: I need a tip on this part. ThreadBesuNodeRunner can only create a Besu instance that doesn't have persistence, meaning it's losing all the state after restart. In-memory KeyValueStorageProvider is provided by BesuControllerModule::provideKeyValueStorageProvider called by dagger. I wonder if there's a non invasive way to invert this dependency and provide a different key value storage, namely I'd like to set it to a persistent version for my testing.

macfarla: @jflo thoughts on dependency nuances here?

jframe (commented):
Overall I think the PR can be simpler by just reusing the consensusContext from the MergeBesuControllerBuilder.

I had a go at doing this by creating it up front in the MergeBesuControllerBuilder and exposing a getter for the TransitionControllerBuilder still a bit ugly but it hides the details in the controller builders.

Here's my attempt if you want to take a look https://github.com/hyperledger/besu/pull/8253

Filter94:
Consumed by [this](https://github.com/hyperledger/besu/pull/8253) PR

PR #8257, opened 2025-02-06
Use dynamic gas calculator to calculate currentBlobGasLimit

siladu:
If a future gas calculator changes the GasCalculator.BLOB_GAS_PER_BLOB value, then ensure it is used in the currentBlobGasLimit calculation

This doesn't impact Prague since the BLOB_GAS_PER_BLOB didn't change.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -481,6 +481,7 @@ public TransactionProcessingResult processTransaction(
         }
       }
 
+      // TODO SLD are the log correct following EIP-7623?
```
macfarla, line 484: do we know enough to resolve this TODO?

macfarla (approved):

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
siladu: I didn't mean to commit that here but it's another issue that needs looking at ideally. I think we have all the info, need to make a call about how this log might be used versus the EIP-7623 spec changes

PR #8262, opened 2025-02-06
Enable Quorum/IBFT1 to Besu migration

pullurib:
## PR description

- Restored and updated the removed IBFT1 code required for parsing and validating IBFT1 extra data.
- Made required changes to enable transitioning to QBFT after IBFT1 block import.
- Added test for importing IBFT1 blocks into a Besu network that transitions to QBFT after the imported IBFT1 blocks.
- Added an end-to-end test case to simulate a 5-node Besu network and validate the block import and transition process.


consensus/common/src/main/java/org/hyperledger/besu/consensus/common/EpochManager.java
```diff
@@ -35,7 +47,17 @@ public EpochManager(final long epochLengthInBlocks) {
...
+      throw new IllegalArgumentException("Block number must be 0 or greater.");
+    }
+    if (startBlock == 0) {
+      return (blockNumber % epochLengthInBlocks) == 0;
+    } else {
+      if (blockNumber < (startBlock - 1)) {
+        return false;
+      }
+      return ((blockNumber - (startBlock - 1)) % epochLengthInBlocks) == 0;
+    }
```
matthew1001, line 60: I feel like this logic could be simplified? Maybe something like:

```
if (startBlock != 0 && blockNumber < (startBlock - 1)) {
   return false;
} else {
   <common logic for isEpoch calculation>
}
```

consensus/ibftlegacy/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+
+jar {
+  archiveBaseName = 'besu-ibftlegacy'
+  manifest {
+    attributes(
+      'Specification-Title': archiveBaseName,
+      'Specification-Version': project.version,
+      'Implementation-Title': archiveBaseName,
+      'Implementation-Version': calculateVersion(),
+      'Commit-Hash': 'UNKNOWN'
```
matthew1001, line 11: Should this be `getGitCommitDetails(40).hash`?

consensus/ibftlegacy/src/main/java/org/hyperledger/besu/consensus/ibftlegacy/IbftBlockHashing.java
```diff
@@ -0,0 +1,180 @@
+/*
```
matthew1001, line 1: I guess I have a general question about what this class is needed for. Is it needed for importing existing IBFT legacy blocks? It looks like it relates to creating commit seals etc. and I wasn't sure if that was required for importing old blocks vs creating new blocks (and I assume we're not supporting the latter).

It may well be that it's required - I just wanted to raise the question to make sure we don't have any unnecessary IBFT legacy classes or functions in the code, only those required for import and transition to QBFT.

besu/src/main/java/org/hyperledger/besu/controller/ConsensusScheduleBesuControllerBuilder.java
```diff
@@ -199,7 +199,7 @@ protected ConsensusContext createConsensusContext(
   protected PluginServiceFactory createAdditionalPluginServices(
       final Blockchain blockchain, final ProtocolContext protocolContext) {
     return besuControllerBuilderSchedule
-        .get(0L)
+        .get(besuControllerBuilderSchedule.keySet().stream().skip(1).findFirst().orElseThrow())
```
matthew1001, line 202: Could do with a quick explanation of this. Presumably instead of getting the first builder schedule we're getting the 2nd one, and throwing if it doesn't exist? What if there is only 1 schedule - should we still be going `get(0L)` in that case?

pullurib: Yes, essentially going with second besu builder in the schedule which would be that of QBFT. There must be two builders in the schedule since this ConsensusScheduleBesuControllerBuilder is used only for migration.

besu/build.gradle
```diff
@@ -23,7 +23,7 @@ jar {
       'Specification-Version': project.version,
       'Implementation-Title': archiveBaseName,
       'Implementation-Version': calculateVersion(),
-      'Commit-Hash': getGitCommitDetails(40).hash
+      'Commit-Hash': 'UNKNOWN'
```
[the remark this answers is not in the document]
pullurib: reverting

besu/src/main/java/org/hyperledger/besu/controller/BesuController.java
```diff
@@ -394,8 +394,7 @@ private BesuControllerBuilder createConsensusScheduleBesuControllerBuilder(
       if (configOptions.isIbft2()) {
         originalControllerBuilder = new IbftBesuControllerBuilder();
       } else if (configOptions.isIbftLegacy()) {
-        throw new IllegalStateException(
-            "IBFT1 (legacy) is no longer supported. Consider using IBFT2 or QBFT.");
+        originalControllerBuilder = new IbftLegacyBesuControllerBuilder();
```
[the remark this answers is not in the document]
pullurib: It's included in the IbftLegacyBesuControllerBuilder constructor

consensus/ibftlegacy/src/main/java/org/hyperledger/besu/consensus/ibftlegacy/IbftBlockHashing.java
pullurib: This is required to parse extradata, verify header and recalculate hash for the block during import/sync

matthew1001:
I've approved the PR @pullurib, couple of last comments:
 - It's probably worth @jframe having a look at these changes if he has some time
 - The gradle verification changes look like we're pulling in some old versions of tuweni. Can you do a build & trivy scan to see if those versions have `HIGH` or `CRITICAL` CVEs in? I wonder if we can just bump the version to use the one currently in `main`?

macfarla:
haven't reviewed in detail but could use a changelog entry :)

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/BftMiningCoordinator.java
```diff
@@ -119,6 +152,41 @@ public void stop() {
...
+  @Override
+  public void subscribe() {
+    if (syncState == null) {
+      return;
+    }
+    syncState.subscribeSyncStatus(
+        syncStatus -> {
+          if (syncState.syncTarget().isPresent()) {
+            // We're syncing so stop doing other stuff
+            LOG.info("Stopping QBFT mining coordinator while we are syncing");
```
jframe, line 164: Should remove QBFT from these messages, as this might not always be QBFT as this class is also used in ibft2

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/bftsoak/BftMiningSoakTest.java
```diff
@@ -213,6 +213,8 @@ public void shouldBeStableDuringLongTest(
     upgradeToLondon(
         minerNode1, minerNode2, minerNode3, minerNode4, lastChainHeight.intValue() + 120);
 
+    Thread.sleep(THREE_MINUTES);
```
jframe, line 216: Is there something better we can do rather than a Thread sleep here? Can we check for a specific condition to occur instead

acceptance-tests/tests/src/test/resources/qbft/migration-ibft1/qbft-migration.json
```diff
@@ -0,0 +1,65 @@
...
+    },
+    "txnSizeLimit": 64,
+    "maxCodeSize": 0,
+    "maxCodeSizeConfig": [
+      {
+        "block": 0,
+        "size": 64
+      }
+    ],
+    "isMPS": false
```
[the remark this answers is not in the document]
pullurib: spillover from quorum genesis, deleted

consensus/ibftlegacy/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+  integrationTestImplementation 'org.mockito:mockito-junit-jupiter'
+
+  testImplementation 'org.assertj:assertj-core'
+  testImplementation 'org.junit.jupiter:junit-jupiter'
+  testImplementation 'org.awaitility:awaitility'
+  testImplementation 'org.mockito:mockito-core'
+  testImplementation 'org.mockito:mockito-junit-jupiter'
+
+
+  testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
```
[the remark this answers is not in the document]
pullurib: Not needed , removed it

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/bftsoak/BftMiningSoakTest.java
pullurib: Checking for blocks mined with a timeout of three minutes instead.

consensus/ibft/src/main/java/org/hyperledger/besu/consensus/ibft/IbftLegacyContext.java
```diff
@@ -0,0 +1,76 @@
...
+package org.hyperledger.besu.consensus.ibft;
+
+import org.hyperledger.besu.consensus.common.EpochManager;
+import org.hyperledger.besu.consensus.common.bft.BftBlockInterface;
+import org.hyperledger.besu.consensus.common.bft.BftContext;
+import org.hyperledger.besu.consensus.common.validator.ValidatorProvider;
+import org.hyperledger.besu.ethereum.ConsensusContext;
+
+/** Holds the BFT specific mutable state. */
+public class IbftLegacyContext extends BftContext {
```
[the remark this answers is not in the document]
pullurib: Thanks, removed it

jframe (approved):

PR #8268, opened 2025-02-07
Allow plugins to propose transactions during block creation

fab-10:


ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/TransactionSelectionResults.java
```diff
@@ -48,27 +48,20 @@ public class TransactionSelectionResults {
...
-      final long gasUsed,
-      final long blobGasUsed) {
+      final Transaction transaction, final TransactionReceipt receipt, final long gasUsed) {
     selectedTransactions.add(transaction);
     transactionsByType
         .computeIfAbsent(transaction.getType(), type -> new ArrayList<>())
         .add(transaction);
     receipts.add(receipt);
     cumulativeGasUsed += gasUsed;
-    cumulativeBlobGasUsed += blobGasUsed;
```
matkt, line 64: why we are removing cumulativeBlobGasUsed ?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/AbstractStatefulTransactionSelector.java
```diff
@@ -0,0 +1,44 @@
...
+    super(context);
+    this.selectorsStateManager = selectorsStateManager;
+    selectorsStateManager.createSelectorState(this, initialState, stateDuplicator);
+  }
+
+  protected S getWorkingState() {
+    return selectorsStateManager.getSelectorWorkingState(this);
+  }
+
+  protected void setWorkingState(final S newState) {
```
matkt, line 41: not completly sure about this name, for me it was the WorldState and after I saw that it's a long 🤔

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/BlockSizeTransactionSelector.java
```diff
@@ -29,37 +29,39 @@
  * evaluating transactions based on block size. It checks if a transaction is too large for the
  * block and determines the selection result accordingly.
  */
-public class BlockSizeTransactionSelector extends AbstractTransactionSelector {
+public class BlockSizeTransactionSelector extends AbstractStatefulTransactionSelector<Long> {
```
matkt, line 32: AbstractStatefulTransactionSelector. I feel that this name should be in conflict with Statefull and stateless worldstate. Just thniking I don't have a strong opinion about that

util/src/main/java/org/hyperledger/besu/util/log4j/plugin/StackTraceMatchFilter.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.logging.log4j.core.filter.AbstractFilter;
+import org.apache.logging.log4j.message.Message;
+
+/** Matches a text in the stack trace */
+@Plugin(
+    name = "StackTraceMatchFilter",
+    category = "Core",
+    elementType = "filter",
+    printObject = true)
+public class StackTraceMatchFilter extends AbstractFilter {
```
matkt, line 35: is this class needed for this PR ?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/TransactionSelectionResults.java
fab-10: Before this class was used to both (1)track selection results and (2)managed some state for the selectors, now the 2 tasks have been separated and in this class are only kept the selection results(1) and since the `cumulativeBlobGasUsed` is not needed as result, but only useful for the stateful selector it has been moved into the `BlobSizeTransactionSelector`

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/txselection/AbstractStatefulPluginTransactionSelector.java
```diff
@@ -0,0 +1,71 @@
...
+      final SelectorsStateManager selectorsStateManager,
+      final S initialState,
+      final SelectorsStateManager.StateDuplicator<S> stateDuplicator) {
+    this.selectorsStateManager = selectorsStateManager;
+    selectorsStateManager.createSelectorState(this, initialState, stateDuplicator);
+  }
+
+  /**
+   * Get the working state for this selector. A working state contains changes that have not yet
+   * commited
```
ahamlat, line 44: * typo fix : committed

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/AbstractStatefulTransactionSelector.java
fab-10: This is about the selector specific state, for example the `BlobSizeTransactionSelector` that needs to keep the cumulative count of how many blobs have been already added by previous txs

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/BlockSizeTransactionSelector.java
fab-10: I agree that _state_ can refer to different things here, so one should read it along with _transaction selector_ to fully qualify, not sure what could be another naming for this

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/AbstractStatefulTransactionSelector.java
matkt: I don't know if it's better but maybe WorkingContext? SelectionContext?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -260,11 +285,26 @@ private void cancelEvaluatingTxWithGraceTime(final FutureTask<Void> txSelectionT
    *     evaluations.
    */
   public TransactionSelectionResults evaluateTransactions(final List<Transaction> transactions) {
-    transactions.forEach(
-        transaction -> evaluateTransaction(new PendingTransaction.Local.Priority(transaction)));
+    selectorsStateManager.blockSelectionStarted();
+
+    transactions.stream()
```
ahamlat, line 290: Streams are not good for performance especially when using map, so I prefer the the previous version.

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -275,10 +315,14 @@ public TransactionSelectionResults evaluateTransactions(final List<Transaction>
    * @return The result of the transaction evaluation process.
    * @throws CancellationException if the transaction selection process is cancelled.
    */
-  private TransactionSelectionResult evaluateTransaction(
-      final PendingTransaction pendingTransaction) {
+  @Override
+  public TransactionSelectionResult evaluatePendingTransaction(
+      final org.hyperledger.besu.datatypes.PendingTransaction pendingTransaction) {
```
ahamlat, line 320: Why do we need to import PendingTransaction with the whole package here ? is there another PendingTransaction ?

util/src/main/java/org/hyperledger/besu/util/log4j/plugin/StackTraceMatchFilter.java
matkt: No if It is necessary for your tests I think it's OK

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/AbstractStatefulTransactionSelector.java
fab-10: Not sure, we already have `BlockSelectionContext` and `TransactionEvaluationContext` here so it can also create confusion.
Do you think using _state_ could be confusing just at a first look, and after reviewing the code for longer the distinction is clearer?

I can also rename this method to `setSelectorWorkingState`, I thought it was redundant since it is in a selector class already, but if it helps readability then no problem

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -288,27 +332,56 @@ private TransactionSelectionResult evaluateTransaction(
       return handleTransactionNotSelected(evaluationContext, selectionResult);
     }
 
-    final WorldUpdater txWorldStateUpdater = blockWorldStateUpdater.updater();
     final TransactionProcessingResult processingResult =
-        processTransaction(evaluationContext.getTransaction(), txWorldStateUpdater);
+        processTransaction(evaluationContext.getTransaction());
+
+    txWorldStateUpdater.markTransactionBoundary();
```
ahamlat, line 338: Could you elaborate on why this is needed ?

ahamlat (commented):
I haven't finished the review, but this is my first round of comments

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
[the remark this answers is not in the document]
matkt: https://github.com/hyperledger/besu/blob/main/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/diffbased/common/worldview/accumulator/DiffBasedWorldStateUpdateAccumulator.java#L609

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
fab-10: yes, for example when we create the actual pending txs in the `evaluateTransactions` method, while this method is from the interface and uses the `PendingTransaction` interface

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
fab-10: reverted

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
fab-10: this is needed in case more that one tx is processed before the commit, for example a plugin can only decide to commit only if txA and txB are both selected, so in this case you use the same world state updater first to execute txA and then txB, but without marking the tx boundary, it is possible that when executing txB there are some addresses already warm-up by txA, thus breaking the gas calculation.
This has no effect without plugin, since here we are committing after every selected, but provides flexibility for plugin to implement selection logic.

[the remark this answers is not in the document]
fab-10: it is called on `StackedUpdater` so if I understood correctly it should be fine

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/AbstractStatefulTransactionSelector.java
matkt: I think it's ok, maybe just add a comment to explain what is this state

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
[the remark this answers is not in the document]
matkt: ok just wanted to be sure we are not using it for Bonsai accumulator

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/selectors/AbstractStatefulTransactionSelector.java
fab-10: done

ahamlat (approved):
LGTM, it would be great if we can have more coverage on `BlockSizeTransactionSelector` class, as this PR changes the way cumulative gas is updated.

plugin-api/src/main/java/org/hyperledger/besu/plugin/services/txselection/SelectorsStateManager.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    /**
+     * Utility to duplicate a long
+     *
+     * @param l a long
+     * @return a copy of the long
+     */
+    static long duplicateLong(final long l) {
+      return l;
```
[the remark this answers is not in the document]
fab-10: This is like the `UnaryOperator::identity` method, since basically there is no need to duplicate a long, it is used by selectors that have a `Long` as state like `BlockSelectionContext`

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -288,27 +331,56 @@ private TransactionSelectionResult evaluateTransaction(
...
+    // only add this tx to the selected set if it is not too late,
+    // this need to be done synchronously to avoid that a concurrent timeout
+    // could start packing a block while we are updating the state here
+    synchronized (isTimeout) {
+      if (!isTimeout.get()) {
+        selectorsStateManager.commit();
+        txWorldStateUpdater.commit();
+        blockWorldStateUpdater.commit();
+        for (final var pendingAction : selectedTxPendingActions) {
+          pendingAction.run();
```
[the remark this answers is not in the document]
fab-10: yes this must be sequential and all pending actions must be completed before leaving the synchronized block

matkt (approved):

PR #8290, opened 2025-02-13
RPC - eth_getLogs: Return block not found instead of empty logs

### LINKED ISSUE #8007: eth_getLogs must return "block unknown" when blockHash is not found

Gabriel-Trintinalia:
## PR description
RPC - eth_getLogs: Return block not found instead of empty logs

## Fixed Issue(s)
Fix #8007


macfarla (approved):
changelog entry

PR #8311, opened 2025-02-17
Remove QbftBlockHashing as this is provided by the block hash

jframe:
## PR description
Removes the QbftBlockHashing interface as the same functionality is provided by creating a commitBlock and using its hash. This makes reusing the Qbft library easier too as there no assumptions that the commit hash can be derived from the blockheader and extradata.


macfarla (approved):
does it need a changelog entry?

jframe:
> does it need a changelog entry?

Don't think so, it's internal refactoring. And there is nothing documented about the QBFT core module at the moment. So don't think we need to do anything.

Issue #8313, opened 2025-02-17
Refactor EthStatsService creation of WebSocket


macfarla:
Vertx has deprecated HttpClient.webSocket() methods
it's now a separate class - https://vertx.io/docs/guides/vertx-5-migration-guide/

Refactor EthStatsService and EthStatsServiceTest to use the new class
```
> Task :ethereum:ethstats:compileJava
Note: /workspace/b2/ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java uses or overrides a deprecated API.
```

Acceptance criteria:
* no deprecation warnings when compiling this module

vitamin43:
Hi, I would like to work on this issue. Could you please assign it to me?

macfarla:
go for it @vitamin43 - any questions you can ask on this issue or in our discord https://discord.com/channels/905194001349627914/938504958909747250

KYash03:
Hi @macfarla - I apologize for the overlap, but I had already started working on this issue before the assignment to @vitamin43. Here's my PR: https://github.com/hyperledger/besu/pull/8327

macfarla:
Thanks @KYash03 looks promising, someone will review. 

 @vitamin43 don't want to waste your time - have a look at other [good first issues](https://github.com/hyperledger/besu/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22)

PR #8327, opened 2025-02-19
refactor: replace HttpClient with WebSocketClient in EthStatsService

KYash03:
Fixes https://github.com/hyperledger/besu/issues/8313

### DIFF
--- a/ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
+++ b/ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
@@ -72,10 +72,9 @@ import java.util.stream.LongStream;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import io.vertx.core.Vertx;
-import io.vertx.core.http.HttpClientOptions;
 import io.vertx.core.http.WebSocket;
+import io.vertx.core.http.WebSocketClient;
 import io.vertx.core.http.WebSocketConnectOptions;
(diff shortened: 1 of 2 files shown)


KYash03:
I seem to have missed that. I've removed`succeededWebSocketEvent` now.

ethereum/ethstats/src/main/java/org/hyperledger/besu/ethstats/EthStatsService.java
```diff
@@ -147,21 +145,9 @@ public EthStatsService(
...
-      final EthStatsConnectOptions ethStatsConnectOptions) {
-    final HttpClientOptions options = new HttpClientOptions();
-
-    if (ethStatsConnectOptions.getCaCert() != null) {
-      options.setPemTrustOptions(
-          new PemTrustOptions().addCertPath(ethStatsConnectOptions.getCaCert().toString()));
-    }
-    return options;
-  }
-
```
fab-10, line 164: This part seems not to have a replacement in the new code, are we loosing the feature to set this option?

KYash03: Yes. It's not needed given the transition to `WebSocketClient`.

fab-10: Please elaborate, because I do not see how it will be possible to specify a custom CA certificate with websockets. (https://besu.hyperledger.org/development/public-networks/reference/cli/options#ethstats-cacert-file)

Issue #7935, opened 2024-11-26
Create storage subcommand supporting ERA1 import/export format


garyschulte:
### Description
In support of dropping chain history, besu will need to support the ERA1 format for chain history pre-merge.  ERA1 data can be used to full sync the chain for specific cases like archive nodes.  Creating a subcommand that imports and ERA1 data will provide the functionality necessary to scan and read ERA1 files.  

For other chains which do not already have ERA1 data generated, or wish to leverage ERA1 for other reasons, we should also implement an export subcommand which implements an accumulator and file writer.  This also will serve to cross-verify the mainnet pre-merge ERA1 exports that have been generated by other execution clients.

### Acceptance Criteria
* ERA1 format support for reading
* ERA1 accumulator for writing
* Storage subcommands for importing and exporting ERA1 files from besu chain data

PR #8329, opened 2025-02-19
7935: Add E2StoreReader

Matilda-Clerke:
## PR description
Adds an Es2StoreReader for reading e2 files such as era and era1 files

## Fixed Issue(s)


util/src/main/java/org/hyperledger/besu/util/e2/E2BeaconState.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package org.hyperledger.besu.util.e2;
+
+/**
+ * Represents a beacon state in an E2 file
+ *
+ * @param beaconState The beacon state
+ * @param slot The slot number
+ */
+public record E2BeaconState(byte[] beaconState, int slot) {}
```
jframe, line 23: We don't need the beacon types as we won't be writing or reading the beacon data from anywhere

util/src/main/java/org/hyperledger/besu/util/e2/E2SignedBeaconBlock.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package org.hyperledger.besu.util.e2;
+
+/**
+ * Represents a signed beacon block in an E2 file
+ *
+ * @param signedBeaconBlock The signed beacon block
+ * @param slot The slot number
+ */
+public record E2SignedBeaconBlock(byte[] signedBeaconBlock, int slot) {}
```
jframe, line 23: Same here, we don't need this type

util/src/main/java/org/hyperledger/besu/util/e2/E2SlotIndex.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.List;
+
+/**
+ * Represents a slot index in an E2 file
+ *
+ * @param startingSlot The first slot number indexed by this slot index
+ * @param indexes The indexes of the slots indexed by this slot index
+ */
+public record E2SlotIndex(long startingSlot, List<Long> indexes) {}
```
jframe, line 25: Also don't need this

util/src/main/java/org/hyperledger/besu/util/e2/E2StoreReaderListener.java
```diff
@@ -0,0 +1,67 @@
...
+   * @param slotIndex the E2SlotIndex
+   */
+  void handleSlotIndex(E2SlotIndex slotIndex);
+
+  /**
+   * Handles the supplied E2SignedBeaconBlock
+   *
+   * @param signedBeaconBlock the E2SignedBeaconBlock
+   */
+  void handleSignedBeaconBlock(E2SignedBeaconBlock signedBeaconBlock);
```
jframe, line 38: can remove the beacon handlers, they won't be in the era1 files we use

util/src/main/java/org/hyperledger/besu/util/e2/E2Type.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.HexFormat;
+
+/** An enumeration of known sections of e2 store files */
+public enum E2Type {
+  /** An empty section */
+  EMPTY(new byte[] {0x00, 0x00}),
+  /** A snappy compressed, SSZ encoded, signed beacon block */
+  COMPRESSED_SIGNED_BEACON_BLOCK(new byte[] {0x01, 0x00}),
+  /** A snappy compressed, SSZ encoded, beacon state */
+  COMPRESSED_BEACON_STATE(new byte[] {0x02, 0x00}),
```
jframe, line 26: Also can remove these as we don't need to parse era files

util/src/main/java/org/hyperledger/besu/util/e2/E2ExecutionBlockBody.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package org.hyperledger.besu.util.e2;
+
+/**
+ * Represents an execution block body in an E2 file
+ *
+ * @param block The execution block
+ * @param slot The slot number
+ */
+public record E2ExecutionBlockBody(byte[] block, int slot) {}
```
jframe, line 23: Slot should be blockIndex from my understanding of the spec

util/src/main/java/org/hyperledger/besu/util/e2/E2ExecutionBlockHeader.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package org.hyperledger.besu.util.e2;
+
+/**
+ * Represents an execution block header in an E2 file
+ *
+ * @param header The execution block header
+ * @param slot The slot number
+ */
+public record E2ExecutionBlockHeader(byte[] header, int slot) {}
```
jframe, line 23: slot should be blockIndex

util/src/main/java/org/hyperledger/besu/util/e2/E2ExecutionBlockReceipts.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package org.hyperledger.besu.util.e2;
+
+/**
+ * Represents an execution block's transaction receipts in an E2 file
+ *
+ * @param receipts The execution block's transaction receipts
+ * @param slot The slot number
+ */
+public record E2ExecutionBlockReceipts(byte[] receipts, int slot) {}
```
jframe, line 23: slot should be blockIndex

util/src/main/java/org/hyperledger/besu/util/io/E2StoreReader.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.common.primitives.Bytes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xerial.snappy.SnappyFramedInputStream;
+
+/** Reads E2 store file such a .era and .era1 */
+public class E2StoreReader {
```
jframe, line 42: Think it would be nicer to have this in the e2 package as it's not general io it's specific to era

util/src/main/java/org/hyperledger/besu/util/io/E2StoreReader.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+        case COMPRESSED_SIGNED_BEACON_BLOCK -> {
+          byte[] compressedSignedBeaconBlockArray = inputStream.readNBytes(length);
+          try (SnappyFramedInputStream decompressionStream =
+              snappyFactory.createFramedInputStream(compressedSignedBeaconBlockArray)) {
+            // TODO: decode with SSZ
+            listener.handleSignedBeaconBlock(
+                new E2SignedBeaconBlock(decompressionStream.readAllBytes(), slot++));
+          }
+        }
```
jframe, line 122: don't need to decode these

util/src/main/java/org/hyperledger/besu/util/io/E2StoreReader.java
```diff
@@ -0,0 +1,177 @@
...
+        case COMPRESSED_EXECUTION_BLOCK_HEADER -> {
+          byte[] compressedExecutionBlockHeader = inputStream.readNBytes(length);
+          try (SnappyFramedInputStream decompressionStream =
```
jframe, line 125: Since we have already read the bytes it would be simpler to use `SnappyCompressor.decompress` to decompress snappy instead of creating the snappy input stream

Matilda-Clerke: Unfortunately, the use of snappy framing means this won't work afaik. When I tried it on the beacon state initially, it failed to decompress.

util/src/main/java/org/hyperledger/besu/util/e2/E2BeaconState.java
Matilda-Clerke: I was initially writing a generic e2 file reader. Are we unlikely to use era files for populating the chain beyond the merge? I realise we can populate it through the p2p network, or potentially from portal, but it shouldn't be difficult to support this too.

util/src/main/java/org/hyperledger/besu/util/era1/Era1Reader.java
```diff
@@ -0,0 +1,130 @@
...
+          try (SnappyFramedInputStream decompressionStream =
+              snappyFactory.createFramedInputStream(compressedReceipts)) {
+            listener.handleExecutionBlockReceipts(
+                new Era1ExecutionBlockReceipts(decompressionStream.readAllBytes(), slot++));
+          }
+        }
+        case BLOCK_INDEX -> {
+          ByteArrayInputStream slotIndexInputStream =
+              new ByteArrayInputStream(inputStream.readNBytes(length));
+          long startingSlot =
```
jframe, line 103: rename to blockIndex and same above for slotIndexInputStream as we are only dealing with block indexes and not slots

util/src/main/java/org/hyperledger/besu/util/era1/Era1Reader.java
```diff
@@ -0,0 +1,130 @@
...
+                "index count does not match number of indexes present for InputStream: {}",
+                inputStream);
+          }
+          listener.handleBlockIndex(new Era1BlockIndex(startingSlot, indexes));
+        }
+      }
+    }
+  }
+
+  private long convertLittleEndianBytesToLong(final byte[] bytes) {
```
jframe, line 124: This should go in EndianUtils and consider using either inline in EndianUtils or directly from bouncy castle https://github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/util/Pack.java#L303 as the bit shift is quite likely more efficient than using a byte buffer

jframe (approved):

util/src/main/java/org/hyperledger/besu/util/era1/Era1Reader.java
```diff
@@ -0,0 +1,129 @@
...
+          }
+          listener.handleBlockIndex(new Era1BlockIndex(startingBlockIndex, indexes));
+        }
+      }
+    }
+  }
+
+  private long convertLittleEndianBytesToLong(final byte[] bytes) {
+    int additionalBytes = Long.BYTES - bytes.length;
+    return Pack.littleEndianToLong(Arrays.concatenate(bytes, new byte[additionalBytes]), 0);
```
pinges, line 127: There is a Pack.littleEndianToLong with a length, so you don't have to do that concatenation: https://github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/util/Pack.java#L319

pinges (commented):
Looking good, just one nit.
Is there a small example era 1 file that we cold use to test this?

util/src/main/java/org/hyperledger/besu/util/era1/Era1Reader.java
Matilda-Clerke: Ooh nice, I'll add that in

Matilda-Clerke:
era1 files can be found at https://mainnet.era1.nimbus.team/ . I did some manual testing with the first few files for this PR, but nothing worth checking in. The next PR is https://github.com/hyperledger/besu/pull/8352 and allows us to use the blocks import subcommand to import era1 files

PR #8330, opened 2025-02-19
Pull in Consensys/tuweni release

siladu:
Migrating from https://github.com/tmio/tuweni to https://github.com/consensys/tuweni/

Also includes:
- reduce check argument args allocations - https://github.com/Consensys/tuweni/pull/10 which should reduce Integer and int[] and therefore gc pressure

Testing: 
- https://github.com/hyperledger/besu/pull/8330#issuecomment-2678014967
- https://github.com/hyperledger/besu/pull/8330#issuecomment-2728372551


CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
 - Allow plugins to propose transactions during block creation [#8268](https://github.com/hyperledger/besu/pull/8268)
 - Update `eth_getLogs` to return a `Block not found` error when the requested block is not found. [#8290](https://github.com/hyperledger/besu/pull/8290)
 - Improve Conflict Detection in Parallelization by Considering Slots to Reduce False Positives. [#7923](https://github.com/hyperledger/besu/pull/7923)
+- Replace tuweni libs with https://github.com/Consensys/tuweni
+- Performance: Consensys/tuweni 25.2.0 includes a signicant improvement that reduces memory allocations across Besu
```
macfarla, line 26:
```suggestion
- Performance: Consensys/tuweni 25.2.0 includes a significant improvement that reduces memory allocations across Besu
```

siladu: Changed to:
> - Performance: Consensys/tuweni 2.6.0 reduces boxing/unboxing overhead on some EVM opcodes, like PushX and Caller

siladu:
Testing:
* Successfully synced holesky, sepolia and mainnet nodes

* [Nethermind Caller benchmark](https://github.com/NethermindEth/gas-benchmarks/tree/main/tests/Caller) - observed absence of PushOperation.staticOperation/checkLength (also checked Push0 and Push0Pop)

nethermind-caller-control

nethermind-caller-tuweni-2 6 0

* [Nethermind Caller benchmark](https://github.com/NethermindEth/gas-benchmarks/tree/main/tests/Caller) - No significant difference in MGas measurements - surprising result, need to validate this.

develop = control
nethermind-caller-control-benchmark-report

develop-8a93eab = this PR
nethermind-caller-tuweni-2 6 0-benchmark-report


* Performance test on mainnet on 2x this PR vs 2x control (main) for two instance sizes: Azure Standard_D4 and Azure Standard_D8.
  * No difference to sync times (expected since majority of sync workload is not executing transactions)
  * No significant difference to newPayload response time - one theory is that the number of push operations are relatively low

D4s

TODO

D8s

TODO

ahamlat:
The CPU profiling is different, we can see that checkLength is not a bottelneck anymore but Nethermind reports the same results in terms of throughput. I would suggest to rerun the tests with profiling enabled on **an Azure VM** and compare between the version that doesn't have this commit with the version that has this commit (PR). The only difference between both versions should be this PR.

siladu:
Retested mainnet sync with excluded io.tmio.tuweni libs and ran for a short period on a sepolia validator to test block production. 
Result was neutral: no noticeable improvement or regression.

daniellehrner (approved):

ahamlat (approved):
I haven't seen any regression neither, but going to keep an eye on it. Let's ship it :ship:

macfarla (approved):

PR #8352, opened 2025-02-26
7935: add era1 format to blocks importer

Matilda-Clerke:
## PR description
Adds the ability to import blocks from era1 files
## Fixed Issue(s)


besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
```diff
@@ -0,0 +1,194 @@
...
+      if (!fastSyncDir.mkdir()) {
+        throw new RuntimeException("Unable to create " + fastSyncDir + " directory");
+      }
+      File pivotBlockHeaderFile =
+          fastSyncDir.toPath().resolve(PIVOT_BLOCK_HEADER_FILE_NAME).toFile();
+      LOG.info("Creating " + pivotBlockHeaderFile);
+      if (!pivotBlockHeaderFile.createNewFile()) {
+        throw new RuntimeException("Unable to create " + pivotBlockHeaderFile);
+      }
+    }
```
garyschulte, line 97: what is the point of the pivot block header file for an era importer?  shouldn't that be strictly a fast/snap sync concern, and far beyond the merge block (which would be the last block in an era1 file) ?

is this just a matter of ensuring we do not truncate block storage when doing a fast/snap sync?  if so - would it make more sense to address it in the fast sync code rather than adding a mock pivot block file here?

garyschulte (commented):
This is a really tight implementation :+1:  One outstanding comment regarding the pivot block concern that IMO doesn't belong in era1 importer.

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
Matilda-Clerke: When I initially ran the first 100 era1 files and then ran besu, it noticed the pivot block header file was missing and attempted to full sync instead, but failed due to lack of trie logs for the imported blocks. Putting in the pivot block header file allows us to trick besu into doing trying to continue a fast/snap/checkpoint sync. This piece is probably the part I'm least confident in, so I'd be happy to hear of other ways to allow besu to sync after initially importing from era1 files.

garyschulte: I will raise with @matkt tomorrow morning at s&t.  I think signaling to snap and fast sync pipelines to behave rationally when we have blockchain data is a good idea.  I am just not sold on a mock pivot block file.

Matilda-Clerke: Ok, let me know what you come up with. It sounds like maybe modifying the sync pipelines to just be a bit more lenient and forgiving of a missing pivot block header will be a more elegant solution.

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
```diff
@@ -0,0 +1,194 @@
...
+        ScheduleBasedBlockHeaderFunctions.create(protocolSchedule);
+    final ProtocolContext context = controller.getProtocolContext();
+
+    Era1Reader reader = new Era1Reader(new SnappyFactory());
+
+    final List<Future<BlockHeader>> headersFutures = new ArrayList<>(ERA1_BLOCK_COUNT_MAX);
+    final List<Future<BlockBody>> bodiesFutures = new ArrayList<>(ERA1_BLOCK_COUNT_MAX);
+    final List<Future<List<TransactionReceipt>>> receiptsFutures =
+        new ArrayList<>(ERA1_BLOCK_COUNT_MAX);
+    reader.read(
```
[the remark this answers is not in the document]
Matilda-Clerke: Maximum size for an era1 file is 8192 blocks. It runs fine locally, but may run out of memory on some lower spec machines, especially as the later blocks are more full. Do you think it's worth sacrificing performance to minimise memory usage?

Matilda-Clerke:
I've removed the code dealing with the pivot block header file, as I anticipate it won't be included in the final code. Easy enough to add back if desired.

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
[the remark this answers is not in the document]
jframe: Looking at files here https://era1.ethportal.net/ max I see is about 500MB, not sure what compression ratio we get with snappy but at most probably only a few GB of memory. Probably ok since you won't be running a full Besu while doing the import.

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
matkt: so we are talking about this check 

```java
else if (fastSyncState.getPivotBlockHeader().isEmpty()
        && protocolContext.getBlockchain().getChainHeadBlockNumber()
            != BlockHeader.GENESIS_BLOCK_NUMBER) {
      LOG.info(
          "Checkpoint sync was requested, but cannot be enabled because the local blockchain is not empty.");
      return Optional.empty();
    }
```

this check in my opinion was done to avoid that the node restarts with a fastsync or snapsync when we restart. this is the only reason I see for this check. before there was a reason because we were saving state from top to bottom and we did not really have a way to know if the state was synced or not. now that we save the root node of the trie at the end. I think we can replace this check with rather a check of the existence of the root node.

If it exists we return empty and if it does not exist we launch the sync

Matilda-Clerke: Sorry, this part of the work is a little above my head. Where should I be looking for this root node? At first I thought maybe you meant `protocolContext.getWorldStateArchive().getWorldState().rootHash()`, but this seems to be populated already after importing the era1 files.

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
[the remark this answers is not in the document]
Matilda-Clerke: Great. It's currently making really good use of the CPU, so hopefully the performance is good.

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
matkt: Populated ? This part just download the blockchain, right? It should return the hash of an empty tree. https://github.com/hyperledger/besu/blob/f92780859d7c3a31929b370cf6fc23e70f8ffb47/datatypes/src/main/java/org/hyperledger/besu/datatypes/Hash.java#L40

I think it can be a good solution. Bonsai will be better because if the state is empty it should return Hash.ZERO while Forest will return the hash of an empty tree. In both cases I think it will tell if the state has been sync

Matilda-Clerke:
Ok, now seeking approval to merge this PR as-is, with additional changes to sync pipelines coming in a new PR

jframe (approved):

Issue #8356, opened 2025-02-26
eth_getTransactionCount() can cause `java.util.NoSuchElementException`


matthew1001:
### Steps to Reproduce
I hit this by running a basic hardhat app looping round doing simple storage `set()` calls at ~ 10TPS to a QBFT chain. Importantly hardhat was requesting the next nonce for every TX, which was frequent enough to eventually hit this race condition:

```
2025-02-26 15:56:43.046+00:00 | vert.x-worker-thread-8 | ERROR | BaseJsonRpcProcessor | Error processing method: eth_getTransactionCount ["0x00e1ed642904f2e8af9108a4383c95e24532e0b8","pending"]
java.util.NoSuchElementException
        at java.base/java.util.TreeMap.key(TreeMap.java:1637)
        at java.base/java.util.TreeMap.lastKey(TreeMap.java:309)
        at org.hyperledger.besu.ethereum.eth.transactions.sorter.PendingTransactionsForSender.maybeNextNonce(PendingTransactionsForSender.java:99)
        at org.hyperledger.besu.ethereum.eth.transactions.sorter.AbstractPendingTransactionsSorter.getNextNonceForSender(AbstractPendingTransactionsSorter.java:419)
        at org.hyperledger.besu.ethereum.eth.transactions.TransactionPool.getNextNonceForSender(TransactionPool.java:569)
        at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.EthGetTransactionCount.pendingResult(EthGetTransactionCount.java:74)
        at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.AbstractBlockParameterOrBlockHashMethod.handleParamTypes(AbstractBlockParameterOrBlockHashMethod.java:100)
        at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.AbstractBlockParameterOrBlockHashMethod.response(AbstractBlockParameterOrBlockHashMethod.java:151)
        at org.hyperledger.besu.ethereum.api.jsonrpc.execution.BaseJsonRpcProcessor.process(BaseJsonRpcProcessor.java:46)
        at org.hyperledger.besu.ethereum.api.jsonrpc.execution.TracedJsonRpcProcessor.process(TracedJsonRpcProcessor.java:54)
        at org.hyperledger.besu.ethereum.api.jsonrpc.execution.TimedJsonRpcProcessor.process(TimedJsonRpcProcessor.java:45)
[...]

PR #8362, opened 2025-02-27
Safely handle race condition between isEmpty() and lastKey() being called

matthew1001:
## PR description
Safely handle cases where a map's `isEmpty()` call returns false, but the map is updated before the call to `lastKey()` is made.

I've looked through the rest of the codebase and fixed the one other place I think we could hit this. For other code paths that use the same combination of `isEmpty()` and `lastKey()` I don't think they are likely to suffer from the same issue. If they do the `NoSuchElementException` will clearly point to where else we need to fix it.

No change log entry as it's just an internal fix that nobody else has reported seeing yet.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/8356


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractTransactionsLayer.java
```diff
@@ -326,11 +327,19 @@ protected TransactionAddedResult addToNextLayer(
...
-      nextLayerDistance = (int) (pendingTransaction.getNonce() - (senderTxs.lastKey() + 1));
+      try {
+        nextLayerDistance = (int) (pendingTransaction.getNonce() - (senderTxs.lastKey() + 1));
+      } catch (NoSuchElementException nse) {
+        // There is a timing condition where isEmpty() returns false, then another thread removes
+        // the last key before our call to lastKey(). There's no benefit to us adding a mutex to
+        // prevent the other thread updating the map so we just catch the exception and treat the
+        // map as empty.
+        nextLayerDistance = distance;
+      }
```
fab-10, line 342: this change is not needed here since each layer is externally synchronized and thread safe

fab-10 (changes requested):
no need to apply the change to the layered txpool, since it is thread safe

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/layered/AbstractTransactionsLayer.java
matthew1001: OK thanks for confirming, I've removed that change

fab-10 (approved):

PR #8363, opened 2025-02-27
Support pending transaction score when saving and restoring txpool

fab-10:

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
@@ -14,6 +14,7 @@
  */
 package org.hyperledger.besu.ethereum.eth.transactions;
 
+import static org.hyperledger.besu.ethereum.eth.transactions.PendingTransaction.MAX_SCORE;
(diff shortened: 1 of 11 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java
```diff
@@ -870,12 +875,19 @@ private void executeLoadFromDisk() {
...
                     .map(
                         line -> {
-                          final boolean isLocal = line.charAt(0) == 'l';
+                          int i = 0;
+                          final var sbScore = new StringBuilder();
+                          while ("1234567890-".indexOf(line.charAt(i)) >= 0) {
+                            sbScore.append(line.charAt(i++));
+                          }
+                          final byte score =
+                              sbScore.isEmpty() ? MAX_SCORE : Byte.parseByte(sbScore.toString());
```
macfarla, line 884: maybe would be clearer to extract 878-884 to a method?

macfarla (commented):
do we need to doc this? where does the score come from?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPool.java

fab-10:
> do we need to doc this? where does the score come from?

Score for txpool come from this https://github.com/hyperledger/besu/pull/7576, and there is no need to update doc, because not saving the score is basically a bug

macfarla (approved):

Issue #6203, opened 2023-11-22
Implement eth_simulatev1


non-fungible-nelson:
was eth_multicallV1

  

### Description
Implementation of https://github.com/ethereum/execution-apis/pull/484
[More detail.](https://mirror.xyz/killaridev.eth/NXR9v4r8b4SWl-hbqJqRQyIAcE4GACobzMObREvL0fo)

This introduces a new RPC method eth_multicallV1 allows users to make complex RPC calls to Ethereum nodes. Compared to eth_call, eth_multicallV1 has following extra features:

* You can encapsulate multiple dependent calls in a single call
* The calls happen inside blocks. You can simulate multiple blocks that can be arbitrary far from each other
* Block variables can be overridden (e.g. time)
* Account state can be overridden for every block (e.g. code and balance)
* It is possible to override precompiles (e.g. ecrecover) with arbitrary EVM code
* ETH transfers produce logs similar to ERC20 logs
* Validation mode. You can choose to do very strict simulation or more relaxed one similar to eth_call

Presentation: https://docs.google.com/presentation/d/1lEaqHTY3ud8pe6VAFwLkb-jdoHpTMBfuF1K9OKy-azs/edit#slide=id.p

As a node runner, I want to be able to have access to all standard RPCs. This includes the speced RPC `eth_multicallV1`. 

### Acceptance Criteria
* Passing any released reference tests in the execution APIs 
* Passing Hive tests: https://github.com/KillariDev/rpctestgen
* Results consistent with [Geth](https://github.com/s1na/go-ethereum/tree/multicall), [Nethermind](https://github.com/NethermindEth/nethermind/tree/feature/eth_multicall) and/or other clients 
* Performance in line with Geth and/or other clients 
* Tested on Mainnet state

Gabriel-Trintinalia:
#8390

PR #8375, opened 2025-03-04
eth_simulateV1 - Fill gaps between block calls

Gabriel-Trintinalia:
## PR description

`eth_simulateV1` allows you to simulate blocks in the future. This PR changes the block simulator to fill the calls with intermediate block state calls if the block number is in the future.

see https://github.com/hyperledger/besu/issues/6203


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockStateCalls.java
```diff
@@ -0,0 +1,134 @@
...
+      final List<BlockStateCall> blockStateCalls,
+      final BlockStateCall blockStateCall,
+      final long currentBlockNumber,
+      final long currentTimestamp) {
+    long blockNumber = blockStateCall.getBlockOverrides().getBlockNumber().orElseThrow();
+    long timestamp = blockStateCall.getBlockOverrides().getTimestamp().orElseThrow();
+    if (blockNumber <= currentBlockNumber) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Block number %d is invalid. It must be greater than %d.",
```
pinges, line 92: Message could be a bit more informative. Something like "trying to add a call at block number x, while current block number is y" or along these lines.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockStateCalls.java
```diff
@@ -0,0 +1,134 @@
...
+              blockNumber, currentBlockNumber));
+    }
+    if (timestamp <= currentTimestamp) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Timestamp %d is invalid. It must be greater than %d.", timestamp, currentTimestamp));
```
pinges, line 98: see the comment above

pinges (commented):
Might need some better messages for exceptions

jframe (approved):

PR #8390, opened 2025-03-06
eth_simulateV1 - Refactoring of Block Simulator

Gabriel-Trintinalia:
## PR description
- The simulation baseBlockHeader must be the outcome of the previous simulation if more than one BlockCall.
- Refactored the Block Simulator for improved readability.
- Implemented a proper return type for the Block Simulator.
- Enabled parameterization of certain components within the Transition Simulator to enhance compatibility with the Block Simulator.
see #6203 6203


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulationResult.java
```diff
@@ -44,15 +41,36 @@ public BlockBody getBlockBody() {
...
+                        block.getHeader().getNumber(),
+                        block.getHash(),
+                        transactionSimulation.result().transaction().getHash(),
+                        block
+                            .getBody()
+                            .getTransactions()
+                            .indexOf(transactionSimulation.result().transaction()),
+                        false)));
+    return logs;
+  }
```
fab-10, line 75: nit: use `.map(...).toList()` instead of creating the list outside the stream

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
```diff
@@ -68,6 +73,18 @@
  * header, transaction receipts, and other relevant information.
  */
 public class BlockSimulator {
+
+  private static final ImmutableTransactionValidationParams STRICT_VALIDATION_PARAMS =
+      ImmutableTransactionValidationParams.builder()
+          .from(TransactionValidationParams.transactionSimulator())
+          .build();
```
fab-10, line 80: It seems you are only aliasing an existing object, so a little optimization
```suggestion
  private static final TransactionValidationParams STRICT_VALIDATION_PARAMS =
     TransactionValidationParams.transactionSimulator());
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
```diff
@@ -68,6 +73,18 @@
...
+
+  private static final ImmutableTransactionValidationParams SIMULATION_PARAMS =
+      ImmutableTransactionValidationParams.builder()
+          .from(TransactionValidationParams.transactionSimulator())
+          .isAllowExceedingBalance(true)
+          .build();
```
fab-10, line 86: same
```suggestion
  private static final TransactionValidationParams SIMULATION_PARAMS =
      TransactionValidationParams.transactionSimulatorAllowExceedingBalance();
```

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockOverrides.java
```diff
@@ -392,13 +366,13 @@ public Builder extraData(final Bytes extraData) {
...
+     * Sets the mix hash
      *
-     * @param mixHashOrPrevRandao the mix hash or previous Randao to set
+     * @param mixHash the mix hash
      * @return the builder instance
      */
-    public Builder mixHashOrPrevRandao(final Hash mixHashOrPrevRandao) {
-      this.mixHashOrPrevRandao = mixHashOrPrevRandao;
+    public Builder mixHashOrPrevRandao(final Hash mixHash) {
+      this.mixHashOrPrevRandao = mixHash;
```
fab-10, line 375: should the method and field be renamed to remove "prevrandao"?

fab-10 (commented):
Some suggestions from that first pass, I will complete the review tomorrow

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
```diff
@@ -114,151 +125,172 @@ public List<BlockSimulationResult> process(
...
-      BlockSimulationResult simulationResult =
+    BlockHeader currentBlockHeader = blockHeader;
+    for (BlockStateCall stateCall : blockStateCalls) {
+      BlockSimulationResult result =
           processBlockStateCall(
-              header, blockStateCall, worldState, blockSimulationParameter.isValidation());
-      simulationResults.add(simulationResult);
+              currentBlockHeader, stateCall, worldState, simulationParameter.isValidation());
+      results.add(result);
+      currentBlockHeader = result.getBlock().getHeader();
```
fab-10, line 149: is it correct to say that previously the block header was always the same for each call to `processBlockStateCall` while now it is update on every loop?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
```diff
@@ -114,151 +125,172 @@ public List<BlockSimulationResult> process(
...
-    long timestamp = blockOverrides.getTimestamp().orElse(header.getTimestamp() + 1);
-    ProtocolSpec newProtocolSpec = protocolSchedule.getForNextBlockHeader(header, timestamp);
+    ProtocolSpec protocolSpec =
+        protocolSchedule.getForNextBlockHeader(
+            baseBlockHeader, blockOverrides.getTimestamp().orElseThrow());
 
-    // Apply block header overrides and state overrides
-    BlockHeader blockHeader = applyBlockHeaderOverrides(header, newProtocolSpec, blockOverrides);
-    blockStateCall.getStateOverrideMap().ifPresent(overrides -> applyStateOverrides(overrides, ws));
+    BlockHeader overridenBaseblockHeader =
```
fab-10, line 173:
```suggestion
    BlockHeader overridenBaseBlockHeader =
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
```diff
@@ -288,94 +320,63 @@ protected void applyStateOverrides(
...
+                    .getBaseFeePerGas()
+                    .orElseGet(
+                        () ->
+                            shouldValidate
+                                ? getNextBaseFee(newProtocolSpec, header, blockNumber)
+                                : Wei.ZERO))
+            .extraData(blockOverrides.getExtraData().orElse(Bytes.EMPTY))
+            .parentBeaconBlockRoot(Bytes32.ZERO)
+            .prevRandao(blockOverrides.getMixHashOrPrevRandao().orElse(Bytes32.ZERO));
+
```
fab-10, line 357: what about moving the functionality of overriding an existing block in the `BlockHeaderBuilder` class for general use, for example `BlockHeaderBuilder::copyWithOverrides`?

fab-10 (commented):
Nice refactor, few suggestions/questions

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulationResult.java
Gabriel-Trintinalia: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
Gabriel-Trintinalia: done

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockOverrides.java
Gabriel-Trintinalia: renamed the field to mixHashOrPrevRandao (the call object accepts both mixHash & prevrandao)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
Gabriel-Trintinalia: Yes, it is correct. The next block simulation needs to be based on the outcome of the previous one. I will update the PR description.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
Gabriel-Trintinalia: done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
Gabriel-Trintinalia: This is a nice suggestion, however in this case, it is not only a copy with overrides, the block simulator is assembling the next block base for the simulation, so it needs data from other sources such as `getNextGasLimit`, `getNextBaseFee`, `miningConfiguration`, etc.

fab-10 (approved):

PR #8394, opened 2025-03-06
Decode deposit log data without web3j

### LINKED ISSUE #8391: Remove Web3j dependency from parsing deposit logs
We use the 3rd party library Web3j to parse the logs of the deposit contract. As this is a critical path and we only use a very small subset of this library in the path we have decided to remove the dependency and replace it with our own parsing.

daniellehrner:
## PR description

Decode the deposit log data without relying on Web3J

## Fixed Issue(s)
fixes #8391 


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/DepositLogDecoder.java
```diff
@@ -14,27 +14,92 @@
...
+    if (data.size() != DEPOSIT_LOG_LENGTH) {
+      throw new IllegalArgumentException(
+          "Invalid deposit log length. Must be "
+              + DEPOSIT_LOG_LENGTH
+              + " bytes, but is "
+              + data.size()
+              + " bytes");
+    }
+
+    final Bytes depositData = data.slice(DATA_START_POSITION);
```
pinges, line 95: You could get rid of this slice if you were to just add the DATA_START_POSITION to each of the offsets.

daniellehrner: Good idea, I've done that.

macfarla (commented):
worth a changelog entry? 
and also can we now change the web3j dependency to be test scope only?

macfarla (approved):

pinges (approved):
LGTM

PR #8397, opened 2025-03-07
QBFT fast recovery by frequent multicasting of RC message 

pullurib:
## PR description
- When a node moves up a round, e.g. from 6 -> 7, it only sends a single round-change message
- This means that if a node misses the round-change message from another node it doesn't know the node is on round 7
- Nodes can publish their current round every N seconds while they are still in that round

Enabled by the experimental  option --Xqbft-fast-recovery . It's very effective when combined with --Xqbft-enable-early-round-change


pullurib:
- Added a Frequent multicaster class to resend a message to validators at regular intervals . This is used for round change messages upon using fast-recovery option
- Moved early-round-change , fast-recovery and frequent multicaster to final state 
- Testcases for frequent mutlicaster unit tests and ATs for new option.
-

macfarla (commented):
could this be a plugin?

pullurib:
@matthew1001 is working on moving QBFT functionality to a plugin . This would be a part of that plugin then.

matthew1001:
I think the plugin version of QBFT could be some way off @pullurib @macfarla Could I ask that we move this PR along in its current shape and the logic will get included in the shift to a QBFT plugin as a matter of coure.

macfarla:
> I think the plugin version of QBFT could be some way off @pullurib @macfarla Could I ask that we move this PR along in its current shape and the logic will get included in the shift to a QBFT plugin as a matter of coure.

sure, I don't think we need to block this on making it a plugin. I'm prob not the right person to review this PR in detail though, someone with more QBFT knowledge should do that.

macfarla:
@pullurib there's a compile error that needs fixing before this can be reviewed in detail

pullurib:
Fixed the compile error and merged latest main

app/src/main/java/org/hyperledger/besu/cli/options/QBFTOptions.java
```diff
@@ -38,6 +38,12 @@ public static QBFTOptions create() {
       hidden = true)
   private boolean enableEarlyRoundChange = false;
 
+  @CommandLine.Option(
+      names = {"--Xqbft-fast-recovery"},
+      description = "Enable fast recovery mode for QBFT consensus (experimental)",
+      hidden = true)
+  private boolean enableFastRecovery = false;
+
```
jframe, line 46: Would be good to also have the recovery time period configurable

consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/FrequentMessageMulticaster.java
```diff
@@ -0,0 +1,107 @@
...
+  @VisibleForTesting
+  public FrequentMessageMulticaster(final ValidatorMulticaster multicaster) {
+    this(multicaster, 5000);
+  }
+
+  private Runnable createPeriodicMulticastTask(
+      final MessageData message, final Collection<Address> denylist) {
+    return () -> {
+      LOG.debug(
+          "Broadcasting round change every {} ms on thread {}",
```
jframe: Since this isn't specific to round change messages:

```suggestion
          "Broadcasting message every {} ms on thread {}",
```

consensus/common/src/test/java/org/hyperledger/besu/consensus/common/bft/FrequentMessageMulticasterTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.jupiter.api.Test;
+
+class FrequentMessageMulticasterTest {
+  private FrequentMessageMulticaster multicaster;
+  private ValidatorMulticaster mockValidatorMulticaster;
+
+  @BeforeEach
+  void setUp() {
+    mockValidatorMulticaster = mock(ValidatorMulticaster.class);
+    multicaster = new FrequentMessageMulticaster(mockValidatorMulticaster, 5000);
```
jframe, line 39: Better to use a smaller interval for testing to decrease the test time taken. Maye 1000 ms instead?

consensus/qbft-core/src/main/java/org/hyperledger/besu/consensus/qbft/core/types/QbftFinalState.java
```diff
@@ -104,4 +104,25 @@ public interface QbftFinalState {
    * @return true if the local node is the proposer for the given round, false otherwise
    */
   boolean isLocalNodeProposerForRound(ConsensusRoundIdentifier roundIdentifier);
+
+  /**
+   * Gets the frequent round change multicaster.
+   *
+   * @return the frequent round change multicaster
+   */
+  public ValidatorMulticaster getFrequentRCMulticaster();
```
jframe, line 113: This is misnamed as it resending more than just round changes

```suggestion
  public ValidatorMulticaster getFrequentMulticaster();
```

consensus/qbft-core/src/main/java/org/hyperledger/besu/consensus/qbft/core/types/QbftFinalState.java
```diff
@@ -104,4 +104,25 @@ public interface QbftFinalState {
...
+   * @return the boolean
+   */
+  public boolean isEarlyRoundChangeEnabled();
+
+  /**
+   * Is fast recovery enabled boolean.
+   *
+   * @return the boolean
+   */
+  public boolean isFastRecoveryEnabled();
```
jframe, line 127: Rather not expose config on the `QbftFinalState`. Is it possible to just pass those flags into the components that need that config?

consensus/qbft-core/src/main/java/org/hyperledger/besu/consensus/qbft/core/statemachine/QbftBlockHeightManager.java
```diff
@@ -97,7 +96,11 @@ public QbftBlockHeightManager(
     this.roundFactory = qbftRoundFactory;
     this.validatorProvider = validatorProvider;
     this.transmitter =
-        new QbftMessageTransmitter(messageFactory, finalState.getValidatorMulticaster());
+        new QbftMessageTransmitter(
+            messageFactory,
+            (finalState.isFastRecoveryEnabled()
+                ? finalState.getFrequentRCMulticaster()
+                : finalState.getValidatorMulticaster()));
```
jframe, line 103: Rather than exposing the `isFastRecoveryEnabled` and `getFrequentRCMulticaster` I think we could just construct the `QbftMessageTransmitter` in `QbftBesuControllerBuilder` configured with the correct multicaster.

siladu:
@pullurib from the commits it looked like maybe you tried an approach using RoundTimer first...is there a reason that didn't work out and you settled on a separate retrying thread that needs to be stopped at the appropriate time?

At first glance, it seems complex to have an executor service as well as a series of Timers managing the messages...but maybe there's a good reason for it?

pullurib:
> @pullurib from the commits it looked like maybe you tried an approach using RoundTimer first...is there a reason that didn't work out and you settled on a separate retrying thread that needs to be stopped at the appropriate time?
> 
> At first glance, it seems complex to have an executor service as well as a series of Timers managing the messages...but maybe there's a good reason for it?

Yeah, needed something to trigger a multicast every few seconds . First attempt had the logic spread across timer and multicaster but ran into some race conditions around timer expiry. Also, frequent multicasting was not exactly timer's concern. So bundled it all together in  FrequentMulticaster

consensus/qbft-core/src/main/java/org/hyperledger/besu/consensus/qbft/core/types/QbftFinalState.java
pullurib: Only the round change message is resent every few seconds within the same round . This multicaster is used only for that purpose .

app/src/main/java/org/hyperledger/besu/cli/options/QBFTOptions.java
pullurib: can make the interval between resending RC messages configurable . I'll add that as an optional parameter to this option

consensus/qbft-core/src/main/java/org/hyperledger/besu/consensus/qbft/core/types/QbftFinalState.java
pullurib: ok, this was convenient than passing down the flags but would you mind throwing some light on why it's not preferred? I understand QbftFinalState to be more like context for qbft to hold components , etc needed for runtime .

PR #8403, opened 2025-03-12
Update sync pipelines to continue from era1 import

Matilda-Clerke:
## PR description
Update FAST sync and derivatives to sync after previously importing blocks from era1 files

## Fixed Issue(s)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/checkpointsync/CheckpointDownloaderFactory.java
```diff
@@ -92,11 +92,13 @@ public static Optional<FastSyncDownloader<?>> createCheckpointDownloader(
...
                   snapContext.addAccountToHealingList(
                       CompactEncoding.bytesToPath(address.addressHash())));
-    } else if (fastSyncState.getPivotBlockHeader().isEmpty()
-        && protocolContext.getBlockchain().getChainHeadBlockNumber()
-            != BlockHeader.GENESIS_BLOCK_NUMBER) {
+    } else if (protocolContext
+        .getWorldStateArchive()
+        .getWorldState()
+        .rootHash()
+        .equals(Hash.EMPTY_TRIE_HASH)) {
```
fab-10, line 99: try to understand the change, should not this condition be negated? I mean if the root hash is not empty?

fab-10 (commented):
is it correct

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/checkpointsync/CheckpointDownloaderFactory.java
Matilda-Clerke: Hmm, you're right. We should be checking that the root hash is not empty. I guess this indicates that actually it's not currently set to the `EMPTY_TRIE_HASH`. I'll have to figure out what it's actually returning and check for that.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/FastDownloaderFactory.java
```diff
@@ -48,6 +48,8 @@
 public class FastDownloaderFactory {
 
   protected static final String FAST_SYNC_FOLDER = "fastsync";
+  protected static final Hash GENESIS_STATE_ROOT =
+      Hash.fromHexString("0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544");
```
fab-10, line 52: Should not this depend on the actual network or genesis used, so the value can change, maybe @matkt has suggestion on how to check if the world state is empty

Matilda-Clerke: Good point. I'm not sure anyone really cares about the ability to import era1 files on sepolia, but it could be useful for testing. Afaik, we don't really know which network the era1 files we're importing are for, so I could add --network and --data-path to allow people to better specify what they're doing. The network option would allow us to just check a different genesis state root, but alternative ways of checking whether a sync is already completed might be preferable

fab-10: to check if the initial sync is done you can use https://github.com/hyperledger/besu/blob/b7a0b91f8469a0631dbde946846a2dc45dc7b2cf/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/state/SyncState.java#L319

matkt: yes clearly it's not good like that. So when we are passing this code the state is not empty and has the genesis modification already ?

matkt: if it is the case why not using ? 

Hash genesisStateRoot =  protocolContext.getBlockchain().getGenesisBlock().getHeader().getStateRoot()

Matilda-Clerke: Ok, I've switched to using `isInitialSyncPhaseDone`, which I think captures the desired behaviour

Matilda-Clerke: > I could add --network and --data-path to allow people to better specify what they're doing
Turns out this isn't necessary, as the user can specify these already under the regular besu command. e.g.
`./bin/besu --network=SEPOLIA blocks import --format=ERA1 ~/era1/*.era1`

fab-10 (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/FastDownloaderFactory.java
```diff
@@ -48,6 +48,8 @@
 public class FastDownloaderFactory {
 
   protected static final String FAST_SYNC_FOLDER = "fastsync";
+  protected static final Hash GENESIS_STATE_ROOT =
```
Gabriel-Trintinalia, line 51: Can we remove this constant?

Matilda-Clerke: Yep, good spotting

macfarla:
maybe worth a changelog entry?

PR #8406, opened 2025-03-12
feature: implement eth_simulateV1

Gabriel-Trintinalia:
## PR description
Adds support for eth_simulateV1

Spec: https://github.com/ethereum/execution-apis/pull/484
Fixes: #6203 

Currently passing 87/97 hive tests


macfarla (commented):
few comments. needs a changelog entry

evm/src/main/java/org/hyperledger/besu/evm/tracing/EthTransferLogOperationTracer.java
```diff
@@ -0,0 +1,109 @@
...
+import com.google.common.collect.ImmutableList;
+import org.apache.tuweni.bytes.Bytes;
+
+/**
+ * Tracer that emits logs for all transfers that occur during the execution of a transaction.
+ *
+ * <p>Logs are emitted for all transfers that occur during the execution of a transaction. This
+ * includes both value transfers and self-destructs.
+ */
+public class EthTransferLogOperationTracer implements OperationTracer {
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: if `traceTransfers` is true, then eth_simulateV1 will also return logs generated by Eth transfers using this tracer. More info can be found here: https://github.com/ethereum/execution-apis/blob/e56d3208789259d0b09fa68e9d8594aa4d73c725/docs/ethsimulatev1-notes.md#eth-transfer-logs

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockOverrides.java
```diff
@@ -353,7 +327,7 @@ public Builder baseFeePerGas(final Wei baseFeePerGas) {
      * @param blobBaseFee the blob base fee to set
      * @return the builder instance
      */
-    public Builder blobBaseFee(final Long blobBaseFee) {
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: AFAIK there is only one plugin using `BlockOverrides` and it does not set the `blobBaseFee`

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockOverrides.java
```diff
@@ -303,17 +288,6 @@ public Builder blockHash(final Hash blockHash) {
       return this;
     }
 
-    /**
-     * Sets the previous Randao.
-     *
-     * @param prevRandao the previous Randao to set
-     * @return the builder instance
-     */
-    public Builder prevRandao(final Bytes32 prevRandao) {
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: No, because `Bytes32` is the base of `Hash`

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthSimulateV1.java
```diff
@@ -0,0 +1,163 @@
...
+            apiConfiguration.getGasCap());
+  }
+
+  @Override
+  public String getName() {
+    return RpcMethod.ETH_SIMULATE_V1.getMethodName();
+  }
+
+  @Override
+  protected BlockParameterOrBlockHash blockParameterOrBlockHash(
```
fab-10, line 77: Should this method also support pending as block tag?

fab-10 (commented):
Partial review. A question about support for `pending` block tag

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
```diff
@@ -135,12 +147,24 @@ public List<BlockSimulationResult> process(
         fillBlockStateCalls(simulationParameter.getBlockStateCalls(), blockHeader);
 
     BlockHeader currentBlockHeader = blockHeader;
+    HashMap<Long, Hash> blockHashCache = new HashMap<>();
```
fab-10, line 150: opt: it seems that the number of elements in the `blockHashCache` and `results` is known so your can size them on creation to avoid resizing. 
Use `HashMap::newHashMap` for maps.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
```diff
@@ -252,39 +314,40 @@ protected BlockCallSimulationResult processTransactions(
...
-    List<TransactionReceipt> receipts = blockCallSimulationResult.getReceipts();
+    List<Transaction> transactions = blockStateCallSimulationResult.getTransactions();
+    List<TransactionReceipt> receipts = blockStateCallSimulationResult.getReceipts();
 
     BlockHeader finalBlockHeader =
         BlockHeaderBuilder.createDefault()
             .populateFrom(blockHeader)
             .ommersHash(BodyValidation.ommersHash(List.of()))
-            .stateRoot(blockOverrides.getStateRoot().orElseGet(ws::rootHash))
+            .stateRoot(blockOverrides.getStateRoot().orElseGet(ws::frontierRootHash))
```
fab-10, line 336: why `frontierRootHash`?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -80,6 +80,10 @@ public class TransactionSimulator {
               SIGNATURE_ALGORITHM.get().getHalfCurveOrder(),
               (byte) 0);
 
+  // Default signature supplier
+  private static final Supplier<SECPSignature> DEFAULT_SIGNATURE_SUPPLIER =
+      Suppliers.memoize(() -> FAKE_SIGNATURE);
```
fab-10, line 85: Could be simplified since the returned value is static

```suggestion
() -> FAKE_SIGNATURE;
```

fab-10 (approved):
LGTM, some minor optimizations, the open question about the behavior of the `pending` block tag, and what about the remaining failing Hive tests.

Gabriel-Trintinalia:
@fab-10 Thanks for the review. I have addressed the suggestions. Regarding the pending block, `eth_simulateV1` will simulate a full block. So when the user chooses "latest", the block simulator will create a pending block whose parent is the latest. So this means that running the simulation on the pending tag would actually skip one block number. If this is required, it can be achieved by overriding the block number using the `blockOverrides` to the latest + 2. The outcome will be two blocks, one which is an empty block, and another with the actual simulation with the empty block as a parent.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
Gabriel-Trintinalia: @fab-10 Since `eth_simulationV1` can simulate up to 256 blocks in the future, I need to retain the changes from each block in the accumulator to build the next block, while also needing the state root to construct the block header. The Frontier Root Hash does not clear the accumulator as the rootHash method does. One option could be to rename it to "intermediateRootHash." Perhaps @matkt has a better suggestion on how I could achieve this?

fab-10:
> @fab-10 Thanks for the review. I have addressed the suggestions. Regarding the pending block, `eth_simulateV1` will simulate a full block. So when the user chooses "latest", the block simulator will create a pending block whose parent is the latest. So this means that running the simulation on the pending tag would actually skip one block number. If this is required, it can be achieved by overriding the block number using the `blockOverrides` to the latest + 2. The outcome will be two blocks, one which is an empty block, and another with the actual simulation with the empty block as a parent.

Fine to ship for me

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/BlockSimulator.java
matkt: why you need to keep what you have in the accumulator ? you just have to persist in your worldstate for each block and start with empty accumulator for each block . no ?

Issue #7614, opened 2024-09-13
Besu increases `gasLimit` when no `--target-gas-limit` is specified


phearnot:
I'm attempting to limit gas usage in a private network, but Besu seems not to honor `gasLimit` from genesis config, and instead increases gas limit up to the value configured for sepolia (`0x1c9c380`). However, the [docs](https://besu.hyperledger.org/public-networks/reference/cli/options#target-gas-limit) for the `--target-gas-limit` command line option state that
> If a value for target-gas-limit is not specified, the block gas limit remains at the value specified in the [genesis file](https://besu.hyperledger.org/public-networks/reference/genesis-items#genesis-block-parameters).

### Steps to reproduce
I'm using the following genesis.json:
```json
{
  "alloc": {},
  "baseFeePerGas": "0x3b9aca00",
  "blobGasUsed": null,
  "coinbase": "0x0000000000000000000000000000000000000000",
  "config": {
    "arrowGlacierBlock": 0,
    "berlinBlock": 0,
    "byzantiumBlock": 0,
    "cancunTime": 0,
    "chainId": 88817,
    "constantinopleBlock": 0,
    "daoForkBlock": 0,
    "eip150Block": 0,
    "eip155Block": 0,
    "eip158Block": 0,
    "ethash": {},
    "grayGlacierBlock": 0,
    "homesteadBlock": 0,
    "istanbulBlock": 0,
    "londonBlock": 0,
    "muirGlacierBlock": 0,
    "petersburgBlock": 0,
    "shanghaiTime": 0,
    "terminalTotalDifficulty": 0,
    "terminalTotalDifficultyPassed": true
  },
  "difficulty": "0x0",
  "excessBlobGas": null,
  "extraData": "0x",
  "gasLimit": "0x1000000",
  "gasUsed": "0x0",
  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "nonce": "0x0",
  "number": "0x0",
  "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "timestamp": "0x660e5e00"
}
```
I'm starting the docker container with besu 24.8.0:
```
docker run -v ./genesis-testnet.json:/etc/genesis.json --rm -p 8551:8551 hyperledger/besu:latest --genesis-file=/etc/genesis.json --engine-rpc-enabled --engine-jwt-disabled --engine-host-allowlist='*' --target-gas-limit=16777216
```
Then I attempt to manually forge the new block:
```
[...]

PR #8409, opened 2025-03-12
Update gas limit logic and cleanup configurations

marcosio:
## PR description
This PR introduces the `gasLimit` field to the Genesis configuration and updates the related CLI logic to handle it. It ensures the gas limit can be defined and adequately propagated from the configuration files to the mining parameters.

## Fixed Issue(s)
Fixes https://github.com/hyperledger/besu/issues/7614


marcosio:
I have different problems with closing it finally:

1. When I run the tests WeiTest.toHumaReadableString and WeiTest.toHumanReadablePaddedString fail due to the "," and "." differences. 
2. In the acceptance tests there are 6 tests with failure.
3. Also, the reference tests are failing.

image
image

siladu (changes requested):
I have updated the issue https://github.com/hyperledger/besu/issues/7614#issuecomment-2735292556 because gasLimit actually works as expected for private networks that use Proof of Authority such as QBFT.

Rather than a bug, I think this is a feature request for Besu to support private networks via the Engine API, in this case to make targetgaslimit more configurable.

The defaulting logic causing the issue is here https://github.com/hyperledger/besu/blob/c99bdbd533707a45fad97017fb964578c3e87fde/consensus/merge/src/main/java/org/hyperledger/besu/consensus/merge/blockcreation/MergeCoordinator.java#L147-L149 
so I think that is the code that needs to change.

The better way to detect a private network is to see if the `--genesisFile` option is set. Then we can say "if privateNetwork then don't set default gas limit".
This may have some impact on mainnet-like devnets that we need to consider.

I don't think you need to read the gasLimit from the genesis file, just avoid this override.

By the way, `gasLimit` is already part of the BlockHeader. If for some reason you do need to read it, then you should be retrieving it via the genesis block, rather than introducing a duplicate config field.

PR #8417, opened 2025-03-13
eth_config prototype

shemnon:
## PR description

A prototype for eth_cofig, sharing operational client configuration data
 via RPC.


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
```diff
@@ -0,0 +1,119 @@
...
+    var contracts =
+        new TreeMap<>(
+            spec.getRequestProcessorCoordinator()
+                .map(RequestProcessorCoordinator::getContractConfigs)
+                .orElse(Map.of()));
+    spec.getBlockHashProcessor()
+        .getHistoryContract()
+        .ifPresent(a -> contracts.put("HISTORY", a.toHexString()));
+    if (!contracts.isEmpty()) {
+      var jsonContracts = result.putObject("contracts");
```
siladu, line 97:
```suggestion
      var jsonContracts = result.putObject("systemContracts");
```
to distinguish from other pre-deployed contracts that might be in the genesis

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
```diff
@@ -0,0 +1,119 @@
...
+
+    var contracts =
+        new TreeMap<>(
+            spec.getRequestProcessorCoordinator()
+                .map(RequestProcessorCoordinator::getContractConfigs)
+                .orElse(Map.of()));
+    spec.getBlockHashProcessor()
+        .getHistoryContract()
+        .ifPresent(a -> contracts.put("HISTORY", a.toHexString()));
+    if (!contracts.isEmpty()) {
```
siladu, line 96: Do we need something for the BEACON_ROOTS_ADDRESS too? https://github.com/ethereum/EIPs/blob/2fb744367e5a65ff3afaf30025cba27ec36b1b3f/EIPS/eip-4788.md?plain=1#L34

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
```diff
@@ -0,0 +1,119 @@
...
+        "chainId", protocolSchedule.getChainId().map(c -> "0x" + c.toString(16)).orElse(null));
+
+    var contracts =
+        new TreeMap<>(
+            spec.getRequestProcessorCoordinator()
+                .map(RequestProcessorCoordinator::getContractConfigs)
+                .orElse(Map.of()));
+    spec.getBlockHashProcessor()
+        .getHistoryContract()
+        .ifPresent(a -> contracts.put("HISTORY", a.toHexString()));
```
siladu, line 95: would be nice to match the EIP name IMO i.e. HISTORY_STORAGE_ADDRESS (and same for the other contracts)

Think precision is good here as perhaps there will be future contracts that require similar names, e.g. the confusion around "withdrawals" vs "withdrawal requests".

siladu (commented):
Adding this as context for reviewers https://hackmd.io/@shemnon/eth_config

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
shemnon: Do you have a naming proposal?  `2935_HISTORY` perhaps?

How about precompiles, should we add the EIP number to the name?  What about modexp, which has had multiple tweaks to the gas schedule, would the eip be the most recent one?

ethereum/api/src/test/resources/org/hyperledger/besu/ethereum/api/jsonrpc/eth/eth_config.json
```diff
@@ -0,0 +1,39 @@
...
+    "jsonrpc": "2.0",
+    "method": "eth_config",
+    "params": []
+  },
+  "response": {
+    "jsonrpc": "2.0",
+    "id": 8,
+    "result": {
+      "current": {
+        "activation": 1444660040,
```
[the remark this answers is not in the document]
shemnon: Two things, (a) the EIP limits itself to post-merge networks and (b) block numbers for mainnet and tesnets will always be less than the point in 2015 when ethereum launched. 

Also, unless we are going to see it on mainnet or the testnets I don't think we should put it in EIPs.  Rollups with shorter block times that present an issue for this can post a RRC.  So it can be moved to activationTime, but I wouldn't specify anything for blocks.

L2s that adopt it and have block based activatons could update the eth_config generation file and add it.

evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
```diff
@@ -58,7 +58,7 @@ public class BigIntegerModularExponentiationPrecompiledContract
    * @param gasCalculator the gas calculator
    */
   public BigIntegerModularExponentiationPrecompiledContract(final GasCalculator gasCalculator) {
-    super("BigIntModExp", gasCalculator);
+    super("EXPMOD", gasCalculator);
```
ahamlat, line 61: I wonder why you didn't choose MODEXP, which is the more widely used name.

shemnon: Will fix.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
siladu: > This would be an example of a true constant being reported, as the beacon root is not configurable in Besu 

Isn't this already the case with the precompiles?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
siladu: I think it should match the name given in the EIP, i.e HISTORY_STORAGE_ADDRESS

I wouldn't object to including the EIP number too, although it is already encoded in the vanity address. However, my comment is less concerned about linking it to an EIP, rather than making it precise enough that a future system contract name is unlikely to collide.

Not sure we should include EIP number to the precompiles but maybe the EIP it was introduced in would be more stable? 
Instead, perhaps more precise naming is warranted there...how likely is it that a future precompile collides? e.g. a new version of modexp alongside the old?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
shemnon: HISTORY_STORAGE_ADDRESS is a variable inside the besu implementations, and had multiple values in devnets.  When I called BEACON_ROOTS_ADDRESS a constant I was referring to Besu code.

siladu:
Reopening as think we should get this in

shemnon:
I'd feel better if at least one other client had implemented it.  But sure, go ahead.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  void generateConfig(final ObjectNode result, final Hardfork forkId, final ProtocolSpec spec) {
+    result.put("activationTime", forkId.milestone());
+
+    ObjectNode blobs = result.putObject("blobSchedule");
+    blobs.put(
+        "baseFeeUpdateFraction", spec.getFeeMarket().getBaseFeeUpdateFraction().longValueExact());
+    blobs.put("max", spec.getGasLimitCalculator().currentBlobGasLimit() / (128 * 1024));
+    blobs.put("target", spec.getGasCalculator().getBlobTarget());
```
macfarla: TODO maxBlobsPerTx added in osaka

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthConfig.java
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright ConsenSys AG.
```
macfarla:
```suggestion
 * Copyright contributors to Besu.
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/DefaultProtocolSchedule.java
```diff
@@ -75,6 +75,25 @@ public ProtocolSpec getByBlockHeader(final ProcessableBlockHeader blockHeader) {
     return null;
   }
 
+  @Override
+  public Optional<ScheduledProtocolSpec> getNextProtocolSpec(final long currentTime) {
```
siladu, line 79: Think we should have a test for this, but can be follow on PR.

siladu (approved):
Wonder if we should add fork names into their as well, for convenience.

PR #8431, opened 2025-03-17
feat: reorder sstore exception that doesn't require storage access

letypequividelespoubelles:

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/operation/SStoreOperation.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/operation/SStoreOperation.java
@@ -72,6 +72,16 @@ public class SStoreOperation extends AbstractOperation {
       return ILLEGAL_STATE_CHANGE;
     }
 
+    final long remainingGas = frame.getRemainingGas();
+
+    if (frame.isStatic()) {
+      return new OperationResult(remainingGas, ExceptionalHaltReason.ILLEGAL_STATE_CHANGE);
+    }
+
+    if (remainingGas <= minimumGasRemaining) {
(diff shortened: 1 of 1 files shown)


matkt (approved):
This one seems to be good. Any concerns @shemnon ?

I also think that this could improve performance @ahamlat  by avoiding some uselesss

> final Supplier<UInt256> currentValueSupplier =
        Suppliers.memoize(() -> account.getStorageValue(key));
    final Supplier<UInt256> originalValueSupplier =
Suppliers.memoize(() -> account.getOriginalStorageValue(key));

OlivierBBB (commented):
LGTM

ahamlat (approved):
Good catch @letypequividelespoubelles.

ahamlat:
As the order of the checks is different with this PR, this still works because both checks 
 ```
 if (remainingGas <= minimumGasRemaining) {
      return new OperationResult(minimumGasRemaining, ExceptionalHaltReason.INSUFFICIENT_GAS);
    }
```
```
if (remainingGas < cost) {
      return new OperationResult(cost, ExceptionalHaltReason.INSUFFICIENT_GAS);
    }
```
return the same halt reason : **ExceptionalHaltReason.INSUFFICIENT_GAS**, I wonder if this can have an impact on tracing as the first parameter of OperationResult can be different with the previous implementation.

OlivierBBB:
@ahamlat we detect these exceptions independently (and prior to the production of the `OperationResult`)

matkt:
I checked other implementation on différent clients and this seems to be the correct one. So I would say that if we have a problem in the tracing the error is already there.

Issue #7695, opened 2024-10-01
Remove old eth protocols


jframe:
Remove old eth protocols older than eth/65. Modern clients don't support these protocols so we are wasting our time connecting to these peers that support these protocols. There is also an opportunity to remove some legacy code by not supporting older protocols e.g. requests without request Id.

PR #8453, opened 2025-03-24
refactor: merge Eth protocol messages into single class

Gabriel-Trintinalia:
This pull request focuses on refactoring the Ethereum protocol message handling in the `ethereum/eth` module. The primary change involves consolidating message constants into a single class, `EthProtocolMessages`, to simplify and centralise the message definitions. Replaced individual message imports (`EthPV62`, `EthPV63`, `EthPV65`) with `EthProtocolMessages` and updated all references accordingly. 

https://github.com/hyperledger/besu/issues/7695


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/EthProtocolMessages.java
```diff
@@ -14,7 +14,7 @@
  */
 package org.hyperledger.besu.ethereum.eth.messages;
 
-public final class EthPV62 {
+public final class EthProtocolMessages {
```
Matilda-Clerke, line 17: Would it make sense for this to be an enum rather than a class of constants?

Matilda-Clerke (approved):
LGTM. Not sure I like that there's no real connection between protocol version and message type, but it seems like the code is set up agnostic to that connection anyway.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/EthProtocolMessages.java
Gabriel-Trintinalia: yes, definitely! I considered doing this in this PR but decided to do it later

Gabriel-Trintinalia:
> Not sure I like that there's no real connection between protocol version and message type

Protocol versions can add, remove, or modify messages, but these messages are not exclusive to any particular version. For instance, the GetReceipts and Receipts messages were introduced in `eth63`, yet they continue to be included in all subsequent protocol versions, even if `eth63` is completely removed from the codebase.

PR #8458, opened 2025-03-25
test: update tests to the latest protocol version

Gabriel-Trintinalia:
## PR description

This pull request updates the test cases and code to use the latest Ethereum protocol version. It replaces older versions (`ETH62`,`ETH63`) with a new constant `LATEST`.

These changes keep the code and tests up to date, improve maintainability, and allow the removal of old protocol versions.

More info:  #7695


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/EthProtocol.java
```diff
@@ -125,4 +125,8 @@ public static EthProtocol get() {
   public static boolean isEth66Compatible(final Capability capability) {
     return NAME.equals(capability.getName()) && capability.getVersion() >= ETH66.getVersion();
   }
+
+  public static Capability getLatestVersion() {
+    return ETH68;
```
usmansaleem, line 130: nit: we can also create a `LATEST` static final variable towards the top of the class and then return that variable here. For example:
```
public static final Capability ETH67 = Capability.create(NAME, EthProtocolVersion.V67);
  public static final Capability ETH68 = Capability.create(NAME, EthProtocolVersion.V68);
  public static final Capability LATEST = ETH68;
```

usmansaleem (approved):
LGTM

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/RespondingEthPeer.java
```diff
@@ -217,12 +218,21 @@ public boolean respond(final Responder responder) {
   }
 
   private void respondToMessage(final Responder responder, final OutgoingMessage msg) {
-    final Optional<MessageData> maybeResponse = responder.respond(msg.capability, msg.messageData);
+    MessageData messageData = msg.messageData;
```
pinges, line 221: I wouldn't mind an if / else here to make it clearer

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/RespondingEthPeer.java
```diff
@@ -217,12 +218,21 @@ public boolean respond(final Responder responder) {
...
+    if (EthProtocol.isEth66Compatible(msg.capability)) {
+      messageData = msg.messageData.unwrapMessageData().getValue();
+    }
+    final Optional<MessageData> maybeResponse = responder.respond(msg.capability, messageData);
     maybeResponse.ifPresent(
         (response) -> {
           if (ethProtocolManager.getSupportedCapabilities().contains(msg.capability)) {
+            var unWrappedResponse = response;
```
pinges, line 229: if / else would make it more readable for me

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/EthProtocolHelper.java
```diff
@@ -0,0 +1,21 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.p2p;
+
+import org.hyperledger.besu.ethereum.p2p.rlpx.wire.Capability;
+
+public class EthProtocolHelper {
```
pinges, line 19: Do we need this class? EthProtocol.LATEST is exactly the same, isn't it?

pinges (commented):
I like how we are now mainly testing against the LATEST protocol version, but does that mean that we are no longer testing the older versions? Should we add tests for the older versions?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/RespondingEthPeer.java
Gabriel-Trintinalia: done, this code will be removed soon anyway :)

ethereum/p2p/src/test/java/org/hyperledger/besu/ethereum/p2p/EthProtocolHelper.java
Gabriel-Trintinalia: done

Gabriel-Trintinalia: yes, I don't want to add a dependency here.

Gabriel-Trintinalia:
@pinges 
> does that mean that we are no longer testing the older versions? 

Correct. Specific changes to old protocols are still being tested by their tests. Everything else was using `Eth63`

> Should we add tests for the older versions?

No, the goal of this PR is to facilitate the removal of the old protocols

PR #8470, opened 2025-03-26
Fastsync removal tests - non breaking changes 

pullurib:
## PR description
- Changes to remove and update tests as part of fastsync feature removal changes 
- These are non breaking changes only - ones that are not dependent on refactored classes , etc.


besu/src/test/resources/complete_config.toml
```diff
@@ -29,7 +29,7 @@ metrics-port=309
 # chain
 genesis-file="/opt/besu/genesis.json" # Path
 network-id=42
-sync-mode="fast"# should be FAST or FULL (or fast or full)
+sync-mode="full"
 fast-sync-min-peers=13
```
macfarla, line 33:
```suggestion
sync-min-peers=13
```

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastDownloaderFactoryTest.java
macfarla, line 1: there's no test for SnapDownloaderFactory which extends FastDownloaderFactory - can this test be adapted instead of deleted

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/worldstate/RequestDataStepTest.java
macfarla, line 1: same here - there's no test for snap/RequestDataStep, can this be adapted instead of deleted

besu/src/test/java/org/hyperledger/besu/chainimport/RlpBlockImporterTest.java
```diff
@@ -69,7 +69,7 @@ public void blockImport() throws IOException {
     final BesuController targetController =
         new BesuController.Builder()
             .fromEthNetworkConfig(
-                EthNetworkConfig.getNetworkConfig(NetworkName.MAINNET), SyncMode.FAST)
+                EthNetworkConfig.getNetworkConfig(NetworkName.MAINNET), SyncMode.SNAP)
```
macfarla, line 72: maybe the RLPBlock importer should be full sync default?

macfarla (commented):
some of the tests should be adapted from FastSync classes to SnapSync where there is no current test, rather than deleted. I didn't check all of them.

macfarla (changes requested):
some of the tests should be adapted from FastSync classes to SnapSync where there is no current test, rather than deleted. I didn't check all of them.

pullurib:
@macfarla Sorry lost track of this . I spent time looking into it but adapting fastsync to snapsync is not a trivial change to include in this PR . I can add a new issue to track that work if that helps and can come back later to it .

macfarla:
> @macfarla Sorry lost track of this . I spent time looking into it but adapting fastsync to snapsync is not a trivial change to include in this PR . I can add a new issue to track that work if that helps and can come back later to it .

ok yep sounds good. let's make sure we don't reduce our test coverage.

PR #8477, opened 2025-03-27
refactor: introduce TransactionReceiptEncodingOptions to TransactionReceiptEncoder

Gabriel-Trintinalia:
The changes mainly involve refactoring the `TransactionReceiptEncoder` to use a new `TransactionReceiptEncodingOptions` class and updating various files to accommodate this new structure.

### Refactoring and Code Improvements:

* **Refactoring `TransactionReceiptEncoder`**:
  - The `TransactionReceiptEncoder` has been refactored to use a new `TransactionReceiptEncodingOptions` class. This refactoring simplifies the encoding process by providing predefined options and removing the need for multiple methods like `writeToForNetwork` and `writeToForStorage`.

* **Introduction of `TransactionReceiptEncodingOptions`**:
  - A new class, `TransactionReceiptEncodingOptions`, has been introduced to encapsulate various encoding options. This class includes predefined options for different encoding scenarios such as network, storage with compaction, storage without compaction, and trie root.


jframe (approved):
LGTM - nice improvement

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptEncodingOptions.java
```diff
@@ -0,0 +1,101 @@
...
+package org.hyperledger.besu.ethereum.core.encoding.receipt;
+
+public class TransactionReceiptEncodingOptions {
+  public static final TransactionReceiptEncodingOptions NETWORK = new Builder().build();
+
+  public static final TransactionReceiptEncodingOptions STORAGE_WITH_COMPACTION =
+      new Builder().withRevertReason(true).withCompactedLogs(true).withBloomFilter(false).build();
+
+  public static final TransactionReceiptEncodingOptions STORAGE_WITHOUT_COMPACTION =
+      new Builder().withRevertReason(true).build();
```
Matilda-Clerke, line 24: Would be nice to `withCompactedLogs(false)` here, in case the default every changes and it no longer matches the constant name

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptEncodingOptions.java
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.core.encoding.receipt;
+
+public class TransactionReceiptEncodingOptions {
```
Matilda-Clerke, line 17: I'm not a fan of classes ending with a plural... it makes me think it should have just been a list. (Don't get me started on `EthPeers`)

In this case, it's accurate that each instance of this class is a set of options, but imo it would be better called `TransactionReceiptEncodingConfiguration` or something similar.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptEncodingOptions.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public static final TransactionReceiptEncodingOptions TRIE_ROOT =
+      new Builder().withOpaqueBytes(false).build();
+
+  private final boolean withRevertReason;
+  private final boolean withCompactedLogs;
+  private final boolean withOpaqueBytes;
+  private final boolean withBloomFilter;
+
+  private TransactionReceiptEncodingOptions(final Builder builder) {
```
Matilda-Clerke, line 34: Also not a fan of passing a builder into a constructor. Personally, I'd change this to a record and just call the constructor in the builder's `build()` method

Matilda-Clerke (approved):

PR #8480, opened 2025-03-27
Fix a pruning bug and add rocksdb storage tuning options

pullurib:
## PR description
Updated:
- Fix a bug that affected pruning on a running chain. When pruning starts for the first time and no block is yet recorded as the storedPruningMark (the earliest retained block), the current block number was incorrectly used. This prevented any blocks before that point from being pruned.
- add rocksdb options to tune the disk usage limits 

Prev for context:
- Run compaction every time pruning is performed 
- add rocksdb options to tune the disk usage limits


siladu:
@pullurib Hi :) 

> Run compaction every time pruning is performed

I am surprised to see this because the BLOCKCHAIN column family uses BlobDB https://github.com/facebook/rocksdb/wiki/BlobDB
which separates the keys from the values (immutable blocks in this case) in order to reduce the write amplication you get from compaction.

Can you please elaborate on why forcing compaction here is helpful and do you have number to share to show an improvement?

This feature may be used to implement EIP-4444 so we are actively looking into it at the moment (maybe for same reason as you?)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -84,6 +87,12 @@ public void onBlockAdded(final BlockAddedEvent event) {
           }
           prunerStorage.setPruningMark(pruningTransaction, currentPruningMark);
           pruningTransaction.commit();
+          if (pruningRequired) {
+            // After pruning data, compact storage to reclaim space
+            if (blockchainStorage instanceof CompactableStorage compactable) {
+              compactable.compact();
+            }
+          }
```
garyschulte, line 95: Totally agree with Simon here.  Forcing compaction because of periodic chain pruning is likely to have implications for block processing performance and IO generally.

I definitely want to see performance stats on this PR.  Charts (grafana) of IO usage with and without this feature, correlated with block execution time.

garyschulte (changes requested):
This big of a change to the i/o profile must be accompanied with performance stats and impact analysis.

@ahamlat @matkt

pullurib:
> Can you please elaborate on why forcing compaction here is helpful and do you have numbers to share to show an improvement?
> FYI, chain-data-pruning feature may be used to implement EIP-4444 so we are actively looking into it at the moment (maybe for same reason as you?)

Hi @siladu ,  yes, this is intended to assist with block history expiry . Database size keeps on growing even after pruning as compaction is needed to clean up deleted keys . Besu restart does compaction and I see that reduction in disk usage then . So idea is to see this change in disk usage without a restart. I saw difference at a high level using this compaction and tweaking options, will run the chain again with pruning as is now, pruning with compaction and pruning with compaction/tweaking options and share results by next week.

> I am no RocksDB expert, but my understanding is that when you delete a BlobsDB entry, it is garbage collected (which is configured for BlobDB specifically). I would expect this would be the thing to tune rather forcing than a compaction.

Right, could be my transactions not resulting in many blob entries due to value size threshold. I'll dig deeper to profile and get numbers .

garyschulte:
>Right, could be my transactions not resulting in many blob entries due to value size threshold. I'll dig deeper to profile and get numbers

Testing on a mainnet ethereum node would be 👍

siladu:
> Hi @siladu , yes, this is intended to assist with block history expiry . 

That good, we are working towards the same goal :)
Perhaps a slight divergence is that a near-term goal for Ethereum mainnet is only to delete a static list of "pre-merge" blocks, not a rolling window. It is related to this EIP: https://eips.ethereum.org/EIPS/eip-7639
EIP-4444 with a rolling window is the eventual goal but may be some months before we implement for mainnet.
That could mean we need ChainDataPruner to work for both cases. Note for EIP-7639, ChainDataPruner may be used to deleted the fixed set of blocks (but retain the headers) so it is only run for a fixed time depending on the state of your node.

> Database size keeps on growing even after pruning as compaction is needed to clean up deleted keys. 

Would be good to understand how BlobDB is working for your network - if it is working, the deleted values should be garbage collected after a period. However, there are still associated SST files that store the pointers to the blob files so perhaps that benefits from compaction, but presumably much less so than if we were using vanilla SST files.
Note, if you are using BONSAI, then each block appended also appends to TRIE_LOG_STORAGE column family. There already exists TrieLogPruner for this which is disabled if you use FULL sync, as discussed with @matthew1001 a while ago.

> Besu restart does compaction and I see that reduction in disk usage then. 

It would be good to understand in detail what exactly the reduction is. A tool you can use for this if you don't already know will show the column family breakdown by SST versus blob file: https://besu.hyperledger.org/public-networks/reference/cli/subcommands#rocksdb-usage

A recent example from mainnet (with CHECKPOINT sync so ~11 million blocks and TRIE_LOG_STORAGE pruning enabled) shows that BLOCKCHAIN column family has 137 GiB of SST storage and 717 GiB of blob file storage
```
|--------------------------------|-----------------|-------------|-----------------|------------------|
| Column Family                  | Keys            | Total Size  | SST Files Size  | Blob Files Size  |
|--------------------------------|-----------------|-------------|-----------------|------------------|
| BLOCKCHAIN                     | 1934724122      | 855 GiB     | 137 GiB         | 717 GiB          |
| VARIABLES                      | 2175            | 35 KiB      | 35 KiB          | 0 B              |
| ACCOUNT_INFO_STATE             | 288572029       | 13 GiB      | 13 GiB          | 0 B              |
| ACCOUNT_STORAGE_STORAGE        | 1279174442      | 56 GiB      | 56 GiB          | 0 B              |
| CODE_STORAGE                   | 1625744         | 5 GiB       | 5 GiB           | 0 B              |
| TRIE_BRANCH_STORAGE            | 2171203227      | 157 GiB     | 157 GiB         | 0 B              |
| TRIE_LOG_STORAGE               | 513             | 139 MiB     | 114 KiB         | 139 MiB          |
|--------------------------------|-----------------|-------------|-----------------|------------------|
| ESTIMATED TOTAL                | 5675302252      | 1088 GiB    | 370 GiB         | 718 GiB          |
|--------------------------------|-----------------|-------------|-----------------|------------------|
```

> I saw difference at a high level using this compaction and tweaking options

Would recommend also looking into BlobDB garbage collection settings.
You can get more info using `besu --config-file config.toml storage rocksdb x-stats`


> Right, could be my transactions not resulting in many blob entries due to value size threshold. I'll dig deeper to profile and get numbers .

Yes, that's interesting. We could look into tweaking that threshold. If you want to perform testing with lots of transactions then https://github.com/MariusVanDerWijden/tx-fuzz is a handy tool for that.
It should work even with a paid-gas network as long as you allocate some ETH to `0xb02A2EdA1b317FBd16760128836B0Ac59B560e9D` address, e.g. https://github.com/siladu/beku-timestamp/commit/21421cd93b4a265bf096de51cb59775e116ccfc5

PR #8488, opened 2025-03-28
Cache number of zero bytes in the payload in the tx

daniellehrner:
## PR description

We need to count the zero bytes of the transaction payload to calculate the intrinsic gas cost. Right now we do this up to 3 times:

1) When we receive a tx and add it to the tx pool
2) At the start of the transaction processing when we validate a tx
3) When we calculate and deduct the intrinsic gas cost during tx processing

This PR caches the number of zero bytes in the tx object, which avoids recounting them and should be more efficient.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Payload.java
```diff
@@ -0,0 +1,73 @@
...
+public class Payload {
+  private final Bytes payloadBytes;
+
+  private Long zeroBytes = null;
+
+  public Payload(final Bytes payloadBytes) {
+    this.payloadBytes = payloadBytes;
+  }
+
+  public Bytes getPayload() {
```
garyschulte, line 30: this would make more sense as getPayloadBytes.  payload.getPayload() is a bit odd.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Payload.java
```diff
@@ -0,0 +1,73 @@
...
+    return payloadBytes;
+  }
+
+  public long zeroBytes() {
```
garyschulte, line 34: sorry another naming thing, suggest getZeroBytesCount()

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Payload.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private long computeZeroBytes() {
+    int zeros = 0;
+    for (int i = 0; i < payloadBytes.size(); i++) {
+      if (payloadBytes.get(i) == 0) {
+        zeros += 1;
+      }
+    }
+    return zeros;
```
garyschulte, line 71: I wonder if/how we can determine this is using simd under-the-hood.  Non-blocking just curious.  @ahamlat ?

garyschulte (commented):
LGTM, just some comments and suggestions

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Payload.java
daniellehrner: Makes sense, I've renamed it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Payload.java
daniellehrner: Makes sense, I've renamed it

daniellehrner:
@garyschulte Do you think it's ready to merged with the latest changes? If yes, could you please approve?

garyschulte (approved):
:+1:

PR #8492, opened 2025-03-31
feat!: remove support for eth/62, eth/63, eth/64 and eth/65

Gabriel-Trintinalia:
## Description

fixes https://github.com/hyperledger/besu/issues/7695

This pull request removes support for the older Ethereum protocol versions `eth/62`, `eth/63`, `eth/64`, and `eth/65`. 

### Changes:

- Remove `eth/62` and `eth/63` Protocol Versions:
  - Eliminate all references and implementations related to eth/62 and eth/63.

- Remove `eth/64` Protocol Version:
  - Remove all references to `legacyForkId`.
  - Deprecate the `--compatibility-eth64-forkid-enabled` flag.
  - Make `ForkId` a required component for constructing a `StatusMessage`.
  - Ensure that decoding and encoding of `ForkId` are always present.
- Remove `eth/65` Protocol Version:
  - Remove the `isEth66Compatible` check, as all newer protocol versions are inherently compatible.
  - Eliminate all checks related to `isEth66Compatible`.


fab-10 (approved):
Nice cleanup!

Gabriel-Trintinalia:
@fab-10  thanks! There may be some clean-up to be done in the `TransactionBroadcaster` as well when this PR is merged since we don't need to announce full transactions anymore

besu/src/main/java/org/hyperledger/besu/cli/options/EthProtocolOptions.java
```diff
@@ -90,13 +88,6 @@ public class EthProtocolOptions implements CLIOptions<EthProtocolConfiguration>
   private PositiveNumber maxGetPooledTransactions =
       PositiveNumber.fromInt(EthProtocolConfiguration.DEFAULT_MAX_GET_POOLED_TRANSACTIONS);
 
-  @CommandLine.Option(
-      names = {LEGACY_ETH_64_FORK_ID_ENABLED},
-      paramLabel = "<Boolean>",
-      description = "Enable the legacy Eth/64 fork id. (default: ${DEFAULT-VALUE})")
-  private Boolean legacyEth64ForkIdEnabled =
```
macfarla, line 97: maybe it's worth keeping just the CLI option as a no-op for one release so we don't break existing configs? Make it a hidden option, flag it as a upcoming breaking change. The rest of the PR is non-breaking?

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/RespondingEthPeer.java
```diff
@@ -218,9 +218,7 @@ public boolean respond(final Responder responder) {
   }
 
   private void respondToMessage(final Responder responder, final OutgoingMessage msg) {
-    boolean supportsRequestId =
-        EthProtocol.isEth66Compatible(msg.capability)
-            && EthProtocol.requestIdCompatible(msg.messageData.getCode());
+    boolean supportsRequestId = EthProtocol.requestIdCompatible(msg.messageData.getCode());
```
pinges, line 221: I thought that all our supported protocols now are request id compatible. Do we still need that method/check?

pinges (commented):
A few comments :-)

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/RespondingEthPeer.java
Gabriel-Trintinalia: The requestIdCompatible is checking if the message itself requires the requestId. (Some messages don't, such as Status, Transactions)

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -178,13 +174,6 @@ private List<Capability> calculateCapabilities(
       final SynchronizerConfiguration synchronizerConfiguration,
       final EthProtocolConfiguration ethProtocolConfiguration) {
     final List<Capability> capabilities = new ArrayList<>();
-
-    if (SyncMode.isFullSync(synchronizerConfiguration.getSyncMode())) {
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: It is not needed. Probably left there to talk to really old archive nodes.

besu/src/main/java/org/hyperledger/besu/cli/options/EthProtocolOptions.java
Gabriel-Trintinalia: done

macfarla (approved):

PR #8493, opened 2025-03-31
Online option to drop pre merge blocks

Matilda-Clerke:
## PR description
Add an online option to prune pre-merge blocks from the database

## Issue
https://github.com/hyperledger/besu/issues/8337


besu/src/main/java/org/hyperledger/besu/services/PreMergeOnlineBlockDataPruner.java
```diff
@@ -0,0 +1,117 @@
...
+  public void run() {
+    if (!SUPPORTED_NETWORKS.contains(networkName)) {
+      LOG.error(
+          "Unable to prune pre-merge blocks and transaction receipts for network: {}", networkName);
+      return;
+    }
+    final long mergeBlockNumber = getMergeBlockNumber(networkName);
+
+    LOG.info("Pruning pre-merge blocks and transaction receipts");
+    for (long startBlockNumber = 0;
```
siladu, line 69: Should we used "earliestBlock" here (which I think eth/69 has the concept of)?
Reason being, what if either we are already pruned, using CHECKPOINT, or if the user restarts before pruning finishes?

besu/src/main/java/org/hyperledger/besu/services/PreMergeOnlineBlockDataPruner.java
```diff
@@ -0,0 +1,117 @@
...
+                });
+      } while (++headerNumber < endBlockNumber);
+      updater.commit();
+
+      if (endBlockNumber % preMergeBlockPrunerRange * 1000 == 0) {
+        LOG.info("Pruned up to block {}", endBlockNumber);
+      }
+
+      try {
+        Thread.sleep(preMergeBlockPrunerSleepTime);
```
siladu, line 102: Rather than hog a thread for the whole application lifetime, we could trigger via a onBlockAdded callback and make the processing async? This pattern already exists for several things, e.g. TrieLogPruner, TransactionPool

besu/src/main/java/org/hyperledger/besu/services/PreMergeOnlineBlockDataPruner.java
```diff
@@ -0,0 +1,117 @@
...
+      } catch (InterruptedException e) {
+        return;
```
siladu, line 104: Would this result in the Pruner thread dying? 
Think we should log an error here at least and/or possibility to recover?

besu/src/main/java/org/hyperledger/besu/services/PreMergeOnlineBlockDataPruner.java
```diff
@@ -0,0 +1,117 @@
...
+        startBlockNumber < mergeBlockNumber;
+        startBlockNumber += preMergeBlockPrunerRange) {
+      BlockchainStorage.Updater updater = blockchainStorage.updater();
+      long headerNumber = startBlockNumber;
+      final long endBlockNumber =
+          Math.min(startBlockNumber + preMergeBlockPrunerRange, mergeBlockNumber);
+      do {
+        blockchainStorage
+            .getBlockHash(headerNumber)
+            .filter((h) -> blockchainStorage.getBlockBody(h).isPresent())
```
siladu, line 79: Similar to the comment above, I think in the "already-pruned" case, we're going to waste two calls per block here, in the worst case (completely pruned) we'll needlessly call this 15_537_393 times.

getBlockBody in particular is expensive.

besu/src/main/java/org/hyperledger/besu/services/PreMergeOnlineBlockDataPruner.java
```diff
@@ -0,0 +1,117 @@
...
+            .ifPresent(
+                (h) -> {
+                  updater.removeTransactionReceipts(h);
+                  updater.removeTotalDifficulty(h);
+                  blockchainStorage
+                      .getBlockBody(h)
```
siladu, line 85: Even in the case where we should prune, this is a duplicated expensive call, would be better to cache this in a local variable.

siladu (commented):
Implementation looks nice a straight forward 👍 

Few comments about edge cases.

Can you share test results please?

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -805,6 +825,19 @@ public BesuController build() {
     if (privacyParameters.getPrivateStorageProvider() != null) {
       closeables.add(privacyParameters.getPrivateStorageProvider());
     }
+    if (preMergeBlockPruningConfiguration.preMergeBlockPruningEnabled()) {
```
siladu, line 828: Would be good to detect if pruning is already completed and skip setting this up

besu/src/main/java/org/hyperledger/besu/services/PreMergeOnlineBlockDataPruner.java
Gabriel-Trintinalia: yes, pruning will have to update the earliest block. I am working on it with eth/69

besu/src/main/java/org/hyperledger/besu/services/PreMergeOnlineBlockDataPruner.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+    LOG.info("Pre-merge block pruning complete");
+  }
+
+  private static long getMergeBlockNumber(final NetworkName network) {
+    return switch (network) {
+      case MAINNET -> MAINNET_MERGE_BLOCK_NUMBER;
+      case SEPOLIA -> SEPOLIA_MERGE_BLOCK_NUMBER;
+      default -> throw new RuntimeException("Unexpected network: " + network);
+    };
```
Gabriel-Trintinalia, line 115: Maybe this information should be here `NetworkName`

Matilda-Clerke:
Some results from testing (green is 100 blocks per 100ms, yellow is 100 blocks per 1000ms):
image
It doesn't seem to have any affect on engineNewPayload or forkchoice performance

siladu:
> Some results from testing (green is 100 blocks per 100ms, yellow is 100 blocks per 1000ms)

Thanks, are "test-3" and "test-4" controls?

Matilda-Clerke:
Correct, test-3 and test-4 were controls. I've shut these nodes down for now, so I'm not sure how long it takes to prune all of the blocks it needs to do. I've now raised https://github.com/hyperledger/besu/pull/8506 as an alternate approach to address some of the issues raised

ahamlat:
How about triggering the drop of x blocks, x is to be defined, after each block was imported with engine_newPayloadVx ?
This way, we're sure to not have an impact on newPayload execution especially on small hardware.
Also, I think we need to have a metric that shows the remaining pre-merge blocks. We need to know if the pruning is finished or ongoing, especially during performance analysis.

PR #8498, opened 2025-03-31
Implement EIP-5920 - PAY opcode

### LINKED ISSUE #8438: Implement PAY opcode
For the upcoming fork Fusaka there's a strong case for a new PAY opcode that can just transfer funds without calling into a contract.
EIP: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5920.md
Discussion on Eth magicians: https://ethereum-magicians.org/t/eip-5920-pay-opcode/11717/37?u=wjmelements

Presentation of PAY opcode: https://docs.google.com/presentation/d/1pZgbr5yLaSNxpwRdixgkAV9mfo_gql448_iSOVKlyvA/edit?usp=sharing

lu-pinto:
## PR description
Implement new PAY opcode for EOF.

## Fixed Issue(s)

fixes #8438


evm/src/main/java/org/hyperledger/besu/evm/operation/PayOperation.java
```diff
@@ -0,0 +1,102 @@
...
+
+    final Address to = Words.toAddress(toAddressBytes);
+    final Wei value = Wei.wrap(frame.getStackItem(0));
+    final boolean hasValue = value.greaterThan(Wei.ZERO);
+    final Account recipient = frame.getWorldUpdater().get(to);
+
+    final boolean accountIsWarm = frame.warmUpAddress(to);
+
+    if (frame.isStatic() && hasValue && !Objects.equals(frame.getSenderAddress(), to)) {
+      return new OperationResult(cost(to, true, recipient, true), ExceptionalHaltReason.ILLEGAL_STATE_CHANGE);
```
jochem-brouwer, line 53: EIP will need clarification for this, but if you use the CALL opcode and you send value (also to self) in a static context, this is not allowed and will throw.

evm/src/main/java/org/hyperledger/besu/evm/operation/PayOperation.java
```diff
@@ -0,0 +1,102 @@
...
+  @Override
+  public OperationResult execute(final MessageFrame frame, final EVM evm) {
+    Code callingCode = frame.getCode();
+    if (callingCode.getEofVersion() == 0) {
+      return InvalidOperation.INVALID_RESULT;
+    }
+
+    final Bytes toAddressBytes = frame.getStackItem(1);
+    if (toAddressBytes.size() > 20
+      && toAddressBytes.numberOfLeadingZeroBytes() < 12) {
```
jochem-brouwer, line 41: Does the stack trim zeros by default? If not, what if I push 21 bytes to the stack with a leading 0 and an address of 20 nonzero bytes? This throws, but should be valid.

jochem-brouwer: `PUSH0 PUSH21 00<20 nonzero bytes> PAY` maybe?

evm/src/main/java/org/hyperledger/besu/evm/operation/PayOperation.java
```diff
@@ -0,0 +1,102 @@
...
+      return new OperationResult(0, ExceptionalHaltReason.ADDRESS_OUT_OF_RANGE);
+    }
+
+    final Address to = Words.toAddress(toAddressBytes);
+    final Wei value = Wei.wrap(frame.getStackItem(0));
```
jochem-brouwer, line 46: I believe the stack is `[addr, val, ...]` and it seems to be implemented `[val, addr, ...]` here. So `stack[0]` should be `val` and `stack[1]` `addr`.
image

evm/src/main/java/org/hyperledger/besu/evm/operation/PayOperation.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private long cost(final Address to, final boolean hasValue, final Account recipient, final boolean accountIsWarm) {
+    long cost = 0;
+    if (hasValue) {
+      cost = gasCalculator().callValueTransferGasCost();
+    }
+    if (accountIsWarm
+      //TODO: what about precompile accounts?
+      || gasCalculator().isPrecompile(to)) {
```
jochem-brouwer, line 90: Precompiles are added to warm addresses by default  
image

https://eips.ethereum.org/EIPS/eip-2929#parameters

jochem-brouwer (commented):
Saw this on discord, did a quick skim-through of this :)

evm/src/main/java/org/hyperledger/besu/evm/operation/PayOperation.java
lu-pinto: Yes but isn't it what StaticCall actually is? A Call with no value that does not change any state. But EIP should be clarified for sure

lu-pinto: One can argue that static calls don't do logs though so you be correct

evm/src/main/java/org/hyperledger/besu/evm/operation/PayOperation.java
lu-pinto: Ok, thanks good to know. This is just the way Besu does it not sure why but in the end it's the same thing.

lu-pinto:
> Saw this on discord, did a quick skim-through of this :)

Thanks for the review, fixes a few things based on that

evm/src/main/java/org/hyperledger/besu/evm/operation/PayOperation.java
jochem-brouwer: > Yes but isn't it what StaticCall actually is? A Call with no value that does not change any state. But EIP should be clarified for sure

Note that I'm talking about a PAY with value, but the PAY is actually to the current address. So the state does not change. The EIP will need clarification for this. But, note for the CALL opcode that if value is being sent in a STATICCALL context, the opcode will exceptionally abort (so even though no state is changed if it would be executed). 

I will open a PR for this against the EIP.

macfarla (commented):
@lu-pinto what do we have to change to make this independent of EOF?

lu-pinto:
> @lu-pinto what do we have to change to make this independent of EOF?

Just adding it at the right hardfork configuration and aligning the return value with current EVM opcodes (*CALL).
There are also some changes happening to PAY as we speak that haven't been merge which I haven't had the time to check: https://github.com/ethereum/EIPs/pull/9590

jflo (approved):

PR #8463, opened 2025-03-26
minor changes to RPC result formatting

macfarla:
Signed-off-by: Sally MacFarlane <macfarla.github@gmail.com>
## PR description
* sort access list storage keys
* ~return 0x0 for zero key rather than 0x~ removed this change from this PR - needs further discussion.

## Fixed Issue(s)
related to #5377
related to  [rpc-compat](https://hivetests2.ethdevops.io/suite.html?suiteid=1742933806-d091f7bdbe99c2cf163c7f54fef02959.json&suitename=rpc-compat#test-23) hive tests

this fixes 
* ~`eth_getProof/get-account-proof-with-storage`~

and reduces the discrepancies in 
* `eth_createAccessList/create-al-contract-eip1559`
* and `eth_createAccessList/create-al-contract`

### Thanks for sending a pull request! Have you done the following?

- [ ] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [ ] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [ ] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [ ] spotless: `./gradlew spotlessApply`
- [ ] unit tests: `./gradlew build`
- [ ] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

PR #8499, opened 2025-04-01
storage proof key - use 0x0 for zero

macfarla:
## PR description
return 0x0 for zero key rather than 0x

Added a test for visibility

## Fixed Issue(s)
related to https://github.com/hyperledger/besu/issues/5377
related to [rpc-compat](https://hivetests2.ethdevops.io/suite.html?suiteid=1742933806-d091f7bdbe99c2cf163c7f54fef02959.json&suitename=rpc-compat#test-23) hive tests

this fixes 1 hive test:
`eth_getProof/get-account-proof-with-storage`

Refs this comment https://github.com/hyperledger/besu/pull/8463/commits/5cb6c26b206936e20550bee1be81a3ed1dfb9928#r2017642440
This was discussed on RPC standards [call](https://github.com/ethereum/pm/issues/1413) and since odd-length hex strings are widely used we will stick with that. Quantity.create handles the zero edge case already.


macfarla:
Note I had to update several `eth_getProof` json spec tests to make the key consistent with this PR (remove leading zero from non-zero key). Using Quantity.create makes the format consistent with other places in the output - same as for nonce, the eth_getProof value and other places too.

I don't think this would be a breaking change - and we'll be now consistent with geth - but flagging for reviewers.

lu-pinto (approved):
LGTM

macfarla:
@lu-pinto based on your feedback and further chat w Gary, adopted your suggestion to keep the leading zeros where we had them before.

PR #8500, opened 2025-04-01
bonsai: commit trielog ahead of state

### LINKED ISSUE #6569: Bonsai RocksDB trielog inconsistency
When Besu doesn't shutdown gracefully due to a forceful kill sometimes there is an inconsistency between the trielog and the world state. This manifests as a worldstate mismatch error caused by a missing parent world state.

e.g.
```
{"@timestamp":"2024-02-12T04:17:36,399","level":"INFO","thread":"EthScheduler-Services-9 (importBlock)","class":"MainnetBlockValidator","message":"Invalid block 7823908 (0x2f63f1be4e566e609f011dfedc7cfe34f78f5d0a97ab88515042bc99de775cac): Optional[Unable to process block because parent world state 0x211c1dcdb86f605b6dac592e5389a11f5f735cbed0485e13ba0d3a05573915f1 is not available]","throwable":""}
```

A transaction was added previously to ensure that the worldstate was committed consistently. The theory is that this should be extended to also include the committing the trielog as well to ensure consistency with trielogs the worldstate.

**Acceptance Criteria**
- Ensure that both worldstate and the trielog are consistent with each other after hard kill of Besu

garyschulte:
## PR description

This PR addresses a regression where bonsai database can be left in a broken state on an abnormal shutdown.  

When besu commits a bonsai worldstate, it commits the trielog and the worldstate.  if the worldstate is committed first, and besu gets killed before it can commit the trielog, upon resuming operations besu will fail to create or import blocks, giving an error similar to: 
>Optional[Unable to process block because parent world state 0xcdee53accfeaa10680a304bc3ab3b721f15caeb61a4ede8ddde994f788892ced is not available]

The fix is to move the trielog commit ahead of the worldstate commit, and add the methods necessary to isolate the commit of both transactions.

Although this PR is very short, there is a lot of complexity hidden in it.  Several edge cases were discovered and addressed during testing:
* trie log manager inadvertently committing world state 
* composed worldstate commit inadvertently causing exceptions for closed trie log transactions
* block creation failure immediately after genesis caused  by caching worldstates at the wrong state of commitment

## Fixed Issue(s)

fixes #6569 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - `eth_getLogs` - empty topic is a wildcard match [#8420](https://github.com/hyperledger/besu/pull/8420)
 - Upgrade spring-security-crypto to address CVE-2025-22228 [#8448](https://github.com/hyperledger/besu/pull/8448)
 - Fix restoring txpool from disk when blob transactions are present [#8481](https://github.com/hyperledger/besu/pull/8481)
+- Fix for bonsai db inconsistency on abnormal shutdown [#8500](https://github.com/hyperledger/besu/pull/8500)
 
 ## 25.3.0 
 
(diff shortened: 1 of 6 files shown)


matkt:
LGTM, if we can run a validator and a node during some days just to be sure it's ok

matkt (approved):

PR #8506, opened 2025-04-02
8337 modify chain data pruner

Matilda-Clerke:
## PR description
Add an online option to prune pre-merge blocks from the database
## Fixed Issue(s)
https://github.com/hyperledger/besu/issues/8337


besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -125,6 +128,9 @@
 public abstract class BesuControllerBuilder implements MiningParameterOverrides {
   private static final Logger LOG = LoggerFactory.getLogger(BesuControllerBuilder.class);
 
+  private static final long MAINNET_MERGE_BLOCK_NUMBER = 15_537_393;
+  private static final long SEPOLIA_MERGE_BLOCK_NUMBER = 1_735_371;
```
Matilda-Clerke, line 132: Opinions on including this in the NetworkName enum? It seems some networks won't really have a merge block number, so 0 would probably be used in those cases

garyschulte: OptionalInt ftw

besu/src/main/java/org/hyperledger/besu/cli/options/ChainPruningOptions.java
```diff
@@ -42,20 +44,32 @@ public class ChainPruningOptions implements CLIOptions<ChainPrunerConfiguration>
...
       description =
           "Enable the chain pruner to actively prune old chain data (default: ${DEFAULT-VALUE})")
   private final Boolean chainDataPruningEnabled = Boolean.FALSE;
 
+  @CommandLine.Option(
+      hidden = true,
+      names = {PRE_MERGE_PRUNING_ENABLED_FLAG},
+      description =
+          "Enable the chain pruner to actively prune pre-merge blocks, but not headers (default: ${DEFAULT-VALUE})")
+  private final Boolean preMergePruningEnabled = Boolean.FALSE;
```
garyschulte, line 62: it would be nice to see chain pruning be contingent on the sync mode in the same way trielog pruning is.  Like a typical node would want to prune pre-merge history, but a full-sync node would not.  

The network pre-merge cut-off could be an empty Optional for all networks except those that have at release time entered the proto-4444 window.

garyschulte (commented):
I was originally thinking of a one-off purge of pre-merge blocks and receipts but this approach makes sense in light of the future eip-4444 moving window

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -87,11 +108,38 @@ public void onBlockAdded(final BlockAddedEvent event) {
...
+            LOG.info(
+                "Done pruning pre-merge blocks. Unsubscribing from block added event observation");
+            unsubscribeRunnable.run();
+          }
+        });
+  }
+
+  private void pruneChainDataAtBlock(
+      final KeyValueStorageTransaction tx, final long blockNumber, final boolean pruneHeaders) {
     final Collection<Hash> oldForkBlocks = prunerStorage.getForkBlocks(blockNumber);
```
pinges, line 137: I can't see where you are setting the prunerStorage fork blocks for the pre merge pruning. I think you meant to just remove the data for the block number in this call, but the way it works actually uses the block number to retrieve the hashes for the blocks to remove.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
Matilda-Clerke: Ooh, I didn't think of using an Optional. That sounds good

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/ChainDataPrunerTest.java
```diff
@@ -45,9 +45,13 @@ public void singleChainPruning() {
     final ChainDataPruner chainDataPruner =
         new ChainDataPruner(
             blockchainStorage,
+            () -> {},
             new ChainDataPrunerStorage(new InMemoryKeyValueStorage()),
+            0,
+            ChainDataPruner.Mode.CHAIN_PRUNING,
```
pinges, line 51: No tests added for pre merge pruning?

pinges (commented):
Some comments

besu/src/main/java/org/hyperledger/besu/cli/options/ChainPruningOptions.java
```diff
@@ -42,20 +44,32 @@ public class ChainPruningOptions implements CLIOptions<ChainPrunerConfiguration>
...
+
   @CommandLine.Option(
       hidden = true,
       names = {CHAIN_PRUNING_BLOCKS_RETAINED_FLAG},
       description =
           "The number of recent blocks for which to keep the chain data. Should be >= "
               + CHAIN_DATA_PRUNING_MIN_BLOCKS_RETAINED_LIMIT
-              + " (default: ${DEFAULT-VALUE})")
+              + " (default: ${DEFAULT-VALUE}). Unused if "
```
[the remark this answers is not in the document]
Matilda-Clerke: Yes, that's the intention

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
Matilda-Clerke: hmm, I think you're right... this is probably not doing anything useful at all. I'll have to rework this I think

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -1159,14 +1178,32 @@ yield new ForestWorldStateArchive(
     };
   }
 
-  private ChainDataPruner createChainPruner(final BlockchainStorage blockchainStorage) {
+  private ChainDataPruner createChainPruner(
+      final BlockchainStorage blockchainStorage, final Runnable unsubscribeRunnable) {
+    NetworkName network =
+        Stream.of(NetworkName.values())
```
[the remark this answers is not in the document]
Matilda-Clerke: Not exactly. It means besu must be run with a valid NetworkName or we throw a runtime exception. That network name then determines the merge block number that we provide to the ChainDataPruner, defaulting to 0 for anything other than mainnet or sepolia

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/chain/ChainDataPrunerTest.java
Matilda-Clerke: Not yet, but I can add some once we're happy with the code

Gabriel-Trintinalia (commented):
I believe this PR shares some similarities with https://github.com/hyperledger/besu/pull/8582.

The 8582 PR includes a flag for saving only headers to the database, which we could consolidate with the flags in this PR. 

Additionally, 8582 utilises the checkpoint at genesis to establish the header cutoff. I suggest we adopt the same approach here, rather than introducing new constants to NetworkName or other locations.

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -1159,14 +1178,32 @@ yield new ForestWorldStateArchive(
...
+        unsubscribeRunnable,
         new ChainDataPrunerStorage(
             storageProvider.getStorageBySegmentIdentifier(
                 KeyValueSegmentIdentifier.CHAIN_PRUNER_STATE)),
-        chainPrunerConfiguration.getChainPruningBlocksRetained(),
-        chainPrunerConfiguration.getChainPruningBlocksFrequency(),
+        switch (network) {
+          case MAINNET -> MAINNET_MERGE_BLOCK_NUMBER;
+          case SEPOLIA -> SEPOLIA_MERGE_BLOCK_NUMBER;
+          default -> 0;
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: I also prefer using the checkpoint at Genesis for pruning configuration, as it aligns with the approach taken in https://github.com/hyperledger/besu/pull/8582

ahamlat (changes requested):
This is not part of the PR, but more an improvement, make the chain pruner threadpool expose the same metrics names as other thread pools. 
To include the metrics in Besu Full dashboard, we need to change
```
MonitoredExecutors.newBoundedThreadPool(
    ChainDataPruner.class.getSimpleName(),
    1,
    1,
    ChainDataPruner.MAX_PRUNING_THREAD_QUEUE_SIZE,
    metricsSystem));
  ```
to
```
MonitoredExecutors.newBoundedThreadPool(
            EthScheduler.class.getSimpleName() + "-ChainDataPruner",
            1,
            1,
            ChainDataPruner.MAX_PRUNING_THREAD_QUEUE_SIZE,
            metricsSystem));
```
One other solution is to integrate the thread pool to EthScheduler.
This is related to the formula we're using to display the metrics of monitred thread pools
```
as the formula currently is something like this
/besu_executors_ethscheduler_([^_]*)_.*/
```

File: besu/src/main/java/org/hyperledger/besu/cli/options/ChainPruningOptions.java (hunk 2 above)
Comment (Matilda-Clerke) [reply besu/src/main/java/org/hyperledger/besu/cli/options/ChainPruningOptions.java:62]: While it would be nice to enable pruning by default, I think users might be upset by their stored history being deleted unexpectedly.

File: besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java [hunk 8]
```diff
@@ -2662,7 +2662,8 @@ private void setMergeConfigOptions() {
 
   /** Set ignorable segments in RocksDB Storage Provider plugin. */
   public void setIgnorableStorageSegments() {
-    if (!unstableChainPruningOptions.getChainDataPruningEnabled()) {
+    if (!unstableChainPruningOptions.getChainDataPruningEnabled()
+        && !unstableChainPruningOptions.getPreMergePruningEnabled()) {
```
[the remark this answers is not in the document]
Matilda-Clerke: I'm not sure. In both cases, we're utilising the same stored variable, so switching between the two haphazardly may cause an issue. For the pre-merge pruning, it may cause blocks to be skipped. It's difficult to predict what would happen the other way around.

Matilda-Clerke:
Addition of the 1 second sleep to avoid conflict with other processes seems to work well. We can see below there's very little impact on newPayload and FCU times
image
Also noteworthy is that using a smaller batch size results in a much smaller disk usage bloat
image

besu/src/main/java/org/hyperledger/besu/cli/BesuCommand.java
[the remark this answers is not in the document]
siladu: > I'm not sure. 

Think we should test this so we know how to advise users in the docs/discord (if there's no way around it in code).

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -16,37 +16,62 @@
...
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ChainDataPruner implements BlockAddedObserver {
-  public static final int MAX_PRUNING_THREAD_QUEUE_SIZE = 16;
   private static final Logger LOG = LoggerFactory.getLogger(ChainDataPruner.class);
+  private static final int LOG_PRE_MERGE_PRUNING_PROGRESS_REPEAT_DELAY = 300;
```
siladu, line 30:
```suggestion
  private static final int LOG_PRE_MERGE_PRUNING_PROGRESS_REPEAT_DELAY_SECONDS = 300;
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -87,6 +112,53 @@ public void onBlockAdded(final BlockAddedEvent event) {
         });
   }
 
+  private void preMergePruningAction() {
+    pruningExecutor.submit(
+        () -> {
+          try {
+            Thread.sleep(1000);
```
siladu, line 119: Would rather we didn't use a sleep, the executor API has delay capabilities built in, e.g. MonitoredExecutors.newScheduledThreadPool

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -87,6 +112,53 @@ public void onBlockAdded(final BlockAddedEvent event) {
...
+            updater.commit();
+            prunerStorage.setPruningMark(pruningTransaction, expectedNewPruningMark);
+            pruningTransaction.commit();
+            LogUtil.throttledLog(
+                () -> LOG.info("Pruned pre-merge blocks up to {}", expectedNewPruningMark),
+                logPreMergePruningProgress,
+                LOG_PRE_MERGE_PRUNING_PROGRESS_REPEAT_DELAY);
+            if (expectedNewPruningMark == mergeBlock) {
+              LOG.info(
+                  "Done pruning pre-merge blocks. Unsubscribing from block added event observation");
```
siladu, line 153: > Done pruning pre-merge blocks

This is useful for the user

> Unsubscribing from block added event observation

This is internal details, I would delete or move to debug.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/storage/keyvalue/KeyValueSegmentIdentifier.java
```diff
@@ -25,7 +25,7 @@
 
 public enum KeyValueSegmentIdentifier implements SegmentIdentifier {
   DEFAULT("default".getBytes(StandardCharsets.UTF_8)),
-  BLOCKCHAIN(new byte[] {1}, EnumSet.allOf(DataStorageFormat.class), true, true, false),
+  BLOCKCHAIN(new byte[] {1}, EnumSet.allOf(DataStorageFormat.class), true, true, true),
```
siladu, line 28: As I've found during testing https://github.com/hyperledger/besu/pull/8599
we should be wary of this because it will amplify disk space, even without any pruning occurring.

With this setting, I think you will find the space increases despite the pruning, at least initially.

In my PR I have made it false here and enabled by a flag at runtime when the configuration is available.

Currently I am thinking the user should temporarily enable GC during pruning and then remove again, but I think more testing is required to ensure we don't breach the 2TB limit.

util/src/main/java/org/hyperledger/besu/util/log/LogUtil.java
```diff
@@ -52,6 +51,23 @@ public static void throttledLog(
...
   }
 
+  /**
+   * Throttles logging to a given logger.
+   *
+   * @param logAction a Runnable which performs the actual log call
+   * @param shouldLog AtomicBoolean to track whether the message should be logged
+   * @param logRepeatDelay delay in seconds between repeated logs
+   */
+  public static void throttledLog(
```
siladu, line 61: Do we need this overload - can't just use the existing method above?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
Matilda-Clerke: Unfortunately, all of the scheduled executors will produce multiple threads when and if needed, which would potentially change the behaviour for the other ChainDataPruner mode as it assumes all threads will be executed in serial. So we have these options:

1. Do a sleep, as currently implemented
2. Use a scheduled executor for both modes and hope the other mode always completes execution of the previous thread before starting a new one
3. Create a scheduled executor and use it just for this mode, probably increasing start up time and memory footprint
4. Use a scheduled executor for both modes, but enforce serial execution for the other mode using a semaphore or something

I'm happy to adjust the code if you point me in the best direction

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
Matilda-Clerke: Debug makes sense

util/src/main/java/org/hyperledger/besu/util/log/LogUtil.java
Matilda-Clerke: With the existing method it seemed weird to pass in a String (which we would have to format separately) when we could just include it in the LOG call in the Runnable lambda. Personally, I'd rather update the other throttledLog calls to the new method than vice versa, but I don't think there's really a downside of having both available.

Matilda-Clerke:
> Good news! What batch sizes did you test?

That was for all batch sizes tested: 1000, 100, and 10

macfarla (commented):
needs a changelog

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/ChainDataPruner.java
```diff
@@ -87,6 +112,56 @@ public void onBlockAdded(final BlockAddedEvent event) {
...
+            }
+            updater.commit();
+            prunerStorage.setPruningMark(pruningTransaction, expectedNewPruningMark);
+            pruningTransaction.commit();
+            LOG.debug("Pruned pre-merge blocks up to {}", expectedNewPruningMark);
+            LogUtil.throttledLog(
+                () -> LOG.info("Pruned pre-merge blocks up to {}", expectedNewPruningMark),
+                logPreMergePruningProgress,
+                LOG_PRE_MERGE_PRUNING_PROGRESS_REPEAT_DELAY_SECONDS);
+            if (expectedNewPruningMark == mergeBlock) {
```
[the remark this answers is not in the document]
Matilda-Clerke: No, it's not possible as we set it like
```final long expectedNewPruningMark = Math.min(storedPruningMark + pruningQuantity, mergeBlock);```
So it will never exceed the mergeBlock. It's worth noting that after removing blocks up to the merge block, we also unsubscribe from listening to `blockAdded` events, preventing the code being called in future

File: besu/src/main/java/org/hyperledger/besu/cli/options/ChainPruningOptions.java [hunk 15]
```diff
@@ -25,11 +25,13 @@
 /** The Chain pruning CLI options. */
 public class ChainPruningOptions implements CLIOptions<ChainPrunerConfiguration> {
   private static final String CHAIN_PRUNING_ENABLED_FLAG = "--Xchain-pruning-enabled";
+  private static final String PRE_MERGE_PRUNING_ENABLED_FLAG = "--Xpre-merge-pruning-enabled";
```
siladu, line 28: Should we use Xhistory-expiry-prune instead? I can merge it in now

Matilda-Clerke:
Latest testing, including enabling database garbage collection, shows fairly minimal impact on newPayload and FCU
image
image

Matilda-Clerke:
Some interesting findings from the latest test run (with double execution issue fixed)
image
All batch sizes eventually cause some increase in newPayload execution time above baseline, but lower batch size causes it to be less impactful over a longer period of time.

siladu (approved):
LGTM 🚢

PR #8472, opened 2025-03-27
estimate gas - don't override gasPrice if not set

macfarla:
Signed-off-by: Sally MacFarlane <macfarla.github@gmail.com>
## PR description
If not set in call params, don't override gasPrice. TransactionSimulator does the work of deciding whether it is set and whether it needs overriding.

## Fixed Issue(s)
This was discovered by looking at rpc-compat hive tests but doesn't actually fix any hive tests directly.
What is fixed by this is transactions will be detected as EIP1559 if no gas price set.

### Thanks for sending a pull request! Have you done the following?

- [ ] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [ ] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [ ] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [ ] spotless: `./gradlew spotlessApply`
- [ ] unit tests: `./gradlew build`
- [ ] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

PR #8508, opened 2025-04-02
create access list - don't override gasPrice if not set

macfarla:
## PR description
follow on from #8472 - this is the same fix for eth_createAccessList

If not set in call params, don't override gasPrice. TransactionSimulator does the work of deciding whether it is set and whether it needs overriding.

Also added a spec test for eth_createAccessList although it does duplicate a hive test so not strictly necessary just a faster feedback loop

##Fixed Issue(s)
This was discovered by looking at rpc-compat hive tests but doesn't actually fix any hive tests directly.
What is fixed by this is transactions will be detected as EIP1559 if no gas price set.


pinges (approved):
LGTM

Gabriel-Trintinalia (approved):

PR #8510, opened 2025-04-03
Add support for grafana-pyroscope to besu

joshuafernandes:
## PR description
Add support for [grafana-pyroscope](https://grafana.com/docs/pyroscope/latest/) to besu

To use, set  the following env vars in compose, k8s etc
```
JAVA_OPTS="-XX:-Inline -javaagent:/opt/pyroscope/lib/pyroscope.jar"
PYROSCOPE_SERVER_ADDRESS=http://my_pyroscope_endpoint:4040
```
image


ahamlat:
This can have a non negligible performance impact as we're disabling inlining. Is it possible to without disabling inlining, otherwise we need to not make it default, but enable it when needed.

ahamlat (commented):
This can have a non negligible performance impact as we're disabling inlining. Is it possible to without disabling inlining, otherwise we need to not make it default, but enable it when needed.

joshuafernandes:
> This can have a non negligible performance impact as we're disabling inlining. Is it possible to without disabling inlining, otherwise we need to not make it default, but enable it when needed.

Lost me @ahamlat . There is 2 ways to inject pyro in:

ahamlat:
I agree that the Java options are on the user side, but we can add that `-XX:-Inline` helps to keep a stacktrace that similar to the code, but we can not add it and still have Pyroscope working fine.
Maybe other options that will be needed from the use side are` -XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints`, that helps all the profilers to have more accurate profiling data.
Can we hold off this until we're able to test it in our Grafana stack ?

ahamlat (approved):
LGTM

macfarla:
maybe worth a changelog entry

Issue #8496, opened 2025-03-31
Share p2p-host via RLPX handshake


tsamsiyu:
### Description
As a software developer working on a project that provisions PoA nodes, I’d like to suggest an optional feature: allowing Besu nodes to share their `p2p-host` during the RLPx handshake. This would help other peers validate local permissions correctly - particularly important in cloud environments where (1) the egress IP differs from the ingress IP, and (2) preserving the ingress sender IP requires intrusive changes to Kubernetes networking.

### Acceptance Criteria
* when starting a new node, I set `p2p-host-share-via-rlpx=true`
* peers accepting request with additional field use this field when adding a node to the maintained peers list

### Versions (Add all that apply)
* Software version: 25.3.0
* Java version: 17
* Docker Version: 27.5.1
* Kubernetes Server version: v1.30.10-eks-bc803b4
* Consensus: QBFT

jframe:
Possibly also the same cause of this hive test failure https://hivetests2.ethdevops.io/viewer.html?suiteid=1743453408-159d9b012784e1bbdb67762ca04e4a0f.json&suitename=discv4&testid=1&file=results%2Fbesu%2Fclient-c3fc5748204eff9c7546419cc3c207e1e0dea770fa4de923f4fd08a3d46f61fa.log

PR #8516, opened 2025-04-03
feat: allow exchanging p2p-host value via rlpx connection

tsamsiyu:
## PR description
This PR adds a new Besu config option `p2p-host-share-via-rlpx`, which is disabled by default. When enabled, it modifies the RLPx Hello message by including the node’s P2P address in the message body. As a result, the bonding node will use this advertised P2P address as the peer’s identity, instead of relying on the IP address from the TCP socket

## Fixed Issue(s)

[#8496](https://github.com/hyperledger/besu/issues/8496)


macfarla:
@matthew1001 any input on this one?

vitali-zevako:
+1 for this feature, currently we need to restart nodes one by one in case of permissioning is enabled, so it finally resolves correct dns names of allowed nodes

PR #8519, opened 2025-04-04
feat: implement EIP-7642 eth/69 subprotocol

Gabriel-Trintinalia:
## Description

This pull request Implements: https://eips.ethereum.org/EIPS/eip-7642 (Eth/69 subprotocol). 

### Changes

* **Integration of eth/69 into the eth protocol manager**
  * Sets the maximum capability default to eth/68 while eth/69 is under testing. This ensures that only nodes explicitly requesting eth/69 will enable it.

* **Support for the Eth/69 receipt format**
  * Removes the bloom filter and introduces a new receipt format:
    ```
    (eth/68): receipt = {legacy-receipt, typed-receipt} with
    typed-receipt = tx-type || rlp(legacy-receipt)

    legacy-receipt = [
        post-state-or-status: {B_32, {0, 1}},
        cumulative-gas: P,
        bloom: B_256,
        logs: [log₁, log₂, ...]
    ]
    (eth/69): receipt = [tx-type, post-state-or-status, cumulative-gas, logs]
    ```

* **Modification of the status message to support Eth/69**
  * Removes `td`
  * Adds `block range`
  * Moves `blockHash` to the end:
    ```
    (eth/68): [version: P, networkid: P, td: P, blockhash: B_32, genesis: B_32, forkid]

    (eth/69): [version: P, networkid: P, genesis: B_32, forkid, earliestBlock: P, latestBlock: P, latestBlockHash: B_32]
    ```

* **Addition of blockRangeBroadcaster**
  * Implements a new message `blockRangeUpdateMessage` (0x11):
    ```
    [earliestBlock: P, latestBlock: P, latestBlockHash: B_32]
    ```
  * Sends `blockRangeUpdateMessage` with the earliest and latest block information to all peers supporting eth/69 every 2 minutes.
  * Updates peer height whenever a `blockRangeUpdateMessage` is received.

* **Wiring and tests**


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/StatusMessage.java
```diff
@@ -131,56 +144,173 @@ private EthStatus status() {
     return status;
   }
 
+  public static StatusMessage.Builder builder() {
+    return new StatusMessage.Builder();
+  }
+
+  public static class Builder {
```
fab-10, line 151: you could use immutables annotation to avoid writing boilerplate code

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockRangeBroadcaster.java
```diff
@@ -0,0 +1,113 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BlockRangeBroadcaster implements UnverifiedForkchoiceListener {
+  private static final Logger LOG = LoggerFactory.getLogger(BlockRangeBroadcaster.class);
+
+  // range update block interval
+  static final int BLOCK_RANGE_UPDATE_INTERVAL = 32; // one slot
```
fab-10, line 37:
```suggestion
  static final int BLOCK_RANGE_UPDATE_INTERVAL = 32; // one epoch
```

BTW at least one epoch, since there could be gaps, but there is no way we will know about that

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockRangeBroadcaster.java
```diff
@@ -0,0 +1,113 @@
...
+  public void onNewUnverifiedForkchoice(final ForkchoiceEvent event) {
+    blockchain
+        .getBlockHeader(event.getHeadBlockHash())
+        .ifPresent(
+            header -> {
+              if (header.getNumber() % BLOCK_RANGE_UPDATE_INTERVAL == 0) {
+                broadcastBlockRange(0L, header.getNumber(), header.getHash());
+              }
+            });
+  }
```
fab-10, line 60: as said before, I think it is better to use block added events, and in ant case making this asynchronous avoid impacting on engine API response time

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockRangeBroadcaster.java
```diff
@@ -0,0 +1,113 @@
...
+          BlockRangeUpdateMessage.readFrom(message.getData());
+      final long earliestBlockNumber = blockRangeUpdateMessage.getEarliestBlockNumber();
+      final long latestBlockNumber = blockRangeUpdateMessage.getLatestBlockNumber();
+      final Hash blockHash = blockRangeUpdateMessage.getBlockHash();
+      LOG.debug(
+          "Received blockRange=[{}, {}, {}] from peer={}",
+          earliestBlockNumber,
+          latestBlockNumber,
+          blockHash,
+          message.getPeer().getLoggableId());
```
fab-10, line 74: since this statement is doing some work and allocations, better to use lambdas for this log

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManagerTest.java
```diff
@@ -272,6 +279,40 @@ public void disconnectNewPoWPeers() {
     }
   }
 
+  @Test
+  public void disconnectOnMissingBlockRangeWhenEth69() {
```
fab-10, line 283: what happens if capability is eth/69 but the peer still sends a status message with total difficulty? is that case worth a test?

besu/src/main/java/org/hyperledger/besu/controller/BesuControllerBuilder.java
```diff
@@ -734,6 +734,14 @@ public BesuController build() {
...
             forkIdManager);
 
+    ethProtocolManager
+        .getBlockRangeBroadcaster()
+        .ifPresent(
+            broadcaster ->
+                protocolContext
+                    .getConsensusContext(MergeContext.class)
+                    .addNewUnverifiedForkchoiceListener(broadcaster));
+
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: Changed to a `ScheduledFutureTaskWithFixedDelay`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockRangeBroadcaster.java
Gabriel-Trintinalia: Changed to a `ScheduledFutureTaskWithFixedDelay`

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/ChainState.java
```diff
@@ -77,6 +91,16 @@ public void update(final Hash blockHash, final long blockNumber) {
     }
   }
 
+  public void update(final Hash blockHash, final long blockNumber, final long earliestBlockHeight) {
+    synchronized (this) {
```
Matilda-Clerke, line 95: Probably out of scope for this PR, but should we consider using a read/write lock instead of basic synchronization for the ChainState. That way, multiple reads can occur at once, but we still have the ability to lock it all down for writing.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -429,6 +462,24 @@ private void handleStatusMessage(final EthPeer peer, final Message message) {
     }
   }
 
+  private void registerStatusReceived(final EthPeer peer, final StatusMessage status) {
+    // pre eth/69 status message
+    if (status.blockRange().isEmpty()) {
```
Matilda-Clerke, line 467: Would it make sense to move this logic into StatusMessage? Maybe outputting the specific eth protocol version the StatusMessage appears to be. Then we could do something like
```
switch (status.getEthProtocolVersion()) {
    case 66, 67, 68 -> //build for eth68
    case 69 -> //build for eth69
}
```

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/BlockRangeUpdateMessage.java
```diff
@@ -0,0 +1,105 @@
...
+
+  public static BlockRangeUpdateMessage create(
+      final long earliestBlockNumber, final long latestBlockNumber, final Hash blockHash)
+      throws IllegalArgumentException {
+    final BlockRangeUpdateMessageData msgData =
+        new BlockRangeUpdateMessageData(earliestBlockNumber, latestBlockNumber, blockHash);
+    final BytesValueRLPOutput out = new BytesValueRLPOutput();
+    msgData.writeTo(out);
+    final Bytes data = out.encoded();
+    return new BlockRangeUpdateMessage(data);
```
Matilda-Clerke, line 50: We're not storing the `BlockRangeUpdateMessageData` in the new object? It has a field for it, and we'll only have to parse the RLP data again later to get it back

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockRangeBroadcaster.java
```diff
@@ -0,0 +1,157 @@
...
+    this.blockchain = blockchain;
+    ethContext
+        .getEthMessages()
+        .subscribe(EthProtocolMessages.BLOCK_RANGE_UPDATE, this::handleBlockRangeUpdateMessage);
+    ethContext
+        .getScheduler()
+        .scheduleFutureTaskWithFixedDelay(
+            this::broadcastBlockRange,
+            Duration.ofSeconds(BLOCK_RANGE_UPDATE_INTERVAL),
+            Duration.ofSeconds(BLOCK_RANGE_UPDATE_INTERVAL));
```
fab-10, line 56: has the spec change to broadcast every minute instead of every 32 blocks?

Gabriel-Trintinalia: Yes, it is up to the node to decide how often it wants to send the update.

https://github.com/ethereum/devp2p/pull/262

The spec mentions every 2 minutes.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/Blockchain.java
```diff
@@ -51,6 +51,13 @@ public interface Blockchain {
    */
   Optional<Hash> getSafeBlock();
 
+  /**
+   * Return the hash of the earliest block in the canonical chain.
+   *
+   * @return The hash of the earliest block in the chain.
+   */
+  Optional<Hash> getEarliest();
```
Matilda-Clerke, line 59: `getEarliest` is a bit ambiguous, could we call it `getEarliestBlockHash`? Likewise for the methods in BlockchainStorage too

Matilda-Clerke (commented):
Looks like we're storing an "earliest block hash" rather than retrieving the earliest block from the database. Does that mean the updated sync process will need to `set` this? And likewise any block pruner should also `set` the earliest block hash after any pruning operations?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/StatusMessage.java
```diff
@@ -206,4 +337,27 @@ public String toString() {
   public String toStringDecoded() {
     return status().toString();
   }
+
+  public static class BlockRange {
```
jframe, line 341: This could be a record

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockRangeBroadcaster.java
```diff
@@ -0,0 +1,174 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BlockRangeBroadcaster {
+  private static final Logger LOG = LoggerFactory.getLogger(BlockRangeBroadcaster.class);
+
+  // range update block interval
+  static final int BLOCK_RANGE_UPDATE_INTERVAL = 60;
```
jframe, line 40: Why not use the recommended default of 2 minutes here?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/Blockchain.java
jframe, line 59: Why not just store the block number instead? It looks like everywhere we use this, we just want the block number and not the block hash so end doing another db call to get the block number from the hash.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -446,10 +497,20 @@ public void blockMined(final Block block) {
...
+            .map(
+                hash ->
+                    blockchain
+                        .getBlockHeader(hash)
+                        .orElseThrow(
+                            () ->
+                                new IllegalStateException(
+                                    "Unable to get earliest block header from blockchain."))
+                        .getNumber())
+            .orElse(0L),
```
pinges, line 513: Should we set the right value if we have done checkpoint sync here?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -787,6 +792,12 @@ public void setSafeBlock(final Hash blockHash) {
     updater.commit();
   }
 
+  public void setEarliest(final Hash blockHash) {
```
pinges, line 795: looks like we are always after the block number when we call getEarliest(). Should we store the block number instead? Then we do not have to make the conversion hash -> block number when we need it.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptEncodingConfiguration.java
```diff
@@ -15,7 +15,20 @@
 package org.hyperledger.besu.ethereum.core.encoding.receipt;
 
 public class TransactionReceiptEncodingConfiguration {
-  public static final TransactionReceiptEncodingConfiguration NETWORK = new Builder().build();
+
```
pinges, line 18: Not sure I really like this class. There are some concepts mixed, like encoding format, and whether we encode the bloom filter. When we are doing eht69 we are not encoding bloom! 
Is this just used for encoding?

pinges (commented):
Few comments

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptDecoder.java
```diff
@@ -28,18 +28,24 @@
...
 
-  /**
-   * Creates a transaction receipt for the given RLP
-   *
-   * @param input the RLP-encoded transaction receipt
-   * @return the transaction receipt
-   */
-  public static TransactionReceipt readFrom(final RLPInput input) {
-    return readFrom(input, true);
-  }
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: I removed it to force the developers to choose whether the revert reason is allowed, and not rely on the default. The default should not be revert reason allowed.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptDecoder.java
```diff
@@ -49,34 +55,116 @@ public static TransactionReceipt readFrom(final RLPInput input) {
...
+    }
+    rlpInput.leaveList();
+    return receipt;
+  }
+
+  private static TransactionReceipt decodeEth69Receipt(
+      final RLPInput input, final RLPInput transactionByteRlp, final RLPInput statusOrStateRoot) {
+    int transactionByte = transactionByteRlp.readIntScalar();
+    final TransactionType transactionType =
+        transactionByte == 0x00 ? TransactionType.FRONTIER : TransactionType.of(transactionByte);
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: yes, but internally FRONTIER is `0xf8`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptDecoder.java
```diff
@@ -49,34 +55,116 @@ public static TransactionReceipt readFrom(final RLPInput input) {
...
+    final List<Log> logs = input.readList(logInput -> Log.readFrom(logInput, false));
+    final LogsBloomFilter bloomFilter = LogsBloomFilter.builder().insertLogs(logs).build();
+    return createReceipt(
+        transactionType, statusOrStateRoot, cumulativeGas, logs, bloomFilter, Optional.empty());
+  }
 
+  private static TransactionReceipt decodeLegacyReceipt(
+      final RLPInput input,
+      final RLPInput firstElement,
+      final RLPInput secondElement,
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: // The first element can be either the transaction type (eth/69 or stateRootOrStatus (eth/68
    // The second element can be either the state root or status (eth/68) or cumulative gas (eth/69)

I have added this comments to the code to clarify

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/ChainState.java
Gabriel-Trintinalia: yes, probably out of scope for this PR

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
Gabriel-Trintinalia: I am not sure if I understood your idea here

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/BlockRangeUpdateMessage.java
Gabriel-Trintinalia: That is how the other messages do as well

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/ChainState.java
```diff
@@ -52,6 +53,10 @@ public long getEstimatedHeight() {
     return estimatedHeight;
   }
 
+  public long getEarliestBlockHeight() {
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: removed

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -79,7 +79,7 @@ public class EthPeers implements PeerSelector {
       Comparator.comparing((final EthPeer p) -> p.getReputation().getScore())
           .thenComparing(CHAIN_HEIGHT);
 
-  public static final Comparator<EthPeer> HEAVIEST_CHAIN =
+  public static final Comparator<EthPeer> TOTAL_DIFFICULT_THEN_HEIGHT =
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: done

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/messages/StatusMessage.java
Gabriel-Trintinalia: changed

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/BlockRangeBroadcaster.java
Gabriel-Trintinalia: I had 60 seconds to make it more often in my tests. Changed back to 2 minutes

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
```diff
@@ -127,7 +127,8 @@ public void handleExecutionBlockReceipts(
                               Bytes.wrap(executionBlockReceipts.receipts()), false);
                       final List<TransactionReceipt> receiptsForBlock = new ArrayList<>();
                       input.readList(
-                          (in) -> receiptsForBlock.add(TransactionReceiptDecoder.readFrom(in)));
+                          (in) ->
+                              receiptsForBlock.add(TransactionReceiptDecoder.readFrom(in, true)));
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: yes, it should not allow the revert reason.  I will update it.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptDecoder.java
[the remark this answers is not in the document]
Gabriel-Trintinalia: See https://github.com/hyperledger/besu/pull/8519#discussion_r2062571450 below. It is a perfect example of this.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptDecoder.java
```diff
@@ -49,34 +55,116 @@ public static TransactionReceipt readFrom(final RLPInput input) {
...
+  }
+
+  private static TransactionReceipt decodeEth69Receipt(
+      final RLPInput input, final RLPInput transactionByteRlp, final RLPInput statusOrStateRoot) {
+    int transactionByte = transactionByteRlp.readIntScalar();
+    final TransactionType transactionType =
+        transactionByte == 0x00 ? TransactionType.FRONTIER : TransactionType.of(transactionByte);
+    final long cumulativeGas = input.readLongScalar();
+    final List<Log> logs = input.readList(logInput -> Log.readFrom(logInput, false));
+    final LogsBloomFilter bloomFilter = LogsBloomFilter.builder().insertLogs(logs).build();
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: Good point

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/chain/DefaultBlockchain.java
```diff
@@ -966,4 +965,30 @@ public Optional<Cache<Hash, Difficulty>> getTotalDifficultyCache() {
...
+    long high = getChainHeadBlockNumber();
+    while (low < high) {
+      long mid = (low + high) / 2;
+      if (getBlockByNumber(mid).isPresent()) {
+        high = mid;
+      } else {
+        low = mid + 1;
+      }
+    }
+    return getBlockByNumber(low).map(block -> block.getHeader().getNumber());
```
Matilda-Clerke, line 992: One problem I've noticed with this: If the genesis block is genuinely our earliest block, this will return block 1. Later in the block range broadcaster, I noticed we're relying on this method returning an empty optional in order to indicate that the start of our block range should be the genesis block, but I can't see any way that will actually happen.

I think it would be best to ensure getEarliestBlockNumber always returns an appropriate value (unless we're pre-syncing?), including returning 0 if that's the earliest block we have.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
Matilda-Clerke: Iirc, it looked like we're trying to determine which eth protocol version the StatusMessage is a part of. I thought it might be handy to move the logic to determine that into the StatusMessage class.

pinges (commented):
Few comments

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -139,8 +140,7 @@ public EthPeers(
     this.clock = clock;
     this.permissioningProviders = permissioningProviders;
     this.maxMessageSize = maxMessageSize;
-    this.bestPeerComparator =
-        currentProtocolSpecSupplier.get().isPoS() ? CHAIN_HEIGHT : TOTAL_DIFFICULTY_THEN_HEIGHT;
+    this.bestPeerComparator = TOTAL_DIFFICULTY_THEN_HEIGHT;
```
Matilda-Clerke, line 143: Wouldn't just `CHAIN_HEIGHT` be the more sensible default given we're always PoS now?

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthProtocolManager.java
Gabriel-Trintinalia: Removed

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptDecoder.java
```diff
@@ -49,34 +55,116 @@ public static TransactionReceipt readFrom(final RLPInput input) {
...
-    RLPInput input = rlpInput;
-    TransactionType transactionType = TransactionType.FRONTIER;
+    // The first byte of the RLP input indicates whether the receipt is typed or flat.
     if (!rlpInput.nextIsList()) {
-      final Bytes typedTransactionReceiptBytes = input.readBytes();
-      transactionType = TransactionType.of(typedTransactionReceiptBytes.get(0));
-      input = new BytesValueRLPInput(typedTransactionReceiptBytes.slice(1), false);
+      return decodeTypedReceipt(rlpInput, revertReasonAllowed);
+    } else {
+      return decodeFlatReceipt(rlpInput, revertReasonAllowed);
```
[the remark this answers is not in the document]
Gabriel-Trintinalia: I will still need the first and second element

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/receipt/TransactionReceiptDecoder.java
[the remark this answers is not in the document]
Gabriel-Trintinalia: Not changing this now as this PR is big already

Matilda-Clerke:
Remaining work for this PR: create sepolia test nodes (eth/69 enabled and disabled via max-capability flag)

Matilda-Clerke (approved):

macfarla (approved):
needs a changelog entry

macfarla:
Is this ready to merge or is there more testing to do eg interop with other ELs?

Matilda-Clerke:
I have a couple of test nodes up still testing sync behaviour for this PR. I'll be aiming to merge it soon

PR #8522, opened 2025-04-04
Avoid allocating a `Set` every time `requestIdCompatible` is called

fab-10:
## PR description

`EthProtocol.requestIdCompatible` is called for every p2p message, and in the previous implementation it was allocating a new `Set` every time, but since the data in the set does not change, it can be created once, moreover to avoid (un)boxing from and to int and hashes, using a `BitSet` optimize the lookup.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/EthProtocol.java
```diff
@@ -46,8 +46,17 @@ public static boolean requestIdCompatible(final int code) {
             EthProtocolMessages.GET_NODE_DATA,
             EthProtocolMessages.NODE_DATA,
             EthProtocolMessages.GET_RECEIPTS,
-            EthProtocolMessages.RECEIPTS)
-        .contains(code);
+            EthProtocolMessages.RECEIPTS);
+    REQUEST_ID_MESSAGES =
+        new BitSet(requestIdMessages.stream().mapToInt(i -> i).max().getAsInt() + 1);
```
macfarla, line 51: is this more optimal than just getting the size of the set?

macfarla (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/EthProtocol.java
ahamlat: We can make it even constant, like 64 which is the size of a long (64 bits).

ahamlat (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/EthProtocol.java
fab-10: the requestIdMessage is the index of the bitset, so the bitset should be large enough for the message with the largest index, since bitset starts from 0 then the optimal size is max index + 1, not sure what do you mean by decoupling?
Then I plan to make `EthProtocolMessages` an enum, so we can directly use EnumSet that is clearer.

PR #8533, opened 2025-04-08
Feature/service manager in protocol context

garyschulte:
## PR description

Adds plugin service manager to protocol context.  The specific motivation is to have plugin services available to AbstractBlockProcessor (in a subsequent PR), but generally this is a logical and convenient way to surface plugin services in the besu codebase.

The scope of the code change is minimal, but there are a lot of test fixture changes.  Also, the ProtocolContext constructor is now protected so we do not have to plumb a bunch of unnecessary items into it for unit tests which directly construct a test protocol context.

- [X] spotless: `./gradlew spotlessApply`
- [X] unit tests: `./gradlew build`


plugin-api/src/main/java/org/hyperledger/besu/plugin/ServiceManager.java
```diff
@@ -52,4 +54,22 @@ public interface ServiceManager {
    *     is unavailable
    */
   <T extends BesuService> Optional<T> getService(Class<T> serviceType);
+
+  /** A toy implementation of ServiceManager, suitable for test and mocks. */
```
siladu, line 58: Could this live in the test packages instead?

siladu (approved):

jflo (approved):

PR #8534, opened 2025-04-09
Add block import tracer provider to plugin-api

garyschulte:
## PR description

Adds the ability to specify a default tracer to AbstractBlockProcessor, so imported blocks may implement tracing.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
@@ -44,7 +44,6 @@ import org.hyperledger.besu.evm.blockhash.BlockHashLookup;
 import org.hyperledger.besu.evm.tracing.OperationTracer;
 import org.hyperledger.besu.evm.worldstate.WorldState;
 import org.hyperledger.besu.evm.worldstate.WorldUpdater;
-import org.hyperledger.besu.plugin.ServiceManager;
 import org.hyperledger.besu.plugin.services.BlockImportTracerProvider;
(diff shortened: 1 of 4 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -96,6 +99,17 @@ protected AbstractBlockProcessor(
...
 
+  private BlockAwareOperationTracer getBlockImportTracer(
+      final ServiceManager pluginContext, final BlockHeader header) {
+
+    // if we have a BlockImportTracerProvider from pluginContext, use it.
+    return Optional.ofNullable(pluginContext)
+        .flatMap(serviceManager -> serviceManager.getService(BlockImportTracerProvider.class))
+        .map(z -> z.getBlockImportTracer(header))
+        // otherwise return NO_TRACING
+        .orElse(BlockAwareOperationTracer.NO_TRACING);
```
fab-10, line 110: not blocking: can the resolution of the tracer to use be done only once since I assume the tracer will not change?

fab-10 (commented):
LGTM, please add a CHANGELOG entry

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
fab-10: I see, if it is not straight forward it could stay like that or just use a memoizing supplier

fab-10 (approved):
CHANGELOG entry missing

garyschulte:
> CHANGELOG entry missing

Ah yes, a memoizing supplier would be 👍

PR #8536, opened 2025-04-09
Default ExtraData to "besu vM.m.p" rather than empty

benaadams:
## PR description

Default ExtraData to "besu vM.m.p" rather than empty e.g. "besu v25.4.0"

This is done similarly by Geth, Nethermind and reth; and will be in latest version of Erigon https://github.com/erigontech/erigon/pull/14419

This helps if are any built block issues to know if is old version (with known fixed issues) or new version


benaadams:
Added changelog, fixed compilation error

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/MiningConfiguration.java
```diff
@@ -213,7 +215,24 @@ public MutableInitValues getMutableInitValues() {
 
   @Value.Immutable
   public interface MutableInitValues {
-    Bytes DEFAULT_EXTRA_DATA = Bytes.EMPTY;
+    // Obtain the raw short version
+    String RAW_SHORT_VERSION = MutableInitValues.class.getPackage().getImplementationVersion();;
+    // Strip everything from the first '-' or '+' onward 
+    // (handles both e.g. "v23.1.1-dev-ac23d311" or "v23.1.1+commit.abcd1234").
+    String SANITIZED_VERSION = RAW_SHORT_VERSION.replaceFirst("[-+].*", "");
```
fab-10, line 222: I think it is better to keep the short commit hash, that helps identifying if it is a released version or a development one, and it should always fit in extra-data.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/MiningConfiguration.java
```diff
@@ -213,7 +215,24 @@ public MutableInitValues getMutableInitValues() {
 
   @Value.Immutable
   public interface MutableInitValues {
-    Bytes DEFAULT_EXTRA_DATA = Bytes.EMPTY;
+    // Obtain the raw short version
+    String RAW_SHORT_VERSION = MutableInitValues.class.getPackage().getImplementationVersion();;
```
fab-10, line 219: I see why you are not able to use `BesuInfo` that is the best fit for this, so I am reworking a bit the code to be able to use it.

fab-10: Create this PR https://github.com/benaadams/besu/pull/1 against your repo

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/MiningConfiguration.java
benaadams: Depends on guarantees of your build pipeline 😀

Keeping the commit hash is more informative

Feel free to absorb or modify this PR

benaadams:
Updated with changes

fab-10:
There are some other tests failing because they expected any empty extra data field, I can take care of them tomorrow

macfarla (approved):
LGTM but still some test failures to fix

benaadams:
> PR [benaadams#2](https://github.com/benaadams/besu/pull/2) that should fix the failing tests

Merged

benaadams:
Seen in the wild

PR #8537, opened 2025-04-10
Tune and make configurable the `PeerTransactionTracker` 

fab-10:
## PR description

In PR #7755 the default behavior of the peer transaction tracker changed from always remember the txs exchanged with a peer (at least until they fits in the cache) to forget a txs has been already exchanged in case it is evicted from the pool for reason not related to the validity.

While the new default solves issues with re-broadcasting the same tx in small networks, it does not works well for public networks, where there are many thousands of pending txs and a lot of independent nodes, since it can cause the broadcast and the re-evaluation of already seen txs, with the waste of network and computation resources.

So this PR changes the current default based on the txpool implementation that is chosen:
* for Layered the default is to not forget on eviction (new behavior), since it is the implementation designed for public networks.
* for Sequencer the default is to forget on eviction (current behavior) since it is the implementation that is used in private/permissioned networks.

Moreover it is possible to overwrite the default behavior using the new experimental option `Xpeer-tracker-forget-evicted-txs`.

Another option has been added to customized the max size of the cache that keeps memory of the tx exchanged with each peers, so it can be tuned to adapt to increases in the pending txs number. The new option is `Xmax-tracked-seen-txs-per-peer` and by default is now `200_000` entries, double from the previous hard coded value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - New metric `besu_peers_peer_count_by_client` to report the count of peers by client name [#8515](https://github.com/hyperledger/besu/pull/8515)- Add Prague fork time for mainnet. [#8521](https://github.com/hyperledger/besu/pull/8521)
 - Add Prague fork time for mainnet. [#8521](https://github.com/hyperledger/besu/pull/8521)
 - Add block import tracer plugin provider, used during block import [#8524](https://github.com/hyperledger/besu/pull/8534)
+- Tune and make configurable the PeerTransactionTracker [#8537](https://github.com/hyperledger/besu/pull/8537)
 #### Dependencies
 ### Bug fixes
 - Fix for storage proof key - if the key is zero use `0x0` not `0x` [#8499](https://github.com/hyperledger/besu/pull/8499)
--- a/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
@@ -296,7 +296,8 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
         hidden = true,
         description =
             "Whether txs evicted due to the pool being full should be removed from peer tracker cache that checks for already known txs (default: false on layered and true on sequenced txpool)",
-        arity = "1")
+        arity = "0..1",
+        fallbackValue = "true")
     private Boolean peerTrackerForgetEvictedTxs;
   }
 
--- a/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
@@ -285,7 +285,7 @@ public class TransactionPoolOptions implements CLIOptions<TransactionPoolConfigu
         paramLabel = "<LONG>",
         hidden = true,
         description =
-            "The number of exchanged txs that are remembered with each peer, to avoid broadcasting duplicates (default: ${DEFAULT-VALUE})",
+            "The number of exchanged txs that are remembered with each peer, to minimize broadcasting duplicates (default: ${DEFAULT-VALUE})",
         arity = "1")
     private int maxTrackedSeenTxsPerPeer =
         TransactionPoolConfiguration.Unstable.DEFAULT_MAX_TRACKED_SEEN_TXS_PER_PEER;
--- a/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
+++ b/besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
@@ -39,6 +39,7 @@ import java.io.File;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import picocli.CommandLine;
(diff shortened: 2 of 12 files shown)


besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -277,6 +279,25 @@ static class Unstable {
...
+        arity = "1")
+    private int maxTrackedSeenTxsPerPeer =
+        TransactionPoolConfiguration.Unstable.DEFAULT_MAX_TRACKED_SEEN_TXS_PER_PEER;
+
+    @CommandLine.Option(
+        names = {PEER_TRACKER_FORGET_EVICTED_TXS_FLAG},
+        paramLabel = "<BOOLEAN>",
+        hidden = true,
+        description =
+            "If txs evicted due to the pool being full should be removed from peer tracker cache that check for already known txs (default: false on layered and true on sequenced)",
```
macfarla, line 298:
```suggestion
            "Whether txs evicted due to the pool being full should be removed from peer tracker cache that checks for already known txs (default: false on layered and true on sequenced txpool)",
```

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -277,6 +279,25 @@ static class Unstable {
         arity = "1")
     private Duration eth65TrxAnnouncedBufferingPeriod =
         TransactionPoolConfiguration.Unstable.ETH65_TRX_ANNOUNCED_BUFFERING_PERIOD;
+
+    @CommandLine.Option(
+        names = {MAX_TRACKED_SEEN_TXS_PER_PEER},
+        paramLabel = "<LONG>",
+        hidden = true,
+        description =
+            "The number of exchanged txs that are remembered with each peer, to avoid broadcasting duplicates (default: ${DEFAULT-VALUE})",
```
macfarla, line 288:
```suggestion
            "The number of exchanged txs that are remembered with each peer, to minimize broadcasting duplicates (default: ${DEFAULT-VALUE})",
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Add support for `eth_simulateV1`. [#8406](https://github.com/hyperledger/besu/pull/8406)
 - New metric `besu_peers_peer_count_by_client` to report the count of peers by client name [#8515](https://github.com/hyperledger/besu/pull/8515)- Add Prague fork time for mainnet. [#8521](https://github.com/hyperledger/besu/pull/8521)
 - Add Prague fork time for mainnet. [#8521](https://github.com/hyperledger/besu/pull/8521)
+- Tune and make configurable the PeerTransactionTracker [#8537](https://github.com/hyperledger/besu/pull/8537)
```
macfarla, line 10:
```suggestion
- Tune and make configurable the PeerTransactionTracker [#8537](https://github.com/hyperledger/besu/pull/8537) with `--Xmax-tracked-seen-txs-per-peer` and `--Xpeer-tracker-forget-evicted-txs`
```

macfarla (approved):

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -247,15 +248,16 @@ static class Sequenced {
...
-      new TransactionPoolOptions.Unstable();
+  private final Unstable unstableOptions = new Unstable();
 
   static class Unstable {
     private static final String TX_MESSAGE_KEEP_ALIVE_SEC_FLAG =
         "--Xincoming-tx-messages-keep-alive-seconds";
-
     private static final String ETH65_TX_ANNOUNCED_BUFFERING_PERIOD_FLAG =
         "--Xeth65-tx-announced-buffering-period-milliseconds";
+    private static final String MAX_TRACKED_SEEN_TXS_PER_PEER = "--Xmax-tracked-seen-txs-per-peer";
```
siladu, line 258: nit: not sure there's a convention for this but maybe could start "--Xpeer-..."` similar to other groups of config.

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
```diff
@@ -277,6 +279,25 @@ static class Unstable {
...
+        description =
+            "The number of exchanged txs that are remembered with each peer, to avoid broadcasting duplicates (default: ${DEFAULT-VALUE})",
+        arity = "1")
+    private int maxTrackedSeenTxsPerPeer =
+        TransactionPoolConfiguration.Unstable.DEFAULT_MAX_TRACKED_SEEN_TXS_PER_PEER;
+
+    @CommandLine.Option(
+        names = {PEER_TRACKER_FORGET_EVICTED_TXS_FLAG},
+        paramLabel = "<BOOLEAN>",
+        hidden = true,
```
siladu, line 296: Does this need a fallback value so `--Xpeer-tracker-forget-evicted-txs` will work without "=true"?

besu/src/test/java/org/hyperledger/besu/cli/options/TransactionPoolOptionsTest.java
```diff
@@ -454,6 +455,40 @@ public void minScoreNonByteValueReturnError() {
...
+        LAYERED.name().toLowerCase(Locale.ROOT));
+  }
+
+  @Test
+  public void explicitlyEnablingForgetEvictedTxsWithLayered() {
+    internalTestSuccess(
+        config -> assertThat(config.getUnstable().getPeerTrackerForgetEvictedTxs()).isTrue(),
+        "--tx-pool",
+        LAYERED.name().toLowerCase(Locale.ROOT),
+        "--Xpeer-tracker-forget-evicted-txs=true");
```
siladu, line 489: Maybe a another test for `--Xpeer-tracker-forget-evicted-txs` using the fallBack value if we want to support that

siladu (approved):

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
fab-10: Good point, the important thing is that _fallback value_ does not ct as a default, because in this case the default is chosen at runtime depending on the txpool implementation

besu/src/main/java/org/hyperledger/besu/cli/options/TransactionPoolOptions.java
fab-10: Added the fallback value

besu/src/test/java/org/hyperledger/besu/cli/options/TransactionPoolOptionsTest.java
fab-10: done

PR #8556, opened 2025-04-16
plugin-api: TrieLogAccumulator add lookups for account and storage

garyschulte:
## PR description
Expose methods to lookup account and storage info in the accumulator plugin-api

Motivation: allow TrieLogFactory plugin to 'decorate' the accumulator maps prior to generating the trielog.  

In this instance we want Linea besu to be able to add state elements to the generated trielog for shomei

- [X] spotless: `./gradlew spotlessApply`
- [X] unit tests: `./gradlew build`

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/common/worldview/accumulator/PathBasedWorldStateUpdateAccumulator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/common/worldview/accumulator/PathBasedWorldStateUpdateAccumulator.java
@@ -250,6 +250,15 @@ public abstract class PathBasedWorldStateUpdateAccumulator<ACCOUNT extends PathB
     return super.getAccount(address);
   }
 
+  /**
+   * Note: PathBasedAccount implements both MutableAccount and AccountValue, so this cast is safe
+   * even though the compiler can't verify it.
+   */
+  @Override
+  public <T extends AccountValue> T getAccountValue(final Address address) {
+    return (T) getAccount(address);
+  }
+
   @Override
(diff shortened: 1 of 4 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/common/worldview/accumulator/PathBasedWorldStateUpdateAccumulator.java
```diff
@@ -250,6 +250,15 @@ public MutableAccount getAccount(final Address address) {
     return super.getAccount(address);
   }
 
+  /**
+   * Note: PathBasedAccount implements both MutableAccount and AccountValue, so this cast is safe
+   * even though the compiler can't verify it.
+   */
+  @Override
+  public <T extends AccountValue> T getAccountValue(final Address address) {
```
matkt, line 258: I am personally not a fan of adding another GetAccount method to the accumulator. I find interface already complex and having several GetAccount methods will not help. Can't we call getaccount and cast at the API plugin part?

garyschulte: I am not certain that account lookup is necessary, I was operating under the assumption that ShomeiTrieLogConverter would fetch the current account state if we include a null for the account prior and updated.  

However, if we do need to do lookups and want to cast the plugin-api accumulator to an explicit type, we will have to pull in all of :ethereum:core just to get the class we want to cast to.  Not really a viable plugin design since it would be coupled to an explicit build of besu.

Right now I am pursuing having shomei do the lookup, so hopefully this whole interface change will not be necessary.

garyschulte:
Extending this interface is no longer necessary

PR #8562, opened 2025-04-17
Let Non validators gossip QBFT messages

pullurib:
## PR description

Today QBFT traffic is only gossiped by other validators

image

A chain configured without direct connectivity between validators won’t produce new blocks

image

Gossiping between non-validators allows more control over the way nodes are deployed and interconnected

image

- Provides more flexibility for different Besu node topologies
- Enables non validators to gossip QBFT messages so that validators peered through non validators can participate in consensus. 
- This would let the edge nodes in a network to be non validators . 
- Code changes 
  - Moved Peerconnection tracking and all peer multicast to Peers class 
  - Most of file changes are refactoring ValidatorMulticaster to PeerMulticaster


jframe:
@pullurib I haven't had a detailed look at this yet. Will the non-validator nodes validate QBFT messages before gossiping them? I am concerned that if the messages aren't validated this could cause DOS attack with invalid messages flooding the network.

pullurib:
> @pullurib I haven't had a detailed look at this yet. Will the non-validator nodes validate QBFT messages before gossiping them? I am concerned that if the messages aren't validated this could cause DOS attack with invalid messages flooding the network.

@jframe yes , all the validation as a validator would is done , gossiped as a validator would and then handled only if the local node is validator

app/src/main/java/org/hyperledger/besu/controller/QbftBesuControllerBuilder.java
```diff
@@ -234,7 +234,8 @@ protected MiningCoordinator createMiningCoordinator(
 
     // NOTE: peers should not be used for accessing the network as it does not enforce the
     // "only send once" filter applied by the UniqueMessageMulticaster.
-    peers = new ValidatorPeers(validatorProvider, Istanbul100SubProtocol.NAME);
+    // Use Peers instead of ValidatorPeers to allow message relay by non-validators
+    peers = new Peers(Istanbul100SubProtocol.NAME);
```
jframe, line 239: My main concern is there will be an increase in duplicate messages and this could fill the event queue. I think might need to do de-duplication earlier before the message is added to the event queue.

pullurib: de-duplication is done at the point where the messages are sent out. That way, as long as all nodes behave, there won't be cycles and flooding of event queue. If there's a faulty node, I don't think there's any guard as of now to protect against the event queue filling up. There's no deny listing for BFT subprotocols. Any message from any node (validator and non validator) must be handled.

PR #8567, opened 2025-04-18
Fail fast on missing system contract addresses

jflo:
Removes hardcoded contract addresses, now they must be provided in genesis files.

Caveat: EVMTool tests have no facility for specifying these addresses, and since the reference tests assume mainnet locations for all system contracrts, the StubGenesisConfigurationOptions are defaulted to those same mainnet addresses.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/requests/RequestContractAddresses.java
```diff
@@ -17,19 +17,22 @@
...
-
-  public RequestContractAddresses(
+  /*
+    public static final Address DEFAULT_WITHDRAWAL_REQUEST_CONTRACT_ADDRESS =
+        Address.fromHexString("0x00000961ef480eb55e80d19ad83579a64c007002");
+    public static final Address DEFAULT_CONSOLIDATION_REQUEST_CONTRACT_ADDRESS =
+        Address.fromHexString("0x0000bbddc7ce488642fb579f8b00f3a590007251");
+    public static final Address DEFAULT_DEPOSIT_CONTRACT_ADDRESS =
+        Address.fromHexString("0x00000000219ab540356cbb839cbe05303d7705fa");
+  */
```
Gabriel-Trintinalia, line 34: Can be removed

besu/src/test/java/org/hyperledger/besu/controller/AbstractBftBesuControllerBuilderTest.java
```diff
@@ -106,6 +107,18 @@ public void setup() throws JsonProcessingException {
...
+        .thenReturn(
+            Optional.of(Address.fromHexString("0x00000000219ab540356cbb839cbe05303d7705fa")));
+    lenient()
+        .when(genesisConfigOptions.getWithdrawalRequestContractAddress())
+        .thenReturn(
+            Optional.of(Address.fromHexString("0x00000961ef480eb55e80d19ad83579a64c007002")));
+    lenient()
+        .when(genesisConfigOptions.getConsolidationRequestContractAddress())
+        .thenReturn(
+            Optional.of(Address.fromHexString("0x0000bbddc7ce488642fb579f8b00f3a590007251")));
```
garyschulte, line 121: it feels wrong to shoe-horn in mainnet system contract addresses for a consensus mechanism test that shouldn't otherwise ever have these contracts 🤔 .  What does the deposit, withdrawal, and consolidation system contract mean in the context of a *Bft consensus chain?

consensus/common/src/test/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilderTest.java
```diff
@@ -45,13 +45,27 @@
...
+  void setup() {
+    when(genesisConfig.getDepositContractAddress())
+        .thenReturn(
+            Optional.of(Address.fromHexString("0x00000000219ab540356cbb839cbe05303d7705fa")));
+    when(genesisConfig.getWithdrawalRequestContractAddress())
+        .thenReturn(
+            Optional.of(Address.fromHexString("0x00000961ef480eb55e80d19ad83579a64c007002")));
+    when(genesisConfig.getConsolidationRequestContractAddress())
+        .thenReturn(
+            Optional.of(Address.fromHexString("0x0000bbddc7ce488642fb579f8b00f3a590007251")));
```
garyschulte, line 66: same comment as above.  These should be fine to be absent as long as we are not using a mainnet genesis config AFAIK

config/src/main/java/org/hyperledger/besu/config/StubGenesisConfigOptions.java
```diff
@@ -467,17 +467,17 @@ public List<Long> getForkBlockTimestamps() {
 
   @Override
   public Optional<Address> getWithdrawalRequestContractAddress() {
-    return Optional.empty();
+    return Optional.of(Address.fromHexString("0x00000961ef480eb55e80d19ad83579a64c007002"));
```
garyschulte, line 470: what do you think about a MainnetGenesisConfigOptions subclass (or interface with default implementations) for these methods?  StubGenesis seems to be used everywhere, not just for mainnet PoS networks

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetProtocolSpecs.java
```diff
@@ -841,88 +843,101 @@ static ProtocolSpecBuilder pragueDefinition(
...
+            .name("Prague");
+    try {
+      RequestContractAddresses requestContractAddresses =
+          RequestContractAddresses.fromGenesis(genesisConfigOptions);
+
+      pragueSpecBuilder.requestProcessorCoordinator(
+          pragueRequestsProcessors(requestContractAddresses));
+    } catch (NoSuchElementException nsee) {
+      LOG.warn("Prague definitions require system contract addresses in genesis");
+      throw nsee;
```
Gabriel-Trintinalia, line 937: Can we add a test for this?

consensus/common/src/test/java/org/hyperledger/besu/consensus/common/bft/BaseBftProtocolScheduleBuilderTest.java
jflo: Yes, same reason as above- not actually needed since conditionals were added.

besu/src/test/java/org/hyperledger/besu/controller/AbstractBftBesuControllerBuilderTest.java
jflo: Ah yes, and this is no longer needed since I added the conditionals around adding the milestones to the schedule. 

Removed.

config/src/main/java/org/hyperledger/besu/config/StubGenesisConfigOptions.java
jflo: I like the idea, but i'm not sure where to draw the boundaries - would a stub genesis only contain frontier features? everything up to the DAO fork? everything up to The Merge?

Perhaps only things used by all PoW, PoS, and PoA networks?

garyschulte (approved):
🚢

PR #8583, opened 2025-04-28
fix: add `NoBlobRule` to Shanghai block header validator

Gabriel-Trintinalia:
## PR description

Adds `NoBlobRule` to the Shanghai block header validator. Used mainly for RLP Block Importer now.


pinges (approved):
LGTM

PR #8589, opened 2025-05-05
Change Maven artifactIds to avoid possible collisions with other libs

fab-10:
## PR description

Currently Besu artifacts are published to Maven, with a `groupId` = `org.hyperledger.besu[.internal]` and the `artifactId` is just the name of the subproject, for example: `p2p`, `core` or plugin-api`.

That works fine for defining a dependency on any Besu lib using the Maven coordinates, since the couple: `groupId` + `artifactId` are unique, but there is a potential issue when packaging the project, since very generic artifactIds, like `core`, `util` and so on, could collide with other project dependencies using the same artifactId and the same calver, since [the Maven repository layout](https://maven.apache.org/repository/layout.html) states that the jar name is only the `artifactId` + `version`, so the `groupId` could not be used to disambiguate when collecting all the jars needed by the app, as the Gradle distribution plugin does for example.

The solution proposed in this PR, is to change the `artifactId` of Besu libs to be automatically set to the _project path_, that is the concatenation of all the project names from the root to the leaf. 
This way: `core` becomes `besu-ethereum-core` or `plugin-api` become `besu-plugin-api`.
This is the approach taken by most common libs like: Vert.x, Netty, etc...

This is a breaking change, since Maven coordinates change and plugins will have to update their dependencies specification.

Known projects that are using Besu dependencies and need to be updated, where possible I will open a PR to update the coordinates and notify the team:
* Teku 
* Web3j
* Hedera
* [Paladin](https://github.com/LF-Decentralized-Trust-labs/paladin/blob/4ad4f21cc8b514ba1e1f2ec640ff9c5431d30b4e/domains/pente/build.gradle#L115) 


garyschulte (approved):
This LGTM - it will help with forks like op-besu as well.  

It is probably worth checking to see what public projects might have dependencies on the current artifactId's.  Cursory glance seems like at least a few web3j artifacts have dependencies, shomei, plugins, etc.  Definitely worth a warning in the release notes.

fab-10:
@garyschulte after some research I added to the description the list of known public projects that are using Besu dependencies, and will notify them about the change

PR #8592, opened 2025-05-07
removes PoWAlgorithm and its references

jflo:
The first of a series of PRs to remove PoW from Besu.


besu/src/main/java/org/hyperledger/besu/cli/options/MiningOptions.java
```diff
@@ -260,32 +259,6 @@ public void validate(
...
-      // Check that mining options are able to work
-      CommandLineUtils.checkOptionDependencies(
-          logger,
-          commandLine,
-          "--miner-enabled",
-          !isMiningEnabled,
-          asList(
-              "--miner-stratum-enabled",
-              "--Xminer-remote-sealers-limit",
-              "--Xminer-remote-sealers-hashrate-ttl"));
```
garyschulte, line 286: this isn't just PoW, this is clique and PoA right?  miner has never been the right name, but IIRC PoA algorithms still need to use miner-enabled and this check I believe is still valid for sealer opts

garyschulte (commented):
bold start.  It isn't clear what we are signaling with this though.  The end of all PoW support in besu, the plugin-izing of it, etc.  I would guess this breaks ETC support?

Another and probably less controversial target would be the stratum mining stuff.

jflo:
> bold start. It isn't clear what we are signaling with this though. The end of all PoW support in besu, the plugin-izing of it, etc. I would guess this breaks ETC support?
> 
> Another and probably less controversial target would be the stratum mining stuff.

yes, it would break ETC support. See this pr for more mining removal:  https://github.com/hyperledger/besu/pull/8454/files

PR #8620, opened 2025-05-10
Reintroduce tx index pruning with performance fix

siladu:
setInstance for SignatureAlgorithmFactory

Replace PR for https://github.com/hyperledger/besu/pull/8590


plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueStorage.java
```diff
@@ -225,7 +225,8 @@ private ColumnFamilyDescriptor createColumnDescriptor(
             .setTtl(0)
             .setCompressionType(CompressionType.LZ4_COMPRESSION)
             .setTableFormatConfig(basedTableConfig)
-            .setLevelCompactionDynamicLevelBytes(dynamicLevelBytes);
+            .setLevelCompactionDynamicLevelBytes(dynamicLevelBytes)
+            .setDisableAutoCompactions(true);
```
ahamlat, line 229: Do you want to disable dynamic compaction on all the column families ? How compaction will happen in this case ?

PR #8629, opened 2025-05-14
Make gas estimation strict by default

fab-10:
## PR description

Second PR of the series that improves gas estimation.

Besu implementation of `eth_gasEstimation` accepts the non standard parameters `strict`, that by default is `false`.
This parameters tells Besu that when processing the tx to estimate the gas used, the sender is allowed to exceeds his balance, and this is achieved setting to `0` all the gas price related field in the tx and in the block header.
This approach is not the standard, and could cause wrong gas estimations for some txs, so this PR set the default value of the `strict` parameter to `true`.

It is still possible to force the previous behavior, explicitly passing the `strict` parameter in the request, set to `false`.

Since `strict` mode does not alter gas pricing parameters, there is the need to calculate the gas limit upper bound that the sender is able to afford according to his balance, and use that upper bound in case it is lower than the block gas limit, when starting the binary search, to avoid tx upfront cost exceed balance errors.

Always in `strict` mode, since the gas pricing parameters are absent, then we need to permit underpriced tx, to skip checks on baseFee and blobBaseFee.

Comparing this PR with current main branch, using https://github.com/Consensys/compate-estimate-gas-plugin, there are less under estimations:

BesuPR8629 3399
BesuNightly 3426


macfarla (approved):
LGTM maybe an update to the docs required

Issue #8602, opened 2025-05-08
EIP-7823 MODEXP upper bounds


jflo:
[EIP-7823](https://eips.ethereum.org/EIPS/eip-7823): Set upper bounds for MODEXP

PR #8632, opened 2025-05-15
EIP-7823 - Modexp upper bounds

siladu:
## PR description

> We introduce an upper bound to the inputs of the precompile, each of the length inputs (length_of_BASE, length_of_EXPONENT and length_of_MODULUS) MUST be less than or equal to 8192 bits (1024 bytes).

> If any of these inputs are larger than the limit, the precompile execution stops, returns an error, and consumes all gas.

## Fixed Issue(s)

Fixes https://github.com/hyperledger/besu/issues/8602

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
@@ -99,10 +99,21 @@ public class BigIntegerModularExponentiationPrecompiledContract
   @Override
   public PrecompileContractResult computePrecompile(
       final Bytes input, @Nonnull final MessageFrame messageFrame) {
+    // https://eips.ethereum.org/EIPS/eip-7823
+    // We introduce an upper bound to the inputs of the precompile,
+    // each of the length inputs (length_of_BASE, length_of_EXPONENT and length_of_MODULUS)
+    // MUST be less than or equal to 8192 bits (1024 bytes).
+    final long length_of_BASE = baseLength(input);
+    final long length_of_EXPONENT = exponentLength(input);
+    final long length_of_MODULUS = modulusLength(input);
+    if (length_of_BASE > 8192 || length_of_EXPONENT > 8192 || length_of_MODULUS > 8192) {
+      return PrecompileContractResult.halt(
+          null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
+    }
     if (useNative) {
-      return computeNative(input);
+      return computeNative(input, length_of_MODULUS);
     } else {
-      return computeDefault(input);
+      return computeDefault(input, length_of_BASE, length_of_EXPONENT, length_of_MODULUS);
     }
   }
 
@@ -113,10 +124,14 @@ public class BigIntegerModularExponentiationPrecompiledContract
    * @return the precompile contract result
    */
   @Nonnull
-  public PrecompileContractResult computeDefault(final Bytes input) {
-    final int baseLength = clampedToInt(baseLength(input));
-    final int exponentLength = clampedToInt(exponentLength(input));
-    final int modulusLength = clampedToInt(modulusLength(input));
+  public PrecompileContractResult computeDefault(
+      final Bytes input,
+      final long length_of_BASE,
+      final long length_of_EXPONENT,
+      final long length_of_MODULUS) {
+    final int baseLength = clampedToInt(length_of_BASE);
+    final int exponentLength = clampedToInt(length_of_EXPONENT);
+    final int modulusLength = clampedToInt(length_of_MODULUS);
     // If baseLength and modulusLength are zero
     // we could have a massively overflowing exp because it wouldn't have been filtered out at the
     // gas cost phase
@@ -247,8 +262,9 @@ public class BigIntegerModularExponentiationPrecompiledContract
    * @param input the input
    * @return the precompile contract result
    */
-  public PrecompileContractResult computeNative(final @Nonnull Bytes input) {
-    final int modulusLength = clampedToInt(modulusLength(input));
+  public PrecompileContractResult computeNative(
+      final @Nonnull Bytes input, final long length_of_MODULUS) {
+    final int modulusLength = clampedToInt(length_of_MODULUS);
     final IntByReference o_len = new IntByReference(modulusLength);
 
     final byte[] result = new byte[modulusLength];


evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
```diff
@@ -52,13 +52,42 @@ public class BigIntegerModularExponentiationPrecompiledContract
...
+
+  /**
+   * Create the Osaka BigIntegerModularExponentiationPrecompiledContract. Introduced in EIP-7823
+   *
+   * @param gasCalculator the gas calculator
+   * @return the BigIntegerModularExponentiationPrecompiledContract
+   */
+  public static BigIntegerModularExponentiationPrecompiledContract osaka(
+      final GasCalculator gasCalculator) {
+    return new BigIntegerModularExponentiationPrecompiledContract(gasCalculator, OSAKA_UPPER_BOUND);
```
lu-pinto, line 90: This looks like an usual design - IMO this static logic does not belong here but can be accessed through `MainnetPrecompiledContracts::byzantium` and `MainnetPrecompiledContracts::osaka`. The duty to configure these things should be on the caller (configuration) not the instance itself.
I would rather leave a default constructor that uses the new Osaka upper bound and a new constructor to set the upper bound.

evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
```diff
@@ -99,24 +128,35 @@ public long gasRequirement(final Bytes input) {
...
+    // each of the length inputs (length_of_BASE, length_of_EXPONENT and length_of_MODULUS)
+    // MUST be less than or equal to 8192 bits (1024 bytes).
+    final long length_of_BASE = baseLength(input);
+    final long length_of_EXPONENT = exponentLength(input);
+    final long length_of_MODULUS = modulusLength(input);
+    if (length_of_BASE > upperBound
+        || length_of_EXPONENT > upperBound
+        || length_of_MODULUS > upperBound) {
+      return PrecompileContractResult.halt(
+          null, Optional.of(ExceptionalHaltReason.PRECOMPILE_ERROR));
```
lu-pinto, line 142: Need to have a unit test for this

evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
```diff
@@ -99,24 +128,35 @@ public long gasRequirement(final Bytes input) {
...
   @Nonnull
-  public PrecompileContractResult computeDefault(final Bytes input) {
-    final int baseLength = clampedToInt(baseLength(input));
-    final int exponentLength = clampedToInt(exponentLength(input));
-    final int modulusLength = clampedToInt(modulusLength(input));
+  private PrecompileContractResult computeDefault(
+      final Bytes input,
+      final long length_of_BASE,
+      final long length_of_EXPONENT,
+      final long length_of_MODULUS) {
```
lu-pinto, line 156: nit: why using underscores in variable names

evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
```diff
@@ -241,14 +281,9 @@ private static long square(final long n) {
...
-   * Compute native precompile contract.
-   *
-   * @param input the input
-   * @return the precompile contract result
-   */
-  public PrecompileContractResult computeNative(final @Nonnull Bytes input) {
-    final int modulusLength = clampedToInt(modulusLength(input));
+  private PrecompileContractResult computeNative(
+      final @Nonnull Bytes input, final long length_of_MODULUS) {
+    final int modulusLength = clampedToInt(length_of_MODULUS);
```
lu-pinto, line 286: nit: same as https://github.com/hyperledger/besu/pull/8632/files#r2099915626

evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
lu-pinto: Let me know what you think https://github.com/hyperledger/besu/commit/9b34bf9366d86db4736fc7f25acd4fcf9d878532

siladu: Disagree that this is an usual design, this is a static factory used everywhere in java projects including Besu, and can be found in textbooks. Static factories typically live on the object being constructed.

In particular, this follows the design in place for e.g `AltBN128AddPrecompiledContract.byzantium`

> Let me know what you think https://github.com/hyperledger/besu/commit/9b34bf9366d86db4736fc7f25acd4fcf9d878532

No objections in general, but I think this look-up style design would be unconventional for Besu.
If you look through MainnetProtocolSpecs, most fork configuration is done by wiring in a fork-specific version of a class, which extends the previous one, e.g. GasCalculator. 

The difference here is we wire in a static factory instead of a whole new version of the class, since it's only one value that is changed.
If there is a performance issue with the extra check though, wiring in e.g. OsakaBigIntegerModularExponentiationPrecompiledContract might be more code but more straight forward.

evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
siladu: This is to match the spec - controversial perhaps, but I find it's really useful if the spec and code values match

evm/src/main/java/org/hyperledger/besu/evm/precompile/BigIntegerModularExponentiationPrecompiledContract.java
siladu: yep, this was first draft for the devnet, planning to productionise next

siladu:
Using Benchmark.java, I see no significant difference between main and this branch. The most significant are highlight in bold but it goes in both directions so maybe just noise...
Screenshot 2025-05-29 at 6 57 51 pm

lu-pinto:
In order to avoid running this off my laptop as it is not great for benchmarking due to other things running, I used a cloud instance. Results for Prague:
image

And results for Osaka:
image

There is some variance in one or 2 test cases but I think it's just noise, test case is too trivial.

siladu:
> Confirm no (performance?) issue introducing explicit upperBound test to Byzantium (Long.MAX_VALUE)

It might just be noise at these levels, but I measure a small impact with profiling, 2% performance impact in the worst case. I think it's negligible and we will be on Fusaka soon enough anyway.

[modexp-benchmark-subcommand-profiling.tar.gz](https://github.com/user-attachments/files/20589332/modexp-benchmark-subcommand-profiling.tar.gz)

siladu:
> Regression test (mainnet sync versus control)

No performance difference spotted on mainnet, profiling for 10 minutes didn't show modexp at all, so must be rarely used.

lu-pinto (approved):

PR #8639, opened 2025-05-16
Dont touch state for CALLs raising a static exception

OlivierBBB:
## PR description

This PR does two things:
- not compute the gas cost of `LOGX` that raise a static exception
- not compute the gas cost of `CALL` that raises a static exception

**Note.** Computing the cost of a value-bearing `CALL` requires a state access: one must determine whether the target address exists in the state. In the code this state access was implicitly performed during `cost(frame, true)`. If the `CALL` raises a static exception this state access is not required; it also pollutes the accumulator. In the fix we therefore remove this cost calculation and replace it with 0. We also remove the cost calculation for `LOGX` operations that raise a static exception.


matkt (approved):
LGTM

lu-pinto (approved):

PR #8658, opened 2025-05-21
refactor: add blob schedule to protocol spec builder

Gabriel-Trintinalia:
## Description

This PR refactors the `ProtocolSpecBuilder` to explicitly require `BlobSchedule` as a constructor argument. By default it is initiated with `NoBlobSchedule`

It also ensures that `FeeMarket`, `GasCalculator`, `GasLimitCalculatorBuilder`, and `BlockHeaderValidatorBuilder` are derived based on the provided BlobSchedule.

### Key Changes

- Made BlobSchedule a required input for ProtocolScheduleBuilder.
- Updated the builder to explicitly accept `FeeMarket`, `GasCalculator`, `GasLimitCalculatorBuilder`, and `BlockHeaderValidatorBuilder` based on the `BlobSchedule`.

This is a non-functional refactor with no changes to existing protocol behavior at runtime.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpecBuilder.java
```diff
@@ -503,6 +515,19 @@ BlockValidator apply(
...
 
+  public interface FeeMarketBuilder {
+    FeeMarket apply(BlobSchedule blobSchedule);
+  }
+
+  public interface BlockHeaderValidatorBuilder {
+    BlockHeaderValidator.Builder apply(FeeMarket feeMarket, GasCalculator gasCalculator);
+  }
+
+  public interface GasLimitCalculatorBuilder {
```
Matilda-Clerke, line 526: Consider adding `@FunctionalInterface` to indicate intended usage of these interfaces

Matilda-Clerke (commented):
I didn't realise that BlobSchedule was being moved out of BlobScheduleOptions and wasn't a new class. Up to you if you want to modify it with my suggestions.

siladu (approved):
Love it. Plus you got rid of one of my TODOs so ⭐

PR #8664, opened 2025-05-22
refactor: remove `PragueFeeMarket`

Gabriel-Trintinalia:
## Description  

The `PragueFeeMarket` is functionally identical to `BlobFeeMarket` (formerly `CancunFeeMarket`), with the sole distinction being the `baseFeeUpdateFraction` parameter, which is determined by the `BlobSchedule`.

### Updates  
- Renamed `CancunFeeMarket` to `BlobFeeMarket`.  
- Removed the `PragueFeeMarket` class, as it is redundant and does not introduce any unique changes.  
  - Removed `FeeMarket` from the `PragueDefinition`, as it is unnecessary. The builder will pass the correct `BlobSchedule` to the inherited `FeeMarket` from the `BlobFeeMarket` definition.  
  - Merged `PragueFeeMarketTest` and `CancunFeeMarketTest` into a single test class, `BlobFeeMarketTest`
- Removed the overloaded constructor from `BlobFeeMarket`; it will now always expect a `baseFeeUpdateFraction`.  
  - Introduced a new `cancunDefault` method in `FeeMarket` to create Blob fee markets with default Cancun `BlobSchedule` parameters, primarily for use in test cases.  
- Renamed the `implementsDataFee` method to `implementsBlobFee`.


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/feemarket/FeeMarketBuilderFactory.java
```diff
@@ -0,0 +1,60 @@
...
+import org.hyperledger.besu.ethereum.mainnet.ProtocolSpecBuilder;
+
+import java.util.Optional;
+
+/**
+ * Factory for creating {@link FeeMarket} instances used in protocol specification configuration.
+ */
+public class FeeMarketBuilderFactory {
+  @FunctionalInterface
+  public interface FeeMarketBuilder {
```
siladu, line 28: nit: It's little confusing having this as well as ProtocolSpecBuilder.FeeMarketBuilder sharing the same name. 
IIUC, they are working together just at two different layers of abstraction.

The outer FeeMarketBuilder accepts only blobSchedule and returns FeeMarket.
The inner represents one case where we delegate to the "inner" builder which accepts more params and return FeeMarket.

I feel like both are more accurately "Factorys" rather than "Builders", but since the outer is following an existing convention I would suggest we change this one to be: `FeeMarketFactory`
Perhaps this interface should also live where the static factories live, i.e. in FeeMarket?

siladu (approved):
Nice refactor 👍 
Just a naming/location nit

pinges (approved):
just a comment about passing in the london block number into the london() method.

siladu (approved):
LGTM

PR #8687, opened 2025-05-23
Replace `@Nonnull` annotation with `@NotNull`

fab-10:
## PR description

`@Nonnull` annotation is part of JSR-305 specification, but the JSR-305 specification is no longer actively maintained, which may limit its adoption in modern projects.

The proposed PR replace it with '@NotNull` from Jakarta validation.

This change is also preliminary to following dependency updates.

### DIFF
--- a/platform/build.gradle
+++ b/platform/build.gradle
@@ -117,6 +117,8 @@ dependencies {
     api 'io.consensys.tuweni:tuweni-toml:2.7.0'
     api 'io.consensys.tuweni:tuweni-units:2.7.0'
 
+    api 'jakarta.validation:jakarta.validation-api:3.0.2'
+
     api 'javax.inject:javax.inject:1'
 
     api 'net.java.dev.jna:jna:5.16.0'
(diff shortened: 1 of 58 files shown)


garyschulte (commented):
Did you consider [checker framework](https://github.com/typetools/checker-framework)?  it seems like it would be a more robust choice, more actively maintained and doesnt have Java EE connotations.

fab-10:
I followed [this guide](https://medium.com/@mesfandiari77/understanding-notnull-vs-nonnull-in-java-a-practical-guide-to-null-safety-5e179e918f37), that does not mention the checker framework, how it could be used as replacement for `@NotNull`

fab-10:
Checker framework also understand Jakarta annotation https://checkerframework.org/manual/#nullness-related-work

It seems they are not solving the same problem, and I just wanted to find a replacement, since `@Nonnull` seems dead, so I am fine keeping the status quo, I didn't want to invest too much in this at the moment.

siladu:
Why not `javax.validation.constraints.NotNull` to avoid third party?

siladu (approved):
LGTM

platform/build.gradle
```diff
@@ -117,6 +117,8 @@ dependencies {
     api 'io.consensys.tuweni:tuweni-toml:2.7.0'
     api 'io.consensys.tuweni:tuweni-units:2.7.0'
 
+    api 'jakarta.validation:jakarta.validation-api:3.0.2'
```
Matilda-Clerke, line 120: Looks like jakarta validation 3.1 was released in April 2024. Any reason we wouldn't want to use it instead?

Matilda-Clerke (approved):

platform/build.gradle
fab-10: The scope of this change was to replace the annotation, in preparation for another PR that will update dependencies, and this was the version that we were already using since it is pulled by `org.hibernate.validator:hibernate-validator:8.0.2.Final`

PR #8695, opened 2025-05-26
chore: increase the default for blob prioritized transactions

Gabriel-Trintinalia:
## PR description

Increase the default for blob prioritized transactions


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/transactions/TransactionPoolConfiguration.java
```diff
@@ -87,7 +87,7 @@ enum Implementation {
   long DEFAULT_PENDING_TRANSACTIONS_LAYER_MAX_CAPACITY_BYTES = 25_000_000L;
   int DEFAULT_MAX_PRIORITIZED_TRANSACTIONS = 4000;
   EnumMap<TransactionType, Integer> DEFAULT_MAX_PRIORITIZED_TRANSACTIONS_BY_TYPE =
-      new EnumMap<>(Map.of(TransactionType.BLOB, 9));
+      new EnumMap<>(Map.of(TransactionType.BLOB, 20));
```
fab-10, line 90: Why 20?

Gabriel-Trintinalia: That is the max being tested on fusaka-devnet-0

fab-10: That's fine, just wanted to know where the number came from

fab-10 (approved):

PR #8701, opened 2025-05-27
Enable Gradle dependency resolution consistency

fab-10:
## PR description

See 
https://docs.gradle.org/current/userguide/dependency_resolution_consistency.html for why this is an improvement


fab-10:
The verification metadata does not mean that we are using different versions of the same artifacts, but that during the dependency resolution Gradle had to touch that artifact version for some reasons, probably because a 3rd party dependency declares it, before it decides with is the right version to use. 
The actual version that we are using is `2.24.3` that is the one declared in our platform, while for reference `2.23.1` is declared by `tech.pegasys.discovery:discovery`.
So Gradle now needs to fetch these additional dependencies, even if they are not used, idk why because this is an implementation detail.

PR #8702, opened 2025-05-27
Cache feature for snapshot database

matkt:
## PR description

This PR refactors the DB snapshot (thanks to @garyschulte's commit) to remove the use of transactions, as a snapshot is immutable by nature.
Additionally, it introduces an optional cache layer at the snapshot level to optimize repeated reads.

Tests : 
 - Checked block building on a sepolia validator
 - Verified oldest_snapshot_seqno to be sure old snapshot are correctly closed
 - Tried calling eth_call in order to verify the performance with the snapshot cache


plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueSnapshot.java
```diff
@@ -62,24 +74,72 @@ public class RocksDBColumnarKeyValueSnapshot
...
+
+  private Optional<byte[]> getFromCacheOrRead(
+      final byte[] segmentId,
+      final byte[] key,
+      final ColumnFamilyHandle handle,
+      final Cache<Bytes, Optional<byte[]>> cache)
+      throws RocksDBException {
+    final Bytes cacheKey = makeCacheKey(segmentId, key);
+    Optional<byte[]> cached = cache.getIfPresent(cacheKey);
+    //noinspection OptionalAssignedToNull
```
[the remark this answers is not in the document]
matkt: it's because I'm checking if optional == null, but it's something we want to do, I will move it at the beginning of the class wuth SuppressWarnings

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulator.java
```diff
@@ -289,7 +290,9 @@ public <U> Optional<U> process(
     try (final MutableWorldState ws = getWorldState(header)) {
 
       WorldUpdater updater = getEffectiveWorldStateUpdater(ws);
-
+      if (ws instanceof BonsaiWorldState bonsaiWorldState) {
+        bonsaiWorldState.disableCacheMerkleTrieLoader();
```
garyschulte, line 294: 👍

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueSnapshot.java
```diff
@@ -62,24 +79,71 @@ public class RocksDBColumnarKeyValueSnapshot
...
   }
 
   @Override
   public Optional<byte[]> get(final SegmentIdentifier segment, final byte[] key)
       throws StorageException {
     throwIfClosed();
-    return snapTx.get(segment, key);
+    try (final OperationTimer.TimingContext ignored = metrics.getReadLatency().startTimer()) {
+      final ColumnFamilyHandle handle = columnFamilyMapper.apply(segment);
+      if (isReadCacheEnabledForSnapshots && segment.isEligibleToHighSpecFlag()) {
```
garyschulte, line 108: requiring high spec would includes blockchain queries (which afaik should only be the corresponding body and header for this snapshot), but omits code.  It might make sense to remove this condition entirely, since snapshots should only be used to access rpc-relevant things. the largest values are going to be block bodies, and they are not omitted by this check.  

I can see how we might end up with multiple copies of the same code cached for each snapshot.  ambivalent about this.  it might be better to have a non-snapshot specific code cache, since the code hash key is deterministic.  

having column family-specific read caches is probably a better approach in general since there are different update patterns for each.  We can make better use of shared caches for several segments like blockchain and code

garyschulte (commented):
LGTM, I think we would do better to have segment specific caches though so we can leverage different update patterns

plugins/rocksdb/src/main/java/org/hyperledger/besu/plugin/services/storage/rocksdb/segmented/RocksDBColumnarKeyValueSnapshot.java
ahamlat: >  it might be better to have a non-snapshot specific code cache, since the code hash key is deterministic.

Not sure to understand this comment. I agree with your comment related to blocks but in this case, it is not an issue, block data calls doesn't go through this code as far as know, I could be wrong. The idea was to exclude caching the result of get calls related to code. We don't want to cache it at all here.

matkt:
> re-enabling auto-merge after the failed metrics CLI passed locally. I still think we could do smarter caching here though.

yes we can try to do a more optimized version if we really want to use this feature in mainnet or by default

Issue #8475, opened 2025-03-27
Eth/69 - Use earliestBlock and latestBlock of Status Message to better find suitable peers


Gabriel-Trintinalia:
Every time we connect to a peer during sync, we request its head to confirm that peer has the block we need and disconnect if it does not. 
With eth/69 we can skip this since the block range is announced in the status message.

This issue should design and implement a solution/strategy using the new peer information and replace the current strategy.

Gabriel-Trintinalia:
https://github.com/hyperledger/besu/pull/8725

PR #8714, opened 2025-05-29
fix(eth/69): validate block range message

Gabriel-Trintinalia:
## PR description
Disconnect peers if the earliest announced block exceeds the latest

### Hive
Enabling eth/69 ensures all existing hive tests (e.g., receipts and status) pass successfully, and the happy path for `BlockRangeUpdate` functions as expected. This PR addresses one new test not covered by the original eth/69 #8519. Despite this fix, the test still fails because the peer is disconnected due to a failure in retrieving chain head info. This disconnection is unnecessary for peers using eth/69, and an issue has been created to address this behavior: https://github.com/hyperledger/besu/issues/8475


Gabriel-Trintinalia:
Enabling eth/69 ensures all existing hive tests (e.g., receipts and status) pass successfully, and the happy path for BlockRangeUpdate functions as expected. However, one new test, not covered by the original eth/69 PR, is addressed in this PR. Despite this fix, the test still fails because the peer is disconnected due to a failure in retrieving chain head info. This disconnection is unnecessary for peers using eth/69, and an issue has been created to address this behavior: https://github.com/hyperledger/besu/issues/8475

macfarla (approved):

PR #8718, opened 2025-05-29
renames to GenesisFile since it is specific to serialization

jflo:
## PR description

First of a few refactoring PRs, this one isolates some renaming for easier review.


siladu (commented):
But what about https://github.com/hyperledger/besu/commit/3b4136dac5ef6bbff8eb039a9459af221fbc89cc 😆

Matilda-Clerke:
I think Simon has a point there... It is in fact an object containing configuration data, typically populated from the content of a file.

daniellehrner (changes requested):
There are still some places that use `GenesisConfig`. For example:

BesuCommand:326 private final Supplier<GenesisFile> genesisConfigSupplier = Suppliers.memoize(this::readGenesisConfig);

BesuCommand:564 names = {"--override-genesis-config"},

EthNetworkConfig:133 public Builder setGenesisConfig(final GenesisFile genesisFile) {

Those are just some examples, but I think it would make sense to double check and rename some of those leftovers where it makes sense.

PR #8725, opened 2025-06-01
feat: remove chain head check if peer supports eth/69

Gabriel-Trintinalia:
## Description 
- Remove chain head check if peer supports eth/69
- Use genesis block to test snap instead of head

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
@@ -29,7 +29,6 @@ import org.hyperledger.besu.ethereum.forkid.ForkIdManager;
 import org.hyperledger.besu.ethereum.mainnet.ProtocolSpec;
 import org.hyperledger.besu.ethereum.p2p.peers.Peer;
 import org.hyperledger.besu.ethereum.p2p.peers.PeerId;
-import org.hyperledger.besu.ethereum.p2p.rlpx.ConnectCallback;
 import org.hyperledger.besu.ethereum.p2p.rlpx.RlpxAgent;
 import org.hyperledger.besu.ethereum.p2p.rlpx.connections.PeerConnection;
 import org.hyperledger.besu.ethereum.p2p.rlpx.wire.PeerClientName;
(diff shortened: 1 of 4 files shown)


Matilda-Clerke (approved):

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -557,80 +556,95 @@ public String toString() {
...
+                try {
+                  checkIsSnapServer(peer);
+                } catch (Exception e) {
+                  throw new RuntimeException(e);
+                }
+              });
+    } else {
+      isServingSnapFuture = CompletableFuture.completedFuture(null);
+    }
+    isServingSnapFuture.thenRun(
```
pinges, line 631: This does not really belong into the check snap server method. Probably return the future and do that in the status exchanged method.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/EthPeers.java
```diff
@@ -557,80 +556,95 @@ public String toString() {
...
+          < trailingPeerRequirements.getMinimumHeightToBeUpToDate()) {
+        if (!(getNumTrailingPeers(trailingPeerRequirements.getMinimumHeightToBeUpToDate())
+            < trailingPeerRequirements.getMaxTrailingPeers())) {
+          LOG.atTrace()
+              .setMessage(
+                  "Adding trailing peer {} would exceed max trailing peers {}. Disconnecting...")
+              .addArgument(peer.getLoggableId())
+              .addArgument(trailingPeerRequirements.getMaxTrailingPeers())
+              .log();
+          peer.disconnect(DisconnectMessage.DisconnectReason.USELESS_PEER_EXCEEDS_TRAILING_PEERS);
```
pinges, line 608: I don't like it very much that we could connect to the peer and then immediately disconnect if it is trailing. Maybe if the check snap server method returns a future we could wait for both tests to finish and then make the decision whether we want to connect or not.

pinges (commented):
Added two comments

PR #8733, opened 2025-06-03
Modify era1blockimporter to mimic full import

Matilda-Clerke:
## PR description
Modify ERA1 import to mimic full import to allow continuation with FULL sync mode to produce a node with full history


besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
```diff
@@ -119,18 +113,7 @@ public void handleExecutionBlockBody(final Era1ExecutionBlockBody executionBlock
...
-                      RLPInput input =
-                          new BytesValueRLPInput(
-                              Bytes.wrap(executionBlockReceipts.receipts()), false);
-                      final List<TransactionReceipt> receiptsForBlock = new ArrayList<>();
-                      input.readList(
-                          (in) ->
-                              receiptsForBlock.add(TransactionReceiptDecoder.readFrom(in, false)));
-                      return receiptsForBlock;
-                    }));
+            // not really necessary, do nothing
```
siladu: Can you elaborate please?

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
```diff
@@ -141,22 +124,14 @@ public void handleBlockIndex(final Era1BlockIndex blockIndex) {
...
       BlockImportResult importResult =
-          blockImporter.importBlockForSyncing(
-              context,
-              block,
-              receiptsFutures.get(i).get(10, TimeUnit.SECONDS),
-              HeaderValidationMode.NONE,
-              HeaderValidationMode.NONE,
-              BodyValidationMode.NONE,
-              false);
+          blockImporter.importBlock(context, block, HeaderValidationMode.SKIP_DETACHED);
```
siladu: Can you explain why SKIP_DETACHED is appropriate here?

IIRC, detached vs attached is if we have the parent available, e.g. importing blocks out of order.
IIUC we are importing blocks in order here so why not ATTACHED?

siladu (commented):
Are there any automated tests for this?

Or can you add some results from manual testing?

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
Matilda-Clerke: It's what's done in `FullImportBlockStep` for FULL sync. We could probably do FULL validation in this case if we want, but given we trust the files and just want to import as quickly as possible, I'm not sure it's desirable. In fact, it may be best to change this to NONE.

What do you think?

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
Matilda-Clerke: The block import process doesn't require the receipts. Don't they get produced when the block is applied to the chain?

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
siladu: I'm not sure it's a given we trust the files, it might be worth having a hash verification somewhere as part of the era import but maybe we can do that quite cheaply as a one-off at the start? ...maybe @garyschulte has an idea how?

Assuming we can trust it then yes perhaps we can skip some checks. I don't understand the sync ValidationMode's well enough to know the appropriate enum to use tbh - maybe @jframe or @pinges does?

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
siladu: > The block import process doesn't require the receipts.

^ That would be a better comment for the code :)

> Don't they get produced when the block is applied to the chain?

I think so, would be good to verify that and add the why to the comment 👍

Matilda-Clerke:
Initial results from manual testing on sepolia
image
We can see upon besu startup, chain height was around 1.4m blocks. For some reason, besu decided to switch to backwards sync.

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
Matilda-Clerke: Looks like basic chain validation is always performed (i.e. checking that parent hash is the same as out current chain head). @Gabriel-Trintinalia suggested that would probably be enough, and we can skip any additional validation. If we want to include validation in the future, it should be easy to change the code anyway.

siladu: > Looks like basic chain validation is always performed

even with NONE?

siladu: ah, I see you updated to NONE so I guess so. Will be interesting to see how that affects sync speed :)

besu/src/main/java/org/hyperledger/besu/chainimport/Era1BlockImporter.java
```diff
@@ -119,44 +113,25 @@ public void handleExecutionBlockBody(final Era1ExecutionBlockBody executionBlock
...
-                      RLPInput input =
-                          new BytesValueRLPInput(
-                              Bytes.wrap(executionBlockReceipts.receipts()), false);
-                      final List<TransactionReceipt> receiptsForBlock = new ArrayList<>();
-                      input.readList(
-                          (in) ->
-                              receiptsForBlock.add(TransactionReceiptDecoder.readFrom(in, false)));
-                      return receiptsForBlock;
-                    }));
+            // Receipts are generated during the import process. Do nothing
```
siladu: I am wondering if it would be a faster sync to import here rather than generate on the fly, but that can be a separate PR. Happy to keep this simple for now

Matilda-Clerke: That would be a pretty substantial change. Currently, we're just using the existing toolkit in the same way as FULL sync.

jframe (commented):
I think this the wrong approach to take. It's turning the import subcommand for era files into a sync command and only for era files. Would prefer that the import subcommand just did an import. And the normal sync process would recognise that blocks are downloaded and skip downloading them.

Matilda-Clerke:
Discussed with Jason. This works as a proof of concept, but not necessarily worth being merged to main. Next, improving UX by updating the FULL sync pipeline to read and process era1 files from disk (or a http source?) and then switching to syncing from devp2p after exhausting available era1 files.

PR #8738, opened 2025-06-04
chore: increase default gas cap to 100M

Gabriel-Trintinalia:
Increase default gas cap since we have test networks with 60M gas limit already


macfarla (approved):
LGTM - maybe check whether the default is mentioned in the docs, do they need updating

PR #8739, opened 2025-06-04
Archive CHANGELOG, split at start of 2024

siladu:
The current CHANGELOG.md is cumbersome and Github struggles to load it.

This PR moves CHANGELOG entries from before 2024 to docs/CHANGELOG_ARCHIVE.md


siladu:
> Sounds like a good idea. How often will we "archive" the changelog file?

Not sure, but this should keep us going for a good while.

macfarla (approved):
only thing I would add is a link from CHANGELOG to the CHANGELOG_ARCHIVE

matkt (approved):

macfarla:
I have another suggestion - what about remove the "upcoming breaking changes" section from all but latest release?

jframe:
Rather than archive the old entries I think we can just remove them. They always available in git anyways or as part of the old release.

jframe (approved):

PR #8753, opened 2025-06-06
Revert "feat: remove chain head check if peer supports eth/69 (#8725)"

Gabriel-Trintinalia:
This reverts commit ed1086d1a91eb7b3044b744d6a76e6dbdd6d2922.

## PR description

There is an issue with how this commit tracks if a peer is serving snap data. Reverting for now as it will need to be better explored.


jframe (approved):
LGTM

PR #8763, opened 2025-06-07
fix(engine): engine_getBlobsV2 missing from engine_exchangeCapabilities

Gabriel-Trintinalia:
This PR resolves an issue where the `engine_getBlobsV2` method was not included in the list of supported methods returned by `engine_exchangeCapabilities`.

As a result, consensus clients incorrectly assumed that the execution layer did not support blob V2 related requests, leading to errors like:

```
[2025-06-05 20:10:57.89]  WARNING execution: Connected execution client does not support some requested 
engine methods methods=['engine_getBlobsV2']
```


ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/ExecutionEngineJsonRpcMethodsTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.jupiter.api.Test;
+
+class ExecutionEngineJsonRpcMethodsTest {
+  /**
+   * Ensures that all methods returned by create() are valid and declared "engine_" methods. This
+   * protects against accidental omissions from the RpcMethod enum, which is used by
+   * engine_exchangeCapabilities
+   */
+  @Test
+  void testGetSupportedMethods() {
```
macfarla, line 47: love it

macfarla (approved):

PR #8765, opened 2025-06-09
EIP-7934 Validate block size

jframe:
## PR description
EIP-7934 Validate block size

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockSizeBlockValidator.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockSizeBlockValidator.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright contributors to Besu.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.ethereum.mainnet;
+
+import org.hyperledger.besu.ethereum.BlockProcessingResult;
+import org.hyperledger.besu.ethereum.MainnetBlockValidator;
+import org.hyperledger.besu.ethereum.ProtocolContext;
+import org.hyperledger.besu.ethereum.core.Block;
+
+/** Validates blocks ensuring they do not exceed the maximum allowed size. */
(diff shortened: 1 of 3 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockSizeBlockValidator.java
```diff
@@ -0,0 +1,67 @@
...
+   * @param blockBodyValidator the BlockBodyValidator used to validate block bodies
+   * @param blockProcessor the BlockProcessor used to process blocks
+   */
+  public BlockSizeBlockValidator(
+      final BlockHeaderValidator blockHeaderValidator,
+      final BlockBodyValidator blockBodyValidator,
+      final BlockProcessor blockProcessor) {
+    super(blockHeaderValidator, blockBodyValidator, blockProcessor);
+  }
+
```
Gabriel-Trintinalia: Since the limit is in the gasCalculator "for now", this class can be removed and the check can be moved to MainnetTransactionValidator.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockSizeBlockValidator.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  public BlockProcessingResult validateAndProcessBlock(
+      final ProtocolContext context,
+      final Block block,
+      final HeaderValidationMode headerValidationMode,
+      final HeaderValidationMode ommerValidationMode,
+      final boolean shouldUpdateHead,
+      final boolean shouldRecordBadBlock) {
+    final int blockSize = block.getSize();
+    if (blockSize > MAX_RLP_BLOCK_SIZE) {
```
Gabriel-Trintinalia: Since we have a method that returns it in the gasCalculator, it is probably better to avoid using the constant.

evm/src/main/java/org/hyperledger/besu/evm/gascalculator/OsakaGasCalculator.java
```diff
@@ -34,6 +34,14 @@
...
  */
 public class OsakaGasCalculator extends PragueGasCalculator {
+  /** The maximum size of a block in bytes */
+  public static final int MAX_BLOCK_SIZE = 10_485_760;
+
+  /** The safety margin to take the CL block overhead into account */
+  public static final int SAFETY_MARGIN = 2_097_152;
+
+  /** The maximum size of an RLP encoded block size in bytes */
+  public static final int MAX_RLP_BLOCK_SIZE = MAX_BLOCK_SIZE - SAFETY_MARGIN;
```
Gabriel-Trintinalia: Can be private since we have a method that returns it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/BlockSizeBlockValidator.java
jframe: Rather keep it as is as in the future in this PR or another the limit being in the GasCalculator will be removed and at that point I'd want this in it's own validator.

PR #8771, opened 2025-06-09
feat: EIP-7939 Implement counting leading zero operation

Gabriel-Trintinalia:
Implements https://eips.ethereum.org/EIPS/eip-7939

This pull request adds support for the EIP-7939 CLZ opcode, implementing the operation itself, adding corresponding tests, and properly registering the new opcode in the mainnet EVM operations.

Introduces CountLeadingZerosOperation which computes the number of leading zeros for a given input.
Provides parameterized tests to verify correct opcode functionality.
Updates MainnetEVMs to register the new CLZ operation.


daniellehrner (dismissed):
We need to add a `case 0x1e` in the switch statement EVM.runToHalt() in order to be able to be called by the EVM

evm/src/main/java/org/hyperledger/besu/evm/EVM.java
```diff
@@ -252,6 +255,10 @@ public void runToHalt(final MessageFrame frame, final OperationTracer tracing) {
               case 0x18 -> XorOperation.staticOperation(frame);
               case 0x19 -> NotOperation.staticOperation(frame);
               case 0x1a -> ByteOperation.staticOperation(frame);
+              case 0x1e ->
+                  enableOsaka
```
lu-pinto, line 259: why can't this be part of the ProtocolSpecs? Don't like to have fork specific info inside these classes. It seems we can load all of the opcodes during configuration. But maybe this is a bigger refactoring than this PR should address

Gabriel-Trintinalia:
> Please add an end to end test for this opcode to make sure it fully works for the fork. Suggestion: look into EvmToolSpecTests

@lu-pinto done

ahamlat:
Please don't merge before performance benchmarking.

evm/src/main/java/org/hyperledger/besu/evm/operation/CountLeadingZerosOperation.java
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Static operation.
+   *
+   * @param frame the frame
+   * @return the operation result
+   */
+  public static OperationResult staticOperation(final MessageFrame frame) {
+    final Bytes value = frame.popStackItem();
+    final int numberOfLeadingZeros = UInt256.fromBytes(value).numberOfLeadingZeros();
+    frame.pushStackItem(Words.intBytes(numberOfLeadingZeros));
```
ahamlat, line 59: Let's benchmark two different implementations, the one suggested here and this one 
`frame.pushStackItem(UInt256.valueOf(value.numberOfLeadingZeros()))`

Gabriel-Trintinalia: @ahamlat 

> Let's benchmark two different implementations, the one suggested here and this one
> `frame.pushStackItem(UInt256.valueOf(value.numberOfLeadingZeros()))`

Thanks for looking into this. The reference tests do not pass with `Bytes.numberOfLeadingZeros()`. I initially was using it but replaced in this commit: https://github.com/hyperledger/besu/pull/8771/commits/e82709ee04f8c6961e6658539afed797df38d375

lu-pinto: do you know why were they not passing? The implementation is literally the same - I'm surprised

Gabriel-Trintinalia: @lu-pinto  It works if Bytes.size() == 32. Wrapping into a `Bytes32` works fine.

lu-pinto: Ok so @ahamlat you could probably do:
```
final int numberOfLeadingZeros = Bytes32.wrap(value).numberOfLeadingZeros();
frame.pushStackItem(Words.intBytes(numberOfLeadingZeros));
```
but I agree, my guess, is less performant.

Gabriel-Trintinalia:
@ahamlat @lu-pinto any updates on the benchmarking? Is there any reason why we couldn't merge it as it is and revisit it when the benchmark is done?

lu-pinto:
So I ran some benchmarks for both cases (I actually had to change the suggestion implementation slightly to make it work):
```
      int numberOfLeadingZeros = value.numberOfLeadingZeros() + Bytes32.SIZE - value.size();
```

Benchmark results show 2-3x slower for the impl in this PR:
```
Benchmark                                                                                                     (bytesHex)  (useBase32Wrapping)  Mode  Cnt   Score   Error  Units
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23                 true  avgt   10  14.550 ± 0.017  ms/op
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23                false  avgt   10  48.058 ± 0.114  ms/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323                 true  avgt   10  15.090 ± 0.068  ms/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323                false  avgt   10  26.924 ± 0.104  ms/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323                 true  avgt   10  14.553 ± 0.019  ms/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323                false  avgt   10  47.515 ± 0.207  ms/op
```

branch where I got my benchmarks published: [lu-pinto/benchmark-8770-clz-opcode](https://github.com/lu-pinto/besu/tree/benchmark-8770-clz-opcode)

ahamlat:
We can even simplify the code and avoid the check.
Instead of 
```
numberOfLeadingZeros = value.numberOfLeadingZeros();
      if (value.size() != Bytes32.SIZE) {
        numberOfLeadingZeros += Bytes32.SIZE - value.size();
      }
```
we can have, as in the case where the check is false, value.size() is called twice, and we already that this call can be time consuming
```
numberOfLeadingZeros = value.numberOfLeadingZeros() +  Bytes32.SIZE - value.size();
 ```

**Comment by Gabriel-Trintinalia** [comment]
@lu-pinto @ahamlat what are the next steps here?

**Comment by ahamlat** [comment]
The benchmarks show that when useBase32Wrapping is true, the results is much better in terms of performance. Which means that the fast implementation is this one 
```
 numberOfLeadingZeros = value.numberOfLeadingZeros();
      if (value.size() != Bytes32.SIZE) {
        numberOfLeadingZeros += Bytes32.SIZE - value.size();
      }
```
Could you replace with this implementation ? remove DO NOT MERGE and make it ready for review ?

**Comment by Gabriel-Trintinalia** [comment]
@ahamlat @lu-pinto 
> The benchmarks show that when useBase32Wrapping is true, the results is much better in terms of performance. Which means that the fast implementation is this one
> 
> ```
>  numberOfLeadingZeros = value.numberOfLeadingZeros();
>       if (value.size() != Bytes32.SIZE) {
>         numberOfLeadingZeros += Bytes32.SIZE - value.size();
>       }
> ```
> 
> Could you replace with this implementation ? remove DO NOT MERGE and make it ready for review ?

Unfortunately, It does not pass the reference tests. I added a failing test to the PR:

`Arguments.of("0xff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff", 8));`

The proposed change above returns 1. 

Screenshot 2025-07-04 at 1 50 12 pm

**Comment by lu-pinto** [comment]
> @ahamlat @lu-pinto
> 
> > The benchmarks show that when useBase32Wrapping is true, the results is much better in terms of performance. Which means that the fast implementation is this one
> > ```
> >  numberOfLeadingZeros = value.numberOfLeadingZeros();
> >       if (value.size() != Bytes32.SIZE) {
> >         numberOfLeadingZeros += Bytes32.SIZE - value.size();
> >       }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could you replace with this implementation ? remove DO NOT MERGE and make it ready for review ?
> 
> Unfortunately, It does not pass the reference tests. I added a failing test to the PR:
> 
> `Arguments.of("0xff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff", 8));`
> 
> The proposed change above returns 1.
> 
> Screenshot 2025-07-04 at 1 50 12 pm

Fixed on the new tip https://github.com/hyperledger/besu/commit/5607b1c070abe52366d1ec180fbe7a0e9fa34faf. Was partially counting bytes instead of bits :)

**Comment by ahamlat** [comment]
We had a chat about it with @lu-pinto . There's one main issue and the unit test is made in a way to reproduce the reference test :
1. The main issue is :
```
 numberOfLeadingZeros += Bytes32.SIZE - value.size();
 ```
 This should be
 ```
 numberOfLeadingZeros += 8 * (Bytes32.SIZE - value.size());
 ```
 2. The second one is not really an issue, but it helped to show the issue in the unit test
 Instead of
 ```
 Bytes input = Bytes.fromHexString(value);
 ```
You can use
 ```
 Bytes input = Bytes32.fromHexString(value);
 ```
But may be we should keep using Bytes instead just to be sure we're handling correctly this case.

evm/src/main/java/org/hyperledger/besu/evm/operation/CountLeadingZerosOperation.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Static operation.
+   *
+   * @param frame the frame
+   * @return the operation result
+   */
+  public static OperationResult staticOperation(final MessageFrame frame) {
+    Bytes value = frame.popStackItem();
+    final int numberOfLeadingZeros;
+    if (value.size() > Bytes32.SIZE) {
```
ahamlat, line 54: @lu-pinto Is this check part of the microbenchmarking results you shared ?

evm/src/main/java/org/hyperledger/besu/evm/EVM.java
ahamlat: 100% agree. @Gabriel-Trintinalia Could you elaborate on this ?

ahamlat: even enableShanghai should not be there from what my checks in the code.

ahamlat: But this is out of the scope of this PR.

Gabriel-Trintinalia: This code was introduced to allow the call of the `staticOperation` directly, and, AFAIK It is an optimisation only

evm/src/main/java/org/hyperledger/besu/evm/operation/CountLeadingZerosOperation.java
lu-pinto: Nope, but I got some fresh benchmarks. These are taken at the tip of this branch:
```
Benchmark                                                                                                               (bytesHex)  Mode  Cnt   Score   Error  Units
OneMillionCountLeadingZerosOperationBenchmark.executeOperation                                                                0x23  avgt   10  12.637 ± 0.009  ms/op
OneMillionCountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10  15.119 ± 0.031  ms/op
OneMillionCountLeadingZerosOperationBenchmark.executeOperation    0x23232323232323232323232323232323232323232323232323232323232323  avgt   10  12.642 ± 0.025  ms/op
OneMillionCountLeadingZerosOperationBenchmark.executeOperation                    0x2323232323232323232323232323232323232323232323  avgt   10  12.648 ± 0.041  ms/op
OneMillionCountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323  avgt   10  12.678 ± 0.047  ms/op
```
It looks in line with what we've seen above.

evm/src/main/java/org/hyperledger/besu/evm/EVM.java
lu-pinto: I think we could get rid of it then when we move to lambdas

evm/src/main/java/org/hyperledger/besu/evm/operation/CountLeadingZerosOperation.java
```diff
@@ -0,0 +1,62 @@
...
+      // should not happen but trim just in case
+      value = value.slice(0, 32);
```
lu-pinto: This is actually not right as we are working with Big endians. I'll push a fix

lu-pinto (approved):

PR #8772, opened 2025-06-09
Add Block size when available

pinges:
## PR description
When creating a block from the RLP set the size of the Block. Previously the size of a Block was calculated by RLP encoding it, which is unnecessary now if we create the block from the RLP.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
@@ -29,10 +29,18 @@ public class Block {
 
   private final BlockHeader header;
   private final BlockBody body;
+  private int size;
+
+  public Block(final BlockHeader header, final BlockBody body, final int size) {
+    this.header = header;
+    this.body = body;
+    this.size = size;
+  }
 
   public Block(final BlockHeader header, final BlockBody body) {
     this.header = header;
     this.body = body;
+    this.size = -1; // Size will be calculated on demand
   }
 
   public BlockHeader getHeader() {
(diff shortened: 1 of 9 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/util/RawBlockIterator.java
```diff
@@ -95,10 +93,7 @@ private void nextBlock() throws IOException {
 
       final Bytes rlpBytes = Bytes.wrap(Bytes.wrapByteBuffer(readBuffer, 0, length).toArray());
       final RLPInput rlp = new BytesValueRLPInput(rlpBytes, false);
-      rlp.enterList();
-      final BlockHeader header = BlockHeader.readFrom(rlp, blockHeaderFunctions);
-      final BlockBody body = BlockBody.readFrom(rlp, blockHeaderFunctions);
-      next = new Block(header, body);
```
jframe: Not sure this is the same as `Block.readFrom`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
```diff
@@ -51,8 +59,11 @@ public Bytes toRlp() {
     return RLP.encode(this::writeTo);
   }
 
-  public int calculateSize() {
-    return toRlp().size();
+  public int getSize() {
```
jframe, line 62: Maybe worth a test for this

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/util/RawBlockIterator.java
pinges: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Block.java
pinges: Added a test

jframe (approved):
LGTM

PR #8793, opened 2025-06-10
Introduce the `TransactionValidatorService` to allow plugins to add custom validation rules

fab-10:
## PR description

This PR fills a gap and allows for plugin to introduce additional transaction validation rules, on top of the standard protocol validation that is already performed, and such the rules introduced by the plugin can only restrict what is valid, and cannot revert any standard validation. 
For example a network that want to ban a specific kind of txs, can do that implementing a plugin that make use of this new service.

The additional validation rules are enforced every time a transaction validation is needed, basically they extend the protocol validation rules for transactions:
- before accepting the tx in the pool
- when importing a block
- when building a block

This new feature complements the other 2 feature that extend the transaction validator, but are meant for different use cases:
- `TransactionPoolValidatorService` is only meant to validate tx before they are accepted in the pool, and does not have any effect on block time validation, and does not alter the protocol rules.
- `PermissioningService` is only meant for validating if the sender is permitted to send the tx, but does not alter in any way the protocol rules. In theory this specific case could be implemented on top of the more generic new `TransactionValidatorService` but this could be the scope of a follow up.

### DIFF
--- a/app/src/main/java/org/hyperledger/besu/services/TransactionValidatorServiceImpl.java
+++ b/app/src/main/java/org/hyperledger/besu/services/TransactionValidatorServiceImpl.java
@@ -35,7 +35,7 @@ public class TransactionValidatorServiceImpl implements TransactionValidatorServ
   public TransactionValidatorServiceImpl() {}
 
   @Override
-  public void registerTransactionValidator(final TransactionValidationRule rule) {
+  public void registerTransactionValidatorRule(final TransactionValidationRule rule) {
     transactionValidationRules.add(rule);
     LOG.info("Registered new transaction validator rule");
   }
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/TransactionValidatorFactory.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/TransactionValidatorFactory.java
@@ -117,10 +117,12 @@ public class TransactionValidatorFactory {
 
   public void setAdditionalValidationRules(
       final List<TransactionValidationRule> additionalValidationRules) {
-    final TransactionValidator baseTxValidator = transactionValidatorSupplier.get();
-    transactionValidatorSupplier =
-        Suppliers.memoize(
-            () -> new ExtendableTransactionValidator(baseTxValidator, additionalValidationRules));
+    if (!additionalValidationRules.isEmpty()) {
+      final TransactionValidator baseTxValidator = transactionValidatorSupplier.get();
+      transactionValidatorSupplier =
+          Suppliers.memoize(
+              () -> new ExtendableTransactionValidator(baseTxValidator, additionalValidationRules));
+    }
   }
 
   public TransactionValidator get() {
--- a/acceptance-tests/test-plugins/src/main/java/org/hyperledger/besu/tests/acceptance/plugins/TestTransactionValidatorPlugin.java
+++ b/acceptance-tests/test-plugins/src/main/java/org/hyperledger/besu/tests/acceptance/plugins/TestTransactionValidatorPlugin.java
@@ -47,7 +47,7 @@ public class TestTransactionValidatorPlugin implements BesuPlugin {
     serviceManager
         .getService(TransactionValidatorService.class)
         .orElseThrow()
-        .registerTransactionValidator(
+        .registerTransactionValidatorRule(
             tx ->
                 enabled && !tx.getType().equals(TransactionType.FRONTIER)
                     ? Optional.of("Only Frontier transactions are allowed here")
--- a/app/src/main/java/org/hyperledger/besu/RunnerBuilder.java
+++ b/app/src/main/java/org/hyperledger/besu/RunnerBuilder.java
@@ -780,10 +780,8 @@ public class RunnerBuilder {
         .map(DefaultPeer::fromEnodeURL)
         .forEach(peerNetwork::addMaintainedConnectionPeer);
 
-    if (!transactionValidatorService.getTransactionValidatorRules().isEmpty()) {
-      protocolSchedule.setAdditionalValidationRules(
-          transactionValidatorService.getTransactionValidatorRules());
-    }
+    protocolSchedule.setAdditionalValidationRules(
+        transactionValidatorService.getTransactionValidatorRules());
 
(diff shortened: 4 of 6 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ExtendableTransactionValidator.java
```diff
@@ -0,0 +1,74 @@
...
+import org.hyperledger.besu.ethereum.transaction.TransactionInvalidReason;
+import org.hyperledger.besu.evm.account.Account;
+import org.hyperledger.besu.plugin.services.txvalidator.TransactionValidationRule;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Validates a transaction based on mainnet rule and other custom rules that can added by plugins.
+ * Note that custom validation rules can only be more restrictive and cannot override standard
```
macfarla:
```suggestion
 * Note that custom validation rules can only be more restrictive and standard
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ExtendableTransactionValidator.java
```diff
@@ -0,0 +1,74 @@
...
+ * mainnet rules.
```
macfarla:
```suggestion
 * mainnet rules will still apply.
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ExtendableTransactionValidator.java
```diff
@@ -0,0 +1,74 @@
...
+import org.hyperledger.besu.ethereum.core.Transaction;
+import org.hyperledger.besu.ethereum.transaction.TransactionInvalidReason;
+import org.hyperledger.besu.evm.account.Account;
+import org.hyperledger.besu.plugin.services.txvalidator.TransactionValidationRule;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Validates a transaction based on mainnet rule and other custom rules that can added by plugins.
```
macfarla:
```suggestion
 * Validates a transaction based on mainnet rule and other custom rules that can be added by plugins.
```

app/src/main/java/org/hyperledger/besu/services/TransactionValidatorServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    transactionValidationRules.add(rule);
+    LOG.info("Registered new transaction validator.");
+  }
+
+  /**
+   * Gets transaction validators.
+   *
+   * @return the transaction validators
+   */
+  public List<TransactionValidationRule> getTransactionValidators() {
```
macfarla: the naming is a bit confusing - should this be getTransactionValidationRules ?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/RpcErrorType.java
```diff
@@ -176,6 +176,8 @@ public enum RpcErrorType implements RpcMethodError {
       -32000, "An invalid transaction with a lower nonce exists"),
   TOTAL_BLOB_GAS_TOO_HIGH(-32000, "Total blob gas too high"),
   PLUGIN_TX_VALIDATOR(-32000, "Plugin has marked the transaction as invalid"),
+  PLUGIN_TX_POOL_VALIDATOR(
+      -32000, "Plugin has marked the transaction as not acceptable by the transaction pool"),
```
macfarla, line 180: there is not much difference between these 2 messages, not sure if this could be made clearer though, since we don't necessarily know why the plugin marks it invalid?

app/src/main/java/org/hyperledger/besu/services/TransactionValidatorServiceImpl.java
fab-10: good catch, I missed to rename this one

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/RpcErrorType.java
fab-10: I thought about that, but I did not find any better message, do you have any suggestion?
At the moment there is no way to surface the actual error in the RPC response, maybe something worth to add in a future PR

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/TransactionValidatorFactory.java
```diff
@@ -113,6 +115,14 @@ public void setPermissionTransactionFilter(
             () -> new PermissionTransactionValidator(baseTxValidator, permissionTransactionFilter));
   }
 
+  public void setAdditionalValidationRules(
+      final List<TransactionValidationRule> additionalValidationRules) {
+    final TransactionValidator baseTxValidator = transactionValidatorSupplier.get();
+    transactionValidatorSupplier =
```
Gabriel-Trintinalia: nit: can we return the `baseTxValidator` if no `additiionalValidationRules` are present?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/RpcErrorType.java
macfarla: fair enough. this is fine for now then.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/TransactionValidatorFactory.java
fab-10: this only called when there are additional rules, because the check to see if the list is empty is done in the `RunnerBuilder` but probably is clearer to move it here

fab-10:
> LGTM. Very nice feature! It gave me an idea to refactor the `MainnetTransactionValidator` to have a list of rules as well. Similar to what we do for the transaction selectors. Then we can extend it by forks similar to what we do the header validation rules. I think I will work on a refactor next week

Indeed, sounds good!

PR #8814, opened 2025-06-12
Change stack size growth strategy to 0.5 rate

lu-pinto:
This PR changes the stack rate growth strategy of `FlexStack` to grow at a 50% rate instead of constantly growing by 32 slots as is currently. The former is the approach adopted by the JDK for `ArrayList`.

With a 50% allocation rate and considering a max stack size of 1024, setting the initial size to 91 one can reach the worse case (1024) in just 6 resizes compared to 32 required with the current approach.
Also the current usage on mainnet has shown that CALLs never go above 150 deep during execution.


lu-pinto:
would like to do some performance testing before merging

lu-pinto:
pushing to performance branch for experimentation

PR #8815, opened 2025-06-12
Introduce global code cache

daniellehrner:
fixed the non-functional code hash and integrated it into PathBasedAccount, the existing EVM.getCode*() methods do not use a cache anymore and has been renamed to reflect that they are only wrapping the bytecode into Code object


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/common/PathBasedAccount.java
```diff
@@ -150,22 +169,56 @@ public void setBalance(final Wei value) {
...
   @Override
   public Bytes getCode() {
-    if (code == null) {
-      code = context.getCode(address, codeHash).orElse(Bytes.EMPTY);
+    return getAnalyzedCode().getBytes();
+  }
+
+  @Override
+  public Code getAnalyzedCode() {
+    final Code cachedCode =
```
ahamlat: I think we should check the code before, and if it not null return it. Why a code would exist but not in the cache ?

ahamlat: You can also avoid this lookup when code != null, put it again anyway as you tested in the unit tests.

evm/src/main/java/org/hyperledger/besu/evm/code/CodeV0.java
```diff
@@ -39,24 +36,37 @@ public class CodeV0 implements Code {
...
   private final CodeSection codeSectionZero;
 
   /**
    * Public constructor.
    *
-   * @param bytes The byte representation of the code.
+   * @param byteCode The byte representation of the code.
+   */
+  public CodeV0(final Bytes byteCode) {
+    this(byteCode, byteCode.isEmpty() ? Hash.EMPTY : Hash.hash(byteCode));
```
ahamlat: I think it is better to keep the memoizer here on calculating the hash of the bytecode.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/common/PathBasedAccount.java
```diff
@@ -150,22 +169,56 @@ public void setBalance(final Wei value) {
...
+    // cache hit, but code was not set before, set it
+    if (cachedCode != null) {
+      code = cachedCode;
+      return cachedCode;
     }
+
+    // cache miss, code not set: get the code from the disk, set it and put it in the cache
+    // code cannot be empty here, as it would have been set to EMPTY_CODE in the constructor
+    final Bytes byteCode = context.getCode(address, codeHash).orElse(Bytes.EMPTY);
+    code = new CodeV0(byteCode);
```
ahamlat: I suggest using this constructor `CodeV0(final Bytes byteCode, final Hash codeHash) `as we already have the hashcode :

ahamlat (changes requested):

macfarla:
@daniellehrner are you waiting on reviews for this one?

daniellehrner:
> @daniellehrner are you waiting on reviews for this one?

@macfarla Yes, exactly. Ameziane said he wanted to review it, but did not have time yet because of other more pressing issues.

ahamlat:
@matkt Do you want to take a last look before approving ?

ahamlat (approved):
LGTM 🚢 
These are items I checked :

- [x] The nodes running this PR have better performances than nodes without this PR
- [x] The profiling shows that the improvement is related to using code cache
- [x] Generated the heap dump and checked that the cache is not consuming more than expected. I also checked that the entries in the cache are not holding bigger ArrayWrappingBytes

PR #8816, opened 2025-06-13
Transactions: Add method to return size when included into a block

pinges:
## PR description
EIP 7934 [Add bytesize limit to blocks](https://ethereum-magicians.org/t/eip-7934-add-bytesize-limit-to-blocks/23589) needs the size of a transaction that is added to the block when it is encoded for the block. The old size() method on transaction would have returned the wrong size for blob transactions. This PR renames the old size() method on the Transaction class to getSizeForAnnouncement() and adds a second method getSizeForBlockInclusion() that returns the size of the transaction when it is added to a block.
Also, when decoding from RLP, the size and the hash of the Transactions is set if possible. Before, when the size() or the hash() methods were called for the first time the transaction needed to be encoded to calculate the hash and the size of the transaction.


datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
```diff
@@ -247,7 +247,14 @@ default Optional<? extends Quantity> getMaxFeePerBlobGas() {
    *
    * @return the size in bytes of the encoded transaction.
    */
-  int getSize();
+  int getSizeForAnnouncement();
```
jframe, line 253: should update the javadoc for this

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -113,12 +113,13 @@ public class Transaction
...
-  protected volatile Address sender;
+  private volatile Address sender;
 
   // Caches the hash used to uniquely identify the transaction.
-  protected volatile Hash hash;
+  private volatile Hash hash;
   // Caches the size in bytes of the encoded transaction.
-  protected volatile int size = -1;
+  private volatile int sizeForAnnouncement;
+  private volatile int sizeForBlockInclusion;
```
jframe: why not initialise with -1 like we used to?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -182,6 +183,9 @@ public static Transaction readFrom(final RLPInput rlpInput, final EncodingContex
    *     otherwise it should contain an address.
    *     <p>The {@code chainId} must be greater than 0 to be applied to a specific chain; otherwise
    *     it will default to any chain.
+   * @param hash the transaction hash
+   * @param sizeForAnnouncement the size of the transaction, used for announcement
+   * @param sizeForBlockInclusion the size of the transaction, used for block inclusion
```
jframe, line 188: It's not clear that these are optional from the parameters. Maybe make them `Option<Integer>` in the params. We can still keep it as int field in the class.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
```diff
@@ -585,23 +595,45 @@ public Hash getHash() {
...
+    }
+  }
+
+  private void memoizeSizeForAnnouncement() {
+    final Bytes pooledBytes =
+        TransactionEncoder.encodeOpaqueBytes(this, EncodingContext.POOLED_TRANSACTION);
+    sizeForAnnouncement = pooledBytes.size();
+    if (!transactionType.supportsBlob() || this.getBlobsWithCommitments().isEmpty()) {
+      sizeForBlockInclusion = pooledBytes.size();
+      if (hash == null) {
```
jframe, line 636: can you add a comment on why this is being done

jframe, line 636: This logic to memoize the hash if not null I think would be clearer if in it's own function. It could then be called from both memoize functions

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoder.java
```diff
@@ -30,22 +30,18 @@ public class TransactionDecoder {
...
-          TransactionType.EIP1559,
-          EIP1559TransactionDecoder::decode,
-          TransactionType.BLOB,
-          BlobTransactionDecoder::decode,
-          TransactionType.DELEGATE_CODE,
-          CodeDelegationTransactionDecoder::decode);
+          TransactionType.ACCESS_LIST, input -> AccessListTransactionDecoder.decode(input),
+          TransactionType.EIP1559, input -> EIP1559TransactionDecoder.decode(input),
+          TransactionType.BLOB, input -> BlobTransactionDecoder.decode(input),
+          TransactionType.DELEGATE_CODE, input -> CodeDelegationTransactionDecoder.decode(input));
```
jframe: These can be method references

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoder.java
```diff
@@ -30,22 +30,18 @@ public class TransactionDecoder {
...
 
   private static final ImmutableMap<TransactionType, Decoder> POOLED_TRANSACTION_DECODERS =
-      ImmutableMap.of(TransactionType.BLOB, BlobPooledTransactionDecoder::decode);
+      ImmutableMap.of(TransactionType.BLOB, input -> BlobPooledTransactionDecoder.decode(input));
```
jframe: This can also be a method reference

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionSizeAndHashTest.java
```diff
@@ -0,0 +1,421 @@
...
+    final Transaction transactionRead =
+        TransactionDecoder.decodeOpaqueBytes(bytes, EncodingContext.BLOCK_BODY);
+
+    assertThat(transactionRead.getSizeForBlockInclusion()).isEqualTo(bytes.size());
+
+    assertThat(transactionRead.getHash()).isEqualTo(Hash.hash(bytes));
+  }
+
+  @Test
+  public void returnsRightSizeWithBlobs1a() {
```
jframe: Think we think need a better name for this and the other numbered tests

datatypes/src/main/java/org/hyperledger/besu/datatypes/Transaction.java
pinges: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
pinges: Both fields are set by the constructor. Setting them to -1 if they are not known is taken care of by the builder.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
pinges: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
pinges: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/Transaction.java
pinges: Done

pinges: Not sure

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoder.java
pinges: Done

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoder.java
pinges: Done

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/core/TransactionSizeAndHashTest.java
pinges: Done

jframe (approved):

PR #8819, opened 2025-06-13
refactor: central control point for genesis used in tests

### LINKED ISSUE #6653: Decouple Acceptance Tests from mainnet.json
Most acceptance tests use mainnet.json, so bugs on adding a new fork (eg cancun) may not get covered until much later in the SDLC.

to avoid finding bugs at the last minute, change this to holesky/sepolia (or make it configurable?)

macfarla:
## PR description
Refactor how the protocol specs are constructed for tests. Introduce a constant, which is currently set to mainnet, but would be easy to change to sepolia or other testnet in future. 

For example, we could run all tests on a specific testnet, leading up to a hardfork. But most of the time, we want to make sure we're able to catch any issues with mainnet.

~Use sepolia protocol specs for test fixtures. This should mean we catch issues earlier because hardfork changes are made to sepolia.json before mainnet.json.~

~Only use mainnet for the test where we're checking specific milestone values~

## Fixed Issue(s)
refs #6653


lu-pinto:
Why not hoodi? probably since it has less state it will be used before than sepolia so we catch issues even earlier

macfarla:
> Why not hoodi? probably since it has less state it will be used before than sepolia so we catch issues even earlier

happy to change this to hoodi. Still soliciting feedback on whether we want to move away from mainnet

siladu:
Downside of this is we have to change whenever the testnets change. 
We also lose some testing of mainnet config which is different to the testnets, so I think we trade off finding one class of issues earlier with maybe not finding a mainnet-specific issue.

Maybe we can test both? 

Or use testnets here and have additional explicit mainnet config testing to mitigate this...or maybe eth_config is sufficient?

Following the Pectra testnets, I would say sepolia (and its successor) is more like to be the first testnet to fork before Hoodi since Hoodi is now seen as a stable pre-prod env for large infra operators.

lu-pinto:
My main thinking was that Hoodi is going to be around for longer (unless another Holeskatastrophe :D ) so less changing around

siladu:
It's true but I think in reality hoodi will fork just before mainnet so sepolia will give more warning...depends what you want to optimise for I guess :)

Gabriel-Trintinalia:
Maybe we could have a constant:
` public static final ProtocolSchedule TESTING_NETWORK = ProtocolScheduleFixture.SEPOLIA;`. In the future, whenever we decide to change, it will be only one place.

macfarla:
> Maybe we could have a constant: ` public static final ProtocolSchedule TESTING_NETWORK = ProtocolScheduleFixture.SEPOLIA;`. In the future, whenever we decide to change, it will be only one place.

made this change

macfarla:
@Gabriel-Trintinalia @siladu any other feedback? Worth making this change?

macfarla:
as discussed, changed back to mainnet but keeping the refactor to make it easier to change in the future @siladu @Gabriel-Trintinalia

PR #8825, opened 2025-06-16
Split EVM executor from EvmSpec

lu-pinto:
## PR description
This PR splits up the EVM executor used in the testing infrastructure into an EvmSpec part which is useful for looking up the precompile contracts belonging to a certain fork. It's also able to get the entire configuration for the EVM at a certain fork by just passing the fork name, which is useful in e.g. `BenchmarkSubCommand`.

I looked around and couldn't find a way to do this elsewhere. In production we only seem to get the protocol schedule through timestamps or block numbers given a pre-populated set of milestones.
Even in production, at some stage as the number of forks grow, this whole set of milestones might be wasteful at startup since the list of configured hardforks doesn't change after the initial setup and until process restart.


evm/src/main/java/org/hyperledger/besu/evm/fluent/EVMExecutor.java
```diff
@@ -75,592 +62,39 @@ public class EVMExecutor {
...
-   * @return the evm executor
-   */
-  public static EVMExecutor experimentalEips(
-      final BigInteger chainId, final EvmConfiguration evmConfiguration) {
-    final EVMExecutor executor =
-        new EVMExecutor(MainnetEVMs.experimentalEips(chainId, evmConfiguration));
-    executor.precompileContractRegistry =
-        MainnetPrecompiledContracts.futureEIPs(executor.evm.getGasCalculator());
-    return executor;
+  public EVMExecutor(final EvmSpec evmSpec) {
```
garyschulte, line 77: 👍

evm/src/main/java/org/hyperledger/besu/evm/precompile/MainnetPrecompiledContracts.java
```diff
@@ -79,11 +79,12 @@ static void populateForByzantium(
     populateForFrontier(registry, gasCalculator);
     registry.put(
         Address.MODEXP,
-        BigIntegerModularExponentiationPrecompiledContract.byzantium(gasCalculator));
-    registry.put(Address.ALTBN128_ADD, AltBN128AddPrecompiledContract.byzantium(gasCalculator));
-    registry.put(Address.ALTBN128_MUL, AltBN128MulPrecompiledContract.byzantium(gasCalculator));
+        new BigIntegerModularExponentiationPrecompiledContract(gasCalculator, Long.MAX_VALUE));
+    registry.put(Address.ALTBN128_ADD, new AltBN128AddPrecompiledContract(gasCalculator, 500L));
+    registry.put(Address.ALTBN128_MUL, new AltBN128MulPrecompiledContract(gasCalculator, 40_000L));
```
garyschulte, line 84: IMO we should have constants for these values, just for clarity.  arguably the old constructors should have used constants also

evm/src/main/java/org/hyperledger/besu/evm/precompile/MainnetPrecompiledContracts.java
```diff
@@ -107,10 +108,11 @@ static PrecompileContractRegistry istanbul(final GasCalculator gasCalculator) {
   static void populateForIstanbul(
       final PrecompileContractRegistry registry, final GasCalculator gasCalculator) {
     populateForByzantium(registry, gasCalculator);
-    registry.put(Address.ALTBN128_ADD, AltBN128AddPrecompiledContract.istanbul(gasCalculator));
-    registry.put(Address.ALTBN128_MUL, AltBN128MulPrecompiledContract.istanbul(gasCalculator));
+    registry.put(Address.ALTBN128_ADD, new AltBN128AddPrecompiledContract(gasCalculator, 150L));
+    registry.put(Address.ALTBN128_MUL, new AltBN128MulPrecompiledContract(gasCalculator, 6_000L));
```
garyschulte, line 112: same same, etc

garyschulte (approved):
Clean refactor 👍 

minor comments

evm/src/main/java/org/hyperledger/besu/evm/precompile/MainnetPrecompiledContracts.java
lu-pinto: I do not disagree, as you said, I will have to figure out from which EIPs they came from so IMO that's a job for another PR. 
Also arguably what kind of info do you want to see in the constant name? I can only think of the EIP from where they originate. Because the fork and what they mean is already conveyed in the constructor's var name and caller location.

PR #8826, opened 2025-06-17
Run and fix tests in AbstractBlockCreatorTest

mirgee:
The abstract test class `AbstractBlockCreatorTest` was never subclassed, so its test methods were not executed along with other tests in the module. After making the class concrete to run the tests, one test was failing due to `libckzg4844jni.so` not being loaded. To load the native library, the test class was made to extend `TrustedSetupClassLoaderExtension` as per convention.


matkt (approved):
good catch

matkt:
@mirgee it seems some tests are failing

mirgee:
@matkt All checks passed on the first commit - can you try re-running the pipeline to make sure it's not an intermittent issue?

Issue #8831, opened 2025-06-19
Clean up RPC responses for pruned blocks


Matilda-Clerke:
As a caller of Besu's RPC endpoints, I want Besu to return sensible results when queried for blocks which have been pruned.

1. Investigate the appropriate response (empty vs error code 4444?)
2. Utilise [Geth's workload test tool](https://github.com/ethereum/go-ethereum/blob/master/cmd/workload/README.md) to identify all RPC endpoiints needing updating 
3. Update code to match expectations

PR #8833, opened 2025-06-19
[History Expiry] Return null when block data isn't found

Matilda-Clerke:
## PR description
Fix unwanted error responses and exceptions by simply returning null when block data isn't found

## Fixed Issue(s)
https://github.com/hyperledger/besu/issues/8831


siladu:
Would be great to link the relevant part of the spec for this. Also @macfarla - wonder if this came up in the rpc hive tests?

Matilda-Clerke:
I actually couldn't find this in the spec. EIP-4444 explicitly places it out of scope, and I wasn't able to find any other relevant EIPs to cover it. As such, I've just fixed the exceptions caused by missing data and ensured empty success responses are returned.

Matilda-Clerke:
Here's one example where a null response should be used when block data is missing: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbyhash

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetBlockTransactionCountByNumber.java
```diff
@@ -48,6 +48,7 @@ protected BlockParameter blockParameter(final JsonRpcRequestContext request) {
   protected String resultByBlockNumber(final JsonRpcRequestContext req, final long blockNumber) {
     return getBlockchainQueries()
         .getTransactionCount(blockNumber)
+        .filter((count) -> count >= 0)
```
siladu:
```suggestion
        .filter((count) -> count != -1)
```

would be more precise and in keeping with `EthGetBlockTransactionCountByHash`.

siladu:
> Here's one example where a null response should be used when block data is missing: [ethereum.org/en/developers/docs/apis/json-rpc#eth_getblockbyhash](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbyhash)

Seems reasonable since this PR is touching `eth_getTransaction...`.

An aside maybe, but in terms of `eth_getBlockBy...`, I guess we can return at least the header data - is that what we're doing already?

siladu (commented):
Minor inline comment. Would be good to have unit or integration test to cover the changes

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetBlockTransactionCountByNumber.java
Matilda-Clerke: Surely any negative could could be used to signify something is incorrect though? checking count is not equal to negative one more tightly couples the code to the implementation of getTransactionCount

siladu: That's a fair point, but I think the inconsistency with `EthGetBlockTransactionCountByHash` is the stronger of my two reasons. I don't mind too much whether it's != -1 or >= 0 but think they should behave the same, so maybe also change `EthGetBlockTransactionCountByHash` to match?

macfarla:
> Would be great to link the relevant part of the spec for this. Also @macfarla - wonder if this came up in the rpc hive tests?

yes this is one of the inconsistencies that has been raised at the RPC standardisation call 
and yes think we are failing some hive tests still because of this https://docs.google.com/spreadsheets/d/1UkIJ9Sgj_Bd4AfjSUOmIIq-t2rfzHUbBDSQR_3glEuw/edit?gid=0#gid=0
haven't gotten back to this for a while

like @Matilda-Clerke says the spec is a bit inconsistent

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetBlockTransactionCountByNumber.java
macfarla: I prefer >= 0

agree they should be consistent

Matilda-Clerke:
Ok, those last 3 commits aim for improve hive tests. I've now taken care of all of the `getBlock` tests expecting null results, but this has caused some additional failures in the `ethSimulate` tests. Most likely, ethSimulate is just propagating the result from getBlock calls, but in order to meet it's RPC spec obligations, it should be translating those responses as appropriate.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/AbstractBlockParameterOrBlockHashMethod.java
Matilda-Clerke, line 1: What an absolute mess this file has caused. Another candidate for a thorough refactoring!

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthGetBlockTransactionCountByNumber.java
Matilda-Clerke: Best solution IMO is to return an `Optional<Integer>` and simply leave it empty instead of relying on a negative number to indicate an issue.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Sunsetting features - for more context on the reasoning behind the deprecation of these features, including alternative options, read [this blog post](https://www.lfdecentralizedtrust.org/blog/sunsetting-tessera-and-simplifying-hyperledger-besu)
   - Stratum mining has been removed (part of PoW) [#8802](https://github.com/hyperledger/besu/pull/8802)
   - PoW RPCs removed: `eth_getWork`, `eth_submitWork`, `eth_getHashrate`, `eth_submitHashrate`, `eth_hashrate`
+- Update block related RPC calls to return a null result, better aligning with JSON-RPC specification. Flow on effect to several other RPC calls (eth_getProof, eth_simulateV1) [#8833](https://github.com/hyperledger/besu/pull/8833)
```
macfarla:
```suggestion
- Update block related RPC calls to return a null result if block not found, better aligning with JSON-RPC specification. Flow on effect to several other RPC calls (eth_getProof, eth_simulateV1) [#8833](https://github.com/hyperledger/besu/pull/8833)
```

ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/AbstractJsonRpcHttpBySpecTest.java
```diff
@@ -246,12 +246,12 @@ private void checkResponse(
               mapper
                   .writerWithDefaultPrettyPrinter()
                   .withoutAttribute("creationMethod")
-                  .writeValueAsString(mapper.readTree(actualResult)))
+                  .writeValueAsString(mapper.readTree(expectedResult)))
```
macfarla: well that's confusing

CHANGELOG.md
siladu: Also eth_call? Think it's worth mentioning if so as it's the most widely used

siladu (commented):
Generally LGTM.

For EthGetBlockBy* 
I'd like to see some units/integration tests illustrating the difference between the expired blocks and non-expired blocks.

Even if they are covered by Hive, I think it's still worth a test within Besu since Hive doesn't currently protect us from regressions.

macfarla (commented):
I built locally and ran the rpc-compat hive tests - it fixes 3 but breaks 2 - frustratingly the ethSimulate calls expect error response rather than null. 

it fixes these 3 hive tests
eth_getBlockReceipts/get-block-receipts-empty (besu_besu-local)
eth_getBlockReceipts/get-block-receipts-future (besu_besu-local)
eth_getBlockReceipts/get-block-receipts-not-found (besu_besu-local)

New failures:
eth_simulateV1/ethSimulate-make-call-with-future-block (besu_besu-local)
eth_simulateV1/ethSimulate-empty-with-block-num-set-plus1 (besu_besu-local)

macfarla:
> I built locally and ran the rpc-compat hive tests - it fixes 3 but breaks 2 - frustratingly the ethSimulate calls expect error response rather than null.
> 
> it fixes these 3 hive tests eth_getBlockReceipts/get-block-receipts-empty (besu_besu-local) eth_getBlockReceipts/get-block-receipts-future (besu_besu-local) eth_getBlockReceipts/get-block-receipts-not-found (besu_besu-local)
> 
> New failures: eth_simulateV1/ethSimulate-make-call-with-future-block (besu_besu-local) eth_simulateV1/ethSimulate-empty-with-block-num-set-plus1 (besu_besu-local)

@Gabriel-Trintinalia will this adversely affect users of eth_simulateV1 ? If not I'm ok with merging this, can refactor once we get certainty on the spec?

Matilda-Clerke:
Seems geth is returning a 4444 error code when a block is not present due to pruning. In order to achieve this, I'm afraid we'd need to refactor as described.

> Yes geth returns error 4444 if we can tell a request cannot be fulfilled due to history prunning. It isn't represented in the json-rpc spec yet because we haven't added error codes to spec officially

https://discord.com/channels/595666850260713488/953349957400821860/1387759809419280455

siladu:
> Seems geth is returning a 4444 error code when a block is not present due to pruning. In order to achieve this, I'm afraid we'd need to refactor as described.

If geth are doing something new and off-spec, e.g. returning a new error instead of the currently specified null, I'd lean towards sticking to the spec as is for now until their approach is merged into the spec...realise for history expiry this might be a bit fuzzy for some parts of the spec 😅 
If there's something unspecified, think we should try to get consensus from all EL clients and then update the spec.

In general I think a reasonable position to take is we will only change besu to match spec/hive tests (and hope they are consistent!)

I think for this PR that means I'd favour null and so no changes to your current branch since it matches the spec and hive, except for eth_simulate?
Due to eth_simulate inconsistency, obv might need to hold off though.

Matilda-Clerke:
As discussed this morning: Decided to revert changes related to hive tests and just fix the initial exceptions being thrown

macfarla (approved):

siladu (approved):
Still no tests to distinguish between a expired vs present block, but can defer to next PR.

Issue #8730, opened 2025-06-03
datatypes module error passes through GHA checks


macfarla:
Seems #8631 broke the docker build

```
buildDocker step failed with 
> Task :datatypes:compileJava FAILED
/home/runner/work/besu/besu/datatypes/src/main/java/org/***/besu/datatypes/BlobType.java:35: warning: [MethodInputParametersMustBeFinal] Method input parameters must be final.
41
  BlobType(int versionId) {
```

https://github.com/hyperledger/besu/actions/runs/15396029562/job/43316941235

Question is, how come this passed other CI checks - need to check where that module is compiled 

Also when I run 
`gw :datatypes:compileJava` locally it succeeds ie I don't get the error above

siladu:
Maybe issue with error prone config?

macfarla:
There was an issue with error-prone - fix here - https://github.com/besu-eth/besu-errorprone-checks/pull/13

PR #8846, opened 2025-06-24
build and run testDocker on PRs

macfarla:
## PR description
Previously testDocker task was only run on merge to main. This PR modifies the workflows so that those checks run at PR stage

* extra workflow file develop-pr.yml was added to run the relevant checks (build, testDocker but not publish) on PR 
* fixes to build.gradle to ensure testDocker can work without docker env which was also supplying docker org and artifact name (in addition to docker creds which are not needed since we're not publishing)

~This PR starts with a commit to reproduce the issue with errorprone check not being picked up until buildDocker (merge to main), and then modifies the workflows so that those checks run at PR stage.~

~* change was made to BlobType (remove final keyword) to trigger the original issue~

~For some reason the errorprone check only fails on besu-arm runner not the ubuntu-arm one or ubuntu-22-04~

~eg on this run you can see the errorprone check fails the build for 1 of 2 runners https://github.com/hyperledger/besu/actions/runs/15864950425?pr=8846~

~and once the final keyword is fixed, both runners succeed~

Errorprone root cause #8730 was fixed separately 
* tests added to confirm this case (final keyword within enum constructor) is covered https://github.com/besu-eth/besu-errorprone-checks/pull/13
* root cause fixed https://github.com/besu-eth/besu-errorprone-checks/pull/13

## Fixed Issue(s)
inspired by #8730


macfarla:
so the error prone PR https://github.com/besu-eth/besu-errorprone-checks/pull/13 fixes the root cause that was flagged initially here - so this besu PR would be just if we want to run build & test docker on PRs rather than just on merge to main. I would argue it's worth it if it doesn't slow down the PR build time. It takes 5 min and it's run in parallel. Maybe we would get enough signal from running on 1 of 2 runners

build.gradle
```diff
@@ -800,9 +800,9 @@ artifactoryPublish {
   dependsOn distZip
 }
 
-def dockerBuildVersion = project.hasProperty('release.releaseVersion') ? project.property('release.releaseVersion') : "${rootProject.version}"
-def dockerOrgName = project.hasProperty('dockerOrgName') ? project.getProperty("dockerOrgName") : "hyperledger"
-def dockerArtifactName = project.hasProperty("dockerArtifactName") ? project.getProperty("dockerArtifactName") : "besu"
+def dockerBuildVersion = project.hasProperty('release.releaseVersion') && project.property('release.releaseVersion') ? project.property('release.releaseVersion') : "${rootProject.version}"
```
usmansaleem, line 830: What does this line achieve? `project.hasProperty('release.releaseVersion') && project.property('release.releaseVersion')`? @macfarla

.github/workflows/develop-pr.yml
```diff
@@ -0,0 +1,81 @@
+name: docker develop build and test only
+
+on:
```
usmansaleem, line 3: You can further enhance this by adding:
```
on:
  pull_request:
    branches:
      - main
```

this will allow this workflow to only trigger for PR requests against `main` branch.

.github/workflows/develop-pr.yml
```diff
@@ -0,0 +1,81 @@
...
+          DATE_TIME=$(date +"%y.%-m")
+          # Get the short SHA of the merge commit
+          SHORT_SHA=${GITHUB_SHA::7}
+          # Construct the build target name
+          BUILD_TARGET_NAME="${DATE_TIME}-develop-${SHORT_SHA}"
+          echo "Build Target Name: $BUILD_TARGET_NAME"
+          # Set the build target name as an environment variable
+          echo "BUILD_TARGET_NAME=${BUILD_TARGET_NAME}" >> $GITHUB_ENV
+
+      - name: Checkout Repo
```
usmansaleem, line 62: @macfarla Checkout Repo must (usually) be the first job i.e. should appear before Prepare.

usmansaleem (approved):
LGTM - few comments that can be incorporated.

build.gradle
macfarla: project.hasProperty is does it exist
project.property will give true if it's not null

kind of like getting a value from a java Optional

.github/workflows/develop-pr.yml
macfarla: I think in this case prepare has to be first because it sets up the platform and some environment variables.

build.gradle
macfarla: basically this set of changes was to make sure docker org name and artifact name never get set to empty

.github/workflows/develop-pr.yml
macfarla: done

PR #8848, opened 2025-06-24
Enable decoding for large newPayloads

siladu:
First pass maxes out StreamReadConstraints.maxStringLength for jackson decoding

TODO
We might want a more reasonable max value, and to only apply it to newPayload/engine API requests


siladu:
Replaced by https://github.com/hyperledger/besu/pull/8877

Issue #7704, opened 2024-10-01
Reimplement LivenessCheck and ReadinessCheck as plugins 


lu-pinto:
### Description
I noticed, while working on a plugin that required to extend the ReadinessCheck, that this check is implemented not through plugins but hardcoded as an RPC method during initialization.
It would be much cleaner if these checks can be implemented as plugins, as in the future if someone needs to do something similar, it would be easier and faster to do it.
- LivenessCheck is a trivial probe to check if process is UP or DOWN
- ReadinessCheck requires 2 checks:
1. Make sure that the sync is within a certain distance from the head of the chain. For this, a `numMaxBlocksBehind` parameter is used to check for distance. This check could be done in the plugin version by registering an `BesuEvent` for `addSyncStatusListener(SyncStatusListener)`.
2. Using a parameter `minPeers` check that the number of peers connected is higher than that value. In the plugin version, I can't see a way to have access to the number of peers but this could easily be exposed by `P2PService`.

### Acceptance Criteria
Readinees and Liveness checks are re-implemented without side effects. Both probes will be functioning the same as before but now implemented as plugins.

7suyash7:
Should this be implemented inside the `plugins` directory, with a new sub-directory called `health-checks`, or can they go inside the existing `rocksdb` directory?

lu-pinto:
IMO, we should go with a sub-directory inside `plugins`. `health` should be enough as a name.

marianakibuuka:
## I think this is an interesting issue for me to take on.

marianakibuuka:
I'm interested in the issue. Can I take it on ?

lu-pinto:
> I'm interested in the issue. Can I take it on ?

Yes all yours

macfarla:
@marianakibuuka any progress on this one?

jevinjojo:
@macfarla @lu-pinto I would like to work on this issue. Can get it assigned.

marianakibuuka:
yes please... take it on.. Thanks

bretk1989:
Do not close

fededlV:
was this issue resolved?  If it's not can I take this one?

marianakibuuka:
Nooo..u can take it on

fededlV:
Hi, thanks for confirming that we can take this on.

I noticed that PR #10167 already has most of the implementation in place, but there are still requested changes around the acceptance test and the plugin lifecycle.

We’d like to help with the remaining work. Our plan is to:

[...]

PR #8854, opened 2025-06-24
Pluggable Health Checks

jevinjojo:
## PR description
Refactors Besu’s health check system to use the plugin framework, making liveness and readiness checks fully pluggable

* Added health check interfaces to plugin-api
* Created plugin implementations for liveness and readiness checks
* Updated main app and plugins to use shared interfaces

## Fixed Issue  
- #7704


macfarla:
thanks for your PR @jevinjojo 
* to fix DCO you need to add signoff to each commit - follow the instructions under "details" of the DCO failure
* make sure to compile and run tests locally

jevinjojo:
@macfarla Ready for review.

macfarla (commented):
have run this locally and the endpoints are functioning as they did previously. would like @fab-10 to have a look from a holistic plugin POV

app/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -813,8 +815,20 @@ public Runner build() {
                   metricsSystem,
                   natService,
                   nonEngineMethods,
-                  new HealthService(new LivenessCheck()),
-                  new HealthService(new ReadinessCheck(peerNetwork, synchronizer))));
+                  new HealthService(
+                      besuPluginContext != null
```
fab-10, line 819: you can remove the `null` checks, since `besuPluginContext` could not be null that this point

app/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -813,8 +815,20 @@ public Runner build() {
...
+                              && besuPluginContext
+                                  .getService(LivenessCheckService.class)
+                                  .isPresent()
+                          ? besuPluginContext.getService(LivenessCheckService.class).get()
+                          : new LivenessCheck()),
+                  new HealthService(
+                      besuPluginContext != null
```
fab-10, line 826: s.a.

app/src/main/java/org/hyperledger/besu/RunnerBuilder.java
```diff
@@ -880,8 +894,20 @@ public Runner build() {
...
+                              && besuPluginContext
+                                  .getService(ReadinessCheckService.class)
+                                  .isPresent()
+                          ? besuPluginContext.getService(ReadinessCheckService.class).get()
+                          : new ReadinessCheck(peerNetwork, synchronizer))));
```
fab-10, line 910: better to avoid this code duplication, suggestion: encapsulate the health checks creation login in methods

plugins/health/HealthPluginServiceFactory.java
```diff
@@ -0,0 +1,20 @@
+package org.hyperledger.besu.plugins.health;
+
+import org.hyperledger.besu.plugin.services.BesuPluginContextImpl;
+import org.hyperledger.besu.plugin.services.PluginServiceFactory;
+import org.hyperledger.besu.plugin.services.p2p.P2PService;
+import org.hyperledger.besu.plugin.services.BesuEvents;
+import org.hyperledger.besu.plugin.services.health.LivenessCheckService;
+import org.hyperledger.besu.plugin.services.health.ReadinessCheckService;
+
+public class HealthPluginServiceFactory implements PluginServiceFactory {
```
fab-10, line 10: This part of the code is never called, since there is some missing glue to make it possible to call the plugin logic, when present, instead of the default logic embedded in Besu.

`PluginServiceFactory` seems only related to `BesuControllerBuilder`s that depend on the consensus algo used, so I would not use that to add the services, but instead I suggest that you review how other services solves this, like `PermissioningService`, `TransactionValidatorService` etc...

plugins/health/ReadinessCheckPluginService.java
```diff
@@ -0,0 +1,73 @@
...
+import org.hyperledger.besu.ethereum.api.jsonrpc.health.HealthService;
+import org.hyperledger.besu.plugin.services.p2p.P2PService;
+import org.hyperledger.besu.plugin.services.BesuEvents;
+import org.hyperledger.besu.plugin.data.SyncStatus;
+import org.hyperledger.besu.plugin.services.BesuService;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReadinessCheckPluginService implements ReadinessCheckService {
```
fab-10, line 12: Since this re-implement the `ReadinessCheck`, it will be nice to remove the latter, and make this the default when no external plugin providing the same is present. 
We have another example of this behavior that is the RocksDB plugin, that is shipped with Besu, but only enabled if no other plugin providing the db is present.

fab-10 (commented):
The idea of making health checks plugable make totally sense, and I think it should be extended in a way to allow for more plugins to provide their own health checks, so that all will be evaluated.
About the PR, there is some more work to do to glue the plugin, because at the moment it should never be called, I suggest to add a new acceptance tests to verify that it is actually called, there quite few example here `acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins`

macfarla:
@jevinjojo can you review and respond to PR review feedback, thanks!

jevinjojo:
@macfarla Would it be alright if I close this PR and open a new one with the updated changes?

macfarla:
> @macfarla Would it be alright if I close this PR and open a new one with the updated changes?

yes absolutely! maybe just ref this one if there's useful context. You should be able to close this since you opened it but lmk if you have issues

macfarla:
superseded by #9127

PR #8860, opened 2025-06-26
Adds an optional code size to the flat db account

Gabriel-Trintinalia:
## PR description
Adds an optional code size to the flat db account

Based on https://github.com/hyperledger/besu/pull/8779

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/worldview/BonsaiWorldState.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/worldview/BonsaiWorldState.java
@@ -184,11 +184,19 @@ public class BonsaiWorldState extends PathBasedWorldState {
               bonsaiUpdater -> bonsaiUpdater.removeAccountInfoState(addressHash));
         } else {
           final Hash addressHash = updatedAccount.getAddressHash();
-          final Bytes accountValue = updatedAccount.serializeAccount();
+          // Serialize the account and store it in the trie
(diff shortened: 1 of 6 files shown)


daniellehrner:
Karim just mentioned that we use the flat db account for responses as snap sync server. We need to remove the code size before sending the response.

Gabriel-Trintinalia:
> Karim just mentioned that we use the flat db account for responses as snap sync server. We need to remove the code size before sending the response.

We decode/encode into PMT already in `AccountRangeMessage`

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/trie/pathbased/bonsai/worldview/BonsaiWorldState.java
```diff
@@ -184,11 +184,19 @@ private void updateTheAccounts(
...
+          accountTrie.put(addressHash, accountValueTrie);
+
+          // Store the account in the flat database with metadata
           maybeStateUpdater.ifPresent(
-              bonsaiUpdater ->
-                  bonsaiUpdater.putAccountInfoState(hashAndSavePreImage(accountKey), accountValue));
-          accountTrie.put(addressHash, accountValue);
+              bonsaiUpdater -> {
+                // Serialize the account with code size for flat database storage
+                final Bytes accountValueFlat = updatedAccount.serializeAccountWithCodeSize();
```
matkt, line 195: can we see the impact on performance to have 2 serialize per account instead of one @ahamlat

matkt (commented):
LGTM regarding impat on Bonsai, we need some performance check before

PR #8863, opened 2025-06-26
fix(evmtool): implement --repeat option in state-test subcommand

### LINKED ISSUE #8855: Warm up state-tests
When comparing tests across clients, it can be useful to have a warmed up version of a state-test.

We can run a state-test with 
`./evmtool state-test my-test.json`

but in order to fake a warmup, we need to do:
`./evmtool state-test my-test.json my-test.json my-test.json my-test.json ....`

EvmToolCommand already has 
```
  @Option(
      names = {"--repeat"},
      description = "Number of times to repeat for benchmarking.")
``` 
but it isn't respected by the sub-sub-commands, 

We can also run benchmarks with warmup but they run specific precompile, not arbitrary state-tests:
`./evmtool benchmark --warmup 10000 --iterations 1000`

Preferred option would be to have the existing --repeat option respected, i.e. ability to run:
`./evmtool --repeat 100 state-test my-test.json` 
...and produce a single warmed up result.

But other options include adding a `--warmup` feature to `state-test` or giving `evmtool benchmark` the ability to run state tests.

Dyslex7c:
## PR description
The `--repeat` option was defined in EvmToolCommand but not used by StateTestSubCommand. This adds support for running state tests multiple times for warmup.

## Fixed Issue(s)

fixes #8855 

### DIFF
--- a/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
+++ b/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
@@ -265,6 +265,10 @@ public class EvmToolCommand implements Runnable {
       description = "display version info")
   boolean versionInfoRequested;
 
+  public Integer getRepeatCount() {
+    return repeat;
+  }
+
   static final Joiner STORAGE_JOINER = Joiner.on(",\n");
   private final EvmToolCommandOptionsModule daggerOptions = new EvmToolCommandOptionsModule();
   PrintWriter out;
--- a/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/StateTestSubCommand.java
+++ b/ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/StateTestSubCommand.java
@@ -189,18 +189,22 @@ public class StateTestSubCommand implements Runnable {
   }
 
   private void executeStateTest(final Map<String, GeneralStateTestCaseSpec> generalStateTests) {
-    for (final Map.Entry<String, GeneralStateTestCaseSpec> generalStateTestEntry :
-        generalStateTests.entrySet()) {
-      if (testName == null || testName.equals(generalStateTestEntry.getKey())) {
-        generalStateTestEntry
-            .getValue()
-            .finalStateSpecs()
-            .forEach((__, specs) -> traceTestSpecs(generalStateTestEntry.getKey(), specs));
+    int repeatCount = Math.max(1, parentCommand.getRepeatCount());
+    for (int i = 0; i < repeatCount; i++) {
+      boolean isLastIteration = (i == repeatCount - 1);
+      for (final Map.Entry<String, GeneralStateTestCaseSpec> generalStateTestEntry :
+          generalStateTests.entrySet()) {
+        if (testName == null || testName.equals(generalStateTestEntry.getKey())) {
+          generalStateTestEntry
+              .getValue()
+              .finalStateSpecs()
+              .forEach((__, specs) -> traceTestSpecs(generalStateTestEntry.getKey(), specs, isLastIteration));
+        }
       }
     }
   }
 
-  private void traceTestSpecs(final String test, final List<GeneralStateTestCaseEipSpec> specs) {
+  private void traceTestSpecs(final String test, final List<GeneralStateTestCaseEipSpec> specs, final boolean isLastIteration) {
     final OperationTracer tracer = // You should have picked Mercy.
         parentCommand.showJsonResults
             ? new StandardJsonTracer(
@@ -332,12 +336,14 @@ public class StateTestSubCommand implements Runnable {
         if (!result.getValidationResult().isValid()) {
           summaryLine.put("error", result.getValidationResult().getErrorMessage());
         }
-        if (parentCommand.showJsonAlloc) {
+        if (parentCommand.showJsonAlloc && isLastIteration) {
           EvmToolCommand.dumpWorldState(worldState, parentCommand.out);
         }
       }
 
-      parentCommand.out.println(summaryLine);
+      if (isLastIteration) {
+        parentCommand.out.println(summaryLine);
+      }
     }
   }
 }


macfarla:
@Dyslex7c you need to add signoff to each commit to fix the DCO check - instructions under "View details" on the failed check

Dyslex7c:
done @macfarla

macfarla:
@Dyslex7c spotless check is failing. Have a look at the checklist in the PR, there are some pointers to running checks locally

Dyslex7c:
@macfarla fixed spotless check

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -265,6 +265,10 @@ void setBytes(final String optionValue) {
       description = "display version info")
   boolean versionInfoRequested;
 
+  public Integer getRepeatCount() {
```
siladu:
```suggestion
  Integer getRepeatCount() {
```

nitpick: only needs to be package-private visibility

siladu:
@Dyslex7c thanks for the PR!

It appears to achieve the warmup nicely...

```
# no warmup: time = 20ms
time ./build/install/besu/bin/evmtool state-test ./ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/valid-state-test.json
{"output":"","gasUsed":"0x148e4","time":20916000,"Mgps":"4.025","test":"callcallcodecall_010_SuicideEnd","fork":"Istanbul","d":0,"g":0,"v":0,"stateRoot":"0x0ff25ed40ebe791efa00daa61bbb998db905b529fa0f8f2afb9e1aa69b63442c","postLogsHash":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","pass":true}
./build/install/besu/bin/evmtool state-test   4.24s user 0.18s system 111% cpu 3.953 total

# warmup: time = 0.02ms
time ./build/install/besu/bin/evmtool --repeat=20000 state-test ./ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/valid-state-test.json
{"output":"","gasUsed":"0x148e4","time":22291,"Mgps":"3777.130","test":"callcallcodecall_010_SuicideEnd","fork":"Istanbul","d":0,"g":0,"v":0,"stateRoot":"0x0ff25ed40ebe791efa00daa61bbb998db905b529fa0f8f2afb9e1aa69b63442c","postLogsHash":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","pass":true}
./build/install/besu/bin/evmtool --repeat=20000 state-test   14.28s user 0.83s system 217% cpu 6.949 total
```

siladu (changes requested):
Warmup is working, but when you use `--trace`, it prints out the repeated trace output. See example below.

Please can you test the other options to make sure they are compatible. I think your approach of only showing the latest iteration's output makes sense.

e.g. compare output of 

```
./build/install/besu/bin/evmtool state-test ./ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/valid-state-test.json --trace
```

and 

```
./build/install/besu/bin/evmtool state-test --repeat 10 ./ethereum/evmtool/src/test/resources/org/hyperledger/besu/evmtool/valid-state-test.json --trace
```

Dyslex7c:
@siladu I've made the changes you've mentioned and also made some other changes like handling repeated output for erroneous transactions, please do check the latest commit

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/StateTestSubCommand.java
```diff
@@ -189,20 +189,29 @@ public void run() {
...
   }
 
-  private void traceTestSpecs(final String test, final List<GeneralStateTestCaseEipSpec> specs) {
+  private void traceTestSpecs(
+      final String test,
+      final List<GeneralStateTestCaseEipSpec> specs,
+      final boolean isLastIteration) {
     final OperationTracer tracer = // You should have picked Mercy.
-        parentCommand.showJsonResults
+        parentCommand.showJsonResults && isLastIteration
```
siladu, line 214: One interesting quirk related to the change: In general, if you enable --trace then that impacts the execution time of the state test (as expected).
When using `--repeat`, only performing trace on the last iteration means we warm up the code without trace and then enable it, which could be unexpected.

If we were to measure the total iteration time and divide by the number of iterations, we would average out the impact from the impact of running trace (though last iteration would still pull up the average). I'm not sure if this is a good thing to do though as is also a bit misleading :)

I don't think this should block the PR though. I think usually you'd either be benchmarking or running trace to investigate, maybe not both together.
Need to think about it more and curious if you have any thoughts on this?

siladu (approved):
LGTM

siladu:
@Dyslex7c there are some flaky tests that should get fixed if you merge upstream main into your PR.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/StateTestSubCommand.java
Dyslex7c: yeah I made this change since this was the easiest possible workaround with minimal changes but I get what you're trying to say. The timing does get skewed while using `--repeat` keeping `--trace` enabled and averaging total time could be misleading due to the outlier effect of tracing in the last iteration. Although I'm not sure how big of an issue this really is, if it is then maybe we have to try and find a better approach.

Dyslex7c: yeah I made this change since this was the easiest possible workaround with minimal changes but I get what you're trying to say. The timing does get skewed while using `--repeat` keeping `--trace` enabled and averaging total time could be misleading due to the outlier effect of tracing in the last iteration. Although I'm not sure how big of an issue this really is, if it is then maybe we have to try and find a better approach.

Dyslex7c:
> @Dyslex7c there are some flaky tests that should get fixed if you merge upstream main into your PR.

merged

PR #8869, opened 2025-06-27
Change stack size growth strategy to 0.5 rate

lu-pinto:
This is a cherry-pick from the performance branch that we used to test the changes in the latest interop.

original PR: https://github.com/hyperledger/besu/pull/8814


evm/src/main/java/org/hyperledger/besu/evm/internal/FlexStack.java
```diff
@@ -163,12 +169,17 @@ public void push(final T operand) {
...
-      expandEntries(Math.min(currentCapacity + INCREMENT, maxSize));
+      final int newCapacity = newLength(currentCapacity, currentCapacity + 1, currentCapacity >> 1);
+      expandEntries(Math.min(newCapacity, maxSize));
     }
     entries[nextTop] = operand;
     top = nextTop;
   }
 
+  private static int newLength(final int oldCapacity, final int minGrowth, final int prefGrowth) {
+    return oldCapacity + Math.max(minGrowth, prefGrowth);
```
macfarla: is there a danger of int overflow

lu-pinto: Good catch @macfarla!
There was also another existing problem when someone sets `maxSize == Integer.MAX_VALUE` which is not a realistic value since the JVM is not able to allocate that many entries. You get something like this:
```
Exception in thread "main" java.lang.OutOfMemoryError: Requested array size exceeds VM limit
	at java.base/java.lang.reflect.Array.newArray(Native Method)
	at java.base/java.lang.reflect.Array.newInstance(Array.java:78)
	at FlexStack.expandEntries(FlexStack.java:124)
	at FlexStack.push(FlexStack.java:143)
	at Test.main(Test.java:6)
```

lu-pinto: Fixed now. Couldn't write tests because we would have to increase the heap size to 16G for such a case so I checked it separately offline.

evm/src/main/java/org/hyperledger/besu/evm/internal/FlexStack.java
```diff
@@ -163,12 +181,21 @@ public void push(final T operand) {
...
     if (nextTop >= currentCapacity) {
-      expandEntries(Math.min(currentCapacity + INCREMENT, maxSize));
+      final int newCapacity = newLength(currentCapacity, currentCapacity + 1, currentCapacity >> 1);
+      expandEntries(Math.min(newCapacity, maxSize));
     }
     entries[nextTop] = operand;
     top = nextTop;
   }
 
+  private int newLength(final int oldCapacity, final int minGrowth, final int prefGrowth) {
```
macfarla:
```suggestion
  private int newLength(final int currentCapacity, final int minGrowth, final int prefGrowth) {
```

evm/src/main/java/org/hyperledger/besu/evm/internal/FlexStack.java
```diff
@@ -30,8 +30,24 @@
...
+  /**
+   * Soft limit imposed for growing arrays. JVMs do not allow to allocate arrays above certain
+   * length and you will get the following exception if trying to do so:
+   *
+   * <p>java.lang.OutOfMemoryError: Requested array size exceeds VM limit
+   *
+   * <p>Therefore the maxSize of any stack is capped to this value. This max value is not arbitrary
+   * and was taken from OpenJDK.
+   */
+  private static final int SOFT_MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
```
macfarla:
```suggestion
  private static final int MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
```

evm/src/main/java/org/hyperledger/besu/evm/internal/FlexStack.java
macfarla: to make consistent with other changes

evm/src/main/java/org/hyperledger/besu/evm/internal/FlexStack.java
```diff
@@ -163,12 +181,21 @@ public void push(final T operand) {
...
-      expandEntries(Math.min(currentCapacity + INCREMENT, maxSize));
+      final int newCapacity = newLength(currentCapacity, currentCapacity + 1, currentCapacity >> 1);
+      expandEntries(Math.min(newCapacity, maxSize));
     }
     entries[nextTop] = operand;
     top = nextTop;
   }
 
+  private int newLength(final int oldCapacity, final int minGrowth, final int prefGrowth) {
+    final int growth = Math.max(minGrowth, prefGrowth);
```
macfarla: from line 184
      final int newCapacity = newLength(currentCapacity, currentCapacity + 1, currentCapacity >> 1);
won't `growth` always end up being currentCapacity + 1 ?
currentCapacity >> 1 gives 0.5 x currentCapacity which will be less?

lu-pinto:
After speaking with @ahamlat today, it's best to assess performance on mainnet first and do some benchmarking for CALL* before pulling this in

evm/src/main/java/org/hyperledger/besu/evm/internal/FlexStack.java
```diff
@@ -163,12 +181,21 @@ public void push(final T operand) {
...
+      final int newCapacity = newLength(currentCapacity, currentCapacity >> 1);
+      expandEntries(newCapacity);
     }
     entries[nextTop] = operand;
     top = nextTop;
   }
 
+  private int newLength(final int oldCapacity, final int prefGrowth) {
+    final int growth = Math.max(1, prefGrowth);
+    if (SOFT_MAX_ARRAY_LENGTH - growth < oldCapacity) {
```
macfarla: if you parameterised SOFT_MAX_ARRAY_LENGTH you could unit test this method

lu-pinto: True but the most important performance improvement is that it only grows maximum 6 times and I don't know how to do that. IMO a benchmark committed should be the way to go.

macfarla:
moving to draft since we need to do performance testing on this before merging

lu-pinto:
I think the added benchmark might unblock this PR, waiting on feedback.
I got the following results running on a `r6a.2xlarge`:

Without this PR:
```
Benchmark                     (stackDepth)  Mode  Cnt     Score    Error  Units
OperandStackBenchmark.fillUp             6  avgt   15    12.804 ±  0.021  ns/op
OperandStackBenchmark.fillUp            15  avgt   15    26.348 ±  0.028  ns/op
OperandStackBenchmark.fillUp            34  avgt   15   145.041 ±  7.118  ns/op
OperandStackBenchmark.fillUp           100  avgt   15   335.065 ±  0.573  ns/op
OperandStackBenchmark.fillUp           234  avgt   15   843.487 ± 16.104  ns/op
OperandStackBenchmark.fillUp           500  avgt   15  2102.937 ±  9.191  ns/op
OperandStackBenchmark.fillUp           800  avgt   15  4502.884 ± 17.090  ns/op
OperandStackBenchmark.fillUp          1024  avgt   15  6943.397 ± 18.922  ns/op
```

With this PR:
```
Benchmark                     (stackDepth)  Mode  Cnt     Score    Error  Units
OperandStackBenchmark.fillUp             6  avgt   15    22.623 ±  0.913  ns/op
OperandStackBenchmark.fillUp            15  avgt   15    30.296 ±  0.149  ns/op
OperandStackBenchmark.fillUp            34  avgt   15    58.666 ±  0.090  ns/op
OperandStackBenchmark.fillUp           100  avgt   15   277.010 ±  0.520  ns/op
OperandStackBenchmark.fillUp           234  avgt   15   700.040 ±  1.873  ns/op
OperandStackBenchmark.fillUp           500  avgt   15  1610.791 ±  1.557  ns/op
OperandStackBenchmark.fillUp           800  avgt   15  2632.757 ± 33.295  ns/op
OperandStackBenchmark.fillUp          1024  avgt   15  3150.211 ± 13.124  ns/op
```

It's noticeable the gains at a higher stack height as expected (x2). On the lower end, up until 32, since there's no resize, there's a slight regression. However, I would like to note that the average used stack height stands at around 75 as we can see below:
image
This is the maximum stack size expansion in each CALL - green line from a node closer to genesis and yellow line from a node closer to HEAD.

siladu (approved):
LGTM.

Regression tested on mainnet and no regressions found.
Also no obvious difference in performance detected.

siladu:
@lu-pinto maybe warrants a changelog entry

PR #8877, opened 2025-06-30
Enable decoding for large RPC requests

siladu:
This is to support large payloads (in engine_newPayload) at least up to 200MGas block sizes (and a bit more). Limit may need revisiting at larger block sizes.

~Max out StreamReadConstraints.maxStringLength for jackson decoding of request and parameters~

~Create Engine API specific EngineJsonRpcParserHandler and EngineJsonRpcParameter containing modified ObjectMapper.~

Go back to globally configured but using a 128MB equivalent limit instead of MAX_VALUE. This aligns with the bodyLimit we were already setting.

A version of this was tested on performance branch via https://github.com/hyperledger/besu/pull/8791 and https://github.com/hyperledger/besu/pull/8796
~but this PR restricts the changes to Engine API so shouldn't open up external DOS vectors since Engine API is assumed to be restricted to internal networks.~

## Why the large limit?

Using the large transaction full of zeros test cases from https://github.com/NethermindEth/gas-benchmarks/blob/main/tests-vm/TxDataZero30M_30M.txt :

```
30M:   ~15M chars = ~15MB
60M:   ~30M chars = ~30MB
200M: ~100M chars = ~100MB
# prediction:
300M: ~150M chars = ~150MB
```

128MB was already used as the engine API HTTP body limit, so lowering this is a potential breaking change. 
128MB is roughly equivalent to 128M chars with UTF-8 encoding.
128MB will support 200Mgas transactions which are used in the performance benchmarking suite.
We should maybe increase this limit even higher if we want to support a 300Mgas full of zeros.

## Testing

No visible impact on mainnet sync or operations

Screenshot 2025-07-25 at 6 06 24 am

Screenshot 2025-07-25 at 5 45 55 am

This PR is able to parse this large newPayload: https://github.com/NethermindEth/gas-benchmarks/blob/main/tests-vm/TxDataZero200M_200M.txt

Also wrote a custom test to cache this large payload and retrieve with getPayload.

Tested with [ethereum/go-ethereum@master/cmd/workload](https://github.com/ethereum/go-ethereum/tree/master/cmd/workload?rgh-link-date=2025-06-24T06%3A02%3A50Z) and behaves the same as control.


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/EngineJsonRpcParserHandler.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(EngineJsonRpcParserHandler.class);
+
+  private final ObjectMapper engineMapper;
+
+  public EngineJsonRpcParserHandler() {
+    this.engineMapper = new ObjectMapper();
+    // Configure with max string length for large payloads
+    StreamReadConstraints src =
+        StreamReadConstraints.builder().maxStringLength(Integer.MAX_VALUE).build();
```
siladu: We could make this a smaller value, but not sure what a good one would be. If we are happy with the assumption that only trusted sources (CLs) should call this API then probably MAX_VALUE is OK.

ahamlat: I was thinking about the same. We need to think about validation as not correlated to who is consuming the endpoint. But the issue here, is how can we make an accurate estimate ?

ahamlat:
When testing on mainnet, I go this error 
```
{"@timestamp":"2025-06-30T09:23:24,844","level":"ERROR","thread":"vert.x-worker-thread-
1","class":"JsonRpcExecutorHandler","message":"engine_forkchoiceUpdatedV3 - Error streaming JSON-RPC 
response","throwable":" java.io.IOException: io.vertx.core.http.HttpClosedException: Connection was closed\n\tat 
org.hyperledger.besu.ethereum.api.jsonrpc.JsonResponseStreamer.stopOnFailureOrClosed(JsonResponseStreamer.java:85)
\n\tat org.hyperledger.besu.ethereum.api.jsonrpc.JsonResponseStreamer.write(JsonResponseStreamer.java:57)\n\tat 
com.fasterxml.jackson.core.json.UTF8JsonGenerator._flushBuffer(UTF8JsonGenerator.java:2241)\n\tat 
com.fasterxml.jackson.core.json.UTF8JsonGenerator.close(UTF8JsonGenerator.java:1232)\n\tat 
com.fasterxml.jackson.databind.ObjectWriter._writeValueAndClose(ObjectWriter.java:1290)\n\tat 
com.fasterxml.jackson.databind.ObjectWriter.writeValue(ObjectWriter.java:1098)\n\tat org.hyperledger.besu.ethereum.api.handlers.JsonRpcObjectExecutor.handleJsonObjectResponse(JsonRpcObjectExecutor.ja
va:80)\n\tat 
org.hyperledger.besu.ethereum.api.handlers.JsonRpcObjectExecutor.execute(JsonRpcObjectExecutor.java:58)\n\tat 
org.hyperledger.besu.ethereum.api.handlers.JsonRpcExecutorHandler.lambda$handler$1(JsonRpcExecutorHandler.java:61)\
n\tat java.base/java.util.Optional.ifPresentOrElse(Optional.java:196)\n\tat 
org.hyperledger.besu.ethereum.api.handlers.JsonRpcExecutorHandler.lambda$handler$3(JsonRpcExecutorHandler.java:58)
\n\tat io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)\n\tat 
io.vertx.core.impl.ContextImpl.lambda$executeBlocking$5(ContextImpl.java:205)\n\tat 
io.vertx.core.impl.ContextInternal.dispatch(ContextInternal.java:270)\n\tat 
io.vertx.core.impl.ContextImpl$1.execute(ContextImpl.java:221)\n\tat 
io.vertx.core.impl.WorkerTask.run(WorkerTask.java:56)\n\tat 
java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)\n\tat 
java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)\n\tat 
io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)\n\tat 
java.base/java.lang.Thread.run(Thread.java:1583)\nCaused by: io.vertx.core.http.HttpClosedException: Connection was 
closed\n"}
```

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/EngineJsonRpcParserHandler.java
```diff
@@ -0,0 +1,99 @@
...
+import io.vertx.core.Handler;
+import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.RoutingContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EngineJsonRpcParserHandler {
```
ahamlat: I suggest adding a comment why this class is needed : because we need to be able to handle bigger payloads ...etc

ahamlat:
This PR needs at least the same level of testing as https://github.com/hyperledger/besu/pull/8847 as it is changing how newPayload requests are deserialized

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java
```diff
@@ -39,6 +43,10 @@ public static Handler<RoutingContext> handler() {
         try {
           ctx.put(ContextKey.REQUEST_BODY_AS_JSON_OBJECT.name(), ctx.getBodyAsJson());
```
ahamlat: It is not related to this PR, but I think we should harmonize both implementations. Instead of always trying de deserialize un Object and fallback to array, check if is a JsonObject or JsonArray before trying to deserialize it.
From my check, `isObject()` and `isArray()` are already calculated so there shouldn't any a lot of overhead on Object use case and should improve Array use case.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/EngineJsonRpcParserHandler.java
siladu: The best I can come up with is to use the largest payload size at X gas limit with Prague/Fusaka gas calculations - but what should that gas limit be without having to remember to extend this limit again in the future?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/EngineJsonRpcParserHandler.java
siladu: done

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java
siladu: I agree, would like to make the changes there as well.

Something to consider with my latest commits though. I had to introduce an intermediate map so there is some allocation overhead, and perhaps some extra processing since it's:
1. `String -> JacksonJsonNode` - similar deserialization step as before this PR.
2. `JacksonJsonNode -> Map` - this conversion might be the more expensive overhead
3. `Map -> VertxJsonObject` - think this is cheap, just wraps the map.

The first commit you reviewed did this:
```
// Convert Jackson JsonNode to Vert.x JsonObject
JsonObject jsonObject = new JsonObject(engineMapper.writeValueAsString(jsonNode));
```
but doing it that way instead of the map hit up against the internal Vertx ObjectMapper's 20K limit again.
It was also not great as was another reserialize/deserialize cycle.

Before this PR, it's only one step: `String -> VertxJsonObject`

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/EngineJsonRpcParserHandler.java
ahamlat: We can fix a higher one, like 200 million gas, and base the calculation on it. Add a comment to track it in the future.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java
ahamlat: Yes, that seems a lot of overhead actually. I see some possible alternatives :
1. From what I understand the 20 limit is related to this configuration option` -Dvertx.jackson.defaultReadMaxStringLength`, why not just increase this value and keep a simple deserialization ?
```
-Dvertx.jackson.defaultReadMaxStringLength=50000
```
We can do it in the code at startup
```
System.setProperty("vertx.jackson.defaultReadMaxStringLength", "50000");
 ```

2. If this solution is not possible, get the body length and apply the complex serialization only one we have large bodies (>19.5k), but I believe 1. should resolve the issue.

Comment (siladu) [reply ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java]: I was trying to avoid a global solution since all the RPCs use this and we only really need this for newPayload (and maybe getPayload?) and engine API is a trusted API.

If we limit to a smaller value maybe it's not such a big deal to do globally though.
That is essentially what I did on the performance branch https://github.com/hyperledger/besu/commit/0eb7aa6ad24ee4bb0212aa3fdc0865790527e5aa
but would want to move it to a more central location like startup as you suggest.

(2) is workable as well, I played with that idea too but I don't think it's worth it since smaller payloads wouldn't suffer from the extra serialization steps anyway.

Comment (ahamlat) [reply ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java]: On 2). we need to measure it to be sure. I think even payloads close to the limit can suffer from the extra serialization.

File: ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/EngineJsonRpcParserHandler.java (hunk 1 above)
Comment (siladu) [reply ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/EngineJsonRpcParserHandler.java]: Updated to use 128MB, see justification in the PR desc.

File: ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java (hunk 3 above)
Comment (siladu) [reply ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java]: As discussed, I went back to using a global setting (if engine API enabled).
This setting will apply to the shared vertx instance which both JSON-RPC and Engine-JSON-RPC use.

File: ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/JsonRpcConfiguration.java [hunk 4]
```diff
@@ -37,7 +37,7 @@ public class JsonRpcConfiguration {
   public static final int DEFAULT_ENGINE_JSON_RPC_PORT = 8551;
   public static final int DEFAULT_MAX_ACTIVE_CONNECTIONS = 80;
   public static final int DEFAULT_MAX_BATCH_SIZE = 1024;
-  public static final long DEFAULT_MAX_REQUEST_CONTENT_LENGTH = 5 * 1024 * 1024; // 5MB
+  public static final long DEFAULT_MAX_REQUEST_CONTENT_LENGTH = 128 * 1024 * 1024; // 128MB
```
ahamlat, line 40: Do we need two constants with the same name and same value in two different files ? Why not keep only one of these ?

siladu: Ideally, we should have one set of this config, but these duplicates are an existing issue, not to sort out for this PR IMO

PR #8886, opened 2025-07-01
fix noisy log message and add CHANGELOG entry

pinges:
## PR description
Changes a log message from info to debug, as well as adding a CHANGELOG entry

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteSyncBlocksTask.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteSyncBlocksTask.java
@@ -81,7 +81,7 @@ public class CompleteSyncBlocksTask extends AbstractCompleteBlocksTask<SyncBlock
     if (incompleteHeaders.isEmpty()) {
       return completedFuture(emptyList());
     }
-    LOG.info(
+    LOG.debug(
         "Requesting bodies to complete {} blocks, starting with {}.",
         incompleteHeaders.size(),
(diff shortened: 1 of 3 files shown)


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/CompleteSyncBlocksTask.java
```diff
@@ -81,7 +81,7 @@ CompletableFuture<List<SyncBlock>> requestBodies(final Optional<EthPeer> assigne
     if (incompleteHeaders.isEmpty()) {
       return completedFuture(emptyList());
     }
-    LOG.info(
+    LOG.debug(
```
siladu, line 84: LOG.atDebug()
will avoid calculating the args when debug is disabled

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/tasks/GetReceiptsForHeadersTask.java
```diff
@@ -91,9 +91,9 @@ private CompletableFuture<Map<BlockHeader, List<TransactionReceipt>>> requestRec
       return CompletableFuture.completedFuture(emptyMap());
     }
     LOG.debug(
```
siladu, line 93: LOG.atDebug()

siladu (changes requested):
Think we should revert the main PR and add this to the next attempt

macfarla:
> Think we should revert the main PR and add this to the next attempt

what's the original PR?

PR #8888, opened 2025-07-01
Improve transaction simulation when no gas pricing is present

fab-10:
## PR description

Follow up to https://github.com/hyperledger/besu/pull/8629

This PR helps getting more accurate simulations of transactions in case there is no gas pricing defined in call parameters.

Before when no gas pricing was defined, and the fee market requires a base fee, the only way to simulate a transaction was to force the tx validation parameters to _allow_exceeding_balance_ or _allow_underpriced_ , since all the gas pricing field of the tx were set to 0. That in some cases could result in less accurate simulations and gas estimations.

With this PR, if there is no gas pricing defined in call parameters and a base fee is required, then at the moment of creating the transaction for the simulation, the gas pricing fields are set to the base fee value of the block header used for the simulation. This way there is no more need to force tx validation parameters to force the simulation, other than the _allow_future_nonce_, increasing the accuracy.

If gas pricing parameters are present in the call parameters, then no adjustment is done, and if they are not enough for the current base fee market, or the sender has not enough balance, the simulation will fails, that is in line with Geth's behavior.

Moreover a new transaction simulation method is exposed, in the Plugin API, to allow plugin to directly simulate using call parameters, to avoid forcing plugins to build a transaction on their own and having to implement gas pricing setting rules that could be not aligned with Besu defaults.

### DIFF
--- a/app/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
+++ b/app/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
@@ -15,14 +15,13 @@
 package org.hyperledger.besu.services;
 
 import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulator;
-import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulatorAllowExceedingBalance;
-import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulatorAllowExceedingBalanceAndFutureNonce;
-import static org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams.transactionSimulatorAllowFutureNonce;
 
 import org.hyperledger.besu.datatypes.Hash;
 import org.hyperledger.besu.datatypes.StateOverrideMap;
 import org.hyperledger.besu.datatypes.Transaction;
 import org.hyperledger.besu.ethereum.chain.Blockchain;
+import org.hyperledger.besu.ethereum.mainnet.ImmutableTransactionValidationParams;
+import org.hyperledger.besu.ethereum.mainnet.TransactionValidationParams;
 import org.hyperledger.besu.ethereum.mainnet.ValidationResult;
 import org.hyperledger.besu.ethereum.processing.TransactionProcessingResult;
 import org.hyperledger.besu.ethereum.transaction.CallParameter;
@@ -34,6 +33,7 @@ import org.hyperledger.besu.plugin.data.ProcessableBlockHeader;
 import org.hyperledger.besu.plugin.data.TransactionSimulationResult;
 import org.hyperledger.besu.plugin.services.TransactionSimulationService;
 
+import java.util.EnumSet;
 import java.util.Optional;
 
 /** TransactionSimulationServiceImpl */
@@ -67,7 +67,7 @@ public class TransactionSimulationServiceImpl implements TransactionSimulationSe
       final Optional<StateOverrideMap> maybeStateOverrides,
       final Hash blockHash,
       final OperationTracer operationTracer,
-      final boolean isAllowExceedingBalance) {
+      final EnumSet<SimulationParameters> simulationParameters) {
 
     final CallParameter callParameter = CallParameter.fromTransaction(transaction);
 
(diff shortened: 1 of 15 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/CallParameter.java
```diff
@@ -38,35 +39,48 @@
 @Value.Immutable
 @JsonDeserialize(as = ImmutableCallParameter.class)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public abstract class CallParameter {
+public abstract class CallParameter implements CallParameters {
```
matkt: doesn't seems to be clear having `CallParameter implements CallParameters`

app/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
```diff
@@ -119,4 +122,32 @@ public Optional<TransactionSimulationResult> simulate(
             (org.hyperledger.besu.ethereum.core.ProcessableBlockHeader) pendingBlockHeader)
         .map(res -> new TransactionSimulationResult(transaction, res.result()));
   }
+
+  @Override
+  public Optional<TransactionSimulationResult> simulate(
+      final CallParameters callParameters,
```
matkt: Why not also using this CallParameters for this method https://github.com/hyperledger/besu/pull/8888/files#diff-59b579d2b76f894769cfbdf325ccb62f3aacea71816c5082f09f0161bce0a830R100-R124 ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/CallParameter.java
fab-10: indeed, I can rename the interface to `CallParameter` to match with the existing class

app/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
fab-10: did some refactoring, I hope it is better

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/CallParameter.java
ahamlat: Why the interface is introduced ?

datatypes/src/main/java/org/hyperledger/besu/datatypes/CallParameter.java
```diff
@@ -0,0 +1,161 @@
...
+ * Interface defining parameters for Ethereum transaction calls and simulations.
+ *
+ * <p>This interface provides access to all the parameters that can be specified when making calls
+ * to the Ethereum network, including both legacy and EIP-1559 transaction parameters, as well as
+ * EIP-4844 blob transaction parameters.
+ *
+ * <p>All parameters are optional to support various call scenarios where only a subset of
+ * parameters may be specified.
+ */
+public interface CallParameter {
```
ahamlat, line 34: I don't see another implementation other than ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/CallParameter.java so not sure to understand why is this introduced ?

app/src/main/java/org/hyperledger/besu/services/TransactionSimulationServiceImpl.java
matkt: regarding SimulationParameters. we cannot use also this one instead of using the boolean ?
```java
final boolean isAllowExceedingBalance
final boolean isAllowFutureNonce
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/CallParameter.java
matkt: is it not better to remove the Interface ? `CallParameter implements CallParameter` feel a bit hard to follow

datatypes/src/main/java/org/hyperledger/besu/datatypes/CallParameter.java
fab-10: The interface is needed just to expose the type in the plugin API, this is the standard way do that in plugin API

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/transaction/CallParameter.java
fab-10: The interface is needed just to expose the type in the plugin API, this is the standard way do that in plugin API, for example also `Transaction` is implemented like that.
We can review this approach, but prefer to keep that eventual refactoring separated from this PR.

ahamlat (approved):
LGTM

PR #8909, opened 2025-07-03
Expose methods to query the hardfork in the Plugin API

fab-10:
## PR description

Expose 2 new method in the Plugin API to let plugins query hardfork by blockheader or the hardfork of the next build block.

Since the `HardforkId` was already part of the Plugin API, it is now used internally when creating the milestone schedule in place of raw string names.

### DIFF
--- a/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeaderBuilder.java
+++ b/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeaderBuilder.java
@@ -133,6 +133,33 @@ public class BlockHeaderBuilder {
         .requestsHash(header.getRequestsHash().orElse(null));
   }
 
+  public static BlockHeaderBuilder fromHeader(
+      final org.hyperledger.besu.plugin.data.BlockHeader header) {
+    return create()
+        .parentHash(header.getParentHash())
+        .ommersHash(header.getOmmersHash())
+        .coinbase(header.getCoinbase())
+        .stateRoot(header.getStateRoot())
+        .transactionsRoot(header.getTransactionsRoot())
+        .receiptsRoot(header.getReceiptsRoot())
+        .logsBloom(new LogsBloomFilter(header.getLogsBloom()))
+        .difficulty(Difficulty.of(header.getDifficulty().getAsBigInteger()))
+        .number(header.getNumber())
+        .gasLimit(header.getGasLimit())
+        .gasUsed(header.getGasUsed())
+        .timestamp(header.getTimestamp())
+        .extraData(header.getExtraData())
+        .baseFee(header.getBaseFee().map(Wei::fromQuantity).orElse(null))
+        .mixHash(header.getMixHash())
+        .nonce(header.getNonce())
+        .prevRandao(header.getPrevRandao().orElse(null))
+        .withdrawalsRoot(header.getWithdrawalsRoot().orElse(null))
+        .blobGasUsed(header.getBlobGasUsed().orElse(null))
+        .excessBlobGas(header.getExcessBlobGas().map(BlobGas::fromQuantity).orElse(null))
+        .parentBeaconBlockRoot(header.getParentBeaconBlockRoot().orElse(null))
+        .requestsHash(header.getRequestsHash().orElse(null));
+  }
+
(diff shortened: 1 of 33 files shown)


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpec.java
```diff
@@ -181,8 +183,8 @@ public ProtocolSpec(
    *
    * @return the protocol specification name
```
macfarla:
```suggestion
   * @return the protocol specification hardfork ID
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSchedule.java
```diff
@@ -32,12 +32,13 @@ public interface ProtocolSchedule {
...
   default ProtocolSpec getForNextBlockHeader(
-      final BlockHeader parentBlockHeader, final long timestampForNextBlock) {
+      final org.hyperledger.besu.plugin.data.BlockHeader parentBlockHeader,
+      final long timestampForNextBlock) {
     final BlockHeader nextBlockHeader =
         BlockHeaderBuilder.fromHeader(parentBlockHeader)
             .number(parentBlockHeader.getNumber() + 1)
             .timestamp(timestampForNextBlock)
-            .parentHash(parentBlockHeader.getHash())
+            .parentHash(parentBlockHeader.getBlockHash())
```
macfarla, line 41: there's no difference between getHash and getBlockHash - right?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeaderBuilder.java
```diff
@@ -133,6 +133,33 @@ public static BlockHeaderBuilder fromHeader(final BlockHeader header) {
         .requestsHash(header.getRequestsHash().orElse(null));
   }
 
+  public static BlockHeaderBuilder fromHeader(
+      final org.hyperledger.besu.plugin.data.BlockHeader header) {
```
macfarla, line 137: BlockHeader implements org.hyperledger.besu.plugin.data.BlockHeader so I think you can change the param type for the existing method rather than duplicating it

line 118
  public static BlockHeaderBuilder fromHeader(final BlockHeader header) {

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSchedule.java
fab-10: Yes the same, but `getHash` is not defined in the interface

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeaderBuilder.java
fab-10: I left both because the former is optimized for the internal `BlockHeader` class, while this new one requires some extra work to convert from `Quantity` and other types.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpec.java
fab-10: done

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/EvmToolCommand.java
```diff
@@ -564,7 +564,7 @@ public void run() {
               .put("output", initialMessageFrame.getOutputData().toHexString())
               .put("gasUsed", "0x" + Long.toHexString(evmGas))
               .put("pass", initialMessageFrame.getExceptionalHaltReason().isEmpty())
-              .put("fork", protocolSpec.getName());
+              .put("fork", protocolSpec.getHardforkId().description());
```
siladu, line 567: Is this String identical to the former getName() or a potential breaking change?

fab-10: yes, I changed the description of the `HardforkId` enum where needed to match the string returned by the name method

PR #8917, opened 2025-07-04
introduce word256 and use it in add, exp, mul & sub op codes

daniellehrner:
## PR description

This PR introduces a new internal representation for 256-bit words in Besu: the Word256 class. It replaces the use of Bytes32 in selected EVM operations—specifically ADD, SUB, MUL and EXP—with Word256 for more efficient arithmetic and bitwise manipulation.

The Word256 class internally uses four long values (little-endian) to avoid heap allocations and improve performance over existing Bytes-based approaches.

At this stage, the EVM operand stack still provides and consumes Bytes objects. As such, values are temporarily converted to and from Word256 via Bytes32.wrap(...) and Word256.fromBytes(...). These conversions are expected to be phased out in future PRs as Word256 becomes the native representation within the EVM.

This change lays the groundwork for improving performance in critical execution paths by reducing allocations and enabling arithmetic operations that are more cache- and JIT-friendly.

Here are the benchmarks for the changed OP codes:

| OP code  | Baseline [ns / op] | PR [ns / op] |
| ------------- | ------------- | ------------- |
| ADD  | 128 | 95 | 
| SUB  | 136  | 89 |
| MUL  | 164  | 103 |
| EXP  | 4518  | 3224 |

Because the OP codes are migrated to Word256 in small batches to keep PRs small, there is currently some significant  overhead in converting Bytes from the stack to Word256 when popping and converting from Word256 to Bytes when pushing to the stack. This overhead is temporary and will be gone once the whole EVM including the OperandStack are migrated to Word256. After the full migrations the benchmark numbers are:

| OP code  | Baseline [ns / op] | Full migration [ns / op] |
| ------------- | ------------- | ------------- |
| ADD  | 128 | 18 | 
| SUB  | 136  | 13 |
| MUL  | 164  | 42 |
| EXP  | 4518  | 3119 |


evm/src/main/java/org/hyperledger/besu/evm/operation/AddOperation.java
```diff
@@ -50,19 +52,10 @@ public Operation.OperationResult executeFixedCostOperation(
...
-    int length = resultArray.length;
-    if (length > 32) {
-      frame.pushStackItem(Bytes.wrap(resultArray, length - 32, 32));
-    } else {
-      frame.pushStackItem(Bytes.wrap(resultArray));
-    }
+    final Word256 a = Word256.fromBytes(frame.popStackItem().toArrayUnsafe());
+    final Word256 b = Word256.fromBytes(frame.popStackItem().toArrayUnsafe());
 
+    frame.pushStackItem(Bytes32.wrap(a.add(b).toBytesArray()));
```
Matilda-Clerke, line 58: Minor nit: Might be cleaner to separate out the actual work of addition from wrapping and pushing onto the frame. The compiler should optimize it to the same opcodes anyway
```suggestion
    final Word256 a = Word256.fromBytes(frame.popStackItem().toArrayUnsafe());
    final Word256 b = Word256.fromBytes(frame.popStackItem().toArrayUnsafe());
    
    final Word256 result = a.add(b);

    frame.pushStackItem(Bytes32.wrap(result.toBytesArray()));
```

evm/src/main/java/org/hyperledger/besu/evm/operation/ExpOperation.java
```diff
@@ -51,30 +50,20 @@ public OperationResult execute(final MessageFrame frame, final EVM evm) {
...
       final MessageFrame frame, final GasCalculator gasCalculator) {
-    final Bytes number = frame.popStackItem();
-    final Bytes power = frame.popStackItem();
+    final Word256 number = Word256.fromBytes(frame.popStackItem().toArrayUnsafe());
+    final Word256 power = Word256.fromBytes(frame.popStackItem().toArrayUnsafe());
 
+    // Compute the number of non-zero bytes needed to represent the exponent.
+    // This is used to scale the gas cost of the EXP opcode linearly with exponent size,
+    // as defined in EIP-160: cost = 10 + (50 * numBytes)
     final int numBytes = (power.bitLength() + 7) / 8;
```
Matilda-Clerke, line 59: Out of interest, where does the +7 come from? and if it was to correct an artifact from `Bytes.bitLength()` is it still required?

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Constants.java
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.besu.evm.word256;
+
+/**
+ * Utility class for defining constants for Word256 values.
+ *
+ * <p>This class provides commonly used Word256 constants such as ZERO, ONE, MINUS_ONE, and MAX.
+ */
+final class Word256Constants {
```
Matilda-Clerke, line 22: Any reason we wouldn't just define these in `Word256`, similar to `BigInteger.ZERO` for example?

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Arithmetic.java
```diff
@@ -0,0 +1,192 @@
...
+   * Multiplies two Word256 values.
+   *
+   * @param a the first Word256 value
+   * @param b the second Word256 value
+   * @return the product of a and b as a new Word256
+   */
+  static Word256 mul(final Word256 a, final Word256 b) {
+    // Extract the limbs (64-bit words), least significant first (l0 = low, l3 = high)
+    final long x0 = a.l0, x1 = a.l1, x2 = a.l2, x3 = a.l3;
+    final long y0 = b.l0, y1 = b.l1, y2 = b.l2, y3 = b.l3;
```
Matilda-Clerke, line 89: Why not use them directly, like in `add` and `sub` methods?

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Arithmetic.java
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * @param base the base Word256 value
+   * @param exponent the exponent Word256 value
+   * @return the result of base^exponent as a new Word256
+   */
+  static Word256 exp(final Word256 base, final Word256 exponent) {
+    // Fast path: any number raised to the power of 0 is 1
+    if (exponent.isZero()) {
+      return Word256Constants.ONE;
+    }
```
Matilda-Clerke, line 160: Worth doing a fast path for `exponent.equals(Word256.ONE)`? It would just return `base` right?

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Arithmetic.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Multiplies two Word256 values.
+   *
+   * @param a the first Word256 value
+   * @param b the second Word256 value
+   * @return the product of a and b as a new Word256
+   */
+  static Word256 mul(final Word256 a, final Word256 b) {
```
Matilda-Clerke, line 86: Worth adding fast paths for when either word is 0 or 1?

- if either word is 0, result will always be 0
- if either word is 1, result will always be the other word

Could also consider a fast path for when one word is a power of 2 (i.e. 2^x) and just bitshift x times (assuming bitshift is more efficient than mul)

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Arithmetic.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Subtracts one Word256 value from another.
+   *
+   * @param a the Word256 value to subtract from
+   * @param b the Word256 value to subtract
+   * @return the result of a - b as a new Word256
+   */
+  static Word256 sub(final Word256 a, final Word256 b) {
```
Matilda-Clerke, line 56: Worth adding a fast path for when b is 0 to just return a?

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Arithmetic.java
```diff
@@ -0,0 +1,192 @@
...
+  private Word256Arithmetic() {}
+
+  /**
+   * Adds two Word256 values.
+   *
+   * @param a the first Word256 value
+   * @param b the second Word256 value
+   * @return the sum of a and b as a new Word256
+   */
+  static Word256 add(final Word256 a, final Word256 b) {
```
Matilda-Clerke, line 34: Worth adding a fast path for when either word is 0 to just return the other word?

Matilda-Clerke (commented):
Very cool change. I added some questions, probably just for my own interest.

I'll leave approval for someone with more experience in this area.

jframe:
Changing it back to draft as we are waiting for performance benchmarking

macfarla:
from @ahamlat re the perf testing - this PR is in good shape, someone just needs to work with @daniellehrner to get it over the line. @lu-pinto can take a look please

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/BinaryOperationBenchmark.java
siladu, line 1: cc @lu-pinto 
I think it would be nice to get a benchmarks-only PR in first, taking into account the discussion on  https://github.com/hyperledger/besu/pull/9025 (once it's finalised).
Then we can record a full benchmark report against main before comparing and including the improvements.

Not a blocker, and could still generate the report locally instead, just nice to have from my POV.

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/BinaryOperationBenchmark.java
```diff
@@ -0,0 +1,66 @@
...
+  public void setUp() {
+    frame = BenchmarkMessageFrameFactory.create();
+    aPool = new Bytes[SAMPLE_SIZE];
+    bPool = new Bytes[SAMPLE_SIZE];
+    RandomInputGenerator.fillPools(aPool, bPool);
+    index = 0;
+  }
+
+  @Benchmark
+  public Operation.OperationResult benchmark() {
```
siladu, line 51: Think it's a good idea to have an additional baseline() benchmark alongside each operation (or at least one to cover all BinaryOperations) that is just measuring the exact same frame.push and frame.pop operations but without the opcode.
Similar to https://github.com/hyperledger/besu/pull/9025/files#diff-72040e1f335d5d94e93946600c5443a642633577ad54fa4ab7d5bc2403efd6fcR105

This way we can easily see e.g. 
ADD takes 18ns
baseline takes 6ns
(and deduce ADD alone takes 12ns)

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Helpers.java
```diff
@@ -0,0 +1,169 @@
...
+    final long y0 = y & 0xFFFFFFFFL;
+    final long y1 = y >>> 32;
+
+    final long w0 = x0 * y0;
+    final long t = x1 * y0 + (w0 >>> 32);
+    final long w1 = t & 0xFFFFFFFFL;
+    final long w2 = t >>> 32;
+
+    final long lo = x * y;
+    final long hi = x1 * y1 + w2 + ((x0 * y1 + w1) >>> 32);
```
siladu, line 59: Think some inline comments would help understanding here

evm/src/main/java/org/hyperledger/besu/evm/word256/Word256Helpers.java
siladu, line 1: This class has some interesting but mysterious code - clearly we're sacrificing readability for performance, which is fine, but how should we evaluate that these are the most performant algorithms? I think citing some references and/or some more detail in the comments would go a long way.

ethereum/core/build.gradle
```diff
@@ -14,6 +14,7 @@
  */
 
 apply plugin: 'java-library'
+apply plugin: 'me.champeau.jmh'
```
lu-pinto, line 17: not required, can remove

daniellehrner:
Closed in favor of UIn256 improvements

PR #8920, opened 2025-07-04
Move Keccak256 benchmark to JMH and fix JMH integration

lu-pinto:
## PR description
This PR fixes the JMH integration that Besu already had so that we can write microbenchmarks again. I also added a benchmark for the keccak256 hashing so that I can demo the benchmarking since other benchmarks in there are currently broken and removed the old one from `Benchmarks.java`.

Output without profiler:
```
$ ./gradlew clean :ethereum:core:jmh -Pf=1 -Pwi=1 -Pi=5 -Pincludes=Keccak256
...
> Task :ethereum:core:jmh
# JMH version: 1.37
# VM version: JDK 21.0.2, OpenJDK 64-Bit Server VM, 21.0.2+13-58
# VM options: -XX:+EnableDynamicAgentLoading
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 1 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.hyperledger.besu.ethereum.vm.operations.Keccak256Benchmark.executeOperation

# Run progress: 0.00% complete, ETA 00:01:00
# Fork: 1 of 1
# Warmup Iteration   1: 2.298 ms/op
Iteration   1: 2.261 ms/opING [1m 3s]
Iteration   2: 2.262 ms/opING [1m 13s]
Iteration   3: 2.265 ms/opING [1m 23s]
Iteration   4: 2.268 ms/opING [1m 33s]
Iteration   5: 2.251 ms/opING [1m 43s]


Result "org.hyperledger.besu.ethereum.vm.operations.Keccak256Benchmark.executeOperation":
  2.261 ±(99.9%) 0.025 ms/op [Average]
  (min, avg, max) = (2.251, 2.261, 2.268), stdev = 0.006
  CI (99.9%): [2.236, 2.286] (assumes normal distribution)


# Run complete. Total time: 00:01:00

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

NOTE: Current JVM experimentally supports Compiler Blackholes, and they are in use. Please exercise
extra caution when trusting the results, look into the generated code to check the benchmark still
works, and factor in a small probability of new VM bugs. Additionally, while comparisons between
different JVMs are already problematic, the performance difference caused by different Blackhole
modes can be very significant. Please make sure you use the consistent Blackhole mode for comparisons.

Benchmark                            Mode  Cnt  Score   Error  Units
Keccak256Benchmark.executeOperation  avgt    5  2.261 ± 0.025  ms/op

Benchmark result is saved to /besu/ethereum/core/build/results/jmh/results.txt

[Incubating] Problems report is available at: file:///besu/build/reports/problems/problems-report.html
```

Output with profiler:
```
$ ./gradlew clean :ethereum:core:jmh -Pf=1 -Pwi=1 -Pi=5 -Pincludes=Keccak256 -PasyncProfiler=<libPath>
...
> Task :ethereum:core:jmh
# JMH version: 1.37
# VM version: JDK 21.0.2, OpenJDK 64-Bit Server VM, 21.0.2+13-58
# VM options: -XX:+EnableDynamicAgentLoading
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 1 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.hyperledger.besu.ethereum.vm.operations.Keccak256Benchmark.executeOperation

# Run progress: 0.00% complete, ETA 00:01:00
# Fork: 1 of 1
# Preparing profilers: AsyncProfiler
# Warmup Iteration   1: 2.340 ms/op
Iteration   1: 2.285 ms/opING [59s]
Iteration   2: 2.272 ms/opING [1m 9s]
Iteration   3: 2.256 ms/opING [1m 19s]
Iteration   4: 2.253 ms/opING [1m 29s]
Iteration   5: 2.267 ms/opING [1m 39s]
                 async: (text only)

# Processing profiler results: AsyncProfiler


Result "org.hyperledger.besu.ethereum.vm.operations.Keccak256Benchmark.executeOperation":
  2.267 ±(99.9%) 0.051 ms/op [Average]
  (min, avg, max) = (2.253, 2.267, 2.285), stdev = 0.013
  CI (99.9%): [2.216, 2.317] (assumes normal distribution)

Secondary result "org.hyperledger.besu.ethereum.vm.operations.Keccak256Benchmark.executeOperation:async":
Async profiler results:
  /besu/ethereum/core/org.hyperledger.besu.ethereum.vm.operations.Keccak256Benchmark.executeOperation-AverageTime/flame-cpu-forward.html
  /besu/ethereum/core/org.hyperledger.besu.ethereum.vm.operations.Keccak256Benchmark.executeOperation-AverageTime/flame-cpu-reverse.html


# Run complete. Total time: 00:01:00

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

NOTE: Current JVM experimentally supports Compiler Blackholes, and they are in use. Please exercise
extra caution when trusting the results, look into the generated code to check the benchmark still
works, and factor in a small probability of new VM bugs. Additionally, while comparisons between
different JVMs are already problematic, the performance difference caused by different Blackhole
modes can be very significant. Please make sure you use the consistent Blackhole mode for comparisons.

Benchmark                                  Mode  Cnt  Score   Error  Units
Keccak256Benchmark.executeOperation        avgt    5  2.267 ± 0.051  ms/op
Keccak256Benchmark.executeOperation:async  avgt         NaN            ---

Benchmark result is saved to /besu/ethereum/core/build/results/jmh/results.txt

[Incubating] Problems report is available at: file:///besu/build/reports/problems/problems-report.html
```

Profiler only records the measurements and ignores the warmup phase, however it uses the warmup phase to warmup the profiler itself which is great.
One can also add other options to the `async=` command using the `;` as a separator. For a list of available options (might not overlap with the async profiler itself as these need to be mapped jmh -> asyncProfiler), check here: https://github.com/openjdk/jmh/blob/master/jmh-core/src/main/java/org/openjdk/jmh/profile/AsyncProfiler.java#L85-L336


daniellehrner:
Would be great to have a documentation on how to run the benchmarks. It seems this is the right command:

```
./gradlew :ethereum:core:jmh -Pinclude=<CLASS_NAME>
```

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/Hashing20kKeccak256Benchmark.java
```diff
@@ -0,0 +1,43 @@
...
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+@State(Scope.Thread)
+@OutputTimeUnit(value = TimeUnit.MILLISECONDS)
+public class Hashing20kKeccak256Benchmark {
+
+  @Param({
+    "32",
```
ahamlat: Is 64 missing here ?

evm/src/test/java/org/hyperledger/besu/evm/precompile/Benchmarks.java
```diff
@@ -87,30 +87,6 @@ private static void benchSecp256r1Verify() {
     }
   }
 
-  private static void benchKeccak256() {
-    fakeFrame.expandMemory(0, 1024);
-    var gasCalculator = EvmSpec.evmSpec().getEvm().getGasCalculator();
-
-    for (int len = 0; len <= 512; len += 8) {
```
ahamlat, line 94: The new benchmark doesn't take into account intermediate bytes sizes like 24 bytes or 40 bytes. What do you think about keeping the same logic in JMH ?

daniellehrner:
I tried this in another PR and the downside with  this PR is that it hardcodes the benchmark mode in build.gradle and this can later not be overwritten by annotations. The PR hardcodes average time, which is most probably what we want. But if somebody wants throughput or single shot time this can’t be changed later without changing build.gradle.

Not sure if it's worth it to make it more flexible and how to do that.

lu-pinto:
> Would be great to have a documentation on how to run the benchmarks. It seems this is the right command:
> 
> ```
> ./gradlew :ethereum:core:jmh -Pinclude=<CLASS_NAME>
> ```

there is a comment in the `build.gradle` file that was modified with this PR: https://github.com/hyperledger/besu/pull/8920/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7L631

Not sure if you mean documentation directly in the task - for that I think we will need to change the plugin itself. This is the best I can do, how does it look?
```
 ./gradlew help --task jmh

> Configure project :
Generating project version as supplied is version not semver: unspecified

> Task :help
Detailed task information for jmh

Paths
     :ethereum:core:jmh
     :ethereum:eth:jmh
     :ethereum:rlp:jmh

Type
     JMHTask (me.champeau.jmh.JMHTask)

Options
     --rerun     Causes the task to be re-run even if up-to-date.

Description
     Usage: gradle jmh -Pf=2 -Pwi=3 -Pi=5 -Pincludes=MyBench -PasyncProfiler=<libPath>  -PasyncProfilerOptions=<options>

     Allows to control JMH execution directly from the command line.
        -Pf=<forks>     How many times to forks a single benchmark. Use 0 to disable forking altogether. Defaults to 2 forks.
        -Pwf=<warmupForks>      How many warmup forks to make for a single benchmark. 0 to disable warmup forks. Defaults to 0 forks.
        -Pwi=<warmupIterations> Number of warmup iterations to do. Defaults to 1.
        -Pi=<iterations>        Number of measurement iterations to do.
        -Pbm=<benchmarkMode>    Benchmark mode. Available modes are: [Throughput/thrpt, AverageTime/avgt, SampleTime/sample, SingleShotTime/ss, All/all]. Defaults to avgt.
        -Pincludes=<includePattern>     Include pattern (regular expression) for benchmarks to be executed. Defaults to including all benchmarks.
        -PasyncProfiler=<libPath>       Library path to fetch the Async profiler from. Default is to disable profiling.
        -PasyncProfilerOptions=<asyncProfilerOptions>   Options to pass on to the Async profiler separated by ';'. Default is to produce a flamegraph with all other default profiler options.
```

evm/src/test/java/org/hyperledger/besu/evm/precompile/Benchmarks.java
lu-pinto: We could add it but IMO there's no difference on those ranges anyway that I feel it's a waste of runtime:
```
Benchmark                                      (inputSize)  Mode  Cnt   Score   Error  Units
Hashing20kKeccak256Benchmark.executeOperation           32  avgt   10   4.513 ± 0.017  ms/op
Hashing20kKeccak256Benchmark.executeOperation           64  avgt   10   4.621 ± 0.018  ms/op
Hashing20kKeccak256Benchmark.executeOperation          128  avgt   10   4.784 ± 0.027  ms/op
Hashing20kKeccak256Benchmark.executeOperation          256  avgt   10   8.675 ± 0.115  ms/op
Hashing20kKeccak256Benchmark.executeOperation          512  avgt   10  16.140 ± 0.082  ms/op
```
Hence why I left it out. If you feel strongly about it let me know. Results from an M3 pro.

lu-pinto:
> I tried this in another PR and the downside with this PR is that it hardcodes the benchmark mode in build.gradle and this can later not be overwritten by annotations.

This is not true. You can override it in cli by using `-Pbm=<benchmarkMode>` and you can set it to whatever is supported in the JMH plugin. See description for gradle jmh now above.

>The PR hardcodes average time, which is most probably what we want. But if somebody wants throughput or single shot time this can’t be changed later without changing build.gradle.

Yes well spotted, that is the case indeed - what is set from CLI overrides what is set in annotations in-code. However, I believe that is already the case because we are using [JMH gradle plugin](https://github.com/melix/jmh-gradle-plugin) which sets defaults. IMO this is a problem with the JMH framework itself as what is set in-code should override any CLI settings. I guess one cannot please both worlds - so openJDK maintainers just went with precedence higher for CLI.
Compared to modifying the JMH gradle plugin or reimplenting the JMH integration ourselves, I don't feel this is deal breaker.

evm/src/test/java/org/hyperledger/besu/evm/precompile/Benchmarks.java
ahamlat: I don't feel strongly about it, but we can at least add 64 as I still don't see it, just to make it coherent.

lu-pinto: yeah sorry, didn't commit - added in now

gradle/verification-metadata.xml
```diff
@@ -5348,6 +5348,14 @@
...
       </component>
+      <component group="org.openjdk.jmh" name="jmh-generator-asm" version="1.37">
+         <artifact name="jmh-generator-asm-1.37.jar">
+            <sha256 value="de29bacc5c3a413215800f57de9017fdda1b3cb6e5359ea0c84ebe13c9610222" origin="Generated by Gradle"/>
+         </artifact>
+         <artifact name="jmh-generator-asm-1.37.pom">
+            <sha256 value="4e79c255b998e0df0bf7b5d9f9cf8ae4dd1c4bd0a462860aaa3598bc09a7f69b" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
       <component group="org.openjdk.jmh" name="jmh-generator-bytecode" version="1.36">
```
macfarla: nit: can remove previous version if not used

build.gradle
```diff
@@ -624,27 +632,42 @@ subprojects {
 
   if (file('src/jmh').directory) {
     apply plugin: 'me.champeau.jmh'
+    tasks.named("jmh") {
+      description = "Usage: gradle jmh -Pf=2 -Pwi=3 -Pi=5 -Pincludes=MyBench -PasyncProfiler=<libPath> -PasyncProfilerOptions=<options>\n" +
+        "\nRun JMH benchmarks in each of the projects. Allows for controlling JMH execution directly from the command line.\n" +
+        "\t-Pf=<forks>\tHow many times to forks a single benchmark. Use 0 to disable forking altogether. Defaults to 2 forks.\n" +
```
macfarla:
```suggestion
        "\t-Pf=<forks>\tHow many times to fork a single benchmark. Use 0 to disable forking altogether. Defaults to 2 forks.\n" +
```

daniellehrner:
> > Would be great to have a documentation on how to run the benchmarks. It seems this is the right command:
> > ```
> > ./gradlew :ethereum:core:jmh -Pinclude=<CLASS_NAME>
> > ```
> 
> there is a comment in the `build.gradle` file that was modified with this PR: https://github.com/hyperledger/besu/pull/8920/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7L631
> 
> Not sure if you mean documentation directly in the task - for that I think we will need to change the plugin itself. This is the best I can do, how does it look?
> 
> ```
>  ./gradlew help --task jmh
> 
> > Configure project :
> Generating project version as supplied is version not semver: unspecified
> 
> > Task :help
> Detailed task information for jmh
> 
> Paths
>      :ethereum:core:jmh
>      :ethereum:eth:jmh
>      :ethereum:rlp:jmh
> 
> Type
>      JMHTask (me.champeau.jmh.JMHTask)
> 
> Options
>      --rerun     Causes the task to be re-run even if up-to-date.
> 
> Description
>      Usage: gradle jmh -Pf=2 -Pwi=3 -Pi=5 -Pincludes=MyBench -PasyncProfiler=<libPath>  -PasyncProfilerOptions=<options>
> 
>      Allows to control JMH execution directly from the command line.
>         -Pf=<forks>     How many times to forks a single benchmark. Use 0 to disable forking altogether. Defaults to 2 forks.
>         -Pwf=<warmupForks>      How many warmup forks to make for a single benchmark. 0 to disable warmup forks. Defaults to 0 forks.
>         -Pwi=<warmupIterations> Number of warmup iterations to do. Defaults to 1.
>         -Pi=<iterations>        Number of measurement iterations to do.
>         -Pbm=<benchmarkMode>    Benchmark mode. Available modes are: [Throughput/thrpt, AverageTime/avgt, SampleTime/sample, SingleShotTime/ss, All/all]. Defaults to avgt.
>         -Pincludes=<includePattern>     Include pattern (regular expression) for benchmarks to be executed. Defaults to including all benchmarks.
>         -PasyncProfiler=<libPath>       Library path to fetch the Async profiler from. Default is to disable profiling.
>         -PasyncProfilerOptions=<asyncProfilerOptions>   Options to pass on to the Async profiler separated by ';'. Default is to produce a flamegraph with all other default profiler options.
> ```

@lu-pinto I was thinking to just add a BENCHMARK.md or similar and link from the the README.md to it. ChatGPT writes those kind of things very quickly.

In the task itself is also good, but I would like a way to document it in a markdown file first, for users to find the task to begin with.

That is not a blocker for me though, we can add it later as well.

build.gradle
lu-pinto: done

gradle/verification-metadata.xml
lu-pinto: done

BENCHMARKING.md
```diff
@@ -0,0 +1,77 @@
...
+
+To profile benchmarks with [Async Profiler](https://github.com/jvm-profiling-tools/async-profiler):
+
+```bash
+./gradlew :ethereum:core:jmh \
+  -Pincludes=SomeBenchmark \
+  -PasyncProfiler=/path/to/libasyncProfiler.so \
+  -PasyncProfilerOptions="flamegraph;output=flames.svg"
+```
+
```
ahamlat, line 59: Can you add here all the options we can see in the help command below ?

ahamlat (approved):
LGTM, the documentation can be improved but it is not blocking this PR.

BENCHMARKING.md
lu-pinto: I prefer not to, because they might change and then we have to modify them in this doc and in the help description

PR #8716, opened 2025-05-29
revised charter for discussion

jflo:
Attention all @hyperledger/besu-maintainers please review and comment on this proposed project charter, requested by LFDT.

PR #8931, opened 2025-07-08
Add majority vote rule to removal of maintainers

matthew1001:
## PR description

See the Besu Charter conversation at https://github.com/hyperledger/besu/pull/8716/files#r2121409576

On the most recent community call we agreed to keep maintainer addition/removal rules in one place, so rather than have them in both the charter and the `Maintainers.md`, we we have them in the latter and the charter would refer to it.

This PR updates `Maintainers.md` to include the text from the charter about requiring a majority of maintainers to vote on removal of an existing maintainer.

There is perhaps a follow on question about whether we do indeed want to require a majority approval to remove a maintainer. That would be a higher bar than is required to become a maintainer in the first place (3 votes + 0 vetoes). Should removing a maintainer be the same?

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -124,6 +124,8 @@ This can occur in the following situations:
     leave such as parental leave and medical leave.
 - Other unspecified circumstances.
 
+If one of the above reasons applies to an existing maintainer, removal of the maintainer is completed through a PR approved by a majority of the existing maintainers.
+
 As for adding a maintainer, the record and governance process for moving a
 maintainer to emeritus status is recorded using review approval in the PR making that change.
 


MAINTAINERS.md
```diff
@@ -124,6 +124,8 @@ This can occur in the following situations:
     leave such as parental leave and medical leave.
 - Other unspecified circumstances.
 
+If one of the above reasons applies to an existing maintainer, removal of the maintainer is completed through a PR approved by a majority of the existing maintainers.
```
macfarla, line 127: the voting for removal is the same as for the PR to add a new maintainer - ie lines 92-100. Majority or no veto within 2 weeks.

macfarla: the 2 lines below (127-128 in the original) allude to this but could be made more explicit. I would prefer updating that over adding this new paragraph.

macfarla: alternative proposal including my suggestions in #8937

matthew1001: Yeah I think your proposal in `8937` looks good. I've added a comment to it, and I'll close this PR with a link to yours

matthew1001:
Closing in lieu of PR https://github.com/hyperledger/besu/pull/8937

PR #8242, opened 2025-02-04
Sync block body

pinges:
## PR description
This PR changes the SNAP and CHECKPOINT sync to reduce the memory requirements by not decoding the block bodies requested from peers. This results in less memory and CPU usage during the sync.
This PR does make sure that the properties of the syncs are the same as before, e.g. transaction indexing and block added events haven't changed. 

Note: this PR does not significantly reduce the sync time on it's own, because the bottleneck is the import step of the download pipeline. It will however allow us to increase the parallelism of the pipeline because of the reduced memory and CPU usage. Once the import step is improved that increase in parallelism will allow us to reduce the sync time.

Note: Not decoding receipts during the sync does improve the memory and CPU usage as well. This has been shown in a POC.

First screenshot:
ScShotWithSBB

CPU% GC young generation during sync with this PR:
withSBB1_GCJ

CPU% GC young generation during sync without this PR:
WithoutSBB-GCJ-04

[...]

PR #8939, opened 2025-07-09
Revert revert sync bodies

pinges:
## PR description
This PR is reverting the revert of Sync block body (#8242) and adding the missing check so the pipeline is not restarted.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/AbstractGetBodiesFromPeerTask.java
```diff
@@ -84,11 +84,14 @@ protected Optional<List<T>> processResponse(
...
-      LOG.debug("Message contains no data. Peer: {}", peer);
+      LOG.atDebug().setMessage("Message contains no data. Peer: {}").addArgument(peer).log();
       return Optional.empty();
     } else if (bodies.size() > headers.size()) {
       // Message doesn't match our request - nothing to do
-      LOG.debug("Message doesn't match our request. Peer: {}", peer);
+      LOG.atDebug()
+          .setMessage("Message doesn't match our request. Peer: {}")
+          .addArgument(peer)
+          .log();
```
Matilda-Clerke: What an odd couple of changes. Is there a reason for this? The commit message states it's making logging more efficient, but I would have thought more method calls would ultimately be less efficient?

siladu: It's using the newer fluent API and delayed evaluation of the arguments.

Calling LOG.debug() will eagerly evaluate the args even if someone only has INFO level set.
Depending on the args, it might not always make a difference but just makes sense to always use atDebug() over debug() I think. 
In this case, not 100% sure the evaluation of EthPeer.toString() is being avoided but if you did something like this it would be:
`LOG.debug(peer.toString())`

siladu (changes requested):
I think this warrants a changelog highlighting the measured improvements

PR #8942, opened 2025-07-09
Copy eip196_g1_add cases and run with benchmark subcommand

siladu:
Signed-off-by: Simon Dudley <simon.dudley@consensys.net>## PR description


ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
```diff
@@ -141,6 +141,13 @@ enum Benchmark {
       scope = LOCAL)
   Optional<Integer> warmTime = Optional.empty();
 
+  @Option(
+      names = {"--warm-invert"},
```
lu-pinto, line 145: I don't think this is a right name. I was trying to figure out what this was and had to look into the code to find out. What about `warmup-data` and set default to true.

Also what if we make this the only way of doing warmups and remove the option? In a real world scenario I don't see the point of warming up data as well, even though Nethermind benchmarks might do a bit of that.

ahamlat: I think it is still interesting to have the options with data warm up, just to compare with other teams sharing other ways of benchmarking, with data warmup.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
```diff
@@ -141,6 +141,13 @@ enum Benchmark {
...
+      description =
+          "Run each test case within each warmup iteration, if the benchmark supports it. This attempts to warm the code without warming the data, i.e. avoids warming CPU caches. This does not affect exec-iterations so they should be set to 1 for this to work. Warm time should be increased as required as the warmup is frontloaded now and dependent on the number of test cases.",
```
ahamlat, line 147: Change the description to delete "This does not affect exec-iterations so they should be set to 1 for this to work."

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/benchmarks/BenchmarkExecutor.java
```diff
@@ -147,7 +154,7 @@ public BenchmarkExecutor(
    * @param evmSpecVersion EVM specification version to run the precompile for.
    */
   public void precompile(
-      final Map<String, Bytes> testCases,
+      final Map<String, Bytes> testCases, // TODO SLD enforce LinkedHashMap?
```
ahamlat, line 168: I don't understand the comment. Do you plan to change the implementation ?

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/BenchmarkSubCommand.java
siladu: yeh I don't like the name either... --cache-bust ?

So you're saying `warmup-data` with false being same as --warm-invert=true? That could work.
Happy to make it default behaviour but need to do some more testing to make sure it makes sense for the other benchmarks.

The issue with not warming the data is things like the time to produce some output becomes dependent on the test case content. Generally increases the total runtime significantly as well. 
For example Ec pairings are really slow when the data isn't warm.
Not saying we shouldn't do it, just that we might need to adjust the tool to accommodate, e.g. warmTime and execTime (I've left that out for now, still a WIP).

Another issue is if there's only one test case we will be warming the data still.

ethereum/evmtool/src/main/java/org/hyperledger/besu/evmtool/benchmarks/BenchmarkExecutor.java
siladu: This is a draft WIP, so yes :)

PR #8948, opened 2025-07-10
Add description for profiling

daniellehrner:
## PR description

Explains how to do profiling for Besu, including setup, profiling depending on CPU architecture and some brief hint where the most interesting parts in the flamegraph are.


PROFILING.md
```diff
@@ -0,0 +1,99 @@
...
+
+This guide explains how to set up and run [Async Profiler](https://github.com/jvm-profiling-tools/async-profiler) to profile a running Besu instance.
+
+---
+
+## 1. Setup
+
+### a. Download and Extract
+
+Download version 4.0 of Async Profiler appropriate for your architecture:
```
ahamlat: We need to refer to Async profiler documentation to download the last release.

PROFILING.md
```diff
@@ -0,0 +1,99 @@
...
+- `-t`: Include thread stack traces.
+- `-i 11ms`: Sampling interval.
+- `-f`: Output HTML flamegraph path.
+
+---
+
+## 4. Analyze the Results
+
+Once complete, open the generated flamegraph (e.g., `/tmp/besu-profile.html`) in your browser to analyze performance bottlenecks.
+
```
ahamlat:
```suggestion
**How to read the flamegraph ? **

- Top boxes: These represent the methods currently running when CPU samples were taken.
- Boxes underneath: These represent methods called earlier, eventually leading to the methods above them.

The width of each box shows how much total time was spent in this method (including any methods it called).

So:

- Wide boxes = Potential hot spots.
- Tall stacks = Deep call chains.

```

PROFILING.md
daniellehrner: I've added a link to their release page and left the wget command for version 4.0 as an example

PROFILING.md
```diff
@@ -0,0 +1,112 @@
...
+- `-t`: Include thread stack traces.
+- `-i 11ms`: Sampling interval.
+- `-f`: Output HTML flamegraph path.
+
+---
+
+## 4. Analyze the Results
+
+Once complete, open the generated flamegraph (e.g., `/tmp/besu-profile.html`) in your browser to analyze performance bottlenecks.
+**How to read the flamegraph ? **
```
ahamlat: For some reason, making it bold didn't work, could you check the preview ?

siladu:
Can docs like this live under /docs rather than top level?
Think we should move BENCHMARKING.md too

PROFILING.md
```diff
@@ -0,0 +1,112 @@
...
+sudo sysctl kernel.perf_event_paranoid=1
+```
+
+These settings may need to be re-applied after a reboot.
+
+---
+
+## 2. Determine CPU Architecture
+
+Check system architecture to choose the correct profiling event:
```
siladu: no reason you can't use -e cpu on x86

PROFILING.md
```diff
@@ -0,0 +1,112 @@
...
+  -t \
+  -i 11ms \
+  -f /tmp/besu-profile.html \
+  <PID>
+```
+
+### Flag explanation
+
+- `-d 300`: Profile duration in seconds.
+- `-e wall|cpu`: Profiling event based on architecture.
```
siladu: I would say it's based on whether you want to measure on-CPU time versus the wall clock time, architectural support is incidental.

PROFILING.md
```diff
@@ -0,0 +1,112 @@
...
+- `-t`: Include thread stack traces.
```
siladu: -t used with -e wall, not sure it's the best option to use with -e cpu

README.md
```diff
@@ -48,6 +48,13 @@ Instructions for how to get started with developing on the Besu codebase. Please
 * [Code Coverage](https://lf-hyperledger.atlassian.net/wiki/spaces/BESU/pages/22154288/Code+coverage)
 * [Logging](https://lf-hyperledger.atlassian.net/wiki/spaces/BESU/pages/22154291/Logging) or the [Documentation's Logging section](https://besu.hyperledger.org/public-networks/how-to/monitor/logging)
 
+### Profiling Besu
+
+Besu supports performance profiling using [Async Profiler](https://github.com/async-profiler/async-profiler), a low-overhead sampling profiler.  
+You can find setup and usage instructions in the [Profiling Guide](PROFILING.md).
+
+Profiling can help identify performance bottlenecks in block processing, transaction validation, and EVM execution.  
+Please ensure the profiler is run as the same user that started the Besu process.
```
ahamlat, line 57: Could you add a troubleshooting section and refer to the one from Async Profiler documentation : https://github.com/async-profiler/async-profiler/blob/5fffdb1eaa538b20e5990ca6b96898ffa157fc91/docs/Troubleshooting.md

ahamlat: Also add a mention to say that there're some issues that are related to the JDK, other issues that are already fixed in the nightly build. If your JVM crashes because Async Profiler check the error you had in the logs in[ async-profiler issues](https://github.com/async-profiler/async-profiler/issues) to see if there is a workaround, a fix to install from the JDK or a fix in the nightly build.

PROFILING.md
ahamlat: I think it is still useful to see which thread is the one that is most on CPU for example.

ahamlat:
```suggestion
- `-t`: Profile threads separately.
```

ahamlat: I suggest a more clear description.

PROFILING.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+Should you need to run it as a specific user (e.g., `execution`), use:
+
+```bash
+sudo -u execution /opt/besu/async-profiler-4.0-linux-*/bin/asprof \
+  -d 300 \
+  -e [wall|cpu] \
+  -t \
+  -i 11ms \
```
ahamlat: I had few crashes on ARM even with -e CPU, I suggest add `--cstack vm` as suggested in this comment by Andrei Pangin
https://github.com/async-profiler/async-profiler/issues/1364#issuecomment-3046940172

docs/PROFILING.md
```diff
@@ -0,0 +1,128 @@
...
+## Troubleshooting
+
+If you experience any issues while using the profiler, refer to the official troubleshooting guide for solutions and common pitfalls:
+[Async Profiler Troubleshooting Guide](https://github.com/async-profiler/async-profiler/blob/5fffdb1eaa538b20e5990ca6b96898ffa157fc91/docs/Troubleshooting.md)
+
+### Additional Notes
+* JDK-specific issues: Some problems are caused by bugs or limitations in the JDK itself (e.g., AsyncGetCallTrace stability). Always check if the issue is known for your JDK version.
+* Nightly Builds: Async Profiler regularly fixes issues in nightly builds that are not yet available in the latest stable release. If you encounter a problem, especially a crash, consider testing with the nightly build of Async Profiler.
+* JVM Crashes: If your JVM crashes when using Async Profiler, carefully inspect the crash logs. You can often find:
+* A known issue already discussed on the Async Profiler GitHub Issues.
```
ahamlat: Could you make these sublist items ?

ahamlat (approved):
LGTM

PR #8486, opened 2025-03-28
Modify and add copies of few ATs using QBFT nodes instead of PoW

pullurib:
## PR description
Adding version of ATs using QBFT nodes to check stability before replacing PoW ATs

### Thanks for sending a pull request! Have you done the following?

- [x] Checked out our [contribution guidelines](https://github.com/hyperledger/besu/blob/main/CONTRIBUTING.md)?
- [x] Considered documentation and added the `doc-change-required` label to this PR [if updates are required](https://wiki.hyperledger.org/display/BESU/Documentation).
- [x] Considered the changelog and included an [update if required](https://wiki.hyperledger.org/display/BESU/Changelog).
- [ ] For database changes (e.g. KeyValueSegmentIdentifier) considered compatibility and performed forwards and backwards compatibility tests

### Locally, you can run these tests to catch failures early:

- [x] spotless: `./gradlew spotlessApply`
- [x] unit tests: `./gradlew build`
- [x] acceptance tests: `./gradlew acceptanceTest`
- [ ] integration tests: `./gradlew integrationTest`
- [ ] reference tests: `./gradlew ethereum:referenceTests:referenceTests`

PR #8949, opened 2025-07-10
Remove PoW ATs which were replaced by QBFT miner based ATs

pullurib:
## PR description
As part of PoW deprecation, some of the ATs using PoW miners had to be replaced with those using other consensus. Corresponding QBFT ATs were added few weeks ago and have been stabilized. So deleting the original ATs now. 

PRs which added new QBFT based ATs : 

- https://github.com/hyperledger/besu/pull/8486
- https://github.com/hyperledger/besu/pull/8689


acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins/ExtendTransactionValidatorPluginTest.java
```diff
@@ -79,15 +82,15 @@ public void nonFrontierTransactionsAreNotAccepted() {
         .hasMessage("Plugin has marked the transaction as invalid");
   }
 
-  @Test
+  /*  @Test
```
macfarla: should this commented out test be removed or reinstated?

pullurib: I believe it should be removed and wanted to check about it . @fab-10 , changing this test from using PoW dev node to QBFT node. QBFT blockcreator doesn't seem to add invalid transactions to block as PoWblockcreator . Should I remove this test ?

fab-10: This test should stay, since it verify that extending the tx validation via plugin works.

Actually the test should be consensus independent, so let's understand why it is not passing with QBFT

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins/BlockchainServicePluginTest.java
```diff
@@ -61,44 +63,4 @@ public void hardforkListIsCorrect() throws IOException {
       assertThat(fileContents.get(i)).isEqualTo(expectedLines.get(i));
     }
   }
-
-  private static final String GENESIS_CONFIG =
```
fab-10, line 65: This custom genesis is needed to create a hardfork schedule to test querying hardforks from plugins, has it been replaced by something else with similar behavior?

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins/ExtendTransactionValidatorPluginTest.java
pullurib: I'll check the blockcreation differences

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins/BlockchainServicePluginTest.java
pullurib: yeah, using a qbft genesis with same forks as this one

acceptance-tests/tests/src/test/java/org/hyperledger/besu/tests/acceptance/plugins/ExtendTransactionValidatorPluginTest.java
matthew1001: Just paging in a little to the conversation. Is it that case that the test checks that plugins can extend validation, and previously it happened to use PoW which allowed incorrect TXNs into blocks. But now we don't have PoW we're using QBFT, which doesn't allow them?

If that's the case, perhaps we need the test to use a non-QBFT plugin node, because if QBFT shouldn't allow these transactions then we can't use it as the plugin for the test.

I might have misunderstood though.

fab-10: This tx validation plugin is trivial, it just accepts frontier txs as valid, and reports all the other types as invalid, so there should be no difference when changing consensus algo

pullurib: Added back the testcase using 4 qbft nodes in cluster. With just two nodes using QBFT, this won't work as both of them have to agree on the block . Using 4 nodes, one of them with the extra transaction validator, 3 would import block and plugin node marks as bad block

macfarla:
Looking at the qbft ATs vs PoW original ones
There's 14 that were added under QBFT so those match 14 of the classes deleted in this PR
```
CreateAccountAcceptanceTest.java
ClusterAcceptanceTest.java
ClusterThreadNodeRunnerAcceptanceTest.java
DeploySmartContractAcceptanceTest.java
DeployTransactionAcceptanceTest.java
EthEstimateGasAcceptanceTest.java
EthEstimateGasZeroToleranceAcceptanceTest.java
EthSendRawTransactionAcceptanceTest.java
EventEmitterAcceptanceTest.java
RevertReasonAcceptanceTest.java
MiningAcceptanceTest.java
NodeLocalConfigPermissioningAcceptanceTest.java
BesuEventsPluginTest.java
NewPendingTransactionAcceptanceTest.java
EthEstimateGasAcceptanceTest.java
EthEstimateGasZeroToleranceAcceptanceTest.java
```
However there's also 1 that is deleted in this PR but not added under QBFT - can this test instead be adapted for QBFT?
```
BlockchainServiceFinalizedBlockPluginTest.java
```

pullurib:
> 
> However there's also 1 that is deleted in this PR but not added under QBFT - can this test instead be adapted for QBFT?
> 
> ```
> BlockchainServiceFinalizedBlockPluginTest.java
> ```

Sorry about the oversight, thanks for checking!  Added the modified AT back .

PR #8959, opened 2025-07-15
Fix the problem we found with #8242

pinges:
## PR description
This PR is adding a check whether a block has already been imported before we are trying to import it. This fixes the problem that we found with PR #8242, which was reverted. This PR reverts that revert.


jframe:
Changes look fine. Did you retest this on some nodes? And were number of restarts in the pipeline decreased?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Add option to trace reference tests during execution [#8878](https://github.com/hyperledger/besu/pull/8878)
 - Expose methods to query hardfork by block header or for the next block in the Plugin API [#8909](https://github.com/hyperledger/besu/pull/8909)
 - Generate distribution dependencies catalog [#8987](https://github.com/hyperledger/besu/pull/8987)
+- Improve the sync perfomance by not RLP decoding bodies during sync. This means we are using less memory and CPU, allowing us to increase the prallelism of the download pripeline. [#8959]
```
siladu:
```suggestion
- Improve the sync performance by not RLP decoding bodies during sync. This means we are using less memory and CPU, allowing us to increase the parallelism of the download pipeline, which has been increased from 4 to 8. Can be reduced again with  `--Xsynchronizer-downloader-parallelism=4` [#8959]
```

siladu (approved):
LGTM - thanks for the extra test results!

PR #8961, opened 2025-07-15
Fix Vert.x deprecation warnings

diega:
## PR description
As part of upgrading to Vert.x 5, following the [official migration guide](https://vertx.io/docs/guides/vertx-5-migration-guide/#_handling_deprecations_and_removals), the first step is to address deprecation warnings.

This PR resolves most of the deprecation warnings in the codebase.

#### Pending Analysis
One deprecation remains that requires further investigation:
- The usage of [`io.vertx.core.Vertx#nettyEventLoopGroup`](https://vertx.io/docs/4.5.16/apidocs/io/vertx/core/Vertx.html#nettyEventLoopGroup--) in [`VertxPeerDiscoveryAgent`](https://github.com/hyperledger/besu/blob/25.7.0/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/VertxPeerDiscoveryAgent.java#L107) class. A suitable replacement for this method has not been identified yet.


ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/TimeoutHandler.java
```diff
@@ -54,7 +54,7 @@ private static void processHandler(
                             timeoutOptions.getTimeoutMillis(),
                             t -> {
                               ctx.fail(timeoutOptions.getErrorCode());
-                              ctx.response().close();
```
macfarla, line 57: is this equivalent?

macfarla (commented):
@diega thanks for the PR, changes look ok from a brief review -  have you tested out that these changes work on a running node with the various configs eg websocket auth etc?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/TimeoutHandler.java
diega: Yes, I think so. The [`i.v.c.http.HttpServerResponse#close`](https://vertx.io/docs/4.5.16/apidocs/io/vertx/core/http/HttpServerResponse.html#close--) doc says:
> instead use [HttpConnection.close()](https://vertx.io/docs/4.5.16/apidocs/io/vertx/core/http/HttpConnection.html#close--) or [reset(long)](https://vertx.io/docs/4.5.16/apidocs/io/vertx/core/http/HttpServerResponse.html#reset-long-)

and I thought that closing the connection looked clearer than calling `reset`

diega:
Hey @macfarla, I haven’t tested WebSocket scenarios beyond the newly added `WebSocketHandshakeTest`. What I did manually test so far was that a node starts and syncs correctly. Let me know if you think there's any specific WebSocket config or flow that’s worth testing manually.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLHttpService.java
```diff
@@ -398,14 +399,7 @@ private void handleGraphQLRequest(final RoutingContext routingContext) {
...
       vertx.executeBlocking(
-          future -> {
-            try {
-              final GraphQLResponse graphQLResponse = process(query, operationName, variables);
-              future.complete(graphQLResponse);
-            } catch (final Exception e) {
-              future.fail(e);
-            }
-          },
+          () -> process(query, operationName, variables),
```
usmansaleem: How would a failure in process be handled?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java
```diff
@@ -33,14 +33,14 @@ private JsonRpcParserHandler() {}
   public static Handler<RoutingContext> handler() {
     return ctx -> {
       final HttpServerResponse response = ctx.response();
-      if (ctx.getBody() == null) {
+      if (ctx.body().buffer() == null) {
```
usmansaleem: I think `body().isEmpty()` is much safer. As it internally checks for null body.

usmansaleem:
@diega @macfarla Re VertxPeetDiscoveryAgent. Vertx will eventually move to Netty 5, hence Vertx deprecated internal Netty references from its public API. https://vertx.io/docs/guides/vertx-5-migration-guide/#_netty_type_usage_removals

We can get around it as suggested in above link, however, I reckon we either deprecate this particular metrics or provide conditional code which should detect underlying netty library and use appropriate method to obtain pending tasks metrics.

macfarla:
@diega some comments for you to respond to; have assigned this PR back to you for action

diega:
Hey @macfarla. Thank you for the heads up. I'm travelling now, so I'll work on this during next week.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/graphql/GraphQLHttpService.java
diega: Any exception fired in the context of the async execution results in a res.failed() in the result handler. I'm adding a test for this case though.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java
diega: good call, I'll change it

diega:
@macfarla, comments addressed. I'll move it back to ready for review

macfarla:
I synced hoodi with this PR, just with default config so didn't test any websocket logic. doesn't seem to break anything

usmansaleem:
@diega Line # 107 [VertxPeerDiscoveryAgent](https://github.com/hyperledger/besu/blob/25.7.0/ethereum/p2p/src/main/java/org/hyperledger/besu/ethereum/p2p/discovery/VertxPeerDiscoveryAgent.java#L107) needs change similar to following because as soon as we move to vertx 5, the current code probably won't compile:
```
 pendingTaskCounter(vertx.nettyEventLoopGroup()));
```
should become:
```
// Note: Netty event loop via Vertx Internal, subject to change.
var group = ((VertxInternal)vertx).nettyEventLoopGroup();
pendingTaskCounter(group);
```
The above should keep working till Vertx change the Netty library version and move it to some other method. At that time, we would get compile time error to sort it out.

@macfarla FYI.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/EngineJsonRpcService.java
```diff
@@ -537,8 +561,8 @@ private void applyTlsConfig(final HttpServerOptions httpServerOptions) {
       try {
         httpServerOptions
             .setSsl(true)
-            .setPfxKeyCertOptions(
-                new PfxOptions()
```
macfarla, line 545: are these equivalent

macfarla: @jflo or @diega do we have any known TLS setups that we can get to test this doesn't break anything?

macfarla (commented):
I would like to be sure that we aren't breaking any features that we aren't using on our canaries - TLS and WebSockets. @jflo do we have contacts with anyone using these that can verify these changes?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/EngineJsonRpcService.java
diega: You are correct, the change wasn't equivalent.
I had originally replaced `PfxOptions` with `JksOptions`, which would have changed the certificate format from **PKCS#12** to **JKS**. Now I’ve kept `PfxOptions` and only replaced the `setPfxKeyCertOptions` call with `setKeyStoreOptions`. This makes the new change equivalent to the original behavior (see [i.v.c.n.TCPSSLOptions#setPfxKeyCertOptions](https://github.com/eclipse-vertx/vert.x/blob/4.5.22/src/main/java/io/vertx/core/net/TCPSSLOptions.java#L521)). Thanks for catching that

diega:
Hey @macfarla, I noticed you made a commit in #9396 that was related to this PR. I thought it would be cleaner to keep the deprecation-fix changes together here, but if you’d rather keep them in separate PRs just let me know and I can split them out without any issue.

diega:
@usmansaleem I feel that adding the cast to `VertxInternal` doesn’t fit well with the scope of this PR. Even with the cast, the call to `nettyEventLoopGroup` stays deprecated in vert.x 4 because `i.v.c.impl.VertxInternal` extends `i.v.c.Vertx`. And even if we expect the migration to vert.x 5 to be mostly about bumping the library version, we’ll still need to update the import anyway, since `VertxInternal` moved from `io.vertx.core.impl` to `io.vertx.core.internal`. Because of that, I think adding the cast will make more sense as part of the overall migration PR. wdyt?

macfarla:
> Hey @macfarla, I noticed you made a commit in #9396 that was related to this PR. I thought it would be cleaner to keep the deprecation-fix changes together here, but if you’d rather keep them in separate PRs just let me know and I can split them out without any issue.

That was unintentional @diega - sorry if that's made merging difficult! I'm making my way through codeQL warnings as a side project. The only changes in this PR I'm concerned about are the WebSocket and TLS changes. So yes if you want to split those out into a separate PR, I'd be happy to merge the rest, while we take a bit more time to verify the Websocket and TLS changes.

macfarla:
@diega I have rolled back the websocket and TLS changes and updated this PR. would be good to get the remaining changes in.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/handlers/JsonRpcParserHandler.java
```diff
@@ -78,14 +78,14 @@ public static Handler<Buffer> ipcHandler(
...
+      if (ctx.body().isEmpty()) {
         errorResponse(response, RpcErrorType.PARSE_ERROR);
       } else {
         try {
-          ctx.put(ContextKey.REQUEST_BODY_AS_JSON_OBJECT.name(), ctx.getBodyAsJson());
+          ctx.put(ContextKey.REQUEST_BODY_AS_JSON_OBJECT.name(), ctx.body().asJsonObject());
         } catch (DecodeException | ClassCastException jsonObjectDecodeException) {
           try {
-            final JsonArray batchRequest = ctx.getBodyAsJsonArray();
+            final JsonArray batchRequest = ctx.body().asJsonArray();
```
usmansaleem, line 88: this looks like repeated code from try block. @macfarla probably not part of this PR, but it needs revisiting.

PR #8975, opened 2025-07-18
Clean up production of the fastSyncFactory in DefaultSynchronizer

Matilda-Clerke:
## PR description
Clean up production of the fastSyncFactory in DefaultSynchronizer to explicitly produce nothing when FULL sync is used (previously relied on SnapDownloaderFactory to return an empty Optional)


pinges (approved):
LGTM

PR #8977, opened 2025-07-18
Upgrade netty

siladu:
Maybe some memory improvements from https://netty.io/news/2025/07/15/4-2-3.html 

> AdaptiveByteBufAllocator improvements
> The AdaptiveByteBufAllocator did see a lot of changes this release to not only fixing bugs but also decrease it's memory overhead dramatically. To ensure we don't regress in the future and also allow others to better understand the characteristics we also added an allocation pattern simulator. For some more details about the improvements please refer to [#15429](https://github.com/netty/netty/pull/15429)

https://github.com/netty/netty/pull/15424
> This greatly enhances chunk reuse, especially so the more in-use ByteBufs we have at any one time.
> In my local testing I see up to a 60% reduction in memory usage, depending on the use case.

https://github.com/netty/netty/pull/15399
> This should greatly reduce memory usage caused by chunks that hang around too long due to fragmentation.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ org.gradle.welcome=never
 # versionappendcommit=true
 
 # Optional, skip dependency verification for dev/debug builds
-# org.gradle.dependency.verification=lenient
+org.gradle.dependency.verification=lenient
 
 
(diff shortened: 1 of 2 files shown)


siladu:
Didn't make any noticeable difference

PR #8981, opened 2025-07-21
feat: add world view to traceStartBlock hook

Gabriel-Trintinalia:
This PR adds a `WorldView` parameter to the `traceStartBlock` hook methods in the `BlockAwareOperationTracer` interface, enabling tracers to access the world state at the beginning of block processing.

- Adds `WorldView` parameter to both `traceStartBlock` method overloads in the `BlockAwareOperationTracer` interface
- Updates all implementations and call sites to pass the world state to the tracer hooks
- Updates test mocks to accommodate the new method signature


matkt (approved):
LGTM

PR #8982, opened 2025-07-21
Use system call for beacon root

Gabriel-Trintinalia:


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/blockhash/CancunBlockHashProcessor.java
```diff
@@ -14,28 +14,46 @@
...
+
+import org.apache.tuweni.bytes.Bytes32;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Processes the beacon block storage if it is present in the block header. */
 public class CancunBlockHashProcessor extends FrontierBlockHashProcessor {
+  private static final Logger LOG = LoggerFactory.getLogger(CancunBlockHashProcessor.class);
+  Address BEACON_ROOTS_ADDRESS =
+      Address.fromHexString("0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02");
```
pinges: Is this class always used when there is a cancun fork? We might have to make this configurable, because e.g. an L2 might want to support cancun, but cannot create a contract at that address.

Gabriel-Trintinalia: If there is no contract or no code in the contract, the system call fails silently.

pinges: That is fine, but if they want to use a system contract for the beacon root, they might have to be able to set a different address.

Gabriel-Trintinalia: I can't see how this PR changed this, it isn't the same thing before?

PR #8984, opened 2025-07-21
Include authorization list in the transaction call object and use it when building transaction during simulation

mirgee:
Currently, the transaction call object doesn't include the authorization list, making it, for example, impossible to simulate type 4 transactions. This PR allows passing the authorization list in the transaction call object and uses it to build the transaction of the appropriate type during simulation.

#### TODO

- [x] Cover new behavior by a test.


ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
```diff
@@ -923,6 +926,47 @@ public void shouldSimulateLegacyTransactionWhenBaseFeeNotZero() {
...
 
+  @Test
+  public void shouldGuessDelegateCodeTransactionTypeWhenAuthorizationsPresent() {
+    final CodeDelegation delegation =
+        new CodeDelegation(BigInteger.ONE, Address.fromHexString("0x1"), 42L, FAKE_SIGNATURE);
+
+    final CallParameter callParameter =
+        legacyTransactionCallParameterBuilder()
+            .codeDelegationAuthorizations(List.of(delegation))
+            .build();
```
Gabriel-Trintinalia: Can we move this to a `codedelegatoinTransactionCallParameterBuilder`? we have a builder for all the other tx types

Gabriel-Trintinalia:
Thanks for the PR. For simulateV1, the `authorization` can be a union type of:
- `chainId`, `address`, `nonce`, `r`, `s`, `yParity` (implemented by this PR)
- `chainId`,`address`, `nonce`, `authority`.  (Can be done in a second PR)

In `eth_call`, `eth_simulate`, `eth_estimateGas`, etc. if `r`,`s`,`yParity` are present then we behave like this PR, and if `authority` is present but `r`,`s`,`yParity` are not then you would pretend like you validated the signature and recovered to the address provided in the `authority` field.

mirgee:
@Gabriel-Trintinalia If I understand correctly, you are pointing out that `org.hyperledger.besu.ethereum.core.CodeDelegation` used in `CallParameter` currently assumes the former case (except with `v` instead of `yParity`), doesn't allow for the latter, and so the functionality is supported only partially. In this case, I would agree with implementing support for the latter case in a separate PR.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/transaction/TransactionSimulatorTest.java
mirgee: Done.

Gabriel-Trintinalia:
> @Gabriel-Trintinalia If I understand correctly, you are pointing out that `org.hyperledger.besu.ethereum.core.CodeDelegation` used in `CallParameter` currently assumes the former case (except with `v` instead of `yParity`), doesn't allow for the latter, and so the functionality is supported only partially. In this case, I would agree with implementing support for the latter case in a separate PR.

What I meant is that `CodeDelegation` derives the `authority` from the signature fields. But in a simulation context, the user might not have a signature and may still want to simulate a call with a specific authority. To support this, we should allow injecting something like a `simulationAuthorizerSupplier` that resolves to the provided authority from the call parameters when applicable.  

It’s not a small change, so it makes sense to handle it in a separate issue.

Gabriel-Trintinalia (commented):
the methods `fromTransaction` in `CallParameter` need to be updated as well

mirgee:
> What I meant is that `CodeDelegation` derives the `authority` from the signature fields. But in a simulation context, the user might not have a signature and may still want to simulate a call with a specific authority. To support this, we should allow injecting something like a `simulationAuthorizerSupplier` that resolves to the provided authority from the call parameters when applicable.
> 
> It’s not a small change, so it makes sense to handle it in a separate issue.

Thank you for clarification, I think we are aligned now. Will do that in a separate PR.

> the methods fromTransaction in CallParameter need to be updated as well

Will do!

Gabriel-Trintinalia:
> > It’s not a small change, so it makes sense to handle it in a separate issue.

No problem. Once this PR is merged, I’ll create a follow-up issue to extend the functionality. I can assign it to you if you’d like to continue working on it. But the extension can be handled later. It’s not a high priority at the moment.

Gabriel-Trintinalia (approved):

mirgee:
@Gabriel-Trintinalia Created the follow-up issue #9043 , can you please assign it to me?

PR #8991, opened 2025-07-22
Extend P2P plugin API

pullurib:
### 🎯 **Overview**
Added comprehensive P2P plugin API to enable consensus plugins (like QBFT) to interact with the P2P network layer without direct access to core classes.

### 🚀 **New Plugin Services**

**P2PMessageService** - Message handling registration/unregistration, peer-to-peer message sending with capability/protocol support, access to peer connections and supported capabilities

**ProtocolManagerService** - Protocol manager registration for custom subprotocols, connection/disconnection lifecycle management, message processing delegation to registered protocols  

**ValidatorNetworkService** - Validator-specific messaging (broadcast to all validators, exclude lists), validator peer discovery and tracking, connection monitoring for consensus participants

### 📦 **New Data Types & Implementation**
Added plugin-safe data types: `PeerConnection` (peer connection interface), `Message`/`MessageData` (message abstractions), `Capability` (protocol capability), `Peer`/`PeerInfo` (peer identification), `DisconnectReason` (connection termination reasons). Implementation includes unified `P2PServiceImpl` with seamless P2P network integration, thread-safe concurrent handler management, and optional validator provider support.

### ✅ **Testing & Changes**
Comprehensive testing with `TestP2PServicePlugin` example and acceptance tests for message counting and connection tracking. Plugin API checksum updated for new interfaces with no breaking changes to existing plugin interfaces.

**Impact:** Enables consensus plugins to manage their own P2P protocols and validator networks while maintaining clean separation from Besu core internals.


Gabriel-Trintinalia:
Hi @pullurib, I like the concept of this PR. It’s useful and heading in a good direction. That said, I would suggest the following improvements:

Break the PR into smaller, focused pieces to make it easier to understand the purpose and scope of each service being introduced. For example, the changes related to the `P2PService` extensions and the new `P2PServiceMessage` are clear and could stand on their own.

The introduction of `ProtocolManagerService` also makes sense, but I believe it would be better to introduce that in a separate PR.

I’m also unsure whether `ValidatorNetworkService` should be part of Besu itself. It seems more appropriate for the plugin to manage this list rather than Besu directly.

Regarding the new classes in the plugin package, I agree that this is the right direction overall. However, I'm not fully convinced by the current interface design, particularly where methods perform actions, like sending in the peer connection interface. I would prefer to keep interfaces focused on data access and instead handle message sending through a method in the `P2PService` layer that resolves the correct peer and dispatches the message accordingly.

I'm happy to discuss the implementation with you and collaborate on this PR. I may also draft a follow-up PR with some key changes from this one to help simplify this PR

Gabriel-Trintinalia:
@pullurib draft a PR with the P2P service parts of this: https://github.com/hyperledger/besu/pull/9039. It is pretty much this PR without the message service, protocol service and validator service. Could you have a look?

pullurib:
Thanks for the suggestions and handling this @Gabriel-Trintinalia . I'll look into the next steps this week

PR #8992, opened 2025-07-23
feat: do not trace system calls

Gabriel-Trintinalia:
This pull request makes a minor update to the `SystemCallProcessor` class to modify how operation tracing is handled during message frame processing. Specifically, it replaces the use of a contextual operation tracer with a no-op tracer.

- Updated the `process` method to use `OperationTracer.NO_TRACING` instead of `context.getOperationTracer()` when processing the message frame stack. This ensures that no operation tracing is performed during the execution of system calls.


Matilda-Clerke (approved):
Seems reasonable. Was the tracing unused? Or too much of a processing burden?

PR #8993, opened 2025-07-23
Rename block hash processor

Gabriel-Trintinalia:
Renamed `BlockHashProcessor` to `PreExecutionProcessor` to better reflect its extended responsibilities. In addition to EIP-2935 (historical block hash), the processor also executes  EIP-4788 (beacon root exposure). The new name more accurately represents its broader role in pre-execution state updates via system calls.


Matilda-Clerke (approved):
Looks like a pretty straight forward rename to me

PR #8995, opened 2025-07-23
chore(pool): remove full transaction sender

Gabriel-Trintinalia:
This PR simplifies transaction broadcasting in Besu by removing support for sending full transactions to peers. Since https://github.com/hyperledger/besu/pull/8492, Besu only connects to peers that support `NEW_POOLED_TRANSACTION_HASHES` (minimum protocol: eth/66) and should only send transaction hashes.

Summary of Changes:

- Removed full transaction broadcasting logic.
- Deleted the TransactionsMessageSender class.
- Updated the transaction pool factory to reflect these changes.
- Removed related tests and mocks for peers without hash support.


Gabriel-Trintinalia:
@fab-10, good point, I wasn't aware of this. I will modify this PR to only get rid of the eth/65 part

Issue #8972, opened 2025-07-17
Investigate errors on Bloat network


ahamlat:
I see this error happening multiple times on the bloat network devnet (perf-devnet-2)

```
2025-07-17 13:10:40.997+00:00 | vert.x-worker-thread-10 | ERROR | BaseJsonRpcProcessor | Error processing method: eth_sendRawTransaction ["0x"]
java.lang.IndexOutOfBoundsException: index is out of bounds
	at org.apache.tuweni.bytes.Checks.checkElementIndex(Utils.java:19) ~[tuweni-bytes-2.7.0.jar:2.7.0]
	at org.apache.tuweni.bytes.ArrayWrappingBytes.get(ArrayWrappingBytes.java:50) ~[tuweni-bytes-2.7.0.jar:2.7.0]
	at org.hyperledger.besu.ethereum.core.encoding.TransactionDecoder.getTransactionType(TransactionDecoder.java:131) ~[besu-ethereum-core-25.6-develop-ebc8b77.jar:25.6-develop-ebc8b77]
	at org.hyperledger.besu.ethereum.core.encoding.TransactionDecoder.decodeOpaqueBytes(TransactionDecoder.java:111) ~[besu-ethereum-core-25.6-develop-ebc8b77.jar:25.6-develop-ebc8b77]
	at org.hyperledger.besu.ethereum.api.util.DomainObjectDecodeUtils.decodeRawTransaction(DomainObjectDecodeUtils.java:32) ~[besu-ethereum-api-25.6-develop-ebc8b77.jar:25.6-develop-ebc8b77]
	at org.hyperledger.besu.ethereum.api.jsonrpc.internal.methods.EthSendRawTransaction.response(EthSendRawTransaction.java:83) ~[besu-ethereum-api-25.6-develop-ebc8b77.jar:25.6-develop-ebc8b77]
	at org.hyperledger.besu.ethereum.api.jsonrpc.execution.BaseJsonRpcProcessor.process(BaseJsonRpcProcessor.java:44) ~[besu-ethereum-api-25.6-develop-ebc8b77.jar:25.6-develop-ebc8b77]
	at org.hyperledger.besu.ethereum.api.jsonrpc.execution.TracedJsonRpcProcessor.process(TracedJsonRpcProcessor.java:54) ~[besu-ethereum-api-25.6-develop-ebc8b77.jar:25.6-develop-ebc8b77]
[...]

PR #8997, opened 2025-07-23
TransactionDecoder: handle empty bytes

macfarla:
## PR description
gracefully handle the case where 0x empty is passed as a transaction.
DomainObjectDecodeUtils.decodeRawTransaction catches IllegalArgumentException so this change means you won't end up with the stack trace in the logs

## Fixed Issue(s)
refs #8972 


ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/TransactionDecoder.java
```diff
@@ -109,6 +109,9 @@ private static Transaction decodeTypedTransaction(
   public static Transaction decodeOpaqueBytes(
       final Bytes opaqueBytes, final EncodingContext context) {
     var transactionType = getTransactionType(opaqueBytes);
+    if (opaqueBytes.isEmpty()) {
```
Gabriel-Trintinalia, line 111: Do you think we should do the test before `getTransactionType`? 
Also, I would suggest changing the exception message to "Empty opaque bytes"

macfarla: done

PR #9009, opened 2025-07-28
Add metrics to peer task executor usage to better compare directly with old system

Matilda-Clerke:

### DIFF
--- a/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/BufferedGetPooledTransactionsFromPeerFetcher.java
+++ b/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/task/BufferedGetPooledTransactionsFromPeerFetcher.java
@@ -25,6 +25,8 @@ import org.hyperledger.besu.ethereum.eth.manager.peertask.PeerTaskExecutorResult
 import org.hyperledger.besu.ethereum.eth.transactions.PeerTransactionTracker;
 import org.hyperledger.besu.ethereum.eth.transactions.TransactionPool;
 import org.hyperledger.besu.ethereum.eth.transactions.TransactionPoolMetrics;
+import org.hyperledger.besu.metrics.BesuMetricCategory;
+import org.hyperledger.besu.plugin.services.metrics.OperationTimer;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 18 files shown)


Matilda-Clerke:
Screenshot from 2025-08-01 11-31-30
Here we can see that the performance is comparable to the old system

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/task/GetHeadersFromPeerTask.java
```diff
@@ -206,21 +208,25 @@ public PeerTaskValidationResponse validateResult(final List<BlockHeader> blockHe
...
-    for (BlockHeader blockHeader : result.result().get()) {
-      if (highestBlockHeader.get().getNumber() < blockHeader.getNumber()) {
-        highestBlockHeader.set(blockHeader);
-      }
+    if (!result.ethPeers().isEmpty() && result.result().isPresent()) {
+      result.result().get().stream()
+          .max(Comparator.comparingLong(ProcessableBlockHeader::getNumber))
+          .ifPresent(
+              (highestBlockHeader) ->
+                  result.ethPeers().getLast().chainState().update(highestBlockHeader));
```
pinges, line 233: this could have worked with the Oprional<>, or is there another reason for changing it to a list? The update here is not that important anymore for PoS. 
Also, we could do that same thing at the previous level, when we receive the blockHeaders.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/task/GetHeadersFromPeerTask.java
```diff
@@ -52,7 +54,17 @@ public GetHeadersFromPeerTask(
       final int skip,
       final Direction direction,
       final ProtocolSchedule protocolSchedule) {
-    this(blockNumber, maxHeaders, skip, direction, 5, protocolSchedule);
+    this(null, blockNumber, maxHeaders, skip, direction, 5, protocolSchedule);
```
pinges: minor: make retries a constant

pinges (approved):
Just a minor change, to add a constant for the default retries. Otherwise LGTM

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/manager/peertask/task/GetHeadersFromPeerTask.java
Matilda-Clerke: The idea with providing a list was to provide context to the caller about which peers were attempted. It's actually not required now, as we're explicitly choosing peers in the pivot block confirmer.

I can change it back to an `Optional`, but it doesn't really do any harm as a `List`

PR #9012, opened 2025-07-28
EIP-7928 - Block-level Access Lists - construction

mirgee:
### Summary
This PR introduces support for Block-level Access List construction in line with [EIP-7928](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7928.md). It adds a new serializable `BlockAccessList` data structure, along with a builder to record block-level account and storage accesses. Additionally, it introduces a `TransactionAccessList` used to track transaction-level accesses during execution.

During transaction processing, a `TransactionAccessList` is passed through the `MessageFrame` into the EVM, which records accessed accounts and storage keys. After processing, its contents are merged into the block-level `BlockAccessList`.

The `TransactionAccessList` only records touched accounts and slots, whereas the transaction-level updater is used to obtain the original and updated values. This approach allows correct read / write detection and handling of reverts.

The `BlockAccessList` is optionally included in the block body and its hash is stored in the block header under the `balHash` field. If present, the list is RLP-encoded.

BALs are constructed during:
* **Block processing** (in `AbstractBlockProcessor`) to validate `balHash` in the header and cache BALs for retrieval.
* **Block simulation** (in `BlockSimulator`) to include BALs in `eth_simulateV1` responses.
* **Block creation** (in `BlockCreator`) to embed BALs in newly built blocks.

### New feature flag
A new feature flag `--Xapi-block-access-list-enabled` enables:
* The `eth_getBlockAccessListByNumber` RPC endpoint.
* Caching BALs for the last `cache-last-blocks` imported blocks.
* Inclusion of BALs in block simulation outputs and call result objects.

### New RPC endpoint
`eth_getBlockAccessListByNumber` returns JSON-formatted block-level access list for a recently imported block from cache, identified by its block number or an alias like "latest", "finalized", "safe".

### New fork configuration
A new placeholder fork config "Amsterdam" (based on Osaka) has been introduced. When this fork is active:
* BALs are included in the block body and `balHash` in the block header of built blocks.
* During block import, if a BAL is present, its hash is compared against `balHash` in the header (mismatches are logged but do not invalidate blocks).
* BALs are always included in block simulation outputs regardless of the feature flag.

### Testing
The implementation has been testing by comparing `eth_simulateV1` outputs to those produced by [eth-bal-analysis](https://github.com/nerolation/eth-bal-analysis).

### TODO
- [x] Add transaction / block processing integration tests (possibly using [execution-spec-tests](https://github.com/ethereum/execution-spec-tests)), test BAL construction in `AbstractBlockProcessorIntegrationTest`.
- [x] Cleanups - pass the information whether BAL functionality was enabled via CLI or fork through `BlockAccessListManager` to avoid passing `isBlockAccessListEnabled` unnecessarily, ...
- [x] The output of `eth_getBlockAccessListByNumber` should include `balHash`
- [x] Make sure all tests pass.
- [ ] Add unit tests for newly added code (?)

Closes: #9080


app/src/test/java/org/hyperledger/besu/RunnerBuilderTest.java
```diff
@@ -229,7 +230,7 @@ public void movingAcrossProtocolSpecsUpdatesNodeRecord() {
               BlockDataGenerator.BlockOptions.create()
                   .setBlockNumber(1 + i)
                   .setParentHash(inMemoryBlockchain.getChainHeadHash()));
-      inMemoryBlockchain.appendBlock(block, gen.receipts(block));
+      inMemoryBlockchain.appendBlock(block, gen.receipts(block), Optional.empty());
```
matkt: maybe we can have .appendBlock(block, gen.receipts(block)) with empty by default

config/src/main/java/org/hyperledger/besu/config/JsonGenesisConfigOptions.java
```diff
@@ -528,6 +533,7 @@ public Map<String, Object> asMap() {
     getCancunEOFTime().ifPresent(l -> builder.put("cancunEOFTime", l));
     getPragueTime().ifPresent(l -> builder.put("pragueTime", l));
     getOsakaTime().ifPresent(l -> builder.put("osakaTime", l));
+    getAmsterdamTime().ifPresent(l -> builder.put("amsterdamTime", l));
```
matkt: I wondering if it will be better to add a first PR with amsterdam only ?

app/src/test/java/org/hyperledger/besu/RunnerBuilderTest.java
mirgee: I considered that and it would make sense if you want to minimize the diff or save yourself the editing effort (too late now :)) but in general I think it's better to avoid overloading if possible - listing the arguments explicitly seems clearer. In this case however it would be justifiable because this argument is almost certainly only temporary and will be removed along with the RPC endpoint.

So OK, will do :ok_hand:

config/src/main/java/org/hyperledger/besu/config/JsonGenesisConfigOptions.java
mirgee: It is a logically separable change, but what would be the purpose of separating it? Do you think it would make review significantly easier?

matkt: yes more in order to reduce the size of the PR, also because on my team they asked to use EXPERIMENTAL_EIPS . So maybe it make more sense to add the eip on this one instead of adding amsterdam

mirgee: What is the reasoning for adding it on `ExperimentalEips` as opposed to `Amsterdam`? Why `ExperimentalEips` and not `FutureEips`?

jflo: Good question and the answer is a bit of a judgement call. FutureEips is for things that are SFI'd (and maybe sometimes CFIs), and ExperimentalEips are for features that are changing faster since they are still being designed.

So it depends on how Glamsterdam headliner selection goes; I think it is likely that FutureEips is the right choice.

app/src/main/java/org/hyperledger/besu/cli/options/ApiConfigurationOptions.java
```diff
@@ -60,6 +60,13 @@ public ApiConfigurationOptions() {}
           "Set to enable gas price and minimum priority fee limit in eth_getGasPrice and eth_feeHistory (default: ${DEFAULT-VALUE})")
   private final Boolean apiGasAndPriorityFeeLimitingEnabled = false;
 
+  @CommandLine.Option(
```
jflo, line 63: Not the best place for this, and I don't think a CLI arg is necessary, because this must be activated at a specific block height, so it is best configured via the genesis file config - by the presence of a FutureEIPs or Amsterdam block height entry.

datatypes/src/main/java/org/hyperledger/besu/datatypes/CallParameter.java
```diff
@@ -158,4 +158,15 @@ public interface CallParameter {
...
+  /**
+   * Returns the list of code delegation authorizations.
+   *
+   * <p>These authorizations specify which contracts are allowed to execute code on behalf of the
+   * sender, enabling advanced execution scenarios such as meta-transactions or contract delegation
+   * patterns.
+   *
+   * @return a {@link List} of {@link CodeDelegation} entries
+   */
+  List<CodeDelegation> getCodeDelegationAuthorizations();
```
jflo: I don't understand why this was added, seems irrelevant to BAL?

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/BlockAccessListParameter.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.tuweni.units.bigints.UInt256;
+
+public class BlockAccessListParameter {
+
+  private final List<AccountChangesParameter> accountChanges;
```
jflo: the suffix 'Parameter' on all of these classes is confusing, since it is not passed into any RPC methods as an input. They should be removed.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/BlockOverridesParameter.java
```diff
@@ -50,15 +49,15 @@ public BlockOverridesParameter(
       @JsonProperty("time") final Optional<UnsignedLongParameter> timestamp,
       @JsonProperty("number") final Optional<UnsignedLongParameter> blockNumber,
       @JsonProperty("hash") final Optional<Hash> blockHash,
-      @JsonProperty("prevRandao") final Optional<Bytes32> prevRandao,
```
jflo: These seem less type safe and i'm not sure how they are relevant to BAL?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
```diff
@@ -152,6 +159,8 @@ public BlockTransactionSelector(
     blockWorldStateUpdater = worldState.updater();
     txWorldStateUpdater = blockWorldStateUpdater.updater();
     blockTxsSelectionMaxTime = miningConfiguration.getBlockTxsSelectionMaxTime();
+    currentTxnLocation = 0;
```
jflo: since this is an internal detail that is going to be updated async, using an AtomicInt or something naturally threadsafe is suggested.

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
```diff
@@ -135,7 +135,7 @@ public abstract class AbstractBlockTransactionSelectorTest {
   protected GenesisConfig genesisConfig;
   protected MutableBlockchain blockchain;
   protected TransactionPool transactionPool;
-  protected MutableWorldState worldState;
```
jflo: Not sure how this or any of the other changes in this class are relevant to BAL?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolScheduleModule.java
```diff
@@ -60,6 +61,7 @@ public ProtocolScheduleBuilder provideProtocolScheduleBuilder(
       final EvmConfiguration evmConfiguration,
       final BadBlockManager badBlockManager,
       final boolean isParallelTxProcessingEnabled,
+      final boolean isBlockAccessListEnabled,
```
jflo, line 64: I don't think you want to do this here, whether or not BAL is enabled would be a field on a ProtocolSpec, not a ProtocolSchedule. Each spec represents a hardfork.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolSpecModule.java
```diff
@@ -37,15 +37,17 @@ public ProtocolSpecModule() {}
...
   @Provides
   @Named("frontier")
   public ProtocolSpecBuilder frontierProtocolSpec(
       final EvmConfiguration evmConfiguration,
       final boolean isParalleltxEnabled,
+      final boolean isBlockAccessListEnabled,
       final MetricsSystem metricsSystem) {
     return MainnetProtocolSpecs.frontierDefinition(
-        evmConfiguration, isParalleltxEnabled, metricsSystem);
+        evmConfiguration, isParalleltxEnabled, isBlockAccessListEnabled, metricsSystem);
```
jflo, line 51: this should be false, and false for all hardforks up till FutureEIPs or perhaps Amsterdam

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
```diff
@@ -141,12 +164,18 @@ public static BlockBody readWrappedBodyFrom(
...
+    final Optional<List<Withdrawal>> withdrawals =
+        input.isEndOfCurrentList()
+            ? Optional.empty()
+            : Optional.of(input.readList(Withdrawal::readFrom));
+    final Optional<BlockAccessList> blockAccessList =
         input.isEndOfCurrentList()
             ? Optional.empty()
-            : Optional.of(input.readList(Withdrawal::readFrom)));
+            : Optional.of(BlockAccessListDecoder.decode(input));
+    return new BlockBody(txs, ommers, withdrawals, blockAccessList);
```
jflo, line 178: Just a warning, whether this is RLP or SSZ is still being debated :)

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/BlockAccessListEncoder.java
```diff
@@ -0,0 +1,81 @@
...
+          acctOut.writeList(
+              acct.storageReads(),
+              (sr, srOut) -> srOut.writeUInt256Scalar(sr.slot().getSlotKey().get()));
+
+          acctOut.writeList(
+              acct.balanceChanges(),
+              (bc, bcOut) -> {
+                bcOut.startList();
+                bcOut.writeInt(bc.txIndex());
+                bcOut.writeBytes(bc.postBalance());
```
jflo, line 56: anyone got any convenient way to limit this to uint128 instead of writing out everything in the Bytes returned by bc.postBalance()?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/difficulty/fixed/FixedDifficultyProtocolSchedule.java
```diff
@@ -38,6 +38,7 @@ public static ProtocolSchedule create(
       final MiningConfiguration miningConfiguration,
       final BadBlockManager badBlockManager,
       final boolean isParallelTxProcessingEnabled,
+      final boolean isBlockAccessListEnabled,
```
jflo, line 41: See prior comments about ProtocolSchedule vs. ProtocolSpecs

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -196,23 +201,48 @@ public BlockProcessingResult processBlock(
...
+      final WorldUpdater blockUpdater = worldState.updater();
       final Transaction transaction = transactions.get(i);
+      final WorldUpdater transactionUpdater = blockUpdater.updater();
       if (!hasAvailableBlockBudget(blockHeader, transaction, currentGasUsed)) {
         return new BlockProcessingResult(Optional.empty(), "provided gas insufficient");
       }
-      final WorldUpdater blockUpdater = worldState.updater();
 
+      final Optional<TransactionAccessList> transactionAccessList =
+          createAccessList(blockAccessListBuilder, i);
```
jflo, line 223: createAccessList is only used here, suggest inlining instead since it is only 1 line long.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -196,23 +201,48 @@ public BlockProcessingResult processBlock(
...
+              transactionUpdater,
               blobGasPrice,
               miningBeneficiary,
               transaction,
               i,
-              blockHashLookup);
+              blockHashLookup,
+              transactionAccessList);
+
+      if (transactionUpdater instanceof StackedUpdater<?, ?> stackedUpdater) {
```
jflo, line 236: WorldUpdaterMode.STACKED can be turned off via config, we should probably log a warning in the else condition to indicate that their configuration has broken BAL.

and probably make it non-configurable?

@matkt what say you?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -338,8 +369,33 @@ public BlockProcessingResult processBlock(
...
+                .getHeader()
+                .getBalHash()
+                .ifPresent(
+                    headerHash -> {
+                      final Hash expectedHash = BodyValidation.balHash(t);
+                      if (!headerHash.equals(expectedHash)) {
+                        LOG.warn(
+                            "{} (header BAL hash)\n!=\n{} (expected BAL hash)",
+                            headerHash,
+                            expectedHash);
```
jflo: This would be a bad block, it needs to also be handled as such. Suspect there is no balHash mismatch testing to have caught that.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/CodeDelegationProcessor.java
```diff
@@ -131,8 +136,10 @@ private void processCodeDelegation(
         return;
       }
       authority = worldUpdater.createAccount(authorizer.get());
+      eip7928AccessList.ifPresent(t -> t.addAccount(authority.getAddress()));
```
jflo, line 139: Not so sure about this one; just because an authority's account has been created, does not mean it will ever be accessed... @daniellehrner any thoughts on this part?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -239,33 +265,34 @@ public TransactionProcessingResult processTransaction(
...
-            maybeCodeDelegationProcessor.get().process(worldState, transaction);
-        warmAddressList.addAll(codeDelegationResult.accessedDelegatorAddresses());
+            maybeCodeDelegationProcessor.get().process(worldState, transaction, eip7928AccessList);
+        eip2930WarmAddressList.addAll(codeDelegationResult.accessedDelegatorAddresses());
         codeDelegationRefund =
             gasCalculator.calculateDelegateCodeGasRefund(
                 (codeDelegationResult.alreadyExistingDelegators()));
 
-        worldState.commit();
+        // worldState.commit();
```
jflo: I don't see any reason to remove this commit? it would still be needed so downstream users that fork this state will get the changes made by codeDelegationProcessor.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -298,7 +325,19 @@ public TransactionProcessingResult processTransaction(
...
               .blockHashLookup(blockHashLookup)
-              .accessListWarmStorage(storageList);
+              .eip2930AccessListWarmStorage(eip2930StorageList);
+
+      if (eip7928AccessList.isPresent()) {
+        commonMessageFrameBuilder.eip7928AccessList(eip7928AccessList.get());
+
+        for (Map.Entry<Address, Bytes32> entry : eip2930StorageList.entries()) {
+          final Address address = entry.getKey();
+          final Bytes32 slotKey = entry.getValue();
```
jflo: Spec does not indicate that warmed slots should be included in BAL, in fact, this would be a consensus breaking change if a user warms a slot and doesn't access it.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -421,7 +465,9 @@ public TransactionProcessingResult processTransaction(
       operationTracer.traceBeforeRewardTransaction(worldUpdater, transaction, coinbaseWeiDelta);
       if (!coinbaseWeiDelta.isZero() || !clearEmptyAccounts) {
         final var coinbase = worldState.getOrCreate(miningBeneficiary);
+        initialFrame.getEip7928AccessList().ifPresent(t -> t.addAccount(coinbase.getAddress()));
```
jflo, line 460: the BAL would also need the amount of the balance change, not just the address being credited.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessorIntegrationTest.java
```diff
@@ -835,6 +1060,24 @@ private Block createBlockWithTransactions(
     return new Block(blockHeader, blockBody);
   }
 
+  private void assertBlockAccessListAddresses(
```
jflo, line 1063: These assertions should depend on the protocol spec; we should only assert on them if the current hardfork has BAL enabled.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServerTest.java
```diff
@@ -882,6 +882,12 @@ void insertTestAccounts(final SnapTestAccount... accounts) {
     }
     storageTrie.commit(updater::putAccountStateTrieNode);
     updater.commit();
+    inMemoryStorage
+        .getWorldStateBlockNumber()
+        .ifPresent(
+            currentBlock ->
+                updateStorageArchiveBlock(
+                    inMemoryStorage.getComposedWorldStateStorage(), currentBlock + 1));
```
jflo: Curious why this is needed, was this not working correctly prior?

evm/src/main/java/org/hyperledger/besu/evm/account/CodeDelegationAccount.java
```diff
@@ -0,0 +1,122 @@
+/*
+ * Copyright contributors to Hyperledger Besu.
```
jflo: Not a bad idea, but this isn't being used anywhere

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -117,6 +117,8 @@ private void clearAccumulatedStateBesidesGasAndOutput(final MessageFrame frame)
     frame.clearLogs();
     frame.clearGasRefund();
 
+    // frame.getEip7928AccessList().ifPresent(Eip7928AccessList::clear);
```
jflo:
```suggestion
```

evm/src/main/java/org/hyperledger/besu/evm/processor/AbstractMessageProcessor.java
```diff
@@ -117,6 +117,8 @@ private void clearAccumulatedStateBesidesGasAndOutput(final MessageFrame frame)
...
+
```
jflo:
```suggestion
```

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockOverrides.java
```diff
@@ -67,7 +67,7 @@ public BlockOverrides(
       final Optional<Hash> stateRoot,
       final Optional<BigInteger> difficulty,
       final Optional<Bytes> extraData,
-      final Optional<Bytes32> mixHashOrPrevRandao) {
```
jflo: whats the reasoning behind relaxing the type safety here?

app/src/main/java/org/hyperledger/besu/cli/options/ApiConfigurationOptions.java
mirgee: The CLI flag serves a purpose very different from protocol-level activation - to enable cross-client validation of BAL construction on networks without protocol-level BAL support.

Since Geth already exposes a similar endpoint, the idea is to run Besu and Geth (and eventually other clients') nodes side-by-side on a network without protocol-level BAL support (e.g. mainnet), continuously query for BALs of recent blocks and compare hashes. The longer this test runs, the more confidence we gain that clients handle edge cases consistently.

This way, the BAL construction logic can be tested not only using activity generated on BAL testnet, but also using activity from other networks.

This endpoint is intended to be temporary and to be removed once BAL generation logic is stable and thoroughly validated across clients.

With regards to placement - it's important to realize that the flag mainly two things:

* adds a new RPC endpoint, and
* includes BALs in simulation results.

I was considering different approaches, like adding a new API namespace and adding optional boolean argument to `eth_simulateV1`, but given the temporary nature decided to go with the simplest option. If the question remains relevant, suggestions welcome :)

datatypes/src/main/java/org/hyperledger/besu/datatypes/CallParameter.java
mirgee: Indeed, this change is irrelevant to BALs themselves. However, it was necessary to fix several bugs in order to test BAL construction through the `eth_simulateV1` endpoint. I separated the fixes into #8983, #8984. Perhaps it would have been clearer to base the PRs on top of each other like `#8983 -> #8984 -> #9012`, but that introduces logistical challenges of its own; I can still do that if you insist, but I would suggest simply merging those PRs before this one to remove the changes unrelated to BALs from the diff, while still allowing for testing.

To explain how `BlockOverrides` and `CallParameter` are related to testing BALs, here is the testing method: The goal was to enable generating BALs for arbitrary blocks specified e.g. by number and compare against the reference implementation. I used `eth_getBlockByNumber` to obtain the transactions and header, built `eth_simulateV1` payload by building `BlockOverrides` from the header and using transactions for `CallParameter`s. The simulation result now contains the block's BAL as JSON, and scripts in [eth-bal-analysis](https://github.com/nerolation/eth-bal-analysis) + Alchemy API allow obtaining the reference BAL as JSON. The diff between those, dumped to a file, should be empty.

[Here](https://github.com/mirgee/eth-bal-analysis/blob/cmp/compare_bal.py) is the testing script.

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/BlockAccessListParameter.java
mirgee: Makes sense :+1: I will move this class to `org.hyperledger.besu.ethereum.api.jsonrpc.internal.results` and rename

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/parameters/BlockOverridesParameter.java
mirgee: This type is only used for deserialization and it is converted to `Bytes32` during construction of `BlockOverrides`.

But I agree it's not ideal. The problem with `Bytes32` is that it is an abstract type and `Jackson` has no clue how to instantiate it. A proper solution would be to implement custom deserializer for `Bytes32` - I will do that in #8983, and we can merge it before this PR?

ethereum/blockcreation/src/main/java/org/hyperledger/besu/ethereum/blockcreation/txselection/BlockTransactionSelector.java
mirgee: Good point, will do! The counter is currently read and updated in strictly single-threaded context, but I guess it's still more prudent to use `AtomicInt` in case parallelism is introduced in the future.

ethereum/blockcreation/src/test/java/org/hyperledger/besu/ethereum/blockcreation/AbstractBlockTransactionSelectorTest.java
mirgee: Good catch - debugging leftover. Will undo!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockBody.java
mirgee: I promise I will not cry if this gets removed ;) And if we do decide to go with RLP, at least we have something already.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
mirgee: This was done as a workaround for
```
error: local variables referenced from a lambda expression must be final or effectively final blockAccessListBuilder.map(b -> new TransactionAccessList(i));
                                                                                                                                                        ^ 
```
because it seemed like the least bad option, but no strong opinion here - open to suggestions.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
mirgee: The idea was to start by collecting data and fixing implementation misalignment without impacting network health, but we can certainly be more aggressive. Maybe seeing all the things that can go wrong made me overly cautious :)

Block validity testing accounting for `balHash` mismatches will be added once the spec-test fixtures are ready. AFAIK, this is the relevant work in progress:
https://github.com/ethereum/execution-spec-tests/pull/1823
https://github.com/ethereum/execution-spec-tests/pull/1866

In the meantime, we have at least some BAL generation coverage in `org.hyperledger.besu.ethereum.mainnet.AbstractBlockProcessorIntegrationTest`.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
mirgee: Yes, good catch. This is something I added not because of spec but to align with `eth-bal-analysis` and planned to raise this to @nerolation - so this is as good opportunity as any.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
mirgee: We use the updater to obtain the balance changes, see `addTransactionLevelAccessList` in `BlockAccessListBuilder`.

ethereum/core/src/test/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessorIntegrationTest.java
mirgee: Now the test essentially simulates the case where the `--Xapi-block-access-list-enabled` is passed, which is a valid way to enable BAL construction during block processing, since to serve the BALs on `eth_getBlockAccessListByNumber`, they must be constructed and cached. But I agree we should also cover the fork-driven activation by a test.

evm/src/main/java/org/hyperledger/besu/evm/account/CodeDelegationAccount.java
mirgee: Hmm, don't remember adding this. Maybe accidental rebase leftover or something, will remove.

plugin-api/src/main/java/org/hyperledger/besu/plugin/data/BlockOverrides.java
mirgee: Discussed [here](https://github.com/hyperledger/besu/pull/9012#discussion_r2252428078).

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolScheduleModule.java
matkt: not sure to have this one, ProtocolSchedule is building ProtocolSpec, so we should pass this field to this class . no ?

app/src/main/java/org/hyperledger/besu/cli/options/ApiConfigurationOptions.java
matkt: Yes we want to have two way to activate it , by fork but also with a flag . like that we can test it without a fork on the current mainnet

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
mirgee: The problem is that committing a `StackedUpdater` modifies its state, namely  `updatedAccounts` and `deletedAccounts`, which are used to update the BAL with transaction-level changes. In other words, BAL must be updated before `StackedUpdater` gets committed, so including this commit currently leads to invalid BALs due to missing changes. We have been testing without it, but if this commit must be included, we need to implement some solution, e.g. make the transaction processor responsible for BAL updates.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolScheduleModule.java
mirgee: I think there is some confusion due to poor naming - `isBlockAccessListEnabled`, here and everywhere else, signals that the `--Xapi-block-access-list-enabled` CLI flag is active. Maybe a better name would be `isBlockAccessListApiEnabled` or `isBlockAccessListRpcEnabled` or `isBlockAccessListCliEnabled` or something to that effect. Or use "debug" somewhere in the name.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
mirgee: My worry is that logging the warning in the else branch here, will be too verbose - perhaps we should log it only once on startup, if journaled updater is enabled?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolSpecModule.java
mirgee: I think this is the same conversation as [this](https://github.com/hyperledger/besu/pull/9012#discussion_r2252416464) and [this](https://github.com/hyperledger/besu/pull/9012#discussion_r2246399134) - let's not duplicate threads.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/difficulty/fixed/FixedDifficultyProtocolSchedule.java
mirgee: I think this is the same conversation as https://github.com/hyperledger/besu/pull/9012#discussion_r2252416464 and https://github.com/hyperledger/besu/pull/9012#discussion_r2246399134 - let's not duplicate threads.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServerTest.java
mirgee: Without this fix, the tests in `SnapServerTest` are failing with `FlatDbMode.ARCHIVE` even on the main branch. Bonsai archive relies on world block number to locate `FlatDb` entries.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
mirgee: When updating the BAL with transaction-level changes, we read the account from the updater and obtain both the original and updated balance from there (see `addTransactionLevelAccessList` in `BlockAccessListBuilder`). In this case we know the account will be in the updater because of this line:
```java
final var coinbase = worldState.getOrCreate(miningBeneficiary);
```

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
nerolation: So depending on the final design (with storage locations for reads vs without), warmed slots that are not in the BAL can only come from 2930 ALs. We could clarify in the EIP that 2930 ALs entries are only duplicated in the BAL if they end up being touched/changed. 
So, I'd not directly put them into the BAL but only when the AL entries were actually touched.

nerolation: In the specs we don't add the entries already:
https://github.com/ethereum/execution-specs/blob/9719aefa9afbaeca10332f78fb55f1d6f20b1eee/src/ethereum/osaka/fork.py#L927

mirgee: It seems that geth with prestate tracer and diffmode off includes warmed but untouched slots in the block trace, which `eth-bal-analysis` evaluates as reads. However, this is becoming less of a problem the closer we are to having reference test fixtures with BALs ready to rely on for testing.

nerolation: Interesting catch. I will double check that and fix it in eth-bal-analysis!

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
mirgee: ~~Can someone please explain to me why this commit needs to be here? In which situations is it necessary and why? Understanding this might be useful for finding the optimal solution. Moreover, it would be useful to have a test which fails without the commit and passes with it.~~
EDIT: Fixed by using a dedicated delegation-specific updater.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
matkt: I think we should find a way to work also with JournaledUpdater

mirgee: > I think we should find a way to work also with JournaledUpdater

Would that be a blocker for this PR? BAL construction is still an experimental feature, so perhaps it's admissible for it to not work with every possible configuration yet?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -238,34 +262,37 @@ public TransactionProcessingResult processTransaction(
...
+            maybeCodeDelegationProcessor
+                .get()
+                .process(delegationUpdater, transaction, eip7928AccessList);
+        eip2930WarmAddressList.addAll(codeDelegationResult.accessedDelegatorAddresses());
         codeDelegationRefund =
             gasCalculator.calculateDelegateCodeGasRefund(
                 (codeDelegationResult.alreadyExistingDelegators()));
-
-        worldState.commit();
+        delegationUpdater.commit();
```
matkt, line 274: Hi @daniellehrner,
Could you please take a look at this modification?

We are not calling commit on the worldState Updater passed into the processTransaction method, but instead we're creating a specific updater for the code delegation.

I recall you mentioned that we should call commit regardless, even if the transaction fails. I believe this approach should still be fine, but I'd prefer to have your feedback to avoid introducing any bugs.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServerTest.java
jflo: if we pull that fix out to its own pr we can get it in main now while BAL is being worked on.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
jflo: I'm ok with supporting JournaledUpdater in a subsequent pr.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
jflo: I see, which would be necessary in cases where the CLI option is used to enable it on mainnet. Ok.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
jflo: If would think createAccessList is going to have future utility, leave it in.

If you don't think createAccessList will get more sophisticated or be used more widely, suggest just declaring a final var and copying i into it, and inlining.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
mirgee: Are we 100% sure we must commit to supporting compatibility of `JournaledUpdater` with BALs? Supporting another updater will significantly expand the set of execution flows involved in BAL construction, increasing the surface area for bugs and maintenance complexity.  It seems reasonable to rather standardize on a single updater (at least for BAL construction), although there may be constraints I’m not aware of that require supporting both?

If indeed it turns out we must maintain support both updaters, I will create an issue and start working on it.

ethereum/eth/src/test/java/org/hyperledger/besu/ethereum/eth/manager/snap/SnapServerTest.java
mirgee: OK, created #9057 for this purpose.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
mirgee: @matkt @jflo Can you please help me understand what is the rationale for supporting two different updaters? My understanding is that `JournaledUpdater` is more memory-efficient - what are the benefits of `StackedUpdater`? In what scenarios is each preferred? I am trying to understand the benefits of supporting both.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolScheduleModule.java
matkt: this flag will also activate BAL during block processing on mainnet ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/encoding/BlockAccessListEncoder.java
matkt: Could you explain more, not sure to got it

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/CodeDelegationProcessor.java
matkt: in this case it's more because we will save it. BAL is the read but also the write

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/components/ProtocolScheduleModule.java
mirgee: @matkt What do you mean by "activate BAL during block processing"? Impact of activating the feature flag on block processing is only that the BALs will be built and cached for the purposes of serving them via the RPC endpoint.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
mirgee: According to @shemnon explanation, `JournaledUpdater` is essentially a performance improvement on `StackedUpdater`, but there are still some rare edge cases where it behaves incorrectly - that's why we keep supporting both for the moment:

> Thee journaled updator fails some pre-shanghai-attack empty account tests, which is why it wasn't a full replacement for stacked updater.
IMHO the default updater should be journaled updater and stacked updater should only ever be used if there are empty accounts in the  data.
The failed IIRC related, and the "empty account touch rule" was involved.
I forgot what the failing tess were, but when a frame reverts we also need to reset the list of "touched empty accounts" to what it was before the revert.  Imangine account E is an empty account, A calls B, B touches E, but then B reverts. A returns.  The account E should not be removed from the merkle trie beause the side effects of B were all removed by the revert.

matkt:
Did several tests , I think we are good for this one

matkt:
@ahamlat do you want do to a performance check before merging ?

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpecBuilder.java
```diff
@@ -344,6 +358,16 @@ public ProtocolSpec build(final ProtocolSchedule protocolSchedule) {
...
         blockValidatorBuilder.apply(blockHeaderValidator, blockBodyValidator, blockProcessor);
     final BlockImporter blockImporter = blockImporterBuilder.apply(blockValidator);
+
+    BlockAccessListFactory finalBalManager = blockAccessListFactory;
+    if (finalBalManager == null && isBlockAccessListEnabled) {
+      finalBalManager = new BlockAccessListFactory(true, false);
+    } else if (finalBalManager != null
+        && isBlockAccessListEnabled
+        && !finalBalManager.isCliActivated()) {
+      finalBalManager = new BlockAccessListFactory(true, finalBalManager.isForkActivated());
```
ahamlat: Not sure to understand this code. Why another instance is created on top of the existing one. Also `isBlockAccessListEnabled` should be the same as `finalBalManager.isCliActivated()`.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
```diff
@@ -350,30 +382,57 @@ public BlockProcessingResult processBlock(
       return new BlockProcessingResult(Optional.empty(), e);
     }
 
+    Optional<BlockAccessList> blockAccessList =
```
ahamlat: To be in phase with the code above and do one check when the feature is not enabled, I suggest using `blockAccessListBuilder.ifPresent`. then create blockAccessList and validate balHash.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpecBuilder.java
mirgee: From the perspective of `ProtocolSpecBuilder`, there is no guarantee that the value of `cliActivated` of the  `BlockAccessListFactory` passed to it via the setter respects `isBlockAccessListEnabled`, so this branch is there to make sure it does. It's not strictly necessary, it's more of a defensive reconciliation which can be removed if you think it's not beneficial.

A new instance is created because I think that `BlockAccessListFactory` should remain immutable.

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/AbstractBlockProcessor.java
mirgee: Simply using `ifPresent` leads to issues related to mutating a variable in the enclosing scope from a lambda. How about this? It avoids using lambdas, performs only one check, but seems to be less clear:
```java
    final Optional<BlockAccessList> maybeBlockAccessList;
    if (blockAccessListBuilder.isPresent()) {
      final BlockAccessList bal = blockAccessListBuilder.get().build();
      maybeBlockAccessList = Optional.of(bal);
      try {
        final Optional<Hash> headerBalHash = block.getHeader().getBalHash();
        if (headerBalHash.isPresent()) {
          final Hash expectedHash = BodyValidation.balHash(bal);
          if (!headerBalHash.get().equals(expectedHash)) {
            LOG.warn(
                "{} (header BAL hash)\n!=\n{} (expected BAL hash)",
                headerBalHash.get(), expectedHash);
          }
        }
      } catch (Exception e) {
        LOG.error("Error validating BAL hash", e);
      }
    } else {
      maybeBlockAccessList = Optional.empty();
    }
```

mirgee: Alternatively, we can do a `map` instead of `ifPresent`:
```java
    final Optional<BlockAccessList> maybeBlockAccessList =
      blockAccessListBuilder.map(builder -> {
        final BlockAccessList bal = builder.build();
        try {
          final Optional<Hash> headerBalHash = block.getHeader().getBalHash();
          if (headerBalHash.isPresent()) {
            final Hash expected = BodyValidation.balHash(bal);
            if (!headerBalHash.get().equals(expected)) {
              LOG.warn(
                  "{} (header BAL hash)\n!=\n{} (expected BAL hash)",
                  headerBalHash.get(), expected);
            }
          }
        } catch (Exception e) {
          LOG.error("Error validating BAL hash", e);
        }
        return bal;
      });
```

ahamlat: The first proposal is pretty clear for me and addresses my comment 👍

ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/ProtocolSpecBuilder.java
ahamlat: Maybe add a comment in this case to make it more clear. It is not ideal but it feels hard to head without a clear comment.

mirgee: Yes, I agree a comment should have been added :+1:

matkt:
Did hoodi fullsync with journaled and stacked mode, also did a mainnet snapsync with this branch and node was running without any issues

PR #9025, opened 2025-07-30
Remove CLI options for JMH and adapt benchmarks

lu-pinto:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Improve the sync performance by not RLP decoding bodies during sync. This means we are using less memory and CPU, allowing us to increase the parallelism of the download pipeline, which has been increased from 4 to 8. Can be reduced again with  `--Xsynchronizer-downloader-parallelism=4` [#8959](https://github.com/hyperledger/besu/pull/8959)
 - Enable decoding for large RPC requests [#8877](https://github.com/hyperledger/besu/pull/8877)
 - Add `--attempt-cache-bust` to evmtool benchmark subcommand [#8985](https://github.com/hyperledger/besu/pull/8985)
+- Add gas usage metric to eth_call [#9019](https://github.com/hyperledger/besu/pull/9019)
 
 #### Fusaka devnets
 - EIP-7910 - `eth_config` JSON-RPC Method [#8417](https://github.com/hyperledger/besu/pull/8417), [#8946](https://github.com/hyperledger/besu/pull/8946), [#9015](https://github.com/hyperledger/besu/pull/9015)
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthCall.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthCall.java
@@ -41,6 +41,10 @@ import org.hyperledger.besu.ethereum.transaction.TransactionInvalidReason;
 import org.hyperledger.besu.ethereum.transaction.TransactionSimulator;
 import org.hyperledger.besu.ethereum.transaction.TransactionSimulatorResult;
 import org.hyperledger.besu.evm.tracing.OperationTracer;
+import org.hyperledger.besu.metrics.BesuMetricCategory;
+import org.hyperledger.besu.plugin.services.MetricsSystem;
+import org.hyperledger.besu.plugin.services.metrics.Counter;
+import org.hyperledger.besu.plugin.services.metrics.LabelledMetric;
 
 import java.util.Optional;
 
@@ -48,11 +52,20 @@ import com.google.common.annotations.VisibleForTesting;
 
 public class EthCall extends AbstractBlockParameterOrBlockHashMethod {
   private final TransactionSimulator transactionSimulator;
+  private final LabelledMetric<Counter> gasUsedCounter;
 
   public EthCall(
-      final BlockchainQueries blockchainQueries, final TransactionSimulator transactionSimulator) {
+      final BlockchainQueries blockchainQueries,
+      final TransactionSimulator transactionSimulator,
+      final MetricsSystem metricsSystem) {
     super(blockchainQueries);
     this.transactionSimulator = transactionSimulator;
+    this.gasUsedCounter =
+        metricsSystem.createLabelledCounter(
+            BesuMetricCategory.RPC,
+            "eth_call_gas_used_total",
+            "Total gas used by eth_call requests",
+            "status");
   }
 
   @Override
@@ -94,22 +107,25 @@ public class EthCall extends AbstractBlockParameterOrBlockHashMethod {
             maybeStateOverrides,
             buildTransactionValidationParams(header, callParams),
             OperationTracer.NO_TRACING,
-            (mutableWorldState, transactionSimulatorResult) -> {
-              return transactionSimulatorResult.map(
-                  result ->
-                      result
-                          .getValidationResult()
-                          .either(
-                              (() ->
-                                  result.isSuccessful()
-                                      ? new JsonRpcSuccessResponse(
-                                          request.getRequest().getId(),
-                                          result.getOutput().toString())
-                                      : errorResponse(request, result)),
-                              reason -> errorResponse(request, result)));
-            },
+            (mutableWorldState, transactionSimulatorResult) ->
+                transactionSimulatorResult.map(
+                    result -> {
+                      long gasUsed = result.getGasEstimate();
+                      if (result.isSuccessful()) {
+                        gasUsedCounter.labels("success").inc(gasUsed);
+                        return new JsonRpcSuccessResponse(
+                            request.getRequest().getId(), result.getOutput().toString());
+                      } else {
+                        gasUsedCounter.labels("error").inc(gasUsed);
+                        return errorResponse(request, result);
+                      }
+                    }),
             header)
-        .orElse(errorResponse(request, INTERNAL_ERROR));
+        .orElseGet(
+            () -> {
+              gasUsedCounter.labels("internal_error").inc(0);
+              return errorResponse(request, INTERNAL_ERROR);
+            });
   }
 
   @VisibleForTesting
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/EthJsonRpcMethods.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/EthJsonRpcMethods.java
@@ -73,6 +73,7 @@ import org.hyperledger.besu.ethereum.eth.transactions.TransactionPool;
 import org.hyperledger.besu.ethereum.mainnet.ProtocolSchedule;
 import org.hyperledger.besu.ethereum.p2p.rlpx.wire.Capability;
 import org.hyperledger.besu.ethereum.transaction.TransactionSimulator;
+import org.hyperledger.besu.plugin.services.MetricsSystem;
 
 import java.util.Map;
 import java.util.Set;
@@ -93,6 +94,7 @@ public class EthJsonRpcMethods extends ApiGroupJsonRpcMethods {
   private final ApiConfiguration apiConfiguration;
   private final GenesisConfigOptions genesisConfigOptions;
   private final TransactionSimulator transactionSimulator;
+  private final MetricsSystem metricsSystem;
 
   public EthJsonRpcMethods(
       final BlockchainQueries blockchainQueries,
@@ -105,7 +107,8 @@ public class EthJsonRpcMethods extends ApiGroupJsonRpcMethods {
       final Set<Capability> supportedCapabilities,
       final ApiConfiguration apiConfiguration,
       final GenesisConfigOptions genesisConfigOptions,
-      final TransactionSimulator transactionSimulator) {
+      final TransactionSimulator transactionSimulator,
+      final MetricsSystem metricsSystem) {
     this.blockchainQueries = blockchainQueries;
     this.synchronizer = synchronizer;
     this.protocolSchedule = protocolSchedule;
@@ -117,6 +120,7 @@ public class EthJsonRpcMethods extends ApiGroupJsonRpcMethods {
     this.apiConfiguration = apiConfiguration;
     this.genesisConfigOptions = genesisConfigOptions;
     this.transactionSimulator = transactionSimulator;
+    this.metricsSystem = metricsSystem;
   }
 
   @Override
@@ -135,7 +139,7 @@ public class EthJsonRpcMethods extends ApiGroupJsonRpcMethods {
         new EthGetBlockReceipts(blockchainQueries, protocolSchedule),
         new EthGetBlockTransactionCountByNumber(blockchainQueries),
         new EthGetBlockTransactionCountByHash(blockchainQueries),
-        new EthCall(blockchainQueries, transactionSimulator),
+        new EthCall(blockchainQueries, transactionSimulator, metricsSystem),
         new EthFeeHistory(protocolSchedule, blockchainQueries, miningCoordinator, apiConfiguration),
         new EthGetCode(blockchainQueries),
         new EthGetLogs(blockchainQueries, apiConfiguration.getMaxLogsRange()),
--- a/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/JsonRpcMethodsFactory.java
+++ b/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/methods/JsonRpcMethodsFactory.java
@@ -136,7 +136,8 @@ public class JsonRpcMethodsFactory {
                   supportedCapabilities,
                   apiConfiguration,
                   genesisConfigOptions,
-                  transactionSimulator),
+                  transactionSimulator,
+                  metricsSystem),
               new NetJsonRpcMethods(
                   p2pNetwork,
                   networkId,
--- a/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthCallTest.java
+++ b/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthCallTest.java
@@ -54,6 +54,10 @@ import org.hyperledger.besu.ethereum.transaction.ImmutableCallParameter;
 import org.hyperledger.besu.ethereum.transaction.PreCloseStateHandler;
 import org.hyperledger.besu.ethereum.transaction.TransactionSimulator;
 import org.hyperledger.besu.ethereum.transaction.TransactionSimulatorResult;
+import org.hyperledger.besu.metrics.BesuMetricCategory;
+import org.hyperledger.besu.plugin.services.MetricsSystem;
+import org.hyperledger.besu.plugin.services.metrics.Counter;
+import org.hyperledger.besu.plugin.services.metrics.LabelledMetric;
 
 import java.util.Map;
 import java.util.Optional;
@@ -81,13 +85,23 @@ public class EthCallTest {
   @Mock private BlockchainQueries blockchainQueries;
   @Mock private TransactionSimulator transactionSimulator;
   @Mock private BlockHeader blockHeader;
+  @Mock private MetricsSystem metricsSystem;
 
   @Captor ArgumentCaptor<PreCloseStateHandler<Optional<JsonRpcResponse>>> mapperCaptor;
   @Captor ArgumentCaptor<CallParameter> callParameterCaptor;
 
   @BeforeEach
   public void setUp() {
-    method = new EthCall(blockchainQueries, transactionSimulator);
+    @SuppressWarnings("unchecked")
+    LabelledMetric<Counter> mockGasUsedCounter = mock(LabelledMetric.class);
+    Counter mockCounter = mock(Counter.class);
+    when(mockGasUsedCounter.labels(any(String[].class))).thenReturn(mockCounter);
+
+    when(metricsSystem.createLabelledCounter(
+            any(BesuMetricCategory.class), any(String.class), any(String.class), any(String.class)))
+        .thenReturn(mockGasUsedCounter);
+
+    method = new EthCall(blockchainQueries, transactionSimulator, metricsSystem);
     blockHeader = mock(BlockHeader.class);
     when(blockHeader.getBlockHash()).thenReturn(Hash.ZERO);
   }


ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
```diff
@@ -78,11 +88,21 @@ public void setUp() {
...
+    for (int i = 0; i < OPERATIONS_PER_INVOCATION; i++) {
       frame.pushStackItem(bytes);
       CountLeadingZerosOperation.staticOperation(frame);
       frame.popStackItem();
     }
   }
+
+  @Benchmark
+  @OperationsPerInvocation(OPERATIONS_PER_INVOCATION)
+  public void baseline() {
```
siladu, line 107:
```suggestion
  public void onePushOnePopBaseline() {
```

I prefer this being more descriptive to help distinguish between opcodes that might require two pushes for example

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
```diff
@@ -78,11 +88,21 @@ public void setUp() {
   }
 
   @Benchmark
+  @OperationsPerInvocation(OPERATIONS_PER_INVOCATION)
```
siladu, line 96: very cool!

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
```diff
@@ -30,15 +30,25 @@
...
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
 
 @State(Scope.Thread)
-@OutputTimeUnit(value = TimeUnit.MILLISECONDS)
-public class OneMillionCountLeadingZerosOperationBenchmark {
+@OutputTimeUnit(value = TimeUnit.NANOSECONDS)
+@Warmup(iterations = 1)
+@Fork(2)
```
siladu: What's the thinking for 2 forks?

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
```diff
@@ -78,11 +88,21 @@ public void setUp() {
   }
 
   @Benchmark
+  @OperationsPerInvocation(OPERATIONS_PER_INVOCATION)
   public void executeOperation() {
-    for (int i = 0; i < 1000000; i++) {
+    for (int i = 0; i < OPERATIONS_PER_INVOCATION; i++) {
```
siladu, line 98: I'm still not sure we want an inner loop...
https://github.com/openjdk/jmh/blob/2a316030b509aa9874dd6ab04e21962ac92cd634/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_11_Loops.java#L47-L49

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
```diff
@@ -30,15 +30,25 @@
...
 import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
 
 @State(Scope.Thread)
-@OutputTimeUnit(value = TimeUnit.MILLISECONDS)
-public class OneMillionCountLeadingZerosOperationBenchmark {
+@OutputTimeUnit(value = TimeUnit.NANOSECONDS)
+@Warmup(iterations = 1)
```
siladu:
```suggestion
@Warmup(iterations = 1, time = 1, timeUnit = TimeUnit.SECONDS)
```

I think we could also add the Measurement explicitly and also reduce default time of 10 to 1.
```
@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
```

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/Keccak256Benchmark.java
```diff
@@ -21,15 +21,24 @@
...
 
 @State(Scope.Thread)
-@OutputTimeUnit(value = TimeUnit.MILLISECONDS)
-public class Hashing20kKeccak256Benchmark {
+@Warmup(iterations = 1)
+@OutputTimeUnit(value = TimeUnit.NANOSECONDS)
+@Fork(2)
+@BenchmarkMode(Mode.AverageTime)
+public class Keccak256Benchmark {
+  private static final int OPERATIONS_PER_INVOCATION = 20_000;
```
siladu: Why is this different to CLZ?

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
lu-pinto: I don't follow as the name includes the class and is already unambiguous enough:
```
Result "org.hyperledger.besu.ethereum.vm.operations.CountLeadingZerosOperationBenchmark.baseline":
  2.798 ±(99.9%) 3.870 ns/op [Average]
  (min, avg, max) = (2.321, 2.798, 4.595), stdev = 1.005
  CI (99.9%): [≈ 0, 6.669] (assumes normal distribution)
```

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
lu-pinto: Can set it to 1 if you feel strongly.
Forks exist for helping benchmarks give consistent results and that there's less chance of something going unnoticed that impacts the results, .e.g. misbehaving JVM, sudden CPU spike, ...

For the average benchmark mode it is comparatively with just increasing the iterations because the average is taken across everything instead of the lower fork average:
```
# Fork: 1 of 2
# Warmup Iteration   1: Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
2.012 ns/op
Iteration   1: 2.005 ns/op
Iteration   2: 2.003 ns/opING [38s]
Iteration   3: 2.002 ns/opING [48s]
Iteration   4: 2.003 ns/opING [58s]
Iteration   5: 2.024 ns/opING [1m 8s]

# Run progress: 25.00% complete, ETA 00:03:02
# Fork: 2 of 2
# Warmup Iteration   1: Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
2.044 ns/op
Iteration   1: 2.031 ns/op
Iteration   2: 2.033 ns/opING [1m 39s]
Iteration   3: 2.036 ns/opING [1m 49s]
Iteration   4: 2.031 ns/opING [1m 59s]
Iteration   5: 2.035 ns/opING [2m 9s]


Result "org.hyperledger.besu.ethereum.vm.operations.CountLeadingZerosOperationBenchmark.baseline":
  2.020 ±(99.9%) 0.023 ns/op [Average]
  (min, avg, max) = (2.002, 2.020, 2.036), stdev = 0.015
  CI (99.9%): [1.997, 2.043] (assumes normal distribution)
```

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
siladu: I don't mean to determine which class but rather describe what the baseline is...

e.g. if you have 
AddOperationBenchmark.baseline()
and
CLZOperationBenchmark.baseline()

you might assume the baseline is the same, but Add has two pushes so it is not comparable to the CLZ baseline.

I think it is a small change to provide some useful signal when you are inspecting a group of results (maybe less useful is you are focussing on analysing a single opcode).

So we could end up something like:

AddOperationBenchmark.twoPushOnePopBaseline()
AndOperationBenchmark.twoPushOnePopBaseline()
CLZOperationBenchmark.onePushOnePopBaseline()
BlobHashOperation.onePushOnePopBaseline()
BaseFeeOperation.onePopBaseline()

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
siladu: I think the default is 5 so wondering what the thinking was for lowering it.

siladu:
Results for this PR running `./gradlew :ethereum:core:jmh -Pincludes=CountLeadingZeros` takes **21 mins** overall...
```
Benchmark                                                                                                     (bytesHex)  Mode  Cnt   Score   Error  Units
CountLeadingZerosOperationBenchmark.baseline                                                                        0x23  avgt   10   4.166 ± 0.011  ns/op
CountLeadingZerosOperationBenchmark.baseline          0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10   4.168 ± 0.015  ns/op
CountLeadingZerosOperationBenchmark.baseline            0x23232323232323232323232323232323232323232323232323232323232323  avgt   10   4.165 ± 0.020  ns/op
CountLeadingZerosOperationBenchmark.baseline                            0x2323232323232323232323232323232323232323232323  avgt   10   4.171 ± 0.022  ns/op
CountLeadingZerosOperationBenchmark.baseline                                            0x232323232323232323232323232323  avgt   10   4.216 ± 0.045  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23  avgt   10  10.929 ± 0.174  ns/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10  11.531 ± 0.203  ns/op
CountLeadingZerosOperationBenchmark.executeOperation    0x23232323232323232323232323232323232323232323232323232323232323  avgt   10  10.970 ± 0.101  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                    0x2323232323232323232323232323232323232323232323  avgt   10  10.817 ± 0.112  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323  avgt   10  10.820 ± 0.045  ns/op
```

Results for this PR with suggested 1 second warm/iteration time takes **2.5 mins** overall...
```
Benchmark                                                                                                     (bytesHex)  Mode  Cnt   Score   Error  Units
CountLeadingZerosOperationBenchmark.baseline                                                                        0x23  avgt   10   5.088 ± 0.129  ns/op
CountLeadingZerosOperationBenchmark.baseline          0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10   5.194 ± 0.263  ns/op
CountLeadingZerosOperationBenchmark.baseline            0x23232323232323232323232323232323232323232323232323232323232323  avgt   10   5.061 ± 0.107  ns/op
CountLeadingZerosOperationBenchmark.baseline                            0x2323232323232323232323232323232323232323232323  avgt   10   5.088 ± 0.297  ns/op
CountLeadingZerosOperationBenchmark.baseline                                            0x232323232323232323232323232323  avgt   10   4.990 ± 0.008  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23  avgt   10  12.994 ± 0.235  ns/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10  13.779 ± 0.377  ns/op
CountLeadingZerosOperationBenchmark.executeOperation    0x23232323232323232323232323232323232323232323232323232323232323  avgt   10  15.685 ± 3.964  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                    0x2323232323232323232323232323232323232323232323  avgt   10  13.070 ± 0.315  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323  avgt   10  13.210 ± 0.422  ns/op
```

We seem to get most of the accuracy at an order of magnitude faster overall benchmark duration. Maybe we can go for 2 seconds to be on the safe side?

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
lu-pinto: I don't think we need that many. This all depends on what machine you run them on. So for a headless machine with nothing else running, which is what we should be using for getting reliable results, I don't get much variance at all. So I feel 2 is the lowest for getting a control run, and we can lower even more the time it takes to get results:
```
# Fork: 1 of 5
# Warmup Iteration   1: Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
4.392 ns/op
Iteration   1: 4.354 ns/op
Iteration   2: 4.343 ns/opING [50s]
Iteration   3: 4.349 ns/opING [51s]
Iteration   4: 4.366 ns/opING [52s]
Iteration   5: 4.372 ns/opING [53s]

# Run progress: 2.00% complete, ETA 00:05:56
# Fork: 2 of 5
# Warmup Iteration   1: Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
4.404 ns/op
Iteration   1: 4.349 ns/op
Iteration   2: 4.348 ns/opING [57s]
Iteration   3: 4.349 ns/opING [58s]
Iteration   4: 4.344 ns/opING [59s]
Iteration   5: 4.345 ns/opING [1m]

# Run progress: 4.00% complete, ETA 00:05:47
# Fork: 3 of 5
# Warmup Iteration   1: Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
4.385 ns/op
Iteration   1: 4.337 ns/op
Iteration   2: 4.329 ns/opING [1m 5s]
Iteration   3: 4.330 ns/opING [1m 6s]
Iteration   4: 4.329 ns/opING [1m 7s]
Iteration   5: 4.322 ns/opING [1m 8s]

# Run progress: 6.00% complete, ETA 00:05:40
# Fork: 4 of 5
# Warmup Iteration   1: Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
4.407 ns/op
Iteration   1: 4.340 ns/op
Iteration   2: 4.330 ns/opING [1m 12s]
Iteration   3: 4.324 ns/opING [1m 13s]
Iteration   4: 4.335 ns/opING [1m 14s]
Iteration   5: 4.335 ns/opING [1m 15s]

# Run progress: 8.00% complete, ETA 00:05:32
# Fork: 5 of 5
# Warmup Iteration   1: Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
4.409 ns/op
Iteration   1: 4.350 ns/op
Iteration   2: 4.341 ns/opING [1m 19s]
Iteration   3: 4.340 ns/opING [1m 20s]
Iteration   4: 4.341 ns/opING [1m 21s]
Iteration   5: 4.340 ns/opING [1m 22s]
```

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
lu-pinto: yes the example you showed is loop unrolling optimizations, basically you unroll the loop to do less iterations and more compute in less iterations. For instance the example you have there turns into something like:
```
for(int i=0; i < operations; i+=4) {
   s += x + y + x + y + x + y + x + y;
}
```
I don't think that's the case in the benchmark, and the proof is also that the results without the loop are not hugely different from without it.

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/Keccak256Benchmark.java
lu-pinto: Because invocations of method under test takes roughly 50x longer than of CLZ

lu-pinto:
> We seem to get most of the accuracy at an order of magnitude faster overall benchmark duration. Maybe we can go for 2 seconds to be on the safe side?

Good find! I wasn't aware defaults run for that long. Absolutely 1sec is more than enough for warmup as we are well above the JIT threshold for everything (framework and prod code). We can go even lower to half a sec iterations actually.
Where did you run the benchmarks? Think this might be because you're running on your laptop. It's pretty flat to me so lowering it seems a good call. I'll make the change.

1sec iterations:
```
Benchmark                                                                                                     (bytesHex)  Mode  Cnt   Score   Error  Units
CountLeadingZerosOperationBenchmark.baseline                                                                        0x23  avgt   25   4.342 ± 0.009  ns/op
CountLeadingZerosOperationBenchmark.baseline          0x2323232323232323232323232323232323232323232323232323232323232323  avgt   25   4.341 ± 0.010  ns/op
CountLeadingZerosOperationBenchmark.baseline            0x23232323232323232323232323232323232323232323232323232323232323  avgt   25   4.950 ± 0.931  ns/op
CountLeadingZerosOperationBenchmark.baseline                            0x2323232323232323232323232323232323232323232323  avgt   25   4.333 ± 0.011  ns/op
CountLeadingZerosOperationBenchmark.baseline                                            0x232323232323232323232323232323  avgt   25   4.333 ± 0.005  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23  avgt   25  12.472 ± 0.290  ns/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323  avgt   25  14.424 ± 0.305  ns/op
CountLeadingZerosOperationBenchmark.executeOperation    0x23232323232323232323232323232323232323232323232323232323232323  avgt   25  12.389 ± 0.238  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                    0x2323232323232323232323232323232323232323232323  avgt   25  12.461 ± 0.319  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323  avgt   25  12.380 ± 0.216  ns/op

```

half a sec iterations with some warmup tweaks finishes in ~1.5min:
```
Benchmark                                                                                                     (bytesHex)  Mode  Cnt   Score   Error  Units
CountLeadingZerosOperationBenchmark.baseline                                                                        0x23  avgt   10   4.333 ± 0.016  ns/op
CountLeadingZerosOperationBenchmark.baseline          0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10   4.358 ± 0.026  ns/op
CountLeadingZerosOperationBenchmark.baseline            0x23232323232323232323232323232323232323232323232323232323232323  avgt   10   4.338 ± 0.010  ns/op
CountLeadingZerosOperationBenchmark.baseline                            0x2323232323232323232323232323232323232323232323  avgt   10   4.642 ± 0.502  ns/op
CountLeadingZerosOperationBenchmark.baseline                                            0x232323232323232323232323232323  avgt   10   4.339 ± 0.022  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23  avgt   10  12.353 ± 0.603  ns/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10  14.283 ± 0.428  ns/op
CountLeadingZerosOperationBenchmark.executeOperation    0x23232323232323232323232323232323232323232323232323232323232323  avgt   10  12.371 ± 0.653  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                    0x2323232323232323232323232323232323232323232323  avgt   10  12.301 ± 0.570  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323  avgt   10  12.338 ± 0.576  ns/op
```

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
```diff
@@ -30,15 +30,26 @@
...
 import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
 
 @State(Scope.Thread)
-@OutputTimeUnit(value = TimeUnit.MILLISECONDS)
-public class OneMillionCountLeadingZerosOperationBenchmark {
+@OutputTimeUnit(value = TimeUnit.NANOSECONDS)
+@Measurement(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
```
lu-pinto: warmups were tailored for an amd x86_64 8CPU ~3GHz

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/Keccak256Benchmark.java
```diff
@@ -21,15 +21,26 @@
...
 import org.openjdk.jmh.annotations.Param;
 import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
 
 @State(Scope.Thread)
-@OutputTimeUnit(value = TimeUnit.MILLISECONDS)
-public class Hashing20kKeccak256Benchmark {
+@Warmup(iterations = 8, time = 500, timeUnit = TimeUnit.MILLISECONDS)
```
lu-pinto: warmups were tailored for an amd x86_64 8CPU ~3GHz

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
lu-pinto: IMO you don't care how you compute the baseline. You just need to know that you have to subtract it

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
siladu: That example was a red herring then.

> the results without the loop are not hugely different from without it.

my main point is what I said in the other thread: if the results are the same without it, why is it needed?

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
siladu: Yeh I can see that point of view too. It's not a dealbreaker, happy to keep it simple and can revisit if comparing baselines across opcodes ever trips us up 👍

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
siladu: If something odd happens in one fork, will that still show up as a high error in the final results? 
Only using 2 forks might skew the results more than 3+ and might not be clear which fork is the "correct" one.

But I guess if we can decide on an acceptable amount of variance, then we can rerun if it doesn't meet that?

siladu: btw, I have noticed some variance on cloud machines, e.g. there's an AWS session agent that has to run so it's not as clean as an on-prem run.

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/Keccak256Benchmark.java
siladu: Feel like we could just adjust the JMH settings like we're doing already, rather than introduce another variable? Unless there's some domain-specific benefit to having the inner loop?

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
```diff
@@ -78,11 +89,23 @@ public void setUp() {
...
+    for (int i = 0; i < OPERATIONS_PER_INVOCATION; i++) {
       frame.pushStackItem(bytes);
       CountLeadingZerosOperation.staticOperation(frame);
       frame.popStackItem();
     }
   }
+
+  @Benchmark
+  @OperationsPerInvocation(OPERATIONS_PER_INVOCATION)
+  @Warmup(iterations = 1, time = 500, timeUnit = TimeUnit.MILLISECONDS)
```
siladu: Seems odd to me to warmup the baseline differently to the test?

Overall, to keep it simple to reason about, I'd rather we choose sufficient warmup settings to cover all cases. 
If test duration is a major issue between different opcodes then could maybe have a slow and a fast category?

I feel like tailoring every opcode precisely to save a few seconds is going to make this exercise onerous for most devs. Tailoring the baseline differently to the test feels too precise IMO.

siladu: If the overall suite ends up taking many minutes/hours anyway, I don't think there's much to be gained from this level of precision.

siladu:
Using m6a.2xlarge benchmark machine, I benched 2 seconds and 1 second with simplified warmup/measurement settings...

with these settings at the class level (so applied to baseline as well):
```
@Warmup(iterations = 1, time = 2000, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 5, time = 2000, timeUnit = TimeUnit.MILLISECONDS)
@Fork(2)
```
takes ~5 mins total

```
Benchmark                                                                                                     (bytesHex)  Mode  Cnt   Score   Error  Units
CountLeadingZerosOperationBenchmark.baseline                                                                        0x23  avgt   10   4.335 ± 0.025  ns/op
CountLeadingZerosOperationBenchmark.baseline          0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10   4.335 ± 0.008  ns/op
CountLeadingZerosOperationBenchmark.baseline            0x23232323232323232323232323232323232323232323232323232323232323  avgt   10   4.326 ± 0.006  ns/op
CountLeadingZerosOperationBenchmark.baseline                            0x2323232323232323232323232323232323232323232323  avgt   10   4.337 ± 0.022  ns/op
CountLeadingZerosOperationBenchmark.baseline                                            0x232323232323232323232323232323  avgt   10   4.347 ± 0.012  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23  avgt   10  12.659 ± 0.040  ns/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10  14.495 ± 0.032  ns/op
CountLeadingZerosOperationBenchmark.executeOperation    0x23232323232323232323232323232323232323232323232323232323232323  avgt   10  12.666 ± 0.015  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                    0x2323232323232323232323232323232323232323232323  avgt   10  12.689 ± 0.047  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323  avgt   10  12.681 ± 0.033  ns/op
```

```
@Warmup(iterations = 1, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Fork(2)
```

takes ~3 mins total

```
Benchmark                                                                                                     (bytesHex)  Mode  Cnt   Score   Error  Units
CountLeadingZerosOperationBenchmark.baseline                                                                        0x23  avgt   10   4.666 ± 0.540  ns/op
CountLeadingZerosOperationBenchmark.baseline          0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10   4.346 ± 0.011  ns/op
CountLeadingZerosOperationBenchmark.baseline            0x23232323232323232323232323232323232323232323232323232323232323  avgt   10   4.637 ± 0.470  ns/op
CountLeadingZerosOperationBenchmark.baseline                            0x2323232323232323232323232323232323232323232323  avgt   10   4.350 ± 0.008  ns/op
CountLeadingZerosOperationBenchmark.baseline                                            0x232323232323232323232323232323  avgt   10   4.341 ± 0.011  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                                                0x23  avgt   10  12.913 ± 0.690  ns/op
CountLeadingZerosOperationBenchmark.executeOperation  0x2323232323232323232323232323232323232323232323232323232323232323  avgt   10  14.672 ± 0.482  ns/op
CountLeadingZerosOperationBenchmark.executeOperation    0x23232323232323232323232323232323232323232323232323232323232323  avgt   10  12.772 ± 0.354  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                    0x2323232323232323232323232323232323232323232323  avgt   10  12.919 ± 0.344  ns/op
CountLeadingZerosOperationBenchmark.executeOperation                                    0x232323232323232323232323232323  avgt   10  12.744 ± 0.263  ns/op
```

Two points: 

1. The difference between 1 second and 2 seconds is an 10x better error rate, but do we care about the measurement being ~1 ns off?

2. In terms of a simple "1 second for everything" versus your precise settings, I agree that the difference in results is negligible and it runs in 1.5 mins instead of 3 mins. So then my question is that time saving worth the extra complexity in having to reason about sufficient warmup etc and maybe readjust in the future, e.g. if the implementations under test change?

lu-pinto:
>In terms of a simple "1 second for everything" versus your precise settings, I agree that the difference in results is negligible and it runs in 1.5 mins instead of 3 mins. So then my question is that time saving worth the extra complexity in having to reason about sufficient warmup etc and maybe readjust in the future, e.g. if the implementations under test change?

The idea to standardize around "1 second for everything" is wrong. Benchmarks have different run times so just setting by time everything assuming it's going to be warmed up sufficiently is wrong. Agree though that a simple 1 off warmup annotation in a class is best.
Now, the number of forks is a different story. I'm willing to set `forks = 3` in gradle as I can't see why a particular benchmark would need a different value for what we are doing.

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
lu-pinto: Ok sure, I buy the case for being more relaxed and having more leeway, more so if it's run on a lower spec machine. In the end, warmup time is tied to the machine you are running on because that will impact the number of iterations you get. I can move the warmup to class wide.
But I don't buy this urge to standardize things generally with just 2 examples (CLZ and Keccak). IMO we should be more relaxed on a case by case basis and then tidy up later when there are more benchmarks.

siladu:
> Benchmarks have different run times so just setting by time everything assuming it's going to be warmed up sufficiently is wrong

I agree with this and my assumption based on very little data is 1 second will be enough to warm up any opcode but that assumption might be wrong (it is plenty for CLZ and should be plenty for Keccak). 
My main concern is we go too precise and it ends up being too low as things develop. 
This is also why (in evmtool) I prefer the approach of controlling the number of invocations instead of time, as it is a guaranteed sufficient warmup.
I think you got my point about warmup headroom. 

If it turns out there are wildly different timings per opcode then different settings per class might be prudent to reduce the overall suite duration, but I am hoping we can end up with something a bit simpler.

---

I think the per invocation threshold we'd have to hit for 1 second warmup to be insufficient (with no inner loop) is
```
1 second / 15000 invocations = 66.7 microseconds
```

Worst case for Keccak is well under this:
```
Keccak256Benchmark.executeOperation          512  avgt   10  2017.382 ± 20.057  ns/op
```

Do we have ops that take longer?

ethereum/core/src/jmh/java/org/hyperledger/besu/ethereum/vm/operations/CountLeadingZerosOperationBenchmark.java
siladu: > we should be more relaxed on a case by case basis and then tidy up later when there are more benchmarks

I think we are saying the same thing? 1 second for all is intended to be this relaxed setting (having more warmup than necessary for most cases). Maybe can tighten up later if it makes sense.

siladu: > warmup time is tied to the machine you are running on

Think we should default settings to our chosen benchmark machine: m6a.2xlarge (agreed this with Amez, feel free to challenge)

ahamlat:
> Do we have ops that take longer?

On opcodes, KECCAK256 is only 30 gas, for EXTCODESIZE, EXTCODECOPY, EXTCODEHASH, SLOAD, SSTORE, TLOAD and TSTORE ... the gas cost is at least 100 when storage is warm. So they should take much more time.

Why do we need to have forks ? We do have parallel transaction processing but not sure it is worth adding forks to the benchmarks.

siladu:
@ahamlat 
> Why do we need to have forks ? We do have parallel transaction processing but not sure it is worth adding forks to the benchmarks.

Discussion thread about this here https://github.com/hyperledger/besu/pull/9025#discussion_r2244800331

siladu (approved):
LGTM

PR #9027, opened 2025-08-01
Invalid jump dest not charged

### LINKED ISSUE #8150: Implement gas cost changes to JUMP* for touching code chunks
https://github.com/gballet/go-ethereum/pull/530

The above changes were done in Geth to now charge code chunk loading even if destination of the JUMP is not a valid JUMPDEST opcode.
This will only be needed for the devnet-8, in the meantime some new reference tests might be created.

lu-pinto:
## PR description
This PR is to align with Geth after https://github.com/gballet/go-ethereum/pull/530.

I noticed we are already accounting for the JUMP and JUMPI charges if jumps are valid because the next PC at the start of the interpreter loop will point to the exact code chunk to charge for but not the case where the destination is not JUMPDEST.
I also moved the touch code chunks before getting the operation because charges should come immediately after we know what's the new PC (start of the loop).

## Fixed Issue(s)

fixes #8150

### DIFF
--- a/evm/src/main/java/org/hyperledger/besu/evm/EVM.java
+++ b/evm/src/main/java/org/hyperledger/besu/evm/EVM.java
@@ -214,27 +214,25 @@ public class EVM {
 
     Operation[] operationArray = operations.getOperations();
     while (frame.getState() == MessageFrame.State.CODE_EXECUTING) {
+
+      final Address contractAddress = frame.getContractAddress();
+      long statelessGas =
+          frame
+              .getAccessWitness()
+              .touchCodeChunks(
+                  contractAddress,
+                  frame.wasCreatedInTransaction(contractAddress),
+                  frame.getPC(),
+                  1,
+                  code.length,
+                  frame.getRemainingGas());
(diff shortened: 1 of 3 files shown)


macfarla:
@lu-pinto does this relate to #8150

evm/src/main/java/org/hyperledger/besu/evm/EVM.java
```diff
@@ -214,27 +214,25 @@ public void runToHalt(final MessageFrame frame, final OperationTracer tracing) {
 
     Operation[] operationArray = operations.getOperations();
     while (frame.getState() == MessageFrame.State.CODE_EXECUTING) {
+
+      final Address contractAddress = frame.getContractAddress();
+      long statelessGas =
+          frame
+              .getAccessWitness()
```
matkt: do we have a test for that ?

lu-pinto: I will revert this bit actually. What prompted me to change it was the STOP opcode but I see it's already handled normally if there's no IOOBE exception

PR #9033, opened 2025-08-05
fix: use base 10 for decimal strings in UnsignedLongParameter

Galoretka:
fix: correct base parsing in UnsignedLongParameter constructor

Fix logical error in UnsignedLongParameter string constructor where
decimal numbers without "0x" prefix were incorrectly parsed as
hexadecimal (base 16) instead of decimal (base 10).

- Change base from 16 to 10 for non-hexadecimal strings
- Maintains correct parsing for "0x" prefixed hexadecimal values
- Ensures "123" parses as decimal 123 instead of hex 291


pinges:
This looks like spam or AI. The code works as it is supposed to.

PR #9038, opened 2025-08-07
Skip header validation if the chain is POS

jframe:
Fix header validation when headers go over the pre-merge to post-merge boundary by skipping header validation as it isn't necessary when the pivot block is from the engine API.

This is currently failing because the `TransitionProtocolSchedule` returns a validator for the post-merge blocks with a pre-merge header validator and in this case as the `PostMergeContext.isPostMerge` will still be false until the header or the block is imported.

Tested sync on mainnet with 2 checkpoint and 2 snap sync nodes. All nodes successfully finished syncing.


ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotSelectorFromSafeBlock.java
```diff
@@ -131,15 +131,15 @@ public CompletableFuture<Void> prepareRetry() {
...
+    return new FastSyncState(safeHash, true);
   }
 
   private FastSyncState selectFallbackBlockAsPivot(final Hash fallbackBlockHash) {
     LOG.debug(
         "Safe block not changed in the last {} min, using a previous head block {} as fallback",
         UNCHANGED_PIVOT_BLOCK_FALLBACK_INTERVAL / 60,
         fallbackBlockHash);
-    return new FastSyncState(fallbackBlockHash);
+    return new FastSyncState(fallbackBlockHash, true);
```
pinges, line 142: Not so sure about this, as the fallback is not a safe block, but the head block from the fcu. I guess the source is still trusted, but the source does not have a lot of confidence in that hash.
I think it is still OK to call it trusted, as our CL has identified this as a valid head. If the chain is reorged we will probably fail to finish with this pivot hash and restart the pipeline. We should only check that we are able to get in sync in the case that we have a block as head that was reorged and we receive a pivot hash that is not on that fork. I think we do have that mechanism ...

jframe: If we fail to reach the pivot block we will select another pivot block and start the sync again. The handleFailure in FastSyncDownloader https://github.com/hyperledger/besu/blob/main/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncDownloader.java#L122 will keep retrying if we get a failure.

Nothing has changed with the behaviour around this. The headers would still be valid, just not the fork we were intending. So the existing re-org behaviour will keep working. I think it is still correct to call this trusted as it is from the CL.

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncStateStorage.java
```diff
@@ -55,8 +56,23 @@ public FastSyncState loadState(final BlockHeaderFunctions blockHeaderFunctions)
         return FastSyncState.EMPTY_SYNC_STATE;
       }
       final Bytes rlp = Bytes.wrap(Files.toByteArray(pivotBlockHeaderFile));
-      return new FastSyncState(
-          BlockHeader.readFrom(new BytesValueRLPInput(rlp, false), blockHeaderFunctions));
+      final BytesValueRLPInput input = new BytesValueRLPInput(rlp, false);
+
+      if (isNewFormat(input)) {
```
Gabriel-Trintinalia: isNewFormat is not very a descriptive name. I suggest isVersionedFormat

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncStateStorage.java
```diff
@@ -73,11 +89,28 @@ public void storeState(final FastSyncState state) {
...
   }
+
+  /**
+   * Determines whether the RLP input follows the new format.
+   *
+   * @param input The RLP input to check
+   * @return true if the input is in the new format, false if it's in the old/legacy format
+   */
+  private boolean isNewFormat(final BytesValueRLPInput input) {
+    input.enterList();
```
Gabriel-Trintinalia: enterList returns the list size. Another way possible way is to check the list size

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncStateStorage.java
jframe: That's a better name, will update it

ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncStateStorage.java
jframe: changed to do that. It's less flexible, so might have update that in the future. But at the moment it's a better check. And we can probably remove this check in near future as this is only useful for restarts during the initial sync.

Gabriel-Trintinalia (commented):
Looking good to me. Few comments

Gabriel-Trintinalia (approved):
lgtm

